PR #32, opened 2018-02-03
Adding installation target in Makefile [jed/make-install]

### LINKED ISSUE #28: make install targets
Install to prefix, recognize `DESTDIR`, install `ceed.h` header plus library. Also install Fortran components if applicable (cf. #19).

jedbrown:
resolves #28


Makefile
```diff
@@ -120,13 +132,21 @@ prove : $(tests) $(examples)
...
-ceed.pc : ceed.pc.template
-	@sed 's:%prefix%:$(abspath .):' $< > $@
+$(ceed.pc) : pkgconfig-prefix = $(abspath .)
+$(OBJDIR)/ceed.pc : pkgconfig-prefix = $(prefix)
+.INTERMEDIATE: $(OBJDIR)/ceed.pc
+%/ceed.pc : ceed.pc.template | $$(@D)/.DIR
+	@sed "s:%prefix%:$(pkgconfig-prefix):" $< > $@
+
+install : $(libceed) $(OBJDIR)/ceed.pc
+	$(INSTALL_DATA) -D -t "$(DESTDIR)$(includedir)/" include/ceed.h
```
tzanio: The BSD version of `install` on my Mac doesn't have these options (and fails). Isn't it simple to just copy these files?

jedbrown: Permissions need to be set (in case umask is not 022 in the build environment) on the installed objects without race conditions. I pushed 1e25a746b1a31e71aaccbb0332aa28240dc5d54d which I have now tested on OSX.

tzanio:
Thanks @jedbrown!

Can you mention this also in README and maybe document `$(DESTDIR)`?

PR #41, opened 2018-03-08
Active/passive fields

jedbrown:
This is a placeholder for conversation about this in-progress branch.  Look at the examples to understand the proposed interface changes.

The backend implementation is incomplete and I would like to reach rough consensus on the interface before propagating to other backends and updating the Fortran interface.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -248,7 +248,7 @@ prove : $(tests) $(examples)
 prv : ;@$(MAKE) $(MFLAGS) V=$(V) prove
 
 alltests := $(tests) $(examples) $(if $(MFEM_DIR),$(mfemexamples)) $(if $(PETSC_DIR),$(petscexamples))
-prove-all : $(alltests)
+prove-all : $(ceed.pc) $(alltests)
 	$(info Testing backends: $(BACKENDS))
 	$(PROVE) $(PROVE_OPTS) --exec 'tests/tap.sh' $(alltests:$(OBJDIR)/%=%)
 
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,11 @@ NDEBUG ?= 1
 LDFLAGS ?=
 UNDERSCORE ?= 1
 
+# MFEM_DIR env variable should point to sibling directory
+ifneq ($(wildcard ../mfem/.*),)
+  MFEM_DIR=../mfem
+endif
+
 # OCCA_DIR env variable should point to OCCA master (github.com/libocca/occa)
 OCCA_DIR ?= ../occa
 
@@ -242,7 +247,7 @@ prove : $(tests) $(examples)
 # run prove target in parallel
 prv : ;@$(MAKE) $(MFLAGS) V=$(V) prove
 
-alltests := $(tests) $(examples) $(mfemexamples) $(if $(PETSC_DIR),$(petscexamples))
+alltests := $(tests) $(examples) $(if $(MFEM_DIR),$(mfemexamples)) $(if $(PETSC_DIR),$(petscexamples))
 prove-all : $(alltests)
 	$(info Testing backends: $(BACKENDS))
 	$(PROVE) $(PROVE_OPTS) --exec 'tests/tap.sh' $(alltests:$(OBJDIR)/%=%)
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ UNDERSCORE ?= 1
 
 # MFEM_DIR env variable should point to sibling directory
 ifneq ($(wildcard ../mfem/.*),)
-  MFEM_DIR=../mfem
+  MFEM_DIR?=../mfem
 endif
 
 # OCCA_DIR env variable should point to OCCA master (github.com/libocca/occa)
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ NVCC = $(CUDA_DIR)/bin/nvcc
 
 # ASAN must be left empty if you don't want to use it
 ASAN ?=
-NDEBUG ?= 1
+NDEBUG ?=
 
 LDFLAGS ?=
 UNDERSCORE ?= 1
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ NVCC = $(CUDA_DIR)/bin/nvcc
 
 # ASAN must be left empty if you don't want to use it
 ASAN ?=
-NDEBUG ?=
+NDEBUG ?= 1
 
 LDFLAGS ?=
 UNDERSCORE ?= 1
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,9 @@ examples  += $(examples.f:examples/ceed/%.f=$(OBJDIR)/%)
 #mfemexamples
 mfemexamples.cpp := $(sort $(wildcard examples/mfem/*.cpp))
 mfemexamples  := $(mfemexamples.cpp:examples/mfem/%.cpp=$(OBJDIR)/mfem-%)
+petscexamples.c := $(sort $(wildcard examples/petsc/*.c))
+petscexamples  := $(petscexamples.c:examples/petsc/%.c=$(OBJDIR)/petsc-%)
+
 # backends/[ref & occa  & magma]
 ref.c      := $(sort $(wildcard backends/ref/*.c))
 template.c := $(sort $(wildcard backends/template/*.c))
@@ -216,6 +219,10 @@ $(OBJDIR)/mfem-% : examples/mfem/%.cpp $(libceed) | $$(@D)/.DIR
 	$(MAKE) -C examples/mfem CEED_DIR=`pwd` $*
 	mv examples/mfem/$* $@
 
+$(OBJDIR)/petsc-% : examples/petsc/%.c $(libceed) | $$(@D)/.DIR
+	$(MAKE) -C examples/petsc CEED_DIR=`pwd` $*
+	mv examples/petsc/$* $@
+
 $(tests) $(examples) : $(libceed)
 $(tests) $(examples) : LDFLAGS += -Wl,-rpath,$(abspath $(LIBDIR)) -L$(LIBDIR)
 
@@ -235,9 +242,10 @@ prove : $(tests) $(examples)
 # run prove target in parallel
 prv : ;@$(MAKE) $(MFLAGS) V=$(V) prove
 
-prove-allexamples : $(tests) $(examples) $(mfemexamples)
+alltests := $(tests) $(examples) $(mfemexamples) $(if $(PETSC_DIR),$(petscexamples))
+prove-all : $(alltests)
 	$(info Testing backends: $(BACKENDS))
-	$(PROVE) $(PROVE_OPTS) --exec 'tests/tap.sh' $(tests:$(OBJDIR)/%=%) $(examples:$(OBJDIR)/%=%) $(mfemexamples:$(OBJDIR)/%=%)
+	$(PROVE) $(PROVE_OPTS) --exec 'tests/tap.sh' $(alltests:$(OBJDIR)/%=%)
 
 examples : $(examples)
 
--- a/backends/magma/ceed-magma.c
+++ b/backends/magma/ceed-magma.c
@@ -590,15 +590,15 @@ static int CeedBasisApply_Magma(CeedBasis basis, CeedTransposeMode tmode,
                                 const CeedScalar *u, CeedScalar *v) {
   int ierr;
   const CeedInt dim = basis->dim;
-  const CeedInt ndof = basis->ndof;
-  const CeedInt nqpt = ndof*CeedPowInt(basis->Q1d, dim);
+  const CeedInt ncomp = basis->ncomp;
+  const CeedInt nqpt = ncomp*CeedPowInt(basis->Q1d, dim);
   const CeedInt add = (tmode == CEED_TRANSPOSE);
 
   CeedDebug("\033[01m[CeedBasisApply_Magma] vsize=%d",
-            ndof*CeedPowInt(basis->P1d, dim));
+            ncomp*CeedPowInt(basis->P1d, dim));
 
   if (tmode == CEED_TRANSPOSE) {
-    const CeedInt vsize = ndof*CeedPowInt(basis->P1d, dim);
+    const CeedInt vsize = ncomp*CeedPowInt(basis->P1d, dim);
     for (CeedInt i = 0; i < vsize; i++)
       v[i] = (CeedScalar) 0;
   }
@@ -607,10 +607,10 @@ static int CeedBasisApply_Magma(CeedBasis basis, CeedTransposeMode tmode,
     if (tmode == CEED_TRANSPOSE) {
       P = basis->Q1d; Q = basis->P1d;
     }
-    CeedInt pre = ndof*CeedPowInt(P, dim-1), post = 1;
-    CeedScalar tmp[2][ndof*Q*CeedPowInt(P>Q?P:Q, dim-1)];
+    CeedInt pre = ncomp*CeedPowInt(P, dim-1), post = 1;
+    CeedScalar tmp[2][ncomp*Q*CeedPowInt(P>Q?P:Q, dim-1)];
     CeedDebug("\033[01m[CeedBasisApply_Magma] tmpsize = %d",
-              ndof*Q*CeedPowInt(P>Q?P:Q, dim-1));
+              ncomp*Q*CeedPowInt(P>Q?P:Q, dim-1));
     for (CeedInt d=0; d<dim; d++) {
       ierr = CeedTensorContract_Magma(basis->ceed, pre, P, post, Q, basis->interp1d,
                                       tmode, add&&(d==dim-1),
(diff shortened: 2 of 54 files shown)


ceed-fortran.c
```diff
@@ -320,27 +319,26 @@ static int CeedQFunction_n = 0;
...
 
-static int CeedQFunctionFortranStub(void *ctx, void *qdata, int nq,
+static int CeedQFunctionFortranStub(void *ctx, int nq,
                                     const CeedScalar *const *u, CeedScalar *const *v) {
   struct fContext *fctx = ctx;
   int ierr;
 
   CeedScalar ctx_=1.0;
-  fctx->f((void*)&ctx_, qdata, &nq, u[0], u[1], u[4], v[0], v[1], &ierr);
+  fctx->f((void*)&ctx_, &nq, u[0], u[1], u[2], v[0], v[1], &ierr);
```
jeremylt: I am concerned about this approach handling up to 16 input fields and 16 output fields. I'm not sure what would be a comparable way in Fortran to handle an array of CeedScalar arrays like we do with indata and outdata in the ref implementation in C, but 32 inputs feels like a bad way to do it.

jedbrown: It is possible (but not portably in F77) to pass arrays of pointers. But most Fortran programmers are used to very long argument lists. I know there are CESM components where >30 argument functions are common.

Cc: @thilinarmtb

thilinarmtb: Thanks for the ping @jedbrown ! For the time being this is okay.  We can try
to figure out a way around this if possible.

ceed-fortran.c
```diff
@@ -331,16 +337,26 @@ static int CeedQFunctionFortranStub(void *ctx, int nq,
...
+          u[8],u[9],u[10],u[11],u[12],u[13],u[14],u[15],
+          v[0],v[1],v[2],v[3],v[4],v[5],v[6],v[7],
+          v[8],v[9],v[10],v[11],v[12],v[13],v[14],v[15],&ierr);
   return ierr;
 }
 
 #define fCeedQFunctionCreateInterior \
     FORTRAN_NAME(ceedqfunctioncreateinterior, CEEDQFUNCTIONCREATEINTERIOR)
 void fCeedQFunctionCreateInterior(int* ceed, int* vlength,
                                   void (*f)(void *ctx, int *nq,
```
jedbrown: You can save some insanity by just using `void (*f)(void)` and casting the function pointer. The user doesn't see this declaration, so they don't care if it's a void function pointer or not.

jeremylt: I can work on that. I just got t20 and t30 to work.

jeremylt:
Status Update: ref backend implementation works, as does the Fortran interface. I did not change the occa backend.

Implementation decisions I made that might be worth discussing:
1 - Each QFunction field can only have 1 emode (none, interp, grad, etc)
2 - A restriction or basis action can be applied to both active and passive fields
3 - Currently every field has it's own restriction, but it might be worth avoiding unneeded RestrictionApplys

jeremylt:
@camierjs @thilinarmtb @stomov @v-dobrev @tzanio @jedbrown 
Now that we released v0.2, it feels like a good time to discuss the active-passive refactor.

This is an API change that would affect all backends, so I tagged someone from every backend. Personally, I find this interface for q-functions and operators to be clearer, more flexible, and easier to maintain.

Thoughts? Concerns?

tests/t30-operator.c
```diff
@@ -2,20 +2,22 @@
 #include <ceed.h>
```
jeremylt: This file is the best summary of the proposed changes to q-functions and operators.

camierjs:
The improvement is clearly on the new QFunction API side, I like it.

However, to be able to do so, I understand that we have to specify the in&out layout, which introduces some overhead that concerns me.

There are some new allocation/memcpy/free needs: qdatas, CeedVectors, CeedScalar** in the CeedOperator struct & QFunction setup.

I was about to kill the elem-for-loop wich hurts today the device backend performances, and this improvement brings some additional complications.

We are still using CeedScalar and not CeedVector which means some CeedVectorGet/Restore, that are generating some traffic between the device.

We could perhaps add also some 'standard' CeedOperatorSetField functions to hide the many NULLs in its signature.

jeremylt:
Do you have some thoughts on how we can mitigate the new overhead? I already anticipate that I will be cleaning up my implementation in Ref to handle some inefficiencies.

I am looking to improve the element for loop as well, and I think this structure would help me in Ref batch elements.

WRT a convenience interface for SetField, perhaps something like CeedOperatorSetFieldVector? @jedbrown thoughts? I personally don't mind the extra NULLs.

jeremylt:
Now this branch and #75 have similarly named constants. (CEED_ELEMRESTRICT_IDENTITY and CEED_RESTRICTION_IDENTITY) @jedbrown @thilinarmtb we should probably discuss and make sure that we are not making things more confusing than they need be with our names.

thilinarmtb:
@jeremylt Yes, agreed. We can make them more clear.

I went through the changes and they look fine to me. Basically what we are trying
here is to be flexible with the fields we can apply CeedOperator, it that right?

I tried building this branch in my laptop and I am getting an error related to OccaOperator.
I am attaching the error log here. I am using gcc version 7.3.0 and building a033851.

[error.log](https://github.com/CEED/libCEED/files/2049853/error.log)

Is there a specific OCCA version I should be using?

thilinarmtb:
Ah, I see. I tried with OCCA back-end turned off.

Everything looks fine. My only concern with Fortran interface is that the argument list
is too long (16). Is it possible that this can grow in future?

jeremylt:
Talking to @jedbrown, the length of the argument was a bit arbitrary but designed to be bigger than we would ever need. We could trim it down in the Fortran interface but leave the possibility for expansion in the C code.

jeremylt:
@camierjs, I added some default OperatorSetField entries (CEED_RESTRICTION_IDENTITY, CEED_BASIS_COLOCATED, CEED_QDATA_NONE) that Ref interprets as NULL. I think this addresses your concern while making the examples and user code clearer.

ceed-qfunction.c
```diff
@@ -92,6 +92,7 @@ int CeedQFunctionAddInput(CeedQFunction qf, const char *fieldname,
 
 int CeedQFunctionAddOutput(CeedQFunction qf, const char *fieldname,
                           CeedInt ncomp, CeedEvalMode emode) {
+  if (emode == CEED_EVAL_WEIGHT) return CeedError(qf->ceed, 1, "Cannot create qfunction output with CEED_EVAL_WEIGHT");
```
jedbrown: Looks like this commit only does the first part of the commit message.

include/ceed.h
```diff
@@ -143,6 +143,10 @@ CEED_EXTERN CeedRequest *const CEED_REQUEST_IMMEDIATE;
 CEED_EXTERN CeedRequest *const CEED_REQUEST_ORDERED;
 CEED_EXTERN int CeedRequestWait(CeedRequest *req);
 
+CEED_EXTERN CeedElemRestriction CEED_RESTRICTION_IDENTITY;
+CEED_EXTERN CeedBasis CEED_BASIS_COLOCATED;
+CEED_EXTERN CeedVector CEED_QDATA_NONE;
```
jedbrown: I think we need a better name because "qdata" is no longer a public concept (I think "none" doesn't quite fit the old meaning either). Perhaps `CEED_VECTOR_ACTIVE` to indicate that this vector will be supplied as an active input or output? The use for weights might need a different name.

Also, we need Doxygen comments for these new values.

jeremylt: Updated

ceed-qfunction.c
jeremylt: Updated

jeremylt:
Update: Ref and MAGMA work with the new API. OCCA doesn't yet.

jeremylt:
@camierjs @thilinarmtb I am working on doing an initial rework of OCCA's operator and qfunction files to support the new API. I have a few questions:

1) It looks like OCCA applies the basis operations to all elements in a batch. If that is so, can I expect the GPU to be able to hold all of the input data for all elements so I can avoid copying each element's data for each qfunction application? I can do a copy of each element's data in and out, but it feels inelegant. 

2) We are using arrays of pointers as qfunction inputs and outputs. I don't see it explicitly in the examples but it looks like that should work fine for OCCA. My question is, can I safely do pointer arithmetic to get to the middle of an array of data for multiple elements safely or are there issues I should be aware of in attempting that?

camierjs:
Thanks for starting the rework: I can help you also.

1. I prepared the application to all elems, but it is not turned ‘on’. However I like to have all the data to be on the GPU, so if it avoids us some copies that's what we should go for.

2. Pointer arithmetic can be done in the qfunctions under the for-loops and not before. That is a limitation, these for-loops are transformed to kernels and all arithmetic done before are lost.

jedbrown:
@camierjs Can you help with completing the qfunction apply changes? What is the best way to move this forward? Do we need to discuss further here, in [gitter](https://gitter.im/CEED), or in a call/screenshare?

camierjs:
Ok, I'll switch to it. I'll reach you if I need some info.

jeremylt:
@camierjs I just pushed my most recent work on OpApply. Right now I have the restrictions and basis action applied to all input data and pointers to that qdata as well as the output qdata saved to indata and outdata. I am a bit murky on how best to link up to the qfunction. right now I am wrapping up the code for output basis action and restriction.

jeremylt:
@camierjs It looks like your branch is pretty close to ready (or ready?). Is that right? @jedbrown and I were hoping to merge active-passive soon so we can finish the new optimized CPU backend before ICOSAHOM18.

camierjs:
Yes Jeremy, I have it working.
I had to revert a little and not apply all the basis elems, I'll see if I can re-enable it.
The OpenCL is also not working because of the new parser.

jeremylt:
Thanks for the update and the help!

jeremylt:
@camierjs The MFEM BPs work now, but they are not working with occa (there was only an okl file for BP1). Here is what I get when I run against `/cpu/occa`:

`[jeremy@kawauso mfem]$ ./bp1 -ceed /cpu/occa
Options used:
   -ceed /cpu/occa
   --mesh /home/jeremy/dev/mfem/data/star.mesh
   --order 1
   --visualization
Number of finite element unknowns: 20801
Computing the quadrature data for the mass operator ... done.
   Iteration :   0  (B r, r) = 0.000892612 ...
L2 projection error: 1.97609
`

And with `/cpu/self/ref`:

`[jeremy@kawauso mfem]$ ./bp1 -ceed /cpu/self/ref
Options used:
   -ceed /cpu/self/ref
   --mesh /home/jeremy/dev/mfem/data/star.mesh
   --order 1
   --visualization
Number of finite element unknowns: 20801
Computing the quadrature data for the mass operator ... done.
   Iteration :   0  (B r, r) = 0.000892612 ...
   Iteration :  18  (B r, r) = 6.77949e-16
Average reduction factor = 0.460626
L2 projection error: 7.34603e-06
`

jeremylt:
@jedbrown @camierjs @thilinarmtb 

Ok, summary of this branch to date: 

1) jed/active-passive now has all updates from master
2) All tests pass for cpu/self/* and cpu/occa backends, to include the CEED ex1
3) MFEM BPs work with Ref, Opt, Tmpl
4) OCCA has problem with MFEM BP1 - @camierjs see above post
5) There is no OCCA .okl file for MFEM BP3
6) I haven't checked MFEM BPs with MAGMA
7) Nek5k BPs have not been updated, I would like to incorporate the work in PR #75 so the branches won't conflict - @thilinarmtb can I work with you to adjust the Nek5k BPs?
8) PETSc example has not been updated

backends/ref/ceed-ref-operator.c
```diff
@@ -18,93 +18,392 @@
 #include <string.h>
 #include "ceed-ref.h"
 
+CeedElemRestriction CEED_RESTRICTION_IDENTITY = NULL;
```
jeremylt: I think this is being seen by other backends when they are built, which is not my intention.

backends/magma/ceed-magma.c
```diff
@@ -700,95 +712,378 @@ static int CeedOperatorDestroy_Magma(CeedOperator op) {
...
+                                       opmagma->numqin, qf->numoutputfields, Q); CeedChk(ierr);
+
+  op->setupdone = 1;
+
   return 0;
 }
 
-static int CeedOperatorGetQData_Magma(CeedOperator op, CeedVector *qdata) {
-  CeedOperator_Magma *impl = op->data;
+static int CeedOperatorApply_Magma(CeedOperator op, CeedVector invec,
```
jeremylt: The previous MAGMA backend was a near copy of the Ref backend. I am not sure why OpApply is not working for MAGMA right now.

jeremylt:
Status:

1) Tests
a) /cpu/self/* pass
b) /*/occa pass (inconsistent however)
c) /gpu/magma pass
2) MFEM BPs
a) /cpu/self/* pass
b) /*/occa issues
c) /gpu/magma not yet tested
3) Nek BPs - WIP, builds but incorrect output
a) 
b) 
c) 
4) PETSc BP - WIP, untested
a) 
b) 
c)

ceed-operator.c
```diff
@@ -63,39 +57,96 @@ int CeedOperatorCreate(Ceed ceed, CeedElemRestriction r, CeedBasis b,
...
+int CeedOperatorApply(CeedOperator op, CeedVector in,
+                      CeedVector out, CeedRequest *request) {
   int ierr;
+  Ceed ceed = op->ceed;
+  CeedQFunction qf = op->qf;
 
-  if (!op->GetQData)
-    return CeedError(op->ceed, 1, "Backend does not support OperatorGetQData");
-  ierr = op->GetQData(op, qdata); CeedChk(ierr);
+  if (op->nfields == 0) return CeedError(ceed, 1, "No operator fields set");
```
jeremylt: I added theses checks for common possible mistakes in setup of Operator fields

Makefile
```diff
@@ -235,9 +242,10 @@ prove : $(tests) $(examples)
 # run prove target in parallel
 prv : ;@$(MAKE) $(MFLAGS) V=$(V) prove
 
-prove-allexamples : $(tests) $(examples) $(mfemexamples)
+alltests := $(tests) $(examples) $(mfemexamples) $(if $(PETSC_DIR),$(petscexamples))
```
jeremylt: Do we want to add logic similar to your PETSc logic to check if MFEM is installed?

jedbrown: Yeah, but not urgent.

jeremylt:
```
[jeremy@kawauso mfem]$ ./bp1 -ceed /cpu/self
Options used:
   -ceed /cpu/self
   --mesh /home/jeremy/dev/mfem/data/star.mesh
   --order 1
   --visualization
   --no-test
Number of finite element unknowns: 20801
   Iteration :   0  (B r, r) = 0.000892612 ...
   Iteration :  18  (B r, r) = 6.77949e-16
Average reduction factor = 0.460626
L2 projection error: 7.34603e-06
```

```
[jeremy@kawauso mfem]$ ./bp1 -ceed /cpu/occa
Options used:
   -ceed /cpu/occa
   --mesh /home/jeremy/dev/mfem/data/star.mesh
   --order 1
   --visualization
   --no-test
Number of finite element unknowns: 20801
   Iteration :   0  (B r, r) = 0.000892612 ...
L2 projection error: 1.97609
```

```
[jeremy@kawauso mfem]$ ./bp1 -ceed /omp/occa
Options used:
   -ceed /omp/occa
   --mesh /home/jeremy/dev/mfem/data/star.mesh
   --order 1
   --visualization
   --no-test
Number of finite element unknowns: 20801
   Iteration :   0  (B r, r) = 0.000892612 ...
L2 projection error: 1.97609
[jeremy@kawauso mfem]$ 
```

I fixed an issue where I was failing to zero outputs prior to writing to them. Now OCCA OpApply does not diverge to `nan` but still stops early in MFEM.

jeremylt:
`make prove-all` has a bug - the .okl files are not detected for the MFEM BPs.

Makefile
jeremylt: Added, but discovered problem with OCCA

jeremylt: Namely, OCCA is having trouble finding the .okl

jedbrown:
It should locate those files using the `__FILE__` argument. Either the mfem makefile isn't providing a full path or that isn't working.

jeremylt:
@steven-roberts, this API change will merge soon, impacting your work on the CUDA backend.

Steven-Roberts:
About how soon approximately?  From a quick look at this branch, it seems like the operator and elem restriction will need some upgrading but the other files keep the same api.  Is that a correct assessment?

jeremylt:
I was hoping for last week, but we hit a couple of small bugs with the Nek examples and a couple of OCCA files. It should be within a week, I hope.

jeremylt:
And yes, it's a change for how operators and qfunctions handle inputs and outputs. 'qdata' is dropped in favor of explicit handling of additional fields.

jeremylt:
I think this branch is finally ready! 

Status 

- Backends
-- `/cpu/self` - Working, all tests pass
-- OCCA - Working, all tests pass
-- MAGMA - Working, all tests pass
- Examples
-- Ceed Example - Working
-- MFEM - Working
-- Nek - WIP, don't want to conflict/duplicate PR #92
-- PETSc - Working

jedbrown:
:beers: are in order, @jeremylt

PR #43, opened 2018-03-17
make: set -install_name using @rpath so library is relocatable

jedbrown:
Uses @v-dobrev's suggestion to use `@rpath` instead of `install_name_tool`


tzanio:
Thanks guys!

PR #45, opened 2018-03-25
[occa] Updated to 1.0 C API

dmed256:
See https://github.com/libocca/occa/releases/tag/v1.0.0-alpha.1


jedbrown:
Thanks. @camierjs Do you want to merge this into your more complicated branch, or can I merge it to 'master' and you can rebase?

jedbrown:
@dmed256 Can you weigh in on OCCA philosophy for installing .okl files? Should we install to `$prefix/lib/ceed/occa/*.okl` and hard-code that absolute path, use `dladdr()` and search on a relative path, or something else? We cannot reference the build directory.

cf. https://github.com/spack/spack/pull/7423#issuecomment-375840822

camierjs:
I'm trying it, but David: i'm hitting an error in src/c/kernel.cpp:118:12: call to member function 'add' is ambiguous:  kArg.add(arg.value.ptr, with some other candidates.

camierjs:
The occa-dev branch is fine for me.

jedbrown:
@camierjs I'm interpreting your response as indication that you will fix #46 for v0.2. Let me know if not.

PR #47, opened 2018-03-26
Use occa gpu backend to run Nek5000 ex1

thilinarmtb:
### TODO
- [x] Update ``run-nek-example.sh`` and see if this is working with MPI
- [x] Undo the changes to Makefile

I will test with the above TODOs and if it works fine, we can merge this.

### DIFF
--- a/examples/nek5000/run-nek-example.sh
+++ b/examples/nek5000/run-nek-example.sh
@@ -26,8 +26,6 @@ if [[ $# -ne 4 ]]; then
   exit 1
 fi
 
-export LD_LIBRARY_PATH=`cd ../../lib; pwd`:${LD_LIBRARY_PATH}
-
 ex=$1
 np=$2
 rea=$3


jedbrown:
@camierjs said previously (https://github.com/CEED/libCEED/pull/42#issuecomment-374709550) that he planned to rebase 'occa-dev'. What is the plan now? Perhaps this branch could be rebased onto current 'occa-dev' and then (more cleanly) onto a forthcoming rebased/squashed 'occa-dev'?

thilinarmtb:
I couldn't get the MPI to work with occa backend but I left the `run-nek-examples.sh` script
it as it is since the CPU  backend can work with MPI. I am okay with merging this as it is.
@camierjs

v-dobrev:
I'm switching the target branch of this PR to be `master`.

v-dobrev:
I switched the target branch, so we can merge this after `occa-dev` is merged.

examples/nek5000/make-nek-examples.sh
```diff
@@ -62,7 +62,7 @@ sed -i.bak -e "s|^#FC=.*|FC=\"$FC\"|" \
     -e "s|^#CC=.*|CC=\"$CC\"|" \
     -e "s|^#SOURCE_ROOT=.*|SOURCE_ROOT=\"$NEK5K_DIR\"|" \
     -e "s|^#FFLAGS=.*|FFLAGS+=\"-I../../include\"|" \
-    -e "s|^#USR_LFLAGS+=.*|USR_LFLAGS+=\"-L../../lib -lceed -fopenmp -L/$OCCA_DIR/lib -locca -lm -lcuda\"|" makenek
```
jedbrown: @thilinarmtb It should only be necessary to link the (shared) libraries that you call directly. This works for me everywhere that I have tested.

examples/nek5000/ex1.usr
```diff
@@ -871,7 +885,7 @@ c-----------------------------------------------------------------------
       integer ceed,err
 
 c     Init ceed library
-      call ceedinit('/cpu/self'//char(0),ceed,err)
+      call ceedinit('/gpu/occa'//char(0),ceed,err)
```
jedbrown: How can we make this a run-time option? (See also the email discussion with @tzanio.) It should be possible to build and run this example even without a CUDA device, but here I needed to edit the code to confirm that it works.

examples/nek5000/make-nek-examples.sh
thilinarmtb: I agree, I was being overly cautious.

examples/nek5000/make-nek-examples.sh
```diff
@@ -49,7 +49,13 @@ if [[ "$#" -eq 1 && "$1" -eq "clean" ]]; then
   exit 0
 fi
 
-: ${NEK5K_DIR:=`cd "../../../Nek5000"; pwd`}
+# Set defaults for the parameters
+if [ ! -f $NEK5K_DIR/bin/makenek ]; then
+  NEK5K_DIR:=`cd "../../../Nek5000"; pwd`
```
jedbrown: This isn't bash syntax.
```
$ ./make-nek-examples.sh                                                                                                               
./make-nek-examples.sh: line 54: NEK5K_DIR:=/home/jed/src/Nek5000: No such file or directory                                                                                                  
cp: cannot stat '/bin/makenek': No such file or directory                                                                                                                                     
sed: can't read makenek: No such file or directory                                                                                                                                            
Building example: ex1 ...                                                                                                                                                                     
./make-nek-examples.sh: line 87: ./makenek: No such file or directory                                                                                                                         
  Building ex1 failed. See ex1.build.log for details.             
```

and besides, I prefer the original behavior (providing a default value). Adding a test to error if that doesn't work would be great.

examples/nek5000/make-nek-examples.sh
```diff
@@ -49,7 +49,13 @@ if [[ "$#" -eq 1 && "$1" -eq "clean" ]]; then
...
+fi
+
+# : ${NEK5K_DIR:=`cd "../../../Nek5000"; pwd`}
```
jedbrown: If you're going to replace code, please actually replace it rather than leaving the old thing sitting around commented. In this case, I think this commit should be amended to go back to the old version.

tzanio:
> @tzanio It looks like you made a bunch of documentation changes inside a merge from upstream. (e9854db has two parents, plus its own changes that are not resolving conflicts in the merge.) Was that intentional?

Yes

examples/nek5000/make-nek-examples.sh
tzanio: Several people have made changes to this code, and I wanted to give @thilinarmtb the opportunity to clean it up in the way he prefers.

examples/nek5000/make-nek-examples.sh
tzanio: Hmm... the old version didn't work for me:
```sh
./make-nek-examples.sh: line 54: cd: ../../../Nek5000: No such file or directory
./make-nek-examples.sh: line 54: NEK5K_DIR:=/home/kolev1/ceed/libCEED/examples/nek5000: No such file or directory
Building example: ex1 ...
  Building ex1 failed. See ex1.build.log for details.
```
Does
```
if [ ! -e $NEK5K_DIR/bin/makenek ]; then
```
work for you?

jedbrown: Pretty sure that error message is coming from the "new" version that you just committed. I would write

    if [ -e ${NEK5K_DIR:=`cd "../../../Nek5000"; pwd`} ]; then
        echo "Failed to find Nek5000 at ${NEK5K_DIR}"
        exit 1
    fi

thilinarmtb:
Shall I merge this branch with the current master? I am making some significant
changes in the `run-nek-examples.sh` file

thilinarmtb:
Nevermind :)

camierjs:
Thilina, could you comment out lines 31~35 in your OKL file?
There is a new offset to set if we want the test with the epsilon to be good.

jedbrown:
@thilinarmtb Yes, probably merge 'master' into your branch (again). These repeated merges make for convoluted history and reflect poor communication/coordination, but we're here now and I think it is locally optimal so go ahead.

jedbrown:
@thilinarmtb Or try to merge this branch first and do those changes in a separate, smaller PR.

thilinarmtb:
Okay, I merged this with current master. I will send my changes to `run-nek-examples.sh` in
a separate PR.

thilinarmtb:
Let's get this merged first.

thilinarmtb:
@camierjs; I will un-comment the lines. Please don't merge this till I test it.

thilinarmtb:
@camierjs : If I un-comment, those lines, my error does not reduce and
cg runs till max number of iterations are reached (100 in this case). Any idea why?

camierjs:
hum, looking, but we are just reading and comparing...

camierjs:
you just have the w, u1 & u4?

thilinarmtb:
I actually don't use the setup kernel, I only use `massf` kernel.

camierjs:
well that is even stranger. Why are you not using the setup?
Do you compute yourself the weights?

camierjs:
ah, line 46: const gCeedScalar *w = ((gCeedScalar*)qdata)**+e*Q;**

camierjs:
I can't get your bp1 working, I have an error: Can't open included file 'ceedf.h'

thilinarmtb:
Yes, for the time being. I can improve the example later to do that in
device.

thilinarmtb:
Update your libCEED master branch, this has been fixed. libCEED make
install didn't use to install ceedf.h before.

camierjs:
can you test with the +e*Q? that should help

thilinarmtb:
Will do, I am not near my machine now.

If this is minor, please merge this PR.

camierjs:
I'm testing *wip*

thilinarmtb:
Thanks !!

camierjs:
Well, your weights are ok, I've tried with boxes 3~7.
"+e*Q" won't change much, as they are all the same, but you should put it.
I don't have the issue with the iterations, it's converging in 17 iteration for box7 with a 4.8754e-7 error for a, b & c.

thilinarmtb:
Yep, that sounds about right !!

examples/nek5000/ex1.okl
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+
+// *****************************************************************************
+kernel void massf(void *ctx, void *qdata, const CeedInt e, CeedInt Q,
+                 const CeedScalar *u, const CeedScalar *bu,
+                 CeedScalar *v, CeedScalar *bv) {
+  for (CeedInt i=0; i<Q; i++; tile(TILE_SIZE)) {
+    if (i >= Q) continue;
+    const gCeedScalar *w = ((gCeedScalar*)qdata);
```
camierjs: const gCeedScalar *w = ((gCeedScalar*)qdata)+e*Q;

thilinarmtb:
Cool !! Thanks for testing out !!

tzanio:
@jedbrown: any objections to merge this?

examples/nek5000/make-nek-examples.sh
```diff
@@ -67,7 +67,7 @@ sed -i.bak -e "s|^#FC=.*|FC=\"$FC\"|" \
     -e "s|^#CC=.*|CC=\"$CC\"|" \
     -e "s|^#SOURCE_ROOT=.*|SOURCE_ROOT=\"$NEK5K_DIR\"|" \
     -e "s|^#FFLAGS=.*|FFLAGS+=\"-I${CEED_DIR}/include\"|" \
-    -e "s|^#USR_LFLAGS+=.*|USR_LFLAGS+=\"-L${CEED_DIR}/lib -lceed\"|" makenek
+    -e "s|^#USR_LFLAGS+=.*|USR_LFLAGS+=\"-L${CEED_DIR}/lib -Wl,-rpath,${CEED_DIR}/lib -lceed -fopenmp -lm\"|" makenek
```
jedbrown: Did you intentionally re-add the `-fopenmp -lm` that @thilinarmtb removed as an undocumented side-effect of 262a692c69733cca68d74174699138f2bd553ee5? What should it be?

tzanio: No. I'm not sure why we didn't get a conflict -- Thilina removed them, while I was adding rpath (needed on my Linux box)

thilinarmtb: They should be removed (I could build without them) and tzanio fixed already.

jedbrown: Note that your need for RPATH is because `run-nek-example.sh` did not set `LD_LIBRARY_PATH` correctly (for an installed libceed). Now with RPATH, it shouldn't need to set mess with LD_LIBRARY_PATH at all. @thilinarmtb Can we remove that line?

    export LD_LIBRARY_PATH=`cd ../../lib; pwd`:${LD_LIBRARY_PATH}

thilinarmtb: Yes, we can remove that. I was going to fix it in the next PR but I pushed that change anyway.

camierjs:
Thilina, do you got back the right results?

thilinarmtb:
This branch, as it is gives me right results.

tzanio:
I am also able to run all examples with all backends with it

thilinarmtb:
Thanks for merging @jedbrown !! Thanks for helping with testing it @tzanio and @camierjs !!

PR #51, opened 2018-03-28
[occa] dladdr for OKL files from libceed

camierjs:
- basis & restrict OKL files are read from: 1. focca field, 2. occa cache, 3. libceed install path.
- qfunctions OKL file is read from: 1. focca field, 2. occa cache.

### DIFF
--- a/backends/occa/ceed-occa-basis.c
+++ b/backends/occa/ceed-occa-basis.c
@@ -15,12 +15,12 @@
 // testbed platforms, in support of the nation's exascale computing imperative.
 #define CEED_DEBUG_COLOR 249
 #include "ceed-occa.h"
-#include <sys/stat.h>
 
 // *****************************************************************************
 // * buildKernel
 // *****************************************************************************
 static int CeedBasisBuildKernel(CeedBasis basis) {
+  int ierr;
   const Ceed ceed = basis->ceed;
   const Ceed_Occa *ceed_data = ceed->data;
   const occaDevice dev = ceed_data->device;
@@ -60,7 +60,7 @@ static int CeedBasisBuildKernel(CeedBasis basis) {
   // ***************************************************************************
   // OpenCL check for this requirement
   const CeedInt elem_tile_size = (nelem>TILE_SIZE)?TILE_SIZE:nelem;
-  // OCCA+MacOS implementation need that for now
+  // OCCA+MacOS implementation needs that for now (if DeviceID targets a CPU)
   const CeedInt tile_size = ocl?1:elem_tile_size;
   occaPropertiesSet(pKR, "defines/TILE_SIZE", occaInt(tile_size));
   dbg("[CeedBasis][BK] TILE_SIZE=%d",tile_size);
@@ -90,10 +90,21 @@ static int CeedBasisBuildKernel(CeedBasis basis) {
   // Test if we can get file's status, if not revert to occa://ceed/*.okl ******
   struct stat buf;
   if (stat(oklPath, &buf)!=0) {
-    dbg("[CeedElemRestriction][Create] Could NOT stat this OKL file: %s",oklPath);
-    dbg("[CeedElemRestriction][Create] Reverting to occa://ceed/*.okl");
-    strcpy(oklPath,"occa://ceed/ceed-occa-basis.okl");
+    dbg("[CeedBasis][BK] Could NOT stat this OKL file: %s",oklPath);
+    dbg("[CeedBasis][BK] Reverting to occa://ceed/*.okl");
+    // Try to stat ceed-occa-basis.okl in occa cache
+    ierr=sprintf(oklPath,"%s/ceed/libraries/ceed/ceed-occa-basis.okl",
+            ceed_data->occa_cache_dir);
+    if (ierr<0) return CeedError(ceed, 1, "With occa_cache_dir basis");
+    if (stat(oklPath, &buf)!=0) {
+      dbg("[CeedBasis][BK] Could NOT stat in OCCA cache: %s",oklPath);
+      // reverting to libceed_dir
+      ierr=sprintf(oklPath,"%s/okl/ceed-occa-basis.okl",ceed_data->libceed_dir);
+      if (ierr<0) return CeedError(ceed, 1, "With libceed_dir basis");
+    }else
+      strcpy(oklPath,"occa://ceed/ceed-occa-basis.okl");
   }
+  dbg("[CeedBasis][BK] final okl file is %s",oklPath);
   // ***************************************************************************
   data->kZero   = occaDeviceBuildKernel(dev,oklPath,"kZero",pKR);
   data->kInterp = occaDeviceBuildKernel(dev,oklPath,"kInterp",pKR);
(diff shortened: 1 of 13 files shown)


backends/occa/ceed-occa-basis.c
```diff
@@ -90,10 +90,21 @@ static int CeedBasisBuildKernel(CeedBasis basis) {
...
+    ierr=sprintf(oklPath,"%s/ceed/libraries/ceed/ceed-occa-basis.okl",
+            ceed_data->occa_cache_dir);
+    if (ierr<0) return CeedError(ceed, 1, "With occa_cache_dir basis");
+    if (stat(oklPath, &buf)!=0) {
+      dbg("[CeedBasis][BK] Could NOT stat in OCCA cache: %s",oklPath);
+      // reverting to libceed_dir
+      ierr=sprintf(oklPath,"%s/okl/ceed-occa-basis.okl",ceed_data->libceed_dir);
+      if (ierr<0) return CeedError(ceed, 1, "With libceed_dir basis");
+    }else
+      strcpy(oklPath,"occa://ceed/ceed-occa-basis.okl");
```
jedbrown: Can this bundle of code be put into an internal API? I'm concerned that the duplication will lead to inconsistent searching behavior.

backends/occa/ceed-occa.c
```diff
@@ -16,6 +16,45 @@
...
+static int CeedDladdr_Occa(Ceed ceed) {   
+  Dl_info info;
+  Ceed_Occa *data = ceed->data;
+  memset(&info,0,sizeof(info));
+  const int ierr = dladdr((void*)&CeedInit,&info);
+  dbg("[CeedDladdr_Occa] found %s", info.dli_fname);
+  if (ierr==0)
+    return CeedError(ceed, 1, "OCCA backend cannot fetch dladdr");
+  // libceed_dir setup & alloc in our data
+  const char *libceed_dir = (char*)info.dli_fname;
```
jedbrown: Isn't it already `const char *`, so no cast is needed?

backends/occa/ceed-occa.c
```diff
@@ -16,6 +16,45 @@
...
+  const int path_len = strlen(libceed_dir);
+  dbg("[CeedDladdr_Occa] path_len=%d", path_len);
+  CeedCalloc(path_len+1,&data->libceed_dir);
```
jedbrown: Missing error check.

backends/occa/ceed-occa.c
```diff
@@ -139,6 +181,16 @@ static int CeedInit_Occa(const char *resource, Ceed ceed) {
     return CeedError(ceed,1, "OCCA backend failed to use CUDA resource");
   if (ocl && strcmp(occaDeviceMode(data->device), "OpenCL"))
     return CeedError(ceed,1, "OCCA backend failed to use OpenCL resource");
+  // populatind our data struct with libceed_dir
+  CeedDladdr_Occa(ceed);
```
jedbrown: missing error check

backends/occa/ceed-occa.c
```diff
@@ -16,6 +16,45 @@
...
+  if (stat(libceed_dir, &buf)!=0)
+    return CeedError(ceed, 1, "OCCA backend cannot stat %s", libceed_dir);
+  // Now remove libceed.so part to get the path
+  char *last_slash = strrchr(libceed_dir,'/');
+  if (!last_slash)
+    return CeedError(ceed, 1,
+                     "OCCA backend cannot locate libceed_dir from %s",
+                     libceed_dir);
+  // remove from last_slash and push it to our data
+  *last_slash=0;
```
jedbrown: We aren't supposed to modify `libceed_dir`, which is `const`. I think you can just 

    memcpy(data->libceed_dir,libceed_dir,last_slash-libceed_dir);

which will be null-terminated because `CeedCalloc` returns zeroed memory.

tzanio:
That was quick, @camierjs 😃

backends/occa/ceed-occa.c
```diff
@@ -16,6 +16,45 @@
...
+  memcpy(data->libceed_dir,libceed_dir,last_slash-libceed_dir+1);
+  dbg("[CeedDladdr_Occa] data->data->libceed_dir: %s", data->libceed_dir);
+  return 0;
+}
+#else // nothing is done elsewhere
+static int CeedDladdr_Occa(Ceed ceed) {   
+  return CeedError(ceed, 1, "OCCA backend dladdr not implemented");
```
jedbrown: The function is called unconditionally, so this would mean unconditional failure (including on Solaris, which provides `dladdr`). I'd rather either have no alternative (so we get a build error instead of run-time error) or return without doing anything (and check that empty `libceed_dir` does not cause corruption elsewhere).

backends/occa/ceed-occa-basis.c
camierjs: I refactored this part, now basis, restrict & even qfunctions use the same API.

backends/occa/ceed-occa.c
camierjs: done

backends/occa/ceed-occa.c
camierjs: checked

backends/occa/ceed-occa.c
camierjs: check

backends/occa/ceed-occa.c
camierjs: outdated with the refactoring

backends/occa/ceed-occa.c
camierjs: removed the ifdef alternative

camierjs:
Should be ready for a last check

jedbrown:
And if there are no conflicts with upstream, I would strongly prefer that you don't do merge from it. In fact, I would rather not see merges from upstream unless you can write a good commit message explaining why it is necessary.  This is in https://github.com/ceed/libceed/blob/master/CONTRIBUTING.md:

> Please avoid "merging from upstream" (like merging 'master' into your feature branch) unless there is a specific reason to do so, in which case you should explain why in the merge commit. [Rationale](https://lwn.net/Articles/328436/) from [Junio](https://gitster.livejournal.com/42247.html) and [Linus](http://yarchive.net/comp/linux/git_merges_from_upstream.html).

camierjs:
In fact, there was conflicts with the "Several improvements/unification between the libCEED examples", I should have changed the default message, you're right.

jedbrown:
Huh, okay. In the future, could we have OCCA just run the preprocessor? Not because of this commenting (which I still don't understand) but because I think it'll be a necessary part of a single source solution.

Do you mind if I remove your "merge from upstream" from this branch?

camierjs:
The next parser should handle that I think (@dmed256).
Yes, remove it.

jedbrown:
I don't know why github claims there are conflicts (presumably because the same files were touched, albeit in independent parts), but it merges cleanly with the command line and passes tests for me. Thanks!

dmed256:
@jedbrown The new preprocessor should be able to handle these cases [(v1.0 tests)](https://github.com/libocca/occa/blob/master/parser_sandbox/tests/preprocessor.cpp)

jedbrown:
@dmed256 Excellent, thanks.

PR #65, opened 2018-03-31
[RFC] occa: do not occaFree kernels

jedbrown:
Repeated creation and freeing of kernels appears to expose a caching bug in OCCA, leading to SEGV.

This patch is used in spack for libceed@0.2 so that the nek5000 example works with `-ceed /cpu/occa`. Thread: https://github.com/spack/spack/pull/7423#issuecomment-377667200 Cc: @tzanio @v-dobrev 

If this can be fixed in the occa-1.0 release (@dmed256) then this PR should not be merged. If it will not be fixed (or is actually not an OCCA issue) then this should be merged and we will release v0.2.1.

From a performance and cleanliness standpoint, the nek5000 example should persist the CeedOperator between CG iterations. This would also hide the bug. @thilinarmtb

PR #66, opened 2018-03-31
[occa] Reverted occaFree until refactor

dmed256:


jedbrown:
It looks like this just undoes #65 (returning to the same state as 'master'). I don't think #65 should be merged now that the problem has been fixed in OCCA. Did you have something else in mind?

dmed256:
@jedbrown Oh oops, closing

PR #67, opened 2018-03-31
backends: add missing CeedVectorRestoreArray calls

jedbrown:
@camierjs This oddly breaks ceed ex1 for me with `/cpu/occa`. Can you test it?

### DIFF
--- a/backends/occa/ceed-occa-operator.c
+++ b/backends/occa/ceed-occa-operator.c
@@ -116,6 +116,7 @@ static int CeedOperatorApply_Occa(CeedOperator op, CeedVector qdata,
                                   data->BEv,data->etmp); CeedChk(ierr);
   // *************************************************************************
   ierr = CeedVectorRestoreArray(etmp, &Eu); CeedChk(ierr);
+  ierr = CeedVectorRestoreArray(qdata, (CeedScalar**)&qd); CeedChk(ierr);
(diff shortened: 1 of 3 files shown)


backends/occa/ceed-occa-operator.c
```diff
@@ -116,6 +116,7 @@ static int CeedOperatorApply_Occa(CeedOperator op, CeedVector qdata,
                                   data->BEv,data->etmp); CeedChk(ierr);
   // *************************************************************************
   ierr = CeedVectorRestoreArray(etmp, &Eu); CeedChk(ierr);
+  ierr = CeedVectorRestoreArray(qdata, (CeedScalar**)&qd); CeedChk(ierr);
```
jedbrown: If I remove this line, I see correct results with the example, but with it, I get 

```
Exact mesh volume    :  2.3561944901923                             
Computed mesh volume :  0                                            
Volume error         : -2.3561944901923  
```

I have not been successful at debugging the difference, but it looks like a memory corruption issue. The array variable being restored is `qd` and is not used again, but somehow the entries of a different vector are being destroyed.

tzanio:
@jedbrown can you verify this fixes the issue and merge?

jedbrown:
Yes, but I can't think about this before tomorrow evening.

PR #70, opened 2018-04-14
[WIP] Add a Makefile for Nek5000 example(s)

thilinarmtb:
## TODO
- [ ] Update documentation
- [ ] Resolve the issue of working with Nek5000 makefile


examples/nek5000/Makefile
```diff
@@ -0,0 +1,34 @@
...
+CC_EXAMPLES = $(if $(findstring mpi,$(CC)),$(FC),mpicc)
+
+NEK5K_EXAMPLES ?= bp1
+
+.PHONY: all examples clean
+
+all: examples
+
+examples:
+	export NEK5K_DIR=$(NEK5K_DIR) && export CEED_DIR=$(CEED_DIR) && \
```
jedbrown: You can just write `NEK5K_DIR=$(NEK5K_DIR) ... ./make-nek-examples.sh`. It's a bit cleaner than having all the exports.

thilinarmtb: Yep, that is better. Will fix it.

thilinarmtb:
@jedbrown : Yes, I will update the documentation.

@tzanio @jedbrown: I have one suggestion. Shall we use a make target to run the bp1 examples as well? That way both building and running examples will be unified across MFEM, PETSc and Nek5000.

jedbrown:
The PETSc, MFEM, and plain libCEED examples are all `./bp1 -ceed /cpu/ref`. Could that be made to work with Nek?

thilinarmtb:
I don't think so because we have to set a SESSION.NAME file which the nek5000
reads in order to find the mesh. See [here](
https://github.com/CEED/libCEED/blob/master/examples/nek5000/run-nek-example.sh#L91).

(This may not be compulsory, I will have to double check. This is
certainly needed if the user wants to change the mesh Nek5000 runs on).

jedbrown:
And Nek executables can't process that as a command-line argument?

thilinarmtb:
@tzanio @jedbrown : Are you guys okay with a single Makefile for all the examples
in top directory calling there own Makefiles? Please have a look at 78d928e. This is
just an idea, we can polish it more.

Makenek script generates its own makefile and I did not want to put another one inside
`examples/nek5000` directory.

thilinarmtb:
@jedbrown : I confirmed with Paul that Nek5000 does not read mesh files as a run time
parameter.

thilinarmtb:
We should probably rename the PR as well if we decide to use this.

jedbrown:
I would not mind that higher level Makefile. It could even be in `libceed/Makefile` (everything should already be in place except the targets) instead of `libceed/examples/Makefile`.

As for hard-coding the mesh, have you tried Firetran? https://jedbrown.org/files/BrownKnepleySmith-RuntimeExtensibilityAndLibrarizationOfSimulationSoftware-2014.pdf

thilinarmtb:
@jedbrown: Thanks for merging. I will update the documentation in a separate PR !!

PR #73, opened 2018-05-02
Remove redundancies in Nek example

thilinarmtb:
This PR removes some of the redundant Ceed structure creations and clean up
the Nek example.


thilinarmtb:
@jedbrown: Thanks for merging !!

PR #74, opened 2018-05-02
examples/nek5000/run-nek-example.sh: error if -b is not provided

jedbrown:
I would prefer to just make a default like there is for all the other examples, but in lieu of that, here is a patch to error when the user invariably forgets.

It would be even better if we could have the makefile build an executable named something other than bp1, then make bp1 a symlink to the script that checks $0 to decide what to do. That way we'd be able to run `nek5000/bp1 -ceed /cpu/self` just like we can for all the other examples. But that's a more intrusive change.

### DIFF
--- a/examples/nek5000/run-nek-example.sh
+++ b/examples/nek5000/run-nek-example.sh
@@ -79,6 +79,12 @@ while [ $# -gt 0 ]; do
   shift
 done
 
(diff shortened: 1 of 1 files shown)


thilinarmtb:
Looks good to me. Merging. Thanks for the PR !!

PR #75, opened 2018-05-05
[WIP][Nek] Update BP1 and add BP3

thilinarmtb:
Also, added a fix so that E-vectors don't need to provide an indices array.

### Summary
- Updated the libCEED reference implementation
- Updated the Fortran interface
- Updated the Nek5000 example

### TODO
- [ ] If the fix is acceptable, fix it in other backends

### DIFF
--- a/backends/ref/ceed-ref-restriction.c
+++ b/backends/ref/ceed-ref-restriction.c
@@ -26,7 +26,7 @@ static int CeedElemRestrictionApply_Ref(CeedElemRestriction r,
   int ierr;
   const CeedScalar *uu;
   CeedScalar *vv;
-  CeedInt esize = r->nelem*r->elemsize;
+  CeedInt esize = r->nelem*r->elemsize*ncomp;
 
   ierr = CeedVectorGetArrayRead(u, CEED_MEM_HOST, &uu); CeedChk(ierr);
   ierr = CeedVectorGetArray(v, CEED_MEM_HOST, &vv); CeedChk(ierr);
--- a/backends/ref/ceed-ref-restriction.c
+++ b/backends/ref/ceed-ref-restriction.c
@@ -26,14 +26,14 @@ static int CeedElemRestrictionApply_Ref(CeedElemRestriction r,
   int ierr;
   const CeedScalar *uu;
   CeedScalar *vv;
-  CeedInt esize = r->nelem*r->elemsize*ncomp;
+  CeedInt esize = r->nelem*r->elemsize;
 
   ierr = CeedVectorGetArrayRead(u, CEED_MEM_HOST, &uu); CeedChk(ierr);
   ierr = CeedVectorGetArray(v, CEED_MEM_HOST, &vv); CeedChk(ierr);
   if (tmode == CEED_NOTRANSPOSE) {
     // Perform: v = r * u
     if (!impl->indices) {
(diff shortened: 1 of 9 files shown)


thilinarmtb:
@jedbrown Thanks for the reviews !! Have been busy this week, will have a look during the weekend.

thilinarmtb:
@jedbrown : I added a named constant for E-vector identity. Let me know if this looks okay.

backends/ref/ceed-ref-restriction.c
```diff
@@ -26,7 +26,7 @@ static int CeedElemRestrictionApply_Ref(CeedElemRestriction r,
   int ierr;
   const CeedScalar *uu;
   CeedScalar *vv;
-  CeedInt esize = r->nelem*r->elemsize;
+  CeedInt esize = r->nelem*r->elemsize*ncomp;
```
thilinarmtb: @jedbrown : Is this change okay? We should take into account number of
components as well right?

thilinarmtb: Nevermind, I fix this in eb5d317

jeremylt:
It is the way the front end is set up.

`int CeedElemRestrictionApply(CeedElemRestriction r, CeedTransposeMode tmode,
                             CeedInt ncomp, CeedTransposeMode lmode,
                             CeedVector u, CeedVector v, CeedRequest *request) {
  CeedInt m,n;
  int ierr;

  if (tmode == CEED_NOTRANSPOSE) {
    m = r->nelem * r->elemsize * ncomp;
    n = r->ndof * ncomp;
  } else {
    m = r->ndof * ncomp;
    n = r->nelem * r->elemsize * ncomp;
  }
  if (n != u->length)
    return CeedError(r->ceed, 2,
                     "Input vector size %d not compatible with element restriction (%d,%d)",
                     u->length, m, n);
  if (m != v->length)
    return CeedError(r->ceed, 2,
                     "Output vector size %d not compatible with element restriction (%d,%d)",
                     v->length, m, n);
  ierr = r->Apply(r, tmode, ncomp, lmode, u, v, request); CeedChk(ierr);
  return 0;
}`

jeremylt:
@thilinarmtb Once we merge the active-passive update and the new CPU backend, I am going to be running some comparisons to the NEK code in https://github.com/CEED/benchmarks. Will the work in this branch help make that comparison more fair/reasonable?

thilinarmtb:
@jeremylt : We still do a redundant data copy for the Fortran interface here:
https://github.com/CEED/libCEED/blob/master/ceed-fortran.c#L94

I need to fix this; There is alignment problem between Fortran address and the C address.
If you ignore this data copy time, I think the comparison should be fair.

jedbrown:
@thilinarmtb Can we merge this as is (without your planned work to avoid data copies)? You can start a new branch for that after we merge #41.

thilinarmtb:
Hi Jed,

Sure, I am fine merging as it is.

jeremylt:
We should make a conscious choice about names for CEED_ELEMRESTRICT_IDENTITY and CEED_RESTRICTION_IDENTITY before we merge, but everything else looks ready to me.

thilinarmtb:
@jeremylt : Yes, I think they are the same as well. I am fine with either names. So the choice is
yours :)

PR #76, opened 2018-06-01
Factorize grad

jeremylt:
Adds ability to calculate colocated grad1d matrix based upon QR factorization.


ceed-basis.c
```diff
@@ -290,46 +290,138 @@ int CeedQRFactorization(CeedScalar *mat, CeedInt m, CeedInt n) {
...
+      if (mat[i+n*i]<10E-14) {
         v[i] = mat[i+n*i]-mu;
       } else {
         v[i] = -sigma/(mat[i+n*i]+mu);
       }
       b = 2*v[i]*v[i]/(sigma+v[i]*v[i]);
-      for (j=i; j<m; j++) {
-        v[m-1-(j-i)] = v[m-1-(j-i)]/v[i];
+      for (j=i+1; j<m; j++) {
+        v[j] = v[j]/v[i];
```
jedbrown: These changes would be better squashed into the previous commit.

Also, it would be easier to read/maintain with a (private) function that applies an elementary reflector, similar to [xLARF](http://www.netlib.org/lapack/explore-html/d8/d9b/group__double_o_t_h_e_rauxiliary_ga39a2b39f56ce497b218c5f48339ee712.html#ga39a2b39f56ce497b218c5f48339ee712) in LAPACK.

Also note that `xGEQRF` returns the scaling vector "`tau`", thus avoiding needing to recompute it (just extra code; we don't care about performance here). That would be a more useful interface IMO.

Let me know if you'd like to change this or just merge the branch as is? (And sorry this comment is so late.)

jeremylt: I did two rounds of squashing cleanup.

Are you talking about making a common function for lines 316-328, 392-404, and 406-418?

ceed-basis.c
```diff
@@ -282,11 +282,95 @@ int CeedBasisView(CeedBasis basis, FILE *stream) {
...
+  case CEED_RIGHT:
+    // Find w = D v
+    for (j=i; j<n; j++) { // Row j
+      w[j] = mat[i+m*j]; // v[i] = 1
+      for (k=i+1; k<m; k++) { // Column k
+        w[j] += mat[k+m*j]*v[k];
+      }
+    }
+    // Calculate D - b w vtranspose
+    for (j=i; j<n; j++) { // Row j
```
jeremylt: @jedbrown I could compact this a bit to account for similarity between lines 318-329 and 332-345, but I think this is what you were getting at.

ceed-basis.c
```diff
@@ -282,121 +282,66 @@ int CeedBasisView(CeedBasis basis, FILE *stream) {
...
-     }
-    tau[i] = b;
+    CeedScalar norm = sqrt(v[i]*v[i] + sigma); // norm of v[i:m]
+    CeedScalar Rii = -copysign(norm, v[i]);
+    v[i] -= Rii;
+    // norm of v[i:m] after modification above and scaling below
+    //   norm = sqrt(v[i]*v[i] + sigma) / v[i];
+    //   tau = 2 / (norm*norm)
+    tau[i] = 2 * v[i]*v[i] / (v[i]*v[i] + sigma);
+    for (j=i+1; j<m; j++) v[j] /= v[i];
```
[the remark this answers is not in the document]
jeremylt: With an already tridiagonal matrix, I get correct values with wrong signs, which I sort of expected

PR #77, opened 2018-06-01
CeedBasisView for Fortran

jeremylt:
CeedBasisView does not currently work in the Fortran interface, as there is no easy way to switch between Fortan UNITs and C FILEs. This branch proposes to implement CeedBasisView by setting fCeedBasisView to output to stdout. The user would then have to pipe the output to file if desired.

I am not sure if this is a good solution, but it is *a* solution to getting CeedBasisView working in the Fortran interface.


thilinarmtb:
@jeremylt : I am not sure about this either and that is why I left it as a todo.
I will talk to some people here as well.

jeremylt:
Sounds good to me. I wanted to add the ability to check a piece of the Fortran interface that I added for factorizing the grad1d matrix (PR #76) and ran into this, so I just put this in as a basic way to verify that I got the same results from Fortran and C.

thilinarmtb:
@jeremylt : I am fine with this PR. I don't think there is an easy way to translate
between Fortran units and C FILE structures (may be this is impossible). So +1
to merge as it is.

PR #78, opened 2018-06-22
cpu/opt first optimization

jeremylt:
This starts the optimized backend by using a colocated grad for CEED_EVAL_GRAD. This PR needs to follow PR #76.

### TODO (Possibly in second PR)
- [ ] Add Tensor Contract over batches of elements
- [ ] Symmetric Tensor Contract?


jeremylt:
Rebuilding this backend and combining with PR #81

PR #79, opened 2018-06-22
Optimize the reference tensor contraction

Steven-Roberts:
This new version of `CeedTensorContract_Ref` is about 50% faster.  Example 1 has about a 35-40% speedup.

### DIFF
--- a/backends/ref/ceed-ref-basis.c
+++ b/backends/ref/ceed-ref-basis.c
@@ -38,18 +38,14 @@ static int CeedTensorContract_Ref(Ceed ceed,
     }
   }
 
-  const CeedScalar *uP = u;
   for (CeedInt a=0; a<A; a++) {
     for (CeedInt b=0; b<B; b++) {
-      CeedScalar *vP = v + a * J * C;
       for (CeedInt j=0; j<J; j++) {
         CeedScalar tq = t[j*tstride0 + b*tstride1];
         for (CeedInt c=0; c<C; c++) {
-          *vP += tq * uP[c];
-          vP++;
+          v[(a*J+j)*C+c] += tq * u[(a*B+b)*C+c];
         }
       }
-      uP += C;
     }
   }
   return 0;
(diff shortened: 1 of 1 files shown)


backends/ref/ceed-ref-basis.c
```diff
@@ -32,17 +32,24 @@ static int CeedTensorContract_Ref(Ceed ceed,
...
-      }
-      for (CeedInt b=0; b<B; b++) {
+    for (CeedInt b=0; b<B; b++) {
+      CeedScalar *vP = v + a * J * C;
+      for (CeedInt j=0; j<J; j++) {
+        CeedScalar tq = t[j*tstride0 + b*tstride1];
         for (CeedInt c=0; c<C; c++) {
-          v[(a*J+j)*C+c] += t[j*tstride0 + b*tstride1] * u[(a*B+b)*C+c];
+          *vP += tq * uP[c];
+          vP++;
```
jedbrown: This sort of pointer arithmetic is explicitly discouraged by most compiler optimization guides. I doubt that it is beneficial and would rather not change it.

Steven-Roberts:
I have tested on a Mac with clang and profiled with Instruments.  I see similar performance results on Linux compiled with gcc.  The primary source of speedup comes from moving the if statement out of the nested loop.  Making this change should have minimal effect on the readability.  The pointer arithmetic gives a 5-10% percent speedup.  I can understand keeping that as is.

tzanio:
Thanks @Steven-Roberts!

@jedbrown, fyi Steven is an intern working with @camierjs this summer.

jedbrown:
Cool. What tests are you running? I would like to try reproducing. (Information about reproducing performance claims should go in commit messages.)

Steven-Roberts:
Sorry for not including more in the commit message.  I have looked at the run times for ceed/ex1 and mfem/bp1 and mfem/bp3.  I start with the default configuration and then try different meshes and orders.  Ex1 gets about a 35% speedup, bp1 stays about the same, and bp3 about 20% for most configs.

PR #80, opened 2018-06-25
[WIP] New Basis shapes

jeremylt:
I built this branch so we can discuss how we want to include tets, pyramids, wedges, etc so libCEED can handle mixed meshes and eventually composite operators.

With this proposal, I am envisioning that for all non-het shapes the user would simply provide the grad and intep matrices of the appropriate dimension. This would make libCEED most suited for hex dominant meshes. Then backends would have a switch for ELEM_TYPE with hets using tensor contract and all other shapes using mat-vec multiply.

Intended Priority: Post ICOSAHOM18


jeremylt:
Rebuilding due to sheer number of active-passive updates and this being a bad first draft

PR #81, opened 2018-06-25
Update README

jeremylt:
Updated README to specify which /cpu/self backend in sample cl commands to run examples, as there are 2, soon to be 3 /cpu/self backends. I think it makes it a bit clearer to new users what is going on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,28 +72,28 @@ and run:
 # libCEED examples on CPU and GPU
 cd examples/ceed
 make
-./ex1 -ceed /cpu/self
+./ex1 -ceed /cpu/self/ref
 ./ex1 -ceed /gpu/occa
 cd ../..
 
 # MFEM+libCEED examples on CPU and GPU
 cd examples/mfem
 make
-./bp1 -ceed /cpu/self -no-vis
+./bp1 -ceed /cpu/self/ref -no-vis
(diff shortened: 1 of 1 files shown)


jedbrown:
Is the idea also that `/cpu/self` may dispatch to the optimized backend soon? Would we want them to still use the reference backend when they follow the directions in the README?

jeremylt:
I was wondering about that too. If I understand how backends are registered right now, it would use ref before opt because ref is listed in the Makefile before opt, which is necessary as opt dispatches through to ref for everything that isn't a basis action so ref needs to be built first.

The example is pretty small, so it should run quickly either way. I could see it being a good idea to point new users to the opt backend though as it would be the 'recommended' CPU backend once it is online. In that case, this would be best after PR #78.

In any case, I think we should specify a particular backend as we have multiple that start with /cpu/self because it isn't clear to new users which /cpu/self backend runs right now.

jedbrown:
To your first point, Ref will always be available (and created explicitly by Opt), but the question is what `/cpu/self` would default to. I think it should be Opt with fallback to Ref if there are cases it cannot handle. We might need a system of priorities for sorting implementations that match the same length prefix.

jeremylt:
I'm combining this with PR#82

PR #82, opened 2018-06-27
[Opt] Add optimized cpu backend with ability to use colocated grad

jeremylt:
Start of the cpu/opt backend with the ability to use colocated grad.

Also incorporating PR #81, adding backend prioritization so '/cpu/self' will default to using '/cpu/self/opt' before '/cpu/self/ref'.

~~ToDo: Add test for CeedBasisApply with CEED_TRANSPOSE and CEED_GRAD to verify this action of the new backend.~~ t17 added to test CeedBasisApply with CEED_TRANSPOSE and CEED_GRAD.

### DIFF
--- a/tests/t17-basis.c
+++ b/tests/t17-basis.c
@@ -1,0 +1,60 @@
+// Test grad in multiple dimensions
+#include <ceed.h>
+#include <math.h>
+
+static CeedScalar Eval(CeedInt dim, const CeedScalar x[]) {
+  CeedScalar result = tanh(x[0] + 0.1);
+  if (dim > 1) result += atan(x[1] + 0.2);
+  if (dim > 2) result += exp(-(x[2] + 0.3)*(x[2] + 0.3));
+  return result;
+}
+
+int main(int argc, char **argv) {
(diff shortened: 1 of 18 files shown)


tests/t17-basis.c
```diff
@@ -0,0 +1,55 @@
...
+      u[i] = Eval(dim, xx);
+    }
+
+    // Check if u' * grad' * ones = uq' * ones
+    CeedBasisCreateTensorH1Lagrange(ceed, dim, 1, Q, Q, CEED_GAUSS, &bug);
+    CeedBasisApply(bug, CEED_NOTRANSPOSE, CEED_EVAL_GRAD, u, uq);
+    CeedBasisApply(bug, CEED_TRANSPOSE, CEED_EVAL_GRAD, ones, dtposeones);
+    for (CeedInt i=0; i<Qdim; i++) {
+      sum1 += dtposeones[i]*u[i];
+    }    for (CeedInt i=0; i<dim*Qdim; i++) {
```
jedbrown: formatting?

PR #83, opened 2018-06-28
tests: activate backends from Makefile; can also use command line

jedbrown:


jeremylt:
It runs well on my machine.

Is the follow-on to this going to be that small executable that reports version, backends, and compile info? I couldn't find documentation on how to make the .so do that for mac and found a few people insisting it wasn't possible.

Issue #84, opened 2018-06-28
[fortran] truncated __FILE__ line


camierjs:
We have an "Unterminated character constant" error if the `__FILE__` size is greater than ~70 characters in the qfunction fortran files.
Is there a portable "line-truncation" option for the compiler we use?

jedbrown:
I would use free-form where this is a non-issue, but I think the Nek crew wants to religiously stick to F77. @thilinarmtb ?

PR #85, opened 2018-06-28
[occa] active passive

camierjs:
Works with all the tests
- OpenCL works, but the OKL files have to be 'aware' of the new parser's restrictions (@dmed256)
- still have to handle `op` and `no-op` cases (coming from an operator *or not*) in the qfunctions
- `examples/mfem/bp1.hpp`: switch `qfunc` to `qdata` field, but the size is not yet good
- `tests/t20-qfunction-f.f` and `tests/t30-operator-f.f`: shifted left the `__FILE__` to add few extra characters, see #84

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ NVCC = $(CUDA_DIR)/bin/nvcc
 
 # ASAN must be left empty if you don't want to use it
 ASAN ?=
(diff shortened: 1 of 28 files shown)


jeremylt:
It looks good to me. I say we merge the branches and then tackle the remaining lose ends.

Issue #87, opened 2018-07-02
Test OPApply 2x


jeremylt:
I found a bug in my active-passive code for ref that was not covered in our tests. I think we should have a test that covers repeated application of the same operator to help future users who are building backends of their own.

jeremylt:
This could be fixed by #23

jeremylt:
Best satisfied by testing in MFEM such as PR #88. Closing

PR #88, opened 2018-07-02
Add MFEM to Travis

### LINKED ISSUE #23: Test MFEM examples in Travis

jeremylt:
I included running bp1 in MFEM in the Travis configuration. I do not think we want to include /examples/mfem/bp1 in our standard test suite, but including it in Travis would resolve #23 and #87.

Thoughts?

For this to work well, I would need to add something like tap.sh to run the MFEM BPs against all backends.


tzanio:
I think we should definitely test with MFEM on travis -- the build doesn't take long, and that's an important use case for us.

.travis.yml
```diff
@@ -10,7 +10,18 @@ install:
 env:
   - FC=gfortran
 
+before_script:
+  - git clone https://github.com/mfem/mfem.git
+  - cd mfem
+  - make config
+  - make test -j 4
```
tzanio: To speed things up a bit, you an replace `make config; make test -j 4` with `make serial`.

jeremylt:
I changed to `make serial`. I'll add a script to test against all /cpu/self/* backends later today.

jeremylt:
@jedbrown My MFEM makefile could be better, but I think this answers #23. I would still like to write a separate test for #87. Your thoughts?

examples/mfem/tap.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
jedbrown: Can we use the `tap.sh` from above instead of copying it here? A copy means people will need to maintain both, but they never do, thus bugs hit more than once.

jeremylt:
The problem I was having with getting the BACKENDS variable. I am hesitant to include this in the standard tests in the libCEED Makefile, but maybe we can add a `make test-mfem`?

jedbrown:
Yes, better to do it as a test-mfem target in the top-level Makefile.

jeremylt:
Ok @jedbrown. I have added `make prove-mfem`. I would like to eventually add `make test-mfem`, but I think this branch works now.

jedbrown:
@jeremylt How is this revised version?

jeremylt:
I like it. It runs better than my makefile.

PR #90, opened 2018-07-04
Fix debug option in Makefile

Steven-Roberts:
It seems `NDEBUG` can never be set to false due to
https://github.com/CEED/libCEED/blob/a460c33b23cf49899be5782d753f25b389750a7f/Makefile#L27
This change fixes this issue and is consistent with the checking of the `UNDERSCORE` variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,13 +49,17 @@ ifneq ($(filter %xlf %xlf_r,$(FC)),)
   FFLAGS = -qpreprocess -qextname -qpic -MMD
 endif
 
-CFLAGS += $(if $(NDEBUG),-O2 -DNDEBUG=1,-g)
+ifeq ($(NDEBUG), 1)
+	CFLAGS += -O2 -DNDEBUG=1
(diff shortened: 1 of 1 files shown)


jeremylt:
And active-passive is merged.

jedbrown:
Any opinions on the above? @Steven-Roberts @camierjs ? Shall I strip the `NDEBUG` stuff and make a single `OPT` or `OPTFLAGS` (used in `CFLAGS`, `CXXFLAGS`, and `FCFLAGS`) that the user can override on the command line?

@Steven-Roberts I forgot to mention that the existing logic make `NDEBUG` "false" by running `make NDEBUG=` instead of `make NDEBUG=0` or similar.

camierjs:
Sure, we can uniform flags to be able to override them.

Steven-Roberts:
Adding a single `OPT` seems like a good idea to me.

jeremylt:
This looks like it was handled by #102, so I'm closing it, but please reopen if I'm wrong.

PR #93, opened 2018-07-21
Jed/travis opt

jedbrown:
Add caching to speed up MFEM build and test with PETSc (fixed version 3.9.3).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,10 @@ compiler:
   - gcc
   - clang
 
-install:
-  - sudo apt-get install gfortran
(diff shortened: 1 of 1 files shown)


jeremylt:
Does the 5 min Travis build time mean that caching is still not working right?

jeremylt:
It looks good to me now. <3 min. 

I'll play with OCCA and the idea of internal/external tests in a later PR. It should be not too bad to add.

tzanio:
Thanks for doing this @jedbrown!

PR #97, opened 2018-07-21
Non-Tensor Bases

jeremylt:
This is an in progress branch adding simplexes/tets, pyramids, and wedges to libCEED.


jeremylt:
Plan for this branch:
- Rebase
- Cut back to just the interface changes enabling non-tensor bases
- Move functionality for creating interp, grad matrices for not-tensor bases to subsequent PR

jedbrown:
That sounds good to me. As we discussed, tetrahedral bases are nontrivial and harder to classify the alternatives.

jeremylt:
Started, but I want to make some variable naming changes that would touch all backends, so I will hold this branch as is for now.

jeremylt:
Everything we need is built out now, though this has not been set up in any backends. I can set this up for ref/opt easily and make a couple of tests though. I'd like to clear this out of the list of PRs soonish.

jedbrown:
That sounds good. One simple test would be to choose triangles or tets and write an example with a fixed order of accuracy, that way you don't need to wade into efficient quadratures and stable bases at arbitrary order.

jeremylt:
I think this is ready to go, or really close. I left room to add `t31*` tests for non-tensor hypercubes, `t33*' for pyramids, and 't34*' for prisms. CodeCov went down for the ref backend because I didn't add a test for basis apply in transpose on non tensor bases and I added lines that don't get hit giving error messages for the eval modes not implemented for basis apply.

include/ceed-impl.h
```diff
@@ -95,10 +97,13 @@ struct CeedBasis_private {
                CeedScalar *);
   int (*Destroy)(CeedBasis);
   int refcount;
+  bool tensorbasis;      /* flag for tensor basis */
   CeedInt dim;           /* topological dimension */
   CeedInt ncomp;         /* number of field components (1 for scalar fields) */
   CeedInt P1d;           /* number of nodes in one dimension */
   CeedInt Q1d;           /* number of quadrature points in one dimension */
+  CeedInt P;             /* total number of nodes */
+  CeedInt Q;             /* total number of quadrature points */
```
jeremylt: I added 'P' and 'Q' because there isn't a conventional mapping from 'P1d' and 'Q1d' to total degrees of freedom or total quadrature points for non-tensor bases.

include/ceed.h
```diff
@@ -242,11 +242,28 @@ typedef enum {
...
+typedef enum {
+  /// Quad/Hex/Hypercube - 1+D shape
+  CEED_HCUBE = 0,
+  /// Simplex/tet - 2+D shape
+  CEED_SIMPLEX = 1,
+  /// Pyramid - 3D shape
+  CEED_PRYAMID = 2,
+  /// Prism - 3D shape
+  CEED_PRISM = 4,
+} CeedElemTopology;
```
jeremylt: I only added tests for 'CEED_SIMPLEX' right now, but for the ref backend it should all be the same. I'm not sure if other backends would want to know what type of non-tensor basis is being applied, but that is easy to add to the private basis data if desirable.

interface/ceed-fortran.c
```diff
@@ -280,14 +280,14 @@ static int CeedBasis_count_max = 0;
 #define fCeedBasisCreateTensorH1Lagrange \
     FORTRAN_NAME(ceedbasiscreatetensorh1lagrange, CEEDBASISCREATETENSORH1LAGRANGE)
 void fCeedBasisCreateTensorH1Lagrange(int *ceed, int *dim,
-                                      int *ndof, int *P, int *Q, int *quadmode, int *basis,
+                                      int *ncomp, int *P, int *Q, int *quadmode, int *basis,
```
jeremylt: Changed for consistency to not conflict with the parameters for 'BasisCreateH1'

tests/t300-basis.c
```diff
@@ -1,6 +1,6 @@
 /// @file
-/// Test ceration and distruction of a H1Lagrange basis
-/// \test Test ceration and distruction of a H1Lagrange basis
+/// Test creation and distruction of a H1Lagrange basis
```
jeremylt: Stupid typo

tests/tap.sh
```diff
@@ -42,8 +42,8 @@ for ((i=0;i<${#backends[@]}; ++i)); do
         continue
     fi
 
-    # grep to skip test if backend cannot handle Blocked Restriction
-    if grep -F -q -e 'Backend does not support ElemRestrictionCreateBlocked' \
+    # grep to skip test if backend support a function
```
jeremylt: Right now I changed tap to 'SKIP' whenever the error message contains 'Backend does not support' but I'd like to move to explicit whitelisting/blacklisting.

jedbrown: Fair for now.

As for white/blacklists, one thought is that we could make a `CeedInitTest("resource", "list,of,components", &ceed)` (or encode requirements in the resource string) that fails immediately if required components are not implemented. Then a backend would declare what it supports up-front and tests would specify programmatically which interface levels are needed. So if a backend claims to support `CeedBasisApply` then we expect every test that uses it to pass. I don't know if that is better than listing backends in the test source files, but I'm concerned that such lists could become messy and we could end up skipping some tests that should not be.

include/ceed.h
jeremylt: So enumerating CEED_LINE, CEED_SQUARE, CEED_CUBE, CEED_TRIANGLE, CEED_TETRAHEDRA separately?

jeremylt:
I will work on making an operator test. That is a really good idea.

Makefile
```diff
@@ -340,7 +340,7 @@ style :
             --suffix=none --preserve-date --formatted \
             --exclude=include/ceedf.h \
             include/*.h interface/*.[ch] tests/*.[ch] backends/*/*.[ch] \
-            examples/*/*.[ch] examples/*/*.[ch]pp
+            examples/*/*.[ch] examples/*/*.[ch]pp -i
```
jeremylt: exclude was throwing an `(unmatched)` error and stopping, so I added `-i` to not stop 'make style' on an exclude error

jeremylt:
`t520` added. I think this makes this PR ready to merge. Eventually I'd like to add a `t521` to get non-zero output from a similar test and `t510' and 't511' to do the same with identical physical coords on a 2D square mesh, but I think that can come when we add non-tensor bases to other backends.

tests/t501-operator-f.f
```diff
@@ -0,0 +1,171 @@
...
+
+      call ceedvectorgetarrayread(v,ceed_mem_host,hv,err)
+      total=0.
+      do i=1,nu
+        total=total+hv(i)
+      enddo
+c      if (abs(total-1.)>1.0D-10) then
+c        write(*,'(A,F12.8,A)')
+c     $  'Computed Area: ',total,' != True Area: 1.0'
+c      endif
```
jeremylt: 'getarrayread' in Fortran does not appear to be working correctly. I wonder if PR #134 will fix this.

tests/t500-operator-f.f
```diff
@@ -134,9 +137,19 @@ program test
...
       call ceedoperatorapply(op_mass,u,v,ceed_request_immediate,err)
 
+      call ceedvectorgetarrayread(v,ceed_mem_host,hv,err)
+      do i=1,nu
+        if (abs(hv(i))>1.0D-10) then
+          write(*,'(A,I1,A,F12.8,A)')
+     $    '[',i,'] v ',hv(i),' != 0.0'
+        endif
+      enddo
+      call ceedvectorrestorearray(v,hv,err)
```
jeremylt: This test did not previously check if the result was in fact correct.....

output/t320-basis-f.out
```diff
@@ -0,0 +1,15 @@
+CeedBasis: dim=2 P=6 Q=4
+        qref:	  0.20000000	  0.60000002	  0.33333334	  0.20000000	  0.20000000	  0.20000000	  0.33333334	  0.60000002
+     qweight:	  0.26041666	  0.26041666	 -0.28125000	  0.26041666
```
jeremylt: I'm not sure why these values computed in Fortran are slightly different that the ones computed in C for grad and interp.

jedbrown: Likely it's because the floating point literals you use to initialize the vectors are single precision. They are promoted to double precision only after computing the values. Look at `qref` and `qweight`.

jeremylt:
Ok, ready to merge now

jeremylt:
This fails Travis but passes locally with OCCA https://github.com/libocca/occa/commit/527494c949e95451f491c710592e9cd91e52b56f

include/ceed.h
jedbrown: If you're handling topology in a switch statement, it's easier for it to be flat. Also, we'll need to classify topology (to have homogeneous batches) and the key is simpler when it's one thing rather than two.

jeremylt:
Flattening the interface dropped the coverage because I now have more cases I'm not explicitly checking in the test suite. I thought it was silly to have the user provide both topology and dimension, but that means I need a switch to extract dimension... but I'm only testing one non-tensor topology.

jeremylt:
Rebased to the new testing convention

interface/ceed-basis.c
```diff
@@ -155,6 +158,87 @@ int CeedBasisCreateTensorH1Lagrange(Ceed ceed, CeedInt dim, CeedInt ncomp,
   return 0;
 }
 
+/**
+  @brief Create a non tensor product basis for H^1 discretizations
+
+  @param ceed       A Ceed object where the CeedBasis will be created
+  @param topo       Topology of element, e.g. hypercube, simplex, ect
+  @param dim        Topological dimension
```
jedbrown: doc

interface/ceed-basis.c
```diff
@@ -155,6 +158,87 @@ int CeedBasisCreateTensorH1Lagrange(Ceed ceed, CeedInt dim, CeedInt ncomp,
...
+  @param ncomp      Number of field components (1 for scalar fields)
+  @param ndof       Total number of nodes
+  @param nqpts      Total number of quadrature points
+  @param interp1d   Row-major nqpts × ndof matrix expressing the values of nodal
```
jedbrown: name for this and below. Look at the `make doc` warnings.

include/ceed.h
```diff
@@ -242,11 +242,34 @@ typedef enum {
...
+  /// Hexehedron - 3D shape
+  CEED_HEX = 2,
+  /// Triangle - 2D shape
+  CEED_TRIANGLE = 4,
+  /// Tetrahedron - 3D shape
+  CEED_TET = 8,
+  /// Pyramid - 3D shape
+  CEED_PYRAMID = 16,
+  /// Prism - 3D shape
+  CEED_PRISM = 32,
```
jedbrown: What is the logic about numbering here? If you had, say, `dim << 16 | topo_type` then you could get the dimension with a uniform operation. Also, it looks like we don't actually care about the topology anywhere in the library, though in case of mixed meshes we'll need to identify "like topologies". I could imagine we eventually add a constructor so the user can create new topologies.

jeremylt: I'll rethink the numbering. I like your suggestion of embedding the dim in the number.

jeremylt: Updated. I am using the first three bits for dimension and the last three for shape. 

001xxx is 1D
010xxx is 2D
100xxx is 3D

Right now I don't see the library caring about the topology, but I imagine that some backends will.

interface/ceed-basis.c
jeremylt: make doc is happy now

interface/ceed-basis.c
```diff
@@ -188,35 +187,15 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt ncomp,
...
-  case CEED_PYRAMID:
-    dim = 3;
-    break;
-  case CEED_PRISM:
-    dim = 3;
-    break;
-  }
+  for (int i=1; i<=3; i++)
+    if (topo & CeedIntPow(2, i + 2))
+      dim = i;
```
jedbrown: We should have a function, even if internal, that gives the dimension for any topology.

jeremylt: The way I have it right now, I am imagining that 12-15 are user 1D topologies, 20-23 are user 2D topologies, and 36-39 are user 3D topologies. That way we can support up to 4 additional topologies of each dimension and then I can make an internal function to replicate those three lines.The only part I'm not clear on is how  best to assign these user topologies to these values.

jedbrown: But why have a small number of user-defined topologies when it's no harder (arguably easier) to have thousands as my previous suggestion of `dim << 16 | counter` would do?

jeremylt: Because I didn't understand how that expression worked and took a stab at making something similar?

jedbrown: Ah, `dim << 16` is left-shift by 16 bits. The `|` is bitwise or, which in this case acts like addition. Printed in hexadecimal, `3 << 16 | 5 == 0x30005`. You get the dimension back with `topo >> 16`. Use an unsigned type.

include/ceedf.h
```diff
@@ -61,25 +73,49 @@ c
...
 c
+
       integer ceed_gauss
       parameter(ceed_gauss         = 0)
 
       integer ceed_gauss_lobatto
       parameter(ceed_gauss_lobatto = 1)
 
+c
+c CeedElemTopology
```
jeremylt: Fortran 77 does not appear to play nice with unsigned integers, but we only need the unsigned integer on the library side, so I put a tuple with the 'recipe' on the Fortran side that gets converted.

jedbrown: Signed and unsigned representations are the same up to `INT_MAX`. I said to use unsigned because right shift has implementation-defined behavior for signed integers. It doesn't really matter which is used in the enum.

jedbrown: Note that in Fortran you can write `parameter(CEED_TRIANGLE = X'20001')`.

jedbrown:
Need to fix spelling (COLOCATED build failure). May want to squash some commits that are not actually independent and/or don't pass tests. (Rebasing onto 'master' is risky with a long branch because you likely don't retest each commit separately.) Exclude `*-f.h` from `make style` (it messes up `tests/t310-basis-f.h`). Looks like it'll be ready after that housekeeping.

jeremylt:
astyle exclude does not support wildcards, but I added `t310-basis-f.h` to the exclude list. I squashed the commits that seem to go together. I indicated the commits where I updated for another PR, so chronologically the tests should pass when merged. I would not be opposed to further squashing if it makes sense to you.

jedbrown:
So a key motivation for having every commit pass is that `git bisect` should succeed for all known tests. We currently fail at 04ea18f8c92c1b3d84a3a2a2dbd6e7c8a91be855 because "not implemented" doesn't arrive until 2ce2473be31f5dd6fd98cbaa57115e6bb3e6c56d. Also, the tests here can't possibly work with Fortran until 9b33debc80b31361e5582d5ff4066f602c0c60d1. There's also a bit of renaming going on in the branch -- we might as well squash that out because it's just noise for someone trying to track the origins of a bug.

tests/t310-basis.c
```diff
@@ -0,0 +1,25 @@
...
+#include "t310-basis.h"
+
+int main(int argc, char **argv) {
+  Ceed ceed;
+  const CeedInt P = 6, Q = 4, dim = 2;
+  CeedBasis b;
+  CeedScalar qref[dim*Q], qweight[Q];
+  CeedScalar interp[P*Q], grad[dim*P*Q];
+
+  buildmats((CeedScalar *) &qref, (CeedScalar *) &qweight,
```
jedbrown: You can ditch the casts. `buildmats(qref, qweight, interp, grad)`.

tests/t310-basis.h
```diff
@@ -0,0 +1,42 @@
+void buildmats(CeedScalar *qref, CeedScalar *qweight, CeedScalar *interp,
```
jedbrown: Generally good practice to make this `static`. (At some point I may make it possible to build the tests into a library or to build one executable instead of needing to link dozens of separate executables. Not important yet though.)

tests/t310-basis.c
jeremylt: without the casts I get 'warning: passing argument ... from incompatible pointer type'

jedbrown: Full code/error message? We should not have casts here.

jeremylt: I can ditch the casts and avoid the warning with MALLOC if that is preferable

jeremylt:
```
/home/jeremy/Dev/libceed/tests/t310-basis.c:14:13: warning: passing argument 1 of ‘buildmats’ from incompatible pointer type [-Wincompatible-pointer-types]
   buildmats(&qref, &qweight, &interp, &grad);
             ^~~~~
In file included from /home/jeremy/Dev/libceed/tests/t310-basis.c:5:
/home/jeremy/Dev/libceed/tests/t310-basis.h:1:34: note: expected ‘CeedScalar *’ {aka ‘double *’} but argument is of type ‘CeedScalar (*)[(0, 8)]’
 staic void buildmats(CeedScalar *qref, CeedScalar *qweight, CeedScalar *interp,
```

jeremylt: Never mind; I got it

tests/t310-basis.h
jeremylt: Done

PR #98, opened 2018-07-21
[WIP] t17-f bug

jeremylt:
@jedbrown here is the `t17-basis-f` bug that I was talking about. Part of the (lengthy) output from the two 3D cases is below. The problem only manifests in the 3D case and the C test works perfectly. Every entry in the application of the transpose in the Fortran code is off by 3-5ish and the error adds up to ~280.

```
[jeremy@pheonix libceed]$ ./build/t17-basis /cpu/self
Sum 1:
11.954906
26.959247
37.632024
46.738399
61.571262
...
1104.169648
1160.009795
1208.067615
1251.295207
1279.877736
Sum 2:
0.857980
1.676995
2.335189
2.597477
2.301182
.....
1278.385231
1278.758357
1279.131483
1279.504609
1279.877736
[jeremy@pheonix libceed]$ ./build/t17-basis-f /cpu/self
 Sum 1:
   8.4254765043861806     
   19.957397471457611     
   27.932510394161067     
   34.772322694048817     
   45.820260058194123    
.....   
   964.97821921494437     
   1003.5077873767808     
   1036.2806850204406     
   1067.6815700674724     
   1084.6436215904973     
 Sum 2:
  0.85798013729573031     
   1.6769953721706286     
   2.3351893919667335     
   2.5974770340153555     
   2.3011814331908287 
.....    
   1278.3852263192830     
   1278.7583523706048     
   1279.1314784219267     
   1279.5046044732485     
   1279.8777305245703     
[3] Error:  1084.643622 !=  1279.877731
```

PR #92, opened 2018-07-11
[WIP] Nek5000 BP5

thilinarmtb:
Still a work in progress.

- Needs refactoring and cleaning the git history

PR #99, opened 2018-07-23
Occa Backend Fixes

Steven-Roberts:
I noticed the memory management for occa vectors seemed a bit off.  When a user passes an array `a` to `CeedVectorSetArray` with `CEED_USE_POINTER` or `CEED_OWN_POINTER` mode then calls `CeedVectorGetArray` to make changes, those changes should be reflected in `a`.  This is the case for the cpu backends but not the occa one.  I added a new test for this, which fails for the current occa backend, but works for this new one.  I then realized that the fortran wrappers also break this "feature."  Perhaps this can be ignored.  Either way, this new implementation is a bit cleaner and more consistent with the ref vector.

Other changes:
- Added missing `occaFree` for vector
- Replaced potentially large array allocation on stack with malloc/free
- Renamed occa properties to support latest version (https://github.com/libocca/occa/commit/d1fd6e0f3b569ebf7614f09f02ab7394a3f2804e)

Some tests with Valgrind indicate there may be some more memory issues.

### DIFF
--- a/backends/occa/ceed-occa-operator.c
+++ b/backends/occa/ceed-occa-operator.c
@@ -296,8 +296,8 @@ static int SyncToHostPointer(CeedVector vec) {
   // hold an output quantity.  This should at least be lazy instead of eager
   // and we should do better about avoiding copies.
   const CeedVector_Occa *outvdata = (CeedVector_Occa*)vec->data;
-  if (outvdata->used_pointer) {
-    occaCopyMemToPtr(outvdata->used_pointer, outvdata->d_array,
+  if (outvdata->h_array) {
+    occaCopyMemToPtr(outvdata->h_array, outvdata->d_array,
                      vec->length * sizeof(CeedScalar), NO_OFFSET, NO_PROPS);
   }
   return 0;
--- a/backends/occa/ceed-occa-restrict.c
+++ b/backends/occa/ceed-occa-restrict.c
@@ -160,7 +160,8 @@ int CeedElemRestrictionCreate_Occa(const CeedElemRestriction r,
   data->d_tindices = occaDeviceMalloc(dev, bytes(r), NULL, NO_PROPS);
   // ***************************************************************************
   CeedInt toffsets[r->ndof+1];
-  CeedInt tindices[r->elemsize*r->nelem];
+  CeedInt *tindices;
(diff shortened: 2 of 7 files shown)


jeremylt:
This may require updates to OpApply in the OCCA backend - we added some logic to force a used array to be updated in order to get OpApply to work in OCCA.

jedbrown:
I like the changes I see (thanks!), but I'm waiting on review and resolving the failing tests in Travis. Do you know why it's failing, @Steven-Roberts? Do you need help debugging?

Steven-Roberts:
It's the new fortran test I created that is failing.  I think I have narrowed down the cause now.  Due to an extra memcpy in `fCeedVectorGetArray`, it no longer provided a pointer alias but a completely different pointer.
https://github.com/CEED/libCEED/blob/0771ec9989bc3dfb577315a80cc0b18e86cb2dd3/ceed-fortran.c#L88
Perhaps the simplest fix is to just remove the test.  Or can this memcpy be removed?

jedbrown:
@thilinarmtb Since `CeedVectorGetArray` is meant to obtain _mutable_ access to the vector, if you are going to make a copy, then `CeedVectorRestoreArray` will need to copy back. (Frankly, that's crazy and we should make it work without copies.)

I think I pointed you to how PETSc does this in Fortran77 with a base offset (https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecGetArray.html) and much more cleanly in Fortran90 (https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecGetArrayF90.html#VecGetArrayF90).

How do you want this to work?

jeremylt:
I think PRs #92, #98, #99, and #101 are all involving a pair of related issues - namely OCCA and Fortran memory access. I think this would be a good topic for discussion this week at the CEED 2AM for the libCEED team. Thoughts @jedbrown, @thilinarmtb, @camierjs?

thilinarmtb:
@jedbrown : I like the Petsc fortran 77 base solution. Will try to get it in.

jedbrown:
@camierjs @Steven-Roberts I've had this PR on hold pending discussion at the meeting and Fortran changes (#123 #134). Should it be rebased/merged after #118 and #134?

PR #100, opened 2018-07-27
Opt CPU Backend - Vectorize across 8 element batches

jeremylt:
I have not done speed tests yet, but the new vectorization is in place.


jeremylt:
# Crude initial timing results:

## Current Master with `/cpu/self/opt`

```
$ time ./build/petsc-bp1 -local 1000000
Process decomposition: 1 1 1
Local elements: 1000000 = 100 100 100
Owned dofs: 1030301 = 101 101 101
KSP cg CONVERGED_RTOL iterations 23 rnorm 8.518818e-08
Pointwise error (max) 2.696996e-04

real	0m8.216s
user	0m7.994s
sys	0m0.217s

```

## New Branch with `/cpu/self/opt`

```
$ time ./build/petsc-bp1 -local 500000
Process decomposition: 1 1 1
Local elements: 1000000 = 100 100 100
Owned dofs: 1030301 = 101 101 101
KSP cg CONVERGED_RTOL iterations 23 rnorm 8.518818e-08
Pointwise error (max) 2.696996e-04

real	0m8.038s
user	0m7.803s
sys	0m0.227s
```

This feels like a rather small improvement, though I could be running too small of a problem or my vectorization could stand to be improved or both.

backends/optimized/ceed-opt-operator.c
```diff
@@ -18,12 +18,451 @@
...
+      // Get emode, ncomp
+      CeedEvalMode emode = qf->inputfields[i].emode;
+      CeedInt ncomp = qf->inputfields[i].ncomp;
+      // Basis action
+      switch(emode) {
+      case CEED_EVAL_NONE:
+        break; // No action
+      case CEED_EVAL_INTERP:
+        data = op->inputfields[i].basis->data;
+        data->batchsize = batchsize;
```
jedbrown: I'm really nervous about setting this here and then leaving it. Truthfully, I would rather _pad_ the input (by replicating data from the last element) so that the batch size is never mutated. Alternatively, we could add an explicit argument to `CeedBasisApply` for the batch size, but that would require adopting a convention for how to index across multiple elements (as a first or last index to the tensor).

backends/optimized/ceed-opt-basis.c
```diff
@@ -119,13 +152,21 @@ static int CeedBasisApply_Opt(CeedBasis basis, CeedTransposeMode tmode,
...
+      // CEED_NOTRANSPOSE - Gradient at quadrature points
+      // CEED_TRANSPOSE - Interpolate back to dofs
       P = basis->Q1d, Q = basis->Q1d;
+      outsize = nqpt*dim;
+      midsize = ncomp*Q*CeedPowInt(P>Q?P:Q, dim-1);
       if (tmode == CEED_TRANSPOSE) {
         P = basis->Q1d, Q = basis->P1d;
+        outsize = ncomp*CeedPowInt(basis->P1d, dim);
       }
       pre = ncomp*CeedPowInt(P, dim-1), post = 1;
```
jedbrown: So this batching looks good, but we need to change one thing. The number of elements should be the last index (which means element restriction needs to be permuted), so that `post = batchsize`. This will make it so that every operation within TensorContract will operate on at least `batchsize` contiguous values, so that a consistent vectorization strategy applies. (Otherwise vectorization is messier and dependent on the sizes `P` and `Q`.) If that change in (temporary) E-vector data layout makes sense and you can implement it, then I can show how to specialize the tensor contraction so the compiler will generate good code. (It's too generic now for the compiler to vectorize well, but we need the data layout change to have an easy path to good implementation -- or any path to good implementation at some data sizes.)

backends/optimized/ceed-opt-operator.c
jeremylt: I found the output to be more of a problem because the tensor contract would try to write into unallocated memory if we aren't careful. Though, input can also be a problem for us because we might have an input with no restriction, so we don't have a natural way to pad that input without doing a special copy operation just for the last batch of elements. I added `batchsize` argument to try to prevent both these overwrites and special copies.

On line 262 I reset `batchsize` back to 1 on the last batch of elements. I am not fully happy with how I did it, but this field was my attempt to pass an argument to the `opt` backend `BasisApply` without reworking the arguments to the function for all backends.

backends/optimized/ceed-opt-basis.c
jeremylt: So we would always apply a restriction operation in this case, even for `CEED_RESTRICTION_IDENTITY` to permute the dofs? That makes sense to me, but I am not sure how we would make this permuting compatible with the `QFunctionApply`. Right now qfunctions require contiguous inputs and outputs. Would we interlace for basis actions, separate for qfunctions, then interlace again for the basis before separating for the final basis action? That feels like a lot of copying.

backends/optimized/ceed-opt-operator.c
jedbrown: I see two viable approaches.
1. `batchsize` is an immutable property of a `CeedBasis` (though setting can be deferred). The caller can query it and is responsible for padding as needed (so that all lanes are valid).
2. `batchsize` is an argument of `CeedBasisApply` and existing callers will call with `batchsize=1`. (Variant: create new `CeedBasisApplyBatch`; `CeedBasisApply` can be implemented by calling `CeedBasisApplyBatch(batchsize=1, ...)`.)

In the former, the `CeedBasis` can statically optimize for a fixed batch size. In the latter, it will pick a path each time it is called. Since GPU batches might be very large [*], the second alternative may be preferable.

[*] The warp size is likely all that is needed in the "inner" dimension for a GPU, but each kernel launch will want to process many elements (likely the entire mesh). That said, an optimized GPU implementation will likely want to fuse BasisApply with other operations, in which case they would have a custom fused operation instead of calling via a generic `CeedBasisApplyBatch`.

backends/optimized/ceed-opt-basis.c
jedbrown: The qfunction doesn't care whether the "next" quadrature point is in the same element or a different element. We just need to keep all Q-vectors in the same ordering (e.g., interlaced within batches), but there are no additional implied permutations/copies.

It is true that we'll need a permutation for `CEED_RESTRICTION_IDENTITY` (and perhaps eventually provide a way for that data to be provided in a batched ordering).

jeremylt:
This PR will be updated following PR #110 and PR #106 to improve handling of batches of elements.

backends/optimized/ceed-opt-operator.c
```diff
@@ -17,13 +17,446 @@
 #include <ceed-impl.h>
 #include <string.h>
 #include "ceed-opt.h"
+#include "../ref/ceed-ref.h"
```
jeremylt: I'm not too happy about this.

jeremylt:
# New timing results:

## Current Master with `/cpu/self/ref`

```
$ time ./build/mfem-bp3 -ceed /cpu/self/ref -o 6
Options used:
   -ceed /cpu/self/ref
   --mesh /home/jeremy/Dev/mfem/data/star.mesh
   --order 6
   --visualization
   --no-test
Number of finite element unknowns: 46561
   Iteration :   0  (B r, r) = 1697.41 ...
   Iteration : 662  (B r, r) = 1.57191e-09
Average reduction factor = 0.97929
L2 projection error: 9.83612e-06

real	0m4.173s
user	0m4.164s
sys	0m0.003s
```

## Current Master with `/cpu/self/opt`

```
$ time ./build/mfem-bp3 -ceed /cpu/self/opt -o 6
Options used:
   -ceed /cpu/self/opt
   --mesh /home/jeremy/Dev/mfem/data/star.mesh
   --order 6
   --visualization
   --no-test
Number of finite element unknowns: 46561
   Iteration :   0  (B r, r) = 1697.41 ...
   Iteration : 662  (B r, r) = 1.57191e-09
Average reduction factor = 0.97929
L2 projection error: 9.83612e-06

real	0m4.202s
user	0m4.192s
sys	0m0.007s
```

## New Branch with `/cpu/self/opt`

```
$ time ./build/mfem-bp3 -ceed /cpu/self/opt -o 6
Options used:
   -ceed /cpu/self/opt
   --mesh /home/jeremy/Dev/mfem/data/star.mesh
   --order 6
   --visualization
   --no-test
Number of finite element unknowns: 46561
   Iteration :   0  (B r, r) = 1697.41 ...
   Iteration : 662  (B r, r) = 1.57191e-09
Average reduction factor = 0.97929
L2 projection error: 9.83612e-06

real	0m2.736s
user	0m2.719s
sys	0m0.013s
```

Much better!

jeremylt:
At some point along the line we stopped testing `/cpu/self/tmpl`. Was that intentional?

backends/optimized/ceed-opt-operator.c
```diff
@@ -17,13 +17,331 @@
...
+    // Output pointers
+    for (CeedInt i=0; i<qf->numoutputfields; i++) {
+      CeedEvalMode emode = qf->outputfields[i].emode;
+      if (emode == CEED_EVAL_NONE) {
+        CeedInt ncomp = qf->outputfields[i].ncomp;
+        impl->outdata[i] = &impl->edata[i + qf->numinputfields][e*Q*ncomp];
+      }
+    }
+    // Q function
+    ierr = CeedQFunctionApply(op->qf, Q*blksize,
```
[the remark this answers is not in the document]
Steven-Roberts: Ok thanks.  Seems pretty straightforward.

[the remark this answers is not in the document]
jeremylt: Yeah. I was actually pretty shocked by how easy it was. I was expecting to have to do something exotic/painful.

backends/optimized/ceed-opt-basis.c
```diff
@@ -57,108 +54,106 @@ static int CeedBasisApply_Opt(CeedBasis basis, CeedInt nelem,
...
-    }
-  }
-  if (emode & CEED_EVAL_GRAD) {
-    CeedInt P = basis->P1d, Q = basis->Q1d;
+  } else if (emode == CEED_EVAL_GRAD) {
     // In CEED_NOTRANSPOSE mode:
-    // u is (P^dim x nc), column-major layout (nc = ncomp)
-    // v is (Q^dim x nc x dim), column-major layout (nc = ncomp)
+    // u is (P^dim x nc) x nelem, column-major layout (nc = ncomp)
+    // v is (Q^dim x nc x dim) x nelem, column-major layout (nc = ncomp)
```
jedbrown: Except the `nelem` part is the contiguous direction. We should adopt a consistent convention about expressing shapes. I would propose the C-inspired `[dim, ncomp, Q^dim, nelem]`. We can do that as part of a sanitation pass later.

jeremylt: I'm good to merge; I'll start the sanitation pass branch today.

Can you put those compile options you were using in your branch please?

jedbrown: ce77c5bc200beacfeecb101312aa5f3112587880

jedbrown: Do you want to cherry-pick cf12fa7df7c3e65edbafc900d2c138497e2eec99 into your branch now?

jeremylt: Updated

PR #106, opened 2018-08-01
Implement RestrictionCreateBlocked

jeremylt:
This PR adds `RestrictionCreateBlocked` and implements applying batch restrictions in Ref.


jeremylt:
I need to make tests, but this should make and apply batched restrictions with padding. The NoTraspose should work for any backend, but only Ref currently strips the padded elements from the last block.

jeremylt:
@steven-roberts @camierjs @stomov
I wanted to call the attention of the GPU team to this PR. It implements interlaced batched element restrictions.   The library level call will appropriately permute and pad the indices vector that would be provided for a 'normal' restriction.

I have been jamming CPU code into your backends to get them to pass tests as we add to/modify the frontend. I feel like my efforts are slowing down your backends because I am not familiar with how to best implement ideas for your backends. You are welcome to add to this PR to best handle batched elements, and specifically stripping out the padding in a way that is fastest for your backends. Also coming is batched Basis actions in a subsequent PR so I can continue to optimize the CPU backend.

YohannDudouit:
I'm not sure I understand how this algorithm works, is the idea to treat elements block per block with the idea to do some sort of cache blocking? How would this idea transpose to GPU, do you want to do this in the shared memory, or directly at warp level using warp shuffle intrinsics (I'm not sure that it would be possible)?

jedbrown:
On a CPU, you usually want to restrict a batch of elements into cache (equivalent to a segment of an E-vector, but the E-vector never exists in full), do the vectorized operations on that batch, and contribute back to the L-vector. On a GPU, you could specify a block size that related to a warp or threadblock size, which would theoretically enable fusing with BasisApply and QFunctionApply. (This matters for kernel launch latency, but also memory bandwidth since E-vectors require significantly more storage than L-vectors, especially at low order. Fusing the element restriction would reduce the total amount of memory accessed and the number of times it would need to be accessed.)

Steven-Roberts:
Is it correct that these changes are to allow SIMD in the opt backend?  So a block contains data from the same quadrature point but on different elements?  And the padding is to ensure everything fits exactly in the SIMD width?

@jedbrown That's an interesting idea for fusing.  I'll need to think more about the pros and cons of that format over the standard layout.

jedbrown:
Yes, the reordering within each block is to exploit SIMD and the padding is so there is no fringe code within BasisApply and QFunctionApply (all loops are multiples of the full SIMD vector length).

I think the primary disadvantage is lower occupancy and the need to use atomics when accumulating into the L-vector instead of having the option of having a single thread sum all contributions into any given entry of the L-vector. The use of shared memory is also likely more size-dependent.

YohannDudouit:
One of the problem we were having when applying the restriction was mostly how efficient we are at reading and writing in the Global memory. Because of the way cachelines are requested by threads, it was difficult to have anything that was even close to be coalesced. We attempted to use shared memory to bypass this, but due to the smaller size of shared memory compared to the number of registers, shared memory seems to be only usable as a circular buffer. We did not manage to have better performance with alternated Global memory load to shared memory and Shared memory read to registers.
I feel that your strategy works on CPU because caches have this hierarchical structure (L3 bigger than L2 bigger than L1 bigger than registers), but on GPU Shared memory being smaller than registers might result in a very low usage of the SMs. Then we could hope for some kind of cache optimization at L2 level but that looked more difficult since all threads are accessing this memory. Am I missing something?

YohannDudouit:
The main advantage I was seeing with storing elements in the E-vector format is that it is a lot easier to read the memory coalesce, with the known drawbacks. You can also fuse the scatter/gather algorithm when you store them in E-vector format. Of course some operations get more expensive like the dot product, but the extra cost could pay for what we save elsewhere.
The problem is that I don't see how we could be efficient on doing the scatter gather if we don't do some kind of tiling with this approach. The basic idea would be that all SMs read the whole tile once, and the SM's threads cherry pick the dofs to gather/scatter in one path (which means that they need to load some indices). With this approach we can read the memory coalesce and write it coalesce to perform gather/scatter, then to apply the operator it can also be read and written coalesce.

jedbrown:
I don't think I'm following your point. The representation of an E-vector is exclusively at the discretion of the backend. It is possible for a caller to define its L-vector to have a discontinuous representation such that element restriction (with a particular E-vector convention) is the identity. That's what Nek5000 does, but there are nontrivial costs, especially at lower order. LibCEED does not aspire to require one form or another.

There is a valid discussion about what L-vector ordering can provide the best performance, and that will likely be different on CPU and GPU due to competing concerns of coalescing and memory locality (though packing MPI send/recv buffers should also be considered). But that discussion belongs in some optimization recommendations for users; the library doesn't get to make that decision.

jeremylt:
This now works for the `/cpu/self`family of backends. I changed tap to skip the new `t06` test for the OCCA and MAGMA backends until this functionality is implemented for those backends.

ceed-elemrestriction.c
```diff
@@ -86,24 +86,11 @@ int CeedElemRestrictionCreate(Ceed ceed, CeedInt nelem, CeedInt elemsize,
...
-  if (nlastelems == 0) nlastelems = blksize;
-  for (int j = 0; j < blksize; j++) {
-    for (int k = 0; k < elemsize; k++) {
-      if (j < nlastelems) {
-        blkindices[shift + k*blksize + j] = indices[shift + k + j*elemsize];
-      } else {
-        blkindices[shift + k*blksize + j] = indices[shift + k +
-                                            (nlastelems - 1)*elemsize];
+        blkindices[e*elemsize + k*blksize + j]
+          = indices[CeedIntMin(e+j,nelem-1)*elemsize + k];
```
jedbrown: @jeremylt What do you think of this loop structure? Can we use it in the other places where `nblk - 1` appears as a special case in `CeedElemRestrictionApply_Ref`?

jeremylt: I did something similar with a ternary operator in Ref but I think this looks cleaner

backends/ref/ceed-ref-restriction.c
```diff
@@ -70,67 +68,43 @@ static int CeedElemRestrictionApply_Ref(CeedElemRestriction r,
...
-      if (nlastelems == 0) nlastelems = blksize;
-      for (CeedInt i = 0; i<blksize*elemsize; i++) {
+      for (CeedInt i = 0; i < blksize*elemsize; i++) {
         if ((i % blksize) < nlastelems) {
           vv[impl->indices[shift + i]] += uu[shift + i];
         }
       }
     } else {
       // u is (elemsize x ncomp x nelem)
       if (lmode == CEED_NOTRANSPOSE) { // vv is (ndof x ncomp), column-major
```
jeremylt: I am suspicious of my logic here after the loop reworking and want to look at again in the morning.

PR #110, opened 2018-08-05
Add nelem argument to BasisApply

jeremylt:
This adds the argument `nelem` to `CeedBasisApply`, facilitating backends handling multiple element batches.

### DIFF
--- a/backends/magma/ceed-magma.c
+++ b/backends/magma/ceed-magma.c
@@ -594,6 +594,10 @@ static int CeedBasisApply_Magma(CeedBasis basis, CeedInt nelem,
   const CeedInt nqpt = ncomp*CeedPowInt(basis->Q1d, dim);
   const CeedInt add = (tmode == CEED_TRANSPOSE);
 
+  if (nelem != 1)
+    return CeedError(basis->ceed, 1,
+                     "This backend does not support BasisApply for multiple elements");
+
   CeedDebug("\033[01m[CeedBasisApply_Magma] vsize=%d",
             ncomp*CeedPowInt(basis->P1d, dim));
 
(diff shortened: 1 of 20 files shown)


backends/magma/ceed-magma.c
```diff
@@ -585,15 +585,19 @@ static int CeedTensorContract_Magma(Ceed ceed,
   return 0;
 }
 
-static int CeedBasisApply_Magma(CeedBasis basis, CeedTransposeMode tmode,
```
jedbrown: Try to avoid introducing trailing whitespace. See git config `core.whitespace` for ways to help notice it.

ceed-basis.c
```diff
@@ -407,12 +407,12 @@ int CeedBasisGetColocatedGrad(CeedBasis basis, CeedScalar *colograd1d) {
 ///     \ref CEED_EVAL_GRAD to obtain gradients.
 /// @param u input vector
 /// @param v output vector
-int CeedBasisApply(CeedBasis basis, CeedTransposeMode tmode, CeedEvalMode emode,
-                   const CeedScalar *u, CeedScalar *v) {
+int CeedBasisApply(CeedBasis basis, CeedInt nelem, CeedTransposeMode tmode,
```
jedbrown: The man page above needs to be updated for the new parameter and implied input and output array ordering.

jeremylt: Good point about the manpages. As for the implied ordering, I am a bit unsure. It didn't sound like the GPU backends would like to interlace elements. Would it be fair to state that the `RestrictionCreateBlocked` for a given backend will enforce the ordering required for the `BasisApply`?

jedbrown:
GPU backends will most likely want the innermost index to be element
number.  It could also be blocked by the warp size or any multiple
thereof, probably with equivalent performance but slightly more
complicated code.

PR #112, opened 2018-08-07
Add codecov.io

jedbrown:


jeremylt:
Do we need to upload an initial coverage report? It looks good to me except for the 

"No coverage uploaded for pull request base (master@a4b3319). Click here to learn what that means."

Issue #105, opened 2018-08-01
CeedVectorSetValue


jeremylt:
Add utility function 'CeedVectorSetValue(CeedVector vec, CeedScalar val)`. This will clean up zeroing out vectors.

PR #114, opened 2018-08-07
Create VectorSetValue and tests

jeremylt:
Addresses issue #105.

Given the number of active PRs with backends, I am saving putting this function into backend code for later.

### DIFF
--- a/ceed-vec.c
+++ b/ceed-vec.c
@@ -64,7 +64,6 @@ int CeedVectorSetArray(CeedVector x, CeedMemType mtype, CeedCopyMode cmode,
 /// Set the array used by a vector, freeing any previously allocated array if applicable.
 ///
 /// @param x Vector
-/// @param mtype Memory type of the array being passed
 /// @param value to be used
 int CeedVectorSetValue(CeedVector x, CeedScalar value) {
   int ierr;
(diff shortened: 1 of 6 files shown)


ceed-vec.c
```diff
@@ -61,6 +61,22 @@ int CeedVectorSetArray(CeedVector x, CeedMemType mtype, CeedCopyMode cmode,
   return 0;
 }
 
+/// Set the array used by a vector, freeing any previously allocated array if applicable.
```
jedbrown: Bad copy-pasta. We'll eventually need this to be implementable by backends (to avoid moving data to/from host). You could add a method pointer and have this implementation use the default when the method pointer is NULL.

jeremylt: Updated

Steven-Roberts:
Having an additional argument for `CeedMemType` could potentially be useful.  There are cases in the occa and cuda backends where only the device array needs to be zeroed out.

jedbrown:
@Steven-Roberts My feeling is that that's up to the backend and the caller doesn't know what would be correct. A typical backend would zero only the device copy (and would flag that the host copy is stale).

A clever backend might even delay zeroing the device copy until it can be fused into the next operation. Does that seem reasonable?

Steven-Roberts:
It could be usefully internally, but it does seem reasonable to keep it out of this library function.

PR #116, opened 2018-08-08
Fix typo in tmpl backend init

jeremylt:
This PR fixes a typo in the tmpl backend that was preventing it from being used/tested.


jeremylt:
And Travis is no longer skipping tmpl.

PR #117, opened 2018-08-09
codecov: create flags

jedbrown:


jeremylt:
This looks very helpful to me - I'm curious to see the new reports

jedbrown:
I don't see any change in the diagnostics. The [documentation](https://docs.codecov.io/docs/flags) talks about submitting different flags separately using `-F flagname`, but when I [do that](https://codecov.io/gh/CEED/libCEED/compare/master...jed%2Fcodecov-flags) in a single batch, it labels them all the same. I was hoping to get summary statements about coverage changes in the `interface/` and `backends/` paths. Does anyone (@dmed256?) have experience with this feature?

dmed256:
Here's what I used to split the coverage a while back
```yaml
coverage:
  precision: 2
  round: down
  range: "70...100"

  status:
    project:
      default: false
      api:
        flags: api
      parser:
        flags: parser
    patch:
      default: false
      api:
        flags: api
      parser:
        flags: parser

flags:
  api:
    paths:
      - include/
      - src/
    ignore:
      - examples/
  parser:
    paths:
      - parser_sandbox/include
      - parser_sandbox/src
    ignore:
      - parser_sandbox/tests

comment:
  layout: 'reach, diff, files'
  behavior: default
  require_changes: false
  require_base: no
  require_head: yes
  branches: null
```

This split the coverage for `api` and `parser` where each covered these paths:
- api
  - include/
  - src/
- parser
  - parser_sandbox/include
  - parser_sandbox/src

dmed256:
Oh, I wonder if it's the compiler flags. Sorry, using my phone so I can't browse too well

https://github.com/libocca/occa/blob/master/scripts/Makefile#L142

jedbrown:
Huh, [no flags found](https://codecov.io/gh/libocca/occa/pull/196?src=pr&el=h1) for that Occa configuration.

jedbrown:
The interface and backends flags are working now after identifying a workaround with codecov support. Apparently this is the ticket number, though I don't have access to view it. https://codecov.freshdesk.com/helpdesk/tickets/290

PR #118, opened 2018-08-09
Cleaning Up Code

jeremylt:
This is a series of minor changes to clean up the codebase.

Issue #64, opened 2018-03-31
Add debug-mode checks that every GetArray has a matching RestoreArray


jedbrown:
`CeedVectorGetArray` and `CeedVectorGetArrayRead` are supposed to always be paired with `CeedVectorRestoreArray` and `CeedVectorRestoreArrayRead` respectively, but sometimes users and even backend developers such as myself accidentally forget. In debug mode, the interface should track access so that it can warn in case of non-exclusive write access or if `CeedVectorDestroy` is called while some caller has access. When available, this functionality could print a stack trace using [backtrace](https://www.gnu.org/software/libc/manual/html_node/Backtraces.html) from glibc or [libbacktrace](https://github.com/ianlancetaylor/libbacktrace).

jeremylt:
PR #120

PR #120, opened 2018-08-11
Vec State Tracking

jeremylt:
I added a state variable to track access to CeedVectors. After PR #118 I will add a test to this branch. This is related to issue #64.

### DIFF
--- a/include/ceed-impl.h
+++ b/include/ceed-impl.h
@@ -68,6 +68,7 @@ struct CeedVector_private {
   int (*Destroy)(CeedVector);
   int refcount;
   CeedInt length;
+  uint64_t state;
   void *data;
 };
 
--- a/interface/ceed-vec.c
+++ b/interface/ceed-vec.c
@@ -49,6 +49,7 @@ int CeedVectorCreate(Ceed ceed, CeedInt length, CeedVector *vec) {
   ceed->refcount++;
   (*vec)->refcount = 1;
   (*vec)->length = length;
+  (*vec)->state = 0;
   ierr = ceed->VecCreate(ceed, length, *vec); CeedChk(ierr);
   return 0;
 }
@@ -69,9 +70,15 @@ int CeedVectorSetArray(CeedVector vec, CeedMemType mtype, CeedCopyMode cmode,
                        CeedScalar *array) {
   int ierr;
 
+  if (vec && (vec->state % 2) == 1)
+    return CeedError(vec->ceed, 1, "Cannot grant CeedVector array access, the access lock is already in use");
+
   if (!vec || !vec->SetArray)
     return CeedError(vec ? vec->ceed : NULL, 1, "Not supported");
+
   ierr = vec->SetArray(vec, mtype, cmode, array); CeedChk(ierr);
+  vec->state += 2;
+
   return 0;
 }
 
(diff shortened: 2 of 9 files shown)


ceed-vec.c
```diff
@@ -96,9 +99,14 @@ int CeedVectorSetValue(CeedVector x, CeedScalar value) {
 /// @sa CeedVectorRestoreArray()
 int CeedVectorGetArray(CeedVector x, CeedMemType mtype, CeedScalar **array) {
   int ierr;
+
+  if (x && (x->state % 2) == 1)
+    return CeedError(x->ceed, 1, "Cannot grant CeedVector array access, a process still has read/write access");
```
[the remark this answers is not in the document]
jeremylt: I agree. I updated the error messages.

jeremylt:
I added 3 tests, two of which are expected to fail, to verify that the state lock works as intended.

The vector test numbering now overlaps with the restriction test numbering, but that will be fixed in PR #118.

include/ceed-impl.h
```diff
@@ -66,6 +66,7 @@ struct CeedVector_private {
   int (*RestoreArrayRead)(CeedVector, const CeedScalar **);
   int (*Destroy)(CeedVector);
   int refcount;
+  int state;
```
jedbrown: Signed overflow is undefined and can happen with a 32-bit integer (a billion time steps is a lot, but not impossible). I think `uint64_t` would be preferable here.

ceed-vec.c
```diff
@@ -96,9 +99,14 @@ int CeedVectorSetValue(CeedVector x, CeedScalar value) {
 /// @sa CeedVectorRestoreArray()
 int CeedVectorGetArray(CeedVector x, CeedMemType mtype, CeedScalar **array) {
   int ierr;
+
+  if (x && (x->state % 2) == 1)
+    return CeedError(x->ceed, 1, "Cannot grant CeedVector array access, the access lock is already in use");
```
jedbrown: I think this check should also go in `CeedVectorGetArrayRead`.

ceed-vec.c
```diff
@@ -165,6 +174,9 @@ CEED_EXTERN int CeedVectorGetLength(CeedVector vec, CeedInt *length) {
 int CeedVectorDestroy(CeedVector *x) {
   int ierr;
 
+  if ((*x) && ((*x)->state % 2) == 1)
```
jedbrown: I think this can go after the check below. It's a more permissive policy, but I don't think we need to error if one user drops their reference while another accesses the array.

ceed-vec.c
jeremylt: I was thinking to allow for multiple read access. That shouldn't be problematic, should it?

jedbrown: Multiple read access is fine (it doesn't modify state), but read access should not be acquired while someone is modifying.

jeremylt: Ah, good point.

jeremylt:
We're good once Travis does its thing

jeremylt:
I think the CodeCov report is simply wrong. It is missing the coverage from tests that check the error messages. It is not a big deal with this PR but I worry that these sort of small errors could lead to an inaccurate picture of our coverage, especially if we add more error confirmation tests.

jedbrown:
The gcov output is written by an `atexit` handler. We need to use a gentler error handler in the test suite. https://stackoverflow.com/questions/20250400/how-can-i-use-gcov-even-when-a-segmentation-fault-happens

interface/ceed-vec.c
```diff
@@ -234,11 +268,17 @@ int CeedVectorDestroy(CeedVector *vec) {
   int ierr;
 
   if (!*vec || --(*vec)->refcount > 0) return 0;
+
+  if ((*vec) && ((*vec)->state % 2) == 1)
```
jedbrown: I forgot to check this in the revision. The `*vec` part isn't needed here because it was already checked above.

Issue #94, opened 2018-07-21
Add OCCA continuous integration


jedbrown:

jeremylt:
Closed by PR #121

PR #121, opened 2018-08-12
Add OCCA to Travis CI

jeremylt:
This PR adds OCCA, specifically `/cpu/occa` and `/omp/occa` to Travis CI. This PR addresses Issue #94.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,9 @@ env:
 install:
   - git clone --depth 1 https://github.com/mfem/mfem.git
   - make -C mfem -j2 serial MFEM_CXXFLAGS=-O
-  - export MFEM_DIR=$PWD/mfem PETSC_INSTALL=$HOME/install/petsc-3.9.3
+  - git clone --depth 1 https://github.com/libocca/occa.git
+  - make -C occa -j2
+  - export OCCA_DIR=$PWD/occa MFEM_DIR=$PWD/mfem PETSC_INSTALL=$HOME/install/petsc-3.9.3
   - test -d $PETSC_INSTALL
     || ( curl -O http://ftp.mcs.anl.gov/pub/petsc/release-snapshots/petsc-lite-3.9.3.tar.gz
         && tar xf petsc-lite-3.9.3.tar.gz


.travis.yml
```diff
@@ -39,3 +42,5 @@ cache:
   ccache: true
   directories:
   - $HOME/install
+  - $HOME/mfem
+  - $HOME/occa
```
jedbrown: Do we want these cached instead of relying on ccache?

jedbrown: With PETSc, we are using a release and have logic to skip the install entirely (no configure, no make, no install; just use `$HOME/install/petsc-3.9.3`). We do not have that for MFEM or OCCA, both of which are built in-source using just `make`. We should delete these two lines. The builds will still be fast due to ccache.

jeremylt: Updated. The run time is faster that I expected this time.

.travis.yml
```diff
@@ -17,6 +17,9 @@ env:
 install:
   - git clone --depth 1 https://github.com/mfem/mfem.git
   - make -C mfem -j2 serial MFEM_CXXFLAGS=-O
+  - git clone --depth 1 https://github.com/libocca/occa.git
+  - make -C occa -j2
+  - export OCCA_DIR=$PWD/occa LD_LIBRARY_DIR+=":${PWD}/lib"
```
jedbrown: What is `LD_LIBRARY_DIR`?

jedbrown: Also, that path doesn't look like it is anything. I expect you can delete that part.

jeremylt: Its in the OCCA build instructions. I sometimes have trouble locally when I skip it.

jeremylt: It's where `libocca.so` lives

jedbrown: `LD_LIBRARY_PATH` is in the build instructions. (`LD_LIBRARY_DIR` isn't a thing as far as I know.) LibCEED uses RPATH to link so the environment variable isn't needed. The library is in `$OCCA_DIR/lib`, not `$PWD/lib`.

jeremylt: Should be all good now. Build times are much more reasonable today

PR #123, opened 2018-08-13
Remove the additional memcpy

thilinarmtb:


ceed-fortran.c
```diff
@@ -79,25 +79,22 @@ void fCeedVectorSetArray(int *vec, int *memtype, int *copymode,
 }
 
 #define fCeedVectorGetArray FORTRAN_NAME(ceedvectorgetarray,CEEDVECTORGETARRAY)
-//TODO Need Fixing, double pointer
-void fCeedVectorGetArray(int *vec, int *memtype, CeedScalar *array, int *err) {
+void fCeedVectorGetArray(int *vec, int *memtype, CeedScalar *array, long *offset,
```
thilinarmtb: @jedbrown : Is the use of `long` here for offset okay? Looks like there is no `ssize_t` equivalent in
Fortran?

jeremylt:
Travis has segfaults on t02, t06, t07, and t08. I think all 4 of them deal with recent changes
t02 deals with setting a vector from a value, and in the test the vector didn't previously have an array
t06 and t08 are our tests for blocked restrictions, so the indices get permuted and padded
t07 is the identity restriction test, so using NULL as the index array

I might have forgotten something important in the Fortran interface for these functions?

ceed-fortran.c
jedbrown: `long` is only 32 bits on LLP64 (e.g., MS Windows). Eventually we'll need to run configure tests to determine sizes, but we would be fine using `int64_t` from `stdint.h` and matching it with 8-byte integer in Fortran.

jedbrown:
@thilinarmtb Please have a look at my rebased branch https://github.com/CEED/libCEED/tree/jed/fortran-api-array which tracks some recent changes that were not propagated to the Nek examples.

I was dealing with memory corruption and needed to modify `SIZE.in` (to reduce Nek's 2 GiB BSS segment that foils Valgrind). Can you look at the bp1 output to let us know if it's okay? If so, we need to apply a few similar changes to `bp3.usr`.

jeremylt:
When I try to run I get `malloc_consolidate(): invalid chunk size`. What did you change in `SIZE.in`?

jedbrown:
@jeremylt You probably got errors from bp3, which I hadn't updated yet. Try that branch again for a few more important bug fixes.

jeremylt:
Ok, now it runs.

jedbrown:
Superseded by #134. Please re-open this PR if you wish to consider it independently.

PR #124, opened 2018-08-15
test: remove /ocl/occa from BACKENDS

jedbrown:
Based on discussion in PR #101, we are temporarily disabling OpenCL
testing.  It should be possible to restore once these issues are
resolved:

  https://github.com/libocca/occa/issues/86
  https://github.com/libocca/occa/issues/166


jeremylt:
It looks like the tests significantly speed up without OCL.

PR #127, opened 2018-08-15
CeedVectorSetValue: fix doc and test prior to setting the array

jedbrown:
When backends provide a fast path, it would be easy to forget that it
is acceptable to set the value before setting or getting an array.

### DIFF
--- a/interface/ceed-vec.c
+++ b/interface/ceed-vec.c
@@ -76,7 +76,7 @@ int CeedVectorSetArray(CeedVector vec, CeedMemType mtype, CeedCopyMode cmode,
 }
 
 /**
-  @brief Set the array used by a CeedVector, allocating the array if applicable
+  @brief Set the CeedVector to a constant value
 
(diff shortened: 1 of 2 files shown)


jeremylt:
I think this is much more comprehensive. 

I don't know if you saw, I added in incrementing the state counter by 2 when `SetArray` and `SetValue` is called before we merged the last branch. Do we want to add some sort of check on vector 'staleness' in a similar sort of test? We might need to add something like `CeedVectorCheckState` to do so. It is the sort of thing that we eventually need when we make the backends go through an API, right?

jedbrown:
PR #120 was basically enforcing some constraints on user behavior. When backends want to use state, we should add interfaces for them to cache and whether the cache remains current. In PETSc, there are vector attributes that are held internally. I don't think we need to do it now.

jeremylt:
Just something to put on the to do list for later then.

PR #131, opened 2018-08-15
Check for env var CEED_ERROR_HANDLER

jeremylt:
Found a minor bug, code segfaults outside of testing because it was comparing "NULL" to "exit".

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -323,8 +323,9 @@ int CeedInit(const char *resource, Ceed *ceed) {
   }
   if (!matchlen) return CeedError(NULL, 1, "No suitable backend");
   ierr = CeedCalloc(1,ceed); CeedChk(ierr);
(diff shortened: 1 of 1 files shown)


interface/ceed.c
```diff
@@ -323,7 +323,7 @@ int CeedInit(const char *resource, Ceed *ceed) {
   }
   if (!matchlen) return CeedError(NULL, 1, "No suitable backend");
   ierr = CeedCalloc(1,ceed); CeedChk(ierr);
-  if (!strcmp(getenv("CEED_ERROR_HANDLER"), "exit"))
+  if (getenv("CEED_ERROR_HANDLER") && !strcmp(getenv("CEED_ERROR_HANDLER"), "exit"))
```
jedbrown: Better to assign to a variable to avoid calling `getenv` twice. It'll also be cleaner if we need to test for other names.

jeremylt: Updated

PR #134, opened 2018-08-16
Jed/fortran api array

jedbrown:
Fix various Fortran issues including no-copy array access and string length handling.

Supersedes #123.

### DIFF
--- a/examples/ceed/ex1.c
+++ b/examples/ceed/ex1.c
@@ -367,7 +367,7 @@ int BuildCartesianRestriction(Ceed ceed, int dim, int nxyz[3], int order,
                             ncomp, CEED_MEM_HOST,
                             CEED_COPY_VALUES, el_dof, restr);
   CeedElemRestrictionCreateIdentity(ceed, num_elem, elem_qpts,
-                                    elem_qpts*num_elem*ncomp,
+                                    elem_qpts*num_elem,
                                     ncomp, restr_i);
   free(el_dof);
   return 0;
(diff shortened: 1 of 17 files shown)


jedbrown:
@thilinarmtb Please check the Nek bp1 and bp3 output. I don't know how to verify if it is acceptable or not. It would be very useful to print and test an error so there is something simple and visible in the output.

jedbrown:
Likely #71. #84 is a Fortran fixed-form issue.

jeremylt:
`t304` is still failing unfortunately.

jeremylt:
Found it, it was an error in the test.

jeremylt:
@jedbrown the output I am getting from Nek looks like what I was seeing in the spring, but faster. However, I'll defer to @thilinarmtb 's opinion, as I am far from expert and I don't have my old log files to compare to.

I think we should be able to copy over mfem bp3 okl, but perhaps we want to wait until @dmed256 is happy with where OCCA is at for parsing pointers in the loops?

Is the plan to add Nek to CI? I am not sure if/how it would integrate nicely into the prove system we have now.

thilinarmtb:
@jedbrown: I am travelling today, will review this tonight.

thilinarmtb:
Looks like bp1 and bp3 are both broken although they could be run without segfaults now.
I will try to fix the two examples.

jedbrown:
@thilinarmtb Any progress? Do you want help?

PR #111, opened 2018-08-06
Magma dev optimizations

stomov:
This is the next level of performance optimizations in the magma backend. Now data stays on the GPU and there are no intermediate communications back and forth (as in the initial version that we used it just around some kernels to test them). This passes "make test" examples and examples/ceed/ex1.c.
Something to discuss is probably how to handle the QFunctions and if the current way OK for now.

PR #139, opened 2018-08-21
Fix and use QFunctionGetNumArgs

jeremylt:
This PR fixes "QFunctionGetNumArgs" to reflect the fact that we only allow one emode per field now and uses it to clean up the backend code. Eventually it would be nice to have similar functions for all private data backends may want to use.

Again, will be added to PR #111 separately.

With this PR, it looks like only the error handling and request functionality in the interface won't be fully tested.


jedbrown:
@jeremylt I think it's because the flags #117 are in 'master', but not in your branch. This codecov feature appears to be pretty unstable, but let's stick with it for a bit and maybe it'll look more reasonable for branches that have a common configuration.

jeremylt:
I could see that being a useful future feature that helps with reducing the total number of basis applies. I know our first version of NS will duplicate interp.

PR #141, opened 2018-08-23
PETSc BP3

jeremylt:
This is a first pass at a PETSc BP3 to clarify what PETSc solvers need.

### DIFF
--- a/examples/petsc/bp3.c
+++ b/examples/petsc/bp3.c
@@ -19,7 +19,7 @@
 //     bp3 -ceed /omp/occa
 //     bp3 -ceed /ocl/occa
 //
-//TESTARGS -ceed {ceed_resource} -test
+//TESTARGS -ceed {ceed_resource} -test -degree 3
 const char help[] = "Solve CEED BP3 using PETSc\n";
 
 #include <petscksp.h>
@@ -101,7 +101,9 @@ static int CreateRestriction(Ceed ceed, const CeedInt melem[3],
 typedef struct User_ *User;
 struct User_ {
   MPI_Comm comm;
-  VecScatter ltog;
+  VecScatter ltog;              // Scatter for all entries
+  VecScatter ltog0;             // Skip Dirichlet values
+  VecScatter gtogD;             // global-to-global; only Dirichlet values
   Vec Xloc, Yloc;
   CeedVector xceed, yceed;
   CeedOperator op;
@@ -109,7 +111,8 @@ struct User_ {
   Ceed ceed;
 };
 
-// This function uses libCEED to compute the action of the diffusion operator
+// This function uses libCEED to compute the action of the Laplacian with
+// Dirichlet boundary conditions
 static PetscErrorCode MatMult_Diff(Mat A, Vec X, Vec Y) {
   PetscErrorCode ierr;
   User user;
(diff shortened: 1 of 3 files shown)


examples/petsc/bp1.c
```diff
@@ -97,44 +98,6 @@ static int CreateRestriction(Ceed ceed, const CeedInt melem[3],
   PetscFunctionReturn(0);
 }
 
-static int Setup(void *ctx, CeedInt Q,
```
jeremylt: Stripped the QFunctions to a headder

examples/petsc/bp1.c
```diff
@@ -376,21 +340,21 @@ int main(int argc, char **argv) {
   CeedQFunctionAddInput(qf_setup, "dx", 3, CEED_EVAL_GRAD);
   CeedQFunctionAddInput(qf_setup, "weight", 1, CEED_EVAL_WEIGHT);
   CeedQFunctionAddOutput(qf_setup, "rho", 1, CEED_EVAL_NONE);
-  CeedQFunctionAddOutput(qf_setup, "target", 1, CEED_EVAL_NONE);
```
jeremylt: Renamed output "target" to "true_soln", added output "rhs"

examples/petsc/bp1.c
```diff
@@ -446,14 +406,24 @@ int main(int argc, char **argv) {
   ierr = MatShellSetOperation(mat, MATOP_MULT, (void(*)(void))MatMult_Mass);
   CHKERRQ(ierr);
 
+  // Get RHS vector
   ierr = MatCreateVecs(mat, &rhs, NULL); CHKERRQ(ierr);
-  ierr = VecZeroEntries(X); CHKERRQ(ierr);
```
jeremylt: Attempted to set up PETSc rhs vector, but something is wrong here

jeremylt: Really, I think I set up/used the PETSc vectors and scatter wrong.

examples/petsc/bp1.c
jeremylt: I'd later like to change the vector to actually be called 'true_soln' as well, but I was going to wait until after we got everything running.

jeremylt:
Update: The example compiles and runs, but I have something wrong still, as the error is too high.

jeremylt:
To do:
Make BP3 actually work
Follow BP convention of 2 extra qpts
Use double dash for args

jeremylt:
BP3 now converges but to the wrong answer

jeremylt:
I picked the frequency and offset for the target function to zero the BCs, which is different than the MFEM version. I think that setting zero BCs does more closely focus on the libCEED part of the work though.

I think this branch is in good shape now. I want to make a couple of consistency changes, but then we can probably merge.

jedbrown:
Do you think it makes sense to merge 164d7109079303b7ec02b71cafcc4350fe473cee (which implements Dirichlet boundary conditions, resulting in a nonsingular operator) into this branch? I kept all the `VecScatter`s in the interest of flexibility, but we could eliminate one if desired.

jeremylt:
That commit seems reasonable to me. I think comes down to what we want to do with this example.

As an example, 164d710 better shows a basic way to incorporate boundary terms. And it is more in line with the MFEM BP3.

For benchmarks, this version more directly shows the performance of just libCEED, but I don't think we are there yet.

jedbrown:
Did you mean to cite a different commit?

jeremylt:
I meant to cite yours. I think your commit gives a better example of boundary conditions rather than leaving it somewhat ambiguously with zero Neumann boundary conditions.

PR #143, opened 2018-08-26
Add testing whitelist

jeremylt:
This adds a testing whitelist to make backend developers explicitly list which tests they will not support

### DIFF
--- a/backends/magma/ceed-magma.c
+++ b/backends/magma/ceed-magma.c
@@ -554,6 +554,13 @@ static int CeedElemRestrictionCreate_Magma(CeedElemRestriction r,
   return 0;
 }
 
+static int CeedElemRestrictionCreate_Magma(CeedElemRestriction r,
+    CeedMemType mtype,
+    CeedCopyMode cmode,
+    const CeedInt *indices) {
+  return CeedError(r->ceed, 1, "Backend does not implement blocked restrictions");
+}
+
 // Contracts on the middle index
 // NOTRANSPOSE: V_ajc = T_jb U_abc
 // TRANSPOSE:   V_ajc = T_bj U_abc
@@ -1056,6 +1063,7 @@ static int CeedInit_Magma(const char *resource, Ceed ceed) {
   ceed->VecCreate = CeedVectorCreate_Magma;
   ceed->BasisCreateTensorH1 = CeedBasisCreateTensorH1_Magma;
   ceed->ElemRestrictionCreate = CeedElemRestrictionCreate_Magma;
+  ceed->ElemRestrictionCreateBlocked = CeedElemRestrictionCreateBlocked_Magma;
   ceed->QFunctionCreate = CeedQFunctionCreate_Magma;
   ceed->OperatorCreate = CeedOperatorCreate_Magma;
   return 0;
--- a/tests/tap.sh
+++ b/tests/tap.sh
@@ -36,35 +36,27 @@ for ((i=0;i<${#backends[@]}; ++i)); do
     # grep to skip test if backend cannot handle resource
     if grep -F -q -e 'backend cannot use resource' \
             -e 'OCCA backend failed' ${output}.err; then
-        printf "ok $i0 # SKIP $1 $backend\n"
-        printf "ok $i1 # SKIP $1 $backend stdout\n"
-        printf "ok $i2 # SKIP $1 $backend stderr\n"
+        printf "ok $i0 # SKIP - occa mode not supported $1 $backend\n"
+        printf "ok $i1 # SKIP - occa mode not supported $1 $backend stdout\n"
+        printf "ok $i2 # SKIP - occa mode not supported $1 $backend stderr\n"
         continue
     fi
 
-    # grep to skip test if backend cannot handle Blocked Restriction
-    if grep -F -q -e 'Backend does not support ElemRestrictionCreateBlocked' \
-            ${output}.err; then
-        printf "ok $i0 # SKIP $1 $backend\n"
-        printf "ok $i1 # SKIP $1 $backend stdout\n"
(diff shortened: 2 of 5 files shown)


tests/tap.sh
```diff
@@ -42,9 +42,9 @@ for ((i=0;i<${#backends[@]}; ++i)); do
         continue
     fi
 
-    # grep to skip test if backend cannot handle Blocked Restriction
-    if grep -F -q -e 'Backend does not support ElemRestrictionCreateBlocked' \
-            ${output}.err; then
+     # grep to skip test if backend chooses to whitelist test
+    if grep -F -q -e 'Backend does not support' ${output}.err \
+             && grep -F -q -e "$1 $backend" tests/whitelist ; then
```
jeremylt: I think it is a safer policy to force backends to explicitly list what functions they do not support so we don't get false passes from the test suite

tests/whitelist
```diff
@@ -0,0 +1,26 @@
...
+t202-elemrestriction /cpu/occa
+t202-elemrestriction /gpu/occa
+t202-elemrestriction /omp/occa
+t202-elemrestriction /ocl/occa
+t202-elemrestriction /gpu/magma
+t202-elemrestriction-f /cpu/occa
+t202-elemrestriction-f /gpu/occa
+t202-elemrestriction-f /omp/occa
+t202-elemrestriction-f /ocl/occa
+t202-elemrestriction-f /gpu/magma
```
jeremylt: I am listing the Fortran and C tests separately. I did it this way because we have had trouble with the Fortran tests not being the same as the C tests and certain backends not playing nice over Fortran.

tests/tap.sh
```diff
@@ -53,15 +53,15 @@ for ((i=0;i<${#backends[@]}; ++i)); do
 
     # grep to pass test t103 on error
     if grep -F -q -e 'Cannot grant CeedVector array access, the access lock is already in use' \
-            ${output}.err; then
+            ${output}.err && [[ "$1" = "t103"* ]] ; then
```
jeremylt: I explicitly added this check so we don't get false positives on other tests.

tests/tap.sh
jedbrown: Could we make this something that a backend implements instead of residing in a common whitelist (that will need to be edited and merged)? One solution would be for the backend to implement the function that it "does not support", but raise an error with a different syntax to indicate that it is intentional.

jeremylt:
This fails in Travis but passes with OCCA https://github.com/libocca/occa/commit/527494c949e95451f491c710592e9cd91e52b56f

backends/magma/ceed-magma.c
```diff
@@ -554,6 +554,13 @@ static int CeedElemRestrictionCreate_Magma(CeedElemRestriction r,
   return 0;
 }
 
+static int CeedElemRestrictionCreate_Magma(CeedElemRestriction r,
+    CeedMemType mtype,
+    CeedCopyMode cmode,
+    const CeedInt *indices) {
+  return CeedError(r->ceed, 1, "Backend does not impliment blocked restrictions");
```
jedbrown: spelling

tests/tap.sh
```diff
@@ -36,35 +36,27 @@ for ((i=0;i<${#backends[@]}; ++i)); do
     # grep to skip test if backend cannot handle resource
     if grep -F -q -e 'backend cannot use resource' \
             -e 'OCCA backend failed' ${output}.err; then
-        printf "ok $i0 # SKIP $1 $backend\n"
-        printf "ok $i1 # SKIP $1 $backend stdout\n"
-        printf "ok $i2 # SKIP $1 $backend stderr\n"
+        printf "ok $i0 # OCCA SKIP $1 $backend\n"
```
jedbrown: I believe [TAP](http://testanything.org/tap-specification.html#skipping-tests) says the `SKIP` must come first.

jeremylt: Correct, fixed

jeremylt:
Huh, I had thought TAP already tallied skips, but I guess not

jeremylt:
Perl might have intentionally moved away from explicitly listing the number of skipped tests in the summary result:
https://grokbase.com/t/perl/qa/092jekprma/test-skip-ing-weirdness

tests/tap.sh
```diff
@@ -36,35 +36,27 @@ for ((i=0;i<${#backends[@]}; ++i)); do
...
-    if grep -F -q -e 'Backend does not support ElemRestrictionCreateBlocked' \
-            ${output}.err; then
-        printf "ok $i0 # SKIP $1 $backend\n"
-        printf "ok $i1 # SKIP $1 $backend stdout\n"
-        printf "ok $i2 # SKIP $1 $backend stderr\n"
+     # grep to skip test if backend chooses to whitelist test
+    if grep -F -q -e 'Backend does not implement' \
+            ${output}.err ; then
+        printf "ok $i0 # TODO - not implemented $1 $backend\n"
+        printf "ok $i1 # TODO - not implemented $1 $backend stdout\n"
```
jedbrown: [Specification](http://testanything.org/tap-specification.html#todo-tests) says TODO should be marked `not ok`. I don't have a problem with prove's behavior, but I think we should try to follow the standard when using directives. There are other TAP consumers including Jenkins, which we might use to get CI on GPU or other DOE hardware/software stacks.

jeremylt: I mixed it up; I thought TODO was another type of conditional pass. Fixed now.

jeremylt:
Code coverage went down because lines that return an error code that gets the test listed as skipped are not counted as covered.

jedbrown:
We aren't slaves to the coverage overlords.

jeremylt:
I wanted to at least set as a norm that we acknowledge and understand why coverage goes down and then we can overrule the codecov check if it makes sense.

PR #144, opened 2018-08-26
[WIP] Additional Test against tagged release of OCCA

jeremylt:
This branch adds testing against a stable branch of OCCA to check if failures in Travis are due to OCCA changes or libCEED regressions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,13 @@ addons:
     - liblapack-dev
 
 env:
-  - FC=gfortran
+  - FC=gfortran OCCA_BRANCH="v1.0.5"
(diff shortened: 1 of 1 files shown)


jeremylt:
We don't have a current stable version of OCCA that works with libCEED

jedbrown:
We can use a specific commit hash.  It doesn't need to be a branch or tag.

jeremylt:
Right, but then we have to clone much more of the repo, to an unknown depth and then reset. That felt like a lot of information to drag around in the cache.

jedbrown:
It isn't cached, it's just a bit slower to clone. The whole repository is 15 MB so it isn't a huge issue.

PR #145, opened 2018-08-27
Test against specific commit of OCCA

jeremylt:
This PR adds checking against a specific version of OCCA to check for failure due to changes in OCCA.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,14 @@ addons:
     - liblapack-dev
 
 env:
-  - FC=gfortran
+  - FC=gfortran OCCA_HEAD=HEAD
+  - FC=gfortran OCCA_HEAD=527494c
 
 install:
   - git clone --depth 1 https://github.com/mfem/mfem.git
   - make -C mfem -j2 serial MFEM_CXXFLAGS=-O
-  - git clone --depth 1 https://github.com/libocca/occa.git
+  - git clone https://github.com/libocca/occa.git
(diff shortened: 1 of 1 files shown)


jeremylt:
And we have the green checkmark again!

jeremylt:
I think this is good solution. Thoughts?

screenshot from 2018-08-27 10-54-13

jedbrown:
That's okay with me. We can collapse once the issue is fixed in Occa. Is there an issue in libocca/occa yet? https://github.com/libocca/occa/issues

jeremylt:
There was an OCCA issue, but it got closed for some reason.

jedbrown:
[The issue](https://github.com/libocca/occa/issues/198) so we don't lose track.

.travis.yml
```diff
@@ -12,12 +12,22 @@ addons:
     - liblapack-dev
 
 env:
-  - FC=gfortran
+  - FC=gfortran OCCA_HEAD=HEAD
+  - FC=gfortran OCCA_HEAD=527494c
```
jedbrown: Surely the second line overwrites the first.

jeremylt: It is like the compiler specification; the build matrix is a combination of the different lines in runtime, environment, and exclusions/inclusions. So we get 2 compiler options x 2 env options = 4 runs

https://docs.travis-ci.com/user/customizing-the-build/#build-matrix

PR #147, opened 2018-08-29
Rename /cpu/self/opt to /cpu/self/vec

jeremylt:
This PR renames `cpu/self/opt` to `cpu/self/vec`, to be more specific and to align with the fact that there will be other optimized CPU backends.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -800,7 +800,7 @@ INPUT                  = . \
                          examples/petsc \
                          backends/ref \
                          backends/occa \
-                         backends/optimized \
+                         backends/blocked \
                          backends/tmpl \
                          backends/magma \
                          tests \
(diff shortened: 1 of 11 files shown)


jeremylt:
`git grep -i /opt' is clean, as is 'git grep -i optimized' and 'git grep -i _Opt"

jeremylt:
Is there a different name that makes more sense to you? I am thinking ahead to making the libXSMM background which should also be an optimized CPU backend, so I'd like to have a name that doesn't imply that this is *the* optimized CPU backend when there are multiple

jeremylt:
I switched to `bch` or `batched` for multi-element batches. I could also use `blocked`. I figure it is as good as any name, but better than `opt`. If this new name is fine by you, I'll merge.

jedbrown:
My only concern is that "batch" is not a term of art in libCEED while "block" is. I would skip the cryptic abbreviations `bch`.

jeremylt:
All set. I can merge it in the morning.

PR #148, opened 2018-08-29
Standardize Backend Function Args, Rework Delegate Mechanism

jeremylt:
This branch is in two parts

- First I standardized the ordering and arguments for backend functions, removing the superfluous 'ceed' argument for bases and vectors
- Second I reworked the `tmpl` mechanism to refer to an optional delegate ceed created at `CeedInit` rather than a held reference ceed. This change allows for simple multi level dispatching (e.g. `tmpl` dispatches to `opt` which dispatches to `ref` for `VectorCreate`)

One flaw with this to fix, this currently does not support mixed level delegation.

### DIFF
--- a/backends/blocked/ceed-blocked-basis.c
+++ b/backends/blocked/ceed-blocked-basis.c
@@ -22,7 +22,7 @@
 // NOTRANSPOSE: V_ajc = T_jb U_abc
 // TRANSPOSE:   V_ajc = T_bj U_abc
 // If Add != 0, "=" is replaced by "+="
-static int CeedTensorContract_Blocked(CeedInt A, CeedInt B, CeedInt C, CeedInt J,
+static int CeedTensorContract_Blocked(Ceed ceed, CeedInt A, CeedInt B, CeedInt C, CeedInt J,
                                   const CeedScalar *restrict t,
                                   CeedTransposeMode tmode, const CeedInt Add,
                                   const CeedScalar *restrict u, CeedScalar *restrict v) {
@@ -73,7 +73,7 @@ static int CeedBasisApply_Blocked(CeedBasis basis, CeedInt nelem,
     CeedInt pre = ncomp*CeedIntPow(P, dim-1), post = nelem;
     CeedScalar tmp[2][nelem*ncomp*Q*CeedIntPow(P>Q?P:Q, dim-1)];
     for (CeedInt d=0; d<dim; d++) {
-      ierr = CeedTensorContract_Blocked(pre, P, post, Q,
+      ierr = CeedTensorContract_Blocked(basis->ceed, pre, P, post, Q,
                                     basis->interp1d, tmode, add&&(d==dim-1),
                                     d==0?u:tmp[d%2], d==dim-1?v:tmp[(d+1)%2]);
       CeedChk(ierr);
@@ -97,7 +97,7 @@ static int CeedBasisApply_Blocked(CeedBasis basis, CeedInt nelem,
     // Interpolate to quadrature points (NoTranspose)
     //  or Grad to quadrature points (Transpose)
     for (CeedInt d=0; d<dim; d++) {
-      ierr = CeedTensorContract_Blocked(pre, P, post, Q,
+      ierr = CeedTensorContract_Blocked(basis->ceed, pre, P, post, Q,
                                     (tmode == CEED_NOTRANSPOSE
                                      ? basis->interp1d
                                      : impl->colograd1d),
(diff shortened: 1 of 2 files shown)


backends/optimized/ceed-opt-basis.c
```diff
@@ -22,8 +22,7 @@
 // NOTRANSPOSE: V_ajc = T_jb U_abc
 // TRANSPOSE:   V_ajc = T_bj U_abc
 // If Add != 0, "=" is replaced by "+="
-static int CeedTensorContract_Opt(Ceed ceed,
```
jedbrown: My philosophy here was that the `ceed` is needed to raise an error if we have reason to do so.

backends/optimized/ceed-opt.c
```diff
@@ -20,10 +20,8 @@
 
 static int CeedDestroy_Opt(Ceed ceed) {
   int ierr;
-  Ceed_Opt *impl = ceed->data;
-  Ceed ceedref = impl->ceedref;
-  ierr = CeedFree(&ceedref); CeedChk(ierr);
-  ierr = CeedFree(&impl); CeedChk(ierr);
+  Ceed ceedref = ceed->delegate;
```
jedbrown: Since the top-level interface sees the delegate, I think it can handle destruction.

interface/ceed-basis.c
```diff
@@ -211,7 +217,7 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt ncomp,
   ierr = CeedMalloc(dim*Q*P,&(*basis)->grad1d); CeedChk(ierr);
   memcpy((*basis)->interp1d, interp, Q*P*sizeof(interp[0]));
   memcpy((*basis)->grad1d, grad, dim*Q*P*sizeof(grad[0]));
-  ierr = ceed->BasisCreateH1(ceed, topo, dim, P, Q, interp, grad, qref,
+  ierr = ceed->BasisCreateH1(topo, dim, P, Q, interp, grad, qref,
```
jedbrown: Is it interesting to pass all these arguments along when they have already been set on the `CeedBasis` above? Seems like maybe we should only do that for args that only the backend can handle (e.g., with `CeedMemType`).

interface/ceed.c
```diff
@@ -335,6 +335,27 @@ int CeedInit(const char *resource, Ceed *ceed) {
...
+  @brief Retrieve a delegate CEED
+
+  @param ceed Address of Ceed to retrieve delegate of,
+              address to save delegate to if one exists
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Utility
+**/
+int CeedGetDelegate(Ceed *ceed) {
```
jedbrown: Overwriting the input argument is weird. What if the caller doesn't want to overwrite?

interface/ceed.c
```diff
@@ -335,6 +335,27 @@ int CeedInit(const char *resource, Ceed *ceed) {
...
+  int ierr;
+
+  if ((*ceed)->delegate) {
+    (*ceed)=(*ceed)->delegate;
+    ierr = CeedGetDelegate(ceed); CeedChk(ierr);
```
jedbrown: I don't think we want to run all the way to the end of the list here.

interface/ceed.c
jeremylt: Fixed

interface/ceed.c
jeremylt: Fixed

interface/ceed-basis.c
jeremylt: Probably not. I left it to make the first commit simpler but I can trim other arguments too. Eventually I'd like the backends to not reach directly into the private struct to get that data, but that's not related to this PR

backends/optimized/ceed-opt.c
jeremylt: Good point

backends/optimized/ceed-opt-basis.c
jeremylt: Restored

interface/ceed-basis.c
jeremylt: Maybe in that 'to come soonish' PR I make a round of getters for those values and convert the backends over to using them while also only passing arguments that are needed. Those two make sense together for me.

jeremylt:
CodeCov goes down because this PR removes 370 lines of code

jeremylt:
@jedbrown, I think I made all of your requested updates, but please let me know if more needs to be done.

jedbrown:
@jeremylt I reset this branch because the rebased tree (that you just pushed this evening) was the same as I had in the merge that I had been testing.

PR #151, opened 2018-09-12
Fix tap.sh and tmpl

jeremylt:
I caught a bad merge/rebase on my part due to funny looking output from Travis. I fixed the part of tap.sh that let this mistake through.

### DIFF
--- a/tests/tap.sh
+++ b/tests/tap.sh
@@ -34,8 +34,7 @@ for ((i=0;i<${#backends[@]}; ++i)); do
     (build/$1 ${args/\{ceed_resource\}/$backend} || false) > ${output}.out 2> ${output}.err
     status=$?
     # grep to skip test if backend cannot handle resource
-    if grep -F -q -e 'backend cannot use resource' \
-            -e 'OCCA backend failed' ${output}.err; then
+    if grep -F -q -e 'OCCA backend failed to use' ${output}.err; then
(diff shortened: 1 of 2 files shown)


tests/tap.sh
```diff
@@ -34,8 +34,8 @@ for ((i=0;i<${#backends[@]}; ++i)); do
     (build/$1 ${args/\{ceed_resource\}/$backend} || false) > ${output}.out 2> ${output}.err
     status=$?
     # grep to skip test if backend cannot handle resource
-    if grep -F -q -e 'backend cannot use resource' \
-            -e 'OCCA backend failed' ${output}.err; then
+    if grep -F -q -e 'backend cannot use resource' ${output}.err | \
+            grep -F -q -e 'OCCA backend failed' ${output}.err; then
```
jedbrown: Why this change? Note that this pipe ignores the first command entirely while the former code asks whether either pattern appears in the file.

jeremylt: Without this the check for the 'OCCA backend failed' was being ignored; the OCCA not supported message was being printed whenever the error message contained "backend cannot use resource" regardless of if it contains "OCCA backend failed'. I am trying to force tap to check for both.

jedbrown: Then use `&&` between those two `grep` invocations or use `awk` as in https://unix.stackexchange.com/questions/55359/how-to-run-grep-with-multiple-and-patterns

PR #152, opened 2018-09-16
Explicit time stepping NS solver

jeremylt:
This PR is for discussing the basic Navier-Stokes example implemented in PETSc. We may wish to include this as our first multi-field example using libCEED.

Some details:
- Basic hex mesh
- No slip, non-penetrating BCs for momentum
- Reflecting BCs for density, energy

Parameter values:
```
  CeedScalar Rd         = 0.287;
  CeedScalar Ts         = 300.;
  CeedScalar p0         = 100000.;
  CeedScalar cp         = 1.004;
  CeedScalar cv         = 0.717;
  CeedScalar g          = 9.81;
  CeedScalar lambda     = -2./3.;
  CeedScalar mu         = 73.;
  CeedScalar k          = 0.02638;
  CeedScalar Pr         = cp*mu/k;
```


jeremylt:
Clang is intermittently failing to detect `PETSC_ARCH` for some reason unrelated to this PR.

jedbrown:
I'm not sure how you would assess that. Clang has no knowledge of `PETSC_ARCH` and indeed shouldn't depend on it in any way.

jeremylt:
I was a bit ambiguous with my phrasing. The Clang build of Travis intermittently fails on this branch with the message `Makefile:51: *** PETSc config not found. Please set PETSC_DIR and PETSC_ARCH.  Stop.` I would guess it's an issue with 'PETSC_ARCH' because 'PETSC_DIR' is explicitly set in Travis on line 40. It's only the Clang build that this occurs on, and it is not consistently occurring.

jedbrown:
It must occasionally run in an environment where the cached `PETSC_INSTALL` is broken. Could change the `.travis.yml` to check for `PETSc.pc` instead of just the existence of a directory.

jeremylt:
Ok - this passes Travis now.

It is still a WIP. I am not sure what we would want the default output method to be and I'd like to verify somehow that the results are correct first.

jeremylt:
We will want to eventually kill this last commit, but I switched to dE/dt + div(E u) = 0, and when running with`./ns -local 100000 -degree 5 -qextra 2` I think I see element edges start to show up. Another problem I noted is that I have to increase the initial velocity (see lines 205, 206 of `ns.h`) whenever I increase resolution, which feels wrong to me.

jeremylt:
I megred the AVX backend so we can run faster tests. For this to work on OCCA we will wil need to merge a fix from the nek-test branch (PR #169).

I think we are pretty close to squashing this down to a commit or two and merging this post #182 and #169.

jeremylt:
I made a branch that is our work squashed down to one commit:
https://github.com/CEED/libCEED/tree/ns-example-rebase

The example bulids and runs, so we should be able to reset this branch to match that branch to clean up the commit history.

jeremylt:
I cleaned up the commit history but kept a branch with the full history just in case:

'clean' copy: https://github.com/CEED/libCEED/tree/ns-example
working copy: https://github.com/CEED/libCEED/tree/ns-example-working

jeremylt:
Travis failures are due to a Travis problem, not due to a problem with this branch.

valeriabarra:
Updated branches:
'clean' copy: https://github.com/CEED/libCEED/tree/ns-example-rebase
working copy: https://github.com/CEED/libCEED/tree/ns-example

jeremylt:
@jedbrown, I think this is ready to merge thanks to @valeriabarra's hard work!

jedbrown:
Let's use OPT from the top level makefile. We could probably put -fopenmp-simd there.

PR #153, opened 2018-09-16
Check for PETSc.pc in travis cache

jeremylt:
This branch checks for `PETSc.pc` rather than just the PETSc directory in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,11 +29,11 @@ install:
   - git clone https://github.com/libocca/occa.git
   - git -C occa reset --hard $OCCA_HEAD
   - make -C occa -j2
-  - export OCCA_DIR=$PWD/occa MFEM_DIR=$PWD/mfem PETSC_INSTALL=$HOME/install/petsc-3.9.3
-  - test -d $PETSC_INSTALL
-    || ( curl -O http://ftp.mcs.anl.gov/pub/petsc/release-snapshots/petsc-lite-3.9.3.tar.gz
-        && tar xf petsc-lite-3.9.3.tar.gz
-        && cd petsc-3.9.3
+  - export OCCA_DIR=$PWD/occa MFEM_DIR=$PWD/mfem PETSC_INSTALL=$HOME/install/petsc-3.10.0
+  - test -s "$PETSC_INSTALL/lib/pkgconfig/PETSc.pc"
+        || ( curl -O http://ftp.mcs.anl.gov/pub/petsc/release-snapshots/petsc-lite-3.10.0.tar.gz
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,8 +30,8 @@ install:
   - git -C occa reset --hard $OCCA_HEAD
   - make -C occa -j2
   - export OCCA_DIR=$PWD/occa MFEM_DIR=$PWD/mfem PETSC_INSTALL=$HOME/install/petsc-3.9.3
-  - test -d $PETSC_INSTALL
-    || ( curl -O http://ftp.mcs.anl.gov/pub/petsc/release-snapshots/petsc-lite-3.9.3.tar.gz
+  - test -z | find $PETSC_INSTALL -name PETSc.pc
```
jedbrown: Just specify the path `test -s "$PETSC_INSTALL/lib/pkgconfig/PETSc.pc"`.

jedbrown: We could upgrade to PETSc 3.10 at any time. It shouldn't have any effect (just a slower Travis build the first time after upgrading).

jeremylt:
In any case, it did not help https://travis-ci.org/CEED/libCEED/jobs/429350975

jedbrown:
Try `test -z | find /any/valid/path -name file-not-found-there` in your shell. It succeeds because `find` does not report a failure if it doesn't find any matching files. Note that `test -z` takes its string argument on the command line (not by pipe (and the pipe would run backward for that to work). In bash syntax, you could write `test -z $(find ...)`, but the logic for that to rebuild is backward. You could do `test -n $(find ...)` which would have suitable logic, but is more fragile than the command I suggested.

.travis.yml
```diff
@@ -30,10 +30,10 @@ install:
...
   - make -C occa -j2
   - export OCCA_DIR=$PWD/occa MFEM_DIR=$PWD/mfem PETSC_INSTALL=$HOME/install/petsc-3.9.3
-  - test -d $PETSC_INSTALL
-    || ( curl -O http://ftp.mcs.anl.gov/pub/petsc/release-snapshots/petsc-lite-3.9.3.tar.gz
-        && tar xf petsc-lite-3.9.3.tar.gz
-        && cd petsc-3.9.3
+  - test -s "$PETSC_INSTALL/lib/pkgconfig/PETSc.pc"
+        || ( curl -O http://ftp.mcs.anl.gov/pub/petsc/release-snapshots/petsc-lite-3.10.tar.gz
+        && tar xf petsc-lite-3.10.tar.gz
+        && cd petsc-3.10
```
jedbrown: Need 3.10 here.

PR #158, opened 2018-09-19
Fixed Nek5000 BP1

thilinarmtb:
Now BP1 with CEED operator converges to the given tolerance.

@jedbrown @jeremylt : Please have a look.

### DIFF
--- a/examples/nek5000/bp1.usr
+++ b/examples/nek5000/bp1.usr
@@ -825,7 +825,7 @@ c     Create ceed basis for mesh and computation
       p=nx1
       q=p+1
       ncomp=1
-      call ceedbasiscreatetensorh1lagrange(ceed,ndim,ncomp,q,q,
+      call ceedbasiscreatetensorh1lagrange(ceed,ndim,ndim,q,q,
      $  ceed_gauss,basisx,err)
       call ceedbasiscreatetensorh1lagrange(ceed,ndim,ncomp,p,q,
      $  ceed_gauss,basisu,err)
--- a/examples/nek5000/bp1.usr
+++ b/examples/nek5000/bp1.usr
@@ -769,7 +769,7 @@ c-----------------------------------------------------------------------
 
       integer p,q,ncomp,edof,ldof
       integer basisu,basisx
-      integer erstrctu,erstrctx,erstrctw
+      integer erstrctu,erstrctx
       integer qdata
       integer qf_mass,qf_setup
       integer op_mass,op_setup
@@ -780,7 +780,7 @@ c-----------------------------------------------------------------------
 
 c     Init ceed library
       call get_spec(spec)
-      call ceedinit(trim(spec)//char(0),ceed,err)
+      call ceedinit(trim(spec),ceed,err)
 
       call geodat1n  ! Set up gf() arrays
 
@@ -825,7 +825,7 @@ c     Create ceed basis for mesh and computation
       p=nx1
       q=p+1
       ncomp=1
-      call ceedbasiscreatetensorh1lagrange(ceed,ndim,ndim,q,q,
+      call ceedbasiscreatetensorh1lagrange(ceed,ndim,ncomp,q,q,
      $  ceed_gauss,basisx,err)
       call ceedbasiscreatetensorh1lagrange(ceed,ndim,ncomp,p,q,
      $  ceed_gauss,basisu,err)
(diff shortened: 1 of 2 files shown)


examples/nek5000/bp1.usr
```diff
@@ -825,23 +825,21 @@ c     Create ceed basis for mesh and computation
       p=nx1
       q=p+1
       ncomp=1
-      call ceedbasiscreatetensorh1lagrange(ceed,ndim,ndim,q,q,
```
jeremylt: It doesn't matter in this case, as you aren't creating or using the coordinates, but basisx should have `ndim` fields, i.e. x,y,z has 3 fields.

examples/nek5000/bp1.usr
```diff
@@ -868,15 +866,15 @@ c     Create a ceed operator
...
       call ceedoperatorsetfield(op_setup,'rho',
-     $  erstrctw,ceed_basis_collocated,
+     $  erstrctx,ceed_basis_collocated,
      $  ceed_vector_active,err)
 
       call ceedoperatorcreate(ceed,qf_mass,
      $  ceed_null,ceed_null,op_mass,err)
       call ceedoperatorsetfield(op_mass,'rho',
-     $  erstrctw,ceed_basis_collocated,
+     $  erstrctx,ceed_basis_collocated,
```
jeremylt: This works here, but in BP3 a separate restriction for the qdata will be needed

examples/nek5000/bp1.usr
```diff
@@ -955,27 +961,28 @@ c     Helmholtz equations
...
+      integer i,e
 
       call ceedvectorsetarray(vec_p1,ceed_mem_host,ceed_use_pointer,
      $  p1,err)
+      call ceedvectorsetarray(vec_ap1,ceed_mem_host,ceed_use_pointer,
+     $  ap1,err)
+
       call ceedoperatorapply(op_mass,vec_p1,vec_ap1,
      $  ceed_request_immediate,err)
-      call ceedvectorgetarray(vec_ap1,ceed_mem_host,ap1,offset,err)
```
jeremylt: Have you had a chance to check on GPU yet? @jedbrown and I were talking, and the library likely needs a function such as `CeedVectorSync` to force a device to host update in situations like this.

examples/nek5000/run-nek-example.sh
```diff
@@ -31,7 +31,7 @@ fi
 # Set defaults for the parameters
 : ${CEED_DIR:=`cd ../../; pwd`}
 nek_examples=(bp1 bp3)
-nek_spec=/cpu/self
+nek_spec=/cpu/self/ref
```
jeremylt: `/cpu/self/ref` isn't the recommended CPU backend. I'd prefer this to be set to `/cpu/self/blocked` or `/cpu/self` (which defaults to `/cpu/self/blocked`)

jeremylt:
Now that this runs successfully, I added Issue #159 to add an Nek5000 test

jeremylt:
Anything I can clarify/help with these comments? It would be nice to have working Nek examples in the release this week.

examples/nek5000/bp1.usr
thilinarmtb: Yes, I agree. Otherwise, I don't think this code will run on the GPU.

jeremylt: I think it makes sense to put this in a later PR. I'll make an issue for this frontend feature.

examples/nek5000/bp1.usr
jeremylt: ^ This was just a FYI comment

examples/nek5000/bp1.usr
thilinarmtb: Can you explain further? I don't quite understand.

jeremylt: In PETSc `bp1.c` lines 320-334 we set up the coordinates vector. It requires 3 components, one for the x coordinate, one for the y coordinate, and one for the z coordinate.

The `ncomp` field in this basis is not ever used, as this example never actually creates a coordinate vector in libCEED and this basis never acts on a coordinate vector, so this field does not have an effect, but technically the coordinates would have 3 components in 3D.

I set `ncomp` to `ndim` when I converted this example in case a future user uses this example as a baseline to make a libCEED Fortran code on a more complex mesh. In that (future) case, that user would need to set `ncomp` to `ndim`, so I modeled that for them.

I'm not sure if I explained it better, so please let me know if I can be more clear!

PR #165, opened 2018-10-15
Update Nek BP3

jeremylt:
This PR updates Nek BP3 and attempts to update the OKL files. The output looks ok to me, but I am not a Nek expert.

You can test the OKLs with `/cpu/occa`.


jedbrown:
Tests passed as far as I could tell so I merged it. Let us know if you notice an issue @thilinarmtb.

PR #167, opened 2018-10-18
Add Getters

jeremylt:
This is a work in progress branch adding getters for the private Ceed object struct data, allowing us to make changes to the interface without changing all backends.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,4 +1,4 @@
-# libCEED Code of Conduct
+# libCEED: Code of Conduct
 
 ## Our Pledge
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# How to Contribute
+# libCEED: How to Contribute
 
 Contributions to libCEED are encouraged.
 <!---
--- a/Doxyfile
+++ b/Doxyfile
@@ -266,7 +266,7 @@ OPTIMIZE_OUTPUT_JAVA   = NO
 # sources. Doxygen will then generate output that is tailored for Fortran.
 # The default value is: NO.
 
-OPTIMIZE_FOR_FORTRAN   = NO
+OPTIMIZE_FOR_FORTRAN   = YES
 
 # Set the OPTIMIZE_OUTPUT_VHDL tag to YES if your project consists of VHDL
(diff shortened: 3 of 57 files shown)


jeremylt:
CodeCov is not happy. The report is providing some interesting highlights about functionality we are not using at this time, such as:

Vector:        state (will use in blocked Operator)
Basis:         qref
OFunction: vector length
Operator:    dqf and dqfT (we knew that)

Still to do:
- Update OCCA backend
- Update MAGMA backend (after PR #111)
- Make getters for QF and Operator fields
- Make getters for `void *data` information
- Make setters for the various functions

End Goal: Remove `#include <ceed-impl.h>` from all backend code

jedbrown:
Why is a getter for Vector state needed?

We should probably remove `dqf` and `dqfT` in favor of a to-be-defined interface to ask for the derivative of a QFunction with respect to a subset of its inputs. (We'll also want derivatives with respect to passive fields if we use libCEED for inference problems.)

jeremylt:
I was thinking that a getter for the state would be good for the blocked backend. If the current permuted evec is up to date, then the blocked backend should not go through the effort of reapplying the restriction.

Good point about dqf and dqfT.

I had the same concern about a single generic function to set backend functions. It feels like a tradeoff between reducing the number of setters and type-safety. I imagine that by the time someone is working on backends we can probably expect them to be able to trace an error like that.

jeremylt:
Of note, CeedChk and malloc/free are in ceed-impl.h. I think to full keep the interface structs private from backends we should split ceed-impl.h, maybe into ceed-impl.h and ceed-interface.h? Thoughts @jedbrown?

jedbrown:
`ceed-backend.h` to provide interfaces for backends that are not needed by ordinary users?

jeremylt:
I like that. I separated the headers out into 3. I made the split work with the current code. I think I want to think more about what goes in backend vs impl though.

jeremylt:
Status:

- This code has getters for everything but the user QFunction function pointer
- I want to push off setters for backend functions to a later PR
- I want to update the MAGMA backend to use these getters after PR #111 (@stomov)

procode200:
ok

jeremylt:
I added in a bunch of small documentation fixes, some related and others just added because I was fixing documentation anyways.

PR #168, opened 2018-10-26
Add OSX to Travis

jeremylt:
This PR addes osx testing to Travis, fixing issue #95.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: c
 
+os:
+  - linux
+  - osx
+
 compiler:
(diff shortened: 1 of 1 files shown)


jeremylt:
I sped this up with ccache on OSX. The slowest part is now building MFEM and OCCA on OSX.

Linux: 2:00-2:30 with 4 builds (gcc vs clang, OCCA head vs OCCA previous)
OSX: 6:00-7:00 with 2 builds (gcc vs clang, OCCA previous)

PR #170, opened 2018-10-30
Add LMode field to CeedOperatorSetField

jeremylt:
This PR addes the LMode field to CeedOperatorSetField. Previously lmode=CEED_NOTRANSPOSE was hardcoded into CeedOperatorApply.

This PR should follow PR #167.

### DIFF
--- a/backends/occa/ceed-occa-restrict.okl
+++ b/backends/occa/ceed-occa-restrict.okl
@@ -81,8 +81,31 @@
     const int rngN = toffsets[i+1];
     for (int d = 0; d < ncomp; ++d) {
       double value = 0.0;
-      for (int j=rng1; j<rngN; ++j)
-        value += uu[d+tindices[j]*ncomp];
+      for (int j=rng1; j<rngN; ++j) {
+        int n = tindices[j] % elemsize;
+        int e = tindices[j] / elemsize;
+        value += uu[(e*ncomp + d)*elemsize + n];
+      }
+      vv[d*ndof+i] = value;
+    }
+  }
+}
+// *****************************************************************************
+@kernel void kRestrict5b(const int ncomp,
+                         const int *tindices,
+                         const int *toffsets,
+                         const double* uu,
+                         double* vv) {
+  for (int i=0; i<ndof; i++; @tile(TILE_SIZE,@outer,@inner)){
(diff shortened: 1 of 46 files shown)


interface/ceed-operator.c
```diff
@@ -93,8 +93,8 @@ int CeedOperatorCreate(Ceed ceed, CeedQFunction qf, CeedQFunction dqf,
   @ref Basic
 **/
 int CeedOperatorSetField(CeedOperator op, const char *fieldname,
-                         CeedElemRestriction r, CeedBasis b,
-                         CeedVector v) {
+                         CeedElemRestriction r, CeedTransposeMode lmode,
```
jedbrown: Documentation above needs to be updated to include and explain the new parameter.

jeremylt: Updated

jedbrown:
Sorry about the slow review. Can we have at least one unit test that uses `CEED_TRANSPOSE` (i.e., interlaced) fields? I think it works (at least with the CPU backends), but it's important to have something in between the scalar unit tests and the Navier-Stokes example.

jeremylt:
I can cook up a variant of t501 that does this pretty easily - I'm do it right after the setters PR is up.

backends/occa/ceed-occa-restrict.okl
```diff
@@ -87,3 +87,20 @@
     }
   }
 }
+// *****************************************************************************
+@kernel void kRestrict5b(const int ncomp,
```
jeremylt: @camierjs if you get a moment, can you look at this? I just wrote the first test for a restriction with interlaced elements and OCCA's kRestrict5b (I just added this) and/or kRestrict2 to handle restrictions with lmode==CEED_TRANSPOSE aren't behaving correctly.

jeremylt:
Update, OCCA doesn't handle multifield transpose restrictions correctly at all. This is just the first time we have tested it. I should have a fix soon.

jeremylt:
Fixed, ready to merge if you are happy with the new unit tests

backends/occa/ceed-occa-restrict.okl
jeremylt: I got it

PR #172, opened 2018-11-14
Setters

jeremylt:
I need to also update MAGMA a little, but this should be good.

### DIFF
--- a/interface/ceed-qfunction.c
+++ b/interface/ceed-qfunction.c
@@ -77,6 +77,8 @@ int CeedQFunctionCreateInterior(Ceed ceed, CeedInt vlength,
   ierr = CeedCalloc(strlen(focca)+1, &focca_copy); CeedChk(ierr);
   strcpy(focca_copy, focca);
   (*qf)->focca = focca_copy;
+  ierr = CeedCalloc(16, &(*qf)->inputfields); CeedChk(ierr);
+  ierr = CeedCalloc(16, &(*qf)->outputfields); CeedChk(ierr);
   ierr = ceed->QFunctionCreate(*qf); CeedChk(ierr);
   return 0;
 }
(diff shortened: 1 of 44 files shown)


interface/ceed.c
```diff
@@ -331,7 +337,41 @@ int CeedInit(const char *resource, Ceed *ceed) {
     (*ceed)->Error = CeedErrorAbort;
   (*ceed)->refcount = 1;
   (*ceed)->data = NULL;
+
+  // Set lookup table
+  foffset foffsets[25] = {
```
jeremylt: I will turn this `25` into a global constant.

interface/ceed.c
```diff
@@ -365,10 +405,50 @@ int CeedSetDelegate(Ceed ceed, Ceed *delegate) {
...
+    strcat(lookupname, fname);
+  }
+
+  // Find and use offset
+  for (CeedInt i = 0; i < 25; i++) {
+    if (!strcmp(ceed->foffsets[i].fname, lookupname)) {
+      size_t offset = ceed->foffsets[i].offset;
+      size_t *fpointer;
+      fpointer = (size_t *)(object + offset);
+      *fpointer = (size_t) f;
```
jeremylt: We don't have checking that the function has the desired inputs, but we never did at this point (setting the pointer).

interface/ceed-qfunction.c
```diff
@@ -220,6 +228,22 @@ int CeedQFunctionGetFOCCA(CeedQFunction qf, char* *focca) {
...
+
+  @param qf              CeedQFunction
+  @param[out] f          Variable to store user function
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+
+int CeedQFunctionGetUserFunction(CeedQFunction qf, int* (*f)) {
```
jeremylt: I'm not super happy with how this works, but it works.

jedbrown: This `f` is not a pointer to a function, just a pointer to data. You could use `void (**f)(void)` if you want a pointer to a generic function pointer.

jeremylt: I will give that a go. I spent quite a while fighting this particular few lines.

jeremylt:
I think this is ready too. I am ~90% sure the MAGMA backend still works, but honestly it will need a separate PR to incorporate all of the getters/setter and streamlining after we can get #111 merged.

PR #173, opened 2018-11-16
CeedVecs Write Access Block With Readers

jeremylt:
This PR blocks CeedVec write access if the vector currently is under read access.


jeremylt:
Note: I rebased this to include the Travis fix from 170.

jeremylt:
I added a check for writers to SetArray as well

jeremylt:
Code coverage goes down slightly because the added tests do not run to the end of the file.

PR #174, opened 2018-11-19
Vec inputs

jeremylt:
This is a ~~WIP~~ PR to switch CeedBasisApply and CeedQFunctionApply to vector inputs to facilitate GPU backends. 

Status:
- Interface converted
- Tests converted
- Ref converted
- Tmpl converted
- Blocked converted
- OCCA converted
- MAGMA converted

Note: This PR is based off of PR #172

Issue #164, opened 2018-10-05
Quid of the libCEED API for GPU


YohannDudouit:
Studying the implementation of Generic GPU kernels for MFEM, I wanted to reiterate my concerns for efficient GPU implementation in libCEED. Take what I say with a pinch of salt though as I might have overlooked many aspects of the problem.

If we take as a primary goal for efficient GPU backend: generating GPU code with performance close to [libParanumal](https://www.paranumal.com/). If I understand correctly libParanumal, the GPU kernels fuse `Prolongation`/`OperatorApply`/`Restriction`, so that the kernel only requires one memory swipe. Since the kernels are highly memory bound, reading and writing the data only once (all computation being hidden by memory transfers) gets close to the best achievable performance for partial assembly with sum factorization.
The libCEED approach is to decompose a kernel into `Prolongation`, `OperatorApply`, `Restriction`, and further decompose `OperatorApply` in `BasisApply`->`QFunctionApply`->`BasisApply`.
This decomposition provides a clear path on CPU for generic operators, we decided in libCEED to implement this genericity through dynamic linking of function pointers according to the intent of the user. This approach is reasonable on CPU since at this level of abstraction the function pointers and optimizations barriers created by them are not an issue.
However, for a GPU, a dynamic function pointer means a GPU kernel launch, and thus a memory swipe. A naive approach would thus result in five memory swipes (`Prolongation`->`BasisApply`->`QFunctionApply`->`BasisApply`->`Restriction`). Since these kernels are highly memory bound, the performance is directly proportional to the inverse of the number of memory swipes. This naive approach would thus result in the **best case scenario** in approximately 20% of the performance of libParanumal. On the other hand, an optimized approach would require to be at the highest level of abstraction of libCEED in order to perform only one memory swipe.
The problem when we implement kernels at the highest level of abstraction is that we loose all the generic decomposition of an operator that libCEED is trying to provide. GPU kernels become nothing more than a single function name that calls a kernel that has to be specialized for every operator desired by the users. This specialization of GPU kernels at the highest level of abstraction would result in a combinatorial explosion of the required GPU kernels.
Since this issue is in conflict with libCEED's primary goal of providing optimized generic operator specified by the user, this means that in order to get good generic GPU performance we need to be able to do code generation. The problem with this is that it seems to split the libCEED CPU and GPU API, and it is not trivial to see how to reunify them.

[...]

Issue #162, opened 2018-09-27
Add CeedVectorSyncArray


jeremylt:
We need to add a new frontend function `CeedVectorSyncArray` to signal to GPU backends that CeedVector data needs to be pushed to the host.

YohannDudouit:
I understand the need for such a functionality, but I think we should be careful on how we implement this, a convenient implementation might reveal to be especially bug prone. Do you have in mind a proposition on what form this synchronisation would have in term of code?

jeremylt:
Implementation would depend upon the specific backend. I am imagining this being just a way for the user to demand that the host be updated with data from the device, for whatever reason. The backend gets to decide how that works.

YohannDudouit:
I mean, did you have an interface in mind? The kind of bugs I was thinking of already arise at the interface level.

jeremylt:
I'm not sure what you mean. All I imagine adding to the frontend is `CeedVectorSync(CeedVector v)`. The backend then would make sure that the data on the host matches the data on the device via whatever means is most appropriate. I know nothing about the best way to copy data from the device to the host.

My hope is that this would let GPU backends only update the host pointer when requested.

Here is some sample user code I'm thinking of in the context of the PETSc example:

```
  ierr = VecGetArrayRead(user->Xloc, (const PetscScalar**)&x); CHKERRQ(ierr);
  ierr = VecGetArray(user->Yloc, &y); CHKERRQ(ierr);
  CeedVectorSetArray(user->xceed, CEED_MEM_HOST, CEED_USE_POINTER, x);
  CeedVectorSetArray(user->yceed, CEED_MEM_HOST, CEED_USE_POINTER, y);

  CeedOperatorApply(user->op, user->xceed, user->yceed,
                    CEED_REQUEST_IMMEDIATE);
  CeedVectorSync(user->yceed);
```

And for the Nek example:

```
      call ceedvectorsetarray(vec_ap1,ceed_mem_host,ceed_use_pointer,
     $  ap1,err)
       call ceedoperatorapply(op_mass,vec_p1,vec_ap1,	
     $  ceed_request_immediate,err)
       call ceedvectorsync(vec_ap1,err)
```

I am hoping this new function clarifies when communication between host and device is required, removing extra communication.

YohannDudouit:
[...]

Issue #86, opened 2018-06-29
Making QFunction Context Const


Steven-Roberts:
The context for a qfunction callback is documented as an in/out parameter. https://github.com/CEED/libCEED/blob/535acf402d3673fa5aa2d037f0e06810d1903c19/ceed-qfunction.c#L46-L47
For the cuda and occa backends, the goal is to have the qfunction being applied in parallel.  If the context is modified in a qfunction, however, this could lead to race conditions and other potential issues.  Is there any reason that the context cannot be declared constant?

jeremylt:
On line 128 of the PETSc bp1.c example, there is a function that is writing to ctx.

Steven-Roberts:
Ahh I had not noticed that.  I suppose the documentation will need to explain that atomics or other synchronization constructs are needed in a situation like this.

jedbrown:
My theory was that if the context was to be used for monitoring, it would only be modified using atomics.  After merging active/passive, we could make a type of passive output that would be reduced with a user-specified operation such as sum or max.

jeremylt:
Is there planned work here?

jeremylt:
See Issue #183

PR #175, opened 2018-11-28
Cuda backend

YohannDudouit:
This PR update the Cuda backend to make it usable as soon as their is a path for QFunction working with Cuda.
- [x] ~~This PR should be merged after #174~~


jeremylt:
FYSA in Travis, the most recent version of OCCA errored for the restrictions on this commit.

backends/cuda/ceed-cuda-basis.c
```diff
@@ -134,21 +134,31 @@ extern "C" __global__ void grad(const CeedInt nelem, const int transpose, const
...
-    pre /= BASIS_Q1D;
-    for (CeedInt i=0; i<pre; i++) {
-      for (CeedInt j=0; j<BASIS_Q1D; j++) {
-        for (CeedInt k=0; k<post; k++) {
-          v[(i*BASIS_Q1D + j)*post + k] = qweight1d[j] * (d == 0 ? 1 : v[(i*BASIS_Q1D + j)*post + k]);
-        }
-      }
-    }
-    post *= BASIS_Q1D;
+// extern "C" __global__ void weight(const CeedScalar * __restrict__ qweight1d, CeedScalar * __restrict__ v) {
```
jeremylt: Once we get ready to mege, it would be nice to clean out any old commented code. I know the OCCA backend has a lot of commented out code, but I think its preferable to delete lines that aren't going to be used. (I realize this is a WIP but I thought this was as good of a place to mention it as any)

YohannDudouit: I kept this one for legacy reasons, I implemented a quick version without thinking too much about performance.

jeremylt: I bring it up because what I've seen in the OCCA backend is that commented code does not get maintained and becomes dated and confusing quickly.

YohannDudouit:
This PR is ready to be reviewed.

jeremylt:
I don't have a CUDA device handy at the moment - did the issue with the Fortran tests and the vectors get sorted out?

Makefile
```diff
@@ -212,9 +214,18 @@ ifneq ($(wildcard $(OCCA_DIR)/lib/libocca.*),)
   BACKENDS += /cpu/occa /gpu/occa /omp/occa
 endif
 
+CUDA_LIB_DIR := $(wildcard $(foreach d,lib lib64,$(CUDA_DIR)/$d/libcudart.${SO_EXT}))
```
jeremylt: We'll want to modify the 'info' part of the makefile to include the CUDA backend

YohannDudouit:
The Fortran tests don't pass but we should discuss if this a problem of the libCEED interface, or of my backend. I can easily have it pass these tests, but I think that what I'm doing is correct and the libCEED interface for Fortran should be able to return the same pointer as the one I gave.

backends/cuda/ceed-cuda-basis.c
```diff
@@ -0,0 +1,278 @@
...
+
+  __shared__ CeedScalar s_mem[BASIS_Q1D * BASIS_P1D + 2 * BASIS_BUF_LEN];
+  CeedScalar *s_interp1d = s_mem;
+  CeedScalar *s_buf1 = s_mem + BASIS_Q1D * BASIS_P1D;
+  CeedScalar *s_buf2 = s_buf1 + BASIS_BUF_LEN;
+  for (CeedInt k = i; k < BASIS_Q1D * BASIS_P1D; k += blockDim.x) {
+    s_interp1d[k] = interp1d[k];
+  }
+
+  const CeedInt P = transpose ? BASIS_Q1D : BASIS_P1D;
```
jeremylt: Will this cover non-tensor bases as well or will that be a future modification?

backends/cuda/ceed-cuda-operator.c
```diff
@@ -0,0 +1,390 @@
...
+                                      lmode, vec, impl->evecs[i],
+                                      request); CeedChk(ierr);
+      // Get evec
+      ierr = CeedVectorGetArrayRead(impl->evecs[i], CEED_MEM_DEVICE,
+                                    (const CeedScalar **) &impl->edata[i]);
+      CeedChk(ierr);
+    }
+  }
+
+  // Output Evecs
```
jeremylt: You could just remove this?

backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -0,0 +1,203 @@
...
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+#include <ceed-impl.h>
+#include <string.h>
+#include <stdio.h>
+#include "ceed-cuda.h"
+
+// int CeedQFunctionApplyElems_Cuda(CeedQFunction qf, const CeedInt Q,
```
jeremylt: QFunctionApply is already agnostic to the number of elements, so I don't think we need to keep this around.

backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -0,0 +1,203 @@
...
+    CeedChk(ierr);
+  }
+
+  for (CeedInt i = 0; i < numoutputfields; i++) {
+    ierr = CeedVectorGetArray(V[i], CEED_MEM_DEVICE, &data->fields.outputs[i]);
+    CeedChk(ierr);
+  }
+
+  // TODO find a way to avoid this systematic memCpy
+  if (qf->ctxsize > 0) {
```
jeremylt: The ctx could be modified, so a copy back may be required.

backends/cuda/ceed-cuda-restriction.c
```diff
@@ -0,0 +1,252 @@
...
+
+  ierr = CeedElemRestrictionSetData(r, (void*)&impl); CeedChk(ierr);
+  ierr = CeedSetBackendFunction(ceed, "ElemRestriction", r, "Apply",
+                                CeedElemRestrictionApply_Cuda); CeedChk(ierr);
+  ierr = CeedSetBackendFunction(ceed, "ElemRestriction", r, "Destroy",
+                                CeedElemRestrictionDestroy_Cuda); CeedChk(ierr);
+  return 0;
+}
+
+int CeedElemRestrictionCreateBlocked_Cuda(const CeedMemType mtype,
```
jeremylt: This is just a question to improve my knowledge: We use blocked restrictions to help us interlace dofs and get faster BasisApply. Would this help CUDA/GPUs at all?

backends/cuda/ceed-cuda-vec.c
```diff
@@ -0,0 +1,352 @@
...
+                                   const CeedMemType mtype,
+                                   CeedScalar **array) {
+  int ierr;
+  Ceed ceed;
+  ierr = CeedVectorGetCeed(vec, &ceed); CeedChk(ierr);
+  CeedVector_Cuda *data;
+  ierr = CeedVectorGetData(vec, (void*)&data); CeedChk(ierr);
+
+  switch (mtype) {
+  case CEED_MEM_HOST:
```
jeremylt: @camierjs I was thinking of trying to add Yohann's dual memory model to the OCCA backend - I think it could help speed up the OCCA performance by removing extra copies from device to host to device. I think it will be more complex though in the case of OCCA as OCCA runs on CPU or GPU based upon the specific backend.

backends/cuda/ceed-cuda-vec.c
```diff
@@ -0,0 +1,352 @@
...
+      data->memState = BOTH_SYNC;
+    }
+    *array = data->d_array;
+    break;
+  }
+  return 0;
+}
+
+// *****************************************************************************
+static int CeedVectorGetArray_Cuda(const CeedVector vec,
```
jeremylt: Is this code different from GetArrayRead? I can't see a difference, but I could be wrong. If it's the same we can just use the same function for both.

backends/cuda/ceed-cuda-qfunction.c
YohannDudouit: My bad, I removed it but didn't save the file...

backends/cuda/ceed-cuda-qfunction.c
YohannDudouit: Yeah, that brings the question if we want a better way to deal with this, because memCpy is expensive, and we might want to have the ability to avoid these expensive copies.

backends/cuda/ceed-cuda-restriction.c
YohannDudouit: This help if you want to hide the Restriction/Prolongation on the CPU. Then you treat elements at E-vector level on GPU, and do the Restriction/Prolongation on CPU. If you do everything on GPU it's unclear to me that there is any benefit.

tests/t500-operator.cu
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
```
jeremylt: This looks so close to the CPU code for the QFunction and close to the OCCA version. I really want to start prototyping some approaches to getting down to one source file for the QFunctions - I know users will make bugs juggling 3 copies of the QFunction because I make bugs juggling 2 copies.

I am ok merging the use of a .cu file for the QFunctions as well only under the understanding that we will work to remove this soon and get down to 1 source file for QFunctions.

backends/cuda/ceed-cuda-vec.c
YohannDudouit: It could also be interesting for the MAGMA (@stomov ) backend, I added a `NONE_SYNC` state, not sure that it's really needed, but it made the synchronisation logic simpler.

backends/cuda/ceed-cuda-vec.c
YohannDudouit: It's not the same, the synchronisation logic is not the same.

YohannDudouit: `GetArrayRead` sync results in `BOTH_SYNC` whereas `GetArray` never results in `BOTH_SYNC` since the array can be modified.

backends/cuda/ceed-cuda-qfunction.c
jeremylt: This deals with #86. Currently ctx is largely constant but in the inteface QFuntions may modify it.

jeremylt:
Does this pass 'make prove-all'? I could see the PETSc BPs requiring #162 in order to work correctly.

backends/cuda/ceed-cuda-basis.c
YohannDudouit: I haven't touched at all the GPU implementations except what wasn't written and the `weight` kernel. I'm working on a separate branch on a similar branch using my GPU implementations instead of @Steven-Roberts 's.
I haven't spent the time to understand the logic of this implementation, and look if it could be extended to non-tensor basis. I imagine that it could be hacked using `BASIS_DIM=1` and `BASIS_Q1D=Q`, since applying the contraction we use is a basically a matrix vector product.

jeremylt: Fair. I think you are right about setting the backend BASIS_DIM=1 and BASIS_Q1D=Q. We do something similar to enable non-tensor bases for Ref and Blocked.

tests/t500-operator.cu
YohannDudouit: The case where it will be different, is for non-linear operators where we compute the geometric factors on the fly. Then using GPU specific code will be unavoidable I think, but the GPU code can easily degenerate to CPU code I think. I'm just a bit worried about memory coalescing, this might require code that is only specific to GPU to load data efficiently, as incorrect memory access easily result in a factor 30 slowdown.

YohannDudouit:
I haven't looked at the example part, I'll try my best, but this requires me to figure out how to use mfem, petsc,  etc on Ray, so this might take some time.

jeremylt:
Hopefully it should just be a question of writing the .cu files if our test suite is working as intended.

tests/t500-operator.cu
jeremylt: Geometry on the fly vs precomputed will require completely different QFunctions and Operators. I think it makes the case for single source files for QFunctions even stronger as it would be better for performance comparisons to swap out different single implimentations of particular operators rather than having to swap out triplicates of the operators and not knowing 100% that they are identical in their action.

tests/t500-operator.cu
```diff
@@ -0,0 +1,39 @@
...
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+// *****************************************************************************
+extern "C" __global__ void setup(void *ctx, CeedInt Q, Fields_Cuda fields) {
+  const CeedScalar *weight = fields.inputs[0], *dxdX = fields.inputs[1];
```
jeremylt: This is another step away from having a single approach to QFunctions that each backend adapts to their own needs. Currently we have three ways that users need to write their inputs to QFunctions (and in three different source files for the exact same operator/qfunction combo)

CPU: array of pointers to scalar array, as per the spec (will eventually be in separate header files)
OCCA: pair of scalar arrays with all inputs/outputs and an array of offsets
CUDA: structs of input/output arrays

I think OCCA and CUDA should be able to have a single approach that also works well for CPU and doesn't require much/any efforts on the part of the user. If there is a performance barrier to providing an array of device arrays, perhaps a common struct of arrays would allow us to unify these aproaches.

I want to make it as easy as possible for users to use libCEED with multiple backends on hetrogeneous architecture without having to do a lot of work maintaining multiple copies of idential logic for their problem.

@jedbrown, @camierjs thoughts on QFunctions inputs/outputs as arrays of pointers to arrays vs single arrays with an array of offsets vs a struct of arrays?

jeremylt:
I can't speak for other reviewers, but to me the main 'TODO's are:

NEED:
- Fortran tests passing
- MFEM and PETSc examples passing (might need #162)
- Verify ctx is coppied corectly
- Squash to #174 + 1 passing commit

NICE TO HAVE:
- Comments clean out
- Non-tensor bases

FUTURE:
- Single source QFunctions (asking the user to write 3 versions of their QFunctions will kill adoption and introduce bugs, IMHO)

backends/cuda/ceed-cuda-operator.c
```diff
@@ -27,7 +27,7 @@ static int CeedOperatorDestroy_Cuda(CeedOperator op) {
     ierr = CeedVectorDestroy(&impl->evecs[i]); CeedChk(ierr);
   }
   ierr = CeedFree(impl->evecs); CeedChk(ierr);
-  ierr = CeedFree(impl->edata); CeedChk(ierr);
+  // ierr = CeedFree(impl->edata); CeedChk(ierr);
```
jeremylt: I wonder why you're getting an error. We're going to want to free that so we don't leak.

YohannDudouit: Sorry, that was lazy of me. Always the same error, I give the pointer I want to deallocate instead of the address of the pointer. I don't know why it didn't show up before though.

YohannDudouit:
@camierjs  The fall-through are on purpose... Is it not legal C any more?
Is the backend responsible that the C tests pass, but not the Fortran? I feel like this is an interface issue, not a backend issue, no?
I still haven't managed to install mfem or petsc on Ray, so I cannot test the examples for the moment to diagnose why they don't work.

jeremylt:
A backend mus pass 'make prove-all' to be accepted. That includes the C tests, Fortran tests, CEED example, PETSc examples, and MFEM examples.

YohannDudouit:
@camierjs t400f t500f t501f t502f pass on Ray, what kind of error do you have?

YohannDudouit:
@camierjs CEED/ex1, mfem/bp1 and mfem/bp3 examples now pass.

YohannDudouit:
@camierjs @jedbrown @jeremylt All tests pass on my side, I had to modify slightly the petsc examples to synchronize the required output vectors.

backends/ref/ceed-ref-operator.c
```diff
@@ -167,6 +167,7 @@ static int CeedOperatorSetup_Ref(CeedOperator op) {
   CeedChk(ierr);
 
   // Outfields
+
```
jeremylt: Stray line added

examples/mfem/bp1.hpp
```diff
@@ -255,5 +255,11 @@ class CeedMassOperator : public mfem::Operator {
     CeedVectorSetArray(u, CEED_MEM_HOST, CEED_USE_POINTER, x.GetData());
     CeedVectorSetArray(v, CEED_MEM_HOST, CEED_USE_POINTER, y.GetData());
     CeedOperatorApply(oper, u, v, CEED_REQUEST_IMMEDIATE);
+
+    const CeedScalar* array;
```
jeremylt: Did you want to similarly comment this as to be updated with CeedVectorSyncArray?

include/ceed.h
```diff
@@ -289,9 +289,11 @@ CEED_EXTERN int CeedLobattoQuadrature(CeedInt Q, CeedScalar *qref1d,
 CEED_EXTERN int CeedQRFactorization(CeedScalar *mat, CeedScalar *tau, CeedInt m,
                                     CeedInt n);
 
+typedef int (*CeedQFunctionCallback)(void *, const CeedInt, const CeedScalar *const *, CeedScalar *const *);
```
jeremylt: Perhaps CeedUserQFunction instead of CeedQFunctionCallback? Names are hard.

examples/petsc/bp1.cu
```diff
@@ -0,0 +1,56 @@
...
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+// *****************************************************************************
+extern "C" __global__ void Setup(void *ctx, CeedInt Q,
+                                 Fields_Cuda fields) {
+  CeedScalar *rho = fields.outputs[0], *true_soln = fields.outputs[1], *rhs = fields.outputs[2];
```
jeremylt: I am ok with merging this third strategy for managing inputs/outputs to QFunction for now, but I think we are long overdue for getting down to one QFunction source. libCEED simply won't be usable if 3 copies of every QFunction have to be maintianed and diverging strategies will impeed the full QFunction Gallery.

I think single source needs to be our next objective after merging the new backends.

@jedbrown, thoughts on adding a third input/output strategy for QFunctions?

backends/ref/ceed-ref-operator.c
YohannDudouit: Removed.

examples/mfem/bp1.hpp
YohannDudouit: TODO comment added.

include/ceed.h
YohannDudouit: Name modification done.

Makefile
```diff
@@ -47,13 +47,13 @@ CUDA_DIR  ?= $(or $(patsubst %/,%,$(dir $(patsubst %/,%,$(dir \
 AFLAGS = -fsanitize=address #-fsanitize=undefined -fno-omit-frame-pointer
 
 OPT    = -O -g
-CFLAGS = -std=c99 $(OPT) -Wall -Wextra -Wno-unused-parameter -fPIC -MMD -MP
-NVCCFLAGS = $(OPT)
+CFLAGS = -std=c99 $(OPT) -Wall -Wextra -Wno-unused-parameter -fPIC -MMD -MP -lstdc++
```
jedbrown: `-lstdc++` is an `LDFLAGS` and presumably only exists if (a) C++ code is compiled in libCEED (depends on configuration) or (b) some other library needs it when using static libraries? Also, some platforms want to use `-lc++`. Probably best to just choose whether to use `LINK.c` or `LINK.cc` (these are implicitly defined by make; see `make -p -f/dev/null` for a complete list).

backends/cuda/cuda-vec.cu
```diff
@@ -0,0 +1,34 @@
...
+#include <cuda.h>
+
+__global__ void setValueK(CeedScalar * __restrict__ vec, CeedInt size, CeedScalar val) {
+  int idx = threadIdx.x + blockDim.x * blockIdx.x;
+  if (idx >= size)
+    return;
+  vec[idx] = val;
+}
+
+extern "C" int DeviceSetValue(CeedScalar* d_array, CeedInt length, CeedScalar val) {
```
jedbrown: Namespacing needed. Check `nm -D lib/libceed.so | grep ' [BDT] ' | grep -v -i ' ceed'` after building to see namespace violations.

backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -0,0 +1,179 @@
...
+    return CeedError(ceed, 1, "Cannot find file's extension!");
+  const size_t cuda_path_len = last_dot - cuda_file;
+  strcpy(&cuda_file[cuda_path_len], ".cu");
+  //*******************
+  FILE *fp;
+  long lSize;
+  char *buffer;
+
+  fp = fopen ( cuda_file, "rb" );
+  if( !fp ) perror(cuda_file),exit(1);
```
jedbrown: Library code should never call `exit()`. Use `CeedError()`.

backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -0,0 +1,179 @@
...
+  fseek( fp, 0L, SEEK_END);
+  lSize = ftell( fp );
+  rewind( fp );
+
+  /* allocate memory for entire content */
+  ierr = CeedCalloc( lSize+1, &buffer ); CeedChk(ierr);
+
+  /* copy the file into the buffer */
+  if( 1!=fread( buffer, lSize, 1, fp) )
+    fclose(fp),free(buffer),fputs("entire read fails",stderr),exit(1);
```
jedbrown: Also here. And just use braces instead of comma chains. Also note that `CeedFree` must be used on `CeedMalloc`ed memory.

Makefile
YohannDudouit: As you know I'm really not comfortable with this kind of things, I had to use this flag to have successful compilation with the Cuda runtime API. I'm not attached to this flag, so if you have a solution that you think better, I have no problem using it.

backends/cuda/cuda-vec.cu
YohannDudouit: I'm sorry but I'm not following, there is no namespace in C, so what do you mean? Is it because this function is not static? I thought about this but I didn't know how to solve this, so if you know the solution I'd appreciate it.

backends/cuda/ceed-cuda-qfunction.c
YohannDudouit: Sorry, I didn't write this code, I'll do your modifications.

backends/cuda/ceed-cuda-qfunction.c
YohannDudouit: Thanks.

examples/mfem/bp1.hpp
```diff
@@ -255,5 +255,11 @@ class CeedMassOperator : public mfem::Operator {
...
+    CeedVectorGetArrayRead(u, CEED_MEM_HOST, &array);
```
jeremylt: 'u' shouldn't need to be synced as its an input

examples/mfem/bp3.hpp
```diff
@@ -315,5 +315,11 @@ class CeedDiffusionOperator : public mfem::Operator {
     CeedVectorSetArray(v, CEED_MEM_HOST, CEED_USE_POINTER, y.GetData());
 
     CeedOperatorApply(oper, u, v, CEED_REQUEST_IMMEDIATE);
+
+    const CeedScalar* array;
+    CeedVectorGetArrayRead(u, CEED_MEM_HOST, &array);
```
jeremylt: Same here, 'u' shouldn't need to be synced

examples/petsc/bp3.cu
```diff
@@ -88,5 +88,4 @@ extern "C" __global__ void Error(void *ctx, CeedInt Q,
        i += blockDim.x * gridDim.x) {
     err[i] = u[i] - target[i];
   }
-  return 0;
```
jeremylt: Eventually the goal is to have all QFunctions return an error code (OCCA, CUDA, and C) as we want the user to be able to add diagnostics to their QFunctions to indicate if something goes awry.

YohannDudouit: __global__ functions have to be void.

examples/mfem/bp1.hpp
YohannDudouit: Thanks, I thought about this but forgot to modify it.

examples/petsc/bp3.cu
jeremylt: I remember some discussion at CEED2AM about how to get it a return value to work for OCCA, but I don't recall all of it. I just recall the desire for the user to set return values.

Makefile
YohannDudouit: Actually, it seems that I needed this at some point but not any more...

backends/cuda/cuda-vec.cu
YohannDudouit: I added `Ceed` in front of `DeviceSetValue`, is it sufficient?

Makefile
jedbrown: If I remove the `-lstdc++`, I see
```
$ lddtree build/t101-vec
build/t101-vec (interpreter => /lib64/ld-linux-x86-64.so.2)
    libceed.so => /home/jed/src/libceed/lib/libceed.so
        libocca.so => /home/jed/src/occa/lib/libocca.so
            libpthread.so.0 => /usr/lib/libpthread.so.0
            librt.so.1 => /usr/lib/librt.so.1
            libdl.so.2 => /usr/lib/libdl.so.2
            libOpenCL.so.1 => /usr/lib/libOpenCL.so.1
            libstdc++.so.6 => /usr/lib/libstdc++.so.6
            libgomp.so.1 => /usr/lib/libgomp.so.1
        libcudart.so.10.0 => /opt/cuda/lib64/libcudart.so.10.0
        libnvrtc.so.10.0 => /opt/cuda/lib64/libnvrtc.so.10.0
        libcuda.so.1 => /usr/lib/libcuda.so.1
            libnvidia-fatbinaryloader.so.415.25 => /usr/lib/libnvidia-fatbinaryloader.so.415.25
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1
    libm.so.6 => /usr/lib/libm.so.6
    libc.so.6 => /usr/lib/libc.so.6
```

which suggests that you are needing it only when OCCA uses static libraries. In that case, we could add `LDLIBS += -lstdc++` when OCCA is enabled. If you see this also with CUDA, then add if _either_ OCCA or CUDA is being used. Some compilers will produce noisy warnings if you use linker flags when only compiling. For example, your version produces
```
clang-7: warning: -Z-reserved-lib-stdc++: 'linker' input unused [-Wunused-command-line-argument]
```
for every compilation unit when I `make CC=clang`.

backends/cuda/cuda-vec.cu
jedbrown: C-style namespaces just prefix the library name. You could call this `CeedCudaDeviceSetValue()`, for example. The `nm` command above will give a list of non-namespaced public symbols.

jedbrown: Normally this could be `CEED_INTERN`, but it appears that nvcc does not understand visibility attributes.

YohannDudouit:
@jedbrown The problem with making the Cuda QFunction a `__device__` function is its signature, because we even less want to have pointers of pointers at the quadrature point level, that would be a disaster. The problem is that to be inlined efficiently, we cannot use 'CeedScalar* inputs[16], CeedScalar* ouputs[16]', we need a more direct access to the data, this is why I used a struct to carry the pointers, the pointers are copied by value (in the kernel argument cache) not accessed through an indirection that needs to be repeated by each thread.
However, maybe that a `__device__` function could use this struct plus the current qIndex and Q, so that we could fuse the `__device__` function contrary to the current state with only a `__global__` function that can't be fused with BasisApply. I already discussed about this with @jeremylt  and I plan to do this when I'll be working on #164 , but it wouldn't get rid of the Fields_Cuda struct (We can rename it though).

YohannDudouit:
@jedbrown Do you have any objection to squashing and merging this as is?

interface/ceed-fortran.c
```diff
@@ -479,7 +479,7 @@ void fCeedQFunctionCreateInterior(int* ceed, int* vlength,
 
   CeedQFunction *qf_ = &CeedQFunction_dict[CeedQFunction_count];
   *err = CeedQFunctionCreateInterior(Ceed_dict[*ceed], *vlength,
-                                     CeedQFunctionFortranStub,focca_c, qf_);
+                                     CeedQFunctionFortranStub, focca, qf_);
```
jedbrown: Why? The string that comes directly from Fortran might not be NULL terminated.

YohannDudouit: Wow, I have no idea how this happened, sorry for that.

include/ceed-impl.h
```diff
@@ -129,12 +129,13 @@ struct CeedQFunction_private {
...
   CeedQFunctionField *outputfields;
   CeedInt numinputfields, numoutputfields;
-  int (*function)(void*, CeedInt, const CeedScalar *const*, CeedScalar *const*);
+  CeedUserQFunction function;
   const char *focca;
   bool fortranstatus;
   void *ctx;      /* user context for function */
   size_t ctxsize; /* size of user context; may be used to copy to a device */
   void *data;     /* backend data */
+  char* spec;     /* the string spec of the qFunction */
```
jedbrown: This is unused in the present PR. Can we keep it out for now?

include/ceed.h
```diff
@@ -289,9 +289,12 @@ CEED_EXTERN int CeedLobattoQuadrature(CeedInt Q, CeedScalar *qref1d,
 CEED_EXTERN int CeedQRFactorization(CeedScalar *mat, CeedScalar *tau, CeedInt m,
                                     CeedInt n);
 
+typedef int (*CeedUserQFunction)(void *, const CeedInt,
```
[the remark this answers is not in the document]
YohannDudouit: Should I remove this `typedef` and use the type directly in `CeedQFunctionCreateInterior`?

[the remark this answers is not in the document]
jedbrown: I'm not opposed to the typedef, but it needs to be documented and named in a consistent way.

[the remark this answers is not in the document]
jedbrown: See 80d7e79dff68893a0ca2b9fda474205682da13d1 and merge it if you think it's complete.

PR #177, opened 2018-12-03
Adds a Vector unit test using CEED_MEM_DEVICE.

YohannDudouit:
Proposition of a first unit test for CeedVector using CEED_MEM_DEVICE.

### DIFF
--- a/examples/nek5000/.gitignore
+++ b/examples/nek5000/.gitignore
@@ -2,7 +2,8 @@ ex1
 ex1.f
 ex1.log*
 *.out
-*.build.log
+*build.log
+*.state
 makefile
 makenek*
 obj/
--- a/examples/nek5000/bp1.usr
+++ b/examples/nek5000/bp1.usr
@@ -714,6 +714,7 @@ C     Solution to BP1 using libCEED
       integer qf_mass,qf_setup
       integer op_mass,op_setup
       real*8  x,y,z
+      integer*8 offset
 
       external massf,masssetupf
 
@@ -784,8 +785,9 @@ C     Create ceed vectors
       call ceedvectorcreate(ceed,ldim*lx*nelt,vec_coords,err)
       call ceedvectorcreate(ceed,nelt*q**ldim,vec_qdata,err)
 
+      offset=0
       call ceedvectorsetarray(vec_coords,ceed_mem_host,
-     $  ceed_use_pointer,coords,err)
+     $  ceed_use_pointer,coords,offset,err)
 
 C     Create ceed qfunctions for masssetupf and massf
       call ceedqfunctioncreateinterior(ceed,1,masssetupf,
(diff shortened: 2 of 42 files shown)


jeremylt:
I rebased this branch to the new master to see what would happend to this new test in Travis.

jeremylt:
It fails, as expected, which is why we should allow the error message 'Can only provide to HOST memory' in tap for this test - Travis won't be able to run with GPUs so this test will always error on Travis.

jeremylt:
As I see it, this PR needs the following changes to be accepted:

- Modify tests/tap.sh to 'SKIP' this test when the error message is 'Can only provide to HOST memory'. (see lines 44-51 of tap.sh for an example)
- Check this yields intended behavior for OCCA (may have to revisit after Device memory added to OCCA backend)
- Rename test to t109 to be compatible with current master

jeremylt:
I put in a Fortran test, but I haven't verified that it works as desired for CUDA.

YohannDudouit:
The Fortran test doesn't work, I suspect a pointer issue, I'll try to investigate more as soon as possible.

jeremylt:
We'll need to add exceptions to the new junit test capability in Jenkins (added for our future GPU CI capability)

jeremylt:
Currently the Fortran test fails for CUDA. @YohannDudouit is investigating. I suspect a problem in the Fortran interface, personally, as I think the 'GetArray' might inherently be CPU only as it currently is for the Fortran interface.

YohannDudouit:
The bug is in `t109-vec-f` line 30. `ceedvectorsetarray` does not have the correct argument, the problem is that I have no idea what the correct arguments are since I don't understand how the Fortran interface works. It seems that `boff` should be use, but the interface for `ceedvectorsetarray` does not take an `offset`, should it be added to the interface of this function?

jeremylt:
@YohannDudouit, I tried your idea. How does this look?

jeremylt:
@jedbrown, this should be pretty close to good now. There was an issue in SetArray.

Short-
The previous Fortran SetArray was:
```call ceedvectorsetarray(x,ceed_mem_host,ceed_use_pointer,xarray,err)```
The proposed new Fortran SetArray is:
```call ceedvectorsetarray(x,ceed_mem_host,ceed_use_pointer,xarray,xoffset,err)```

I added `xoffset` because Fortran may not have a direct pointer to the array, especially with a vector set by the array of another vector or device memory.

We could also add a function to our Fortran interface to avoid the offsets altogether,
Code proposed by @YohannDudouit:
```c
external caddress
real*8 val
int*8 add
call caddress(val, add)
```
```fortran
void caddress(void* val, long int* add){
  *add = val;
}
```

YohannDudouit:
All tests pass like this on Ray.

valeriabarra:
Looks good to me.

valeriabarra:
Looks good to me.

Issue #37, opened 2018-02-23
Create gallery of QFunctions (mass, Laplace, etc.)


jedbrown:
Public interface something like

```c
int CeedQFunctionCreateFromGallery(Ceed ceed, const char *spec, CeedQFunction *qf);
```

where `spec` is something like `"Mass"` or `"Mass:dim=3,rho=10"`. (Many possible syntax choices for options.)

jedbrown:
Like all libCEED objects, each `QFunction` instance is specific to the given `Ceed`. The Gallery code should execute for any backend. If a backend provides a specialized implementation, that will be bundled with the backend, not with the Gallery.

A `CeedOperator` is not complete until its fields have been specified. I'd prefer for the Gallery to return a "complete" `QFunction` instead of a partly created `Operator`, but I suppose a convenience function could be created so the user would need to see one fewer objects.

jeremylt:
In my mind, a single call of `CeedQfunctionCreateFromGallery(ceed, "Mass")` would replace lines 349-353 of the PETSc `bp1.c` file, and the QFunction would not need to be given as it is in the `bp1.h` file.

We couldn't do this completely at the CeedOperator level because we don't know the user basis, restriction, vectors, etc in advance, so the user would still need to set that information.

YohannDudouit:
I am currently working on integrating libParanumal in libCEED, but there is still some obscure corners in the way it will interact with the `QFunctionGallery`.
From libParanumal side I get the operator (`CeedOperator` level) doing something like:
```cpp
mesh->device.buildKernel(fileName, kernelName, kernelInfo);
```
But I cannot make this function call at the moment where I call `CeedQFunctionCreateFromGallery(ceed, opName, qf)` since `mesh` and `kernelInfo` need information at `CeedOperator` level to create the kernel.
So, does that mean that I should put all the information in the `void* data` of the `CeedQFunction_private` when calling `CeedQFunctionCreateFromGallery(ceed, opName, qf)` (`qf->data` would contain just the `opName`)? And then I extract the information (`opName`) contained in `data` and use it when calling `CeedOperatorCreate(ceed, qf, NULL, NULL, op)`, and from the information of `ceed` and `opName` I create the libParanumal `mesh` and `kernelInfo` and then call `mesh->device.builKernel(...)` to create the libParanumal operator?
Is it what you had in mind @jedbrown and @jeremylt ?

jeremylt:
[...]

PR #178, opened 2018-12-06
[WIP] [Discussion] QFunction Gallery

jeremylt:
I am creating this PR as a starting place for us to discuss how we want the gallery to work. As I see it, the gallery needs to do four things

- Streamline setup for common QFunctions
- Facilitate using highly optimized versions of known QFunctions
- Document QFunction requirements: inputs, outputs, data layout, etc
- Provide examples for users to make their own complex QFunctions

I think that we should be careful - this gallery should be based on the QFunction itself, not the bases or geometry, as we get that information from the CeedBasis. We shouldn't have a bunch of different QFunctions for each problem (bp1massp6q8het, bp1massp7q10tet, libparanumalbp1massp5q6tet, etc)

I think we should keep in mind our end goal of only having to write the source of the QFunctions once (rather than a C/Fortran, OCCA, CUDA, one-for-every-backend copy of the code).

Issue #37

### DIFF
--- a/gallery/bp1mass.h
+++ b/gallery/bp1mass.h
@@ -1,0 +1,39 @@
+// Copyright (c) 2017, Lawrence Livermore National Security, LLC. Produced at
+// the Lawrence Livermore National Laboratory. LLNL-CODE-734707. All Rights
+// reserved. See files LICENSE and NOTICE for details.
+//
+// This file is part of CEED, a collection of benchmarks, miniapps, software
+// libraries and APIs for efficient high-order finite element and spectral
+// element discretizations for exascale applications. For more information and
+// source code availability see http://github.com/ceed.
+//
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+/// @file
+/// libCEED QFunctions for setting up BP1 mass operator
+
+// *****************************************************************************
+static int Fields(Ceed ceed, CeedQFunction *qf) {
+
+  // Create the Q-function that defines the action of the mass operator.
+  CeedQFunctionCreateInterior(ceed, 1, Mass, __FILE__ ":Mass", &qf);
+  CeedQFunctionAddInput(qf, "u", 1, CEED_EVAL_INTERP);
+  CeedQFunctionAddInput(qf, "rho", 1, CEED_EVAL_NONE);
+  CeedQFunctionAddOutput(qf, "v", 1, CEED_EVAL_INTERP);
+}
+// *****************************************************************************
+static int Mass(void *ctx, CeedInt Q,
+                const CeedScalar *const *in, CeedScalar *const *out) {
+  const CeedScalar *u = in[0], *rho = in[1];
+  CeedScalar *v = out[0];
+  for (CeedInt i=0; i<Q; i++) {
+    v[i] = rho[i] * u[i];
+  }
+  return 0;
+}
+// *****************************************************************************
--- a/interface/ceed-qfunction.c
+++ b/interface/ceed-qfunction.c
@@ -85,6 +85,32 @@ int CeedQFunctionCreateInterior(Ceed ceed, CeedInt vlength,
 }
 
 /**
+  @brief Create a CeedQFunction from a named CeedQFunction in the gallery
+
+  @param ceed       A Ceed object where the CeedQFunction will be created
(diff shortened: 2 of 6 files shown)


gallery/bp1mass.h
```diff
@@ -0,0 +1,39 @@
...
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+/// @file
+/// libCEED QFunctions for setting up BP1 mass operator
+
+// *****************************************************************************
+static int Fields(Ceed ceed, CeedQFunction *qf) {
```
jeremylt: A created a folder with sample files of the info we need - I'm not married to this layout but it's a start. This function is the setup lines the user would have run if they made the qfunction outside of the gallery.

gallery/bp1mass.h
```diff
@@ -0,0 +1,39 @@
...
+
+  // Create the Q-function that defines the action of the mass operator.
+  CeedQFunctionCreateInterior(ceed, 1, Mass, __FILE__ ":Mass", &qf);
+  CeedQFunctionAddInput(qf, "u", 1, CEED_EVAL_INTERP);
+  CeedQFunctionAddInput(qf, "rho", 1, CEED_EVAL_NONE);
+  CeedQFunctionAddOutput(qf, "v", 1, CEED_EVAL_INTERP);
+}
+// *****************************************************************************
+static int Mass(void *ctx, CeedInt Q,
```
jeremylt: and this is the qfunction itself. Ideally I'd like to use this as a base and other backends (OCCA, CUDA, etc) can use this to build anything specific to that backend (like the OKL file).

interface/ceed-qfunction.c
```diff
@@ -81,6 +81,32 @@ int CeedQFunctionCreateInterior(Ceed ceed, CeedInt vlength,
...
+  @param ceed       A Ceed object where the CeedQFunction will be created
+  @param name       Name of the QFuction in the gallery
+  @param[out] qf    Address of the variable where the newly created
+                     CeedQFunction will be stored
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Basic
+**/
+int CeedQFunctionCreateFromGallery(Ceed ceed, const char *name,
```
jeremylt: Intentionally left just as comments for now, I see this function as needing to 
1) find the requested qfunction (folder search or registered functions search)
2) set up the requested qfunction (see above about the fields, etc)

jeremylt:
A topic for discussion: Do we want a OperatorCreateFromGallery?

Personally, I don't think such a function would help us. Unless we have an operator gallery for every set of bases and restrictions we will still need to make calls to OperatorSetField, so swapping OperatorCreate for OperatorCreateFromGallery doesn't really do anything for us, as far as I can tell.

jedbrown:
@jeremylt As is, `CeedOperatorCreate` just takes a `QFunction`. There could be a helper function that is implemented as
```
CeedQFunctionCreateFromGallery(..., &qf);
CeedOperatorCreate(ceed, qf, &op); // references qf
CeedQFunctionDestroy(&qf); // drop my reference
```
It won't particularly hurt (just makes an interface that may need to evolve if/when we support multiple QFunctions per Operator), but also isn't at all important so I wouldn't bother spending time on it now.

jeremylt:
Fair enough, though I think we would need to keep the refrence qf until the Operator is destroyed - it contains a lot of field data that is not stored in the Operator fields.

jedbrown:
The Operator holds a reference to the QFunction and will drop it properly when it's destroyed. The user doesn't need to know about it. Impl as written above.

jeremylt:
Good point

gallery/bp1mass.h
YohannDudouit: Ok this will work as long as the data accessed are of stride `Q`, if the layout is organized differently then something much more advanced has to be done on GPU.

jeremylt: Everything should be loops in term of Q

YohannDudouit: However, this prevents us from fusing operations on a GPU since this function is __global__, unfortunately making it a __device__ function doesn't resolve completely the problem, but might be closer to what we want. However, making the QFunction a __device__ function change the level at which the QFunction is described, indeed we cannot include the for loop on threads:
```
for (int i = blockIdx.x * blockDim.x + threadIdx.x;
    i < Q;
    i += blockDim.x * gridDim.x)
```
since the parallelization strategy is decided at a higher level (the level of the fused __global__ function), and cannot be at the __device__ function level.

jeremylt: Ideally the user should have to only specify the names of the input/output variables (CeedScalar *rho = in[0], etc) and the action at the quadrature points (v[i] = rho[i] * u[i]). If a more complex loop control structure is required, we can work to adapt the code generated to best match the needs of OCCA or CUDA.

PR #179, opened 2018-12-07
LibParanumal backend

YohannDudouit:
A proposition for integrating libParanumal kernels in libCEED.
Contains:
- libParnumal backend
- a dummy version of the QFunction gallery.
- t900 test that uses the Gallery


backends/libParanumal/ceed-libparanumal.c
```diff
@@ -28,12 +28,12 @@ static int CeedInit_libparanumal(const char *resource, Ceed ceed) {
 
   ierr = CeedSetBackendFunction(ceed, "Ceed", ceed, "OperatorCreate",
                                 CeedOperatorCreate_libparanumal); CeedChk(ierr);
-  ierr = CeedSetBackendFunction(ceed, "Ceed", ceed, "OperatorDestroy",
-                                CeedOperatorDestroy_libparanumal); CeedChk(ierr);
+  // ierr = CeedSetBackendFunction(ceed, "Ceed", ceed, "OperatorDestroy",
```
jeremylt: What trouble are you having with the interface - I would like to know for my PR that created the setter.

jeremylt: OperatorDestroy and QFunctionDestroy are associated with the Operator and the QFunction, not the Ceed

tzanio:
I have trouble commenting on this PR -- I get the following error
```
Start commit oid is not part of the pull request
```
Does anybody else have the same issue?

jeremylt:
@tzanio I am getting the same problem on a few files

jeremylt:
Wrt the gallery:
We will want to do this carefully - we don't want to modify this function into a long if-else chain in the long term. I started the discussion in PR #178 on how we can do this in a maintainable way that is backend agnostic

jeremylt:
Overall: I think it's a very solid start

- This is after PR #174, so some of the changes actually come from that PR
- I think we will want to condense this down to a handful of clean commits
- I am worried about the extensibility of the current OperatorApply code for libParanumal backend
- QFunction apply is currently empty but returns a sucess value after doing nothing; instead this function should return 'does not impliment' if you do not want this backend to do QFunctionApply outside of an operator or should delegate to OCCA if you want to still maintain that capability

include/ceed.h
```diff
@@ -292,20 +292,27 @@ CEED_EXTERN int CeedQRFactorization(CeedScalar *mat, CeedScalar *tau, CeedInt m,
...
-                                   const CeedScalar *const *u,
-                                   CeedScalar *const *v);
+                                   CeedVector* u, CeedVector* v);
 CEED_EXTERN int CeedQFunctionDestroy(CeedQFunction *qf);
 
 CEED_EXTERN int CeedOperatorCreate(Ceed ceed, CeedQFunction qf,
                                    CeedQFunction dqf, CeedQFunction dqfT,
                                    CeedOperator *op);
+
+CEED_EXTERN int CeedOperatorCreateFromGallery(Ceed ceed, char *spec,
```
jeremylt: We can't do an OperatorCreateFromGallery..... We would need to do one for every basis/restriction/ect combo...

interface/ceed-qfunction.c
```diff
@@ -77,10 +78,73 @@ int CeedQFunctionCreateInterior(Ceed ceed, CeedInt vlength,
...
+  @param ceed       A Ceed object where the CeedQFunction will be created
+  @param vlength    Vector length.  Caller must ensure that number of quadrature
+                    points is a multiple of vlength.
+  @param spec       The specification of the desired qFunction
+  @param[out] qf    Address of the variable where the newly created
+                     CeedQFunction will be stored
+
+  @return An error code: 0 - success, otherwise - failure
+**/
+int CeedQFunctionCreateInteriorFromGallery(Ceed ceed, CeedInt vlength,
```
jeremylt: We will want to do this carefully - we don't want to modify this function into a long if-else chain in the long term. I started the discussion in PR #178 on how we can do this in a maintainable way that is backend agnostic

tests/t900-libparanumal.c
```diff
@@ -0,0 +1,38 @@
+/// @file
+/// Test creation and destruction of a CEED object
+/// \test Test creation and destruction of a CEED object
+#include <ceed.h>
+
+int main(int argc, char **argv) {
+  Ceed ceed;
```
jeremylt: This should eventually become a t520 series test - if this is a basic elliptic PDE on hexes, then all backends should be able to impliment it from the gallery

I am confused as to where the data for the vectors is coming from - they are not ever created!

Also, the restrictions should be destroyed at the end of the test.

backends/libParanumal/ceed-libparanumal-operator.c
```diff
@@ -0,0 +1,141 @@
...
+  ierr = CeedOperatorGetData(op, (void*)&impl); CeedChk(ierr);
+
+  occaFree(impl->kernelInfo);
+  occaFree(impl->kernel);
+
+  ierr = CeedFree(&op->data); CeedChk(ierr);
+  return 0;
+}
+
+static int CeedOperatorSetup_libparanumal(CeedOperator op) {
```
jeremylt: I am worried about the extensibility of the current OperatorApply code for libParanumal backend

backends/libParanumal/ceed-libparanumal-qfunction.c
```diff
@@ -0,0 +1,26 @@
...
+// element discretizations for exascale applications. For more information and
+// source code availability see http://github.com/ceed.
+//
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+#include "ceed-libparanumal.h"
```
jeremylt: This backend does not impliment QFunctionApply. If this is the case, then it should have a dummy implimentation that returns an error message to that effect:

```
// * CeedBasisCreateH1_Occa
int CeedBasisCreateH1_Occa(CeedElemTopology topo, CeedInt dim,
                          CeedInt ndof, CeedInt nqpts,
                          const CeedScalar *interp,
                          const CeedScalar *grad,
                          const CeedScalar *qref,
                          const CeedScalar *qweight,
                          CeedBasis basis) {
  return CeedError(basis->ceed, 1, "Backend does not implement non-tensor bases");
}
```

include/ceed.h
YohannDudouit: There was a strong wishes for the Gallery being at Operator level, I tested but this doesn't need to be kept.

interface/ceed-qfunction.c
YohannDudouit: Since the discussion on the Gallery was idle I had to implement something, I went for the most straightforward as this is not the purpose of this PR.

tests/t900-libparanumal.c
YohannDudouit: This test is not full, those fields will have to be initialized indeed.

interface/ceed-qfunction.c
jeremylt: Fair enough. The interface is almost the same one I proposed, so it should be easy to swap in our agreed upon version once we make progress on that front.

backends/libParanumal/ceed-libparanumal-qfunction.c
jeremylt: I think in terms of usability, it would be ideal to use the default OCCA implimentation unless there is a libParanumal implimentation so that the user has something (with OperatorApply too?). Perhaps some backends should have READMEs for information like that?

include/ceed.h
YohannDudouit: We could imagine the `OperatorGallery` being able to call a subset of the `QFunctionGallery` and wrapping the `QFunction` in an `Operator`.

jeremylt: Agreed. I think it is preferable to make a fully formed QFunction rather than a half formed Operator as we will still have to set the Operator fields either way.

Makefile
```diff
@@ -272,7 +280,7 @@ $(tests) $(examples) : $(libceed)
 $(tests) $(examples) : LDFLAGS += -Wl,-rpath,$(abspath $(LIBDIR)) -L$(LIBDIR)
 
 run-% : $(OBJDIR)/%
-	@tests/tap.sh $(<:build/%=%)
+	#@tests/tap.sh $(<:build/%=%)
```
tzanio: Was this comment intentional?

YohannDudouit: It comes from a PR I merged. @jeremylt I imagine this comment should not be merged?

jeremylt: Correct, that shouldn't be there. I think it came from the CUDA branch.

jeremylt:
I think there might be some updats to 174 that might need to make their way into here, t305-f shouldn't be failing on Travis.

It seems strange to me that t900 is passing for /cpu/ref

jeremylt:
I can't speak for other reviewers, but to me the main 'TODO's are:

NEED:
- Make sure all functions are implemented, delegated, or give an error
- Resolve t900, and if its staying whitelist it for all other backends
- Fortran tests passing
- MFEM and PETSc examples passing (might need #162)
- Squash to #174 + 1 passing commit

NICE TO HAVE:
- Comments clean out

FUTURE:
- Full development of gallery

jeremylt:
Will this still work after the changes to OCCA, or will it be better to restart this work with a proper QFunction gallery (which will be easier with the new format)

jeremylt:
This is far enough out of date that I'll close the PR until the changes to master and the OCCA revamp can be incorporated.

PR #185, opened 2018-12-19
Improve QFunction ctx for Fortran interface

jeremylt:
@YohannDudouit discovered an quirk of the Fortran interface that will create issues for the CUDA/OCCA backends.

Specifically, the Fortran interface uses the `ctx` field to hold the Fortran function pointer *and* any user context. This functionality was not fully fleshed out to let users set QFunction context and have it used in backends that are not using the Fortran user QFunction (OCCA, CUDA). The current implimentation interferes with OCCA and CUDA's ability to directly copy over and use the `ctx` data.

 This needs some work, but I would appreciate thoughts on my approach.


YohannDudouit:
In order to avoid a ton of expensive memCpy for GPU backends, I think it would be good to have four kinds of context, a static one (needs to be transferred only once), a input dynamic one (memCopied from host to device before every `QFunctionApply`), an output dynamic one (memCopied  from device to host after every `QFunctionApply`). And the fourth one for the Fortran interface.

jeremylt:
This fix bypasses the need for a special Fortran context, but it may become a tad complex if we move to three contexts. 

Long term, maybe just a simple flag to indicate that the context was updated would be better.

YohannDudouit:
And how do you read this flag? memCpy? The problem of memCpy is mostly the overhead associated with it.

backends/occa/ceed-occa-qfunction.c
```diff
@@ -134,7 +134,13 @@ static int CeedQFunctionApply_Occa(CeedQFunction qf, CeedInt Q,
   // ***************************************************************************
   void *ctx;
   if (cbytes>0) {
-    ierr = CeedQFunctionGetContext(qf, &ctx); CeedChk(ierr);
+    bool fortranstatus;
```
jeremylt: If the backend does not use the Fortran interface to access the Fortran version user QFunction, then the backend is responsible for making a special call to get the true context data. This would apply to OCCA and CUDA.

include/ceed-impl.h
```diff
@@ -131,11 +131,34 @@ struct CeedQFunction_private {
   CeedInt numinputfields, numoutputfields;
   int (*function)(void*, CeedInt, const CeedScalar *const*, CeedScalar *const*);
   const char *focca;
+  bool fortranstatus;
```
jeremylt: I'm not commited to this particular variable name. It works but I'm open to a better name.

tests/t401-qfunction-f.f
```diff
@@ -0,0 +1,126 @@
+c-----------------------------------------------------------------------
```
jeremylt: This is the first time we have tested that the Fortran ctx works. This does not currently come up in the Nek5k PR.

interface/ceed-fortran.c
```diff
@@ -438,53 +438,37 @@ static int CeedQFunction_count = 0;
...
-  void *innerctx;
-};
-
 static int CeedQFunctionFortranStub(void *ctx, int nq,
                                     const CeedScalar *const *u, CeedScalar *const *v) {
-  struct fContext *fctx = ctx;
+  fContext *fctx = ctx;
   int ierr;
 
-  CeedScalar ctx_=1.0;
```
jeremylt: Previously we were simply suppling the user QFunction with false data for the ctx.

include/ceed-impl.h
```diff
@@ -131,11 +131,34 @@ struct CeedQFunction_private {
...
   void *ctx;      /* user context for function */
   size_t ctxsize; /* size of user context; may be used to copy to a device */
   void *data;     /* backend data */
 };
 
+/// Struct to handle the context data to use the Fortran QFunction stub
+/// @ingroup CeedQFunction
+typedef struct {
+  CeedScalar *innerctx;
```
jeremylt: I set the Fortran context to only be arrays of scalars. I'm not entirely sure Fortran77 does user defined structs in the same way C90 does, and this should still offer the desired functionality, as ints and bools can be saved into scalars and then recovered.

jeremylt:
Side note: CodeCov got confused. This will raise coverage.

include/ceed-impl.h
YohannDudouit: I think it would be better to separate the ctx and fortran_ctx. There is no reason why the backend should have access to the fortran_ctx.

jeremylt:
I think overhauling the ctx overall (Issue #86) is a bigger and separate PR after this one. In this PR I just want to fix the fact that currently the Fortran does not have the ability to set and pass context to the QFunction.

include/ceed-impl.h
jeremylt: The backend only would have to consider the fortran_ctx if the backend choses to bypass the Fortran interface when applying the QFunction. I fixed the implementation in the Fortran interface.

Here is the issue: The user provided Fortran function is wrapped in a C function to match the libCEED spec. But QFunctions only have 4 arguments, ctx, number of quad pts, input, and output. In order to have the C wrapped Fortran function match this signature, we need a way to pass in the true Fortran pointer, and ctx is the best place to do that.

include/ceed-impl.h
jedbrown: F77 doesn't have structs. F90 does, but they repeatedly botched the interface specification such that it's only practical without vendor-specific extensions after F2008+TS29113 (or F2018).

jeremylt:
I would like to merge this fix as soon as we can because the current code is hard broke, but I agree that we should overhaul the ctx overall. I would like to see some way to specify dynamic vs static ctx as it has been a repeated request from the GPU backends to only copy ctx back and forth host vs device if absolutely needed.

PR #190, opened 2019-01-07
Simplify accessing of inner context

jeremylt:
This PR simplifies access to the 'true' user context for backends that do not call through the C/Fortran user interface when using user QFunctions written in Fortran (such as CUDA and OCCA).

### DIFF
--- a/interface/ceed-qfunction.c
+++ b/interface/ceed-qfunction.c
@@ -299,7 +299,7 @@ int CeedQFunctionGetFortranStatus(CeedQFunction qf, bool *fortranstatus) {
 }
 
 /**
-  @brief Get Fortran global context for a CeedQFunction
+  @brief Get true user context for a CeedQFunction
 
   @param qf              CeedQFunction
   @param[out] ctx        Variable to store context data values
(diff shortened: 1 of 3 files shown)


interface/ceed-qfunction.c
```diff
@@ -309,13 +309,14 @@ int CeedQFunctionGetFortranStatus(CeedQFunction qf, bool *fortranstatus) {
...
 **/
 
-int CeedQFunctionGetFortranContext(CeedQFunction qf, void* *ctx) {
-  if (!qf->fortranstatus)
-    return CeedError(qf->ceed, 1,
-                     "QFunction was not set using Fortran");
+int CeedQFunctionGetInnerContext(CeedQFunction qf, void* *ctx) {
+  if (qf->fortranstatus) {
+    fContext *fctx = qf->ctx;
+    *ctx = fctx->innerctx;
```
jedbrown: Fine for now, but more extensible (possibly needed in the future) would be to have a method (function pointer) that is called if it exists so that this if statement doesn't possibly need to be built out to handle future cases.

PR #192, opened 2019-01-11
Set up CI with Azure Pipelines

jedbrown:
Runs in jedbrown/petsc Docker container (https://github.com/jedbrown/jedbrown-dockerfiles). We could add other packages like MFEM and OCCA to the container.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,20 @@
+# C/C++ with GCC
+# Build your C/C++ project with GCC using make.
+# Add steps that publish test results, save build artifacts, deploy, and more:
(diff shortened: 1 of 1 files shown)


jeremylt:
The setup looks good to me, though I can't currently see the results (jeremy.thompson@colorado.edu).

Is this to get parallel CI?

jedbrown:
Try now, under "show all checks -> details" or directly at https://dev.azure.com/CEED-ECP/libCEED/_build/results?buildId=12

The ultimate intent is to get CI using GPUs, but that will involve some further work to configure a custom Agent for Azure Pipelines.

jeremylt:
I can see it now. It would be fantastic to get GPU CI.

PR #195, opened 2019-01-15
Simple Cuda backend using one thread per element

YohannDudouit:
This backend uses a simple approach where each element is processed by one GPU thread. This approach is expected to be efficient for 1D and 2D problems, but very inefficient as soon as the kernels start to spill data out of registers, which should arise around `Q1D=4` for 3D problems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,8 @@ ref.c      := $(sort $(wildcard backends/ref/*.c))
 template.c := $(sort $(wildcard backends/template/*.c))
 cuda.c     := $(sort $(wildcard backends/cuda/*.c))
 cuda.cu    := $(sort $(wildcard backends/cuda/*.cu))
+cuda-reg.c := $(sort $(wildcard backends/cuda-reg/*.c))
+cuda-reg.cu:= $(sort $(wildcard backends/cuda-reg/*.cu))
 blocked.c  := $(sort $(wildcard backends/blocked/*.c))
 avx.c      := $(sort $(wildcard backends/avx/*.c))
 xsmm.c     := $(sort $(wildcard backends/xsmm/*.c))
@@ -263,9 +265,9 @@ ifneq ($(CUDA_LIB_DIR),)
   $(libceed) : CFLAGS += -I$(CUDA_DIR)/include
   $(libceed) : LDFLAGS += -L$(CUDA_LIB_DIR) -Wl,-rpath,$(abspath $(CUDA_LIB_DIR))
   $(libceed) : LDLIBS += -lcudart -lnvrtc -lcuda
-  libceed.c  += $(cuda.c)
-  libceed.cu += $(cuda.cu)
-  BACKENDS += /gpu/cuda
+  libceed.c  += $(cuda.c) $(cuda-reg.c)
+  libceed.cu += $(cuda.cu) $(cuda-reg.cu)
+  BACKENDS += /gpu/cuda/ref /gpu/cuda/reg
 endif
 
 # MAGMA Backend
--- a/README.md
+++ b/README.md
@@ -97,8 +97,8 @@ There are multiple supported backends, which can be selected at runtime in the e
 | `/gpu/occa`              | CUDA OCCA kernels                                 |
 | `/omp/occa`              | OpenMP OCCA kernels                               |
 | `/ocl/occa`              | OpenCL OCCA kernels                               |
-| `/gpu/cuda/ref`          | Pure CUDA kernels                                 |
-| `/gpu/cuda/reg`          | Pure CUDA kernels                                 |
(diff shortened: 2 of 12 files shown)


backends/cuda-reg/ceed-cuda-reg.c
```diff
@@ -0,0 +1,52 @@
...
+  ierr = CeedSetBackendFunction(ceed, "Ceed", ceed, "BasisCreateTensorH1",
+                                CeedBasisCreateTensorH1_Cuda_reg); CeedChk(ierr);
+  ierr = CeedSetBackendFunction(ceed, "Ceed", ceed, "BasisCreateH1",
+                                CeedBasisCreateH1_Cuda_reg); CeedChk(ierr);
+  return 0;
+}
+
+__attribute__((constructor))
+static void Register(void) {
+  CeedRegister("/gpu/cuda/reg", CeedInit_Cuda_reg, 20);
```
jeremylt: If you set this '20' and the same '20' in the two CUDA backends to be different, then it indicates which backend should be used for '/gpu/cuda' (lower is higher priority)

YohannDudouit: Thank you, I never paid attention to this parameter before.

jeremylt:
This looks tightly linked to your previous CUDA backend. One thing to consider that could reduce duplication is my approach to the pair of libXSMM backends (https://github.com/CEED/libCEED/tree/libxsmm/backends/xsmm). I put both in separate folders and shared as much common files as I could. This helps code duplication and maintenance, in my opinion.

backends/cuda/ceed-cuda.c
```diff
@@ -128,5 +141,5 @@ static int CeedInit_Cuda(const char *resource, Ceed ceed) {
 
 __attribute__((constructor))
 static void Register(void) {
-  CeedRegister("/gpu/cuda", CeedInit_Cuda, 20);
+  CeedRegister("/gpu/cuda/ref", CeedInit_Cuda, 20);
```
jeremylt: See above about setting this to a different value to give priority to one version over another.

Makefile
```diff
@@ -123,6 +123,8 @@ ref.c      := $(sort $(wildcard backends/ref/*.c))
 template.c := $(sort $(wildcard backends/template/*.c))
 cuda.c     := $(sort $(wildcard backends/cuda/*.c))
 cuda.cu    := $(sort $(wildcard backends/cuda/*.cu))
+cuda-reg.c := $(sort $(wildcard backends/cuda-reg/*.c))
```
jeremylt: If these end up being maintained in the same folder, then you can have fewer changes here.

YohannDudouit:
@jeremylt I left this backend separated from the original cuda backend since I also have a different Restriction/Prolongation, tell me what you think, I'm completely open to merging it with the cuda backend.

YohannDudouit:
I think the `ignoring #prama unroll` is because they are in a `Quote` macro, but I believe they are needed. I'll do a test.

jeremylt:
@YohannDudouit, as a heads up, we merged a new `make style` that actually works, so if you've run `make style` on this branch the old version might have missed a couple of things.

YohannDudouit:
I removed the `#pragma unroll`, the warning should have disappeared, I'll investigate the impact on performance later.

YohannDudouit:
@jedbrown I'm about to squash and merge this branch, do you have any objection?

.travis.yml
```diff
@@ -16,8 +16,8 @@ addons:
   apt:
```
jeremylt: I'm confused as to why there are changes in `travis.yml`

jeremylt: Ah, it's rebase is strangely including old changes

jedbrown:
There's a lot of unrelated commits in this branch? Is that a mistake/bad rebase?

YohannDudouit:
Yes it's a mistake/bad rebase, if you know how to solve this, I'm happy to do it.

jeremylt:
I think I got it. Does it look good to you?

README.md
```diff
@@ -97,7 +97,8 @@ There are multiple supported backends, which can be selected at runtime in the e
 | `/gpu/occa`              | CUDA OCCA kernels                                 |
 | `/omp/occa`              | OpenMP OCCA kernels                               |
 | `/ocl/occa`              | OpenCL OCCA kernels                               |
-| `/gpu/cuda`              | Pure CUDA kernels                                 |
+| `/gpu/cuda/ref`          | Pure CUDA kernels                                 |
```
jeremylt: Could you please update these descriptions to briefly describe the difference between these two backends?

YohannDudouit:
Thank you @jeremylt for fixing my rebase mess.

PR #197, opened 2019-01-22
Operator Composition

jeremylt:
This adds operator composition to libCEED.

Due to memory management considerations, I am leaving the implementation details to each backend.

ToDo:
- [x] Add to Ref
- [x] Add to Blocked/tmp/AVX
- [x] Add C tests
- [x] Add Fortran tests


jeremylt:
I need to add Fortran tests, but this is what I have set up for operator composition. As I said above, I left implementation to the backends due to expressed memory concerns for GPUs.

backends/blocked/ceed-blocked-operator.c
```diff
@@ -381,12 +381,18 @@ static int CeedOperatorApply_Blocked(CeedOperator op, CeedVector invec,
...
   // Zero lvecs
-  for (CeedInt i=0; i<numoutputfields; i++) {
-    ierr = CeedOperatorFieldGetVector(opoutputfields[i], &vec); CeedChk(ierr);
-    if (vec == CEED_VECTOR_ACTIVE)
-      vec = outvec;
-    ierr = CeedVectorSetValue(vec, 0.0); CeedChk(ierr);
-  }
+    for (CeedInt i=0; i<numoutputfields; i++) {
+      ierr = CeedOperatorFieldGetVector(opoutputfields[i], &vec); CeedChk(ierr);
+      if (vec == CEED_VECTOR_ACTIVE) {
```
jeremylt: I set the convention that only the active output gets added to - the assumption being that passive outputs would be linked to the specific sub-operators.

backends/blocked/ceed-blocked.h
```diff
@@ -22,6 +22,7 @@ typedef struct {
 } CeedBasis_Blocked;
 
 typedef struct {
+  bool add;
```
jeremylt: I added this same argument to the top of the Ref and Blocked struct to avoid duplicating the composite operator apply, but this may be fragile.

backends/magma/ceed-magma.c
```diff
@@ -1196,6 +1196,13 @@ static int CeedOperatorCreate_Magma(CeedOperator op) {
   return 0;
 }
 
+int CeedCompositeOperatorCreate_Magma(CeedOperator op) {
```
jeremylt: "Does not support" added to all non /cpu/self backends

include/ceed-backend.h
```diff
@@ -26,6 +26,7 @@
 
 #define CEED_MAX_RESOURCE_LEN 1024
 #define CEED_ALIGN 64
+#define CEED_COMPOSITE_MAX 16
```
jeremylt: I think we may want to do something like this for the total number of input/output fields and other constants.

interface/ceed-operator.c
```diff
@@ -170,13 +230,17 @@ int CeedOperatorApply(CeedOperator op, CeedVector in,
...
   CeedQFunction qf = op->qf;
 
-  if (op->nfields == 0) return CeedError(ceed, 1, "No operator fields set");
-  if (op->nfields < qf->numinputfields + qf->numoutputfields) return CeedError(
-          ceed, 1, "Not all operator fields set");
-  if (op->numelements == 0) return CeedError(ceed, 1,
-                                     "At least one restriction required");
-  if (op->numqpoints == 0) return CeedError(ceed, 1,
-                                    "At least one non-collocated basis required");
+  if (op->composite) {
```
jeremylt: I decided to use the same 'OperatorApply' to make it easier on the user. I don't think it adds much complexity here.

tests/t520-operator.c
```diff
@@ -0,0 +1,268 @@
...
+#include <math.h>
+#include "t310-basis.h"
+
+/* The mesh comprises of two rows of 3 quadralaterals followed by one row
+     of 6 triangles:
+   _ _ _
+  |_|_|_|
+  |_|_|_|
+  |/|/|/|
+
```
jeremylt: This example is on a mixed mesh - We can also support mapping the same dofs to different quadrature points, as would be helpful for the wind turbine examples

tests/t520-operator.c
```diff
@@ -0,0 +1,268 @@
...
+  // ---- Mass Tet
+  CeedOperatorCreate(ceed, qf_massTet, NULL, NULL, &op_massTet);
+  CeedOperatorSetField(op_massTet, "rho", ErestrictuiTet, CEED_NOTRANSPOSE,
+                       CEED_BASIS_COLLOCATED, qdataTet);
+  CeedOperatorSetField(op_massTet, "u", ErestrictuTet, CEED_NOTRANSPOSE,
+                       buTet, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(op_massTet, "v", ErestrictuTet, CEED_NOTRANSPOSE,
+                       buTet, CEED_VECTOR_ACTIVE);
+
+  // Hex Elements
```
jeremylt: The duplication feels tiresome, but I think it is largely due to the nature of the test I built.

YohannDudouit:
I'm having some doubts about the approach proposed for the composed operators. I would think that an approach where we evaluate first the `Interp` and the `Grad` at the same time, and then the different qFunctions choose to use either `u` or `grad(u)` (already computed), and then we apply either the `Interp` and `Grad` to the result of the qFunctions would be a better approach.
The current algorithmic approach is something like:
```
forall(operator) {
  v += opertor(u);
}
```
The algorithmic approach I'm proposing is something like:
```
interp = BasisApply(u, B, noTranspose, EvalInterp);
grad   = BasisApply(u, B, noTranspose, EvalGrad);
forall(qfunction) {
  [u_q, grad_u_q] += qfunction(interp, grad);
}
v = BasisApply(u_q, B, transpose, EvalInterp) + BasisApply(grad_u_q, B, transpose, EvalGrad);
```
This approach could be optimized further to avoid evaluating Interp or Grad when not required by any of the operators, but otherwise this reduces the algorithmic complexity and more importantly the data movements.
For instance with an equation of the type:
`grad(u).grad(v) + grad(u).(vq) + (uq).grad(v) + uv` that would be 4 operators with the proposed approach, two evaluations of grad(u), grad(v), interp(u), interp(v) whereas only one with the second approach, besides `u` is read 4 times with the first approach, but only one time with the second approach.

jeremylt:
The approach I have for CPUs is as you described, but I left the GPU implementation to the curators of those backends.

YohannDudouit:
My remark was not particularly intended for GPUs.

jeremylt:
Then I'm not sure how reordering the work to lump together different interp and grad operations with different matrices helps on CPUs. I'm open to improvements, I'm just not sure how this would help.

YohannDudouit:
It helps exactly the same way as it does on GPUs, less computation and less data movements.

jeremylt:
But it's neither of those. The sub operators operate on completely different elements

jeremylt:
@YohannDudouit pointed out that 'composition' might not be exactly the term most people might expect. To clarify, this new interface is intended to be used to aggregate multiple operators with different restrictions and bases, with either different or the same qfunction. This is not intended to compose operators working on the same restrictions and bases with different qfunctions.

jeremylt:
Good catch, thank you. I fixed it.

jeremylt:
Build is failing due to some bizarre condition in the Travis copy of the libXSMM repo. I'll fix it later.

backends/blocked/ceed-blocked.h
jeremylt: @jedbrown this is the one thing I'd most want your opinion on before merging.

jeremylt:
The code coverage checks are somehow mangled.

PR #208, opened 2019-03-08
CeedBasisApply check

valeriabarra:
This PR checks for compatibility of basis topological and geometric dimensions with lengths of input/output vectors passed to CeedBasisApply. 
This problem arose in PR #196 , where there was an attempt to apply an operator with vectors that had different topological (2D) and geometric (3D) dimensions.
Test [`tests/t308-basis.c`](tests/t308-basis.c) has been added to check for this failure.


interface/ceed-basis.c
```diff
@@ -607,8 +607,20 @@ int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *colograd1d) {
...
                    CeedEvalMode emode, CeedVector u, CeedVector v) {
   int ierr;
+  CeedInt ulength, vlength, ndof, nqpt;
   if (!basis->Apply) return CeedError(basis->ceed, 1,
-                                        "Backend does not support BasisApply");
+                                      "Backend does not support BasisApply");
+  // check compatibility of topological and geometrical dimensions
+  ierr = CeedBasisGetNumNodes(basis, &ndof); CeedChk(ierr);
+  ierr = CeedBasisGetNumQuadraturePoints(basis, &nqpt); CeedChk(ierr);
+  if (u && v) {
```
jeremylt: There must always be a `v`, so we will always have something to check. What about instead of lines 616-618 we declare `ulength=0` on 610, always do `CeedVectoGetLength` on `v`, do `CeedVectorGetLength` on `u` if `u` is not `NULL` and then always run the check on lines 619-622?

tests/tap.sh
```diff
@@ -62,6 +62,15 @@ for ((i=0;i<${#backends[@]}; ++i)); do
         continue
     fi
 
+    # grep to pass test t308 on error
```
jeremylt: We're good now, but as our test suite grows I think we'll have to occasionally check that our exceptions in here still uniquely identify error messages to make sure that 'expected failures' happened the right way.

interface/ceed-basis.c
valeriabarra: Done. Thanks!

PR #213, opened 2019-03-12
[DISCUSS] Add a new function in the interface to access the E-Vector layout.

YohannDudouit:
This opens a discussion to abstract the intermediary E-Vector data layout. Currently, it is imposed to backends to use `e-layout(dof, comp, elem) = dof + comp*ndof + elem*ncomp*ndof`. This PR adds a function that allows to access the index through a function, this is inefficient, but knowing the index is only used in tests currently. This would allow backends to use intermediary data layout optimized for their specific algorithms.

### DIFF
--- a/interface/ceed-elemrestriction.c
+++ b/interface/ceed-elemrestriction.c
@@ -309,6 +309,25 @@ int CeedElemRestrictionApply(CeedElemRestriction rstr, CeedTransposeMode tmode,
 }
 
 /**
+  @brief Get the Index of a degrees of freedom in the E-vector
+
+  @param rstr             CeedElemRestriction
+  @param[in] dof          The dof index in the element
+  @param[in] comp         The component index in the element
+  @param[in] elem         The element index
+  @param[out] index       The index of the dof in the E-Vector
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+int CeedElemRestrictionGetELayoutIndex(CeedElemRestriction rstr,
+                                       int dof, int comp, int elem, int *index) {
+  ierr = rstr->ELayout(dof, comp, elem, index); CeedChk(ierr);
+  return 0;
+}
+
+/**
   @brief Get the Ceed associated with a CeedElemRestriction
 
   @param rstr             CeedElemRestriction


interface/ceed-elemrestriction.c
```diff
@@ -308,6 +308,25 @@ int CeedElemRestrictionApply(CeedElemRestriction rstr, CeedTransposeMode tmode,
...
+  @param[in] comp         The component index in the element
+  @param[in] elem         The element index
+  @param[out] index       The index of the dof in the E-Vector
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+int CeedElemRestrictionGetELayoutIndex(CeedElemRestriction rstr,
+                                       int dof, int comp, int elem, int *index) {
```
jedbrown: `CeedInt`.  Should perhaps operate on a batch?

YohannDudouit:
I don't agree with you, the current design imposes the E-Vector through the validation of the tests, and it imposes its data format. I believe the restriction operator tests are the ones that impose the data structure.

jedbrown:
`CeedElemRestrictionApply` exists for testing and debugging purposes, not to be called in an application.

jedbrown:
`CeedElemRestrictionApply` exists for testing and debugging purposes, not to be called in an application. Moreover, it need not be implemented by a backend (could fall back to reference implementation, for example) if the author does not consider it useful for testing/debugging.

YohannDudouit:
I'm confused, the rule I've heard is "a backend has to pass all tests", the restriction tests impose a data format, can I just ignore these tests? Do you mean that `CeedElemRestrictionApply` could use a different restriction than `CeedOperatorApply`?

jedbrown:
Yes to the latter. We could also let it fail and when exclude those tests from the backend, but I think that would be less useful for a user wishing to understand whether they have defined an `CeedElemRestriction` correctly. (Of course such a user could just use a different backend for debugging.) In any case, `CeedOperatorApply` definitely has no obligation to call `CeedElemRestrictionApply`.

YohannDudouit:
Okay I'll try to do this in the cuda-reg backend to see if I understand, thank you for your answers.

jeremylt:
Is this closed/resolved?

YohannDudouit:
It is not resolved but I think it is closed.

PR #214, opened 2019-03-14
Add CeedVectorSyncArray

jeremylt:
This PR adds a minor function - CeedVectorSyncArray. This function signals to the backend to update the memory `mtype` associated with a CeedVector `vec`. The default implementation is to call `CeedVectorGetArrayRead` followed by `CeedVectorRestoreArrayRead`, but a backend can override this.

See Issue #162


jeremylt:
@valeriabarra, unfortunately I don't think that there is without GPU testing

valeriabarra:
> @valeriabarra, unfortunately I don't think that there is without GPU testing

Ah, right!

PR #217, opened 2019-03-20
lWIP] Improve LIBXSMM Backend Performance

jeremylt:
This PR increases the performance of the LIBXSMM backend by creating an of array kernel pointers when the TensorContract object is created.

### DIFF
--- a/backends/xsmm/ceed-xsmm-tensor.c
+++ b/backends/xsmm/ceed-xsmm-tensor.c
@@ -14,33 +14,26 @@
 // software, applications, hardware, advanced system engineering and early
 // testbed platforms, in support of the nation's exascale computing imperative.
 
+#include <math.h>
 #include <string.h>
-#include <libxsmm.h>
 #include "ceed-xsmm.h"
 
-// Blocked Tensor Contract
-static int CeedTensorContract_Xsmm_Blocked(CeedTensorContract contract,
-    CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
-    CeedTransposeMode tmode, const CeedInt Add, const CeedScalar *restrict u,
-    CeedScalar *restrict v) {
-  CeedScalar alpha = 1.0, beta = 1.0;
-  char transu = 'N', transt = 'N';
-  if (tmode == CEED_TRANSPOSE)
-    transt = 'T';
-
-  if (!Add)
-    beta = 0.0;
-
-  for (CeedInt a=0; a<A; a++)
-    // libXSMM GEMM
-    libxsmm_dgemm(&transu, &transt, &C, &J, &B,
-                  &alpha, &u[a*B*C], NULL, &t[0], NULL,
-                  &beta, &v[a*J*C], NULL);
-  return 0;
+// Utility functions for index in pointer array
+int CeedGetXsmmInd_Tensor(CeedInt nelem, CeedInt add, CeedTransposeMode tmode,
+                          CeedInt B, CeedInt C, CeedInt J, CeedInt currdim,
+                          CeedInt dim) {
+  return (nelem == 8 ? 1:0)*4*2*dim + (add ? 1:0)*4*dim +
+         (tmode ? 1:0)*2*dim + (B == J ? 1:0)*dim + currdim;
 }
 
-// Serial Tensor Contact
-static int CeedTensorContract_Xsmm_Serial(CeedTensorContract contract,
+int CeedGetXsmmInd_NonTensor(CeedInt add, CeedInt P, CeedInt Q, CeedInt B,
+                             CeedInt C, CeedInt J) {
+  return (C == 8 ? 1:0)*4*2 + (add ? 1:0)*4 +
+         (B == P ? (J == Q ? 0:1) : (B == Q ? 2:3));
+}
+
+// Default Tensor Contact
+static int CeedTensorContract_Xsmm_Default(CeedTensorContract contract,
     CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
     CeedTransposeMode tmode, const CeedInt Add, const CeedScalar *restrict u,
     CeedScalar *restrict v) {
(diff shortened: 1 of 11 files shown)


jedbrown:
Cool, do you have any update on performance results?

jeremylt:
Not yet. I am working on the scheme for the function pointers right now

jeremylt:
This probably can be improved upon, but I have everything set to use kernel pointers but fallback to the old implementation. It works and passes the test suite. However, sometimes `libxsmm_dmmdispatch` returns 0 rather than a valid pointer, so I need to figure out what is going on with that. (pehaps @alheinecke can help us figure that one out?) That would not prevent this branch from being merged though.

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -73,27 +66,112 @@ static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A,
...
+              const int flags = LIBXSMM_GEMM_FLAGS('N', tmode ? 'T' : 'N');
+              CeedInt B = grad ? impl->Q : (tmode ? impl->Q : impl->P),
+                      J = grad ? impl->Q : (tmode ? impl->P : impl->Q),
+                      C = nelem*CeedIntPow(J, dim);
+              int ind = CeedGetXsmmInd_Tensor(nelem, add, tmode, B, C, J,
+                                              impl->dim);
+              CeedScalar alpha = 1.0, beta = 1.0;
+              if (!add) beta = 0.0;
+              // TODO: This call does not allways actually return a kernel pointer
+              impl->kernels[ind] = libxsmm_dmmdispatch(C, J, B,
```
jeremylt: Here is the call that sometimes does not return a pointer.

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -73,27 +66,112 @@ static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A,
...
+          for (CeedInt grad = 1; grad <= impl->dim; grad+=impl->dim-1) {
+            const int flags = LIBXSMM_GEMM_FLAGS('N', tmode ? 'T' : 'N');
+            CeedInt B = tmode ? impl->P : grad*impl->Q,
+                    J = tmode ? grad*impl->Q : impl->P;
+            int ind = CeedGetXsmmInd_NonTensor(nelem, add, impl->P, impl->Q,
+                                               B, J);
+            CeedScalar alpha = 1.0, beta = 1.0;
+            if (!add) beta = 0.0;
+            // TODO: This call does not allways actually return a kernel pointer
+            impl->kernels[ind] = libxsmm_dmmdispatch(nelem, J, B,
```
jeremylt: And here

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -73,27 +66,112 @@ static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A,
...
+  else
+    ind = CeedGetXsmmInd_NonTensor(nelem, add?1:0, impl->P, impl->Q, B, J);
+
+  // Run kernel or fallback to default implementation
+  if (C != 1 && ind < impl->numkernels && ind >= 0 && impl->kernels[ind])
+    for (CeedInt a=0; a<A; a++)
+      impl->kernels[ind](&u[a*B*C], &t[0], &v[a*J*C], NULL, NULL, NULL);
   else
-    CeedTensorContract_Xsmm_Blocked(contract, A, B, C, J, t, tmode, Add, u, v);
+    CeedTensorContract_Xsmm_Default(contract, A, B, C, J, t, tmode, add, u, v);
```
jeremylt: We fall back to the default if a valid kernel was not stored or nelem != 1 or 8.

alheinecke:
sure, can you share values of the lixsmm_dmmdispatch which lead to 0 pointers. 0 pointer means that JITer didn't like the inputs.

jeremylt:
@alheinecke I have a meeting to go to now, but I'll reply as soon as I can with some combinations that did not play nice. I'm not sure how soon that will be though.

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -73,27 +66,112 @@ static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A,
...
+
+  // Get kernel index
+  if (impl->tensorbasis)
+    ind = CeedGetXsmmInd_Tensor(nelem, add?1:0, tmode==CEED_TRANSPOSE?1:0, B, C,
+                                J, impl->dim);
+  else
+    ind = CeedGetXsmmInd_NonTensor(nelem, add?1:0, impl->P, impl->Q, B, J);
+
+  // Run kernel or fallback to default implementation
+  if (C != 1 && ind < impl->numkernels && ind >= 0 && impl->kernels[ind])
```
jeremylt: I'd like to cut this conditional down in size, if its possible.

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -73,27 +66,112 @@ static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A,
...
+  CeedTensorContract_Xsmm *impl;
+  ierr = CeedCalloc(1, &impl); CeedChk(ierr);
+
+  // Set up pointers to kernels
+  ierr = CeedBasisGetTensorStatus(basis, &impl->tensorbasis); CeedChk(ierr);
+  if (impl->tensorbasis) {
+    ierr = CeedBasisGetNumNodes1D(basis, &impl->P); CeedChk(ierr);
+    ierr = CeedBasisGetNumQuadraturePoints1D(basis, &impl->Q); CeedChk(ierr);
+    ierr = CeedBasisGetDimension(basis, &impl->dim); CeedChk(ierr);
+    // Set up kernel pointer array
```
jeremylt: The kernel set up feels a bit hard to read to me? Is it at all understandable to someone who didn't write it?

jeremylt:
I'm getting an error in the serial backend in the PETSc BPs. I will investigate.

backends/xsmm/ceed-xsmm-tensor.c
jedbrown: Should we error if no valid kernel?

jeremylt:
Right now I fall back to calling the gemm (what we were doing before) if we didn't create a valid kernel, so we will still get good results.

backends/xsmm/ceed-xsmm-tensor.c
jedbrown: Okay, but why does it sometimes return NULL? Are you passing invalid arguments or it just doesn't process some sizes? Since this has performance implications, I'd like for us to understand which code path is taken, not just "randomly" use a slower implementation.

alheinecke:
yes, as I said I'm interested in the cases which return NULL. This means we couldn't JIT a kernel or there was a problem. However, I recently run the following combination for N={1..100}xM={1..100}xK={1..100}xtransb={N,T}xtransa={N}xalpha={1}xbeta={0,1} and didn't see a problem.

jeremylt:
Agreed, I want to know what is going on. I'm just not sure if we want to error on a failed kernel build in production though.

jedbrown:
Can you find a configuration that fails so we can work out why it fails? Hopefully it's something we can fix.

jeremylt:
Yeah, I can do that. A few in the test suite were misbehaving.

jeremylt:
The kernels all fail to build when transb='T'.

alheinecke:
Are you using one of the latest commits of LIBXSMM? 
 
(for samples/xgemm it works, this is eaa89b37c3517815a149849dd20738b2dd4168ce)
./kernel 32 32 32 32 32 32 1 1 0 0 0 1 nopf DP 100
------------------------------------------------
RUNNING (32x32) X (32x32)^T = (32x32), DP
------------------------------------------------
function pointer address: 4165e040
0.007781s for creating jit
0.002551s for C
2.569121 GFLOPS for C
0.000399s for libxsmm
16.430843 GFLOPS for libxsmm
max. error: 0.000000
------------------------------------------------

jeremylt:
I just checked with the most recent LIBXSMM master - no change. I previously was using master from 3 days ago.

alheinecke:
Do you see the problem also with the driver I pointed out above in your LIBXSMM clone?

jeremylt:
I'm having trouble building it - Investigating

alheinecke:
Hmm, I looked here:
https://github.com/CEED/libCEED/blob/e8de80aa80aca257ce2ea575647590b47fd89778/backends/xsmm/ceed-xsmm-tensor.c#L134
and 
https://github.com/CEED/libCEED/blob/e8de80aa80aca257ce2ea575647590b47fd89778/backends/xsmm/ceed-xsmm-tensor.c#L161

and this looks good. 

The only thing I do different is also specifying lda/ldb/ldc, but the code here looks good:
https://github.com/hfp/libxsmm/blob/master/src/libxsmm_main.c#L1994
And I believe we also have the braces correctly for your macro call:
https://github.com/hfp/libxsmm/blob/master/include/libxsmm_frontend.h#L171

alheinecke:
BTW, this is how I do it:
int l_flags = LIBXSMM_GEMM_FLAGS('N', 'N');
if ( l_trans_b ) {
  l_flags |= LIBXSMM_GEMM_FLAG_TRANS_B;
}

jeremylt:
It is working now. It seems that every so often my `git pull` mangles my copy of LIBXSMM and I need to delete it and start over.

Now my kernels all build. I should get even better performance now.

The only time we call a gemm instead of using a kernel call is when C=1.

jeremylt:
@jedbrown, I added an error message if the kernel fails to build. @alheinecke, thanks for the help!

jedbrown:
`git pull --ff-only` would prevent merging into some local changes. `git fetch && git reset --hard origin/master` makes HEAD look like upstream.

alheinecke:
Great! Please share the new performance data.

BTW: we are generating the LIBXSMM library frontend with python scripts during "make". After a pull it's always good/recommended to build LIBXSMM with "make realclean && make". This deletes any of these generated files. Not sure where this hick-up is coming from. I have filed an issue against LIBXSMM (https://github.com/hfp/libxsmm/issues/326). If by any means you still have the original hash you started  last week from, that would be super helpful.

jeremylt:
Ah, I haven't been using `make realclean`, so that probably was it.

hfp:
>  sometimes libxsmm_dmmdispatch returns 0 rather than a valid pointer

This is fixed in LIBXSMM/master revision and will be released in v1.11.

PR #218, opened 2019-03-21
Add Jenkinsfile

jedbrown:


jedbrown:
I think this is ready if we defer GPU support to another PR; cf. #219.

jedbrown:
We can add XSMM and other libraries to the image, but I'd like to have CEED-2.0 releases of them (and/or automate a pipeline for updating images, but probably not on every push).

Issue #193, opened 2019-01-12
clang-tidy


jedbrown:
clang-tidy path/to/source.c -- -Iinclude

highlights a number of places where we forgot to check return values or made invalid assumptions. There's also noise about issues that it can't show statically are actually fine, but it would be worth either adjusting the code to help the static analysis or disabling those checks at the finest possible granularity.

jeremylt:
I did a first past to handle the return values and some of the faulty logic. It still complains about `strcpy` and some of `ceed.c` and `ceed-vec.c`.

jedbrown:
Would be good practice to avoid `strcpy` anyway.

jeremylt:
Good point. There will be a follow-up PR to do more, I just wanted to make sure we get the easiest parts before v0. 5.

jeremylt:
Ok, the `strcpy`issues have been resolved. All that remains is "noise about issues that it can't show statically are actually fine".

PR #223, opened 2019-03-27
Strcpy fix

jeremylt:
This PR changes a few instances of strcpy and strcat to strncpy and strncat. See Issue #193.

After this, I think that all that remains on clang-tidy is "issues that it can't show statically are actually fine".

### DIFF
--- a/examples/navier-stokes/navierstokes.c
+++ b/examples/navier-stokes/navierstokes.c
@@ -250,7 +250,7 @@ int main(int argc, char **argv) {
   User user;
   char ceedresource[4096] = "/cpu/self";
   PetscFunctionList icsflist = NULL, qflist = NULL;
-  char problemtype[256] = "advection";
+  char problemtype[PETSC_MAX_PATH_LEN] = "advection";
   PetscInt degree, qextra, localNelem, lsize, outputfreq,
            steps, melem[3], mdof[3], p[3], irank[3], ldof[3], contsteps;
   PetscMPIInt size, rank;
@@ -358,7 +358,7 @@ int main(int argc, char **argv) {
   qextra = 2;
   ierr = PetscOptionsInt("-qextra", "Number of extra quadrature points",
                          NULL, qextra, &qextra, NULL); CHKERRQ(ierr);
-  PetscStrcpy(user->outputfolder, ".");
+  PetscStrncpy(user->outputfolder, ".", 2);
   ierr = PetscOptionsString("-of", "Output folder",
                             NULL, user->outputfolder, user->outputfolder,
                             sizeof(user->outputfolder), NULL); CHKERRQ(ierr);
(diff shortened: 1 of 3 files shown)


jeremylt:
Updated

examples/navier-stokes/navierstokes.c
```diff
@@ -358,7 +358,7 @@ int main(int argc, char **argv) {
   qextra = 2;
   ierr = PetscOptionsInt("-qextra", "Number of extra quadrature points",
                          NULL, qextra, &qextra, NULL); CHKERRQ(ierr);
-  PetscStrcpy(user->outputfolder, ".");
+  PetscStrncpy(user->outputfolder, ".", 4);
```
valeriabarra: why 4? "." is one character

examples/navier-stokes/navierstokes.c
```diff
@@ -647,7 +647,7 @@ int main(int argc, char **argv) {
   if (!icsfp)
       return CeedError(ceed, 1, "Function not found in the list");
   char str[256] = __FILE__":ICs";
-  strcat(str, problemtype);
+  PetscStrlcat(str, problemtype, 256);
```
valeriabarra: I'm not sure this is safe enough. Should we make it safer by stating that the maximum number of character copied into str should be no more than sizeof(str) - strlen(str)?

examples/navier-stokes/navierstokes.c
```diff
@@ -660,8 +660,8 @@ int main(int argc, char **argv) {
   PetscFunctionListFind(qflist, problemtype, &fp);
   if (!fp)
       return CeedError(ceed, 1, "Function not found in the list");
-  strcpy(str, __FILE__":");
-  strcat(str, problemtype);
+  PetscStrncpy(str, __FILE__":", 12);
```
valeriabarra: The constant `__FILE__` is likely to be longer than 12 characters

examples/navier-stokes/navierstokes.c
jeremylt: Should be 2, I miscounted.

examples/navier-stokes/navierstokes.c
jeremylt: Good point, I'll switch it to 256.

examples/navier-stokes/navierstokes.c
jeremylt: PetscStrlcat makes sure that we don't go past the declared length of str.

jedbrown: Note that you can use `sizeof(str)` in situations like this.

examples/navier-stokes/navierstokes.c
```diff
@@ -647,7 +647,7 @@ int main(int argc, char **argv) {
   if (!icsfp)
       return CeedError(ceed, 1, "Function not found in the list");
   char str[256] = __FILE__":ICs";
```
jedbrown: And there is `PETSC_MAX_PATH_LEN`, which may be better than a magic number.

jeremylt: Updated

Issue #181, opened 2018-12-07
Zero/Verify Output Fields Set in QFunctionApply


jeremylt:
In development of the NS example, we discovered that a QFunction author may not set all output values. This may lead to junk values in the output of the Operator. I should avoid this by modifying PR #174 to zero out the output vecs for the QFunction prior to using them.

I think I will also try to make a quick version of SetValue if the value is zero (wrapped in the same function call).

@jedbrown, @valeriabarra

jedbrown:
An alternative that would save the (small) performance cost of extra zeroing would be to verify by populating the output vectors with NaN (or using Valgrind's API to mark it undefined) and confirm that it is defined after.

jeremylt:
I thought about that as well. I suppose one question is if we want to adopt the convention that QFunction authors must explicitly set all fields or if we want to adopt the convention that all fields not set are zero. This is a nonissue if we want users to explicitly set all outputs for clarity.

Setting the memory to undefined/NaN and checking could help users debug complex QFunctions with branching logic.

I was thinking of switching to using memset  for SetValue with 0.0 to make the zeroing almost negligable.

jedbrown:
I think users should generally not declare a dependency if they don't intend to set output fields. An alternative would be to explicitly adopt the convention that QFunctions always _sum_ into their output fields, so that multiple QFunctions could be called in sequence. We'd have to test that they don't overwrite. There isn't much difference between memset and a loop in user code, FWIW.

YohannDudouit:
I don't think this is a good idea, always _sum_ means one more full memory swipe (We need to to read before writing), this has a non negligible impact on performance (especially for GPUs).

jedbrown:
I don't think it's so clear-cut.

1. explicit zeroing has the same concern about memory traffic
2. it may be resident in cache and/or fused/unrolled into registers by a JIT
3. if we want to have the abstraction of composing QFunctions, then the alternative to summing is to write into separate memory and sum those outside the QFunctions (more register/cache pressure)

jeremylt:
I agree that for clarity/debugability/etc it is preferable that users set all outputs. 

WRT summing, I think this might be the preferable long term approach as we move to operators with multiple QFunctions. It is a lot of movement to write into a separate location and then add each time. I'm thinking if there is a clean way to do that without requiring the user code to write +=.

[...]

PR #224, opened 2019-03-27
Check QFunction Output Vecs

jeremylt:
This PR is a minor change to ~~zero~~ *set* QFunction output vecs *as undefined* prior to use in the CPU backends. See Issue #181.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,14 @@ addons:
     - gfortran
     - liblapack-dev
     - libopenblas-dev
+    - valgrind-dbg
   homebrew:
     packages:
     - ccache
     - gcc
     - lapack
     - openblas
+    - valgrind
 
 env:
   - FC=gfortran
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ addons:
     - gcc
     - lapack
     - openblas
-    - valgrind
 
 env:
   - FC=gfortran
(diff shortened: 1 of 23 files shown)


jedbrown:
Do you want to merge this? The extra memory touch may not be free, so may not be desired, especially in non-reference backends. I think identifying non-conforming qfunctions is better than always paying an extra memory touch.

jeremylt:
I rearranged this initialize the qdata vectors to `NAN` and put in a warning about incomplete QFunctions in the documentation. I'm still thinking about a better long term solution.

jeremylt:
CI stalled due to Jenkins error.

jedbrown:
I'm going to leave this out of v0.4 because NAN seriously impacts performance on some architectures. Marking blocks of memory undefined using the Valgrind API may be a better option. We can decide what to do after tagging v0.4.

jeremylt:
I updated this. Now in the `/cpu/self/*` backends the output arrays of the QFunctions are flagged as undefined prior to use. According to Valgrind this *should* have no performance impact when Valgrind is not running.

ToDo: I think we need to somehow guard this include
```
#include <valgrind/memcheck.h>
```
in case Valgrind is not installed or is installed to a unrecognized location (see the Azure CI failure). This could be another reason to make a specific `memcheck` backend.

jeremylt:
@jedbrown this should be good now. I made a separate `/cpu/self/ref/memcheck` backend because Valgrind might not be installed (it wan't on CI initially). This backend flags QFunction output memory as undefined before each QFApply.

jeremylt:
This will pair nicely with the multi-delegate capability, letting someone run this QFunctiion code + AVX or XSMM TensorContract.

PR #225, opened 2019-03-27
Add CeedGetPreferredMemType

jeremylt:
This PR adds `CeedGetPreferredMemType(Ceed, CeedMemType *)`. This helps integration of libCEED into packages such as MFEM or PETSc which support CPU and GPU vectors.


backends/cuda/ceed-cuda.c
```diff
@@ -99,6 +99,12 @@ int run_kernel(Ceed ceed, CUfunction kernel, const int gridSize,
   return 0;
 }
 
+int CeedGetPreferredMemType_Cuda(CeedMemType *type) {
+  *type = CEED_MEM_HOST;
```
YohannDudouit: `CEED_MEM_DEVICE`?

interface/ceed.c
```diff
@@ -434,6 +435,27 @@ int CeedSetDelegate(Ceed ceed, Ceed *delegate) {
...
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Basic
+**/
+int CeedGetPreferredMemType(Ceed ceed, CeedMemType *type) {
+  int ierr;
+  if (ceed->GetPreferredMemType) {
+    ierr = ceed->GetPreferredMemType(type); CeedChk(ierr);
+  } else {
+    *type = CEED_MEM_DEVICE;
```
YohannDudouit: `CEED_MEM_HOST`?

jeremylt:
Good catch

backends/cuda/ceed-cuda.c
```diff
@@ -99,6 +99,12 @@ int run_kernel(Ceed ceed, CUfunction kernel, const int gridSize,
   return 0;
 }
 
+int CeedGetPreferredMemType_Cuda(CeedMemType *type) {
```
jedbrown: `static`?

include/ceed-impl.h
```diff
@@ -28,7 +28,7 @@
 #define CEED_MAX_RESOURCE_LEN 1024
 #define CEED_ALIGN 64
 
-#define CEED_NUM_BACKEND_FUNCTIONS 29
+#define CEED_NUM_BACKEND_FUNCTIONS 30
```
jedbrown: Future: We should use `offsetof` to avoid magic numbers that need to be maintained.

tests/t001-ceed.c
```diff
@@ -0,0 +1,19 @@
+/// @file
+/// Test creation and destruction of a CEED object
+/// \test Test creation and destruction of a CEED object
+#include <ceed.h>
+#include <limits.h>
+
+int main(int argc, char **argv) {
+  Ceed ceed;
+  CeedMemType type = INT_MAX;
```
jedbrown: I guess I don't see why this is meaningfully distinct from the test below, but the idea with numbers was to make the sorted list have some meaning and be easy to navigate. If we have multiple tests with the same number, we lose that.

jeremylt: Stray file, removed

backends/cuda/ceed-cuda.c
jeremylt: fixed

include/ceed-impl.h
jeremylt: I'm using `offsetof` to calculate the lookup table of backend functions; this number is the total number of backend functions across all objects. I guess I'm not sure how `offsetof` helps me know how many function pointers are in all of the object?

jedbrown: 1. Why `ceedoffsetof` when `offsetof` is defined in `stddef.h`?
2. The translation table does not need to be allocated statically.

jeremylt: 1. I was having difficulty getting `offsetof` and our opaque pointers to work together.
2. Good point - I can fix that up pretty easily.

jedbrown: Syntax would be `offsetof(struct Ceed_private, Error)`. Note that you can avoid repetition by using a macro with stringify `#` and concatenation `##`.

jedbrown: Also, `CeedSetBackendFunction` should not mix data and function pointers. Better something like this:

```c
@@ -482,9 +482,8 @@ int CeedSetBackendFunction(Ceed ceed,
   for (CeedInt i = 0; i < CEED_NUM_BACKEND_FUNCTIONS; i++) {
     if (!strcmp(ceed->foffsets[i].fname, lookupname)) {
       size_t offset = ceed->foffsets[i].offset;
-      size_t *fpointer;
-      fpointer = (size_t *)(object + offset);
-      *fpointer = (size_t) f;
+      int (**fpointer)(void) = (int (**)(void))((char*)object + offset);
+      *fpointer = f;
       return 0;
     }
   }
```

jedbrown: I was thinking something like (untested)
```c
#define CEED_FTABLE_ENTRY(class, method) \
  {#class #method, offsetof(struct class ## _private, method)}
```
then
```c
  foffset foffsets[] = {
    CEED_FTABLE_ENTRY(Ceed, Error),
    ...
    CEED_FTABLE_ENTRY(Ceed, ElemRestrictionCreateBlocked),
    ...
    CEED_FTABLE_ENTRY(CeedVector, SetArray),
```

jeremylt: Added in PR #232

PR #226, opened 2019-03-28
Ns fix alignment

valeriabarra:
This just fix a wrong alignment that was introduced in the NS PR #152


tests/t306-basis.c
```diff
@@ -2,6 +2,7 @@
 /// Test QR Factorization
 /// \test Test QR Factorization
 #include <ceed.h>
+#include <ceed-impl.h>
```
jeremylt: I don't think this is needed. Otherwise I'm happy to squash+merge this.

valeriabarra: I'm not even sure when that was introduced.. I'll remove it

examples/petsc/Makefile
```diff
@@ -14,6 +14,9 @@ LDLIBS = $(call pkgconf, --libs-only-l $(PETSc.pc) $(ceed.pc)) -lm
 bp.c := $(sort $(wildcard bp*.c))
 bp := $(bp.c:%.c=%)
 
+ns.c := ns.c
```
jeremylt: Sorry to keep missing things - The changes in this file are a holdover from when the Navier-Stokes example was in the PETSc folder, right?

valeriabarra: Thanks for noticing that. Yeah, that was very old

PR #228, opened 2019-03-28
Restriction Block

jeremylt:
This PR adds the ability to restrict to/from a single block of a blocked restriction. This ~~hopefully~~ provides a mild performance enhancement for the `cpu/self/*/blocked` backends.


jeremylt:
The CI is currently failing because of Nek, not because of this PR. Investigating.

backends/blocked/ceed-blocked-operator.c
```diff
@@ -282,29 +322,46 @@ static int CeedOperatorApply_Blocked(CeedOperator op, CeedVector invec,
       CeedChk(ierr);
       ierr = CeedQFunctionFieldGetNumComponents(qfinputfields[i], &ncomp);
       CeedChk(ierr);
+      // Restrict active input block
```
jeremylt: For now I am applying this only to the active input - I am concerned that applying this to all inputs will decrease performance by extra memory movement for unchanged input vectors

backends/blocked/ceed-blocked-operator.c
```diff
@@ -384,39 +445,37 @@ static int CeedOperatorApply_Blocked(CeedOperator op, CeedVector invec,
       case CEED_EVAL_CURL:
         break; // Not implimented
       }
-    }
-  }
+      // Restrict active output block
+      if (i == activeout) {
+        ierr = CeedOperatorFieldGetLMode(opoutputfields[i], &lmode);
```
jeremylt: I am currently only doing this for the active output, but I can (and probably should) switch to all outputs

.travis.yml
```diff
@@ -51,68 +51,24 @@ install:
...
-        if [[ $? -eq 0 ]]; then
-          git -C occa reset --hard $OCCA_HEAD;
-        else
-          rm -rf occa
-          && git clone https://github.com/libocca/occa.git
-          && git -C occa reset --hard $OCCA_HEAD;
-        fi
-    else
-        git clone https://github.com/libocca/occa.git
+  - git clone https://github.com/libocca/occa.git
```
jeremylt: All of the `git` logic was still running into problems, so I stopped caching anything but PETSc. Travis runs in about the same time.

backends/blocked/ceed-blocked-operator.c
jeremylt: Updated

jeremylt:
I included the full evector version of the blocked backend as an option. Right now the performance picture is a bit muddied.

jeremylt:
Jenkins stall
```
Cannot contact jnlp-pod-bmtbd-rxf37: hudson.remoting.RequestAbortedException: java.nio.channels.ClosedChannelException
```

jedbrown:
I restarted the Jenkins run.

jeremylt:
I cherry picked two of these commits to put it PR #231; I'll rebase this branch after that branch is merged.

jeremylt:
Failure due to Jenkins timeout.

jedbrown:
Restarted and completed. I don't understand the error. Internal routing on GKE should be reliable.

jeremylt:
I set up the serial AVX and XSMM to use this blocking technique as well. This now has some uninteresting intermediate commits, so I'd squash+merge.

interface/ceed-elemrestriction.c
```diff
@@ -308,6 +308,54 @@ int CeedElemRestrictionApply(CeedElemRestriction rstr, CeedTransposeMode tmode,
...
 
+/**
+  @brief Restrict an L-vector to a block of an E-vector or apply transpose
+
+  @param rstr    CeedElemRestriction
+  @param block   Block number to restrict to/from
+  @param tmode   Apply restriction or transpose
+  @param lmode   Ordering of the ncomp components
+  @param u       Input vector (of size @a ndof when tmode=CEED_NOTRANSPOSE)
+  @param v       Output vector (of size @a nelem * @a elemsize when tmode=CEED_NOTRANSPOSE)
```
jedbrown: These numbers aren't directly defined here. An example might be useful.

interface/ceed-elemrestriction.c
```diff
@@ -308,6 +308,54 @@ int CeedElemRestrictionApply(CeedElemRestriction rstr, CeedTransposeMode tmode,
...
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+int CeedElemRestrictionApplyBlock(CeedElemRestriction rstr, CeedInt block,
+                                  CeedTransposeMode tmode,
+                                  CeedTransposeMode lmode,
+                                  CeedVector u, CeedVector v,
+                                  CeedRequest *request) {
```
jedbrown: Should the interface be designed to support a zero-copy implementation when element restriction is the identity?

jeremylt: Two questions. 
A) Would this only be for single element processing? Multi element processing needs a copy to do the shuffle.
B) Would it be better to modify this interface or have the backend handle the special case of single element blocks with an identity restriction.

interface/ceed-elemrestriction.c
jeremylt: Updated:

```
  @param block   Block number to restrict to/from, i.e. block=0 will handle
                 elements [0 : blksize-1] and block=3 will handle elements
                 [3*blksize : 3*blksize-1]
```

interface/ceed-elemrestriction.c
jedbrown: I think this ordering is most relevant at high order where single-element processing is competitive. One could also imagine an E-vector with block/interlaced elements and an "identity" restriction in that ordering.

I'm concerned about public interfaces that may have poor granularity for some architectures. It lures users into writing fragile code that needs multiple branches for different hardware or for different application choices.

One option for zero-copy that does not change granularity is for the function to return the location where the E-vector data lives, thus having the option to return a pointer to the data where it lies, versus in a buffer.

jeremylt: I agree that this is important for performance, but I'm trying to understand the best place to put the extra complexity.

Are you worried about a user calling `CeedElemRestrictionApply` in their code or a backend calling `CeedElemRestrictionApply`?

I don't see a way that an identity restriction in transpose mode avoids branching logic in the backend's `CeedOperatorApply` code. That decision reaches all the way back to the decision of where to write the output of `CeedBasisApply` after the QFunction, or possibly where to write the output of `CeedQFunctionApply` for `CEED_EVAL_NONE`. In notransopose mode we can, and at one point used to, make the evec point to the lvec data for inputs with an identity restriction (that was dropped in the current blocked backend because of the interlacing), so that side isn't as tricky.

I worry about an interface that returns a data location rather than a vector causing issues for the multi-memory model in different backends.

Currently we copy full evecs, even for identity restrictions. Does better handling of identity restrictions best fit in this PR or a follow-on?

jedbrown: I'm mainly concerned about user code. We maintain all the backends in existence at this point, though we want to eventually have a stable way to support backends maintained externally.

With the current interface, libCEED could determine that the restriction is the identity and make the output vector alias the block of the input vector (instead of copying into separate memory). It could mark the result read-only, but would need some way of releasing the aliased vector. Perhaps a Get/Restore interface. For the transpose mode, we could Get a block of the E-vector for writing and Restore it when done.
* Identity restriction and matching memory spaces
  * Get sets a writeable view; Restore drops the view
* Otherwise
  * Get ensures a writeable buffer; Restore applies the transpose restriction.

jeremylt: This is starting to sound like a follow-on PR in terms of scope. This PR already provides performance enchantment, so maybe we merge this one and start a new PR better handling identity restrictions?

jedbrown: Yes, that's fine with me.

PR #231, opened 2019-04-01
Travis updates

jeremylt:
This PR changes the Travis yaml to fix the failing Nek tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ os:
 
 dist: xenial
 
-osx_image: xcode10.1
+osx_image: xcode10.2
 
 compiler:
   - gcc
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,11 +50,11 @@ install:
   - cd Nek5000/tools && ./maketools genbox genmap reatore2 && cd ../..;
   - export NEK5K_DIR=$PWD/Nek5000 PATH=$PWD/Nek5000/bin:$PATH;
 # PETSc
-  - export PETSC_INSTALL=$HOME/install/petsc-3.10.0
+  - export PETSC_INSTALL=$HOME/install/petsc-3.11.0
   - test -s "$PETSC_INSTALL/lib/pkgconfig/PETSc.pc"
-        || ( curl -O http://ftp.mcs.anl.gov/pub/petsc/release-snapshots/petsc-lite-3.10.0.tar.gz
-        && tar xf petsc-lite-3.10.0.tar.gz
-        && cd petsc-3.10.0
+        || ( curl -O http://ftp.mcs.anl.gov/pub/petsc/release-snapshots/petsc-lite-3.11.0.tar.gz
+        && tar xf petsc-lite-3.11.0.tar.gz
+        && cd petsc-3.11.0
         && ./configure --with-debugging=0 COPTFLAGS=-O --with-mpi=0 --with-fc=0 --with-cxx=0 --prefix=$PETSC_INSTALL
         && make
         && make install )
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,93 +26,30 @@ addons:
     - openblas
 
 env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -51,68 +51,24 @@ install:
...
-        if [[ $? -eq 0 ]]; then
-          git -C occa reset --hard $OCCA_HEAD;
-        else
-          rm -rf occa
-          && git clone https://github.com/libocca/occa.git
-          && git -C occa reset --hard $OCCA_HEAD;
-        fi
-    else
-        git clone https://github.com/libocca/occa.git
+  - git clone https://github.com/libocca/occa.git
```
jedbrown: Let's add `--depth=1` to all the clones. Reduces size from 17M to 600K in this case.

jedbrown: Erm, won't work for a specific hash. I guess this https://stackoverflow.com/questions/31278902/how-to-shallow-clone-a-specific-commit-with-depth-1

jeremylt: I dropped the `--depth=1` argument on OCCA last year when we were having trouble with it's stability effecting our tests. Do we want to drop the testing against a 'known good' commit of OCCA now that we haven't had that issue in a while?

jeremylt:
The builds are still failing on Nek. I'll work on figuring out what they changed in the last 2 days.

jeremylt:
Nek isn't properly sending optional flags over to their make system. I'm filing a bug with them.

.travis.yml
```diff
@@ -26,93 +26,31 @@ addons:
...
-          && git clone --depth 1 https://github.com/Nek5000/Nek5000.git;
-        fi
-    else
-        git clone --depth 1 https://github.com/Nek5000/Nek5000.git;
-    fi
+  - git clone https://github.com/Nek5000/Nek5000.git && 
+      git -C Nek5000 reset --hard d5c351b;
   - cd Nek5000/tools && ./maketools genbox genmap reatore2 && cd ../..;
-  - export NEK5K_DIR=$HOME/install/Nek5000 PATH=$HOME/install/Nek5000/bin:$PATH;
+  - export NEK5K_DIR=$PWD/Nek5000 PATH=$PWD/Nek5000/bin:$PATH;
```
jeremylt: This should be temporary - @thilinarmtb said they'll work on fixing the bug I found. Until then we can test against Nek5000 19.0-rc2

.travis.yml
jeremylt: I dropped it for OCCA, but am temporarily pulling more for Nek. See below

jeremylt:
Nek master fixed

jeremylt:
Travis is now passing without these changes but I think they still will help prevent issues. Sometimes the more complicated git commands were causing Travis failures.

jedbrown:
Travis seems to be super slow.

jeremylt:
Agreed. The 2 OSX instance limit and a current OSX backlog isn't helping us. Our Linux jobs are running more or less promptly.

jeremylt:
Travis is wrapping up, but with these changes it looks like the 2 Linux builds will take 4-5 minutes each and 2 OSX builds will take 10-11 minutes each, so a total run time of 28-32 minutes (hopefully). This varies based on instance limits, etc.

jeremylt:
One of the biggest sources of variability in the CI times is codecov, as this is not uncommon, especially on OSX:
```
    -> Pinging Codecov
https://codecov.io/upload/v4?package=bash-8a28df4&token=&branch=travis-fix&commit=0dc1c4ca56f363fd35d05f33059c5b80c51813d5&build=3536.4&build_url=&name=&tag=&slug=CEED%2FlibCEED&service=travis&flags=tests&pr=false&job=514409454
    -> Uploading
    X> Failed to upload
    -> Sleeping for 30s and trying again...
```

jedbrown:
I'm not aware of a similarly attractive code coverage tool.

PR #234, opened 2019-04-04
QFunction Inputs

jeremylt:
This ~~WIP~~ PR is converting CPU, CUDA, and OCCA QFunctions to use a common input/output strategy in order to move towards single source QFunctions.


jeremylt:
@dmed256 @camierjs @YohannDudouit 

I know that OCCA backends are coming soon. 

I am working on converting the OKL QFunctions from
```
@kernel void setup(void *ctx, CeedInt Q,
                   const int *iOf7, const int *oOf7, 
                   const CeedScalar *in, CeedScalar *out) {
  const CeedScalar *weight = in + iOf7[0];
  const CeedScalar *J = in + iOf7[1];
  CeedScalar *rho = out + oOf7[0];

  for (int i=0; i<Q; i++; @tile(TILE_SIZE,@outer,@inner)) {
    rho[i] = weight[i] * (J[i+Q*0]*J[i+Q*3] - J[i+Q*1]*J[i+Q*2]);
  }
}
```
to
```
@kernel void setup(void *ctx, CeedInt Q,
                   CeedQFunctionArguments ags) {
  const CeedScalar *weight = args.in[0];
  const CeedScalar *J = args.in[1];
  CeedScalar *rho = args.out0];

  for (int i=0; i<Q; i++; @tile(TILE_SIZE,@outer,@inner)) {
    rho[i] = weight[i] * (J[i+Q*0]*J[i+Q*3] - J[i+Q*1]*J[i+Q*2]);
  }
}
```
where
```

typedef struct {
  const CeedScalar *in[16];
  CeedScalar *out[16];
} CeedQFunctionArguments;
```

This should move us significantly closer to a single source QFunction, which lets us have a QFunction gallery (and thus helps the libParanumal backend) and help me for adding AD.

dmed256:
@jeremylt The kernels seem to parse properly in Serial and OpenMP modes 🎉 

However, there are still issues with defining variables outside of `@outer` loops

```cpp
@kernel void f_build_mass(int *ctx, CeedInt Q,
                          const int *iOf7, const int *oOf7,
                          const CeedScalar *u, CeedScalar *v) {
  const CeedScalar *J = u + iOf7[0];
  const CeedScalar *qw = u + iOf7[1];
  CeedScalar *qd = v + oOf7[0];
  //const struct BuildContext *bc = (struct BuildContext*)ctx;
  const int dim = ctx[0];
  const int space_dim = ctx[1];
  const CeedInt dims = dim + 10*space_dim;

  for (int i=0; i<Q; i++; @tile(TILE_SIZE,@outer,@inner)) {
```
The variables should be captured and passed to the generated CUDA/OpenCL/HIP kernels in the `@outer` loop regions.

dmed256:
I'll put supporting variable capture as a top priority!

jeremylt:
I was just about to test on CUDA, so thanks for the heads up. For now it should be relatively easy to move the
```
  for (int i=0; i<Q; i++; @tile(TILE_SIZE,@outer,@inner)) {
```
line to the right spot.

dmed256:
Sounds good!

Cnce libocca/occa#222 is done I can update the kernels to avoid defining these values at each loop for the CPU modes (shouldn't affect GPU modes). We might be able to remove the `iOf7` variables as well, something I can try to help with later.

jedbrown:
Note that we've also had discussions of making the q-functions not be OCCA kernels, but merely called from kernels (inlinable since we have their source for the JIT). In that case, we can just call with `Q=1` on GPUs and let the outer kernel constructed by libCEED handle strides. Similar for NVRTC.

YohannDudouit:
To me the `Q=1` hack doesn't work since `Q` does not only appear in the loop boundary.

jedbrown:
@YohannDudouit Can you explain?

jeremylt:
I see this as a relatively 'quick' task that at least gets us much closer to single source. If we can get down to loop contents and location as the only differences, I see that as a win.

dmed256:
@jedbrown 👍 I like the idea of specifying only q update and JIT compiling the kernel we build around that q update function/code

Something like this?
```cpp
@kernel void mass(void *ctx, CeedInt Q,
                  const int *iOf7, const int *oOf7,
                  const CeedScalar *in, CeedScalar *out) {
  for (int i=0; i<Q; i++; @tile(TILE_SIZE,@outer,@inner)) {
    const CeedScalar *rho = in + iOf7[0];
    const CeedScalar *u = in + iOf7[1];
    CeedScalar *v = out + oOf7[0];

    v[i] = rho[i] * u[i];
  }
```
&darr;
```cpp
CeedScalar massUpdate(const CeedScalar rho,
                      const CeedScalar u) {
  return rho[i] * u[i];
}
```

jedbrown:
Note that the input and output fields could have multiple components, but it could be something similar to what you wrote. Alternatively/preferably we could call the source of the CPU-style qfunctions with `Q=1` available to the JIT. The compiler should be able to inline and eliminate the load/stores implicit in that model, but would not prevent a non-JIT compiler from doing CPU and OMP target optimizations.

jeremylt:
@jedbrown I think @YohannDudouit is referring to QFunctions like:
```
static int Diff(void *ctx, CeedInt Q, CeedQFunctionArguments args) {
  const CeedScalar *ug = args.in[0], *qd = args.in[1];
  CeedScalar *vg = args.out[0];

  for (CeedInt i=0; i<Q; i++) {
    const CeedScalar ug0 = ug[i+Q*0];
    const CeedScalar ug1 = ug[i+Q*1];
    const CeedScalar ug2 = ug[i+Q*2];
    vg[i+Q*0] = qd[i+Q*0]*ug0 + qd[i+Q*1]*ug1 + qd[i+Q*2]*ug2;
    vg[i+Q*1] = qd[i+Q*1]*ug0 + qd[i+Q*3]*ug1 + qd[i+Q*4]*ug2;
    vg[i+Q*2] = qd[i+Q*2]*ug0 + qd[i+Q*4]*ug1 + qd[i+Q*5]*ug2;
  }
  return 0;
}
```

jedbrown:
Ah, that `Q` is both number of points to process and "leading dimension" for multi-dimensional arrays? We could support that by either making a separate stride argument or expanding the pointers so that indexing inside a q-function is ultimately all 1D arrays. Thoughts?

YohannDudouit:
Yes that's what I was referring to, if you make `Q=1` you change the data layout in some way. We could maybe get around this by loading `ug` into register first, and then give the register array pointer instead of the initial pointer, otherwise we would loose the nice coalesce memory access when Q is large.

We were discussing yesterday with @dmed256 to have a signature of the type:
```
void massUpdate(const int q, const int Q, const CeedScalar* in, CeedScalar* out) {
  const CeedScalar rho = in[0][q];
  const CeedScalar u = in[1][q];
  CeedScalar v = out[o][q];
  v = rho* u;
}
```
The concern was that this would be inefficient on CPU because you can't inline this. So we started thinking of a Macro that would generate the previous function as a string, and also a CPU function the way it is currently donein libCEED.

jedbrown:
What about this?
```c
static int Diff(void *ctx, CeedInt Q, CeedInt N,
  const CeedScalar *ug = in[0], *qd = in[1];
  CeedScalar *vg = out[0];
  for (CeedInt i=0; i<Q; i++) {
    const CeedScalar ug0 = ug[i+N*0];
    const CeedScalar ug1 = ug[i+N*1];
    const CeedScalar ug2 = ug[i+N*2];
    vg[i+N*0] = qd[i+N*0]*ug0 + qd[i+N*1]*ug1 + qd[i+N*2]*ug2;
    vg[i+N*1] = qd[i+N*1]*ug0 + qd[i+N*3]*ug1 + qd[i+N*4]*ug2;
    vg[i+N*2] = qd[i+N*2]*ug0 + qd[i+N*4]*ug1 + qd[i+N*5]*ug2;
  }
  return 0;
}
```

YohannDudouit:
I think you always access with `i=0` the different inputs/outputs, thus you would need to change the pointer?

jedbrown:
Sure, the caller needs to increment their base pointers for the arrays, but that indexing needs to be done somewhere and it should still inline fine.

jeremylt:
I just confirmed, this works for the OCCA and CUDA backends.

jeremylt:
Travis is erroring because MFEM is not building. Investigating.

jeremylt:
I added the stride argument. Now we have:

```
static int Diff(void *ctx, CeedInt Q, CeedInt N, CeedQFunctionArguments args) {
  const CeedScalar *ug = args.in[0], *qd = args.in[1];
  CeedScalar *vg = args.out[0];

  for (CeedInt i=0; i<Q; i++) {
    const CeedScalar ug0 = ug[i+N*0];
    const CeedScalar ug1 = ug[i+N*1];
    const CeedScalar ug2 = ug[i+N*2];
    vg[i+N*0] = qd[i+N*0]*ug0 + qd[i+N*1]*ug1 + qd[i+N*2]*ug2;
    vg[i+N*1] = qd[i+N*1]*ug0 + qd[i+N*3]*ug1 + qd[i+N*4]*ug2;
    vg[i+N*2] = qd[i+N*2]*ug0 + qd[i+N*4]*ug1 + qd[i+N*5]*ug2;
  }
  return 0;
}
```

@YohannDudouit, is this closer to something that the CUDA backend can ingest to create the information in the `.cu` files?

jeremylt:
TODO: Replace `__FILE__` with `__DIR__`/`PATH(foo.h)` if it is accepted in PR #243.

include/ceed.h
```diff
@@ -298,23 +304,22 @@ CEED_EXTERN int CeedQRFactorization(CeedScalar *mat, CeedScalar *tau, CeedInt m,
...
-/// @param out - array of pointers to each output array in the order provided
-///              using CeedQFunctionAddOutput().  The shapes are as above for
-///              \a in.
+/// @param args - struct of pointers to each argument in the order provided
+///              by the user in CeedQFunctionAddInput().  Each array has shape
+///              `[dim, ncomp, Q]` where `dim` is the geometric dimension for
+///              \ref CEED_EVAL_GRAD (`dim=1` for \ref CEED_EVAL_INTERP) and
+///              `ncomp` is the number of field components (`ncomp=1` for
+///              scalar fields).  This results in indexing the `i`th input at
+///              quadarture point `j` as `in[i][(d*ncomp + c)*Q + j]` and the
```
jedbrown: "quadarture" (spelling)

tests/t400-qfunction-f.okl
```diff
@@ -18,18 +18,17 @@
 typedef int CeedInt;
 typedef double CeedScalar;
 // OCCA parser doesn't like __global here
-typedef /*__global*/ double gCeedScalar;
+//typedef __global double gCeedScalar;
```
jedbrown: Commented-out dead code smells bad.

tests/t400-qfunction.cu
```diff
@@ -15,9 +15,11 @@
 // testbed platforms, in support of the nation's exascale computing imperative.
 
 // *****************************************************************************
-extern "C" __global__ void setup(void *ctx, CeedInt Q, Fields_Cuda fields) {
-  const CeedScalar *w = (const CeedScalar *)fields.inputs[0];
-  CeedScalar *qdata = fields.outputs[0];
+extern "C" __global__ void setup(void *ctx, CeedInt Q,
+                                 CeedQFunctionArguments args) {
```
jedbrown: We added the `N` argument to support CUDA layouts, but aren't using it here?

tests/t400-qfunction-f.okl
jeremylt: Borked change and revert. I'll track down all versions of this.

jeremylt:
@YohannDudouit can you verify that the tests to the CUDA backends in this branch work?

Also, I'm going to squash all of these down when I merge. Can you help me draft a summary of why we're using the args struct?

jeremylt:
Draft merge/squash message:

Merge PR 234, more closely align CUDA and CPU QFunctions, use struct of inputs/outputs to avoid double indirection for GPU performance.

jeremylt:
Update: @YohannDudouit is working on a CUDA codegen backend and may have some refinements to this PR based on that work.

jeremylt:
Closed in favor of #288

Issue #219, opened 2019-03-22
CI on GPUs


jedbrown:
If anyone can set up a Docker image for a Jenkins agent that has CUDA, we can turn on CI for GPUs. See this issue:  https://github.com/CEED/kubernetes-jenkins/issues/1

jeremylt:
It appears I don't have access to the link.

jedbrown:
Should work now.

PR #235, opened 2019-04-05
Test gitlab-ci

jedbrown:


jedbrown:
This can merge after #591

jedbrown:
This is partially addresses #219 in that it provides CI for HIP on ROCm devices. We still don't have CI for CUDA.

jedbrown:
gitlab.com/ceed is taken, but I could shift the GitLab pipeline to gitlab.com/libceed/libceed (instead of routing it through my fork)

@jeremylt It looks like you have a mirror/integration configured? https://gitlab.com/jeremylt/libCEED/-/pipelines/170497959
Should it be disabled?

jeremylt:
It probably should be. I was playing around with something that I didn't quite get to work.

jeremylt:
Question - do we want to do code coverage of these backends as well?

jedbrown:
That would be great. Noether has `gcov` installed. Could you set it up in `.gitlab-ci.yml`?

.gitlab-ci.yml
```diff
@@ -0,0 +1,18 @@
...
+  - test
+
+noether-rocm:
+  stage: test
+  tags:
+    - rocm
+  image: jedbrown/rocm:latest
+  script:
+    - make -j$(nproc)
+    - echo '[{"subject":"/","metrics":[{"name":"Transfer Size (KB)","value":"19.5","desiredSize":"smaller"},{"name":"Speed Index","value":0,"desiredSize":"smaller"},{"name":"Total Score","value":92,"desiredSize":"larger"},{"name":"Requests","value":4,"desiredSize":"smaller"}]}]' > performance.json
```
jedbrown, line 11: @jeremylt This is a placeholder that would show performance if we used GitLab Merge Requests. I'm hoping that like tests, it will eventually become available in the pipeline view.

jeremylt: Would you mind if I throw a `make info` in here? It's in our other builds and I've found it useful for quick debugging in the past.

jeremylt:
> That would be great. Noether has `gcov` installed. Could you set it up in `.gitlab-ci.yml`?

Would it be reasonable to add lcov to Noether? Without it the results from Travis and Noether will differ quite a lot.

.gitlab-ci.yml
jedbrown: Go for it.

jedbrown:
I installed `lcov`.

jeremylt:
It looks like GCC 9.3 and 10.1 has a regression that affects coverage: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95847 (I verified locally with 10.1 and on Noether with 9.3). Travis uses GCC 7.5 so we don't see this issue there.

We can
a) use a different compiler
b) roll back to an old gfortran
c) drop coverage until the compiler bug is fixed

My preference would be to install Clang on Noether if its not too much bother.

jedbrown:
Noether has clang-10 and gcc-8/gfortran-8 (and 9 and 10).

jeremylt:
Weird. I get nothing from `clang --version`

jedbrown:
Tab is your friend
```
$ clang-10 --version
Debian clang version 10.0.1-+rc4-1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

jeremylt:
Yeah, I just found that. I got confused because my system `clang` points to `clang-10` like `gcc` points to `gcc-10` on Noether.

I just pushed using gcc8; everything compiles and runs fine when I ssh in and use gcc8.

jeremylt:
Ok, I think everything but the CodeCov tokens is set - I don't I have the right permissions for that.
```
HTTP 400
Please provide the repository token to upload reports via `-t :repository-token`
```
Also, we could add
```
    - clang-tidy --version && TIDY_OPTS="-fix-errors" make -j$(nproc) tidy && git diff --exit-code;
```
but unless I'm missing it, `clang-tidy` isn't on Noether yet.

jedbrown:
Codecov report now includes HIP. https://codecov.io/gh/CEED/libCEED/tree/jed%2Fgitlab-ci/backends/hip
Congrats @nbeams on 96% coverage without metrics!

jeremylt:
There is some stuff that should be excluded, like certain error messages. I can add those exclusion markers in a follow-up for CUDA and HIP if we are ready to merge this now.

jedbrown:
Yeah, let's do that in a follow-up.

PR #237, opened 2019-04-17
Travis MFEM fix

jeremylt:
This PR fixes a small problem when building MFEM on Travis. This problem does not occur on my local machine.

To see the issue: https://travis-ci.org/CEED/libCEED/builds/521378750

The flag added to fix: `-std=c++11`


jeremylt:
Thanks, I'll shorten it up before merging.

PR #239, opened 2019-04-29
Add delegates for specific objects

jeremylt:
This PR adds the ability to delegate to different reference CEEDs for different objects. For example, we will be able to use `/cpu/self/ref/blocked` as the basic delegate, `/cpu/self/xsmm/blocked` as the TensorContract delegate, and `/cpu/self/juliadiff/blocked` as the QFunction delegate for our future "AD via Julia Package" backend.

This approach will help us avoid combinatorial explosion in the list of CPU backends (e.g. `/cpu/self/xsmm/blocked/juliadiff/futurecoolstuff`.

A note for testing: I am testing this by making a template backend that uses 4 different `/cpu/self/ref/blocked` CEEDs. I want to test this functionality without needing XSMM or AVX. This still provides an adequate test as these 4 CEEDs are independent. (Though if I'm careful and it's desirable I can make the test a mix of `/cpu/self/ref/blocked` and `/cpu/self/ref/serial`.)

### DIFF
--- a/backends/template/ceed-tmpl-sub.c
+++ b/backends/template/ceed-tmpl-sub.c
@@ -1,0 +1,52 @@
+// Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory. LLNL-CODE-734707.
+// All Rights reserved. See files LICENSE and NOTICE for details.
+//
+// This file is part of CEED, a collection of benchmarks, miniapps, software
+// libraries and APIs for efficient high-order finite element and spectral
+// element discretizations for exascale applications. For more information and
+// source code availability see http://github.com/ceed.
+//
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+#include <ceed-backend.h>
+#include <string.h>
+
+static int CeedInit_Tmpl(const char *resource, Ceed ceed) {
+  int ierr;
+  if (strcmp(resource, "/cpu/self")
+      && strcmp(resource, "/cpu/self/tmpl/sub"))
+    return CeedError(ceed, 1, "Tmpl backend cannot use resource: %s", resource);
+
+  Ceed ceedref;
+  // Create refrence CEED that implementation will be dispatched
+  //   through unless overridden
+  CeedInit("/cpu/self/ref/blocked", &ceedref);
+  ierr = CeedSetDelegate(ceed, &ceedref); CeedChk(ierr);
+
+  // Create refrence CEED for objects
+  Ceed basisceedref;
+  CeedInit("/cpu/self/ref/blocked", &basisceedref);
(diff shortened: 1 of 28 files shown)


jeremylt:
ok, something went wierd with the branching, but I'll fix it up.

interface/ceed.c
```diff
@@ -436,6 +436,68 @@ int CeedSetDelegate(Ceed ceed, Ceed *delegate) {
...
+  @brief Set a delegate CEED for a specific object type
+
+  @param ceed           Ceed to set delegate of
+  @param[out] delegate  Address to set the delegate to
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+int CeedSetObjectDelegate(Ceed ceed, Ceed *delegate, const char *objname) {
```
jedbrown: This function only ever uses `*delegate` so presumably it can accept `Ceed delegate` instead.

backends/template/ceed-tmpl-sub.c
```diff
@@ -0,0 +1,52 @@
...
+  CeedInit("/cpu/self/ref/blocked", &opceedref);
+  ierr = CeedSetObjectDelegate(ceed, &opceedref, "Operator");
+  CeedChk(ierr);
+
+  return 0;
+}
+
+__attribute__((constructor))
+static void Register(void) {
+  CeedRegister("/cpu/self/tmpl/sub", CeedInit_Tmpl, 70);
```
jedbrown: Does this need to be a backend in the library or could it go in a test so we don't have to document and test it for all users?

interface/ceed.c
jeremylt: Fixed, and for `CeedSetDelegate` as well.

backends/template/ceed-tmpl-sub.c
jeremylt: Perhaps we should replace the current `/cpu/self/tmpl` backend with `/cpu/self/tmpl/sub`?

This feature is intended to create backends from multiple different backends, so it feels like making a backend and running it through the test suite is the right way to test it?

I agree that the documentation needs to be improved.

As it is, I'm not wild about the template backend being built in a standard build of the library. Maybe we need a way to build the template backend(s) only for testing, like in Travis?

jedbrown: We could build a `libceed-test.so` if all the tests need to run it, or we could just have tests that need it define it (by including a header or linking a common object file).

jeremylt: I started the work but the Makefile dependencies are broken in parallel.

jedbrown: Please have a look at https://github.com/CEED/libCEED/tree/jed/libceed-test

The `prove` target would still need to filter tmpl out of backends for non-t* (it's done in the makefile for `test` and `junit` targets.

jedbrown: Branch is updated, but there is some Travis issue that I haven't been able to reproduce locally.

jeremylt: I rebased to your branch and squashed out the debugging commits. I think we're doing pretty good right now. I left many of the commits, and 686de4a specifically as a separate so that I can more easily change our solution to the `--no-as-needed` issue in the future.

jeremylt:
Arg, I've had the same thing occur intermittently in the past. @thilinarmtb any thoughts on why we occasionally get failing results on Nek BP1 that dissappear on reruns?

PR #241, opened 2019-05-01
Update nek5000 example scripts

thilinarmtb:


jedbrown:
@thilinarmtb What's the status here?

jeremylt:
@thilinarmtb, anything I can help with to get this wrapped up?

examples/nek5000/generate-boxes.sh
```diff
@@ -54,13 +54,18 @@ function genbb()
   cp $1.box ttt.box
   echo
   echo "Running genbox ..."
-  printf "ttt.box\n" | $NEK5K_DIR/bin/genbox || return 1
+  if [ -z ${NEKTOOLS_DIR+x} ]; then
```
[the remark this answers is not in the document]
thilinarmtb: I will remove the +x it is not relevant now.

thilinarmtb:
It is almost done. Let me test locally and fix it today if necessary.

jedbrown:
@thilinarmtb What do we need now to get Travis passing on the merge? A rebase onto (or merge from) 'master' may be needed.

thilinarmtb:
I will fix the Travis error. Looks like it can't find `generate-boxes.sh`

jedbrown:
@thilinarmtb  At some point (here or elsewhere), can you prevent the Nek tests from writing `xyzb6.fld01` to the root directory? Also, is it possible that the two Nek tests interfering with each other's file IO is causing the intermittent test failures?

thilinarmtb:
I know why they fail and I am trying to fix it in a cleaner way. I am basically trying
to get rid of all the scripts and replace it with one in a cleaner way. 

`xyz{mesh}.fld01` is dumped when there is an issue with the mesh -- which we have in
this case. I will fix this soon.

thilinarmtb:
@jedbrown @jeremylt  I think this is ready. I can do further polishing later.
Please squash the commits together.

jeremylt:
This is failing locally for me, with the box folder `b6` being built at the top level of the libCEED directory instead of the build directory for some strange reason. I think its not properly handling when `NEK5K_DIR` is not set but Nek is in a sibling directory to libCEED. However, setting `NEK5K_DIR` does not fix the problem.

Error:
```
nek-bp1 ................. 
1..21
ok 1 nek-bp1 /cpu/self/ref/serial
not ok 2 nek-bp1 /cpu/self/ref/serial stdout
# + At line 153 of file genbox.f (unit = 8)
# + Fortran runtime error: File already opened in another unit
# + 
# + Error termination. Backtrace:
# + #0  0x7faabf949b0d in already_open
# + at /build/gcc/src/gcc/libgfortran/io/open.c:714
# + #1  0x5644cb55c5e4 in ???
# + #2  0x5644cb55026e in ???
# + #3  0x7faabf3e5ce2 in ???
# + #4  0x5644cb5502ad in ???
# + #5  0xffffffffffffffff in ???
# + Error generating box. See /home/jeremy/Dev/libCEED/b6/log for details.
not ok 3 nek-bp1 /cpu/self/ref/serial stderr
# +build/nek-bp1: line 228: cd: ./boxes: No such file or directory
# +sed: can't read b.box: No such file or directory
# +cp: cannot stat 'b1e.rea': No such file or directory
# +build/nek-bp1: line 273: cd: ./boxes/b6: No such file or directory
# +build/nek-bp1: line 281: build/bp1.cpuselfrefserial.log.1.b6: No such file or directory
# +grep: build/bp1.cpuselfrefserial.log*: No such file or directory
```

.travis.yml
```diff
@@ -49,7 +49,7 @@ install:
 # Nek5k
   - git clone --depth 1 https://github.com/Nek5000/Nek5000.git;
   - cd Nek5000/tools && ./maketools genbox genmap reatore2 && cd ../..;
-  - export NEK5K_DIR=$PWD/Nek5000 PATH=$PWD/Nek5000/bin:$PATH;
+  - export NEK5K_DIR=$PWD/Nek5000 NEK_BOX_DIR=./build/boxes PATH=$PWD/Nek5000/bin:$PATH;
```
jedbrown: I don't like this because `make prove-all` doesn't work unless the user manually sets `NEK_BOX_DIR`. Does this need to be a default set in `Makefile`?

jeremylt: I think this issue is why I previously set the scrit to build the boxes in the `examples/nek5000` dir and copied them over rather than trying to build them in the `build` dir. It avoids this special case.

jedbrown: Surely it's possible for `genbox` to put the boxes in a location provided as a command-line argument?

jeremylt: I'm not sure, but we can definitely make it so for this script. @thilinarmtb, I also noticed that `make examples` fails, though it does it on master as well.

Current status of ways to build the Nek examples in this branch:

`make examples` from top level - fails
```
make: *** No rule to make target 'build/nek-bp1', needed by 'examples'.  Stop.
```

`make nek` in `examples` folder - fails 
```cd nek5000 && NEK5K_DIR=/home/jeremy/Dev/Nek5000 CEED_DIR=/home/jeremy/Dev/libCEED \
	 FC=mpif77 CC=mpicc \
	 ./nek-examples.sh -e bp1
Box size not specified. Try setting it with -b option.
Try ./nek-examples.sh -h for more help.
make: *** [Makefile:42: nek] Error 1
```

`./nek-examples.sh -m` in `examples/nek5000` folder - succeeds

`make prove` at  top level - fails as described above without special variable

jedbrown: @thilinarmtb What's your timeline to address the above so we can merge?

thilinarmtb: @jedbrown : I will try to finish before this weekend. Bit busy these days.

thilinarmtb: @jedbrown: It is not possible for genbox to put the boxes in a designated dir. NEKBOX_DIR is a default that needs to be set in makefile. How do you suggest we resolve this?

jedbrown: If it were me, I'd add a `-o`/`--output` option to `genbox`. We shouldn't allow our productivity to suffer for eternity because Paul neglected a command-line utility best practice in 1994. If not that, the obvious thing would be to have a makefile rule that runs `genbox` in a temp directory and moves the result. We can't use `examples/nek5000/Makefile` because `makenek` writes its own (I still don't understand why that's considered acceptable), but we can write such a thing into the shell script or a higher level makefile (e.g., `examples/Makefile`). The complexity cannot be forced upon the user.

thilinarmtb: Sure, I will fix this thing by tomorrow.

thilinarmtb:
Now `make prove` works but not `make prove-all`.

jeremylt:
Right, `make prove` and `make prove-all` are not building the library first.

Makefile
```diff
@@ -407,15 +411,12 @@ tst : ;@$(MAKE) $(MFLAGS) V=$(V) test
 ctc-% : $(ctests);@$(foreach tst,$(ctests),$(tst) /cpu/$*;)
 
 prove : BACKENDS += $(TEST_BACKENDS)
-prove : $(matched_prereq)
+prove :
```
jeremylt: Either `$(matched)` or `$(matched_req)` is needed here.

jeremylt: Update: Adding `$(matched)` is fixing everything but the Nek examples for me

jeremylt: Error message:
```
ok 1 nek-bp1 /cpu/self/ref/serial
not ok 2 nek-bp1 /cpu/self/ref/serial stdout
# + build/nek-bp1: line 213: /home/jeremy/Dev/libCEED/genbox: No such file or directory
# + Error generating box. See /home/jeremy/Dev/libCEED/build/boxes/b6/log for details.
not ok 3 nek-bp1 /cpu/self/ref/serial stderr
# +build/nek-bp1: line 40: cd: ../../../Nek5000: No such file or directory
# +build/nek-bp1: line 41: cd: /home/jeremy/Dev/libCEED/bin: No such file or directory
# +build/nek-bp1: line 297: build/bp1.cpuselfrefserial.log.1.b6: No such file or directory
# +grep: build/bp1.cpuselfrefserial.log*: No such file or directory
```

thilinarmtb:
So, in order for make prove-all to work, we need to pass NEK5K_DIR and CEED_DIR to tap.sh; 
Let me know what is the best way to do it rather than building and copying. This needs additional
commands in makefiles specific to nek, which I don't think is nice.

Makefile
```diff
@@ -407,16 +411,15 @@ tst : ;@$(MAKE) $(MFLAGS) V=$(V) test
...
 	$(PROVE) $(PROVE_OPTS) --exec 'tests/tap.sh' $(matched:$(OBJDIR)/%=%)
 # run prove target in parallel
 prv : ;@$(MAKE) $(MFLAGS) V=$(V) prove
 
-prepnektests:
-	(export CC FC && cd examples && make prepnektests)
-
+.ONESHELL :
 prove-all :
+	+export NEK5K_DIR='$(NEK5K_DIR)' CEED_DIR='$(PWD)'
```
jeremylt: @jedbrown @thilinarmtb how about this approach to pass `NEK5K_DIR` and `CEED_DIR` to the shell where we test

thilinarmtb: Looks good.

jedbrown: I think each of the lines run in different shells. Also, there are other targets to run nek tests. We can export those variables globally.

jeremylt: I added `.OHESHELL` to get all of the lines in the same shell.

jeremylt: And now bp1/bp3 is failing to build on Travis. Investigating.

jedbrown: That can have other consequences. I'd rather [export](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html) those variables specifically.

jeremylt: I thought that was what the highlighted line does, but the following line ignored the export without `.ONESHELL`.

jedbrown: The `+` says "the thing you're running behaves like make; it may use the jobserver, for example". The command is run in the shell. `export` is also a make keyword (see my link).

jeremylt:
Found the issue:
```
FATAL ERROR: Cannot find mpif77!
```
Travis doesn't have MPI right now. Currently, we are building and running for the test suite with gfortran instead of mpif77.

jeremylt:
Now `make prove-all` works but `make prove-all -j2` fails. I BP1 and BP3's make processes compete with each other and conflict with `-j2`.

jedbrown:
Those should be made separate.  There's no good reason for them to be bundled together.

jeremylt:
Part of the build process is common, and I think the common parts/files are conflicting when one jobserver tries to make BP1 and the other tries to make BP3.

Makefile
```diff
@@ -395,29 +402,27 @@ allexamples = $(examples) $(external_examples)
...
-prepnektests:
-	(export CC FC && cd examples && make prepnektests)
-
-prove-all :
-	+$(MAKE) prove realsearch=%
+prove-all-prep : lib
+	$(MAKE) -j1 examples
+prove-all : prove-all-prep
+	+$(MAKE) CC=$(CC) FC=$(FC) CEED_DIR=`pwd` NEK5K_DIR="$(abspath $(NEK5K_DIR))" \
+prove realsearch=%
```
jeremylt: How about this? It runs in parallel (I believe) but builds in serial to avoid the Nek conflicts?

examples/nek5000/nek-examples.sh
```diff
@@ -0,0 +1,329 @@
...
+###############################################################################
+# DONT'T TOUCH WHAT FOLLOWS !!!
+###############################################################################
+# Set defaults for the parameters
+: ${NEK5K_DIR:=`cd "../../../Nek5000"; pwd`}
+: ${NEK5K_TOOLS_DIR:=`cd "${NEK5K_DIR}/bin"; pwd`}
+: ${CEED_DIR:=`cd "../../"; pwd`}
+: ${FC:="mpif77"}
+: ${CC:="mpicc"}
+: ${MPI:=0}
```
jeremylt: Set default to serial - Travis doesn't need to build w/mpifort

examples/nek5000/nek-examples.sh
```diff
@@ -0,0 +1,329 @@
...
+  if [ ${nek_verbose} = "true" ]; then
+    echo "Cleaning ..."
+  fi
+
+  if [ -f ./makenek ]; then
+    printf "y\n" | NEK_SOURCE_ROOT=${NEK5K_DIR} ./makenek clean 2>&1 >> /dev/null
+  fi
+  rm makenek* SESSION.NAME 2> /dev/null
+  for i in `seq 1 6`; do
+    rm -f bp$i bp$i.f bp$i*log* SIZE 2> /dev/null
```
jeremylt: Added `SIZE` to clean - above we say it is removed on cleaning

Makefile
```diff
@@ -124,25 +125,25 @@ tests.f   := $(sort $(wildcard tests/t[0-9][0-9][0-9]-*.f90))
 tests     := $(tests.c:tests/%.c=$(OBJDIR)/%)
 ctests    := $(tests)
 tests     += $(tests.f:tests/%.f90=$(OBJDIR)/%)
-#examples
+# Examples
```
jeremylt: Minor makefile style consistency updates

Makefile
```diff
@@ -465,7 +470,7 @@ install : $(libceed) $(OBJDIR)/ceed.pc
 
 cln clean :
 	$(RM) -r $(OBJDIR) $(LIBDIR)
-	$(MAKE) -C examples clean
+	$(MAKE) -C examples clean NEK5K_DIR="$(abspath $(NEK5K_DIR))"
```
jeremylt: Clean was missing the Nek dir before I added this

examples/Makefile
```diff
@@ -24,9 +24,9 @@ CONFIG_MK ?= $(or $(wildcard $(MFEM_DIR)/config/config.mk),$(wildcard\
   $(MFEM_DIR)/share/mfem/config.mk),invalid_mfem_dir)
 
 NEK5K_DIR ?= $(abspath ../../Nek5000)
-NEK5K_FC = $(if $(findstring mpi,$(FC)),$(FC),mpif77)
-NEK5K_CC = $(if $(findstring mpi,$(CC)),$(FC),mpicc)
-NEK5K_EXAMPLES ?= bp1
+NEK5K_FC:= $(if $(FC),$(FC),mpif77)
+NEK5K_CC:= $(if $(CC),$(CC),mpicc)
```
jeremylt: I set this to use FC/CC unless we explicitly want to use mpifort/mpicc

jeremylt:
@jedbrown, @thilinarmtb I think everything works now. Overall thoughts?

Thanks @thilinarmtb for overhauling the unwieldy number of scripts we had before.

Makefile
jedbrown: Hmm, only the Nek part needs to be serial. This seems to serialize more than necessary.

examples/nek5000/nek-examples.sh
jedbrown: This might surprise a user who wishes to experiment with Nek. I think I'd rather see Travis `export MPI=0` instead of hard-coding it here.

Makefile
jeremylt: I think my latest commit only does the Nek examples in serial
```
prove-all-prep :
	+$(MAKE) NEK5K_DIR= examples
	$(MAKE) -j1 examples
```

examples/nek5000/nek-examples.sh
jeremylt: Updated

examples/Makefile
jeremylt: Updated to use mpif77/mpicc unless `MPI=0`

thilinarmtb:
Looks good. Thanks for the help @jeremylt !

Makefile
jeremylt: @jedbrown, I think I fixed this. Are we ok to merge?

jedbrown:
`make search prove=nek`?
```
make: 'lib' with optional backends: /cpu/self/ref/memcheck /cpu/self/avx/serial /cpu/self/avx/blocked /cpu/self/xsmm/serial /cpu/self/xsmm/blocked /cpu/occa /gpu/occa /omp/occa /cpu/self/tmpl /cpu/self/tmpl/sub
make -C examples CC=gcc FC=gfortran MPI= CEED_DIR=`pwd` NEK5K_DIR="/home/jed/src/Nek5000" \
  NEK5K_EXAMPLES=bp1 nek
make -C examples CC=gcc FC=gfortran MPI= CEED_DIR=`pwd` NEK5K_DIR="/home/jed/src/Nek5000" \
  NEK5K_EXAMPLES=bp3 nek
make[1]: Entering directory '/home/jed/src/libceed/examples'
make[1]: Entering directory '/home/jed/src/libceed/examples'
cd nek5000 && NEK5K_DIR=/home/jed/src/Nek5000 CEED_DIR=/home/jed/src/libceed \
         FC=gfortran CC=gcc MPI= \
         ./nek-examples.sh -m -e "bp1"
cd nek5000 && NEK5K_DIR=/home/jed/src/Nek5000 CEED_DIR=/home/jed/src/libceed \
         FC=gfortran CC=gcc MPI= \
         ./nek-examples.sh -m -e "bp3"
cp: cannot create regular file './makenek': File exists
Building example: bp1 ...
Building example: bp3 ...
  Building bp1 failed. See bp1.build.log for details.
make[1]: *** [Makefile:42: nek] Error 1
make[1]: Leaving directory '/home/jed/src/libceed/examples'
make: *** [Makefile:365: build/nek-bp1] Error 2
make: *** Waiting for unfinished jobs....
```

jeremylt:
I didn't know that was an intended way to run the examples. I'll try to force Nek to oblige.

jeremylt:
I think I fix it now.

jedbrown:
Sorry about the round-about. We also need the likes of `make test search=nek` and `make junit search=nek`. Why does the script need to build both instead of building each Nek BP individually? Would it help to put them in separate subdirectories so the `makenek`s don't collide?

The Rule of Make is that each command should create exactly one output (that matters to the build; `*.d` files are okay because they only matter the next time that target is being built). When we have one command that creates two outputs, each of which we'd like to depend on separately, we get stuck with these leaky workarounds.

jeremylt:
I think independent subdirectories that duplicate the contents of the current Nek directory would be needed to avoid needing to build the Nek examples sequentially. And I'm not 100% sure that would work because both directories would be reaching back to the Nek directory and running code from there.

There are two parts to getting the Nek examples to build and run for prove, junit, and test. They need to be built separately (or possibly isolated from each other) and they need the CEED_DIR and NEK5K_DIR environment variables to be exported.

jeremylt:
So....

I got `make prove/junit/test` to work, with/without `search=*`.

I renamed the `examples/nek5000` dir because 
`__FILE__` causes issues in Travis, as
`     $  /home/travis/build/CEED/libCEED/examples/nek5000/build/bp1/bp1.f`
is too long at 73 characters.

As I feared, Nek on Travis appears to be having some issue with two processes trying to build the third party dependencies simultaneously (https://travis-ci.org/CEED/libCEED/jobs/561607000 line 1140-1156). I can't seem to replicate this locally, so debugging is slow.

jeremylt:
I have chased the issue into the bowels of Nek. 
Our Makefile calls `nek-examples.sh`
That copies and calls `makenek` (Bash script)
That calls `Nek5000/bin/nekconfig` (another Bash script)
That calls `Nek5000/core/makenek.inc` (another Bash script)
That has a function `build_3rdparty()`
This function seems to choke when our two different Make instances call `Nek5000/3rd_party/gslib/install`
This script forces the removal and recreation of the folder gslib (inside the folder `Nek5000/3rd_party/gslib`), and I believe this creates a race condition. 
As far as I can tell, because we are 5 scripts deep, both Make instances are unaware that they are both trying to remove, create, and then unpack a tarball into the same folder so they can both run Make in said folder.

Unfortunately, going into `Nek5000/3rd_party/gslib` and running `install` ourselves before we run `make prove` does not seem to help us at all because `makenek` still wants to rebuild the entire set of 3rd party dependencies on our first run of `make prove`.
However, if we run `make prove-all -j2` a second time after a fresh clone of Nek, then we don't trigger the rebuild of the 3rd party dependencies. This is why I was having trouble reproducing locally at first.

To reproduce:
1) Run `make clean` in libCEED
2) Remove your local Nek5k
3) Clone Nek5k
4) Run `make prove-all -j2` or `make search=nek prove/junit/test` in libCEED

Running one of the tests in `Nek5000/short_tests` can prevent the rebuild, but then there seems to be something wrong with in this situation, as `make prove` in libCEED then errors with messages such as
```
# +At line 23 of file /home/jeremy/Dev/Nek5000/core/reader_rea.f (unit = 9, file = '/home/jeremy/Dev/libCEED/build/boxes/b4/b4.rea')
# +Fortran runtime error: End of file
# +
# +Error termination. Backtrace:
# +#0  0x7f6d8587c08f in finalize_transfer
# +at /build/gcc/src/gcc/libgfortran/io/transfer.c:4109
# +#1  0x56357d35dc4a in rdparam_
# +at /home/jeremy/Dev/Nek5000/core/reader_rea.f:23
# +#2  0x56357d2b06e6 in readat_
# +at /home/jeremy/Dev/Nek5000/core/connect2.f:39
# +#3  0x56357d28e698 in nek_init_
# +at /home/jeremy/Dev/Nek5000/core/drive1.f:74
# +#4  0x56357d28876b in nekton
# +at /home/jeremy/Dev/Nek5000/core/drive.f:7
# +#5  0x56357d28876b in main
# +at /home/jeremy/Dev/Nek5000/core/drive.f:11
ok 4 nek-bp1 /cpu/self/ref/blocked
ok 5 nek-bp1 /cpu/self/ref/blocked stdout
```

jeremylt:
I bypassed the issue by single sourcing the Nek examples. Now one file runs both of the Nek BPs. The two files were 2 lines different anyways, so it makes more sense like this. Now there is only one file to build, so we aren't fighting the Nek build process's hardcoded serial build requirement.

What works now, in parallel and serial:
`make prove-all`
`make prove search=nek`
`make test search=nek`
`make junit search=nek`
`make examples`
`make nek` in the `examples` directory
`./nek-examples -m` in the `examples/nek` directory

jeremylt:
One minor issue. I ran the tests above with `MPI=0`. When `MPI=1` or `MPI` is unset (same thing), then `make examples` or `make nekexamples` at the top level has an issue where Nek can't seem to find `mpif.h`. This does not occur when `MPI=0` or `make nek` is called in the `examples` folder.

jeremylt:
Fixed it. Because Nek is running on scripts, it handles environment variables strangely and sometimes ignores our Nek master script, `nek-examples.sh`, or the makefile that calls it, `examples/Makefile`, changing the environment variables `FC` and `CC` to `mpif77` and `mpicc` instead of the top level Makefile default of `gfortran` and `gcc`. This leads to a mix of building with `gfortran` and `mpif77` which, naturally, conflict with each other.

Makefile
```diff
@@ -360,6 +372,12 @@ $(OBJDIR)/mfem-% : examples/mfem/%.cpp $(libceed) | $$(@D)/.DIR
 	  MFEM_DIR="$(abspath $(MFEM_DIR))" $*
 	mv examples/mfem/$* $@
 
+$(OBJDIR)/nek-% : examples/nek/%.usr examples/nek/nek-examples.sh $(libceed) | $$(@D)/.DIR
```
jedbrown: If any two Nek examples conflict with each other, should this just say `nek-bps` instead of the wildcard `nek-%` (to avoid a future developer accidentally recreating the present problem)?

jeremylt: Updated

jeremylt:
1. Moved. I set up a Makefile in the Nek folder that seems to be handling the complexity fine.
2. Documented
3. I think we can and should squash

PR #243, opened 2019-05-06
Add PETSc BP2, BP4, BP5, BP6

jeremylt:
This PR refactors the PETSc BP1 and BP3 while adding PETSc BP2, BP4, BP5, and BP6.


jeremylt:
Travis is failing because of Nek BP1 failing to build, which is unrelated to this PR. Investigating.

jeremylt:
`make benchmark` could be improved, but the benchmark scripts have been updated.

examples/navier-stokes/README.md
```diff
@@ -92,7 +92,9 @@ Energy Density:
 ## Density Current
 
 This problem solves the full compressible Navier-Stokes equations, using
-operator composition and design of coupled solvers in the context of atmospheric modeling.
+operator composition and design of coupled solvers in the context of atmospheric modeling. This problem was uses the formulation given in Semi-Implicit Formulations
```
valeriabarra: Thanks for adding the reference paper. Remove "was".

jeremylt: Good catch. Fixed.

Makefile
```diff
@@ -381,11 +381,11 @@ all: $(alltests)
 examples : $(allexamples)
 
 # Benchmarks
-allbenchmarks = petsc-bp1 petsc-bp3
+allbenchmarks = petsc-bps
 bench_targets = $(addprefix bench-,$(allbenchmarks))
 .PHONY: $(bench_targets) benchmarks
 $(bench_targets): bench-%: $(OBJDIR)/%
-	cd benchmarks && ./benchmark.sh --ceed "$(BACKENDS)" -r $(*).sh
+	cd benchmarks && ./benchmark.sh --ceed "$(BACKENDS)" -r petsc-bp3.sh
```
jeremylt: This should be fixed to run some subset of (petsc-bp1 petsc-bp2 petsc-bp3 petsc-bp4 petsc-bp5 petsc-bp6) as the default for `make benchmark`. The trick is, petsc-bps needs to be made and the petsc-bp1 petsc-bp2 petsc-bp3 petsc-bp4 petsc-bp5 petsc-bp6 scripts need to be run in the benchmark folder.

benchmarks/petsc-bps.sh
```diff
@@ -0,0 +1,53 @@
...
+#
+# The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+# a collaborative effort of two U.S. Department of Energy organizations (Office
+# of Science and the National Nuclear Security Administration) responsible for
+# the planning and preparation of a capable exascale ecosystem, including
+# software, applications, hardware, advanced system engineering and early
+# testbed platforms, in support of the nation's exascale computing imperative.
+
+
+function run_petsc_tests()
```
jedbrown: Why this pattern of sourcing and then the function? Also note that you can pass arguments
```
.: . filename [arguments]
    Execute commands from a file in the current shell.
```

examples/petsc/Makefile
```diff
@@ -11,8 +11,8 @@ LDFLAGS = $(call pkgconf, --libs-only-L --libs-only-other $(PETSc.pc) $(ceed.pc)
 LDFLAGS += $(patsubst -L%, $(call pkgconf, --variable=ldflag_rpath $(PETSc.pc))%, $(call pkgconf, --libs-only-L $(PETSc.pc) $(ceed.pc)))
 LDLIBS = $(call pkgconf, --libs-only-l $(PETSc.pc) $(ceed.pc)) -lm
 
-bp.c := $(sort $(wildcard bp*.c))
-bp := $(bp.c:%.c=%)
+bps.c := bps.c
+bp := $(bps.c:%.c=%)
```
jedbrown: Spelled as `./bps` in the README.

examples/petsc/bps.c
```diff
@@ -112,6 +116,48 @@ struct User_ {
   Ceed ceed;
 };
 
+typedef enum {
+  CEED_BP1, CEED_BP2, CEED_BP3, CEED_BP4, CEED_BP5, CEED_BP6
+} bpType;
+static const char *const bpOptions[] = {"bp1","bp2","bp3","bp4","bp5","bp6",
```
jedbrown: PETSc convention would be to call these `*Types` rather than `*Options`.

examples/petsc/bps.c
```diff
@@ -397,39 +457,94 @@ int main(int argc, char **argv) {
...
 
-  // Create the Q-function that builds the diff operator (i.e. computes its
+  // Create the Q-function that builds the operator (i.e. computes its
   // quadrature data) and set its context data.
-  CeedQFunctionCreateInterior(ceed, 1,
-                              Setup, __FILE__ ":Setup", &qf_setup);
+  switch (bpChoice) {
+    case CEED_BP1:
+      CeedQFunctionCreateInterior(ceed, 1,
+                                  SetupMass, __FILE__ ":SetupMass", &qf_setup);
```
jedbrown: These don't actually come from this file, but rather `bp1.h`, right?

examples/petsc/bps.c
```diff
@@ -397,39 +457,94 @@ int main(int argc, char **argv) {
...
-  CeedQFunctionAddOutput(qf_diff, "v", 1, CEED_EVAL_GRAD);
+  if (bpChoice == CEED_BP1 || bpChoice == CEED_BP2)
+    CeedQFunctionAddOutput(qf_setup, "rho", 1, CEED_EVAL_NONE);
+  else
+    CeedQFunctionAddOutput(qf_setup, "rho", 6, CEED_EVAL_NONE);
+  CeedQFunctionAddOutput(qf_setup, "true_soln", vscale, CEED_EVAL_NONE);
+  CeedQFunctionAddOutput(qf_setup, "rhs", vscale, CEED_EVAL_INTERP);
+
+  // Set up PDE operator
+  switch (bpChoice) {
```
jedbrown: Rather than the sequence of switch statements, it might be cleaner to have an array of structs or even just helper functions for each bp.

benchmarks/petsc-bps.sh
jeremylt: Honestly, I was repeating the style already there; I was trying not to change too much from what was already there. There are already 9 variables being passed in this fashion from `benchmark.sh`.

examples/petsc/bps.c
jeremylt: I'll think on how to rearrange this, but I agree that it can be better.

examples/petsc/bps.c
jeremylt: I created `__DIR__` and `PATH(headderfile.h)` to handle this. If this is reasonable, I'll transfer it to the other files, especially in #234.

examples/petsc/bps.c
jeremylt: Fixed

examples/petsc/Makefile
jeremylt: Fixed

examples/petsc/bps.c
jeremylt: Actually, it seems that OCCA likes what I did less. Investigating.

jeremylt: Fixed. There are a few instances of this already in the test suite, but we've gotten lucky because `foo.c` and `foo.h` shared the same filename.

examples/petsc/bps.c
jeremylt: Much cleaner now

jeremylt:
I'm basing my multigrid example on the improved BP3/BP4, so this should be merged before that example.

benchmarks/petsc-bps.sh
jeremylt: I dropped this and am using a single script for the PETSc BPs. I think this addresses your concern.

Makefile
jeremylt: Fixed. Default is `make benchmarks` runs BP1 and BP3 but the script can be called in the benchmarks folder to run the other BPs.

Makefile
```diff
@@ -392,7 +392,7 @@ all: $(alltests)
 examples : $(allexamples)
 
 # Benchmarks
-allbenchmarks = petsc-bp1 petsc-bp3
+allbenchmarks = petsc-bps
 bench_targets = $(addprefix bench-,$(allbenchmarks))
 .PHONY: $(bench_targets) benchmarks
 $(bench_targets): bench-%: $(OBJDIR)/%
```
jeremylt: One downside to `make benchmarks` (exists currently as well) is that it runs against *all* backends, to include tmpl and memcheck. I don't think this is a huge issue, as you can run the benchmarks script at better granularity directly.

jeremylt: Note, this is partially changed with `libceed_test.so`, so only the memcheck backend is run in the bench environment when it doesn't really make sense.

examples/petsc/bps.c
```diff
@@ -497,17 +601,22 @@ int main(int argc, char **argv) {
...
   CeedVectorDestroy(&rhsceed);
 
   ierr = KSPCreate(comm, &ksp); CHKERRQ(ierr);
   {
     PC pc;
     ierr = KSPGetPC(ksp, &pc); CHKERRQ(ierr);
-    ierr = PCSetType(pc, PCNONE); CHKERRQ(ierr);
+    if (bpChoice == CEED_BP1 || bpChoice == CEED_BP2) {
+      ierr = PCSetType(pc, PCJACOBI); CHKERRQ(ierr);
+      ierr = PCJacobiSetType(pc, PC_JACOBI_ROWSUM); CHKERRQ(ierr);
```
jedbrown: This could become a run-time option. Nothing in [the spec](https://ceed.exascaleproject.org/bps/) says that Jacobi preconditioning should be used (though it speeds up convergence for these cases.

examples/petsc/bps.c
```diff
@@ -497,17 +601,22 @@ int main(int argc, char **argv) {
...
+    } else {
+      ierr = PCSetType(pc, PCNONE); CHKERRQ(ierr);
+    }
     ierr = KSPSetType(ksp, KSPCG); CHKERRQ(ierr);
```
jedbrown: In this PR or a later one, we should probably use `KSP_NORM_NATURAL` because it's (a) actually monotone and (b) has one fewer reduction per iteration.

jeremylt:
We should be all set here. I'll squash+merge once Travis wraps up.

PR #244, opened 2019-05-06
Fix: Replace a few CeedChk with CeedChk_Cuda

YohannDudouit:
There was a few `CeedChk` being used instead of `CeedChk_Cu` in the `cuda` and `cuda-reg` backends.


jeremylt:
It looks like this PR does as described. Just for my curiosity, what does `CeedChk_Cu` do that `CeedChk` doesn't?

backends/cuda/ceed-cuda.c
```diff
@@ -23,6 +23,7 @@
 int compile(Ceed ceed, const char *source, CUmodule *module,
             const CeedInt numopts, ...) {
   int ierr;
+  cudaFree(0);//Make sure a Context exists for nvrtc
```
jeremylt: Is the usual/recommended fix?

YohannDudouit: Surprisingly it seems that it is the usual fix: https://stackoverflow.com/questions/10415204/how-to-create-a-cuda-context

jeremylt: Huh. CUDA can be weird. Thanks for the info!

YohannDudouit:
I think `CeedChk_Cu` gets the error message from the Cuda API before sending a `CeedError`
```
#define CeedChk_Cu(ceed, x) \
do { \
  CUresult result = x; \
  if (result != CUDA_SUCCESS) { \
    const char *msg; \
    cuGetErrorName(result, &msg); \
    return CeedError((ceed), result, msg); \
  } \
} while (0)
```

PR #246, opened 2019-05-09
Add PetscObjectSetName to set solution's vec name

valeriabarra:
This very small PR adds the object name to the state vector in our NS example so that it can be displayed in the output in vts/vtr formats, after the latest Petsc patch.


jeremylt:
Looks good to me.

PR #247, opened 2019-05-10
Cuda backend using shared memory

YohannDudouit:
This backend proposes a Cuda implementation using shared memory.

PR #248, opened 2019-05-13
Simplices handling for Cuda backends

YohannDudouit:
This PR adds a simple algorithm to treat simplices on for Cuda backends.


backends/cuda/ceed-cuda-basis.c
```diff
@@ -415,6 +500,67 @@ int CeedBasisApply_Cuda(CeedBasis basis, const CeedInt nelem,
   return 0;
 }
 
+int CeedBasisApplyNonTensor_Cuda(CeedBasis basis, const CeedInt nelem,
```
jeremylt: It looks like this will need the adjustment to remove "ceed-impl.h" as well.

jeremylt:
Am I right that this follows #247?

backends/cuda/ceed-cuda-basis.c
YohannDudouit: That's right, but most of these are already in cuda-shared branch, I'll rebase once the shared-memory backend gets merged into master.

YohannDudouit:
It doesn't completely follows #247 , but was implemented afterwards, I initially thought to put the non-tensor kernels in cuda-shared backend but then I realized that the implementation was relatively trivial and unoptimized and thus decided to put it in the cuda-ref backend. I could remove all the work from the cuda-shared backend to clean this PR!

YohannDudouit:
Refactored in #249

PR #251, opened 2019-05-18
Add all backends to Doxygen

jeremylt:
This is a small PR to fix the fact that Doxygen currently misses half of our backends.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -800,9 +800,16 @@ INPUT                  = . \
                          examples/petsc \
                          examples/navier-stokes \
                          backends/ref \
-                         backends/occa \
                          backends/blocked \
                          backends/template \
+                         backends/memcheck \
+                         backends/opt \
+                         backends/avx \
+                         backends/xsmm \
+                         backends/cuda \
+                         backends/cuda-reg \
+                         backends/cuda-shared \
+                         backends/occa \
                          backends/magma \
                          tests \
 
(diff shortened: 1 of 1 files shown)


jeremylt:
I was thinking about that. I organized them by 'backend family', i.e. the CPU backends are grouped, the CUDA backends are grouped, etc. I'm fine with whatever organizational schema makes the most sense to the most people.

valeriabarra:
Since in this list the family label (`/cpu/` or `/gpu/`) is not present, I'd keep them in alphabetic order for easy future insertions/deletions.

jeremylt:
Fair enough; changed

valeriabarra:
Would this recursive alphabetization make it extremely pedantic? Or just tidier?
```
                         . \
                         backends/avx \
                         backends/blocked \
                         backends/cuda \
                         backends/cuda-reg \
                         backends/cuda-shared \
                         backends/magma \
                         backends/memcheck \
                         backends/occa \
                         backends/opt \
                         backends/ref \
                         backends/template \
                         backends/xsmm \
                         doc \
                         include \
                         interface \
                         examples/ceed \
                         examples/mfem \
                         examples/nek5000 \
                         examples/petsc \
                         examples/navier-stokes \
                         tests \
```

jeremylt:
I left it alone because the original ordering is roughly ordered by line length, but I have no opinion either way.

Issue #253, opened 2019-05-22
Makefile Valgrind Check Error


jeremylt:
On Cori and Theta we get the error message:

```
fatal error: valgrind/memcheck.h: No such file or directory
 #include <valgrind/memcheck.h>
          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

We need to rework the Valgrind check in the Makefile. The headers are only included in a 'development' or 'debugging' install of Valgrind, but the check currently looks for any version of Valgrind.

jedbrown:
Resolved by #258.

PR #258, opened 2019-05-24
Check for devel Valgrind

jeremylt:
This PR improves the check for Valgrind. See Issue #253.

I'm not sure if there is a cleaner way to check for the Valgrind headers in Make.


Makefile
```diff
@@ -241,7 +242,10 @@ libceed.c += $(blocked.c)
 libceed.c += $(opt.c)
 
 # Memcheck Backend
-ifneq ($(shell which valgrind 2> /dev/null),)
+MEMCHK_STATUS = Disabled
+MEMCHK := $(shell if echo "\#include <valgrind/memcheck.h>" | $(CC) -x c -c - >/dev/null 2>&1; then echo 1; fi;)
```
jedbrown: This leaves behind a file `-.o`, but it should be sufficient to `$(CC) -E -`, which does not. Pushed this update.

PR #261, opened 2019-05-29
Update top level .gitignore

valeriabarra:
The `.gitignore` had some old references to test files from when they used to be in the main folder, instead of the `build` subfolder, like now


jedbrown:
@jeremylt Object files don't go in the `tests/` now. The patterns predate putting that stuff in `build/`. (We now ignore the whole `build/` directory.)

PR #262, opened 2019-05-29
OCCA Backend clang-tidy fixes

jeremylt:
This PR is a series of small changes to the OCCA backend for `clang-tidy` issues.

### DIFF
--- a/backends/occa/ceed-occa-basis.c
+++ b/backends/occa/ceed-occa-basis.c
@@ -107,24 +107,26 @@ static int CeedBasisBuildKernel(CeedBasis basis) {
 // * CeedScalars are used here, not CeedVectors: we don't touch it yet
 // *****************************************************************************
 static int CeedTensorContract_Occa(CeedInt A, CeedInt B, CeedInt C, CeedInt J,
-                                   const CeedScalar *t, CeedTransposeMode tmode,
-                                   const CeedInt Add,
-                                   const CeedScalar *u, CeedScalar *v) {
-  const CeedInt transpose = tmode == CEED_TRANSPOSE;
-  const CeedInt tstride0 = transpose?1:B;
-  const CeedInt tstride1 = transpose?J:1;
-  for (CeedInt a=0; a<A; a++) {
-    for (CeedInt j=0; j<J; j++) {
-      if (!Add)
-        for (CeedInt c=0; c<C; c++)
-          v[(a*J+j)*C+c] = 0.0;
-      for (CeedInt b=0; b<B; b++) {
-        for (CeedInt c=0; c<C; c++) {
-          v[(a*J+j)*C+c] += t[j*tstride0 + b*tstride1] * u[(a*B+b)*C+c];
-        }
-      }
-    }
+                                   const CeedScalar *restrict t,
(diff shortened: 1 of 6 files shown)


jeremylt:
Strangely, two warnings remain even though I believe them to be incorrect.
```
/home/travis/build/CEED/libCEED/backends/occa/ceed-occa-basis.c:127:26: warning: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage [clang-analyzer-core.uninitialized.Assign]
          v[(a*J+j)*C+c] += tq * u[(a*B+b)*C+c];
```
```
/home/travis/build/CEED/libCEED/backends/occa/ceed-occa-basis.c:127:32: warning: The right operand of '*' is a garbage value [clang-analyzer-core.UndefinedBinaryOperatorResult]
          v[(a*J+j)*C+c] += tq * u[(a*B+b)*C+c];
```

backends/occa/ceed-occa-basis.c
```diff
@@ -125,6 +125,7 @@ static int CeedTensorContract_Occa(CeedInt A, CeedInt B, CeedInt C, CeedInt J,
       for (CeedInt j=0; j<J; j++) {
         CeedScalar tq = t[j*tstride0 + b*tstride1];
         for (CeedInt c=0; c<C; c++)
+          // NOLINTNEXTLINE: False positive for unitalized values
```
jeremylt: I added `NOLINT` for the offending line. This is an uncommon false positive.

jeremylt:
I turned on the `make tidy` check for Travis's Linux builds. We could also get it running on OSX with some work, but that seemed superfluous.

backends/occa/ceed-occa-basis.c
jedbrown: Yeah, it thinks `A*J*C` could be non-positive, but asserting that that is not the case wasn't fixing it for me either. The NOLINT is fine by me.

backends/occa/ceed-occa-okl.c
```diff
@@ -32,7 +32,7 @@ int CeedOklPath_Occa(const Ceed ceed, const char *c_src_file,
   if (!last_dot)
     return CeedError(ceed, 1, "Cannot find file's extension!");
   const size_t okl_path_len = last_dot - okl;
-  strcpy(&okl[okl_path_len],".okl");
+  strncpy(&okl[okl_path_len],".okl",5);
```
jedbrown: As a technicality, it is possible to construct paths for which this causes memory corruption

jeremylt: Yeah, I'm not super happy it. I put in a note for this to be updated in the OCCA overhaul.

jeremylt:
> Looks good. We could also consider enabling non-default checks (`readability-*`, etc.; use `clang-tidy -list-checks -checks=\*` for a full list).

I looked at the number of warnings generated by the readability checks. It looks like a full additional PR of work for the Fortran interface alone.

jedbrown:
Yup, I'm not surprised. We aren't beholden to the clang-tidy overlords, but some of those checks would likely help find bugs early.

PR #263, opened 2019-05-29
Update postprocessing scripts

valeriabarra:
This very small PR updates the post-processing scripts after that the command line option "qextra" was removed in PR #243

### DIFF
--- a/benchmarks/petsc-bps.sh
+++ b/benchmarks/petsc-bps.sh
@@ -51,3 +51,5 @@ function run_tests()
       done
    done
 }
+
+test_required_examples="petsc-bps"
--- a/benchmarks/postprocess-base.py
+++ b/benchmarks/postprocess-base.py
@@ -107,16 +107,14 @@ while True:
       parts=line.split()
       i=0
       while i < len(parts):
-         if parts[i]=='-qextra':
-            i=i+1
-            data['quadrature-pts']=int(parts[i])
-         elif parts[i]=='-degree':
+         if parts[i]=='-degree':
             i=i+1
             data['order']=int(parts[i])
          elif parts[i]=='-ceed':
             i=i+1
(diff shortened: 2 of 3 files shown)


jeremylt:
You can still specify qextra as a runtime option, but it is filled in automatically if you specify the BP unless you a set it. BP 5/6 have only qextra=1. Setting it in the BP script is redundant.

jedbrown:
Maybe I'm missing something, but can we ensure that it appears somewhere in the output and that the postprocessing script parses it from the output rather than assuming that it has not been changed by the user?

Regarding plotting/postprocessing, we should work toward a solution where we only parse output from the various files and return a pandas dataframe with one row per observation. That can be written as CSV or consumed by a single plotter that will work for all the benchmarks and implementations (MFEM, Nek, etc.). That may be an issue for https://github.com/ceed/benchmarks, however.

jeremylt:
We could add a way to add extra flags through the benchmark script, which would then be picked up by the plotting script and added to the plots if needed. Always adding qextra makes messy plots, IMHO.

It would be lovely to rework the plotting scripts. It looks like in the benchmark repo they slightly customized the scripts to each package, so updating all of the slightly different scripts will take some work. Perhaps we should make a solution here and then take it to the benchmarks repo.

jedbrown:
I guess what I mean is that the output when using `-benchmark` should include the full specification, such that we don't need to know the input arguments or maintain consistency in defaults. I.e., there should be an explicit statement containing P and Q.

Then, the plotting script can drop mention of O when it uses the standard benchmark spec, but we won't get misleading plots if we were to run a nonstandard configuration.

jeremylt:
The BP name is an option that `petsc-bps.c` keeps track of. We could output it as part of our revamp of the scripts: #263

Issue #265, opened 2019-05-31
Update Benchmark Plotting Scripts


jeremylt:
The postprocessing scripts should be updated to use a Pandas dataframe and the BP file should be updated to output more complete documentation of the parameters.

Eventually, it would be good to update the benchmarks repo with our work on improving the postprocessing scripts.

valeriabarra:
Yeah there are also some other things that I didn't get to improve in PR #263.
1) there is another occurrence of 'q= p + %i+ ' on line 146 for the legend in the file `postprocess-plot.py`. That case didn't occur in my testing. 
2) now there is the variable `test_required_examples='petdc-bps' ` assigned in the file `petsc-bps.sh` that eventually creates the title of the plots with "petsc-bps", regardless of which one it is. I wasn't sure now that we have one single script for all petsc-bps how to handle this situation for this variable (perhaps with a list?)

PR #266, opened 2019-05-31
Improve petsc-bps benchmark output

jeremylt:
This PR starts to improve the `petsc-bps` output. See #263. Please feel free to add more commits to this branch.

Sample output:
```
[jeremy@pheonix libCEED]$ ./build/petsc-bps -problem BP6 -ceed /cpu/self/avx/blocked -benchmark -degree 7 -qextra 3

CEED Benchmark Problem 6:
  backend        : /cpu/self/avx/blocked
  p              : 8
  q              : 11
  Global dofs    : 960
  Process decomp : 1 1 1
  Local elements : 2 = 1 1 2
  Owned dofs     : 960 = 8 8 15
  KSP cg DIVERGED_ITS iterations 20 rnorm 6.241902e-04
  CG solve time  : 0.0377591 (0.0377591) sec.
  DOFs/sec in CG : 1.52546 (1.52546) million.
  Pointwise error (max) 4.921658e-04
```

### DIFF
--- a/benchmarks/postprocess-plot.py
+++ b/benchmarks/postprocess-plot.py
@@ -28,6 +28,7 @@ show_figures=1        # display the figures on the screen?
 
 
 #####   Load the data
+import pandas as pd
 exec(compile(open('postprocess-base.py').read(), 'postprocess-base.py', 'exec'))
 
 
@@ -41,17 +42,14 @@ rcParams['font.sans-serif'].insert(0,'Noto Sans')
 rcParams['font.sans-serif'].insert(1,'Open Sans')
 rcParams['figure.figsize']=[10, 8] # default: 8 x 6
 
-cm=get_cmap('Set1') # 'Accent', 'Dark2', 'Set1', 'Set2', 'Set3'
-if '_segmentdata' in cm.__dict__:
-   cm_size=len(cm.__dict__['_segmentdata']['red'])
-elif 'colors' in cm.__dict__:
-   cm_size=len(cm.__dict__['colors'])
-colors=[cm(1.*i/(cm_size-1)) for i in range(cm_size)]
+cm_size=16
+colors=['dimgrey','black','saddlebrown','firebrick','red','orange',
+        'gold','lightgreen','green','cyan','teal','blue','navy',
+        'purple','magenta','pink']
 
-##### Get runs
+##### Get test names
 sel_runs=runs
-tests=list(set([run['test'].rsplit('/',1)[-1].rsplit('.sh',1)[0]
-                for run in sel_runs]))
+tests=list(sel_runs.test.unique())
 test=tests[0]
 
 ##### Run information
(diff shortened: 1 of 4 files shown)


examples/petsc/bps.c
```diff
@@ -408,15 +410,24 @@ int main(int argc, char **argv) {
   ierr = VecSetUp(X); CHKERRQ(ierr);
 
   if (!test_mode) {
+      ierr = PetscPrintf(comm,
+                         "\nCEED Benchmark Problem %d:\n"
+                         "  backend        : %s\n"
+                         "  p              : %d\n"
+                         "  q              : %d\n",
```
jedbrown: Longer (unambiguous) names for these would be better.

Are there any parsing changes that should go into the benchmarks/ scripts?

jeremylt: I can expand the names.

There are parsing changes that would be required, but I wanted to get opinions on the sample output first.

jeremylt:
Updated with longer descriptions:

```
[jeremy@pheonix libCEED]$ ./build/petsc-bps -problem BP6 -ceed /cpu/self/avx/blocked -benchmark -degree 7 -qextra 3

-- CEED Benchmark Problem 6 -- libCEED + PETSc --
  libCEED:
    libCEED backend                    : /cpu/self/avx/blocked
  Mesh:
    Mesh order (p)                     : 8
    Number of 1D Quadrature points (q) : 11
    Global DOFs                        : 960
    Process decomposition              : 1 1 1
    Local elements                     : 2 = 1 1 2
    Owned dofs                         : 960 = 8 8 15
  KSP:
    KSP Convergence                    : DIVERGED_ITS
    Total KSP Iterations               : 20
    Final rnorm                        : 6.241902e-04
  Performance:
    CG Solve Time                      : 0.0307491 (0.0307491) sec
    DOFs/sec in CG                     : 1.87323 (1.87323) million
    Pointwise error (max)              : 4.921658e-04
```

jedbrown:
Better. "mesh order" doesn't distinguish polynomial degree (polynomial order) from L^2 order of convergence. How about "Number of 1D basis nodes" or "Number of 1D interpolation points"?

jeremylt:
Good point. Updated:

```
[jeremy@pheonix libCEED]$ ./build/petsc-bps -problem BP6 -ceed /cpu/self/avx/blocked -benchmark -degree 8 -qextra 3

-- CEED Benchmark Problem 6 -- libCEED + PETSc --
  libCEED:
    libCEED backend                    : /cpu/self/avx/blocked
  Mesh:
    Number of 1D Basis Nodes  (p)      : 9
    Number of 1D Quadrature Points (q) : 12
    Global DOFs                        : 729
    Process Decomposition              : 1 1 1
    Local Elements                     : 1 = 1 1 1
    Owned DOFs                         : 729 = 9 9 9
  KSP:
    KSP Convergence                    : DIVERGED_ITS
    Total KSP Iterations               : 20
    Final rnorm                        : 4.755065e-08
  Performance:
    CG Solve Time                      : 0.0417695 (0.0417695) sec
    DOFs/Sec in CG                     : 1.04718 (1.04718) million
    Pointwise Error (max)              : 2.330416e-03
```

jedbrown:
You're approaching valid YAML syntax. :-)

jeremylt:
Plotting now works with the new output - ready to start cleaning up the scripts.

jeremylt:
I think this helps with those two. Thanks for pointing the comment out.

jedbrown:
Script is broken with python3 and the file names include newlines, which is unusual.

jeremylt:
That's rude of it. I'll look at it Monday. I want to largely rebuild it into something cleaner.

valeriabarra:
@jeremylt I've taken care of it, don't worry about it.

valeriabarra:
@jed I've tested it now with both Python 2.7.15rc1 and  Python 3.6.7 and it works just fine

valeriabarra:
@jeremylt I I've taken care of the legend, too, that was part 1 of my [comment](https://github.com/CEED/libCEED/issues/265#issuecomment-497729971)

jeremylt:
I did some tidying and converting to pandas.

jeremylt:
I'm happy with these improvements. Do we need any changes?

valeriabarra:
> I'm happy with these improvements. Do we need any changes?

I personally had to shorten the title of my plots a lot, because having that whole string was too long

jeremylt:
How about this for a plot title:
`(1 node, 4 tasks), /cpu/self/ref/serial, PETSc BP1`

That's what I have in the latest commit

valeriabarra:
That looks fine for 1 node. But do we want to keep the 'tasks per node' for when we run on multiple nodes, or we're ok with total tasks?

jeremylt:
Total tasks seems ok for me, but I don't know what people expect to see more (@jedbrown?). It is a matter of simple division, so I don't think its a big deal and I think more compact is more better.

jedbrown:
I might use "ranks" instead of "tasks" since some programming models have tasks and threads. MPI ranks is a precise term.
Could limit horizontal space using `1 node \times 16 ranks \times 4 threads` (and elide the threads part when threads=1). I don't feel strongly about it.

jeremylt:
I changed tasks to ranks. Right now we don't have threads as separate information.

benchmarks/postprocess-plot.py
```diff
@@ -156,9 +156,9 @@
       plot(y/slope2,y,'k-',label='%g iter/s'%(slope2/vdim))
 
    # Plot information
-   title('(%i node%s, %i ranks), %s, %s'%(
+   title(r'%i node%s $\times$ %i ranks, %s, %s'%(
```
jeremylt: Right now we have total ranks rather than ranks per process, so times doesn't make sense here. Do we want total ranks or ranks per node? Nodes x Ranks Per Node = Total Ranks takes up a lot of space, which is why I had left it as (Nodes, Total Ranks).

jedbrown: I slightly prefer `nodes \times ranks_per_node` because mental multiplication is easier than division, resources like memory bandwidth are largely shared on each node, and architectures can have a pretty big range in number of cores and relative power of those cores.

jeremylt: Updated to ranks per node. I think we're good to merge this now.

valeriabarra:
Looks good, thanks.

PR #267, opened 2019-06-02
Add -include directive to Makefile 

valeriabarra:
Added -include directive to top-level Makefile to allow easy import of custom build configurations (harmless if file config.mk does not exist)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -500,4 +500,30 @@ print-% :
 	$(info )
 	@true
 
+# "make configure" will autodetect any variables not passed on the
+# command line, caching the result in config.mk to be used on any
+# subsequent invocations of make.  For example,
+#
+#   make configure CC=/path/to/my/cc CUDA_DIR=/opt/cuda
+#   make
+#   make prove
+configure :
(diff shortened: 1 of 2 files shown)


valeriabarra:
Thanks for adding that. It looks good. I also had to add `CUDA_LIB_DIR_STUBS` (please see my [comment](https://github.com/CEED/libCEED/pull/249#issuecomment-498050154) ). Would it be good to add it here as well?

jeremylt:
Ready to merge?

valeriabarra:
> Ready to merge?

Currently testing. Will merge when done.

valeriabarra:
@jedbrown I've added  `@: > config.mk` so that if one runs  `make configure` multiple times, it does not append to the same  `config.mk` file multiple times. Moreover, I've added the `-a` option to the tee command for CC, FC, and NVCC. If it looks ok to you, I can merge it.

jedbrown:
Yeah, all but the first line was meant to be `tee -a`. Your solution is fine too, and less error-prone if one were to reorder. I pushed one more minor commit, and it should be ready to merge.

PR #269, opened 2019-06-04
First set of optimizations for the cuda backends.

YohannDudouit:
This PR optimizes the cuda-reg restriction (also used by the cuda-shared backend), and the basis apply of the cuda-shared backend.

### DIFF
--- a/backends/cuda-reg/ceed-cuda-reg-restriction.c
+++ b/backends/cuda-reg/ceed-cuda-reg-restriction.c
@@ -23,7 +23,7 @@ static const char *restrictionkernels = QUOTE(
 extern "C" __global__ void noTrNoTr(const CeedInt nelem,
                                     const CeedInt *__restrict__ indices,
                                     const CeedScalar *__restrict__ u,
-CeedScalar *__restrict__ v) {
+                                    CeedScalar *__restrict__ v) {
   if (indices) {
     for(CeedInt dof = blockIdx.x * blockDim.x + threadIdx.x;
         dof < nelem*RESTRICTION_ELEMSIZE;
@@ -227,7 +227,8 @@ extern "C" __global__ void trTrIdentity(const CeedInt nelem,
     v [ RESTRICTION_NCOMP * (s + RESTRICTION_ELEMSIZE * e) + d ] = u[i];
   }
 }
-                                        );
+
+);
 
 static int CeedElemRestrictionApply_Cuda_reg(CeedElemRestriction r,
     CeedTransposeMode tmode, CeedTransposeMode lmode,
@@ -260,7 +261,7 @@ static int CeedElemRestrictionApply_Cuda_reg(CeedElemRestriction r,
     ierr = CeedElemRestrictionGetElementSize(r, &elemsize); CeedChk(ierr);
     void *args[] = {&nelem, &impl->d_ind, &d_u, &d_v};
     CeedInt blocksize = elemsize<1024?(elemsize>32?elemsize:32):1024;
-    ierr = run_kernel(ceed, kernel, CeedDivUpInt(ndof, blocksize), blocksize,
+    ierr = CeedRunKernelCuda(ceed, kernel, CeedDivUpInt(ndof, blocksize), blocksize,
                       args); CeedChk(ierr);
   } else {
     if (impl->d_ind) {
@@ -270,8 +271,8 @@ static int CeedElemRestrictionApply_Cuda_reg(CeedElemRestriction r,
         kernel = impl->trTr;
       }
       void *args[] = {&impl->d_tindices, &impl->d_toffsets, &d_u, &d_v};
-      ierr = run_kernel(ceed, kernel, CeedDivUpInt(ndof, blocksize), blocksize,
-                        args); CeedChk(ierr);
+      ierr = CeedRunKernelCuda(ceed, kernel, CeedDivUpInt(ndof, blocksize), blocksize,
+                               args); CeedChk(ierr);
     } else {
       if (lmode == CEED_NOTRANSPOSE) {
         kernel = impl->trNoTrIdentity;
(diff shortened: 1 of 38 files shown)


backends/cuda-reg/ceed-cuda-reg.c
```diff
@@ -20,6 +20,11 @@
 #include "ceed-cuda-reg.h"
 #include "../cuda/ceed-cuda.h"
 
+static int CeedGetPreferredMemType_Cuda_reg(CeedMemType *type) {
```
jeremylt: I fixed the delegation, so this shouldn't be needed anymore.

backends/cuda-shared/ceed-cuda-shared.c
```diff
@@ -22,6 +22,11 @@
 #include <cuda_runtime.h>
 #include "ceed-cuda-shared.h"
 
+static int CeedGetPreferredMemType_Cuda_shared(CeedMemType *type) {
```
jeremylt: See above

backends/cuda-shared/ceed-cuda-shared.c
```diff
@@ -40,8 +45,13 @@ static int CeedInit_Cuda_shared(const char *resource, Ceed ceed) {
 
   Ceed_Cuda_shared *data;
   ierr = CeedCalloc(1,&data); CeedChk(ierr);
+  data->Q1d  = -1;
```
jeremylt: These are at the `CEED` level?

backends/cuda-reg/ceed-cuda-reg-restriction.c
```diff
@@ -19,35 +19,33 @@
 #include "../cuda/ceed-cuda.h"
 
 static const char *restrictionkernels = QUOTE(
-    extern "C" __global__ void noTrNoTr(const CeedInt nelem,
-                                        const CeedInt *__restrict__ indices,
-                                        const CeedScalar *__restrict__ u,
+
```
jeremylt: Is this indenting change also in #249?

backends/cuda-shared/ceed-cuda-shared.c
YohannDudouit: I think this is where they make the most sense, since this tells which basis is currently loaded in the constant memory, however it's buggy and doesn't bring any performance gain, I'll remove it.

backends/cuda-reg/ceed-cuda-reg-restriction.c
YohannDudouit: I believe so.

backends/cuda-reg/ceed-cuda-reg.c
YohannDudouit: done.

backends/cuda-shared/ceed-cuda-shared.c
YohannDudouit: done.

jeremylt:
Ready to merge?

YohannDudouit:
@jeremylt we might want to merge #249 first because of the namespacing modifications, then I'll make sure that the modifications are also in this PR, and I'll be ready to merge.

jedbrown:
Do you get a zillion of these warnings?
```
./include/ceed.h:128:3: warning: operand of ?: changes signedness from ‘int’ to ‘CUresult’ {aka ‘enum cudaError_enum’} due to unsignedness of other operand [-Wsign-compare]
```

tcew:
I am curious about performance aspects.

It is not plausible to review GPU code by eye and claim it is acceptable. 

When I evaluate my own GPU code I work iteratively with the compiler, profiling tools, and roofline models to understand the performance characteristics of the code.

At some points I also reject some optimizations based on code support and logistic reasons.

If these codes exceed previous performance that is excellent. Is this the case ?

jedbrown:
I pushed a commit to resolve `-Wsign-compare`.

@tcew The sort of review that might be most valuable from you is in terms of strategy (is this expected to be excessively fragile?) and maintainability.

For assessing/tracking performance, I think #281 (and #219) will be extremely useful.

YohannDudouit:
@tcew The `cuda-shared` backend had the best performance of all GPU backends at the time of the PR. Now, the `cuda-gen` ( #275 ) achieves the best performance of all GPU backends.
To benchmark backends I use mfem 4.0 (for the GPU support) and a branch that should be merged soon that integrate libCEED into MFEM. I also use profiling tools to achieve the performance analysis.

tcew:
> @tcew The `cuda-shared` backend had the best performance of all GPU backends at the time of the PR. Now, the `cuda-gen` ( #275 ) achieves the best performance of all GPU backends.
> To benchmark backends I use mfem 4.0 (for the GPU support) and a branch that should be merged soon that integrate libCEED into MFEM. I also use profiling tools to achieve the performance analysis.

@YohannDudouit That is really cool!  It would be neat if you can attach some info like that to future review requests. Writing the code is really only a small part of the work package - and without more context it is hard to get at what the deltas imply. Finally, it would be helpful if future review requests were maybe a bit more specific - i.e. point to key piece(s) of code that might be priority for review. 

@jedbrown I think you are correct that keeping track of the performance of the code over time is going to be super important. I guess the output will need to capture a lot of state including test device, backend, problem size(s) etc. Figuring out what it all means is going to be tricky though.

When transitioning between the main GPU platforms different optimizations can give mixed performance results. I am finding balancing performance between CUDA and HIP execution optimality is tricky for instance. In some cases the performance requirements in CUDA tank HIP performance and vice versa.

jedbrown:
> I am finding balancing performance between CUDA and HIP execution optimality is tricky for instance. In some cases the performance requirements in CUDA tank HIP performance and vice versa.

This sort of non-portable performance is critical feedback for vendors and other developers to be aware of. Even anecdotes in which you've understood what's going on would be interesting to see. It's hard to build a mental model of architectural differences and performance consequences purely on vendor documentation and the success stories that appear in talks and papers.

tcew:
> > I am finding balancing performance between CUDA and HIP execution optimality is tricky for instance. In some cases the performance requirements in CUDA tank HIP performance and vice versa.
> 
> This sort of non-portable performance is critical feedback for vendors and other developers to be aware of. Even anecdotes in which you've understood what's going on would be interesting to see. It's hard to build a mental model of architectural differences and performance consequences purely on vendor documentation and the success stories that appear in talks and papers.

I am working directly with AMD. Our usage case is esoteric, but the feedback has been welcome.

The interplay between thread model, compiler, and hardware is very nonlinear in particular at the edge of register pressure.

jedbrown:
@YohannDudouit Can you please resolve the merge conflicts (a consequence of style changes in #249)?

PR #274, opened 2019-06-10
Fix Underinterpolation  in /cpu/self Backends

jeremylt:
This ~~WIP~~ PR fixes the `cpu/self/` backends when `P > Q`. This will use dim**2 contractions for grad, rather than 2*dim, so if this becomes important, we can optimize in the future.


jeremylt:
I'm having some weird issue in the 3D testing, so there is some small problem. Not sure where, as I copied the basis code from before collocated bases. XSMM works but the other CPU backends don't, so it's probably something minor and stupid.

jeremylt:
junit seems to have not run...

PR #276, opened 2019-06-17
t506, Single QFunction Multiple Operators

jeremylt:
This PR adds a test that shows a bug in the OCCA backend that prevents reuse of QFunctions across operators.

@YohannDudouit, how does this new test do with the CUDA backends?


jeremylt:
This bug will be addressed by @dmed256's planned OCCA backend overhaul. At that time we can merge this new test.

jeremylt:
Will be able to merge after #305

jeremylt:
Thanks. I'll merge after #305.

Issue #252, opened 2019-05-21
Add Identity QFunction


jeremylt:
Add an identity QFunction. This would especially be useful for restriction/interpolation operations for the multigrid example.

PR #279, opened 2019-06-18
Add Identity QFunctions

jeremylt:
This WIP PR adds `CEED_QFUNCTION_IDENTITY` to make restriction/interpolation operators more efficient. For this QFunction, each qfunction infield writes directly into each qfunction outfield.

i.e.
```
static int IdentQF(void *ctx, CeedInt Q,
  const CeedScalar *u = in[0];
  CeedScalar *v = out[0];
  for (CeedInt i=0; i<Q; i++) {
    v[i] = u[i];
  }
  return 0;
}
```
we replicate this without the extra copy for any number of input/output pairs.

TODO:
- ~~Update interface~~
- ~~Add test for CEED_QFUNCTION_IDENTITY~~
- Update OCCA backend
- Update CUDA backends (@YohannDudouit, I will run my ideas by you for this)

This resolves issue #252.

### DIFF
--- a/gallery/identity/ceed-identity.c
+++ b/gallery/identity/ceed-identity.c
@@ -1,0 +1,55 @@
+// Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory. LLNL-CODE-734707.
+// All Rights reserved. See files LICENSE and NOTICE for details.
+//
+// This file is part of CEED, a collection of benchmarks, miniapps, software
+// libraries and APIs for efficient high-order finite element and spectral
+// element discretizations for exascale applications. For more information and
+// source code availability see http://github.com/ceed.
+//
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+#include <string.h>
+#include "ceed-backend.h"
+#include "ceed-identity.h"
+
+/**
+ This file is not compiled into libCEED. This file provides a template to
+   build additional gallery QFunctions. Copy this file and the header to a
+   new folder in this directory and modify.
+**/
+
+/**
+  @brief Set fields for new Ceed QFunction
+**/
+static int CeedQFunctionInit_Identity(Ceed ceed, const char *requested,
+    CeedQFunction qf) {
+  int ierr;
+
+  // Check QFunction name
+  const char *name = "Identity";
+  if (strcmp(name, requested))
+    return CeedError(ceed, 1, "QFunction '%s' does not match requested name: %s",
+                     name, requested);
+
+  // Add QFunction fields
+  ierr = CeedQFunctionAddInput(qf, "input", 1, CEED_EVAL_INTERP); CeedChk(ierr);
+  ierr = CeedQFunctionAddOutput(qf, "output", 1, CEED_EVAL_INTERP);
+  CeedChk(ierr);
+
+  return 0;
+}
+
+/**
+  @brief Register new Ceed QFunction
+**/
+__attribute__((constructor))
+static void Register(void) {
+  CeedQFunctionRegister("Identity", Identity_loc, 1, Identity,
+                        CeedQFunctionInit_Identity);
+}
--- a/gallery/identity/ceed-identity.c
+++ b/gallery/identity/ceed-identity.c
@@ -19,13 +19,7 @@
 #include "ceed-identity.h"
 
 /**
- This file is not compiled into libCEED. This file provides a template to
-   build additional gallery QFunctions. Copy this file and the header to a
-   new folder in this directory and modify.
-**/
-
-/**
-  @brief Set fields for new Ceed QFunction
+  @brief Set fields identity QFunction that copies inputs directly into outputs
 **/
 static int CeedQFunctionInit_Identity(Ceed ceed, const char *requested,
     CeedQFunction qf) {
@@ -46,7 +40,7 @@ static int CeedQFunctionInit_Identity(Ceed ceed, const char *requested,
 }
 
 /**
-  @brief Register new Ceed QFunction
+  @brief Register identity QFunction that copies inputs directly into outputs
 **/
 __attribute__((constructor))
 static void Register(void) {
--- a/gallery/identity/ceed-identity.h
+++ b/gallery/identity/ceed-identity.h
@@ -27,7 +27,7 @@ CEED_QFUNCTION(Identity)(void *ctx, const CeedInt Q,
                          const CeedScalar *const *in,
                          CeedScalar *const *out) {
   // Ctx holds field size
-  const CeedInt size = *(CeedInt *)ctx;
+  const CeedInt size = ctx ? *(CeedInt *)ctx : 1;
 
   // in[0] is input, size (Q*size)
   const CeedScalar *input = in[0];
--- a/gallery/identity/ceed-identity.h
+++ b/gallery/identity/ceed-identity.h
@@ -15,13 +15,7 @@
 // testbed platforms, in support of the nation's exascale computing imperative.
 
 /**
- This file is not compiled into libCEED. This file provides a template to
-   build additional gallery QFunctions. Copy this file and the registeration/
-   initalization .c file to a new folder in this directory and modify.
-**/
-
-/**
-  @brief New Ceed QFunction
+  @brief  Identity QFunction that copies inputs directly into outputs
 **/
 CEED_QFUNCTION(Identity)(void *ctx, const CeedInt Q,
                          const CeedScalar *const *in,
--- a/gallery/identity/ceed-identity.h
+++ b/gallery/identity/ceed-identity.h
@@ -1,0 +1,44 @@
+// Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory. LLNL-CODE-734707.
+// All Rights reserved. See files LICENSE and NOTICE for details.
+//
+// This file is part of CEED, a collection of benchmarks, miniapps, software
+// libraries and APIs for efficient high-order finite element and spectral
+// element discretizations for exascale applications. For more information and
+// source code availability see http://github.com/ceed.
+//
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+/**
+ This file is not compiled into libCEED. This file provides a template to
+   build additional gallery QFunctions. Copy this file and the registeration/
+   initalization .c file to a new folder in this directory and modify.
+**/
+
+/**
+  @brief New Ceed QFunction
+**/
+CEED_QFUNCTION(Identity)(void *ctx, const CeedInt Q,
+                         const CeedScalar *const *in,
+                         CeedScalar *const *out) {
+  // Ctx holds field size
+  const CeedInt size = *(CeedInt *)ctx;
+
+  // in[0] is input, size (Q*size)
+  const CeedScalar *input = in[0];
+  // out[0] is output, size (Q*size)
+  CeedScalar *output = out[0];
+
+  // Quadrature point loop
+  CeedPragmaSIMD
+  for (CeedInt i=0; i<Q*size; i++) {
+    output[i] = input[i];
+  } // End of Quadrature Point Loop
+
+  return 0;
+}
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -25,7 +25,9 @@
 ///   @{
 
 /**
-  @brief Create an operator from element restriction, basis, and QFunction
+  @brief Create a CeedOperator and associate a CeedQFunction. A CeedBasis and
+           CeedElemRestriction can be associated with CeedQFunction fields with
+           \ref CeedOperatorSetField.
 
   @param ceed    A Ceed object where the CeedOperator will be created
   @param qf      QFunction defining the action of the operator at quadrature points
--- a/interface/ceed-qfunction.c
+++ b/interface/ceed-qfunction.c
@@ -166,7 +166,12 @@ int CeedQFunctionCreateInteriorByName(Ceed ceed,  const char *name,
 
 /**
   @brief Create an identity CeedQFunction. Inputs are written into outputs in
-           the order given.
+           the order given. This is useful for CeedOperators of the form
+           A = G^T B^T I B G. These operators can be represented with only the
+           action of a CeedRestriction and CeedBasis, such as restriction and
+           prolongation operators for p-multigrid. Backends may optimize
+           CeedOperators with this CeedQFunction to avoid the copy of input data
+           to output fields by using the same memory location for both.
 
   @param ceed       A Ceed object where the CeedQFunction will be created
   @param size       Size of the qfunction fields
@@ -622,7 +627,7 @@ int CeedQFunctionDestroy(CeedQFunction *qf) {
   ierr = CeedFree(&(*qf)->inputfields); CeedChk(ierr);
   ierr = CeedFree(&(*qf)->outputfields); CeedChk(ierr);
   // Free ctx if identity
-  if ((*qf)->identity && (*qf)->ctx) {
+  if ((*qf)->identity) {
     ierr = CeedFree(&(*qf)->ctx); CeedChk(ierr);
   }
 
--- a/interface/ceed-qfunction.c
+++ b/interface/ceed-qfunction.c
@@ -183,10 +183,12 @@ int CeedQFunctionCreateIdentity(Ceed ceed, CeedInt size, CeedQFunction *qf) {
   ierr = CeedQFunctionCreateInteriorByName(ceed, "Identity", qf); CeedChk(ierr);
 
   (*qf)->identity = 1;
-  CeedInt *ctx;
-  ierr = CeedCalloc(1, &ctx); CeedChk(ierr);
-  ctx[0] = size;
-  ierr = CeedQFunctionSetContext(*qf, ctx, sizeof(ctx)); CeedChk(ierr);
+  if (size > 1) {
+    CeedInt *ctx;
+    ierr = CeedCalloc(1, &ctx); CeedChk(ierr);
+    ctx[0] = size;
+    ierr = CeedQFunctionSetContext(*qf, ctx, sizeof(ctx)); CeedChk(ierr);
+  }
 
   return 0;
 }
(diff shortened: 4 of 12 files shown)


jeremylt:
Update: Identity QFunctions added and working on CPU backends. Need to add to CUDA backends.

jedbrown:
We've got these references to `CEED_QFUNCTION_IDENTITY` despite it just being a flag in the struct.
```
$ git grep _IDENTITY
backends/blocked/ceed-blocked-operator.c:                         "CEED_EVAL_NONE for a matching input and output does not make sense with CEED_QFUNCTION_IDENTITY");
backends/opt/ceed-opt-operator.c:                     "CEED_QFUNCTION_IDENTITY must have the same number of inputs and outputs");
backends/opt/ceed-opt-operator.c:                         "CEED_EVAL_NONE for a matching input and output does not make sense with CEED_QFUNCTION_IDENTITY");
backends/ref/ceed-ref-operator.c:                     "CEED_QFUNCTION_IDENTITY must have the same number of inputs and outputs");
backends/ref/ceed-ref-operator.c:                         "CEED_EVAL_NONE for a matching input and output does not make sense with CEED_QFUNCTION_IDENTITY");
```

jeremylt:
> We've got these references to `CEED_QFUNCTION_IDENTITY` despite it just being a flag in the struct.
> 
> ```
> $ git grep _IDENTITY
> backends/blocked/ceed-blocked-operator.c:                         "CEED_EVAL_NONE for a matching input and output does not make sense with CEED_QFUNCTION_IDENTITY");
> backends/opt/ceed-opt-operator.c:                     "CEED_QFUNCTION_IDENTITY must have the same number of inputs and outputs");
> backends/opt/ceed-opt-operator.c:                         "CEED_EVAL_NONE for a matching input and output does not make sense with CEED_QFUNCTION_IDENTITY");
> backends/ref/ceed-ref-operator.c:                     "CEED_QFUNCTION_IDENTITY must have the same number of inputs and outputs");
> backends/ref/ceed-ref-operator.c:                         "CEED_EVAL_NONE for a matching input and output does not make sense with CEED_QFUNCTION_IDENTITY");
> ```

Fixed it. Incomplete changeover from different approach.

jeremylt:
This PR works as is, so it is a question if we want to add support in other backends (CUDA and OCCA) now or a later PR. @jedbrown, @YohannDudouit thoughts?

backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -177,3 +177,10 @@ int CeedQFunctionCreate_Cuda(CeedQFunction qf) {
                                 CeedQFunctionDestroy_Cuda); CeedChk(ierr);
   return 0;
 }
+
+int CeedQFunctionCreateIdentity_Cuda(CeedQFunction qf) {
+  int ierr;
+  Ceed ceed;
+  ierr = CeedQFunctionGetCeed(qf, &ceed); CeedChk(ierr);
+  return CeedError(ceed, 1, "Backend does not implement Identity QFunctions");
```
jedbrown: Surely it can be implemented in a naive way (write a normal QFunction).

jeremylt: Copying over the vectors with `CEED_COPY_VALUES` is probably the easiest

jeremylt: updated

backends/cuda/ceed-cuda-operator.c
```diff
@@ -277,9 +287,22 @@ static int CeedOperatorApply_Cuda(CeedOperator op, CeedVector invec,
       CeedChk(ierr);
     }
   }
+
   // Q function
-  ierr = CeedQFunctionApply(qf, numelements * Q, impl->qvecsin, impl->qvecsout);
-  CeedChk(ierr);
+  if (identityQF) {
```
jeremylt: @YohannDudouit I added this as a naive implementation of the identity QFunction. This definitely can be improved upon.

backends/cuda/ceed-cuda-vec.c
```diff
@@ -61,9 +61,10 @@ static int CeedVectorSetArrayHost_Cuda(const CeedVector vec,
   case CEED_COPY_VALUES: ;
     CeedInt length;
     ierr = CeedVectorGetLength(vec, &length); CeedChk(ierr);
-    ierr = CeedMalloc(length, &data->h_array_allocated); CeedChk(ierr);
+    if (!data->h_array_allocated) {
```
jeremylt: @YohannDudouit I added this because I found in the OCCA backend that SetArray to set a memory location followed by SetArray to copy data was not copying data into the previously set memory location. Also added below for the device memory.

jeremylt:
Naive implementation added to CUDA, OCCA, and MAGMA, tested in CUDA only. We need testing in CUDA (MAGMA backend currently does not compile on master).

backends/cuda/ceed-cuda-vec.c
YohannDudouit: I'm not sure I completely agree with this, if you do `SetArray(vec, CEED_USE_POINTER, &array)` followed by `SetArray(vec, CEED_COPY_VALUES, &array)` it would replace `h_array` by a newly allocated vector, but if you do `SetArray(vec, CEED_OWN_POINTER, &array)` followed by `SetArray(vec, CEED_COPY_VALUES, &array)` then it copies into `h_array`, is it the expected behaviour?

YohannDudouit: Are you sure it's not:
```c++
if (!data->h_array)
```
?

backends/cuda/ceed-cuda-vec.c
```diff
@@ -88,10 +89,11 @@ static int CeedVectorSetArrayDevice_Cuda(const CeedVector vec,
 
   switch (cmode) {
   case CEED_COPY_VALUES:
-    ierr = cudaMalloc((void **)&data->d_array_allocated, bytes(vec));
-    CeedChk_Cu(ceed, ierr);
-    data->d_array = data->d_array_allocated;
-
+    if (!data->d_array_allocated) {
```
YohannDudouit: Same as previously, I think you want to check:
```c++
if (!data->d_array)
```

backends/cuda/ceed-cuda-operator.c
YohannDudouit: I'm not fan of polluting the qFunction code with this special case, especially because identity should really be handle at the operator level to be optimized (to avoid restriction/prolongation). Thus I think that having an identity qFunction should be enough for now:
```c++
__global__ void identity_apply(...) {
  const CeedScalar* u = fields.in[0];
  CeedScalar* v = fields.out[0];
  for(...){
   v[i] = u[i];
  }
}
```

backends/cuda/ceed-cuda-vec.c
jeremylt: Good point

jeremylt:
The identity needs to handle arbitrary numbers of input/output pairs. Can you make a .cu that does that?

YohannDudouit:
Do we really need a generic identity? Why the user wouldn't write the identity he needs according to the number of inputs/outputs?

jeremylt:
I'm concerned about copying the entire qvector for no reason. For my multigrid example, its an extra (degree-1)*2 copies of the qvector per CG iteration.

backends/cuda/ceed-cuda-operator.c
jeremylt: Note, this is an identity QFunction, not an identity operator. Restriction/prolongation are still needed.

backends/opt/ceed-opt-operator.c
```diff
@@ -193,18 +225,36 @@ static int CeedOperatorSetup_Opt(CeedOperator op) {
...
+      ierr = CeedQFunctionFieldGetEvalMode(qfinputfields[i], &inmode);
+      CeedChk(ierr);
+      CeedEvalMode outmode;
+      ierr = CeedQFunctionFieldGetEvalMode(qfoutputfields[i], &outmode);
+      CeedChk(ierr);
+      if (inmode == CEED_EVAL_NONE && outmode == CEED_EVAL_NONE) {
+        return CeedError(ceed, 1,
+                         "CEED_EVAL_NONE for a matching input and output does not make sense with identity QFunction");
+      } else {
+        impl->qvecsout[i] = impl->qvecsin[i];
```
jeremylt: I set the output qvecs and input qvecs to be the same to avoid copying

backends/opt/ceed-opt-operator.c
```diff
@@ -379,8 +433,10 @@ static inline int CeedOperatorApply_Opt(CeedOperator op,
     }
 
     // Q function
-    ierr = CeedQFunctionApply(qf, Q*blksize, impl->qvecsin, impl->qvecsout);
-    CeedChk(ierr);
+    if (!identityQF) {
```
jeremylt: I only call the QFunction if it's not identity.

backends/opt/ceed-opt-operator.c
YohannDudouit: Why wouldn't you do a `CEED_EVAL_NONE` with memcpy instead?

jeremylt: I'm trying to avoid a memcopy altogether.

YohannDudouit: What I meant is that I think that applying `Restriction`+`Prolongation` and doing nothing for the identity qFunction is more expensive than memcpy the input to the output (without `Restriction`+`Prolongation`). If it's really an identity operator, then the fastest should be memcpy, or am I missing something?

jeremylt: There is still a basis action. The QFunction is identity, but the Operator is not. I have two operators in multigrid that do something with an identity QFunction.

Interpolate Operator:
Restrict Coarse, Basis Interp Coarse to Fine, QFunction Identity, Basis Eval None, Prolongate Fine
Restriction Operator:
Restrict Fine, Basis Interp Fine to Coarse, QFunction Identity, Basis Eval None, Prolongate Coarse

Currently CEED_EVAL_NONE is a no operation event in OpApply; I use the Qvec as the Evec, or vice versa. This PR extends that idea so I don't do a QFunction that is basically a memcopy for no reason when I can use the input Qvecs as the output Qvecs.

YohannDudouit:
I'm not a big fan that `CeedQFunctionCreateIdentity_*` for each backend is an empty function, shouldn't we just remove the backend specific `CeedQFunctionCreateIdentity_*`?

jeremylt:
I don't have strong feelings either way. I wanted to leave the option to do something there, just in case.

jedbrown:
I guess I'd think there could be a reference implementation that naively copies from input to output vector. Such an implementation would be kernel-specific now, but would be unified once we manage to unify our qfunctions. I think it's valuable to have such a reference implementation because it lowers the bar for writing new backends. They don't have to special-case identity if they don't feel like it, and all apps will still run correctly.

jeremylt:
Is there a clean way to write a qfunction to handle a generic number of input/output pairs?

If we want a fallback for all backends, it seems like having the interface default to something like below with the backend's preferred memtype would be easier. We would be able to capture all backend qfunction formats with a single function. 
```
for (int i=0; i<numinputfields; i++) {
      const CeedScalar *q;
      ierr = CeedVectorGetArrayRead(in[i], CEED_MEM_DEVICE, &q);
      CeedChk(ierr);
      ierr = CeedVectorSetArray(out[i], CEED_MEM_DEVICE,
                                CEED_COPY_VALUES, (CeedScalar *)q);
      CeedChk(ierr);
      ierr = CeedVectorRestoreArrayRead(in[i], &q); CeedChk(ierr);
    }
```

jedbrown:
I think users should have to name fields. The fallback constructor could put the number of in/out pairs into the context for its generic implementation, which just loops over pairs and copies.

jeremylt:
Currently in this PR the user has to name the qfunction fields. It would make a mess of operator code to do otherwise. The arrays of vectors U and V are already ordered with the pairs lined up as given by the user. I think my 'default' implementation above would be easier than what I have now and easier than making a generic qfunction that handles 1-16 pairs.

jedbrown:
My point was that can pass the integer number of fields in through the context. You don't need "16" pairs, just
```c
int Mass(void *ctx, int Q, const double *const *in, double *const *out) {
  int nfields = *(int*)ctx;
  for (int f=0; f<nfields; f++) {
    const double *u = in[f];
    double *v = out[f];
    for (int i=0; i<Q; i++) {
      v[i] = u[i];
    }
  }
  return 0;
}
```

interface/ceed-qfunction.c
```diff
@@ -75,6 +76,85 @@ int CeedQFunctionCreateInterior(Ceed ceed, CeedInt vlength,
   return 0;
 }
 
+/**
+  @brief Default implementation of CeedQFunctionApplyIdentity.
```
jeremylt: I set up this as a backend agnostic default.

jeremylt:
I get what you're saying with the context, my brain just went a different direction.

It feels easier to me to fall back to the new identity apply I set up for any backend rather than making the qfunction. I think it's less code both now and in the future.

interface/ceed-qfunction.c
```diff
@@ -75,6 +76,85 @@ int CeedQFunctionCreateInterior(Ceed ceed, CeedInt vlength,
...
+
+  // Write inputs to outputs
+  for (int i=0; i<qf->numinputfields; i++) {
+        const CeedScalar *q;
+        ierr = CeedVectorGetArrayRead(u[i], memtype, &q);
+        CeedChk(ierr);
+        ierr = CeedVectorSetArray(v[i], memtype,
+                                  CEED_COPY_VALUES, (CeedScalar *)q);
+        CeedChk(ierr);
+        ierr = CeedVectorRestoreArrayRead(u[i], &q); CeedChk(ierr);
```
jedbrown: Identation looks funny.

interface/ceed-qfunction.c
jedbrown: I guess I was concerned about a backend that would apply in blocks (without ever creating a full E-vector). You're currently managing this by creating short work vectors and still calling `CeedQFunctionApply`, but your non-QFunction C code can't be transferred to a device, for example.

If you like the current implementation, we can stick with it, but we'll want to keep this alternate in mind.

interface/ceed-qfunction.c
jeremylt: got it

interface/ceed-qfunction.c
jeremylt: I'm not sure I understand your concern. With this we use the backend COPY_VALUES, so it will be using device operations on the CUDA backend, for instance.

YohannDudouit: If you take a look at the `cuda-gen` backend for instance, you never have access from the host to any E-vector nor Q-vector for instance.

jeremylt: Ah, that makes sense. Though, I'd imagine that cuda-gen will want to do something smarter than run a meaningless qfunction, right?

YohannDudouit: Yes, that's true, but I think @jedbrown has a point, especially considering that some backends are lagging on the development. Each of this new little feature also make writing a new backend from scratch more challenging. But I think it's okay if you don't want to have this default implementation, even though that's a small weakness from a design point of view.

jedbrown: Well, run-time compilation with access to the identity code (as it does), may be able to elide it there. My concern is that we want backends to be able to fuse and schedule composite backends (e.g., mixed topology/mixed order meshes using CUDA Streams or Graphs), but this approach obliges the backend to either handle identity as a special case or to have the reference identity handling prevent that fusion.

jeremylt: Yeah, I don't like my approach now (the fact that OCCA changed something since I last updated it locally and my approach doesn't work on OCCA doesn't help). I was trying to avoid dragging around a collection of the different identity qfunction formats. Really, it would be nice to do this after CUDA and CPU share a source and we start the qfunction gallery.

jeremylt: I was able to fix the OCCA issue. I'm happy with everything but the generic implementation in here.

jeremylt:
I will finish this after #311.

jeremylt:
I am rebuilding this with the gallery. After #306 is merged, I'll add the optimizations to skip the unnecessary memcpy in CPU backends.

jeremylt:
This makes sense to me to add now and then do backend specific optimizations in follow on PRs. I've been trying to make smaller, easier to review PRs.

interface/ceed-qfunction.c
```diff
@@ -575,6 +621,10 @@ int CeedQFunctionDestroy(CeedQFunction *qf) {
   }
   ierr = CeedFree(&(*qf)->inputfields); CeedChk(ierr);
   ierr = CeedFree(&(*qf)->outputfields); CeedChk(ierr);
+  // Free ctx if identity
+  if ((*qf)->identity && (*qf)->ctx) {
```
jedbrown:
```suggestion
  if ((*qf)->identity) {
```
It's safe to `CeedFree`, so may as well keep the conditional simple.

tests/tap.sh
```diff
@@ -103,7 +103,7 @@ for ((i=0;i<${#backends[@]};++i)); do
 
     # grep to skip tests t410 and ex1 for OCCA' \
```
jedbrown: Perhaps better to explain why some tests need this special treatment (and why it doesn't need to be replicated for junit.py) than to incompletely reproduce the conditional.

interface/ceed-qfunction.c
```diff
@@ -163,6 +164,35 @@ int CeedQFunctionCreateInteriorByName(Ceed ceed,  const char *name,
   return 0;
 }
 
+/**
+  @brief Create an identity CeedQFunction. Inputs are written into outputs in
+           the order given.
```
jedbrown: Could the description explain why one might use this? Perhaps sketch an example where this enables a backend to fast-path?

tests/tap.sh
jeremylt: It is a temporary exception for the OCCA backend. This will be removed when #305 is merged. I can add the exception to junit for uniformity if desired - I left it alone only because this is supposed to be temporary and we don't currently test the OCCA backend with junit.

interface/ceed-qfunction.c
jeremylt: Updated

interface/ceed-operator.c
```diff
@@ -25,7 +25,9 @@
 ///   @{
 
 /**
-  @brief Create an operator from element restriction, basis, and QFunction
+  @brief Create a CeedOperator and associate a CeedQFunction. A CeedBasis and
```
jeremylt: The operator documentation was dated.

gallery/identity/ceed-identity.h
```diff
@@ -0,0 +1,44 @@
...
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+/**
+ This file is not compiled into libCEED. This file provides a template to
+   build additional gallery QFunctions. Copy this file and the registeration/
```
jedbrown: spelling "registration"

gallery/identity/ceed-identity.c
```diff
@@ -0,0 +1,55 @@
...
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+#include <string.h>
+#include "ceed-backend.h"
+#include "ceed-identity.h"
+
+/**
+ This file is not compiled into libCEED. This file provides a template to
```
jedbrown: This just isn't true. It is compiled into `libceed.so`. It doesn't expose any public symbols, but it does have a static constructor. Same comment in `ceed-identity.h`.

jedbrown:
Looks good to squash-merge.

PR #288, opened 2019-07-01
[WIP][DISCUSS] Single source QFunction proposal

YohannDudouit:
This PR proposes a `qFunction` format that would allow the same code to be used on host and on device.
The `qFunction` is in the `.qf` files, but could be moved to `.h` files. That would require cpu backends to define the macro `CEED_QFUNCTION`.


dmed256:
@YohannDudouit I like this design since we should be able to use it with OCCA kernels

I took one of your `.qf` files and was able to translate it right away!

I've looked at your JIT-compiled kernel and will probably base the revisited OCCA version on your CUDA one. That way I can leverage the work you did on speedup up the CUDA version and make OCCA fast on GPU versions!

@jeremylt What do you think?

### Details

Setting

```cpp
#define CEED_QFUNCTION inline

typedef int CeedInt;
typedef double CeedScalar;

<bp1.qf-source>

// Dummy kernel to avoid failing `occa translate`
@kernel void foo(const int N) {
  for (int i = 0; i < N; ++i; @tile(16, @outer, @inner)) {
  }
}
```

and running

```bash
occa translate --mode CUDA examples/nek5000/bp1.qf
```

ends up translating the to:

```cpp
typedef int CeedInt;
typedef double CeedScalar;

__device__ inline void masssetupf(void *ctx,
                                  CeedInt Q,
                                  const CeedScalar * const *in,
                                  CeedScalar * const *out) {
  CeedScalar *rho = out[0], *rhs = out[1];
  const CeedScalar *x = in[0];
  const CeedScalar *J = in[1];
  const CeedScalar *w = in[2];
  for (CeedInt i = 0; i < Q; i++) {
    CeedScalar det = (J[i + Q * 0] * (J[i + Q * 4] * J[i + Q * 8] - J[i + Q * 5] * J[i + Q * 7]) - J[i + Q * 1] * (J[i + Q * 3] * J[i + Q * 8] - J[i + Q * 5] * J[i + Q * 6]) + J[i + Q * 2] * (J[i + Q * 3] * J[i + Q * 7] - J[i + Q * 4] * J[i + Q * 6]));
    rho[i] = det * w[i];
    rhs[i] = rho[i] * w[i] * sqrt(x[i] * x[i] + x[i + Q] * x[i + Q] + x[i + 2 * Q] * x[i + 2 * Q]);
  }
}

__device__ inline void massf(void *ctx,
                             CeedInt Q,
                             const CeedScalar * const *in,
                             CeedScalar * const *out) {
  const CeedScalar *u = in[0], *rho = in[1];
  CeedScalar *v = out[0];
  for (CeedInt i = 0; i < Q; i++) {
    v[i] = rho[i] * u[i];
  }
}

// Ignore the dummy kernel
extern "C" __global__ void _occa_foo_0(const int N) {
  {
    int _occa_tiled_i = 0 + (16 * blockIdx.x);
    {
      int i = _occa_tiled_i + threadIdx.x;
      if (i < N) {}
    }
  }
}
```

jedbrown:
This looks great! Is the memory coalescing discussion from earlier moot because you can always call with Q=1 or for some other reason? I'm also open to a macro in front of the `for` loop that can expand to a pragma (OpenMP or OpenACC).

YohannDudouit:
@jedbrown The trick is that it's only an `inline __device__` function on GPU, thus we can build around this function as we like. My concern was that the compiler couldn't map registers to the double pointer `CeedScalar**`, but apparently he is because the jit compilation allows tailored code. The memory coalescing can be handled by the kernel using that `inline __device__` qfunction, and thus doesn't seem to be a concern any more. I'll write tomorrow a way for previous Cuda backends to use `.qf` files and not `.cu` files as currently.

jeremylt:
Looks great! Thank you!

Re CPU backends - We want to support CPU backends that don't do JiT. One way to do this is to convert to `.h` and have the macro fill the current info and compile it into the user code.

YohannDudouit:
@jeremylt I'll do this once I'm done with rewriting the qFunction for the previous cuda backends. Any recommendation where to define `CEED_QFUNCTION` for CPU backends?

jedbrown:
Good point about dropping the return values. I suppose atomic write to a common value only in case of error may not be too bad (assuming the errors are very rare). I think I'd still rather return an error code even if it isn't checked by GPU backends, but would let it go if it's a genuine problem to support.

In `ceed.h`, I would do something like this
```c
#ifndef CEED_QFUNCTION
#  define CEED_QFUNCTION static
#endif
```
so that other backends don't have to `#undef`.

dmed256:
```
#ifndef CEED_QFUNCTION
#  define CEED_QFUNCTION static
#endif
```
^ Should we define it as an empty macro by default? Header guards should normally prevent duplicate symbols.

jeremylt:
Since the CPU backends are the only non-JiT backends, we want to set the default to `static` rather than empty to get the right compiled CPU code.

jedbrown:
It isn't incorrect if not `static`, but I prefer not to expose the symbols when not necessary.

YohannDudouit:
It's okay if `CEED_QFUNCTION` is defined differently than what I use for GPUs, I'm jit compiling the code anyway so I won't see the value you give to `CEED_QFUNCTION`. I dropped the return value, but I could put it back, however using `CeedError` is more problematic as GPU kernels don't return values, and I'm not sure we want to do a `memcpy` for that. I guess that there should be a way to use Cuda errors if we really want to return an error code.

jedbrown:
The qfunction is a device function, not a kernel. I wouldn't mind if the kernel (which calls the qfunction) just ignores that value. I like the return value because it helps to give a useful error message when debugging a qfunction on the CPU.

YohannDudouit:
Okay, I'll add the return value.

YohannDudouit:
@jedbrown Are you okay with having to define the qFunctions in a separate `.h` file? Otherwise the jit compilation would most certainly fail.

jedbrown:
Yeah, I'm fine with using `.h` files that can be included from JIT. Some day the JIT could learn to filter/ignore other stuff to relax that restriction, but I don't think it's important any time soon (if ever).

YohannDudouit:
I'm having some issues handling math functions in a clean way, for functions like `sqrt`, `sin`, `cos`. I can omit the standard `#include <math.h>` it gives me some warnings, but it works overall (not clean though). If I include `math.h` the jit compilation fails. Any idea to solve this in a clean way?
(@jedbrown look at the petsc examples)

jedbrown:
I see math functions are predefined in CUDA compilation units. Seems like we can do
```c
#ifndef __CUDACC__
#  include <math.h.>
#endif
```
It may be useful to make a `ceed-qfunction.h` the defines types (`CeedScalar`) and stuff like this that would break NVRTC. (It's fine to include `math.h` in `*.cu` source complied by `nvcc`, but if NVRTC can't handle it, we can work around it like this.)

YohannDudouit:
@jedbrown Thanks, that worked. I'm a bit worried that user will have to do this guarded include though.

jeremylt:
In master, I'm explicitly passing the name of the 5 different `.h` files the QFunctions are in for the single `bps.c` in the PETSc folder. (See below)

Thoughts about pulling this up to the examples folder makefile or the toplevel makefile?

In the PETSc examples makefile:
```
-D__DIR__=\"$(dir $(abspath $(lastword $<)))\"
```
In `bps.c`:
```
#define PATH(BASE) __DIR__ #BASE
```
```
fOCCA = PATH(bp1.h:SetupMass)
```
We should consider renaming the `fOCCA` parameter and make sure we are clear that the user needs to set this with the true filename and path of the header that contains the qfunction.

jedbrown:
That repetition was why I suggested `ceed-qfunction.h` to contain that logic. Is that a problem because NVRTC would need to find CEED headers?

It would be handy to know why including `math.h` fails with nvrtc, but not with nvcc.

jedbrown:
Mmm, this macro stuff is actually problematic. Paths containing words such as `linux` and `unix` can be mangled by macros that are in scope. https://github.com/mfem/mfem/issues/974
Also, macros starting with `_` are reserved by the compiler/environment.
IIRC, there was a problem that `__FILE__` with some compilers does not contain an absolute path and there was no workaround short of manually defining the path? Which compiler is that?

YohannDudouit:
Here is the compilation error I get, it seems to be nvrtc specific:
```
not ok 21 petsc-bp1 /gpu/cuda/ref stderr
# +/home/dudouit1/Projects/libCEED/backends/cuda/ceed-cuda.c:70 in compile(): NVRTC_ERROR_COMPILATION
# +/usr/include/bits/mathcalls.h(202): error: linkage specification is incompatible with previous "isinf"
# +__nv_nvrtc_builtin_header.h(88973): here
# +
# +/usr/include/bits/mathcalls.h(235): error: linkage specification is incompatible with previous "isnan"
# +__nv_nvrtc_builtin_header.h(88970): here
```

YohannDudouit:
Including headers is not trivial with nvrtc: [Basic usage of nvrtc](https://docs.nvidia.com/cuda/nvrtc/index.html#basic-usage). I'm not sure I understand what [jitify is doing](https://github.com/NVIDIA/jitify/blob/master/jitify.hpp#L1601).

jeremylt:
@jedbrown, I'm happy to update that macro to be more standards compliant or portable, but it does seem to me that we want to facilitate the user writing their qfunctions in multiple header files for situations like our BP1-6 PETSc example. I found the basis for our current solution on a stack overflow page somewhere.

The issue with `__FILE__` is with the Intel Fortran compiler on CU Summit.

jeremylt:
So we remember: We will need to rebase or merge master into this branch before merging this to master. This branch was cut before the PETSc BP updates made it into master and merging this without somehow rebasing or merging in master first will make that folder into a mess.

YohannDudouit:
The problem with `__FILE__` is that it depends how your makefile is written, even with gcc, if you compile using absolute or relative path, `__FILE__` contains what has been used during compilation.

jeremylt:
I checked for that on CU's machine. On that machine ifort ignores the absolute path we give in the makefile and uses something else.

jedbrown:
With respect to Fortran, we're putting qfunctions into headers anyway (when we wish to use JIT). I believe it is the case that any `#include <file.h>` (which generally must have corresponding `-I/path` flags) produces the absolute path within that header. This may be dirty CPP, but:
```c
// defined like this in CPU header; as you like for JIT
  static int name

// User code
CEED_QFUNCTION(mass)(void *ctx,CeedInt Q,
  const CeedScalar *u = in[0], *rho = in[1];
  CeedScalar *v = out[0];
  for (CeedInt i=0; i<Q; i++) {
    v[i] = rho[i] * u[i];
  }
  return 0;
}
```
Then we create a QFunction by `CeedQFunctionCreateInterior(..., mass, mass_loc, &qf_mass);`

jeremylt:
Great or crazy idea based upon your perspective on Fortran. Since we will need a C qfunction header anyway for GPU support, is there a way to cleanly use that for the user QFunction on the CPU side?

jedbrown:
Yeah, I'd make it the default CPU definition.
I don't think there's any JIT for us to do with Fortran qfunctions at present (it's surely out of scope for this PR). So let's assume the qfunctions are implemented in C. It's possible to use ISO C bindings (Fortran2003) to access C symbols, but not in F77. We could compile the C (in a separate compilation unit) with a symbol name that is mangled for Fortran, then call it from a simple Fortran stub, but I think that's circuitous. Easier to have a setup function written in C (with mangled symbol to be callable from Fortran) that creates a QFunction and returns the Fortran handle. I don't know if this mode of operation will be very popular except perhaps for comparisons with Nek.

jeremylt:
Now that CUDA Gen has been merged, can we ke this into a PR to master? I believe that we all generally agree on this approach and it's just a matter of ironing out details.

include/ceed.h
```diff
@@ -54,6 +54,9 @@
 #else
 #  define CEED_EXTERN extern
 #endif
+#ifndef CEED_QFUNCTION
+#  define CEED_QFUNCTION static
+#endif
```
jeremylt: I think I'd change this to @jedbrown's suggestion of
```
  static int name
```

so that we can use 
```
  CeedQFunctionCreateInterior(ceed, 1, mass, mass_loc, &qf_mass);
```

jeremylt: This lets us have qfunction files name differently than the base file without my crazy machinations it the PETSc unified BP code.

YohannDudouit:
Closed in favour of #304

PR #292, opened 2019-07-12
Fix NS bug

valeriabarra:
This small PR fixes a bug due to misordering in the derivatives of **u**, in the Navier-Stokes example, noticed by @LeilaGhaffari [https://github.com/LeilaGhaffari/libCEED/commit/c52a89a2daefcc37fbdea45fef0b6f6523d35a1a](here).
It is currently under testing with implicit TS, but it didn't significantly change the solution behavior with the explicit TS.

### DIFF
--- a/examples/navier-stokes/densitycurrent.h
+++ b/examples/navier-stokes/densitycurrent.h
@@ -228,16 +228,16 @@ static int DC(void *ctx, CeedInt Q,
                                   dq[i+(0+5*1)*Q],
                                   dq[i+(0+5*2)*Q]
                                };
-    const CeedScalar du[9]   = { (dq[i+(0+5*1)*Q] - drho[0]*u[0]) / rho,
-                                 (dq[i+(1+5*1)*Q] - drho[0]*u[1]) / rho,
-                                 (dq[i+(2+5*1)*Q] - drho[0]*u[2]) / rho,
-                                 (dq[i+(0+5*2)*Q] - drho[1]*u[0]) / rho,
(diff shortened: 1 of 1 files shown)


examples/navier-stokes/densitycurrent.h
```diff
@@ -217,86 +217,84 @@ static int DC(void *ctx, CeedInt Q,
...
+    const CeedScalar drho[3]  =  { dq[i+(0+5*0)*Q],
+                                   dq[i+(0+5*1)*Q],
+                                   dq[i+(0+5*2)*Q]
+                                 };
     const CeedScalar du[3][3] = {{(dq[i+(1+5*0)*Q] - drho[0]*u[0]) / rho,
-                                  (dq[i+(1+5*1)*Q] - drho[1]*u[0]) / rho,
-                                  (dq[i+(1+5*2)*Q] - drho[2]*u[0]) / rho},
-                                 {(dq[i+(2+5*0)*Q] - drho[0]*u[1]) / rho,
+                                  (dq[i+(1+5*1)*Q] - drho[0]*u[1]) / rho,
+                                  (dq[i+(1+5*2)*Q] - drho[0]*u[2]) / rho},
```
jedbrown: What is going on here? As I understand it, these three should come from the identity

    grad(rho u) = grad(rho) u + rho grad(u)

and the next three are

    grad(rho v) = grad(rho) v + rho grad(v)

The change introduced in this commit would mix `v` and `w` into the expression involving `rho u`.

PR #293, opened 2019-07-12
[WIP][Discussion]  Add ElemRstrApplyMean

jeremylt:
This WIP PR adds an element restriction with averaging of shared DoFs (none, arithmetic, or harmonic).

This will be useful for certain preconditioning operators.

Thoughts on my proposed interface?

Mean Types:
```c
/// Denotes which averaging technique should be used for a transpose restriction
/// with shared DoFs
typedef enum {
  /// No mean
  CEED_MEAN_NONE,
  /// Arithmetic mean
  CEED_MEAN_ARITHMETIC,
  /// Harmonic mean
  CEED_MEAN_HARMONIC
} CeedMeanType;
```

RestrictionApplyMean options:
```c
CEED_EXTERN int CeedElemRestrictionApplyMean(CeedElemRestriction rstr,
    CeedTransposeMode tmode, CeedTransposeMode lmode, CeedMeanType mtype,
    CeedVector u, CeedVector ru, CeedRequest *request);

CEED_EXTERN int CeedElemRestrictionApplyMeanBlock(CeedElemRestriction rstr,
    CeedInt block, CeedTransposeMode tmode, CeedTransposeMode lmode,
    CeedMeanType mtype, CeedVector u, CeedVector ru, CeedRequest *request);
CEED_EXTERN int CeedElemRestrictionApplyMeanBlockFinalize(CeedElemRestriction rstr,
    CeedTransposeMode tmode, CeedMeanType mtype, CeedVector ru,
    CeedRequest *request);
```

`CEED_MEAN_ARITHMETIC` will be straightforward, but `CEED_MEAN_HARMONIC` will be more involved.

### DIFF
--- a/include/ceed-impl.h
+++ b/include/ceed-impl.h
@@ -95,6 +95,13 @@ struct CeedElemRestriction_private {
                CeedVector, CeedVector, CeedRequest *);
   int (*ApplyBlock)(CeedElemRestriction, CeedInt, CeedTransposeMode,
                     CeedTransposeMode, CeedVector, CeedVector, CeedRequest *);
+  int (*ApplyMean)(CeedElemRestriction, CeedTransposeMode, CeedTransposeMode,
+                   CeedMeanType, CeedVector, CeedVector, CeedRequest *);
+  int (*ApplyMeanBlock)(CeedElemRestriction, CeedInt, CeedTransposeMode,
+                        CeedMeanType, CeedTransposeMode, CeedVector, CeedVector,
+                        CeedRequest *);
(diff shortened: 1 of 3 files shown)


jedbrown:
I can see how this would be useful, but we generally don't want the math to depend on parallelism. So how does one use this interface in parallel without having results that differ when some dofs are shared with other ranks?

jeremylt:
Good point, I think parallel makes a mess of the harmonic mean part, and the arithmetic mean part can already be done with what is in the library by the user and the parent program in a way that works fine in parallel.

jedbrown:
BDDC never needs a harmonic mean, just (possibly) a weighted mean. We can compute the desired weights with an auxiliary/setup qfunction.

PR #295, opened 2019-07-12
BugFix - Resource strcmp in LIBXSMM backends

jeremylt:
This PR fixes a small bug in the LIBXSMM backend.

### DIFF
--- a/backends/ref/ceed-ref.c
+++ b/backends/ref/ceed-ref.c
@@ -18,9 +18,10 @@
 
 static int CeedInit_Ref(const char *resource, Ceed ceed) {
   int ierr;
-  if (strcmp(resource, "/cpu/self")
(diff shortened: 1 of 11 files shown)


jeremylt:
Good idea. Blocked generally gives best performance unless the element size is rather large, so I will use that for all backends.

backends/ref/ceed-ref.c
```diff
@@ -18,9 +18,10 @@
 
 static int CeedInit_Ref(const char *resource, Ceed ceed) {
   int ierr;
-  if (strcmp(resource, "/cpu/self")
+  if (strcmp(resource, "/cpu/self") && strcmp(resource, "/cpu/self/ref")
```
jeremylt: I set serial ref to be the default rather than blocked - it is the simplest, which is the purpose of ref.

PR #296, opened 2019-07-15
examples/petsc: fix max reduction type and address

jedbrown:


jeremylt:
It looks good to me. For my curiosity, does this just run faster? It seems like a parameter change to do the same reduce.

jedbrown:
You were doing the reduction on pointer values, then printing the local value from rank 0.On Jul 14, 2019 19:01, Jeremy L Thompson <notifications@github.com> wrote:It looks good to me. For my curiosity, does this just run faster? It seems like a parameter change to do the same reduce.

PR #298, opened 2019-07-18
Fix ref backend priorities

jeremylt:
I had set `/cpu/self/ref/serial` to be the default for `/cpu/self/ref` but I did not update the priorities to match.


jeremylt:
@jedbrown, it looks like Azure is getting some strange docker error. Investigating.

PR #302, opened 2019-07-30
Create DMPlex PETSc BPs

jeremylt:
This WIP branch merges @jedbrown and @oanaoana's work adding DMPlex to the PETSc examples with the unified `bps.c` code.

~~Right now BP1-2 work and BP3 seems to work.~~ BP1-6 work

~~This branch needs cleanup as Valgrind indicates there is a lot memory in use at the end.~~ done


jeremylt:
Edit: BP1-6 all work, but BP4/6 seem to have higher error than BP3/5, which has to be an error.

jedbrown:
I take it you're comparing to the `bps.c` variants?

jeremylt:
`make prove-all` works locally if you have the latest PETSc master.

jeremylt:
@jed, here is what I mean by BP4/6 having higher error than BP3/5:
```
[jeremy@pheonix petsc]$ ./bpsdmplex -petscspace_degree 3 -problem bp1

-- CEED Benchmark Problem 1 -- libCEED + PETSc --
  libCEED:
    libCEED Backend                    : /cpu/self
  Mesh:
    Number of 1D Basis Nodes (p)       : 4
    Number of 1D Quadrature Points (q) : 5
    Global DOFs                        : 1000
    Owned DOFs                         : 1000
  KSP:
    Total KSP Iterations               : 15
    Final rnorm                        : 2.825725e-11
    Pointwise Error (max)              : 2.605215e-03
[jeremy@pheonix petsc]$ ./bpsdmplex -petscspace_degree 3 -problem bp2

-- CEED Benchmark Problem 2 -- libCEED + PETSc --
  libCEED:
    libCEED Backend                    : /cpu/self
  Mesh:
    Number of 1D Basis Nodes (p)       : 4
    Number of 1D Quadrature Points (q) : 5
    Global DOFs                        : 1000
    Owned DOFs                         : 1000
  KSP:
    Total KSP Iterations               : 15
    Final rnorm                        : 4.847511e-11
    Pointwise Error (max)              : 2.605215e-03
[jeremy@pheonix petsc]$ ./bpsdmplex -petscspace_degree 3 -problem bp3

-- CEED Benchmark Problem 3 -- libCEED + PETSc --
  libCEED:
    libCEED Backend                    : /cpu/self
  Mesh:
    Number of 1D Basis Nodes (p)       : 4
    Number of 1D Quadrature Points (q) : 5
    Global DOFs                        : 512
    Owned DOFs                         : 512
  KSP:
    Total KSP Iterations               : 9
    Final rnorm                        : 2.503918e-16
    Pointwise Error (max)              : 4.244352e-02
[jeremy@pheonix petsc]$ ./bpsdmplex -petscspace_degree 3 -problem bp4

-- CEED Benchmark Problem 4 -- libCEED + PETSc --
  libCEED:
    libCEED Backend                    : /cpu/self
  Mesh:
    Number of 1D Basis Nodes (p)       : 4
    Number of 1D Quadrature Points (q) : 5
    Global DOFs                        : 1000
    Owned DOFs                         : 1000
  KSP:
    Total KSP Iterations               : 77
    Final rnorm                        : 3.146844e-10
    Pointwise Error (max)              : 2.546926e+00
```

Notice that BP1/2 have the same number of CG its and max error. For `bps.c`, BP3/4 and BP5/6 have the same number of CG its and max error as well but not for `bpsdmplex.c`. I must be accounting for the multiplicity incorrectly.

examples/petsc/bpsdmplex.c
```diff
@@ -0,0 +1,747 @@
...
+  ierr = DMGetDimension(dm, &dim); CHKERRQ(ierr);
+  ierr = PetscFECreateDefault(PETSC_COMM_SELF, dim, vscale, PETSC_FALSE, NULL,
+                              PETSC_DETERMINE, &fe);
+  CHKERRQ(ierr);
+  ierr = DMSetFromOptions(dm); CHKERRQ(ierr);
+  ierr = DMAddField(dm, NULL, (PetscObject)fe); CHKERRQ(ierr);
+  ierr = DMCreateDS(dm); CHKERRQ(ierr);
+  if (enforce_bc) {
+    marker_ids[0] = vscale;
+    ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, "wall", "marker", 0, 0, NULL,
```
jeremylt: This is where the BP4/6 problem lies, I believe.

jedbrown: Are the boundary conditions homogeneous or inhomogeneous? If inhomogeneous, you'll need `DMPlexInsertBoundaryValues`.
`DMGlobalToLocal` populates only the inessential dofs. That's usually what you want when applying a linear operator (leaving the others zero), but not when constructing a solution. Note that you may also need to lift those contributions to the RHS in this mode.

jeremylt: We have homogeneous BCs. The BP 3/5 result is correct, so I think that there must be some issue when jumping from 1 component to 3.

examples/petsc/bpsdmplex.c
```diff
@@ -74,8 +74,9 @@ static int CreateRestrictionPlex(Ceed ceed, CeedInt P, CeedInt ncomp,
                                    &indices, NULL); CHKERRQ(ierr);
     for (i=0; i<numindices; i+=ncomp) {
       for (PetscInt j=0; j<ncomp; j++) {
-        if (indices[i+j] != indices[i]+j) SETERRQ1(PETSC_COMM_SELF,
-          PETSC_ERR_ARG_INCOMP, "Cell %D closure indices not interlaced", c);
+        if (indices[i+j] != indices[i] + copysign(j, indices[i]))
+          SETERRQ1(PETSC_COMM_SELF, PETSC_ERR_ARG_INCOMP,
+                   "Cell %D closure indices not interlaced", c);
       }
```
jeremylt: This update seems to fix it. The check for the indices interlacing didn't account for essential BCs being encoded as `-(loc + 1)`.

jeremylt:
It all appears to work now. I am testing Travis with PETSc 3.11.3

examples/petsc/bpsdmplex.c
jedbrown: Yeah, I have a more general version of this function in `jed/ns-dmplex`. That could eventually move into a library (though it depends on both libceed and libpetsc, so either creates a dependency or needs to be in a separate library).

jeremylt: Ah. I must have missed it - I got a bit turned around between all of files in that folder.

Given those options, it seems like a separate small utility library might be the way to go to prevent the extra dependency. Though if PETSc + CEED for GPU finite elements gets more traction, it might belong as part of that effort.

jeremylt:
Note: Codecov is decreasing because we are adding a PETSc example without including it in Travis.

jedbrown:
Any objection to updating Travis to use PETSc 'master'?

jeremylt:
I can do that easily enough. I would still want to do a lightweight build against a specific commit so we can cache instead of building each time. Do you have recommended options for `./configure` to make the lightest build that does what we need?

jeremylt:
@jedbrown, Travis should be set with current PETSc master, but feel free to make any tweaks you like.

jeremylt:
Apt currently doesn't work for Travis running Bionic: https://github.com/travis-ci/travis-build/pull/1754

It looks like this is about to be changed, so I'll try again with Bionic in the morning.

jeremylt:
Travis with Bionic is still broken after the merge of https://github.com/travis-ci/travis-build/pull/1754 and it did not appear to fix the issue in any case.

I dropped Bionic and just manually set `vscale` as is done in `bps.c`.

jeremylt:
@oanaoana @jedbrown I fixed the BCs.

examples/petsc/bpsdmplex.c
```diff
@@ -0,0 +1,752 @@
...
+PetscErrorCode BCsMass(PetscInt dim, PetscReal time, const PetscReal x[],
+                       PetscInt vscale, PetscScalar *u, void *ctx) {
+  for (PetscInt i = 0; i < vscale; i++)
+    u[i] = PetscSqrtScalar(PetscSqr(x[0]) + PetscSqr(x[1]) +
+                           PetscSqr(x[2]));
+
+  return 0;
+}
+
+PetscErrorCode BCsDiff(PetscInt dim, PetscReal time, const PetscReal x[],
```
jeremylt: @jedbrown, with single source QFunctions on the horizon, I did not use this function in the setup QFunctions like you did in your branch. It would cause problems for the code generation backends.

examples/petsc/bpsdmplex.c
```diff
@@ -100,6 +100,45 @@ static int CreateRestrictionPlex(Ceed ceed, CeedInt P, CeedInt ncomp,
...
+           sin(M_PI*(c[2] + k[2]*x[2]));
+
+  return 0;
+}
+
+// Create BC label
+static PetscErrorCode CreateBCLabel(DM dm, const char name[]) {
+  int ierr;
+  DMLabel label;
+
```
jedbrown: Should use `PetscFunctionBeginUser;`

examples/petsc/bpsdmplex.c
```diff
@@ -100,6 +100,45 @@ static int CreateRestrictionPlex(Ceed ceed, CeedInt P, CeedInt ncomp,
...
 
+// BC functions
+PetscErrorCode BCsMass(PetscInt dim, PetscReal time, const PetscReal x[],
+                       PetscInt vscale, PetscScalar *u, void *ctx) {
+  for (PetscInt i = 0; i < vscale; i++)
+    u[i] = PetscSqrtScalar(PetscSqr(x[0]) + PetscSqr(x[1]) +
+                           PetscSqr(x[2]));
+
+  return 0;
+}
```
jedbrown: This is never called due to `enforce_bcs = false`, so why not just use `NULL`?

examples/petsc/bpsdmplex.c
jedbrown: Yeah, okay. That can be left for the future.

examples/petsc/bpsdmplex.c
jeremylt: Currently `enforce_bcs` is a command line option, so a user could call `-enforce_bcs -problem bp1`.

examples/petsc/bpsdmplex.c
jeremylt: Fixed

jedbrown: Still not here? It should be used in any function that calls `CHKERRQ`.

jeremylt: Ah, I had added it to the BC functions. Now its on this function.

PR #304, opened 2019-08-02
Single Source QFunction

YohannDudouit:
This PR proposes a `qFunction` format that would allow the same code to be used on host and on device.
The `qFunction` is in the `.h` files. This introduce the `CEED_QFUNCTION` macro that has to be used to define the `qFunctions`.

- [x] Remove .cu files
- [x] Remove .qf files
- [ ] Remove .okl files


include/ceed.h
```diff
@@ -54,6 +54,9 @@
 #else
 #  define CEED_EXTERN extern
 #endif
+#ifndef CEED_QFUNCTION
```
jeremylt: Comment copied over:

I think I'd change this to @jedbrown's suggestion of
```
  static int name
```

so that we can use 
```
  CeedQFunctionCreateInterior(ceed, 1, mass, mass_loc, &qf_mass);
```

This lets us have qfunction files name differently than the base file without my crazy machinations it the PETSc unified BP code.

jeremylt:
With this change, is it possible/a good idea to have `CeedQFunctionApply_Cuda_gen` run `CeedQFunctionApply_Cuda` when it's being called without an operator?

dmed256:
> I don't see the .okl files as a reason to hold back on this PR if it's ready now, personally. I'm not sure what @dmed256's timeline for the new OCCA backend is, but this branch alone is a big improvement to the library.

:+1:

jeremylt:
I added the improved `CEED_QFUNCTION` macro.

jeremylt:
All we need now is the new `bpsdmplex.c`, but otherwise I think this is ready.

jeremylt:
@YohannDudouit, I think we can merge this now, but feel free to change back the label if I'm wrong.

jeremylt:
Almost ready to merge?

jeremylt:
@YohannDudouit, we're all set. The honor is all yours! Could you please put this message (below) at the top of the Squash+Merge commit message? Thanks for your hard work on this branch!

`[NEWS] Breaking change: QFunctionField parameter 'ncomp' changed to 'size'. This change requires setting the previous value of 'ncomp' to 'ncomp*dim' when adding a QFunctionField with eval mode 'CEED_EVAL_GRAD'.`

jeremylt:
> Strictly speaking, function pointers and data pointers cannot be converted. Instead of keeping these alongside each other, what about either using the enum list as in examples/petsc/bps.c or creating a 4-line helper function (called via `PetscFunctionListAdd`) that sets the four fields?

Fixed. Thank you.

jedbrown:
Thanks, @jeremylt. I think this is ready to merge. The `.okl` files mentioned in the checklist at the top of this PR can be handled in #305.

jedbrown:
:tada:  Hooray! This is a major achievement. Thanks, @YohannDudouit and @jeremylt!

PR #307, opened 2019-08-13
Change default NS problemtype to density_current

valeriabarra:
It makes more sense to have the default problemtype to be the density_current example rather than the advection test case, that was intended to be an internal tool for validation of code, and technically does not fall under the umbrella of the NS equations, but it's a transport equation.


valeriabarra:
@jeremylt this is so minimal and orthogonal to PR #304, that I think it can go first, no?

PR #310, opened 2019-08-19
Fix atomicAdd in cuda-gen backend

YohannDudouit:
The preprocessor guards for `atomicAdd` were not having the intended behaviour, this was preventing `nvrtc` to use the hardware `atomicAdd` for compute capability superior to 60 (This was unfortunately not creating a compilation error). This PR proposes a fix to this issue.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -21,20 +21,7 @@
 #include "../cuda-reg/ceed-cuda-reg.h"
 #include "../cuda-shared/ceed-cuda-shared.h"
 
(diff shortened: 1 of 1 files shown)


YohannDudouit:
The problem is not so much related to GPU, the preprocessor instruction `#if __CUDA_ARCH__ < 600` was being processed by the C preprocessor, and thus didn't appear in the string meant for nvrtc. @jedbrown  I'm open to any suggestion as I would like to add some `#pragma unroll` to the string but because of the same issue I cannot.

jeremylt:
Ah, I see. Thanks for explaining.

jedbrown:
I would define a macro using `_Pragma("unroll")` so you don't need guards for each usage.

PR #311, opened 2019-08-19
[WIP] QFunction Create by Name

jeremylt:
This PR creates a QFunction from a gallery by name. I will rebase to master and add some QFunctions and tests after #304 is merged.


jeremylt:
With Single Source merged, I'll complete this tomorrow.

jeremylt:
I need to add tests for all gallery QFs and add a temporary test exception for OCCA.

@YohannDudouit, this should work great with the CUDA backends; I need to test it but it's a pretty straightforward PR.

gallery/ceed-applyMass.c
```diff
@@ -0,0 +1,67 @@
...
+
+#ifndef __CUDACC__
+#  include "ceed-backend.h"
+#  include <string.h>
+#endif
+
+/**
+  @brief Ceed QFunction for applying mass matrix
+**/
+CEED_QFUNCTION(applyMass)(void *ctx, const CeedInt Q,
```
jeremylt: @YohannDudouit, would this file work for CUDA, or do I need to separate the library init code (the two functions at the bottom) from the QFunction code here?

jeremylt:
Ready for review. The only open detail I have is if I need to separate the init code and QFunction code.

I set up a bit of a strange ex2, and I'm open to changing it. I'm using the diffusion operator to compute surface area by intentionally omitting the surface integrals in the weak form. It works, but as you'd expect it has pretty high error. I was going for something simple with a non-zero result.

gallery/diff1d/ceed-applyDiff1D.c
```diff
@@ -0,0 +1,51 @@
...
+  int ierr;
+
+  // Check QFunction name
+  if (strcmp(name, "applyDiff1D"))
+    return CeedError(ceed, 1, "QFunction does not mach name: %s", name);
+
+  // Add QFunction fields
+  const CeedInt dim = 1;
+  ierr = CeedQFunctionAddInput(qf, "du", dim, CEED_EVAL_GRAD); CeedChk(ierr);
+  ierr = CeedQFunctionAddInput(qf, "qdata", dim*(dim+1)/2, CEED_EVAL_NONE);
```
jedbrown: Should we call a helper instead of copy/pasting this dimension-independent code? I know it's just a few lines.

gallery/diff2d/ceed-applyDiff2D.h
```diff
@@ -0,0 +1,38 @@
...
+
+  // out[0] is output to multiply against gradient v, shape [2, nc=1, Q]
+  CeedScalar *dv = out[0];
+
+  // Quadrature point loop
+  for (CeedInt i=0; i<Q; i++) {
+    const CeedScalar du0 = du[i+Q*0];
+    const CeedScalar du1 = du[i+Q*1];
+    dv[i+Q*0] = qd[i+Q*0]*du0 + qd[i+Q*1]*du1;
+    dv[i+Q*1] = qd[i+Q*1]*du0 + qd[i+Q*2]*du1;
```
jedbrown: Should we use a [Voigt/Mandel](https://en.wikipedia.org/wiki/Voigt_notation#Mnemonic_rule) ordering convention for symmetric tensors?  Whatever it is, we should document the convention. I think you've been consistent thus far (albeit a different convention from what you'll find in the literature).

gallery/diff3d/ceed-buildDiff3D.h
```diff
@@ -0,0 +1,65 @@
...
+    const CeedScalar J33 = J[i+Q*8];
+    const CeedScalar A11 = J22*J33 - J23*J32;
+    const CeedScalar A12 = J13*J32 - J12*J33;
+    const CeedScalar A13 = J12*J23 - J13*J22;
+    const CeedScalar A21 = J23*J31 - J21*J33;
+    const CeedScalar A22 = J11*J33 - J13*J31;
+    const CeedScalar A23 = J13*J21 - J11*J23;
+    const CeedScalar A31 = J21*J32 - J22*J31;
+    const CeedScalar A32 = J12*J31 - J11*J32;
+    const CeedScalar A33 = J11*J22 - J12*J21;
```
jedbrown: Would it be possible to use an inline function or (*gasp*) macro for this part, or would that conflict with qfunction processing? It ends up being a lot of copy-pasta.

gallery/diff1d/ceed-applyDiff1D.c
```diff
@@ -0,0 +1,51 @@
...
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+#include <string.h>
+#include "ceed-backend.h"
+#include "ceed-applyDiff1D.h"
+
+/**
+  @brief Set fields for Ceed QFunction building the geometric data for the 1D
+           diff operator
```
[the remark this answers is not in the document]
jeremylt: Wrt "diff" - I mainly left it for uniformity with our examples

Wrt suffix vs prefix, I don't have a strong preference.

gallery/diff1d/ceed-applyDiff1D.c
jeremylt: I'm trying to make the content of each folder independent from each other, as much as I can.I don't love the duplication, but it seems to make it easier for future developers/users to follow. 

Related - Copying ctx is a performance hurdle we haven't sorted out yet for CUDA, so I provided dimensioned versions of this code to give the best performance.

gallery/diff2d/ceed-applyDiff2D.h
jeremylt: Yeah, we've been following the convention setup by MFEM BP3 when it was first merged. I'm happy to change it for all examples and this new code if its desirable.

jedbrown: Maybe we should create and issue to poll the libCEED community.

gallery/diff1d/ceed-applyDiff1D.c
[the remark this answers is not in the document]
jeremylt: Switched to suffix; after reflection I think its better and it makes the build logs and file structure cleaner.

gallery/diff3d/ceed-buildDiff3D.h
jeremylt: What about what I just pushed? I can condense lines 48-53 once we decide if we want to switch how we store the matrix.

gallery/diff3d/ceed-diff3DBuild.h
```diff
@@ -0,0 +1,57 @@
...
+  // out[0] is qdata, size (Q)
+  CeedScalar *qd = out[0];
+
+  // Quadrature point loop
+  for (CeedInt i=0; i<Q; i++) {
+    // Compute the adjoint
+    CeedScalar A[3][3];
+    for (CeedInt j=0; j<3; j++)
+      for (CeedInt k=0; k<3; k++)
+//      A[k][j] = J[j+1][k+1]*J[j+2][k+2] - J[j+1][k+2]*J[j+2][k+1]
```
jedbrown: Commented code is smelly code. Looks good otherwise.

jeremylt: That's not intended as dead code but rather a simplified version of the code below for reference/readability. I can try to make that clearer? Without the VLAs, messy indices like `i+Q*((j+1)%3+3*((k+1)%3))` are somewhat slow to parse by humans, imho.

jedbrown: Oh, gotcha.

jeremylt:
Personally, with the name field being pretty easy to update I'd prefer to lean towards simpler names like Poisson3D for now and add specificity if we find there are a lot of similar QFunctions we want to add before v1.0.

I think adding a piece of documentation at the top of these files (or in a README) with the weak formulation and a list of inputs/outputs would be good. I'm still somewhat annoyed at the lack of easy options for displaying equations in markdown though.

jedbrown:
It's slightly ugly/fragile, but equations work in Doxygen Markdown. http://www.doxygen.nl/manual/formulas.html
`doc/libCEEDapi.md` has junky workarounds, I think because it used to be in README.md and we wanted it to look alright rendered on GitHub.

We could also consider Sphinx for documentation. I think it looks a lot better than Doxygen, at least for everything but literal API documentation. There are tools like Breathe to get Doxygen into Sphinx, FWIW.

jedbrown:
Pre-1.0, I think it's fine to change. But at some point, they should become stable (possibly versioned) because people will depend on them and mention them by name in papers, etc.

jeremylt:
Diff renamed to Poisson.

gallery/diff3d/ceed-diff3DBuild.h
jeremylt: Added words to clarify

jeremylt:
What do you think about adding equations with the weak forms to the documentation as part of a separate PR to switch to something nicer like Sphinx?

jedbrown:
Yeah, separate PR is fine. Over on the PETSc project, we're looking to adopt Sphinx and would like to be able to generate docs similar to SciPy's, but there is code to be written to work with Sowing (PETSc's doc syntax system). There is Doxygen syntax support via [Breathe](https://breathe.readthedocs.io/en/latest/), but I have yet to see a C or C++ project with Breath-Sphinx docs that look like SciPy's.

* A tutorial page: https://docs.scipy.org/doc/scipy/reference/tutorial/optimize.html
* An API doc page: https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize

jeremylt:
I'll do it after this PR is merged, as there are a number of small doc fixes in here.

jeremylt:
I switched conventions for the BP3-6 qdata. I think we are good to merge.

include/ceed.h
```diff
@@ -27,19 +27,19 @@
 /// @defgroup CeedOperator CeedOperator: composed FE-type operations on vectors
 ///
 /// @page FunctionCategories libCEED: Types of Functions
-/// @subsection Types of Functions
+/// @section Library Functions
```
jedbrown: Should this one by `Library Library Functions`, or drop this level of nesting because it's captured in the @page header? I don't know how this is intended to be used.

Screenshot_20190911_161919

interface/ceed-qfunction.c
```diff
@@ -67,15 +79,83 @@ int CeedQFunctionCreateInterior(Ceed ceed, CeedInt vlength,
...
+
+  @ref Advanced
+**/
+int CeedQFunctionRegister(const char *name, const char *source,
+                          CeedInt vlength, CeedQFunctionUser f,
+                          int (*init)(Ceed, const char *, CeedQFunction)) {
+  if (num_qfunctions >= sizeof(qfunctions) / sizeof(qfunctions[0])) {
+    return CeedError(NULL, 1, "Too many gallery QFunctions");
+  }
+  strncpy(qfunctions[num_qfunctions].name, name, strlen(name)+1);
```
jedbrown: The length `n` is meant to be the length of the destination buffer (but `strncpy` is error-prone because it doesn't NULL terminate the result if there isn't enough space. You'll often see `strncpy(dest, src, dest_size); dest[dest_size-1] = 0;` to ensure a NULL-terminated result. BSDs provide `strlcpy` in `string.h`, but it isn't part of the standard. C11 offers `strcpy_s`, which is the interface you'd like to have, but it isn't standard.

examples/ceed/ex1.c
```diff
@@ -154,12 +175,25 @@ int main(int argc, const char *argv[]) {
   // Create the Q-function that builds the mass operator (i.e. computes its
   // quadrature data) and set its context data.
   CeedQFunction build_qfunc;
-  CeedQFunctionCreateInterior(ceed, 1, f_build_mass,
-                              f_build_mass_loc, &build_qfunc);
-  CeedQFunctionAddInput(build_qfunc, "dx", ncompx*dim, CEED_EVAL_GRAD);
-  CeedQFunctionAddInput(build_qfunc, "weights", 1, CEED_EVAL_WEIGHT);
-  CeedQFunctionAddOutput(build_qfunc, "rho", 1, CEED_EVAL_NONE);
-  CeedQFunctionSetContext(build_qfunc, &build_ctx, sizeof(build_ctx));
+  switch(1) {
```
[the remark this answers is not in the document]
jeremylt: I'll add a command line option.

examples/ceed/ex1.c
```diff
@@ -154,12 +175,25 @@ int main(int argc, const char *argv[]) {
...
+    case (1): {
+      // This creates the QFunction via the gallery.
+      char name[13] = "mass", buffer[2];
+      sprintf(buffer, "%d", dim);
+      strcat(name, buffer); strcat(name, "DBuild");
+      CeedQFunctionCreateInteriorByName(ceed, name, &build_qfunc);
```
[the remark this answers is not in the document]
jeremylt: We tend to use camelCase with variables, which is what I used initially, but I'm happy to switch to PascalCase.

include/ceed.h
jeremylt: I'm not sure what is going on there. I get the same issue when I build master documentation locally. This is a bit of a Doxygen hack to get 'libCEED provides three....' at the top of the page.

jeremylt:
WRT `make style`, I am similarly dissatisfied with the output. I end up having to revert some of the more annoying changes every time I run it. I'm not sure what to do on that front.

examples/ceed/ex1.c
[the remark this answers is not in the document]
jedbrown: I don't judge on local variable names, but our public interfaces all start with uppercase and I think of this as a sort of public interface.

include/ceed.h
jedbrown: Does it not work to put it as leading text under `@page`?
http://www.doxygen.nl/manual/commands.html#cmdpage

PR #313, opened 2019-08-21
NS example: Remove repeated CeedVectorDestroy

valeriabarra:
I've just realized that we were destroying the xceed vector twice. It was harmless, but since this example should serve as a tutorial, it's better to keep a correct form for new users to read.

### DIFF
--- a/examples/navier-stokes/navierstokes.c
+++ b/examples/navier-stokes/navierstokes.c
@@ -895,7 +895,6 @@ int main(int argc, char **argv) {
   CHKERRQ(ierr);
 
   // Clean up
-  CeedVectorDestroy(&xceed);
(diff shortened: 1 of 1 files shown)


examples/navier-stokes/navierstokes.c
```diff
@@ -895,7 +895,6 @@ int main(int argc, char **argv) {
   CHKERRQ(ierr);
 
   // Clean up
-  CeedVectorDestroy(&xceed);
```
jeremylt: Given my druthers, I'd rather destroy xceed and xcorners here after use like we do with q0ceed, multevec, and multlvec.

valeriabarra: Ok. Put it back there and also moved the other ones for onesvec and xcorners for consistency

PR #314, opened 2019-08-22
Update DoF to Node and Style Changes

jeremylt:
This PR
1. Updates references to DoF to Node for correctness
2. Adds `make style` updates


jeremylt:
@YohannDudouit, I still need to check that I didn't break anything when I changed the CUDA code, but this should be a simple change for references to NDoF to NNode.

valeriabarra:
Should we also update the examples suite?

jeremylt:
That's a good point. I can work on it Monday maybe?

jeremylt:
I updated ex1 and the PETSc examples (to include NS). I left Nek and MFEM examples as they are for now because they are BP1/BP3 scalar examples. I can update them if desired though, as DoF=Node in those cases.

PR #316, opened 2019-08-23
Update Codecov

jeremylt:
This PR excludes lines that are skipped in the tests because of intentional failures and checks that don't print in a passing test suite. Also, I stop doing CodeCov on OSX, as it's very slow there and doesn't add anything we aren't getting on the faster linux runs.

This PR is also faster because we generate the report once and submit it 4x rather than generating and submitting 4x.

I did not add exclusion markers to the interface or backends. I can see the benefit in adding markers to error messages we intend to not test, such as when a backend does not respond to a particular resource string, but that is a bigger discussion and I wanted this PR to be small.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,11 +73,13 @@ script:
     fi
 
 after_success:
-  - lcov --directory . --capture --output-file coverage.info
-  - bash <(curl -s https://codecov.io/bash) -F interface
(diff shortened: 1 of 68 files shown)


jeremylt:
CodeCov isn't posting the projected coverage, but it seems to think that it will go down?

PR #318, opened 2019-08-25
Update sign in advection example

jeremylt:
I believe we have a sign error in the unused version of the advection qfunction.

### DIFF
--- a/examples/navier-stokes/advection.h
+++ b/examples/navier-stokes/advection.h
@@ -135,45 +135,45 @@ static int Advection(void *ctx, CeedInt Q,
   for (CeedInt i=0; i<Q; i++) {
     // Setup
     // -- Interp in
-    const CeedScalar rho     =   q[i+0*Q];
-    const CeedScalar u[3]    = { q[i+1*Q] / rho,
-                                 q[i+2*Q] / rho,
-                                 q[i+3*Q] / rho
-                               };
-    const CeedScalar E       =   q[i+4*Q];
-    // -- Grad in
-    const CeedScalar drho[3] = {  dq[i+(0+5*0)*Q],
-                                  dq[i+(0+5*1)*Q],
-                                  dq[i+(0+5*2)*Q]
-                               };
-    const CeedScalar du[3][3] = {{(dq[i+(1+5*0)*Q] - drho[0]*u[0]) / rho,
-                                  (dq[i+(1+5*1)*Q] - drho[1]*u[0]) / rho,
-                                  (dq[i+(1+5*2)*Q] - drho[2]*u[0]) / rho},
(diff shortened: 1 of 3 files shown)


examples/navier-stokes/advection.h
```diff
@@ -210,7 +210,7 @@ static int Advection(void *ctx, CeedInt Q,
       dv[i+(4+0*5)*Q] = 0;
       dv[i+(4+1*5)*Q] = 0;
       dv[i+(4+2*5)*Q] = 0;
-      v[i+4*Q]   = E*(du[0]*wBJ[0] + du[3]*wBJ[1] + du[6]*wBJ[2]);
+      v[i+4*Q]   =-E*(du[0]*wBJ[0] + du[3]*wBJ[1] + du[6]*wBJ[2]);
```
jedbrown: I agree that this is needed. We might want to avoid using `-=` in the future because it breaks symmetry among terms. I'm not sure why not to create one multiline expression in this case.

jeremylt: I have no idea why we did it this way in the past, but I can change that easily.

jeremylt: Updated

examples/navier-stokes/advection.h
```diff
@@ -200,19 +200,19 @@ static int Advection(void *ctx, CeedInt Q,
...
     if (0) {
       dv[i+(4+0*5)*Q] = 0;
       dv[i+(4+1*5)*Q] = 0;
       dv[i+(4+2*5)*Q] = 0;
-      v[i+4*Q]   = E*(du[0]*wBJ[0] + du[3]*wBJ[1] + du[6]*wBJ[2]);
-      v[i+4*Q]  -= E*(du[1]*wBJ[3] + du[4]*wBJ[4] + du[7]*wBJ[5]);
-      v[i+4*Q]  -= E*(du[2]*wBJ[6] + du[5]*wBJ[7] + du[8]*wBJ[8]);
+      v[i+4*Q] = - (E*(du[0]*wBJ[0] + du[3]*wBJ[1] + du[6]*wBJ[2]) +
+                    E*(du[1]*wBJ[3] + du[4]*wBJ[4] + du[7]*wBJ[5]) +
+                    E*(du[2]*wBJ[6] + du[5]*wBJ[7] + du[8]*wBJ[8]);
```
jedbrown: Looks like `E` may as well come outside.

jeremylt: Right. Looking at the latest version, I think we very well might have done a copy-paste when we first did version 2 as a debugging check, left it pretty similar to version 1, and then never cleaned it up after adding the `if(0)`.

PR #323, opened 2019-08-28
NS: clean geometric factors

valeriabarra:
As per our conversation, it is cleaner to adopt a notation `x_i,j` (Jacobian `dx_i / dX_j` in indicial notation) for the change of coordinate matrix, and consequently, `X_i,j` for its inverse.

### DIFF
--- a/examples/navier-stokes/densitycurrent.h
+++ b/examples/navier-stokes/densitycurrent.h
@@ -222,66 +222,53 @@ static int DC(void *ctx, CeedInt Q,
   for (CeedInt i=0; i<Q; i++) {
     // Setup
     // -- Interp in
-    const CeedScalar rho       =    q[0][i];
-    const CeedScalar u[3]      =  { q[1][i] / rho,
+    const CeedScalar rho        =   q[0][i];
+    const CeedScalar u[3]       = { q[1][i] / rho,
                                     q[2][i] / rho,
                                     q[3][i] / rho
                                   };
-    const CeedScalar E         =    q[4][i];
+    const CeedScalar E          =   q[4][i];
     // -- Grad in
-    const CeedScalar drho[3]   =  { dq[0][0][i],
+    const CeedScalar drho[3]    = { dq[0][0][i],
                                     dq[1][0][i],
                                     dq[2][0][i]
                                   };
-    const CeedScalar du[3][3]  = {{(dq[0][1][i] - drho[0]*u[0]) / rho,
-                                   (dq[1][1][i] - drho[1]*u[0]) / rho,
-                                   (dq[2][1][i] - drho[2]*u[0]) / rho},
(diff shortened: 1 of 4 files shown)


examples/navier-stokes/densitycurrent.h
```diff
@@ -222,56 +222,53 @@ static int DC(void *ctx, CeedInt Q,
...
+                                   {qdata[7][i],
+                                    qdata[8][i],
+                                    qdata[9][i]}
                                   };
+    // -- Grad-to-Grad qdata
+    // ---- dXdx_j,k * dXdx_k,j
+    CeedScalar dXdxdXdxT[3][3];
+    for (int j=0; j<3; j++) {
+      for (int k=0; k<3; k++) {
+        dXdxdXdxT[j][k] = dXdx[j][k] * dXdx[k][j];
```
jedbrown: It's supposed to be matrix multiplication, so perhaps `dXdx[j][l] * dXdx[k][l]` (summed over `l`).

valeriabarra: Of course! Sorry about that!

valeriabarra:
> I added a commit with minor style consistency updates (though I don't feel super strongly about them) and a documentation update for `w detJ` vs only `detJ`. This looks good to me though.

I had aligned all the q's and dq's to start at the same column with a multi-row cursor

jeremylt:
I can add that back. I had them aligned with themselves  but not each other after realigning the equal signs.

examples/navier-stokes/densitycurrent.h
```diff
@@ -243,30 +243,34 @@ static int DC(void *ctx, CeedInt Q,
...
     // ---- dXdx_j,k * dXdx_k,j
     CeedScalar dXdxdXdxT[3][3];
-    for (int j=0; j<3; j++)
-      for (int k=0; k<3; k++)
-        dXdxdXdxT[j][k] = dXdx[j][k] * dXdx[k][j];
+      for (int j=0; j<3; j++) {
+        for (int k=0; k<3; k++) {
+          dXdxdXdxT[j][k] = 0;
+          for (int l=0; l<3; l++)
+            dXdxdXdxT[j][k] += dXdx[j][l]*dXdx[k][l];
```
valeriabarra: @jedbrown I did the vanilla implementation without loop-tiling. Do we want to do that? I wasn't sure

valeriabarra:
> I can add that back. I had them aligned with themselves but not each other after realigning the equal signs.

I realigned q, dq's and qdata

examples/navier-stokes/densitycurrent.h
jedbrown: We're expecting the compiler to unroll and transform all of this anyway. (Also, tiling is a cache optimization, but this is so small it fits in registers.)

PR #324, opened 2019-08-28
Doc - Add CUDA-Gen to Doxygen

jeremylt:


jeremylt:
Yeah, it just adds another folder in the file tree and everything seems to be in order to me.

PR #329, opened 2019-08-29
Fix QFunction Ncomp vs Size

jeremylt:
This changes ncomp to size for QFunction fields, fixing OUT_OF_RESOURCES errors we were having with #304.


backends/occa/ceed-occa-qfunction-op.c
```diff
@@ -152,6 +153,8 @@ int CeedQFunctionAllocOpOut_Occa(CeedQFunction qf, CeedInt Q,
       odx+=1;
       break;
     case CEED_EVAL_GRAD:
+      assert(dim>0);
```
jedbrown: Can we produce an error message instead of an `assert`?

jeremylt: I think a better fix is to change the initial value above from 0 to 1 and put the error check in the interface.

PR #333, opened 2019-09-03
Navier-Stokes: add support for periodic boundary conditions

jedbrown:


jedbrown:
These changes are contained in `examples/navier-stokes`. The Travis failures are unrelated.

examples/navier-stokes/navierstokes.c
```diff
@@ -880,6 +904,7 @@ int main(int argc, char **argv) {
   CeedVectorRestoreArray(multlvec, &mult);
 
   // Destroy mult vecs
```
valeriabarra: Looks good. Something very minor. I would update the comment here with `// Destroy nodes and mult vecs` or something along those lines

PR #334, opened 2019-09-03
Work around LIBXSMM bug

jeremylt:
In this PR I fix Travis to LIBXSMM commit 07e360a to prevent whatever memory management bug they introduced from causing CI failures.
See https://github.com/hfp/libxsmm/issues/348


jeremylt:
It looks like the commit in question is failing their CI system. If they sometimes commit to master builds that fail CI, as it appears given their GitHub, then perhaps we want to use a fixed commit or the latest release instead.

PR #335, opened 2019-09-04
Homogenize examples by using ncomp and dim

valeriabarra:
This very simple PR tries to homogenize the example suite by using variables `ncomp` and `dim` where appropriate. We had some discrepancies, among example files in `examples/ceed`, `examples/petsc` and `examples/mfem`.

### DIFF
--- a/examples/ceed/ex1.c
+++ b/examples/ceed/ex1.c
@@ -53,6 +53,7 @@ CeedScalar TransformMeshCoords(int dim, int mesh_size, CeedVector mesh_coords);
 int main(int argc, const char *argv[]) {
   const char *ceed_spec = "/cpu/self";
   int dim        = 3;           // dimension of the mesh
+  int ncomp      = 3;           // number of field components
   int mesh_order = 4;           // polynomial degree for the mesh
   int sol_order  = 4;           // polynomial degree for the solution
   int num_qpts   = sol_order+2; // number of 1D quadrature points
@@ -109,7 +110,7 @@ int main(int argc, const char *argv[]) {
 
   // Construct the mesh and solution bases.
   CeedBasis mesh_basis, sol_basis;
-  CeedBasisCreateTensorH1Lagrange(ceed, dim, dim, mesh_order+1, num_qpts,
+  CeedBasisCreateTensorH1Lagrange(ceed, dim, ncomp, mesh_order+1, num_qpts,
                                   CEED_GAUSS, &mesh_basis);
   CeedBasisCreateTensorH1Lagrange(ceed, dim, 1, sol_order+1, num_qpts,
                                   CEED_GAUSS, &sol_basis);
@@ -129,7 +130,7 @@ int main(int argc, const char *argv[]) {
   // representations.
   CeedInt mesh_size, sol_size;
   CeedElemRestriction mesh_restr, sol_restr, mesh_restr_i, sol_restr_i;
-  BuildCartesianRestriction(ceed, dim, nxyz, mesh_order, dim, &mesh_size,
+  BuildCartesianRestriction(ceed, dim, nxyz, mesh_order, ncomp, &mesh_size,
                             num_qpts, &mesh_restr, &mesh_restr_i);
   BuildCartesianRestriction(ceed, dim, nxyz, sol_order, 1, &sol_size,
                             num_qpts, &sol_restr, &sol_restr_i);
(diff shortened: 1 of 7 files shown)


examples/ceed/ex1.c
```diff
@@ -53,6 +53,7 @@ CeedScalar TransformMeshCoords(int dim, int mesh_size, CeedVector mesh_coords);
 int main(int argc, const char *argv[]) {
   const char *ceed_spec = "/cpu/self";
   int dim        = 3;           // dimension of the mesh
+  int ncomp      = 3;           // number of field components
```
jeremylt: I think we need to be careful here, as 3 is the number of components for the geometry, not for the problem being solved.

examples/mfem/bp1.hpp
```diff
@@ -115,7 +115,8 @@ class CeedMassOperator : public mfem::Operator {
     const int ir_order = 2*(order + 2) - 1; // <-----
     const mfem::IntegrationRule &ir =
       mfem::IntRules.Get(mfem::Geometry::SEGMENT, ir_order);
-    CeedInt nqpts, nelem = mesh->GetNE(), dim = mesh->SpaceDimension();
+    CeedInt nqpts, nelem = mesh->GetNE(), dim = mesh->SpaceDimension(),
```
jeremylt: For BP1/BP3 we should have ncomp = 1, ncomp_geo = 3. I like your use of ncompx below.

examples/navier-stokes/navierstokes.c
```diff
@@ -746,7 +747,7 @@ int main(int argc, char **argv) {
         }
       }
     }
-    CeedVectorCreate(ceed, len*3, &xcorners);
+    CeedVectorCreate(ceed, len*dim, &xcorners);
```
jeremylt: len*ncompx?

examples/navier-stokes/navierstokes.c
```diff
@@ -733,7 +734,7 @@ int main(int argc, char **argv) {
     CeedScalar *xloc;
     CeedInt shape[3] = {melem[0]+1, melem[1]+1, melem[2]+1}, len =
                          shape[0]*shape[1]*shape[2];
-    xloc = malloc(len*3*sizeof xloc[0]);
+    xloc = malloc(len*dim*sizeof xloc[0]);
```
jeremylt: len*ncompx?

examples/petsc/bps.c
```diff
@@ -597,7 +598,7 @@ int main(int argc, char **argv) {
     CeedScalar *xloc;
     CeedInt shape[3] = {melem[0]+1, melem[1]+1, melem[2]+1}, len =
                          shape[0]*shape[1]*shape[2];
-    xloc = malloc(len*3*sizeof xloc[0]);
+    xloc = malloc(len*dim*sizeof xloc[0]);
```
jeremylt: len*ncompx?

examples/petsc/bps.c
```diff
@@ -610,48 +611,48 @@ int main(int argc, char **argv) {
         }
       }
     }
-    CeedVectorCreate(ceed, len*3, &xcoord);
+    CeedVectorCreate(ceed, len*dim, &xcoord);
```
jeremylt: len*ncompx?

examples/ceed/ex1.c
valeriabarra: Right. I thought it could be ambiguous. I have renamed it `ncompx` as it is in the petsc/bps now

examples/petsc/bps.c
valeriabarra: Thank you Jeremy!

valeriabarra:
> Do we want to call out the fact that ncompx=dim in a comment?
> 
> Also, we should probably also add this to `examples/nek/bps/bps.usr`.

Not sure if the comment would add anything.
They already used `ndim` and `ncomp` in  `examples/nek/bps/bps.usr`.

jeremylt:
Good point on the comment.
wrt Nek, we aren't using ncomp on lines 814, 820, 824, 1130, 1136, and 1140 when we add qfunction fields (we are just using 1). I was also wondering if we wanted to use `ncompx` instead of `ldim` on lines 806, 808, 1122, and 1124 like do with the other examples in this PR.

valeriabarra:
> Good point on the comment.
> wrt Nek, we aren't using ncomp on lines 814, 820, 824, 1130, 1136, and 1140 when we add qfunction fields (we are just using 1). I was also wondering if we wanted to use `ncompx` instead of `ldim` on lines 806, 808, 1122, and 1124 like do with the other examples in this PR.

Done. Thanks!

PR #338, opened 2019-09-10
Add CeedPragmaOMP to bps

valeriabarra:
This PR adds the CeedPragmaOMP preprocessor directive, that was included only in the NS-example before, to all BPs (excluded the MFEM ones for now).

### DIFF
--- a/examples/nek/bps/bps.usr
+++ b/examples/nek/bps/bps.usr
@@ -18,15 +18,6 @@ C> @file
 C> Mass and diffusion operators examples using Nek5000
 C TESTARGS -c {ceed_resource} -e bp1 -n 1 -b 4 -test
 
-C PragmaOMP preprocessor directive
-#ifndef CeedPragmaOMP
-#  ifdef _OPENMP
-#    define CeedPragmaOMP(a) !$omp a
-#  else
-#    define CeedPragmaOMP(a)
-#  endif
-#endif
-
 C-----------------------------------------------------------------------
       subroutine masssetupf(ctx,q,u1,u2,u3,u4,u5,u6,u7,
      $  u8,u9,u10,u11,u12,u13,u14,u15,u16,v1,v2,v3,v4,v5,v6,v7,v8,
@@ -45,7 +36,6 @@ C     Input: u1,u2,u3,q             Output: v1,v2,ierr
       real*8 jacmq
 
 C     Quadrature Point Loop
-CeedPragmaOMP(simd)
       do i=1,q
         a11=u2(i+q*0)
         a12=u2(i+q*3)
@@ -98,7 +88,6 @@ C     Input: u1,u2,q                Output: v1,ierr
       real*8 v1(q)
 
 C     Quadrature Point Loop
-CeedPragmaOMP(simd)
       do i=1,q
         v1(i)=u2(i)*u1(i)
       enddo
@@ -124,7 +113,6 @@ C     Input: u1,u2,u3,q             Output: v1,v2,ierr
       real*8 c(3),k(3)
 
 C     Quadrature Point Loop
-CeedPragmaOMP(simd)
       do i=1,q
         pi = 3.14159265358979323846
 
@@ -195,7 +183,6 @@ C     Input: u1,u2,q                Output: v1,ierr
       real*8 v1(3*q)
 
 C     Quadrature Point Loop
-CeedPragmaOMP(simd)
       do i=1,q
         v1(i+0*q)=
      $     u2(i+0*q)*u1(i)+u2(i+1*q)*u1(i+q)+u2(i+2*q)*u1(i+2*q)
--- a/examples/nek/bps/bps.usr
+++ b/examples/nek/bps/bps.usr
@@ -18,6 +18,16 @@ C> @file
 C> Mass and diffusion operators examples using Nek5000
 C TESTARGS -c {ceed_resource} -e bp1 -n 1 -b 4 -test
 
+C PragmaOMP preprocessor directive
+#ifndef CeedPragmaOMP
+#  ifdef _OPENMP
+#    define CeedPragmaOMP_(a) _Pragma(#a)
+#    define CeedPragmaOMP(a) CeedPragmaOMP_(omp a)
+#  else
+#    define CeedPragmaOMP(a)
+#  endif
+#endif
+
 C-----------------------------------------------------------------------
       subroutine masssetupf(ctx,q,u1,u2,u3,u4,u5,u6,u7,
      $  u8,u9,u10,u11,u12,u13,u14,u15,u16,v1,v2,v3,v4,v5,v6,v7,v8,
@@ -35,6 +45,8 @@ C     Input: u1,u2,u3,q             Output: v1,v2,ierr
       real*8 g11,g12,g13,g21,g22,g23,g31,g32,g33
       real*8 jacmq
 
+C     Quadrature Point Loop
+      CeedPragmaOMP(simd)
       do i=1,q
         a11=u2(i+q*0)
         a12=u2(i+q*3)
@@ -86,6 +98,8 @@ C     Input: u1,u2,q                Output: v1,ierr
       real*8 u2(q)
       real*8 v1(q)
 
+C     Quadrature Point Loop
+      CeedPragmaOMP(simd)
       do i=1,q
         v1(i)=u2(i)*u1(i)
       enddo
@@ -110,6 +124,8 @@ C     Input: u1,u2,u3,q             Output: v1,v2,ierr
       real*8 jacmq,scl
       real*8 c(3),k(3)
 
+C     Quadrature Point Loop
+      CeedPragmaOMP(simd)
       do i=1,q
         pi = 3.14159265358979323846
 
@@ -179,6 +195,8 @@ C     Input: u1,u2,q                Output: v1,ierr
       real*8 u2(6*q)
       real*8 v1(3*q)
 
+C     Quadrature Point Loop
+      CeedPragmaOMP(simd)
       do i=1,q
         v1(i+0*q)=
      $     u2(i+0*q)*u1(i)+u2(i+1*q)*u1(i+q)+u2(i+2*q)*u1(i+2*q)
--- a/examples/nek/bps/bps.usr
+++ b/examples/nek/bps/bps.usr
@@ -21,7 +21,7 @@ C TESTARGS -c {ceed_resource} -e bp1 -n 1 -b 4 -test
 C PragmaOMP preprocessor directive
 #ifndef CeedPragmaOMP
 #  ifdef _OPENMP
-#    define CeedPragmaOMP_(a) _Pragma(#a)
+#    define CeedPragmaOMP_(a) C$#a
 #    define CeedPragmaOMP(a) CeedPragmaOMP_(omp a)
 #  else
 #    define CeedPragmaOMP(a)
@@ -46,7 +46,7 @@ C     Input: u1,u2,u3,q             Output: v1,v2,ierr
       real*8 jacmq
 
 C     Quadrature Point Loop
-      CeedPragmaOMP(simd)
+CeedPragmaOMP(simd)
       do i=1,q
         a11=u2(i+q*0)
         a12=u2(i+q*3)
@@ -99,7 +99,7 @@ C     Input: u1,u2,q                Output: v1,ierr
       real*8 v1(q)
 
 C     Quadrature Point Loop
-      CeedPragmaOMP(simd)
+CeedPragmaOMP(simd)
       do i=1,q
         v1(i)=u2(i)*u1(i)
       enddo
@@ -125,7 +125,7 @@ C     Input: u1,u2,u3,q             Output: v1,v2,ierr
       real*8 c(3),k(3)
 
 C     Quadrature Point Loop
-      CeedPragmaOMP(simd)
+CeedPragmaOMP(simd)
       do i=1,q
         pi = 3.14159265358979323846
 
@@ -196,7 +196,7 @@ C     Input: u1,u2,q                Output: v1,ierr
       real*8 v1(3*q)
 
 C     Quadrature Point Loop
-      CeedPragmaOMP(simd)
+CeedPragmaOMP(simd)
       do i=1,q
         v1(i+0*q)=
      $     u2(i+0*q)*u1(i)+u2(i+1*q)*u1(i+q)+u2(i+2*q)*u1(i+2*q)
--- a/examples/nek/bps/bps.usr
+++ b/examples/nek/bps/bps.usr
@@ -21,8 +21,7 @@ C TESTARGS -c {ceed_resource} -e bp1 -n 1 -b 4 -test
 C PragmaOMP preprocessor directive
 #ifndef CeedPragmaOMP
 #  ifdef _OPENMP
-#    define CeedPragmaOMP_(a) C$#a
-#    define CeedPragmaOMP(a) CeedPragmaOMP_(omp a)
+#    define CeedPragmaOMP(a) !$omp a
 #  else
 #    define CeedPragmaOMP(a)
 #  endif
--- a/examples/petsc/bp1.h
+++ b/examples/petsc/bp1.h
@@ -38,7 +38,7 @@ CEED_QFUNCTION(SetupMass)(void *ctx, const CeedInt Q,
     true_soln[i] = sqrt(x[i]*x[i] + x[i+Q]*x[i+Q] + x[i+2*Q]*x[i+2*Q]);
 
     rhs[i] = rho[i] * true_soln[i];
-  }
+  } // End of Quadrature Point Loop
   return 0;
 }
 
@@ -49,8 +49,8 @@ CEED_QFUNCTION(Mass)(void *ctx, const CeedInt Q,
 
   // Quadrature Point Loop
   CeedPragmaOMP(simd)
-  for (CeedInt i=0; i<Q; i++) {
+  for (CeedInt i=0; i<Q; i++)
     v[i] = rho[i] * u[i];
-  }
+
   return 0;
 }
--- a/examples/petsc/bp1.h
+++ b/examples/petsc/bp1.h
@@ -28,7 +28,7 @@ CEED_QFUNCTION(SetupMass)(void *ctx, const CeedInt Q,
   CeedScalar *rho = out[0], *true_soln = out[1], *rhs = out[2];
 
   // Quadrature Point Loop
-  CeedPragmaOMP(simd)
+  CeedPragmaSIMD
   for (CeedInt i=0; i<Q; i++) {
     const CeedScalar det = (J[i+Q*0]*(J[i+Q*4]*J[i+Q*8] - J[i+Q*5]*J[i+Q*7]) -
                             J[i+Q*1]*(J[i+Q*3]*J[i+Q*8] - J[i+Q*5]*J[i+Q*6]) +
@@ -48,7 +48,7 @@ CEED_QFUNCTION(Mass)(void *ctx, const CeedInt Q,
   CeedScalar *v = out[0];
 
   // Quadrature Point Loop
-  CeedPragmaOMP(simd)
+  CeedPragmaSIMD
   for (CeedInt i=0; i<Q; i++)
     v[i] = rho[i] * u[i];
 
--- a/examples/petsc/bp1.h
+++ b/examples/petsc/bp1.h
@@ -17,15 +17,6 @@
 /// @file
 /// libCEED QFunctions for mass operator example using PETSc
 
-#ifndef CeedPragmaOMP
-#  ifdef _OPENMP
-#    define CeedPragmaOMP_(a) _Pragma(#a)
-#    define CeedPragmaOMP(a) CeedPragmaOMP_(omp a)
-#  else
-#    define CeedPragmaOMP(a)
-#  endif
-#endif
-
 #ifndef __CUDACC__
 #  include <math.h>
 #endif
--- a/examples/petsc/bp1.h
+++ b/examples/petsc/bp1.h
@@ -17,6 +17,15 @@
 /// @file
 /// libCEED QFunctions for mass operator example using PETSc
 
+#ifndef CeedPragmaOMP
+#  ifdef _OPENMP
+#    define CeedPragmaOMP_(a) _Pragma(#a)
+#    define CeedPragmaOMP(a) CeedPragmaOMP_(omp a)
+#  else
+#    define CeedPragmaOMP(a)
+#  endif
+#endif
+
 #ifndef __CUDACC__
 #  include <math.h>
 #endif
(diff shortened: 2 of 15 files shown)


valeriabarra:
@YohannDudouit could you please help me identify where to add this directive in the MFEM examples? It is supposed to sit in front of the quadrature point loop. Thanks!

jeremylt:
I added the Pragma to the MFEM BPs. Have we demonstrated a performance boost with/without this for the BPs?

valeriabarra:
> I added the Pragma to the MFEM BPs. Have we demonstrated a performance boost with/without this for the BPs?
> 
> It seems reasonable to add these to the Gallery QFs once we merge this.

Thanks Jeremy! I was looking in the hpp files (my IDE hadn't load the MFEM h files and that's why I could't find where to add it!). 

@jed has carefully inspected the vectorized instructions generated with these directives using `objdump -d --prefix-addresses`

examples/petsc/bp1.h
```diff
@@ -45,8 +47,10 @@ CEED_QFUNCTION(Mass)(void *ctx, const CeedInt Q,
   const CeedScalar *u = in[0], *rho = in[1];
   CeedScalar *v = out[0];
 
-  for (CeedInt i=0; i<Q; i++) {
+  // Quadrature Point Loop
+  CeedPragmaOMP(simd)
+  for (CeedInt i=0; i<Q; i++)
     v[i] = rho[i] * u[i];
-  }
+
```
jeremylt: I left the braces around the loop body for uniformity, though I suppose it does not make a big difference.

examples/petsc/bp3.h
```diff
@@ -76,13 +78,29 @@ CEED_QFUNCTION(Diff)(void *ctx, const CeedInt Q,
...
+                                      qd[i+4*Q],
+                                      qd[i+5*Q]
+                                     };
+
+    const CeedInt idx[3][3] = {{0, 1, 2}, {1, 3, 4}, {2, 4, 5}}; // symmetric matrix indices
+
+    for (int j=0; j<3; j++) {
+      vg[i+(0+j*3)*Q]  = 0;
+      for (int k=0; k<3; k++)
+        vg[i+(0+j*3)*Q] += uJ[k] * dXdxdXdxT[idx[j][k]];
```
jeremylt: Since the PETSc BPs are our benchmarks, I think it would be good to verify that this commit doesn't change our performance. I know it shouldn't, but it seems worth the peace of mind to double check.

examples/petsc/bp3.h
```diff
@@ -81,12 +81,27 @@ CEED_QFUNCTION(Diff)(void *ctx, const CeedInt Q,
...
+                                      qd[i+2*Q],
+                                      qd[i+3*Q],
+                                      qd[i+4*Q],
+                                      qd[i+5*Q]
+                                     };
+
+    const CeedInt idx[3][3] = {{0, 1, 2}, {1, 3, 4}, {2, 4, 5}}; // symmetric matrix indices
+
+    for (int j=0; j<3; j++) {
+      vg[i+(0+j*3)*Q]  = 0;
```
jeremylt: Change to `[i+Q*j]`?

examples/petsc/bp3.h
```diff
@@ -76,13 +78,30 @@ CEED_QFUNCTION(Diff)(void *ctx, const CeedInt Q,
...
   for (CeedInt i=0; i<Q; i++) {
-    const CeedScalar ug0 = ug[i+Q*0];
-    const CeedScalar ug1 = ug[i+Q*1];
-    const CeedScalar ug2 = ug[i+Q*2];
-    vg[i+Q*0] = qd[i+Q*0]*ug0 + qd[i+Q*1]*ug1 + qd[i+Q*2]*ug2;
-    vg[i+Q*1] = qd[i+Q*1]*ug0 + qd[i+Q*3]*ug1 + qd[i+Q*4]*ug2;
-    vg[i+Q*2] = qd[i+Q*2]*ug0 + qd[i+Q*4]*ug1 + qd[i+Q*5]*ug2;
-  }
+    // Read spatial derivatives of u
+    const CeedScalar uJ[3]        =  {ug[i+Q*0],
```
jeremylt: `uJ`? I'm not sure what J means. Jacobian?

examples/petsc/bp3.h
valeriabarra: I have compared the benchmarks. Here you can find the bp3 currently on Master:
https://bit.ly/2kdVguU
and the new version with the loop:
https://bit.ly/2kG7RqX
Same comparison for bp4, current Master version:
https://bit.ly/2kG7VXJ
And version with the loop:
https://bit.ly/2lN1l1W

valeriabarra: For small numbers of local elements, it seems a bit faster with the loop. For larger number of elements there seems to be only a bit of noise between the two versions

valeriabarra:
@jed I have inspected this version of the code with the disassembler but I'm not sure I had compiled with the same configuration you were testing it with.

examples/petsc/bp3.h
valeriabarra: Yup. It was more appropriate for BP4, where u has multiple components

examples/nek/bps/bps.usr
```diff
@@ -18,6 +18,16 @@ C> @file
...
 
+C PragmaOMP preprocessor directive
+#ifndef CeedPragmaOMP
+#  ifdef _OPENMP
+#    define CeedPragmaOMP_(a) _Pragma(#a)
+#    define CeedPragmaOMP(a) CeedPragmaOMP_(omp a)
+#  else
+#    define CeedPragmaOMP(a)
+#  endif
+#endif
```
jedbrown: I'm pretty sure Fortran won't be defining `_OPENMP`, so these will always be inactive. But if your Fortran does define that macro, you'll get a syntax error because Fortran doesn't understand `_Pragma`. OpenMP syntax is different for Fortran (directives go in comments, `!$omp simd` or `C$omp simd`).

examples/petsc/bp3.h
valeriabarra: Should I call it gradu or something else?

examples/petsc/bp4.h
```diff
@@ -86,30 +88,51 @@ CEED_QFUNCTION(Diff3)(void *ctx, const CeedInt Q,
...
   for (CeedInt i=0; i<Q; i++) {
-    // Component 1
-    const CeedScalar ug00 = ug[i+(0+0*3)*Q];
-    const CeedScalar ug01 = ug[i+(0+1*3)*Q];
-    const CeedScalar ug02 = ug[i+(0+2*3)*Q];
-    vg[i+(0+0*3)*Q] = qd[i+Q*0]*ug00 + qd[i+Q*1]*ug01 + qd[i+Q*2]*ug02;
-    vg[i+(0+1*3)*Q] = qd[i+Q*1]*ug00 + qd[i+Q*3]*ug01 + qd[i+Q*4]*ug02;
-    vg[i+(0+2*3)*Q] = qd[i+Q*2]*ug00 + qd[i+Q*4]*ug01 + qd[i+Q*5]*ug02;
+    // Read spatial derivatives of u components
+    const CeedScalar uJ[3][3]     = {{ug[i+(0+0*3)*Q],
```
jedbrown: Is there a mnemonic behind `uJ`? Should we call it `du` or `dudX` or something like that?

examples/petsc/bp4.h
```diff
@@ -86,30 +88,51 @@ CEED_QFUNCTION(Diff3)(void *ctx, const CeedInt Q,
...
+       for (int k=0; k<3; k++)
+         vg[i+(1+j*3)*Q] += uJ[1][k] * dXdxdXdxT[idx[j][k]];
+     }
+
+     // Component 3
+     for (int j=0; j<3; j++) {
+       vg[i+(2+j*3)*Q]  = 0;
+       for (int k=0; k<3; k++)
+         vg[i+(2+j*3)*Q] += uJ[2][k] * dXdxdXdxT[idx[j][k]];
+     }
```
jedbrown: Why three separate loops instead of nesting?

examples/petsc/bp3.h
valeriabarra: You can also see the plots for BP3 Master version:
http://bit.ly/Masterplot
Loop version:
http://bit.ly/LoopVersion

Same for BP4 Master version:
http://bit.ly/MasterBP4
Loop Version:
http://bit.ly/LoopVersionBP4

jedbrown:
The indirect indexing in bp4.h is problematic.
```
0000000000007950 <Mass3+0x30> vmovsd xmm0,QWORD PTR [r10+rax*8]
0000000000007956 <Mass3+0x36> mov    rcx,rax
0000000000007959 <Mass3+0x39> vmulsd xmm1,xmm0,QWORD PTR [r8+rax*8]
000000000000795f <Mass3+0x3f> vmovsd QWORD PTR [rdi+rax*8],xmm1
0000000000007964 <Mass3+0x44> vmulsd xmm1,xmm0,QWORD PTR [r9+rax*8]
000000000000796a <Mass3+0x4a> vmovsd QWORD PTR [rsi+rax*8],xmm1
000000000000796f <Mass3+0x4f> vmulsd xmm0,xmm0,QWORD PTR [r11+rax*8]
0000000000007975 <Mass3+0x55> vmovsd QWORD PTR [rdx+rax*8],xmm0
000000000000797a <Mass3+0x5a> inc    rax
000000000000797d <Mass3+0x5d> cmp    rbx,rcx
0000000000007980 <Mass3+0x60> jne    0000000000007950 <Mass3+0x30>
```

examples/petsc/bp4.h
valeriabarra: Right.

valeriabarra:
> We can delete redefining `CeedPragmaOMP` within examples/navier-stokes/.

Done. Thanks for the reminder!

jedbrown:
The missed vectorization is related to aliasing. I can make it go using `--param vect-max-version-for-alias-checks=200`, but that's undesirable. `restrict` qualifiers are not doing all they should be doing.

examples/petsc/bp4.h
valeriabarra: Since **u** is a vector in this case, I simply thought `uJ` would be the most natural notation since that represents all spatial derivatives of all components of **u**, hence a Jacobian. But I can change it, of course

valeriabarra:
> The missed vectorization is related to aliasing. I can make it go using `--param vect-max-version-for-alias-checks=200`, but that's undesirable. `restrict` qualifiers are not doing all they should be doing.

Ok, so should I leave the version with the indirect indexing?

examples/nek/bps/bps.usr
valeriabarra: Thanks! I have updated it now and checked that `gfortran -cpp -E -dM -fopenmp` gives `#define CeedPragmaOMP(a) CeedPragmaOMP_(omp a)` in the list of defines.
However, I cannot test it at the moment as I never got to install Nek.

jedbrown:
I'd say don't worry about it now. I have this variant:
```c
  CeedPragmaOMP(simd)
  for (CeedInt i=0; i<Q; i++) {
    // Read spatial derivatives of u components
    const CeedScalar uJ[3][3]     = {{ug[i+(0+0*3)*Q],
                                      ug[i+(0+1*3)*Q],
                                      ug[i+(0+2*3)*Q]},
                                     {ug[i+(1+0*3)*Q],
                                      ug[i+(1+1*3)*Q],
                                      ug[i+(1+2*3)*Q]},
                                     {ug[i+(2+0*3)*Q],
                                      ug[i+(2+1*3)*Q],
                                      ug[i+(2+2*3)*Q]}
                                    };
    // Read qdata
    const CeedScalar dXdxdXdxT[3][3] = {{qd[i+0*Q],
                                         qd[i+1*Q],
                                         qd[i+2*Q]},
                                        {qd[i+1*Q],
                                         qd[i+3*Q],
                                         qd[i+4*Q]},
                                        {qd[i+2*Q],
                                         qd[i+4*Q],
                                         qd[i+5*Q]}
                                       };

    for (int k=0; k<3; k++) // k = component
      for (int j=0; j<3; j++) // j = direction of vg
        vg[i+(k+j*3)*Q] = (uJ[0][k] * dXdxdXdxT[0][j] +
                           uJ[1][k] * dXdxdXdxT[1][j] +
                           uJ[2][k] * dXdxdXdxT[2][j]);
  } // End of Quadrature Point Loop
```
for which
```
/home/jed/usr/ccache/mpich/bin/mpicc  -std=c99 -fPIC -Wall -Wwrite-strings -Wno-strict-aliasing -Wno-unknown-pragmas -fstack-protector -fvisibility=hidden -O3 -march=native -ftree-vectorize -ffast-math  -fopenmp-simd -fopt-info-vec-missed --param vect-max-version-for-alias-checks=20 -fopt-info-vec-optimized -D__DIR__=\"/home/jed/src/libceed/examples/petsc/\" -I/home/jed/petsc/mpich-opt/include -I/home/jed/petsc/include -I/home/jed/src/libceed/include  -L/home/jed/petsc/mpich-opt/lib -L/home/jed/src/libceed/lib   -Wl,-rpath,/home/jed/petsc/mpich-opt/lib  -Wl,-rpath,/home/jed/src/libceed/lib /home/jed/src/libceed/examples/petsc/bps.c -o bps \
  -lpetsc -lceed  -lm
[...]
/home/jed/src/libceed/examples/petsc/bp4.h:93:3: missed: couldn't vectorize loop
/home/jed/src/libceed/examples/petsc/bp4.h:93:3: missed: number of versioning for alias run-time tests exceeds 20 (--param vect-max-version-for-alias-checks)
/home/jed/src/libceed/examples/petsc/bp4.h:35:3: missed: couldn't vectorize loop
/home/jed/src/libceed/examples/petsc/bp4.h:35:3: missed: number of versioning for alias run-time tests exceeds 20 (--param vect-max-version-for-alias-checks)
``
The line 93 "missed" optimization isn't accurate because
```
0000000000007b30 <Mass3+0x210> vmovupd ymm0,YMMWORD PTR [r8+r10*1]
0000000000007b36 <Mass3+0x216> vmulpd ymm1,ymm0,YMMWORD PTR [rcx+r10*1]
0000000000007b3c <Mass3+0x21c> vmovupd YMMWORD PTR [rdx+r10*1],ymm1
0000000000007b42 <Mass3+0x222> vmulpd ymm1,ymm0,YMMWORD PTR [r9+r10*1]
0000000000007b48 <Mass3+0x228> vmovupd YMMWORD PTR [rax+r10*1],ymm1
0000000000007b4e <Mass3+0x22e> vmulpd ymm0,ymm0,YMMWORD PTR [rbx+r10*1]
0000000000007b54 <Mass3+0x234> vmovupd YMMWORD PTR [rdi+r10*1],ymm0
0000000000007b5a <Mass3+0x23a> add    r10,0x20
0000000000007b5e <Mass3+0x23e> cmp    r10,r11
0000000000007b61 <Mass3+0x241> jne    0000000000007b30 <Mass3+0x210>
```
This is with gcc-9.1. Adding `restrict` qualifiers to the relevant pointers does not change the "missed" optimization. There is still a lot of code handling variants that I think shouldn't exist because `restrict` should prevent such aliasing. It would be useful to "reduce" this test case to a minimal example of a missed optimization (or diagnostic incorrectly saying it is missed), then file a bug report with gcc (https://gcc.gnu.org/bugs/) and LLVM (which fails to vectorize). But let's not let this hold up our merging of this PR.

**Comment by valeriabarra** [comment]
I like your variant! I was checking that with 3 nested loops we were actually getting a bit worse performance. I had thought of  defining `dXdxdXdxT[3][3]` as well, but I wasn't sure if it was ok wasting the allocation since we know it's symmetric in this case

**Comment by jedbrown** [comment]
It isn't a real "allocation" since these loops are all being unrolled and optimized.

**Comment by valeriabarra** [comment]
@jedbrown I have updated it with your version, but I had to modify it since I think you accidentally swapped the `uJ[*][k]` indices.

**Comment by valeriabarra** [comment]
@jed @jeremylt if you want to compare the performance plots, here is the latest version
http://bit.ly/LatestBP4
vs Master:
http://bit.ly/masterBP4

File: examples/nek/bps/bps.usr (hunk 5 above)
Comment (jedbrown) [reply examples/nek/bps/bps.usr]: It'll fail because the string is being quoted (needed for `_Pragma`). This test works for me in isolation:
```c
#ifndef CeedPragmaOMP
#  ifdef _OPENMP
#    define CeedPragmaOMP(a) !$omp a
#  else
#    define CeedPragmaOMP(a)
#  endif
#endif
```

examples/petsc/bp3.h
valeriabarra: Sorry that was just a bad commit that had slipped through

examples/nek/bps/bps.usr
valeriabarra: Fixed. Thanks!

jedbrown:
I notice a couple issues. One is that `-fopenmp-simd` does not define the `_OPENMP` macro, so this is probably doing nothing at present. The other is an apparent GCC bug in which `#pragma omp simd` actually _prevents_ vectorization. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91732

Until this bug is fixed (and it may take a while to be fixed, let alone propagate to users), perhaps we should just create `CeedPragmaSIMD`, which expands to `_Pragma("GCC ivdep")` on recent GCC, falling back to `_Pragma("omp simd")` for other vendors.

jedbrown:
I pushed the above-described change. Feel free to modify/squash. I observe
```
./bps -local 100000 -problem bp4 -benchmark
[...]                                       
  Performance:                                                                                 
    CG Solve Time                      : 0.230695 (0.230695) sec                               
    DoFs/Sec in CG                     : 2.77356 (2.77356) million                       
```
becoming (with this change)
```
$ ./bps -local 100000 -problem bp4 -benchmark                                                                      
[...]                                                                                                                                   
  Performance:                                                                                                                                                                                 
    CG Solve Time                      : 0.214863 (0.214863) sec                                                                                                                               
    DoFs/Sec in CG                     : 2.97792 (2.97792) million     
```
and assembly of this variety:
```
             0000000000008eeb <Diff3+0x1fb> vfmadd231pd ymm12,ymm1,ymm10
0000000000008ef0 <Diff3+0x200> vfmadd231pd ymm12,ymm9,ymm2
0000000000008ef5 <Diff3+0x205> vmovupd YMMWORD PTR [rdi+rax*1],ymm12
0000000000008efa <Diff3+0x20a> vmulpd ymm12,ymm11,ymm7
0000000000008efe <Diff3+0x20e> vmulpd ymm11,ymm11,ymm3
0000000000008f02 <Diff3+0x212> vmulpd ymm7,ymm6,ymm7
0000000000008f06 <Diff3+0x216> mov    rdi,QWORD PTR [rsp-0x40]
0000000000008f0b <Diff3+0x21b> vfmadd231pd ymm12,ymm10,ymm0
0000000000008f10 <Diff3+0x220> vfmadd132pd ymm10,ymm11,ymm2
0000000000008f15 <Diff3+0x225> vfmadd231pd ymm12,ymm9,ymm3
0000000000008f1a <Diff3+0x22a> vfmadd132pd ymm9,ymm10,ymm8
0000000000008f1f <Diff3+0x22f> vmovupd YMMWORD PTR [rdi+rax*1],ymm12
0000000000008f24 <Diff3+0x234> mov    rdi,QWORD PTR [rsp-0x30]
0000000000008f29 <Diff3+0x239> vmovupd YMMWORD PTR [rdi+rax*1],ymm9
0000000000008f2e <Diff3+0x23e> vmulpd ymm9,ymm6,ymm0
0000000000008f32 <Diff3+0x242> vfmadd132pd ymm0,ymm7,ymm5
0000000000008f37 <Diff3+0x247> vfmadd132pd ymm1,ymm9,ymm5
0000000000008f3c <Diff3+0x24c> vfmadd231pd ymm0,ymm4,ymm3
0000000000008f41 <Diff3+0x251> vmulpd ymm3,ymm6,ymm3
0000000000008f45 <Diff3+0x255> vfmadd231pd ymm1,ymm4,ymm2
0000000000008f4a <Diff3+0x25a> vfmadd132pd ymm2,ymm3,ymm5
0000000000008f4f <Diff3+0x25f> vmovupd YMMWORD PTR [r14+rax*1],ymm1
0000000000008f55 <Diff3+0x265> mov    rdi,QWORD PTR [rsp-0x38]
0000000000008f5a <Diff3+0x26a> vfmadd132pd ymm4,ymm2,ymm8
0000000000008f5f <Diff3+0x26f> vmovupd YMMWORD PTR [rdi+rax*1],ymm0
0000000000008f64 <Diff3+0x274> vmovupd YMMWORD PTR [r15+rax*1],ymm4
0000000000008f6a <Diff3+0x27a> add    rax,0x20
0000000000008f6e <Diff3+0x27e> cmp    rax,QWORD PTR [rsp+0x20]
0000000000008f73 <Diff3+0x283> jne    0000000000008e30 <Diff3+0x140>
```
versus scalar code otherwise (including with `omp simd`, due to the aforementioned gcc bug).

valeriabarra:
Thank you very much!

examples/petsc/bp3.h
```diff
@@ -31,7 +31,7 @@ CEED_QFUNCTION(SetupDiff)(void *ctx, const CeedInt Q,
   CeedScalar *qd = out[0], *true_soln = out[1], *rhs = out[2];
 
   // Quadrature Point Loop
-  CeedPragmaOMP(simd)
+  #pragma omp simd
```
valeriabarra: Hey @jedbrown why using this version, different from `CeedPragmaSIMD`?

jedbrown: My mistake; pushed an amended commit.

examples/petsc/bp3.h
```diff
@@ -75,33 +75,33 @@ CEED_QFUNCTION(SetupDiff)(void *ctx, const CeedInt Q,
 
 CEED_QFUNCTION(Diff)(void *ctx, const CeedInt Q,
                      const CeedScalar *const *in, CeedScalar *const *out) {
-  const CeedScalar *ug = in[0], *qd = in[1];
-  CeedScalar *vg = out[0];
+  const CeedScalar *restrict ug = in[0], *restrict qd = in[1];
```
valeriabarra: @jedbrown do we want to keep these `restrict` qualifiers around, even though you've seen they are ineffective? If that is the case, should I homogenize it for all other examples?

jedbrown: Let's remove them since C++ and CUDA (https://devblogs.nvidia.com/cuda-pro-tip-optimize-pointer-aliasing/) use `__restrict__` and it seems to be unnecessary with these pragmas.

valeriabarra: Removed them.

examples/nek/bps/bps.usr
valeriabarra: @jed I have now updated the Nek example to the SIMD version. (I have found this reference https://gcc.gnu.org/onlinedocs/gfortran/IVDEP-directive.html).
Again, I cannot test this at the moment. I think it should be good.

PR #339, opened 2019-09-10
Fix postprocessing script

valeriabarra:
This quick PR fixes the `post-process-base.py` for plotting, after PR #314

### DIFF
--- a/benchmarks/postprocess-base.py
+++ b/benchmarks/postprocess-base.py
@@ -102,7 +102,7 @@ while True:
       elif 'Global nodes' in line:
          data['num_unknowns']=int(line.split(':')[1])
          if data['case']=='vector':
-            data['num_unknowns']/=3
+            data['num_unknowns']*=3
       ## Number of elements
       elif 'Local Elements' in line:
          data['num_elem']=int(line.split(':')[1].split()[0])*data['num_procs']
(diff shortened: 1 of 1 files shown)


jeremylt:
We went wrong somewhere, as the output for plotting should be DoFs/Sec, where DoFs are given by nnodes*(dofs/node), so nnodes for BP 1, 3, 5 and 3*nnodes for BP 2, 4, 6. Nodes/Sec doesn't make sense for BP 2, 4, 6.

jeremylt:
Ah, I see. We need to modify 'Global nodes' to account for DoFs/Node. This involves changing more in the Python script, but these two changes are needed as well.

benchmarks/postprocess-base.py
```diff
@@ -101,6 +101,8 @@
       ## Total DOFs
       elif 'Global nodes' in line:
          data['num_unknowns']=int(line.split(':')[1])
+         if data['case']=='vector':
+            data['num_unknowns']/=3
```
valeriabarra: @jeremy is this what you meant?

jeremylt: The other way around, I think. `num_unknowns` comes from the line giving `Global nodes`, and in the vector case `num_unknowns` is 3x larger than `Global nodes`, as `Global nodes` is `gsize/ncompu` and we really want `gsize`, the total number of DoFs we are solving for.

valeriabarra: I think I now got what you meant. Thanks!

PR #341, opened 2019-09-13
Improved performance of cuda-gen backend

YohannDudouit:
These improvements on the cuda-gen backend are the result of the joint work with @tcew inspired by the optimizations from libParanumal. This mostly improves the performance of BP3.


backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -659,6 +674,38 @@ inline __device__ void gradTranspose3d(BackendData& data, const CeedScalar *__re
   }
 }
 
+template <int NCOMP, int P1d, int Q1d>
+inline __device__ void gradColo3d(BackendData& data, const CeedScalar *__restrict__ r_U,
```
valeriabarra: Not to be pedantic, but we changed all the `colocated` that were misspelled, to `collocated`, a while ago. This would re-introduce it to the library

backends/cuda-gen/ceed-cuda-gen.h
```diff
@@ -33,17 +33,17 @@ typedef struct {
...
-  CUfunction interp;
-  CUfunction grad;
-  CUfunction weight;
-  CeedScalar *d_interp1d;
-  CeedScalar *d_grad1d;
-  CeedScalar *d_qweight1d;
-  CeedScalar *c_B;
-  CeedScalar *c_G;
-} CeedBasis_Cuda_gen;
+// typedef struct {
```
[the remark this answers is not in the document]
YohannDudouit: That's a good point, I can do a some cleaning in the same time.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
YohannDudouit: Thank you, I'll correct the typo.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -1106,7 +1106,7 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
         break;
       case CEED_EVAL_GRAD:
         code << "  CeedScalar r_q"<<i<<"[ncomp_in_"<<i<<"*Dim];\n";
-        //TODO put the following in colograd3d
+        //TODO put the following in collograd3d
```
valeriabarra: Is this still TODO or the comment is outdated?

YohannDudouit: Done.

tcew:
Is there anything in particular I should look at? 

Is the performance at the levels we saw during your visit ?

YohannDudouit:
Before:
[MFEM_bp3_ceed-cuda-gen_lassen_gcc](https://github.com/CEED/libCEED/files/3617850/plot2_MFEM_ceed-cuda_bp3_lassen_gcc_N001_pn1.pdf)

After:
[MFEM_bp3_ceed-cuda-gen-new_lassen_gcc](https://github.com/CEED/libCEED/files/3617848/plot2_MFEM_ceed-cuda-new_bp3_lassen_gcc_N001_pn1.pdf)

tzanio:
Will this make it in libCEED-0.5?

YohannDudouit:
@tcew Looking at the results in my previous post it looks like the performance are on par with libParanumal. I read in the result table you did that I had a bit lower performance than in these plots (which is a good news). As you can see I pushed up to order 14, I imagine that your approach might run out of shared memory at lower order.
One interesting point of comparison could be to know if libParanumal is reaching the asymptotic regime faster than libCEED/cuda-gen, which would give an edge to your implementation for "strong scalability". As @v-dobrev was observing, the 0.8 performance is at approximately 1 million dofs for cuda-gen, which limits "strong scalability".

jeremylt:
Passes on CU's machine, and passes merged with master.

That's a nice performance boost.

Currently it doesn't support P<Q, as you mentioned. Does this do something sensible for non-tensor bases? Or would there be a future modification to support 3 (maybe 4) cases
- Tensor basis, P<=Q w/collocated
- Tensor basis, P>Q w/o collocated
- Non-tensor basis
- Tensor basis, P=Q and Lobatto quadrature (BP5/6) (Its very special case. We added it for CPU, but that was an easy modification.)

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -500,12 +518,17 @@ inline __device__ void writeQuadsTranspose3d(BackendData& data, const CeedInt nq
 template <int NCOMP, int P1d, int Q1d>
 inline __device__ void ContractX3d(BackendData& data,
                                    const CeedScalar *U, const CeedScalar *B, CeedScalar *V) {
+  CeedScalar r_B[P1d];
+  for (int i = 0; i < P1d; ++i)
+  {
```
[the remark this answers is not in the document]
jeremylt: @YohannDudouit, if you run `make style`, you'll want to only keep the changes for cuda-gen. I ran and merged a commit with `make style` last week and did the fun process of manually reverting the 'bad' style updates for the rest of the code.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -125,7 +125,7 @@ static int CeedOperatorApply_Cuda_gen(CeedOperator op, CeedVector invec,
     ierr = CeedRunKernelDimSharedCuda(ceed, data->op, grid, Q1d, Q1d, elemsPerBlock,
                                       sharedMem, opargs);
   } else if (dim==3) {
-    const CeedInt elemsPerBlock = Q1d<8? 4 : 1;
+    const CeedInt elemsPerBlock = Q1d<6? 4 : (Q1d<8? 2 : 1);
```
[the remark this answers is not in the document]
YohannDudouit: I'm not sure. This heuristic was found experimentally on BP3, it could certainly be improved for complex operators later. The idea is mostly to have a number of active threads per block as close as possible to a warp size (multiple of 32), not so much about the amount of resources used per thread. The number of components mostly impact the number of registers used by each threads, so I'm not sure it's relevant here. It would be relevant if we could choose between different implementations (that use more or less registers, shared mem, etc), but I only have one implementation in cuda-gen for now. I thought about unifying all cuda backends so that I can use the code in cuda-gen with any of the previous implementation (it's mostly using the functions from cuda-shared reformated for cuda-gen). This would require time with very uncertain interest, so it's not very high in my priority list for the moment.

YohannDudouit:
@jedbrown None of the cuda backend handles the case `Q<P`. I have to do a PR to fix this, it's not too difficult, but it impacts many parts of the code.

YohannDudouit:
@jeremylt I made sure that Q>P works without collocated gradient. This backend doesn't handle non-tensor basis for the moment, I will do a PR to fix this in the near future. I also don't treat BP5/6 as a special case for the moment.

jeremylt:
Yes, the PETSc tests passed for me.

PR #342, opened 2019-09-13
Make examples consistent

valeriabarra:
We had several inconsistencies in our examples suite. Some used `qd` for qdata (but we have used the prefix/suffix `d` for derivatives in other examples), some used `rho`. Some used `w` for quadrature weights, some used `qw`.
I tried to clean up a bit.


valeriabarra:
I didn't mean to push commits `85974f1`, `b100d62`, and `d249df7` from master in.

jeremylt:
Rebasing against origin/master should remove those commits.

Issue #199, opened 2019-01-28
Linking BLAS


jeremylt:
On CU Summit using the Intel compilers, the libXSMM backend requires linking against MKL instead of BLAS. Currently this requires a manual Makefile edit.

alheinecke:
you can build LIBXSMM with BLAS=0 then this should go away, of course when using libxsmm_dgemm and JIT is failing, then there is no longer a call.

@hfp if this is bad, then we should add 3 nested loops.

Alex

hfp:
> libXSMM backend requires linking against MKL instead of BLAS

Let me restate, what you wrote:

1. Linking any BLAS (for instance OpenBLAS) is not "good enough" (MKL seems a hard dependency).
2. You did not intend to remove BLAS as a dependency (which is what Alex suggested aka `BLAS=0`).

Can you confirm?  If (1) is true, it is a bug and I will definitely fix it i.e., we want LIBXSMM to have no dependency a specific LAPACK/BLAS library (any BLAS library should be fine).  So, (2) is not our intended way to fix (1) but allows to remove LAPACK/BLAS dependency if a fallback is not needed.

jeremylt:
To clarify, I don't see this as a LIBXSMM issue. For whatever reason, CU Summit's linking against BLAS isn't working quite right so I'm linking against MKL instead. 

I'd prefer not to specify to our users any particular way to make libraries that our backends may use, so I'd prefer not to insist that our users make LIBXSMM with `BLAS=0`.

I think this is a libCEED Makefile fix where I link against BLAS *or* MKL. I don't think a fallback is needed, so it doesn't change our performance at all but rather the ease of building our software.

hfp:
Thanks!

Here are some comments:

* `MKL=0` can force using another LAPACK/BLAS library (LIBXSMM's Makefile is supposed to work with MKL, OpenBLAS, or RefLapack/RefBlas with preference in the order listed). If RefBLAS is preferred over OpenBLAS, `OPENBLAS=0` is similiar to `MKL=0`. There are more keys to help picking up BLAS on systems with more specific directory layout. This is only about convenience with out-of-the-box builds, libxsmm actually relies only on `dgemm_` and `sgemm_`symbols (not a specific BLAS implementation).
[...]

PR #352, opened 2019-09-17
Makefile - drop BLAS link for XSMM

jeremylt:
This fixes Issue #199 by dropping the unneeded BLAS link from the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,14 +290,7 @@ endif
 # libXSMM Backends
 ifneq ($(wildcard $(XSMM_DIR)/lib/libxsmm.*),)
   $(libceeds) : LDFLAGS += -L$(XSMM_DIR)/lib -Wl,-rpath,$(abspath $(XSMM_DIR)/lib)
-  $(libceeds) : LDLIBS += -lxsmm -ldl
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -290,14 +290,7 @@ endif
 # libXSMM Backends
 ifneq ($(wildcard $(XSMM_DIR)/lib/libxsmm.*),)
   $(libceeds) : LDFLAGS += -L$(XSMM_DIR)/lib -Wl,-rpath,$(abspath $(XSMM_DIR)/lib)
-  $(libceeds) : LDLIBS += -lxsmm -ldl
-  MKL ?= 0
```
jedbrown: Perhaps instead of (or in addition to) the `MKL` boolean, we should respond to `MKLROOT` being set.

jeremylt: That would work. We need the BLAS/MKL fallback for performance reasons, so we can't simply use libxsmmnoblas.pc as discussed in the issue.

jeremylt: Updated. Now we *should* detect MKLROOT and link against MKL if installed and BLAS otherwise.

PR #356, opened 2019-09-18
Codecov Exceptions

jeremylt:
This PR adds more Codecov exceptions so we can more easily see what we are not testing rather than highlighting lines we do not intend to test.


jeremylt:
Ready for review. I'll add a later PR to hit some of the lines we are missing with new tests.

PR #360, opened 2019-09-19
Jed/config flags and string processing

jedbrown:


interface/ceed.c
```diff
@@ -519,11 +519,9 @@ int CeedSetObjectDelegate(Ceed ceed, Ceed delegate, const char *objname) {
 
   // Set object delegate
   ceed->objdelegates[count].delegate = delegate;
-  ierr = CeedCalloc(strlen(objname)+1, &ceed->objdelegates[count].objname);
-  CeedChk(ierr);
-  CeedInt len = strlen(ceed->objdelegates[count].objname);
```
jedbrown: I think this must have been a legit bug. The calloc above would result in `len = 0` (maybe you meant to `strlen(objname)` instead?), which then copies nothing below.

jeremylt: Ah, yes. I was, wondering what why the delegation wasn't correct in CodeCov. Thanks

jeremylt: I'm having trouble seeing if lines 478-480 of ceed.c are now covered in this branch.

jedbrown: It is covered by `t300-basis`, for example.

PR #371, opened 2019-10-02
Minor MAGMA Updates

jeremylt:
I made three small changes
1) `make style`
2) use switch in `ceed-magma-basis.c`
3) drop unused `magma_dtensor_contract.c`

I still need to test, but I am open to more changes to bring this backend more in line with the style of the other backends.


jeremylt:
Passes on CU's machine. I think there might be more unused functions and code we can remove.

PR #306, opened 2019-08-07
Assemble Linear QFunction

jeremylt:
~This WIP PR adds assembly of the QFunction.

ToDo:
- [x] QFunction in ref
- [x] QFunction diagonal in ref
- [x] Nonlinear QFs
- [x] Refactor into operator code
~- [ ] Dispatch to ref~
- [x] Tests

This is an intermediate step in supporting preconditioners.~
This will be reworked following the efforts in #304. Also, I will focus on assembling linear QFunctions for now to support next month's talks.

PR #372, opened 2019-10-02
FDM Element Inverse

jeremylt:
WIP

This is downstream of #306


interface/ceed-operator.c
```diff
@@ -248,6 +248,166 @@ int CeedCompositeOperatorAddSub(CeedOperator compositeop,
...
+                          for each element
+  @param[out] qdata     CeedVector to hold qdata for fdminv
+  @param request        Address of CeedRequest for non-blocking completion, else
+                          CEED_REQUEST_IMMEDIATE
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+int CeedOperatorCreateFDMElementInverse(CeedOperator op, CeedOperator *fdminv,
```
jeremylt: @jedbrown, this is a WIP downstream of PR #306, but this is most of the code needed to provide per element inverses that we can (ideally) plug into @stefanozampini's branch. The exciting bit that will actually take some math is lines 374-377 below. I am going to start with a simple average of the QFunction values with qweight divided out and go from there.

interface/ceed-operator.c
```diff
@@ -248,6 +248,166 @@ int CeedCompositeOperatorAddSub(CeedOperator compositeop,
...
+  ierr = CeedFree(&grad); CeedChk(ierr);
+
+  // Assemble QFunction
+  CeedVector assembled;
+  CeedElemRestriction rstr;
+  ierr =  CeedOperatorAssembleLinearQFunction(op, &assembled, &rstr, request);
+  CeedChk(ierr);
+  ierr = CeedElemRestrictionDestroy(&rstr); CeedChk(ierr);
+
+  // Calculate element averages
```
jedbrown: Suppose I have a target grid function `f[i][j][k]` that I would like to approximate via the product `x[i]*y[j]*z[k]` (perhaps with each of those factors positive). We can defer for now, but it would be cool to "solve" this problem. I think doing so would be useful in boundary layers, for example.

jeremylt: Right, I would like to do much better than a simple average here. The average gets us to where we can run simple problems on simple meshes, but I plan on developing more sophisticated diagonals to handle more complex problems. My next step will be taking `lambda` and changing `I kron I kron lambda + I kron lambda kron I + lambda kron I kron I` (for 3D Laplacian) to  `I kron I kron lambda_z + I kron lambda_y kron I + lambda_x kron I kron I` to handle variation along reference axes.

jeremylt:
With #375 ready for review, I'll be able to pick this back up next week. With the groundwork in that branch, this should only be a day or two of work.

jeremylt:
I need to flesh out the test and verify correctness, but this is working now.

jeremylt:
Once I get this working, I'll work on revamping OpApply in Ref so we can delegate AssembleLinearQF, AssembleLinearDiag, and CreateFDMElementInverse though Ref to provide some initial capability to GPU backends.

interface/ceed-basis.c
```diff
@@ -639,8 +643,13 @@ int CeedSymmetricSchurDecomposition(Ceed ceed, CeedScalar *mat,
...
     //   tau = 2 / (norm*norm)
-    tau[i] = 2 * v[i]*v[i] / (v[i]*v[i] + sigma);
-    for (CeedInt j=i+1; j<n-1; j++) v[j] /= v[i];
+    if (fabs(sigma) > 1e-16) {
+      tau[i] = 2 * v[i]*v[i] / (v[i]*v[i] + sigma);
+      for (CeedInt j=i+1; j<n-1; j++)
+        v[j] /= v[i];
+    } else {
+      tau[i] = 0;
+    }
```
jeremylt: @jedbrown, with denationalization of mass and stiffness matrices we get division my zero sometimes here. I'm switching to the Householder vector computation in Golub and van Loan which checks for that division by zero, but I'm probably missing something small because a few matrix entries are off by a sign.

interface/ceed-basis.c
```diff
@@ -634,13 +634,21 @@ int CeedSymmetricSchurDecomposition(Ceed ceed, CeedScalar *mat,
       sigma += v[j] * v[j];
     }
     CeedScalar norm = sqrt(v[i]*v[i] + sigma); // norm of v[i:n-1]
-    CeedScalar Rii = -copysign(norm, v[i]);
-    v[i] -= Rii;
+    CeedScalar Rii = copysign(norm, v[i]);
```
jeremylt: This `Rii` does not seem to be present in Golub and van Loan's version, but removing it and substituting their use of `norm` into the code below produces results off by more than a sign error. Substituting these changes as is with the use of `Rii` into QR above yields very wrong results, so this can't be correct.

jedbrown: I lack context at the moment, but Householder prefers the "further" projection, which I think is the prior code (with the negative).
TB-Householder2

jeremylt:
Fixed and it works!
I need a Fortran test and then we can review this initial version.

jeremylt:
It looks like Travis has an error because the GCC and Clang runs for t305 have equivalent but different factorizations. Rude.

jeremylt:
Fixed the 305 issue - X may be different, but X X^T = M^-1is unique.

interface/ceed-operator.c
```diff
@@ -503,6 +504,195 @@ int CeedOperatorAssembleLinearDiagonal(CeedOperator op, CeedVector *assembled,
...
+  @brief Build a FDM based approximate inverse for each element for a
+           CeedOperator
+
+  This returns a CeedOperator and CeedVector to apply a Fast Diagonalization
+    Method based approximate inverse. This function obtains the simultaneous
+    diagonalization for the 1D mass and Laplacian operators,
+      M = V^T V, K = V^T S V.
+    The assembled QFunction is used to modify the eigenvalues from simultaneous
+    diagonalization and obtain an approximate inverse of the form
+      V^T S^hat V.
```
jedbrown: What is assumed about the input CeedOperator? Does the QFunction matter? Can it be composite? I assume there are a few assumptions.

tests/output/t302-basis-f.out
```diff
@@ -1,11 +1,11 @@
...
-collograd[2]:  0.32869822 -1.47067023  0.38441439  0.75755761
-collograd[3]: -0.58062817  2.10878235 -4.86015442  3.33200024
+collograd[0]:  3.00000000 -4.04508497  1.54508497 -0.50000000
+collograd[1]:  0.80901699  0.00000000 -1.11803399  0.30901699
+collograd[2]: -0.30901699  1.11803399  0.00000000 -0.80901699
+collograd[3]:  0.50000000 -1.54508497  4.04508497 -3.00000000
+collograd[0]: -2.97732648  3.61240959  0.42440915 -4.17292696
+collograd[1]: -0.88463145  0.06263227  0.56306894  1.37442322
+collograd[2]:  0.34431271 -1.52560211  0.49593798  0.54828267
+collograd[3]:  0.13546013 -0.41042147  0.25437347 -6.26545429
```
jedbrown: Maybe I missed it, but why these changes?

jeremylt: There was a divide by zero showing up, so I had to tweak the simultaneous diagonalization to make it more stable.

interface/ceed-operator.c
jeremylt: Right, for now it's non-composite operators with linear qfunctions. I'll document that. I plan on relaxing that in the future.

jedbrown:
Values changed a lot (opposite sign) so please confirm that they are correct. I'd think sign matters for a collocated gradient.On Nov 24, 2019 07:53, Jeremy L Thompson <notifications@github.com> wrote:@jeremylt commented on this pull request.



In tests/output/t302-basis-f.out:
> -collograd[0]: -3.00000000  4.04508497 -1.54508497  0.50000000
-collograd[1]: -0.80901699  0.00000000  1.11803399 -0.30901699
-collograd[2]:  0.30901699 -1.11803399  0.00000000  0.80901699
-collograd[3]: -0.50000000  1.54508497 -4.04508497  3.00000000
-collograd[0]: -3.33200024  4.86015442 -2.10878235  0.58062817
-collograd[1]: -0.75755761 -0.38441439  1.47067023 -0.32869822
-collograd[2]:  0.32869822 -1.47067023  0.38441439  0.75755761
-collograd[3]: -0.58062817  2.10878235 -4.86015442  3.33200024
+collograd[0]:  3.00000000 -4.04508497  1.54508497 -0.50000000
+collograd[1]:  0.80901699  0.00000000 -1.11803399  0.30901699
+collograd[2]: -0.30901699  1.11803399  0.00000000 -0.80901699
+collograd[3]:  0.50000000 -1.54508497  4.04508497 -3.00000000
+collograd[0]: -2.97732648  3.61240959  0.42440915 -4.17292696
+collograd[1]: -0.88463145  0.06263227  0.56306894  1.37442322
+collograd[2]:  0.34431271 -1.52560211  0.49593798  0.54828267
+collograd[3]:  0.13546013 -0.41042147  0.25437347 -6.26545429

There was a divide by zero showing up, so I had to tweak the simultaneous diagonalization to make it more stable.

jeremylt:
Yeah, I'll double check again. The BPs and the rest of the test suite are happy, with all of the CPU backends using the calculated collocated gradient. I think I had a theory on why this was the case, but I don't recall it right now.

jedbrown:
The sign cancels if you take a second derivative (D^T K D), so make sure the sign is tested via a first derivative.

jeremylt:
Right. We have a test that checks that if we set the values at nodes with some polynomial f(x), then D u = v gives the values at f'(x) in 1/2/3D. Those tests are passing, but I don't recall how I convinced myself as to *why* the changed values work.

stefanozampini:
I assume this solves the full element problem adding a non-zero diagonal term right? How can use this to solve for the element interior?

jeremylt:
Right now I don't do anything with the zero eigenvalue for the Laplacian. Improving the handling of the eigenvalues is coming. 

The new test, t540, shows sample use.

jedbrown:
Interior Dirichlet solves will come soon, in a subsequent PR. Stefano, would you prefer the solver for interior (after boundary conditions have been lifted to the RHS), or the element solve with in-place Dirichlet conditions?

stefanozampini:
In BDDC, if A = [A_II, A_IB; A_BI A_BB] is the subdomain Neumann matrix,  the Dirichlet solve is for A_II only. The "lifting" is done explicitly by multiplying by A_IB (Schur complement).
I'm not sure what is the best way of doing it. It would be nice to have a MatCreateSubMatrix_CEED that is aware of the fact that the index set used gives rise to a bunch of uncoupled Dirichlet element problems.

jedbrown:
Yes, but we're interested in vectorizing over elements. If we could apply an in-place interior solve, it would hide the internal details (which depend on element order and architecture) from the BDDC solver.

stefanozampini:
we need a PC for this. What do you mean by vectorizing over elements? apply the same element solve to multiple rhs?

jeremylt:
The FDM inverse is formed from the same generalized eigenvectors for each element but different eigenvalues on each element. Depending on order, we can get better performance by interlacing 8 elements for the tensor contractions for these eigenvectors.
I.e. for `FDMinv = V^T S V` only `S` differs on each element so applying `V` over 8 element batches can give better vectorization and performance.

interface/ceed-basis.c
```diff
@@ -579,7 +579,11 @@ int CeedQRFactorization(Ceed ceed, CeedScalar *mat, CeedScalar *tau,
     // norm of v[i:m] after modification above and scaling below
     //   norm = sqrt(v[i]*v[i] + sigma) / v[i];
     //   tau = 2 / (norm*norm)
-    tau[i] = 2 * v[i]*v[i] / (v[i]*v[i] + sigma);
+    if (sigma > 1e-15)
+      tau[i] = 2 * v[i]*v[i] / (v[i]*v[i] + sigma);
+    else
+      tau[i] = 0;
+
```
jeremylt: This change is needed for the SymmetricSchur but makes a mess of the QRFactorization. I have dropped it.

tests/output/t302-basis-f.out
jeremylt: Added a test and reverted this change. The symmetry of BP3 was indeed masking an issue.

jedbrown: Thanks.

include/ceed-impl.h
```diff
@@ -23,12 +23,7 @@
 #include <ceed-backend.h>
 #include <stdbool.h>
 
-#define CEED_INTERN CEED_EXTERN __attribute__((visibility ("hidden")))
-
-#define CEED_MAX_RESOURCE_LEN 1024
-#define CEED_ALIGN 64
-
-#define CEED_COMPOSITE_MAX 16
+#define CEED_EPSILON 1E-16
```
jeremylt: @jedbrown, I made this a preprocessor macro for now. I figure we can get fancy when we get around to 
```
/// We can discuss ways to avoid forcing these to be compile-time decisions,
///   but let's leave that for later.
```
for `CeedScalar`

jeremylt:
This should be good to merge. The only CI issue is going to be the OCCA OpenCL issue. I plan on doing a performance refactor after I fix the fallback mechanism in a followup PR.

Issue #374, opened 2019-10-03
CUDA Gen Multiple Output Mode Error


jeremylt:
The backend `/gpu/cuda/gen` errors when an operator has several `CeedEvalMode` writing to the same vector.

To reproduce, see `t532-operator`

Valgrind stack trace:
```
==32498== Process terminating with default action of signal 6 (SIGABRT): dumping core
==32498==    at 0xC0D9207: raise (in /usr/lib64/libc-2.17.so)
==32498==    by 0xC0DA8F7: abort (in /usr/lib64/libc-2.17.so)
==32498==    by 0x4E4B0FB: CeedErrorAbort (ceed.c:143)
==32498==    by 0x4E4B1D0: CeedErrorImpl (ceed.c:108)
==32498==    by 0x4E4AA73: CeedVectorGetArray (ceed-vec.c:196)
==32498==    by 0x4E5D46C: CeedOperatorApply_Cuda_gen (ceed-cuda-gen-operator.c:91)
==32498==    by 0x401A7A: main (t532-operator.c:132)
```

`/gpu/cuda/gen` attempts to gain write access to the same vector twice, once for `CEED_EVAL_INTERP` and once for `CEED_EVAL_GRAD`.

This bug is not present in any other CUDA backend.

YohannDudouit:
I can make a temporary fix for this issue. The problem is that multiple write access are requested, and I'm not convinced that this error is not the expected behavior. Since this operator requires multiple write access, maybe that all backends should produce an error.
I can provide a fix (hack?) for the `CEED_VECTOR_ACTIVE` output, because it's easy to check, and avoid the multiple write access. For any output, it would require that I check the output `CeedVector`s together to see if I have a multiple write access, and instead of using `CeedVectorGetArray` use the cached value. I think that this kind of behavior would bypass the memory system of the `CeedVector`, because what we're really dealing with is a multiple write access I think.

I think it would make more sense to have the possibility to combine certain eval modes like: `CEED_EVAL_INTERP`, `CEED_EVAL_GRAD`, `CEED_EVAL_DIV`, `CEED_EVAL_CURL`.
This would also prevent to read/write multiple times the same input/output.
Any opinion on this @jedbrown ?

jedbrown:
[...]

PR #376, opened 2019-10-07
Add Non-tensor Basis Transpose Grad Tests

jeremylt:


jeremylt:
This doesn't fix the problem you found @YohannDudouit, but I think this is still a good test to add.

tests/t522-operator.h
```diff
@@ -0,0 +1,65 @@
...
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+CEED_QFUNCTION(setup)(void *ctx, const CeedInt Q,
+                      const CeedScalar *const *in,
+                      CeedScalar *const *out) {
+  const CeedScalar *qw = in[0], *J = in[1];
+  CeedScalar *qd = out[0];
+
+  for (CeedInt i=0; i<Q; i++) {
+    // Stored in Voigt convention
```
valeriabarra: I would push
```
// qdata stored in Voigt convention
        qd: 0 2
            2 1
```
 on line 32. Otherwise it seems like also J is stored in Voigt convention, and it's not.

jeremylt: All set

jeremylt:
Added t325 to test for bug found by @nbeams. Thank you again

jeremylt:
Just checked the new test on CUDA - all tests pass (except for the known issue #374)

PR #380, opened 2019-10-11
WIP - Proposed Delegation for Assemble Linear QF, Assemble Diag

jeremylt:
This might be a way to delegate to /cpu/self/ref/serial to get AssembleLinearQF and AssembleLinearDiagonal for all operators. I think I need to put in some SyncArrays to handle the work vectors that I put in AssembleLinearQF, but I'm not 100% sure if that will be enough. Thoughts on this general approach?

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -58,7 +58,7 @@ int CeedOperatorCreate(Ceed ceed, CeedQFunction qf, CeedQFunction dqf,
     return 0;
   }
 
-  ierr = CeedCalloc(1,op); CeedChk(ierr);
+  ierr = CeedCalloc(1, op); CeedChk(ierr);
   (*op)->ceed = ceed;
   ceed->refcount++;
   (*op)->refcount = 1;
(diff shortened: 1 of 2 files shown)


interface/ceed-operator.c
```diff
@@ -303,7 +303,28 @@ int CeedOperatorAssembleLinearQFunction(CeedOperator op, CeedVector *assembled,
...
+    ierr = CeedCalloc(1, &opref); CeedChk(ierr);
+    memcpy(&opref, &op, sizeof(op));
+    opref->data = NULL;
+    ierr = ceedref->OperatorCreate(opref); CeedChk(ierr);
+
+    // Assemble
+    ierr = op->AssembleLinearQFunction(opref, assembled, rstr, request);
+
+    // Cleanup
+    ierr =  opref->Destroy(opref); CeedChk(ierr);
```
jedbrown: Does the copy above mean that some arrays could be freed by `opref->Destroy`? Perhaps it's okay with strict programming practices.

jeremylt: In theory, `opref->Destroy` should only destroy backend specific data, so I think we are ok there.

Where we are running into issues is with data synchronization inside the assembly of the QF. i.e. I will update a work vector's contents, but the vector that the work vector looks into does not know that its data has been updated.

jeremylt: Perhaps some sort of Vec+offset workflow would help with this. I would drop the work vectors and just keep track of an index range to update for each input field/dim.

jeremylt:
Looking at this again and thinking more - using host pointers to avoid extra memcopies is so baked into the ref backend that I do not think this will work unless we were to completely rewrite how the ref backend works from the ground up. Closing.

PR #383, opened 2019-10-16
Facilitate Empty Operators

jeremylt:
Process decomposition may provide a restriction that is empty (0 elements), which currently causes an error in a libCEED operator as if the user did not define a restriction. This PR allows a user to intentionally set a restriction with 0 elements. If there are 0 elements, then the backend code is not called. (proper backend code should run fine with 0 elements, but it seems silly to make the function calls for an empty operator)

Thanks @valeriabarra for catching this.

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -107,6 +107,7 @@ int CeedCompositeOperatorCreate(Ceed ceed, CeedOperator *op) {
   (*op)->ceed = ceed;
   ceed->refcount++;
   (*op)->composite = true;
+  (*op)->numelements = -2;
   ierr = CeedCalloc(16, &(*op)->suboperators); CeedChk(ierr);
   ierr = ceed->CompositeOperatorCreate(*op); CeedChk(ierr);
   return 0;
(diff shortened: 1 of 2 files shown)


interface/ceed-operator.c
```diff
@@ -106,6 +107,7 @@ int CeedCompositeOperatorCreate(Ceed ceed, CeedOperator *op) {
   (*op)->ceed = ceed;
   ceed->refcount++;
   (*op)->composite = true;
+  (*op)->numelements = -2;
```
jedbrown: What does `-2` mean?

jeremylt: I can add a comment.
Old schema
0 - unset
1+ - set
New schema
-2 - composite operator
-1 - unset
0+ - set

I'm using -2 as a nonzero value that isn't -1 to simplify the errorchecking logic.

jedbrown: Please document this schema. It looks hazardous in this PR because you only check `numelements != -1` instead of `numelements >= 0` (though perhaps it is intended that those places be unreachable for composite operators).

jeremylt: I could also go real crazy and set a boolian flag if a restriction has been added and then only accept values >=0. It would probably be clearer.

jeremylt:
Fixed. Much more readable with a flag.

PR #384, opened 2019-10-16
Fix Bug in CPU Collocated Basis Grad

jeremylt:
I made a bug in the CPU collocated basis grad code. This fixes it and adjusts the PETSc BPs test parameters to catch this case in the future.


jeremylt:
I'll merge after Travis

PR #385, opened 2019-10-16
Azure - use junit realsearch=%

jeremylt:
As discussed, this tests the PETSc examples on Azure.


jeremylt:
It looks like the Docker image is older than PETSc 3.11.3.

jedbrown:
I updated the Docker container.

jeremylt:
That fixed it; thanks.

PR #386, opened 2019-10-16
Minor AVX Backend Changes

jeremylt:
I 
1) Dropped some dead code
2) Cleaned up PETSc bps a bit
3) Changed the bps testing args to make sure we get all AVX combos tested

### DIFF
--- a/examples/petsc/bps.c
+++ b/examples/petsc/bps.c
@@ -35,7 +35,7 @@
 //     bps -problem bp5 -ceed /omp/occa
 //     bps -problem bp6 -ceed /ocl/occa
 //
-//TESTARGS -ceed {ceed_resource} -test -problem bp2 -degree 3
+//TESTARGS -ceed {ceed_resource} -test -problem bp2 -degree 5 -qextra 5
 
 /// @file
 /// CEED BPs example using PETSc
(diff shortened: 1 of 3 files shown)


examples/petsc/bps.c
```diff
@@ -77,16 +77,14 @@ static PetscInt GlobalStart(const PetscInt p[3], const PetscInt irank[3],
...
+  for (PetscInt i=0; i<p[0]; i++)
+    for (PetscInt j=0; j<p[1]; j++)
       for (PetscInt k=0; k<p[2]; k++) {
         PetscInt mnodes[3], ijkrank[] = {i,j,k};
         if (i == irank[0] && j == irank[1] && k == irank[2]) return start;
         GlobalNodes(p, ijkrank, degree, melem, mnodes);
         start += mnodes[0] * mnodes[1] * mnodes[2];
       }
-    }
-  }
```
jedbrown: I guess I don't much care, but leaving the braces in is more conventional in this case, even though it eats up vertical space on your screen.

jeremylt: I can revert it - we've just largely removed the braces when not strictly required.

jeremylt:
Found the right combo - will merge when CI finishes

PR #391, opened 2019-10-17
Codecov - cover gallery

jeremylt:
I wanted to make sure we are actually fully testing the gallery to so we know that code is good.


jeremylt:
Not all examples are being covered despite
```
  examples:
    paths:
      - examples/
```
oddly

PR #392, opened 2019-10-17
Get Full Ceed Resource Name

jeremylt:
When the user specifies `/cpu/self` or `/gpu`, we should support a way to tell the user what backend is being used.

I didn't implement in Fortran and add a Fortran test because I didn't really want to fiddle with passing strings between C and F77, but I can work on it if it is desirable.

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -405,6 +405,13 @@ int CeedInit(const char *resource, Ceed *ceed) {
   // Backend specific setup
   ierr = backends[matchidx].init(resource, *ceed); CeedChk(ierr);
 
+  // Copy resource prefix, if backend setup sucessful
+  size_t len = strlen(backends[matchidx].prefix);
+  char *tmp;
+  ierr = CeedCalloc(len+1, &tmp); CeedChk(ierr);
+  memcpy(tmp, backends[matchidx].prefix, len+1);
+  (*ceed)->resource = tmp;
+
   return 0;
 }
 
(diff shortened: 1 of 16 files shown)


YohannDudouit:
Thanks this will be useful to detect when users do a typo in the name of the backend they ask for, and to know which backend is run when `/cpu/self` and `/gpu` is the spec.

interface/ceed.c
```diff
@@ -636,6 +643,22 @@ int CeedSetData(Ceed ceed, void* *data) {
...
+
+  @param ceed            Ceed to get resource name of
+  @param[out] resource   Variable to store resource name
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Basic
+**/
+
+int CeedGetResource(Ceed ceed, char* *resource) {
```
jedbrown: Should not return a mutable pointer. Also, `char **resource` is just fine. No need to mix syntax conventions just because it's a `**`.

jeremylt: In this case would it be typical to allocate the space, make a copy, and expect the user to free it? I thought about that but wasn't sure.

We've done `char* *whatever` in a few places, so I'll change it globally.

jedbrown: I don't object to returning a pointer to `const`, it just needs to be `const`.

jeremylt: Fixed

jeremylt:
Edited the initial commit to use const.
Added a second commit to fix mixed syntax array pointers globally.

PR #393, opened 2019-10-18
Doc - small fixes

jeremylt:


interface/ceed-elemrestriction.c
```diff
@@ -347,7 +347,9 @@ int CeedElemRestrictionApply(CeedElemRestriction rstr, CeedTransposeMode tmode,
                    indicates the component is the innermost index in
                    ordering of the l-vector
                    tmode=CEED_NOTRANSPOSE)
-  @param v       Output vector (of size @a nelem * @a elemsize when
+  @param u       Input vector (of size @a nnodes * @a ncomp when
+                   tmode=CEED_NOTRANSPOSE)
+  @param v       Output vector (of size @a blksize * @a elemsize when
                    tmode=CEED_NOTRANSPOSE)
```
jedbrown: Should we use "shape" here to communicate how to index, versus "size" which is ambiguous?

PR #394, opened 2019-10-18
Add PETSc example areaplex

valeriabarra:
This example uses the mass operator to compute the surface area of a cube using PETSc with DMPlex.

This example uses the followig aspects that were not seen in previous examples, for new libCEED users to learn :
- handling of problems in which topological dimension is different from geometrical dimension
- handling of non-square coordinate transformation dx/dX Jacobian with intermediate coordinate transformation from 3D physical space, to 2D local and reference spaces


jeremylt:
So, with DMPlex being the base code for all of our examples, perhaps we should rename things
bps -> bpsraw
bpsdmplex -> bps
areadmplex -> area

That way we have the future
bpsdmplexsphere -> bpssphere

We would need to change the benchmarks to run against bpsraw and trace documentation/readme, but otherwise it would be a small change.

examples/petsc/areadmplex.c
```diff
@@ -0,0 +1,368 @@
...
+  CeedInit(ceedresource, &ceed);
+
+  // Print summary
+  P = degree + 1;
+  Q = P + qextra;
+  const char *usedresource;
+  CeedGetResource(ceed, &usedresource);
+  if (!test_mode) {
+    ierr = PetscPrintf(comm,
+                       "\n-- CEED -- libCEED + PETSc Surface Area problem --\n"
```
jeremylt: Maybe -- libCEED + PETSc -- Surface Area Problem --
? The 'CEED' out front is strange to me

jeremylt:
It looks like many of the functions in `setupareaplex.h` aren't being used.

examples/petsc/areadmplex.c
valeriabarra: Sure

valeriabarra:
> It looks like many of the functions in `setupareaplex.h` aren't being used.

That is a stray file!! Thanks for catching

valeriabarra:
> So, with DMPlex being the base code for all of our examples, perhaps we should rename things
> bps -> bpsraw
> bpsdmplex -> bps
> areadmplex -> area
> 
> That way we have the future
> bpsdmplexsphere -> bpssphere
> 
> We would need to change the benchmarks to run against bpsraw and trace documentation/readme, but otherwise it would be a small change.

Sure, if everyone likes the proposed convention, I can change it

jeremylt:
I think this just needs some discussion in the README, otherwise LGTM.

The only other thing would be resolution on the naming conventions we want to take in this folder.

valeriabarra:
> I think this just needs some discussion in the README, otherwise LGTM.
> 
> The only other thing would be resolution on the naming conventions we want to take in this folder.

Thanks Jeremy. I've updated the README. 

For the naming convention I would like to hear at least someone else's opinion so we can see if there is a consensus or majority.

jedbrown:
Are we going to use it for non-BP qfunctions? I don't really care what the directory is called, but let's stop moving them (after this time, if it's important).

valeriabarra:
> Are we going to use it for non-BP qfunctions? I don't really care what the directory is called, but let's stop moving them (after this time, if it's important).

The naming convention Jeremy proposed did not regard the directory, but the `.c` files and executables: 

bps -> bpsraw
bpsdmplex -> bps
areadmplex -> area

jedbrown:
Oh, that looks fine to me. I think the DMPlex code is cleaner and more relevant unless someone really wants raw grid handling.

examples/petsc/README.md
```diff
@@ -32,9 +32,9 @@ In addition to the common arguments, the following arguments may be set:
 This code solves the CEED bakeoff problems on a unstructured grid using DMPlex
 with p-multigrid implemented in PCMG. This example requires a PETSc version later than 3.11.3.
 
-To build, run `make multigrid`.
+To build, run `make multigrid`
 
-To run, `./multigrid -ceed [ceed-resource] -problem bp[1-6] -degree [degree]`.
+To run, `./multigrid -ceed [ceed-resource] -problem bp[1-6] -degree [degree]`
 
 In addition to the common arguments, the following arguments may be set:
```
valeriabarra: @jeremy I can't tag line 46 here. But I'm wondering if "The following arguments can be specified for both examples:" refers to the bpsdmplex and multigrid examples, or actually to all of above. I can update it, if necessary.

jeremylt: Those six arguments apply to bpsraw, bps, and multigrid.

valeriabarra: Ok then it was outdated. I can update it.

Ps: I'm not sure why Fortran tests would fail with Clang, since I've only updated this README in this commit

valeriabarra: PPs: I've removed the periods at the end of those lines in case an inexperienced user copies and pastes the Readme line to run, and does not realize the period is not necessary to run

valeriabarra:
Git didn't like the renaming of files for some reason. My `git status` was all clear and nice, before the commit, showing the renamed files and modified files:
```
	modified:   ../../benchmarks/README.md
	modified:   ../../benchmarks/benchmark.sh
	renamed:    ../../benchmarks/petsc-bps.sh -> ../../benchmarks/petsc-bpsraw.sh
	modified:   ../README.md
	modified:   Makefile
	modified:   README.md
	renamed:    areadmplex.c -> area.c
	renamed:    bpsdmplex.c -> bps.c
	renamed:    bps.c -> bpsraw.c
```

examples/petsc/README.md
```diff
@@ -43,11 +43,30 @@ In addition to the common arguments, the following arguments may be set:
...
+
+This example uses the mass matrix to compute the surface area of a cube, defined via DMPlex. 
+
+To build, run `make area`
+
+To run, `./area -ceed [ceed-resource] -petscspace_degree [degree]`
+
+### Command line arguments
+
+The following arguments can be specified for both examples:
```
jeremylt: can be specified for *this example*?

valeriabarra: Ah great catch, thank you! Forgot to update that line

jeremylt: I think you also want examples -> example at the end of the line.

valeriabarra: Of course! Sorry!

PR #397, opened 2019-10-23
Travis - COVERAGE=1 only for linux

jeremylt:
I think Travis OSX is not playing nice with some homebrew updates. I think this gets around the issue. (We aren't running codecov on OSX anyways for speed.)

I'll flag folks for review once Travis comes back good.


jeremylt:
This fixes it.

PR #399, opened 2019-10-27
Restriction Optimizations

jeremylt:
This branch is a work in progress trying to improve the performance of the `/cpu/self/ref` element restriction.

I need to see how well this does on a Skylake, but in my first commit I have made blocksize a compile time parameter. This change provides a mild benefit on my laptop. My hope is that for `blksize=8` this will help an AVX512 machine vectorize intelligently.

```
Low Order - p=1

W/O Change Blocked
  13.85%  bpsraw   libceed.so              [.] CeedElemRestrictionApply_
   6.42%  bpsraw   [unknown]               [k] 0xffffffffb8400b07
   6.33%  bpsraw   bpsraw                  [.] SetupDiffRhs

W/Change Blocked
   9.68%  bpsraw   libceed.so              [.] CeedElemRestrictionApplyB
   6.47%  bpsraw   [unknown]               [k] 0xffffffffb8400b07
   6.23%  bpsraw   bpsraw                  [.] SetupDiffRhs

W/O Change Serial
   7.51%  bpsraw   libceed.so              [.] CeedTensorContractApply_X
   6.89%  bpsraw   libceed.so              [.] CeedBasisApply_Ref
   6.86%  bpsraw   libceed.so              [.] CeedElemRestrictionApply_

W/Change Serial
   7.80%  bpsraw   libceed.so              [.] CeedTensorContractApply_◆
   6.93%  bpsraw   libceed.so              [.] CeedBasisApply_Ref
   5.41%  bpsraw   libopenblasp-r0.3.6.so  [.] dgemm_kernel_HASWELL
   5.32%  bpsraw   libopenblasp-r0.3.6.so  [.] dgemm_tn
   4.92%  bpsraw   libceed.so              [.] CeedElemRestrictionApply

High Order - p=9

W/O Change Blocked
  12.87%  bpsraw   bpsraw                  [.] Diff
  11.22%  bpsraw   libceed.so              [.] CeedElemRestrictionApply_
   9.85%  bpsraw   libc-2.29.so            [.] __memmove_avx_unaligned_e

W/Change Blocked
  12.94%  bpsraw   bpsraw                  [.] Diff
  10.74%  bpsraw   libceed.so              [.] CeedElemRestrictionApplyB
   9.67%  bpsraw   libc-2.29.so            [.] __memmove_avx_unaligned_e

W/O Change Serial
  10.30%  bpsraw   bpsraw                  [.] Diff
  10.18%  bpsraw   libceed.so              [.] CeedElemRestrictionApply_
[...]

PR #400, opened 2019-10-28
Some cleaning

valeriabarra:
This is a PR of small things that I've noticed here and there.

### DIFF
--- a/examples/petsc/bpsraw.c
+++ b/examples/petsc/bpsraw.c
@@ -809,6 +809,12 @@ int main(int argc, char **argv) {
     ierr = KSPGetIterationNumber(ksp, &its); CHKERRQ(ierr);
     ierr = KSPGetResidualNorm(ksp, &rnorm); CHKERRQ(ierr);
     if (!test_mode || reason < 0 || rnorm > 1e-8) {
+      CeedInt gsize;
+      ierr = VecGetSize(X, &gsize); CHKERRQ(ierr);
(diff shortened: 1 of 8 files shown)


examples/petsc/bpsraw.c
```diff
@@ -809,6 +809,12 @@ int main(int argc, char **argv) {
     ierr = KSPGetIterationNumber(ksp, &its); CHKERRQ(ierr);
     ierr = KSPGetResidualNorm(ksp, &rnorm); CHKERRQ(ierr);
     if (!test_mode || reason < 0 || rnorm > 1e-8) {
+      CeedInt gsize;
```
jeremylt: I think we already have gsize floating around and defined. Should we remove these extra two lines in all 4 examples?

jeremylt:
Would you like to cherry pick https://github.com/CEED/libCEED/pull/399/commits/36437134ccd3f2e6dabf0771a357ccd4721cfd16 into this branch as well? It fits thematically.

examples/petsc/bpsraw.c
valeriabarra: Done, but it was only in this one example that it was declared in the block scope

valeriabarra:
> Would you like to cherry pick [3643713](https://github.com/CEED/libCEED/commit/36437134ccd3f2e6dabf0771a357ccd4721cfd16) into this branch as well? It fits thematically.

Done. If fits. Thanks! I had forgotten to update that after the naming convention update

PR #405, opened 2019-10-30
Fix Style Indentation Errors

### LINKED ISSUE #396: Make Style
`make style` introduces several errors that have to be manually reverted.

jeremylt:
Astyle struggles with some of our indentation. This PR make some small changes and adds `INDENT-OFF` where needed so that we can run `make style` without needing manual fixes.

Fixes #396

### DIFF
--- a/backends/cuda-reg/ceed-cuda-reg-basis.c
+++ b/backends/cuda-reg/ceed-cuda-reg-basis.c
@@ -21,6 +21,7 @@
 
 //*********************
 // reg kernels
+// *INDENT-OFF*
 static const char *kernels3dreg = QUOTE(
 
 typedef CeedScalar real;
@@ -266,7 +267,7 @@ inline __device__ void Contract3d(const real *A, const real *B,
     for (int a3 = 0; a3 < nA3; a3++)
       for (int b2 = 0; b2 < nB2; b2++)
         for (int t = 0; t < nB1; t++) {
-          T[a2 + a3*nA2 + b2*nA2*nA3] += B[b2*nB1 + t] * A[a3*nA2*nA1 + a2*nA1 + t];
+          T[a2 + a3*nA2 + b2*nA2*nA3] += B[b2*nB1+t] * A[a3*nA2*nA1+a2*nA1+t];
         }
 }
 
@@ -278,7 +279,7 @@ inline __device__ void ContractTranspose3d(const real *A, const real *B,
     for (int a3 = 0; a3 < nA3; a3++)
       for (int b1 = 0; b1 < nB1; b1++)
         for (int t = 0; t < nB2; t++) {
-          T[a2 + a3*nA2 + b1*nA2*nA3] += B[t*nB1 + b1] * A[a3*nA2*nA1 + a2*nA1 + t];
+          T[a2 + a3*nA2 + b1*nA2*nA3] += B[t*nB1+b1] * A[a3*nA2*nA1+a2*nA1+t];
         }
 }
 
@@ -468,6 +469,7 @@ extern "C" __global__ void weight(const CeedInt nelem,
 }
 
 );
+// *INDENT-ON*
 
 int CeedCudaInitInterp(CeedScalar *d_B, CeedInt P1d, CeedInt Q1d,
                        CeedScalar **c_B);
--- a/examples/ceed/ex2.h
+++ b/examples/ceed/ex2.h
@@ -111,10 +111,12 @@ CEED_QFUNCTION(f_apply_diff)(void *ctx, const CeedInt Q,
       // Stored in Voigt convention
       // 0 2
       // 2 1
+      // *INDENT-OFF*
       const CeedScalar dXdxdXdxT[2][2] = {{qdata[i+0*Q],
                                            qdata[i+2*Q]},
                                           {qdata[i+2*Q],
                                            qdata[i+1*Q]}};
+      // *INDENT-ON*
       // j = direction of vg
       for (int j=0; j<2; j++)
         vg[i+j*Q] = (du[0] * dXdxdXdxT[0][j] +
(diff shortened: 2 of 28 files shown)


backends/cuda-reg/ceed-cuda-reg-basis.c
```diff
@@ -21,6 +21,7 @@
 
 //*********************
 // reg kernels
+// *INDENT-OFF*
```
jeremylt: Asyle has no clue how to handle these quoted kernels.

examples/mfem/bp1.hpp
```diff
@@ -115,8 +115,8 @@ class CeedMassOperator : public mfem::Operator {
     const int ir_order = 2*(order + 2) - 1; // <-----
     const mfem::IntegrationRule &ir =
       mfem::IntRules.Get(mfem::Geometry::SEGMENT, ir_order);
-    CeedInt nqpts, nelem = mesh->GetNE(), dim = mesh->SpaceDimension(),
-            ncompx = dim;
+    CeedInt nelem = mesh->GetNE(), dim = mesh->SpaceDimension(),
```
jeremylt: When there is a mix of declarations with and without initializations, astyle indents correctly when the initializations come first.

examples/navier-stokes/densitycurrent.h
```diff
@@ -78,15 +78,21 @@
 CEED_QFUNCTION(ICsDC)(void *ctx, CeedInt Q,
                       const CeedScalar *const *in, CeedScalar *const *out) {
 
+// *INDENT-OFF*
```
jeremylt: Astyle tries to reverse indent this. The behavior of preproc linting inside of functions is apparently undefined and won't be fixed.

gallery/poisson3d/ceed-poisson3dapply.h
```diff
@@ -42,6 +42,7 @@ CEED_QFUNCTION(Poisson3DApply)(void *ctx, const CeedInt Q,
     // 0 5 4
     // 5 1 3
     // 4 3 2
+    // *INDENT-OFF*
     const CeedScalar dXdxdXdxT[3][3] = {{qd[i+0*Q],
```
jeremylt: Astyle really hates writing this style and wants to explode the array so that braces each have their own line.

jedbrown: We could consider packing such arrays via a loop or even a macro/function call (though we'd have to make it available on the CUDA/JIT side as well).

jeremylt: Right, I think that would be preferable so we can minimize the use of `INDENT-OFF`

jeremylt: Updated

examples/ceed/ex2.h
```diff
@@ -103,20 +103,21 @@ CEED_QFUNCTION(f_apply_diff)(void *ctx, const CeedInt Q,
...
-      const CeedScalar dXdxdXdxT[2][2] = {{qdata[i+0*Q],
-                                           qdata[i+2*Q]},
-                                          {qdata[i+2*Q],
-                                           qdata[i+1*Q]}};
-      // *INDENT-ON*
+      const CeedInt idx[] = {0, 2, 2, 1};
+      CeedScalar dXdxdXdxT[2][2];
+      for (CeedInt m=0; m<2; m++)
+        for (CeedInt n=0; n<2; n++)
+          dXdxdXdxT[m][n] = qdata[i+idx[n+2*m]*Q];
```
valeriabarra: Are we sure this is the preferred way we want this? It seems to me that it is harder to read and therefore to spot bugs, than fixing indentation style by hand, if/when needed

examples/ceed/ex2.h
```diff
@@ -111,11 +111,11 @@ CEED_QFUNCTION(f_apply_diff)(void *ctx, const CeedInt Q,
       // Stored in Voigt convention
       // 0 2
       // 2 1
-      const CeedInt idx[] = {0, 2, 2, 1};
+      const CeedInt idx[2][2] = {{0, 2}, {2, 1}};
       CeedScalar dXdxdXdxT[2][2];
       for (CeedInt m=0; m<2; m++)
         for (CeedInt n=0; n<2; n++)
-          dXdxdXdxT[m][n] = qdata[i+idx[n+2*m]*Q];
+          dXdxdXdxT[m][n] = qdata[i+idx[m][n]*Q];
```
valeriabarra: With the 2D array indices it reads a bit better, but it still looks to me more time consuming to read/debug/write this than the previous version

examples/ceed/ex2.h
jeremylt: I'm happy with dropping 2f40c9d502a896ccbf8765d62c5bdd42b5843f70 and cfd0a8060ed6616b70ae77e2243fb3d74a7dd253 if everyone prefers, but without doing something `make style` is useless - fixing errors in 20+ files every time is not practical.

I think this comes back to the same discussion in https://github.com/CEED/libCEED/pull/323

jeremylt:
@jedbrown, any strong preference between

```
    const CeedInt idx[2][2] = {{0, 2}, {2, 1}};
    CeedScalar dXdxdXdxT[2][2];
    for (CeedInt m=0; m<2; m++)
      for (CeedInt n=0; n<2; n++)
```
and 
```
    const CeedScalar dXdxdXdxT[2][2] = {{qd[i+0*Q],
                                         qd[i+2*Q]},
                                        {qd[i+2*Q],
                                         qd[i+1*Q]}
                                       };
```

I think that's the last thing to settle on this PR.

jedbrown:
I have a mild preference for the latter unless we're encapsulating that pattern in a macro so that it's a one-liner in QFunctions. We may want that eventually, but I don't see a pressing need to do it now.

jeremylt:
Updated to drop the loops. This should be good to go now and we can use `make style` again.

valeriabarra:
Just curious, will we have to filter-out the python interface once we have that up, to preserve the intended indentations there?

LGTM though. Thanks.

jeremylt:
I'm not sure how this will handle python. I'm curious to see.

PR #406, opened 2019-10-30
Delegate AssembleLinearQF to ref/serial when not impl

jeremylt:
This provides `AssembleLinearQF` to the GPU backends via `/cpu/self/ref/serial` so we can do diag based preconditioning for the GPU backends soon. I'll refactor `AssembleLinearDiag` downstream of this.

I also fixed an OCCA backend bug where they did not support multiple operator fields writing into the same output vector.

### DIFF
--- a/backends/occa/ceed-occa-restrict.okl
+++ b/backends/occa/ceed-occa-restrict.okl
@@ -61,7 +61,7 @@
   for (int i=0; i<nnodes; i++; @tile(TILE_SIZE,@outer,@inner)){
     const int rng1 = toffsets[i];
     const int rngN = toffsets[i+1];
-    double value = 0.0;
+    double value = vv[i];
     for (int j=rng1; j<rngN; ++j){
       const int tid = tindices[j];
       value += uu[tid];
@@ -80,7 +80,7 @@
     const int rng1 = toffsets[i];
     const int rngN = toffsets[i+1];
     for (int d = 0; d < ncomp; ++d) {
-      double value = 0.0;
+      double value = vv[d*nnodes+i];
       for (int j=rng1; j<rngN; ++j) {
         int n = tindices[j] % elemsize;
         int e = tindices[j] / elemsize;
@@ -101,7 +101,7 @@
     const int rng1 = toffsets[i];
     const int rngN = toffsets[i+1];
     for (int d = 0; d < ncomp; ++d) {
-      double value = 0.0;
+      double value = vv[d+i*ncomp];
       for (int j=rng1; j<rngN; ++j) {
         int n = tindices[j] % elemsize;
         int e = tindices[j] / elemsize;
--- a/include/ceed-backend.h
+++ b/include/ceed-backend.h
@@ -183,6 +183,7 @@ CEED_EXTERN int CeedOperatorGetFallbackResource(CeedOperator op,
     const char **resource);
 CEED_EXTERN int CeedOperatorSetFallbackResource(CeedOperator op,
     const char *resource);
+CEED_EXTERN int CeedOperatorFallbackGetParentCeed(CeedOperator op, Ceed *ceed);
 CEED_EXTERN int CeedOperatorGetData(CeedOperator op, void **data);
 CEED_EXTERN int CeedOperatorSetData(CeedOperator op, void **data);
 CEED_EXTERN int CeedOperatorSetSetupDone(CeedOperator op);
--- a/include/ceed-backend.h
+++ b/include/ceed-backend.h
@@ -179,6 +179,10 @@ CEED_EXTERN int CeedOperatorGetQFunction(CeedOperator op, CeedQFunction *qf);
 CEED_EXTERN int CeedOperatorGetNumSub(CeedOperator op, CeedInt *numsub);
 CEED_EXTERN int CeedOperatorGetSubList(CeedOperator op,
                                        CeedOperator **suboperators);
+CEED_EXTERN int CeedOperatorGetFallbackResource(CeedOperator op,
+    const char **resource);
+CEED_EXTERN int CeedOperatorSetFallbackResource(CeedOperator op,
+    const char *resource);
 CEED_EXTERN int CeedOperatorGetData(CeedOperator op, void **data);
 CEED_EXTERN int CeedOperatorSetData(CeedOperator op, void **data);
 CEED_EXTERN int CeedOperatorSetSetupDone(CeedOperator op);
--- a/include/ceed-backend.h
+++ b/include/ceed-backend.h
@@ -65,6 +65,11 @@ CEED_EXTERN int CeedGetObjectDelegate(Ceed ceed, Ceed *delegate,
                                       const char *objname);
 CEED_EXTERN int CeedSetObjectDelegate(Ceed ceed, Ceed delegate,
                                       const char *objname);
+CEED_EXTERN int CeedGetOperatorFallbackResource(Ceed ceed,
+    const char **resource);
+CEED_EXTERN int CeedSetOperatorFallbackResource(Ceed ceed,
+    const char *resource);
+CEED_EXTERN int CeedGetOperatorFallbackParentCeed(Ceed ceed, Ceed *parent);
 CEED_EXTERN int CeedSetBackendFunction(Ceed ceed,
                                        const char *type, void *object,
                                        const char *fname, int (*f)());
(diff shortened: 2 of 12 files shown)


backends/occa/ceed-occa-restrict.okl
```diff
@@ -61,7 +61,7 @@
   for (int i=0; i<nnodes; i++; @tile(TILE_SIZE,@outer,@inner)){
     const int rng1 = toffsets[i];
     const int rngN = toffsets[i+1];
-    double value = 0.0;
+    double value = vv[i];
```
jeremylt: I can isolate this to a separate commit if desirable.

jeremylt:
I added a testing exception for OCCA. With this change, the OCCA backends attempt to run `petsc-multigrid` but fail because the OCCA backends do not currently support QFunction reuse.

This should be fixed by #305.

jeremylt:
A Fortran test that was passing locally is failing on Travis. Investigating, but it's probably something small.

jeremylt:
It was a small bug in the QFunction on the Fortran side. Fixed now.

interface/ceed-operator.c
```diff
@@ -304,8 +304,58 @@ int CeedOperatorAssembleLinearQFunction(CeedOperator op, CeedVector *assembled,
...
     // LCOV_EXCL_STOP
   }
-  ierr = op->AssembleLinearQFunction(op, assembled, rstr, request);
-  CeedChk(ierr);
+  if (op->AssembleLinearQFunction) {
+    ierr = op->AssembleLinearQFunction(op, assembled, rstr, request);
+    CeedChk(ierr);
+  } else { // Delegate to /cpu/self/ref/serial
+    Ceed ceedref;
+    ierr = CeedInit("/cpu/self/ref/serial", &ceedref); CeedChk(ierr);
```
jedbrown: It isn't the first use of this pattern (thus perhaps out of scope for this PR), but should backends be responsible for identifying their delegate (`ceedref`), instead of losing that symmetry here in interface code? It could be stored in `ceed->ceedref`.

jeremylt: Backends do generally specify their own delegate and are responsible for their creation (but not destruction). A `ceed` can specify a default delegate as well as delegates for specific objects; however, delegates are treated as all-or-nothing with respect to specific objects. The code is often interdependent enough that I have been assuming that we can't mix and match functions from different `ceed`s with the same object.

```
struct Ceed_private {
  const char *resource;
  Ceed delegate;
  Ceed parent;
  objdelegate *objdelegates;
  int objdelegatecount;
...
```

We don't currently have a way for any backend to delegate only `CeedOperatorAssembleLinearQFunction` while implementing all other `CeedOperator` functions.

This is the first use of a single function delegation. Whatever we adopt here I'd like to also use for `AssembleLinearDiagonal` and `CreateElementFDMInverse`, so it seems like the right time to change it if we want to.

I can add
1) carrying around this operator+qfunction clone for the life of the operator and/or
2) letting the backend specify this 'delegate' (maybe 'fallback'? we're overusing the phrase 'delegate' to mean two/three different things here) but default to `/cpu/self/ref/serial`

if that would be more preferable/flexible/extensible/ect

jeremylt: @jedbrown, I refactored to make it clear that this is different that our typical delegation mechanism and to allow backends to select something other than `/cpu/self/ref/serial` if desired. What do you think?

interface/ceed-operator.c
```diff
@@ -248,6 +252,50 @@ int CeedCompositeOperatorAddSub(CeedOperator compositeop, CeedOperator subop) {
...
+**/
+int CeedOperatorCreateFallback(CeedOperator op) {
+  int ierr;
+
+  // Fallback Ceed
+  const char *resource;
+  ierr = CeedOperatorGetFallbackResource(op, &resource);
+  Ceed ceedref;
+  ierr = CeedInit(resource, &ceedref); CeedChk(ierr);
+  op->fallbackceed = ceedref;
```
jedbrown: Can we have one `ceed->fallback` instead of one per operator?

interface/ceed-operator.c
```diff
@@ -248,6 +252,50 @@ int CeedCompositeOperatorAddSub(CeedOperator compositeop, CeedOperator subop) {
...
+  // Clone Op
+  CeedOperator opref;
+  ierr = CeedCalloc(1, &opref); CeedChk(ierr);
+  memcpy(opref, op, sizeof(*opref)); CeedChk(ierr);
+  opref->data = NULL;
+  opref->setupdone = 0;
+  opref->ceed = ceedref;
+  ierr = ceedref->OperatorCreate(opref); CeedChk(ierr);
+  opref->parentceed = op->ceed;
+  op->fallback = opref;
```
jedbrown: Are we making hanging references above? Is it possible to infinite loop?

interface/ceed-operator.c
```diff
@@ -747,6 +806,65 @@ int CeedOperatorGetData(CeedOperator op, void **data) {
...
+  @param op            CeedOperator
+  @param[out] resource Variable to store fallback resource
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+
+int CeedOperatorGetFallbackResource(CeedOperator op, const char **resource) {
+  *resource = (const char *)op->fallbackresource;
```
jedbrown: Documentation should be explicit that string cannot outlive the `CeedOperator`, and could become invalid if `CeedOperatorSetFallbackResource` is called.

interface/ceed-operator.c
```diff
@@ -905,6 +1023,16 @@ int CeedOperatorDestroy(CeedOperator *op) {
   ierr = CeedQFunctionDestroy(&(*op)->dqf); CeedChk(ierr);
   ierr = CeedQFunctionDestroy(&(*op)->dqfT); CeedChk(ierr);
 
+  // Destroy fallback
+  if ((*op)->fallback) {
+    ierr = (*op)->qffallback->Destroy((*op)->qffallback); CeedChk(ierr);
+    ierr = CeedFree(&(*op)->qffallback); CeedChk(ierr);
+    ierr = (*op)->fallback->Destroy((*op)->fallback); CeedChk(ierr);
```
jedbrown: We can't we `CeedOperatorDestroy(&(*op)->fallback);`?

interface/ceed-operator.c
jeremylt: Easy enough to do - it would be consistent per Ceed

interface/ceed-operator.c
jeremylt: Updated

interface/ceed-operator.c
jeremylt: The fallback contains references to memory that has been freed, so we need to only call the backend destructor code.

interface/ceed-operator.c
jeremylt: I don't believe its possible to infinite loop as written. It runs Valgrind clean when I build with only the CPU backends and use a fallback rather than the `/cpu/self/ref/blocked` impl of `AssembleLinearQF`.

jedbrown: What if the delegate doesn't implement the desired method and attempts to delegate to itself? This may not be possible at present, but I'd be concerned about it happening in the future.

jeremylt: I put in a check that a ceed cannot delegate to itself.

jeremylt:
@jedbrown, good to merge now? I think I addressed your comments and I think its much better code now.

I'm thinking that I should refactor FDM Element Inverse and Assemble Linear Diag in a follow-on to both this and #372 to keep the PRs smaller.

interface/ceed-operator.c
```diff
@@ -248,6 +248,58 @@ int CeedCompositeOperatorAddSub(CeedOperator compositeop, CeedOperator subop) {
   return 0;
 }
 
+/**
+  @brief Duplicate a CeedOperator with a refrence Ceed to fallback for advanced
```
jedbrown:
```suggestion
  @brief Duplicate a CeedOperator with a reference Ceed to fallback for advanced
```

include/ceed-backend.h
```diff
@@ -65,6 +65,11 @@ CEED_EXTERN int CeedGetObjectDelegate(Ceed ceed, Ceed *delegate,
                                       const char *objname);
 CEED_EXTERN int CeedSetObjectDelegate(Ceed ceed, Ceed delegate,
                                       const char *objname);
+CEED_EXTERN int CeedGetOperatorFallbackResource(Ceed ceed,
+    const char **resource);
+CEED_EXTERN int CeedSetOperatorFallbackResource(Ceed ceed,
+    const char *resource);
+CEED_EXTERN int CeedGetOperatorFallbackParentCeed(Ceed ceed, Ceed *parent);
```
jedbrown: Are these necessarily `Operator` fallbacks, or does `CeedGetFallbackResource` make sense?

jeremylt: Currently only Operators need this fallback idea, but I think that we could more generally use this fallback idea in the future.

jeremylt: Updated

PR #407, opened 2019-10-31
Travis - use xcode11.2

jeremylt:
Homebrew updating has been slowing down our Travis OSX builds by ~6 minutes. Switching to a newer OSX image helps fix this.


jeremylt:
That did the trick. We now are looking at Linux builds ~6-7 min and OSX builds ~10 min.

PR #409, opened 2019-11-01
CPU - avoid extra memcpy for identity qfunctions

jeremylt:
This avoids the extra memcpy in identity qfunctions for CPU backends and should improve the performance of our multigrid example.

Edit: it looks like it speeds up p-multigrid by ~1 sec with 20**3 7th order elements and coarsening by halves. This officially makes our p-multigrid faster in the KSP than no PC. The next performance hurdle is fixing `AssembleLinearDiag`, downstream of #406.


jeremylt:
I included a refactor of `CeedQFunctionCreateIdentity` so that we support more than emode `CEED_EVAL_INTERP`.

backends/blocked/ceed-blocked-operator.c
```diff
@@ -561,6 +588,12 @@ static int CeedOperatorAssembleLinearQFunction_Blocked(CeedOperator op,
   // Setup
   ierr = CeedOperatorSetup_Blocked(op); CeedChk(ierr);
 
+  // Check for identity
+  if (impl->identityqf)
+   // LCOV_EXCL_START
+    return CeedError(ceed, 1, "Assembling identity qfunction does not make sense");
+  // LCOV_EXCL_STOP
+
   // Input Evecs and Restriction
```
jeremylt: Note, this should be moved to `ceed-qfunction.c`.

jeremylt: Done

backends/blocked/ceed-blocked-operator.c
```diff
@@ -40,7 +40,9 @@ static int CeedOperatorDestroy_Blocked(CeedOperator op) {
 
   for (CeedInt i=0; i<impl->numeout; i++) {
     ierr = CeedVectorDestroy(&impl->evecsout[i]); CeedChk(ierr);
-    ierr = CeedVectorDestroy(&impl->qvecsout[i]); CeedChk(ierr);
+    if (!impl->identityqf) {
+      ierr = CeedVectorDestroy(&impl->qvecsout[i]); CeedChk(ierr);
+    }
```
jedbrown: This is okay, but perhaps we should have `qvecsout_alloc[]` so that destruction responsibilities are clear up-front?

backends/blocked/ceed-blocked-operator.c
```diff
@@ -198,12 +201,28 @@ static int CeedOperatorSetup_Blocked(CeedOperator op) {
...
+    ierr = CeedQFunctionGetFields(qf, &infields, &outfields); CeedChk(ierr);
+
+    for (CeedInt i=0; i<numinputfields; i++) {
+      ierr = CeedQFunctionFieldGetEvalMode(infields[i], &inmode);
+      CeedChk(ierr);
+      ierr = CeedQFunctionFieldGetEvalMode(outfields[i], &outmode);
+      CeedChk(ierr);
+
+      ierr = CeedVectorDestroy(&impl->qvecsout[i]); CeedChk(ierr);
+      impl->qvecsout[i] = impl->qvecsin[i];
```
jedbrown: Or, perhaps better, just increase reference count here?

backends/blocked/ceed-blocked-operator.c
```diff
@@ -561,6 +582,12 @@ static int CeedOperatorAssembleLinearQFunction_Blocked(CeedOperator op,
   // Setup
   ierr = CeedOperatorSetup_Blocked(op); CeedChk(ierr);
 
+  // Check for identity
+  if (impl->identityqf)
+    // LCOV_EXCL_START
+    return CeedError(ceed, 1, "Assembling identity qfunction does not make sense");
```
jedbrown: Maybe "not supported"? I don't see why it "doesn't make sense"; it could be implemented with a fast path some day.

backends/blocked/ceed-blocked-operator.c
jeremylt: Increased the refcount

backends/blocked/ceed-blocked-operator.c
jeremylt: In this case the qfunction is just a memcopy, so we could return a vector of 1s if desirable.

Updated the message.

PR #414, opened 2019-11-13
Python Interface

jeremylt:
Here is a PR to make discussions about the Python interface easier to follow.

ToDo:
- [x] Pick location for `libceed-qfunctions.c` tools
- [x] Double check all documentation
- [x] Convert to Pytest
- [x] Add CI - new Travis matrix option
- [x] Add OperatorApplyAdd


interface/ceed-python/ceed.py
```diff
@@ -26,26 +26,26 @@
 # Ceed Enums
 # ------------------------------------------------------------------------------
 # CeedMemType
-MEM_HOST       = lib.CEED_MEM_HOST
-MEM_DEVICE     = lib.CEED_MEM_DEVICE
+CEED_MEM_HOST       = lib.CEED_MEM_HOST
+CEED_MEM_DEVICE     = lib.CEED_MEM_DEVICE
```
jeremylt: It seems to me that this might as well match the globals in the other interfaces. Not sure why I used the shorter name at first.

interface/ceed-python/ceed_vector.py
```diff
@@ -162,3 +162,17 @@ def __init__(self, ceed, pointer):
...
+  def __init__(self):
+    # CeedVector object
+    self._pointer = [lib.CEED_VECTOR_ACTIVE]
+
+    # Reference to Ceed
+    self._ceed = None
+
+CEED_VECTOR_ACTIVE = _VectorActive()
+
+# ------------------------------------------------------------------------------
```
jeremylt: We'll need to do this for CEED_BASIS_COLLOCATED and CEED_VECTOR_NONE

interface/ceed-python/tests(will want to move eventually)/t100-vec.py
```diff
@@ -2,7 +2,7 @@
 # Test creation, setting, reading, restoring, and destroying of a vector
 
 import sys
-from libceed import MEM_HOST, USE_POINTER
+from libceed import CEED_MEM_HOST, CEED_USE_POINTER
```
jeremylt: Do we want to try to make sure it's reasonable to do `from libceed import *` if people want?

interface/ceed-python/ceed.py
jedbrown: Writing `ceed.CEED_MEM_HOST` feels clunky.

jedbrown: Also, does this parameter need to be explicit? Seems like it'd be a property of the array (NumPy or whatever). When getting an array, perhaps it should be specified (if/when we support an array container that can manage GPU arrays).

interface/ceed-python/tests(will want to move eventually)/t100-vec.py
jedbrown: No, I'd say that's an anti-pattern, and it'll be cumbersome when people follow reasonable practices.

valeriabarra: What if these constants were declared as members inside the class `Ceed`? Would `Ceed.MEM_HOST` still feel clunky? In that case, we would't need to import each of the constants in every single file

jeremylt: It should work to do `libceed.MEM_HOST', right?

jedbrown:
```python
from libceed import Ceed
ceed = Ceed('/cpu/self')
qf = ceed.QFunction(...)
qf.add_input('foo', 3, libceed.EVAL_INTERP)
```
doesn't work because `libceed` is not defined. This would work, and I think is entirely acceptable and simplest to implement:
```python
import libceed
ceed = libceed.Ceed('/cpu/self')
qf = ceed.QFunction(...)
qf.add_input('foo', 3, libceed.EVAL_INTERP)
```

Alternatively, we could make it an attribute of the relevant class/classes, e.g.,
```python
qf.add_input('foo', 3, ceed.EVAL_INTERP)
```
or
```python
qf.add_input('foo', 3, qf.EVAL_INTERP)
```

interface/ceed-python/ceed_constants.py
```diff
@@ -0,0 +1,85 @@
...
+                  PYRAMID:  "ceed_pyramid",
+                  PRISM:    "ceed_prism",
+                  HEX:      "ceed_hex"}
+
+# ------------------------------------------------------------------------------
+# Ceed Constants
+# ------------------------------------------------------------------------------
+# Requests
+REQUEST_IMMEDIATE = lib.CEED_REQUEST_IMMEDIATE
+REQUEST_ORDERED = lib.CEED_REQUEST_ORDERED
```
jeremylt: @jedbrown, in the future, will users need to make request objects beyond `REQUEST_IMMEDIATE` and `REQUEST_ORDERED`? I am trying to decide if we need to wrap these in objects for future compatibility. We have to wrap the vectors below to get the interface to work nicely.

jedbrown: It might be possible to expose requests as futures, but I would assume `IMMEDIATE` for this PR. https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future

interface/ceed-python/ceed.py
jeremylt: We can set MEM_HOST as the default and then only call with MEM_DEVICE when we have an option to do so

jedbrown: Yeah, keyword arguments are the way to go, and will make the Python syntax a lot cleaner than we can do in C.

interface/ceed-python/libceed/ceed_constants.py
```diff
@@ -80,15 +79,23 @@
...
 REQUEST_IMMEDIATE = lib.CEED_REQUEST_IMMEDIATE
 REQUEST_ORDERED = lib.CEED_REQUEST_ORDERED
 
+# Object shell
+class _CeedConstantObject(ABC):
+  """Shell for holding constant Vector and Basis constants."""
+
+  def __init__(self, constant):
+    self._pointer = [constant]
+
```
jeremylt: I just realized that this is a cleaner way to wrap these constants we use

interface/ceed-python/libceed/ceed_constants.py
```diff
@@ -94,6 +94,7 @@ def __init__(self, constant):
 # Vectors
 VECTOR_ACTIVE = _CeedConstantObject(lib.CEED_VECTOR_ACTIVE)
 VECTOR_NONE = _CeedConstantObject(lib.CEED_VECTOR_NONE)
+VECTOR_NULL = _CeedConstantObject(ffi.NULL)
```
jeremylt: What about a more general purpose `libceed.NULL`? We use null arguments in a few places.

jedbrown: Keyword argument defaulting to `None`. Let's not define our own thing.

valeriabarra:
I was about to define the default argument to NULL, but I would have had to
swap input and output arguments order, to put that last, and I thought it
was not a good idea to swap. I thought it would be confusing for the user
because it would result the opposite of the C version of the function.

jeremylt:
Really, we should modify the C/Fortran API to accept CEED_VECTOR_NONE here. It should be a valid input vector for BasisApply with CEED_EVAL_WEIGHT. Then we don't need anything different here. I can fix it in the morning.

interface/ceed-python/ceed_cffi/build-ceed.py
```diff
@@ -0,0 +1,230 @@
...
+    CeedOperator subop);
+  int CeedOperatorAssembleLinearQFunction(CeedOperator op,
+    CeedVector *assembled, CeedElemRestriction *rstr, CeedRequest *request);
+  int CeedOperatorAssembleLinearDiagonal(CeedOperator op,
+    CeedVector *assembled, CeedRequest *request);
+  int CeedOperatorApply(CeedOperator op, CeedVector in,
+                                  CeedVector out, CeedRequest *request);
+  int CeedOperatorDestroy(CeedOperator *op);
+  """
+)
```
jedbrown: Is there a reason we can't obtain the above by reading `ceed.h`? See `f.read()` patterns at https://cffi.readthedocs.io/en/latest/embedding.html#embedding-and-extending or https://dbader.org/blog/python-cffi

jeremylt: Huh. I wonder if that is new(er), as much of the online discussion of cffi states that manually imported the header is(was) required.

interface/ceed-python/tests/t202-elemrestriction-py.py
```diff
@@ -0,0 +1,34 @@
...
+
+  y = ceed.Vector(2*blksize*2)
+  y.set_value(0)
+
+  r.apply(libceed.NOTRANSPOSE, libceed.NOTRANSPOSE, x, y)
+
+  print(y)
+
+  x.set_value(0)
+  r.apply(libceed.TRANSPOSE, libceed.NOTRANSPOSE, y, x)
```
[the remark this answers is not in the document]
jeremylt: `r` still needs to know if its going from lvec to evec or evec to lvec. We should probably make the lmode and tmode named fields for clarity.

At this time, libCEED vectors don't really have a 'transpose' property, so I'm not sure if we want to/how to add a property to the python API that the base library doesn't have.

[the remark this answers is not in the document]
jedbrown: Hmm, well suppose we were using NumPy and had restriction as a matrix `R`. Then we would write `evec = R @ lvec` or `lvec += R.T @ evec`. Making new objects to express transposes is cumbersome in C, but easy in Python. I think `R.T.apply(evec, lvec)` would be cleaner and more Pythonic than flags.

We should probably eventually implement `evec = R @ lvec` to automatically create a vector of the appropriate size, with `R.apply(lvec, evec)` being the way to reuse memory. (Most users of Python won't care that much about precise memory accounting and want, first and foremost, the cleanest possible interface.)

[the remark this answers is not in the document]
jeremylt: I added `r.T.apply(...)`. `lmode` is still a keyword arg right now, but assuming `NOTRANSPOSE`.

jeremylt:
Next on my list is getting the QFunctions and Operators to work. I'd really like to this branch with a working Python interface before SC19 if we can.

jeremylt:
I'll add the QF and Op prints in the morning.

jeremylt:
To Run t500:
1) Build libCEED
2) in `interface/ceed-python/ceed_cffi`
```
python setup.py build_ext
python setup.py develop --user

```
3) in `interface/ceed-python/libceed`

```
python setup.py build
python setup.py develop --user

```
4) in `interface/ceed-python/tests`

```
python setup-qfunctions.py build
python setup.py install --user

```
5) copy `build/lib.*/libceed_qfunctions.cpython*.so` to `qfs.so`
6) run `python t500-operator-py.py /cpu/self`

jeremylt:
I'll investigate the failing CI tomorrow. I'll probably fix it by smashing this down to 1 python commit + some small library fix commits. We can keep polishing from there.

valeriabarra:
> I'll investigate the failing CI tomorrow. I'll probably fix it by smashing this down to 1 python commit + some small library fix commits. We can keep polishing from there.

I would preserve quite a bit more than a python commit

jeremylt:
If you have a commit structure in mind, you are welcome to tackle it. I don't have the time to rebase 177 commits into something cogent. I was thinking of smashing it so we both get attribution. We could just smash on the merge.

jeremylt:
I updated the ElemRestriction/Basis transpose applies to avoid mutation, as disussed

jeremylt:
I update the CFFI build script to open ceed.h and exclude the appropriate lines. We may have to occasionally update the exclusions, but this seems much easier to maintain that manually copying and pasting.

jeremylt:
The big open ToDos are:
1) Update/finish test suite
2) Finish documentation
3) Decide on file locations, build process
4) CI

jeremylt:
I take that back, my cdef approach was horribly broken

jeremylt:
And its fixed now

jeremylt:
QF single source build slightly updated. According to PEP, distutils *must* append lots of extra data into the filename, so we will always get `libceed_qfunctions.textwedontcareabout.so`

jeremylt:
I think it can be cleaned up, but I added `make python` at the top level.

interface/ceed-python/Makefile
```diff
@@ -0,0 +1,30 @@
...
+# software, applications, hardware, advanced system engineering and early
+# testbed platforms, in support of the nation's exascale computing imperative.
+
+PYFLAGS ?=
+
+ceed-cffi:
+	python setup-ceed-cffi.py build_ext install $(PYFLAGS)
+
+libceed-python:
+	python setup-libceed.py build install $(PYFLAGS)
```
jeremylt: I added this `PYFLAGS` variable to add install options like `--user`. I'm not sure exactly what the convention is, as usually folks don't call `setup.py` for modules via make

jeremylt:
Updated ToDo at the top with the remaining items. Everything works, so its just a matter of polish and testing.

We should merge #423 first

jeremylt:
Failing CI issue fixed in https://github.com/CEED/libCEED/pull/423/commits/70dc8cf2bd34a7cdb108a64db8b4531da9c5c3b6

valeriabarra:
I think we should rename the `interface/ceed-python` folder just `interface/python`. Any objections? Should I go ahead and rename it?

valeriabarra:
> I think we should rename the `interface/ceed-python` folder just `interface/python`. Any objections? Should I go ahead and rename it?

As well as `tests/python-tests` only `tests/python`.

jeremylt:
I made the folder `ceed-python` to match the other files in that folder. WRT the tests, I think that's a reasonable change.

jeremylt:
Travis runs `make pytest` and the build runs successfully. We just need the fixes in #423 and this branch will pass Travis.

jeremylt:
We need to add OperatorApplyAdd. It should be small/quick.

jeremylt:
Any strong feelings about including Python 2 in the CI matrix? I left it off since its on its last legs, but some folks use it a fair amount.

jedbrown:
I don't think we should bother with python2. It's EOL in a month.

jedbrown:
I think this is looking great. It would be nice to support `python setup.py install --prefix=/path/to/prefix`. This is presently awkward through `make PYFLAGS=--prefix=/path/to/prefix` and must be re-specified to run Python tests.

In practice, I like to use `python setup.py develop --prefix=...` so that I don't need to reinstall to test each incremental change. The same can be done using `pip install -e`.

I think it would be okay to have a top-level directory called `python`, containing the Python interfaces.

Although it can be in a subsequent PR, we should index in PyPI so that `pip install libceed` will work. PETSc is installable that way (see its top-level `setup.py`), though I don't know if it follows current best practices.

jeremylt:
Moved to a top level folder named `python`

jeremylt:
@jedbrown Now if the user sets `prefix`, then it will automatically be added to the pyflags.

I think we are ready to merge. @valeriabarra, what do you think?

The Travis failure is due to an unrelated OCCA issue with the OpenCL backend.

I intentionally left the advanced operator functions (AssembleLinearQF, etc) out for now since those interfaces are in flux.

valeriabarra:
> @jedbrown Now if the user sets `prefix`, then it will automatically be added to the pyflags.
> 
> I think we are ready to merge. @valeriabarra, what do you think?
> 
> The Travis failure is due to an unrelated OCCA issue.
> 
> I intentionally left the advanced operator functions (AssembleLinearQF, etc) out for now since those interfaces are in flux.

I can do a pass on all documentation and check that last checkbox off. Then, I believe we are good to go. But I feel I should not be a major reviewer of this PR since I was also a main contributor...

We can leave the implementation and optimization of user-python-code of QFunctions for another PR. And I can start working on an outreach/tutorial jupyter notebook PR soon.

jeremylt:
We have *a* way to do user defined Qfunctions via distutils. The example is in the test folder.

valeriabarra:
> We have _a_ way to do user defined Qfunctions via distutils. The example is in the test folder. I'd like to support more in the future, but we do support the single source files, which was important to me.

Yes, I was aware of that, and I agree. I was just saying that if we want to support the user choice to use Numba, we could leave it for another PR. 

I will do the pass through the doc now.

jeremylt:
Ah, I missread. Multi-tasking is hard.

valeriabarra:
Done! My only question is: do we want to write anything more in the README?
LGTM otherwise, and I will wait for someone else's review.

jeremylt:
I expanded the readme a bit since the test suite is currently our only examples of use.

jeremylt:
Testing updated for master and set Travis to run pytest in a verbose mode.

jeremylt:
There is a bizarre Travis error only on the push, not on the PR.
https://travis-ci.org/CEED/libCEED/jobs/622909093?utm_medium=notification&utm_source=github_status
I'm not sure what the deal is, but I'm passing make prove-all locally , this brand passes when merged into master, and this PR does not touch anything outside of the python files/folders, so this is probably not a serious issue but rather some bizarre bug.

jedbrown:
1. We have some commits that add binary (`*.pyc`) files and subsequently removes them. Those pairs of commits should be squashed together. It'd be okay to squash-merge if you prefer.
2. `make python` should not "install" outside the working directory. We could use virtualenv for testing. There should be a `requirements.txt` file. I think `setup.py` should go at the top level; my understanding is that is necessary for `pip install`, which we'll want.
3. The instructions need to work without write access to `/usr/local/`.

jeremylt:
1) I was thinking we should squash+merge. 260 in progress commits would be difficult to convert into a handful of smaller commits unless we rebuild the branch.

jeremylt:
@jedbrown, re 2) and installing in the working directory, do you mean that the default behavior should be using the `--user` option?

jedbrown:
Squashing is fine with me, but remember to give attribution to both of you since only one can be the author of the squashed commit.

The global install is still an issue, and shouldn't be required for pytest. I don't have a problem with postponing `pip install` support, but we'll want it eventually. It looks like [this](https://github.com/bast/python-cffi-demo) is a good demo that uses pytest and cffi with a native library (built using CMake, but easy to substitute in our `make install prefix=...`). [This project](https://github.com/angr/angr) is a complete package that follows best practices as I understand them. (PETSc and petsc4py are split, and petsc4py uses cython instead of cffi, so it isn't a great example.) See also [packaging guidelines](https://packaging.python.org/tutorials/packaging-projects/) for what `setup.py` needs to be pip-installable. Ideally, we should check that the following works:
```
virtualenv TEST
source TEST/bin/activate
pip install git+https://github.com/ceed/libceed@python/interface-v1
python -c 'import libceed'
```

jedbrown:
`--user` is still an install. Suppose I have a libceed installed that I'm using in a different project. I check out a branch in some clone of libceed and run its tests. That shouldn't break my install. See the demo repository linked above for what seems like a sane workflow to me.

jeremylt:
I lifted `setup.py` to the top level. I think I can get `setup.py` to work like angr soon.

jeremylt:
Getting closer. `python setup.py build install` works fine but `pip install` is a mess.

jeremylt:
setup works, pip is still a mess.

Here's the issue: 
Pip makes a temp copy of the library somewhere, builds libCEED and the python module, and erases the temp folder (and thus the library and it's .so)
CFFI needs to know where to find `libCEED.so`, but Pip deleted it.

I'm trying to bully Pip into building libCEED in place and building the python module however it feels like. When I build libCEED and pip install the python module separately, everything works fine so long as `libceed.so` can be found at runtime, added to `LD_LIBRARY_PATH`.

Short: I need Pip to keep a persistent copy of the library and CFFI needs to know where to find `libceed.so`

jeremylt:
I have had some success at making pip copy the `.o`, `.so`, ect files we need, but I've had trouble getting CFFI to find these files

jeremylt:
I have a setup that will work so long as I can get the final install dir.

CFFI needs
```
  runtime_library_dirs=[os.path.abspath("path/to/final/lib/dir")] # library path, at runtime
```

jeremylt:
Getting pip to work is going to take some effort. I have what I think is a clean setup for only using `python setup.py foo`.  ]I removed the makefile commands as I don't think they add much. Do we want to merge this and fight pip later or continue to fight pip now?

jeremylt:
We're down to 20 commits. The original commit history of ~300 commits can be found here: https://github.com/CEED/libCEED/commits/python/interface-v1-original

valeriabarra:
Are we all good on this PR now?

jeremylt:
@valeriabarra, I'm happy to merge now if @jedbrown finds this improved install/build process adequate.

`pip install .` now works in the top level folder, so long as libCEED is installed at the location provided by the env variable `CEED_DIR`

jeremylt:
I updated Travis to use `pip` and updated the README.

README.md
```diff
@@ -239,6 +239,13 @@ To install libCEED for Python, run
 
     python setup.py build install
 
+with the desired setuptools options, such as `--user`.
+
+Alternatively, if libCEED in installed in the directory specified by environment
```
valeriabarra: Is installed...
And
... specified by the environment variable

jeremylt: Thanks

PR #415, opened 2019-11-13
Tests - add t106 SyncArray and t107 VectorView

jeremylt:
Adds simple tests for interface calls - helps with tutorials


jeremylt:
Passes clean on a GPU node of CU's cluster.

PR #417, opened 2019-11-16
None Args

jeremylt:
This PR removes our use of  `NULL` in the API in a few places in favor of `CEED_VECTOR_NONE` and `CEED_QFUNCTION_NONE` for clarity and typechecking. This will also help the python API.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-qfunction.c
+++ b/backends/cuda-gen/ceed-cuda-gen-qfunction.c
@@ -60,7 +60,7 @@ static int loadCudaFunction(CeedQFunction qf, char *c_src_file) {
   char *buffer;
 
   fp = fopen ( cuda_file, "rb" );
-  if( !fp ) CeedError(ceed, 1, "Couldn't open the Cuda file for the QFunction.");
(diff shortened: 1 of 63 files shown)


jeremylt:
CodeCov drop comes from the fact that this exposes that we aren't using `dqf` or `dqfT` at this time.

backends/cuda-gen/ceed-cuda-gen-qfunction.c
```diff
@@ -107,7 +107,7 @@ int CeedQFunctionCreate_Cuda_gen(CeedQFunction qf) {
 
   char *source;
   ierr = CeedQFunctionGetSourcePath(qf, &source); CeedChk(ierr);
-  const char *funname = strrchr(source, ':') + 1;
+  cons char *funname = strrchr(source, ':') + 1;
```
jedbrown: ?
This was clearly missed by testing.

jeremylt: Woops, I'll fix it in a small PR here in a bit

PR #421, opened 2019-11-18
Add Ceed_Cuda struct to Ceed_Cuda_ref/shared/gen.

SanderA:
While running with the backends /gpu/cuda/shared and /gpu/cuda/gen I got the CUDA_ERROR_INVALID_DEVICE error.
This was basically caused by casting the empty Ceed_Cuda_ref/shared/gen structs to the Ceed_Cuda struct and trying to access the deviceId member.

This PR adds the Ceed_Cuda struct to the Ceed_Cuda_ref/shared/gen structs so that the latter can be upcast to the former.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -828,10 +828,8 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
 
   // Add atomicAdd function for old NVidia architectures
   struct cudaDeviceProp prop;
-  Ceed delegate;
-  CeedGetDelegate(ceed, &delegate);
   Ceed_Cuda *ceed_data;
-  ierr = CeedGetData(delegate, (void **)&ceed_data); CeedChk(ierr);
+  ierr = CeedGetData(ceed, (void **)&ceed_data); CeedChk(ierr);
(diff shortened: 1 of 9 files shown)


jedbrown:
:wave: Thanks for this PR!  How would one reproduce the error condition you encountered? We'd like to make sure such things are tested.

SanderA:
I got these by just running the PETSc benchmarks. LibCeed was installed through PETSc. I can send the configure.log, but I'm not sure where to post it?

jeremylt:
Side note: the PETSc benchmarks will not look great for the CUDA backends until we merge 294

jedbrown:
You can attach files by drag & drop into this thread. Just steps to reproduce would be fine for now. More need for #219.

SanderA:
I can get the error by running make petscexamples && cd benchmarks && ./benchmark.sh --run petsc-bpsraw.sh -n 1 -p 1 -c /gpu/cuda/gen -b bp3.

jeremylt:
@YohannDudouit, thoughts on this PR?

jedbrown:
@YohannDudouit Could you please look at this? We'd like to resolve this issue (if still present) before release.

YohannDudouit:
I don't manage to reproduce the issue, the change looks reasonable to me, it mostly seems to be factorizing code to avoid code duplication.

PR #424, opened 2019-11-22
Enable MAGMA non-tensor basis

nbeams:
This pull request contains updates related to adding non-tensor basis capability to the MAGMA backend.

- Adds new routines for non-tensor basis actions
- Adds MAGMA element restrictions using a different E-vector layout.  This requires whitelisting some element restriction tests (t204-7)
- Updates tensor basis actions to use new layout
- Also updates tensor basis actions to remove loops over component 

Current concerns for all reviewers:
- Is the method of skipping the element restriction tests, output message, etc. OK? (in tap.sh)
- Other style/interface change requests
- Tests t412, t532, and t534-6 currently fail.  However, they also fail when I run with the cuda/ref backend, so I am not sure if the failure is due to any changes made in this PR.  Can someone confirm or deny any known issues for these tests?

Specific for MAGMA team: 
- Element restriction create function - further tests and/or return to MAGMA calls as before


jeremylt:
`tap.sh` isn't my favorite place to make lots of exceptions, but since you need to expose `CeedElemRestrictionApply` for the delegate CUDA backend, this is probably fine.

I'm investigating the t53* issues

jeremylt:
One issue with using a version of ElemRestrictionApply that doesn't pass the test suite - our delegation to manage building preconditioning operators won't work for the MAGMA backend

nbeams:
@jeremylt, can you please elaborate a bit more about the preconditioning operators?  Does this refer to something in a current branch/PR?  Or something currently in libCEED that I'm not aware of?

jeremylt:
We have fallback to the CPU ref backend to assemble the Qfunctions as a sort of sparse matrix, to assemble the Operator diagonal, and (soon) to assemble an element inverse based on Fast Diagonalization. I to have to fix a big in the implementation for the CUDA backend, but you can see the implementation working for the OCCA backend.

This is based on restrictions and bases that pass the test suite as written. Any backend that doesn't pass the test suite would have to implement the functionality or indicate that they do not support these functions.

nbeams:
Ah, I see.  So we'd have the option to implement `AssembleLinearDiagonal` and `AssembleLinearQFunction` for MAGMA (and would have to implement `CeedOperatorCreate` in order to properly set these functions, with some sort of situation like the fallback operator you added).  

But at the moment in the element-fdm-inverse branch, it doesn't seem like `CreateFDMElementInverse` is available for the backends to optionally override, like the two AssembleLinear* functions are. Is that correct?  I realize this is a feature branch and perhaps still somewhat in flux, so would there be future plans to allow backends to override the FDM inverse creation?  Or would we be forced to have an error if someone tries to call that function with the MAGMA backend?

jeremylt:
There is some work coming up to refractor that branch and let backends implement their own of all three. This is an area of active development. They all rely upon Qfunction assembly, so that's really the tricky part.

nbeams:
I noticed after merging in some changes from master yesterday that now the previously-failing t53* tests are skipped with a "not implemented" message.  I assume this comes through the cuda/ref operator code since we are using that in MAGMA.

jeremylt:
Correct. I dropped the fallback mechanism after seeing that the issue was rather complex to fix.

jeremylt:
I haven't forgotten about this PR - I'm in between some personal travel for the holidays. I will be able to test/review next week.

nbeams:
No worries!  I was wondering about this PR in light of our discussion on #425 -- would we want to go ahead and merge this with the required t20* whitelisting, then add the tests back for the MAGMA backend if the end result of 425 is a setup where MAGMA could pass those tests again?

backends/magma/ceed-magma-restriction.c
```diff
@@ -0,0 +1,202 @@
...
+        magma_setvector(size, sizeof(CeedInt), indices, 1, impl->dindices, 1);
+      }
+      break;
+    case CEED_USE_POINTER:
+      if (indices != NULL) {
+        ierr = magma_malloc( (void**)&impl->dindices,
+                             size * sizeof(CeedInt)); CeedChk(ierr);
+        // TODO: magma calls
+        //ierr = cudaMemcpy(impl->dindices, indices, size * sizeof(CeedInt),
+        //                  cudaMemcpyHostToDevice);
```
jeremylt: Dead code

tests/tap.sh
```diff
@@ -121,6 +121,15 @@ for ((i=0;i<${#backends[@]};++i)); do
...
     fi
 
+    # grep to skip t204-7 for MAGMA (different layout used for E-vectors)
+    if [[ "$backend" = *"magma" ]] \
+            && [[ "$1" = "t20"[4-7]* ]] ; then
+        printf "ok $i0 # SKIP - backend uses different E-vector layout $1 $backend\n"
+        printf "ok $i1 # SKIP - backend uses different E-vector layout $1 $backend stdout\n"
+        printf "ok $i2 # SKIP - backend uses different E-vector layout $1 $backend stderr\n"
+        continue
+    fi
```
jeremylt: I think that this is fine for now, but we'll want a better long term strategy #425

nbeams:
Thanks @jeremylt.  You're wanting to wait for @YohannDudouit's approval before merging, right?

jeremylt:
Right

nbeams:
@YohannDudouit, do you think you'll have time to review this soon?  I'd like to get it merged before everyone leaves for vacation, if possible.

jeremylt:
It worked for me. If we want to merge this now, I can rebase and adjust #447.

jedbrown:
Your preference on order. I approved #447.

jedbrown:
@nbeams @stomov @abdelfattah83 Is it okay if we squash-merge this PR? It has a lot of intermediate commits. Alternatively, you can use `git rebase -i` to squash them into roughly bisectable commits to preserve authorship. If we squash, we'll include `Thanks-to` or `Co-developed-by` tags for Stan and Ahmad.

nbeams:
I tried several times to come up with a mostly-squashed history using `git rebase -i`.  Most times I got the "Cannot squash without previous commit" error, I guess because the commit history is rather muddled with merges from master?  One time it seemed like it was going to let me complete the rebase, but when it gave me the combined commit message to edit, the commit messages it listed did not correspond at all to the ones I'd picked... (They were all commit messages from things merged from master, with none of the magma-related ones I'd picked.)  I don't really have much experience with squashing this way, so maybe someone else can get something to work with a fairly clean history (and let me know if it seems obvious what I was doing wrong from this description).

Obviously, it's ultimately up to Stan and Ahmad.

stomov:
Squash-merging sounds good to me. Thanks.

PR #428, opened 2019-12-06
Small doc fixes

valeriabarra:
While revising our Python interface documentation in PR #414 I found some small documentation inconsistencies. This is a small PR to improve the documentation of the C API.


valeriabarra:
Travis is failing due to OCCA, but nothing related to this commit, which is only documentation

PR #430, opened 2019-12-09
[WIP] jrwrigh/ns_dmplex

jrwrigh:
WIP PR for adding error calculation to the `advection` problem of the navier-stokes example, specifically dealing with nsplex.

TODO List copied from @jedbrown's [comment](https://github.com/CEED/libCEED/pull/430#issuecomment-564710930):

* [x]  Change `ctxSetup` to a struct in `nsplex.c` and inside each `ICs*`
* [x]  Add `time` to `ctxSetup` and flag in `problemType` indicating that nonzero time is supported
* [x]  Enable smooth initial conditions in advection and/or advection2d
* [x]  Split `op_ics` and multiplicity correction to PETSc global Vec into callable function with parameter `time`
* [x]  Implement nonzero time for advection and call after `TSSolve` (with `TSGetTime`) when nonzero time is supported
* [x]  Use `VecAXPY(Q, -1.0, Qexact); VecNorm(Q, NORM_MAX, &norm)` and print when nonzero time is supported
* [ ]  Convergence study: order of accuracy under mesh refinement for each polynomial degree, execution time, etc. in Pandas DataFrame like below:

| degree | delta_x | ndofs | num_time_steps | max_error | execution_time | num_procs |
| --- | --- | --- | --- | --- | --- | --- |

### DIFF
--- a/examples/navier-stokes/navierstokes.c
+++ b/examples/navier-stokes/navierstokes.c
@@ -152,7 +152,6 @@ static int CreateRestriction(Ceed ceed, const CeedInt melem[3],
 
 // PETSc user data
 typedef struct UserNS_ *UserNS;
-typedef struct Units_ *Units;
 
 struct UserNS_ {
   MPI_Comm comm;


jrwrigh:
Main roadblock right now is getting `*.vtu` files to be written for the final iteration/solution. Any help in this area would be appreciated, as I can't find a simple way to get it done. 

Technically I could set `-output_freq` to 1, but I'm afraid that would mess with comparisons between the different fidelity simulations (those that must have less aggressive time steps will be "punished" more than those that don't).

jedbrown:
1. You can call `TSMonitor_NS` directly, as is done for the initial condition.
2. The VTU output doesn't preserve the full function space, so you're just evaluating at vertices. That's okay-ish for a max-norm error, but poor for an L^2 error. You can compute L^2 error by integration. See `examples/petsc/` for an example.

jedbrown:
Another reason to print errors as part of the normal run is that we would like to add accuracy tests in continuous integration. For advection, I would just rotate a smooth cosine bump by some angle theta and report error.

For Euler (zero viscosity NS), I would use Example 2.2 from the attached and let it move a small distance.

[ZhangZhangShu-OrderOfAccuracyAndNumericalPerformanceTwoClassesFiniteVolumeWENO-2011.pdf](https://github.com/CEED/libCEED/files/3938042/ZhangZhangShu-OrderOfAccuracyAndNumericalPerformanceTwoClassesFiniteVolumeWENO-2011.pdf)

The philosophy is that we run a grid refinement test, observe design order of accuracy (hopefully), and then dial the resolution and duration back to something that runs fast for CI.

jrwrigh:
I've copied over the functions and corrected the differences in variable names. 

Only major architecture change I made was moving the `User` and `Units` struct definition to `error_header.h`, as they're required by the `ComputeErrorMax` function. 

Next step I'm not sure on is where to define the values of the `target` vector. I believe @jedbrown  had mentioned at some point making another qfunction to populate that vector, or defining it in the `Error` qfunction (already in `error_header.h`). 

For calculating the exact solution, I'm imagining doing something similar to my python script; have a "setup" where I calculate the center of the bubble (which is a function of the final time) and an "evaluate" which actually returns the true solution (which is a function of the bubble center and the quadrature coordinate). How would I get the quadrature coordinate location?

examples/navier-stokes/nsplex.c
```diff
@@ -207,37 +208,37 @@ static int VectorPlacePetscVec(CeedVector c, Vec p) {
...
+//   PetscScalar Kelvin;
+//   // derived units
+//   PetscScalar Pascal;
+//   PetscScalar JperkgK;
+//   PetscScalar mpersquareds;
+//   PetscScalar WpermK;
+//   PetscScalar kgpercubicm;
+//   PetscScalar kgpersquaredms;
+//   PetscScalar Joulepercubicm;
+// };
```
jedbrown: Just delete this code if you're moving it to a common header.

examples/navier-stokes/nsplex.c
```diff
@@ -868,6 +869,31 @@ int main(int argc, char **argv) {
...
 
+  // Copied from ../petsc/bps.c
+  // Create the error Q-function
+  CeedQFunction qf_error;
+  CeedOperator op_error;
+  CeedVector target;
+  CeedQFunctionCreateInterior(ceed, 1, Error,
+                              Error_loc, &qf_error);
+  CeedQFunctionAddInput(qf_error, "q", ncompq, CEED_EVAL_INTERP);
+  CeedQFunctionAddInput(qf_error, "true_soln", ncompq, CEED_EVAL_NONE);
```
jedbrown: The true solution depends on time, which we'll pass in through the context. One way to leverage code is to add `time` to the context for initial conditions, and have that error when `time != 0.0` unless it's supported. In that case, you'd call initial conditions with this new time and collocated output to define the exact solution, then use this differencing to make it an error, also collocated at quadrature points.

jrwrigh: What do you mean when you say the "context"? Is that a struct somewhere?

jedbrown: See `CeedQFunctionSetContext`. It's how we pass in parameters like viscosity and gravitational acceleration.

jedbrown:
- [ ] Change `ctxSetup` to a struct in `nsplex.c` and inside each `ICs*`
- [ ] Add `time` to `ctxSetup` and flag in `problemType` indicating that nonzero time is supported
- [ ] Enable smooth initial conditions in advection and/or advection2d
- [ ] Split `op_ics` and multiplicity correction to PETSc global Vec into callable function with parameter `time`
- [ ] Implement nonzero time for advection and call after `TSSolve` (with `TSGetTime`) when nonzero time is supported
- [ ] Use `VecAXPY(Q, -1.0, Qexact); VecNorm(Q, NORM_MAX, &norm)` and print when nonzero time is supported
- [ ] Convergence study: order of accuracy under mesh refinement for each polynomial degree, execution time, etc. in Pandas DataFrame like below

| degree | delta_x | ndofs | num_time_steps | max_error | execution_time | num_procs |
|----|----|----|----|----|----|---|

jrwrigh:
All the libCEED side of the code is now done and successfully running. However, I'm getting some unexpected phenomenon.

1) Solution diverges in 1 time step for grids greater that ~ 40x40x1 for `-petscspace_degree 1` and ~20x20x1 for `-petscspace_degree 2`. I wasn't seeing this behavior before. This still occurs when switching back to the original IC as well (`continutityBubble=-1` in `advection.h`).

2) For the low resolution meshes I can run, `-petscspace_degree 1` reports an error ~1 for a 10x10x1 grid, but for `-petscspace_degree 2`, it's reporting `24.` (as in exactly 24 for double printing). Possibly a consequence of the low mesh resolution, but still kind of weird.

jedbrown:
I pushed a couple fixes to your branch. Try
```
 ./nsplex -ts_monitor -petscspace_degree 3 -dm_plex_box_faces 40,40 -ts_adapt_dt_max .01 -ts_max_time 1 -problem adve
ction2d -dm_view -CtauS .3 -implicit -snes_mf -ksp_atol 1e-5 -snes_type ksponly -ts_view -stab none -ts_type rosw
[...]
Max Error: 0.00283378
```

* As the spatial discretization gets more accurate (from grid refinement or increased polynomial degree), you'll need to also make the time integrator more accurate. This can be done by increasing its order of accuracy or by reducing the time step (either by shrinking `-ts_atol` and `-ts_rtol` or by fixing a maximum time step `-ts_adapt_dt_max`).
* The default time integrator is BDF2. You "could" use `-ts_bdf_order 3` to increase the order of accuracy, but this won't because transport has near-imaginary eigenvalues. BDF3 and higher are not stable on the imaginary axis. So if you stick with BDF2, you'll need to use smaller steps when refining.
* For Galerkin (and SU, but not SUPG), you can use explicit integrators. Otherwise, `-ts_type rosw` and `-ts_type arkimex` have A-stable methods.
* As spatial and (implicit) temporal integration get more accurate, iterative solver tolerances will need to get tighter.

examples/navier-stokes/nsplex.c
```diff
@@ -722,13 +741,19 @@ int main(int argc, char **argv) {
   {  // Print grid information
     CeedInt gnodes, onodes;
     int comm_size;
+    char box_faces_str[PETSC_MAX_PATH_LEN];
```
jrwrigh: Is there a better size to use here other than `PETSC_MAX_PATH_LEN`?

jrwrigh:
This should be ready to pull. The only issues from the testing (that I can see) are with `navierstokes.c`. (mainly just the fact that I moved the `User` struct to `common.h`, so there's a redefinition error).

jeremylt:
I'll add the `1-In Review` tag once CI is passing.

jeremylt:
WRT the build errors you're seeing, its more complicated than just redefinition.

I believe the `Units_` struct is the same between the two codebases. If so, the redundant definition in `navierstokes.c` can be simply removed.

The two versions of the `User_` struct are different. To prevent the build errors, you'd need to reconcile the differences between the structs or give them separate names. We're eventually going to drop `navierstokes.c`, so renaming is probably the easiest fix.

examples/navier-stokes/nsplex.c
```diff
@@ -1028,10 +1034,35 @@ int main(int argc, char **argv) {
...
 
   // Solve
+  start = clock();
   ierr = TSSolve(ts, Q); CHKERRQ(ierr);
+  ierr = TSGetSolveTime(ts,&ftime); CHKERRQ(ierr);
+  end = clock();
+  cpu_time_used = ((double) (end - start)) / CLOCKS_PER_SEC;
+  ierr = PetscPrintf(PETSC_COMM_WORLD,
+                    "Time taken for solution: %g\n",
+                    (double)cpu_time_used); CHKERRQ(ierr);
```
jeremylt: If you want to time the solution, I'd use lines `267-272` of `bps.c` in the `examples/petsc` folder.

examples/navier-stokes/nsplex.c
```diff
@@ -892,6 +917,7 @@ int main(int argc, char **argv) {
     if (qf_ifunction) CeedQFunctionSetContext(qf_ifunction, &ctxAdvection2d, sizeof ctxAdvection2d);
   }
 
+
```
jeremylt: Stray line added

examples/navier-stokes/nsplex.c
jeremylt: I think this should be fine.

examples/navier-stokes/nsplex.c
jrwrigh: Fixed in 043888d

examples/navier-stokes/nsplex.c
jrwrigh: Fixed in 6fc7875

jrwrigh:
`navier-stokes.c` now builds in CI. I believe the current problems are now dealing with the the test suite. Any guidance on how to adjust the stdout tests would be helpful.

jeremylt:
In case you can't see the error message: 
image

Also, it would be nice to remove these warnings:
image

examples/navier-stokes/nsplex.c
```diff
@@ -722,13 +739,19 @@ int main(int argc, char **argv) {
   {  // Print grid information
     CeedInt gnodes, onodes;
     int comm_size;
+    char box_faces_str[PETSC_MAX_PATH_LEN]="NONE";
```
jeremylt: Our typical style would be `char box_faces_str[PETSC_MAX_PATH_LEN] = "NONE";`

jeremylt:
The output files the test suite checks against are found here: https://github.com/jrwrigh/libCEED/tree/jrwrigh/ns-dmplex/tests/output

jrwrigh:
Yeah, I saw the error messages. I can see that it's doing a `diff` between some file called `ns-nsplex.out` and the STDOUT of the program, but I don't see the file in `tests/output` like it says it should be. I'm guessing it's generated, but I'm not sure how to change it.

tests/output/ns-nsplex.out
```diff
@@ -1,3 +1,6 @@
+Degree of FEM Space: 1
 Global FEM nodes: 0 on 1 ranks
 Local FEM nodes: 8 (0 owned)
+dm_plex_box_faces: NONE
+Time taken for solution: 0.00363827
```
jrwrigh: Any way to have this accept any number value? Like a regex:

```suggestion
Time taken for solution: (?:\d?\.\d+)
```

jeremylt: You'd have to rework lines 130-147 of `tests/tap.sh` to make that work. 

What I'd probably do is check that the desire value is within tolerances in the code and print some sort of good/bad message in test mode for that line. Though, does wallclock time help us check something important here?

examples/navier-stokes/nsplex.c
```diff
@@ -1028,10 +1031,34 @@ int main(int argc, char **argv) {
   user->dt = dt;
 
   // Solve
+  start = MPI_Wtime();
   ierr = TSSolve(ts, Q); CHKERRQ(ierr);
+  ierr = TSGetSolveTime(ts,&ftime); CHKERRQ(ierr);
+  cpu_time_used = MPI_Wtime() - start;
```
jeremylt: I'd flip the order of these two lines. Also, you probably want `ierr = PetscBarrier((PetscObject)ts); CHKERRQ(ierr);` between `start = ...` and `TSSolve`.

examples/navier-stokes/nsplex.c
```diff
@@ -1028,10 +1031,34 @@ int main(int argc, char **argv) {
...
+  ierr = PetscPrintf(PETSC_COMM_WORLD,
+                    "Time taken for solution: %g\n",
+                    (double)cpu_time_used); CHKERRQ(ierr);
```
jeremylt: You'll want an `MPI_Allreduce` here

Example from `bps.c`:
`
        ierr = MPI_Allreduce(&my_rt, &rt_max, 1, MPI_DOUBLE, MPI_MAX, comm);
        CHKERRQ(ierr);`

tests/output/ns-nsplex.out
jeremylt: In `bps.c`, we have a command line arg for testing mode, and we modify/reduce output if we are in testing mode.

Example:
```
  // Print summary
  if (!test_mode) {
    PetscInt P = degree + 1, Q = P + qextra;
    const char *usedresource;
    CeedGetResource(ceed, &usedresource);
    ierr = PetscPrintf(comm,
                       "\n-- CEED Benchmark Problem %d -- libCEED + PETSc --\n"
                       "  libCEED:\n"
                       "    libCEED Backend                    : %s\n"
                       "  Mesh:\n"
                       "    Number of 1D Basis Nodes (p)       : %d\n"
                       "    Number of 1D Quadrature Points (q) : %d\n"
                       "    Global nodes                       : %D\n"
                       "    Owned nodes                        : %D\n",
                       bpChoice+1, usedresource, P, Q, gsize/ncompu,
                       lsize/ncompu); CHKERRQ(ierr);
  }
```

jrwrigh: The wall clock time is just a measurement, so we could leave it off completely for testing if I setup a separate testing mode.

jeremylt: I'm personally a big fan of 'pretty' output normally and stripped down output in testing mode. I generally only print in test mode on an error.

I've mentioned `bps.c` a few times in this review because this Navier-Stokes code started as a modified version of that file. We have since improved that example so there are fixes in there for some of the things you're hitting.

jrwrigh:
A few questions:
 - Should I keep "Max Error" in the when `-test` flag is done? 
     - May be useful for testing, but also results would more than likely differ with different backends/architectures
 - To remove the compiler warnings, just remove the unused functions?
    - `Setup2d`, `IFunction_DC`, and `IFunction_Advection` inside `navierstokes.c`

jeremylt:
For testing made, I generally prefer no output unless there is an error.

For the compiler warnings, I would look at the multigrid and bps example. We included a ifdef guard on some of the functions in setup.h.

@jedbrown, what do you think about remaining to navierstokes.c and navierstokesraw.c to match the BPs? Long term believe the plan is to keep only the dmplex code, it seems silly to give it a more complex name.

jedbrown:
The place where the structured/unstructured codes diverge is in boundary conditions, which is something we need to do a lot of work on in the coming months. The BPs don't really have to deal with that, but I don't see value in keeping a "raw" NS.

jrwrigh:
> For testing made, I generally prefer no output unless there is an error.
So I should remove the `tests/output/ns-[nsplex|navier-stokes].out` files? Wouldn't that leave the nsplex without any available testing?

jeremylt:
The `.out` files provide a predetermined output that the test should be checked against if it produces output. This is separate from if a file is in the test suite.

There is no `.out` file for the ex*.c, MFEM examples, NEK examples, or the PETSc BP examples because they only print output on an error in test mode, but they are all part of the test suite.

jrwrigh:
So I should prevent any printing to stdout when the `-test` flag is used and remove the `tests/output/ns-[nsplex|navier-stokes].out` files?

jeremylt:
That's what I'd do. I am trying to migrate out test suite to only use output files when we are specifically testing the ability of the library or test to generate a specific print statement.

jrwrigh:
I believe this is now good to go. The error on travis is with one of the build jobs not being able to clone OCCA. All the other instantiations pass fine. Let me know if there's anything else left to be done.

jeremylt:
I'm seeing one last warning on compilation.
```
/home/travis/build/CEED/libCEED/examples/navier-stokes/nsplex.c: In function ‘main’:

/home/travis/build/CEED/libCEED/examples/navier-stokes/nsplex.c:872:5: warning: ‘problem’ may be used uninitialized in this function [-Wmaybe-uninitialized]

     CeedQFunctionCreateInterior(ceed, 1, problem->apply_ifunction,

     ^
```

examples/navier-stokes/densitycurrent.h
```diff
@@ -474,6 +475,7 @@ CEED_QFUNCTION(DC)(void *ctx, CeedInt Q,
 //                                       (diffussive terms will be added later)
 //
 // *******************************************************************************
+#ifdef nsplex
 CEED_QFUNCTION(IFunction_DC)(void *ctx, CeedInt Q,
                    const CeedScalar *const *in, CeedScalar *const *out) {
```
jrwrigh: So I can't reproduce [that warning](https://github.com/CEED/libCEED/pull/430/#issuecomment-581662639) (the `problem->apply_ifunction` no initialized) on my laptop, but I'm guessing that's due to gcc versions being different. 

My only guess as to why the warning shows is that gcc isn't reading the `#ifdef` directive correctly/well on the definition of `IFunction_DC`, which was used to get rid of the original compiler warnings (about unused functions when compiling `navierstokes.c`). Any ideas on how to resolve this?

jeremylt: I can't reproduce it locally with GCC or Clang. Given that we plan on removing `navierstokes.c` and renaming `nsplex.c' as such, this issue will go away so I don't see it as a blocking issue here.

jedbrown:
I had committed the faux pas of rebasing `jed/ns-dmplex` after you started working on it, so I rebased your branch onto my branch and all seems to work, including sensible convergence results on advection2d under grid refinement. Could you have a look at https://github.com/CEED/libCEED/tree/jrwrigh/ns-dmplex, and if you don't object to what I've done, reset the branch in your repository to my version of your branch (this will update the PR) and I'll merge to my branch (which I think is nearly mergable to master).

jedbrown:
A tentative merge of this branch is contained in #458 which is prepared for master.

PR #431, opened 2019-12-11
Operator - move AssembleDiagonal to fallback

jeremylt:
This is the AssembleDiag/FDM Inverse Refactor.

- [x] Move AssembleDiag
- [x] Refactor AssembleDiag for performance
- [x] Move FDM Inverse

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -668,7 +668,6 @@ static int CeedOperatorAssembleLinearQFunction_Ref(CeedOperator op,
   return 0;
 }
 
-
 // Assemble Linear Diagonal
 static int CeedOperatorAssembleLinearDiagonal_Ref(CeedOperator op,
     CeedVector *assembled, CeedRequest *request) {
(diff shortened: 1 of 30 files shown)


jeremylt:
Operator assembly is still slow, but less so with the latest commit. This approach opens more opportunities for optimization.

jeremylt:
There is still room for improvement on the performance of `AssembleLinearDiagonal`, but this is a big improvement and seems like a good amount of work to consider merging.

backends/ref/ceed-ref-operator.c
```diff
@@ -668,6 +668,403 @@ static int CeedOperatorAssembleLinearQFunction_Ref(CeedOperator op,
...
+  ierr = CeedBasisGetInterp(basisout, &interpout); CeedChk(ierr);
+  ierr = CeedBasisGetGrad(basisin, &gradin); CeedChk(ierr);
+  ierr = CeedBasisGetGrad(basisout, &gradout); CeedChk(ierr);
+  // Each element and component
+  CeedPragmaSIMD
+  for (CeedInt e=0; e<nelem; e++) {
+    CeedPragmaSIMD
+    for (CeedInt q=0; q<nqpts; q++) {
+      CeedPragmaSIMD
+      for (CeedInt n=0; n<nnodes; n++) {
```
jedbrown: I don't think this is correct with respect to the pragmas that we use (https://gcc.gnu.org/onlinedocs/gcc/Loop-Specific-Pragmas.html) or `omp simd`, which is really a hint to use that loop as the SIMD index. At a minimum, I would remove it from the `q` loop.

backends/ref/ceed-ref-operator.c
```diff
@@ -668,6 +668,403 @@ static int CeedOperatorAssembleLinearQFunction_Ref(CeedOperator op,
...
+        if (interp)
+          qdataarray[(e*ncomp+c)*nnodes+n] = 1;
+        if (grad)
+          for (CeedInt d=0; d<dim; d++) {
+            CeedInt i = (n / CeedIntPow(P1d, d)) % P1d;
+            qdataarray[(e*ncomp+c)*nnodes+n] += lambda[i];
+          }
+        qdataarray[(e*ncomp+c)*nnodes+n] = 1 / (elemavg[e] *
+                                                qdataarray[(e*ncomp+c)*nnodes+n]);
```
jedbrown: What is the performance hot-spot of this PR? I'd be interested in seeing a profile.

include/ceed-backend.h
```diff
@@ -200,4 +200,8 @@ CEED_EXTERN int CeedOperatorFieldGetLMode(CeedOperatorField opfield,
 CEED_EXTERN int CeedOperatorFieldGetVector(CeedOperatorField opfield,
     CeedVector *vec);
 
+CEED_INTERN int CeedMatrixMultiply(Ceed ceed, CeedScalar *matA,
+                                   CeedScalar *matB, CeedScalar *matC,
+                                   CeedInt m, CeedInt n, CeedInt kk);
+
```
jedbrown: What memory space do its arguments reside in? That this is a reference implementation not intended for performance should probably be added to the docs.

interface/ceed-basis.c
```diff
@@ -1107,85 +1152,60 @@ int CeedBasisGetInterp(CeedBasis basis, CeedScalar **interp) {
...
+    for (CeedInt i=0; i<basis->dim*basis->Q*basis->P; i++)
+      basis->grad[i] = 1.0;
+
+    // Calculate
+    for (CeedInt d=0; d<basis->dim; d++) {
+      for (CeedInt i=0; i<basis->dim; i++) {
+        CeedPragmaSIMD
+        for (CeedInt qpt=0; qpt<basis->Q; qpt++) {
+          CeedPragmaSIMD
+          for (CeedInt node=0; node<basis->P; node++) {
```
jedbrown: I think these pragmas are also problematic.

If you're going to have the naive code below rather than hoisting it out/restructuring the loop, it's probably best to skip all pragmas because they imply that this code is meant to be high-performance.

backends/ref/ceed-ref-operator.c
jeremylt: I'm pretty sure its that loop on lines `797-854`. The function `CeedOperatorAssembleLinearDiagonal` accounts for 80% of the run time.

include/ceed-backend.h
jeremylt: The caller is responsible for the memory of A, B, and C.

jedbrown: Sure, but some of it is `const` and we should probably be explicit that they should be valid CPU pointers (we don't have a `CeedMemType`).

jeremylt: Right. It was obvious when this was an interface only function. I'll fix it up.

backends/ref/ceed-ref-operator.c
```diff
@@ -16,39 +16,9 @@
...
-  ierr = CeedFree(&impl); CeedChk(ierr);
-  return 0;
-}
-
-/*
-  Setup infields or outfields
- */
+//------------------------------------------------------------------------------
+// Setup Input/Output Fields
+//------------------------------------------------------------------------------
```
jeremylt: With the new fallback, this file is at >1100 lines and was getting hard for me to visually navigate, so I added these little comment blocks to help visually separate the functions.

include/ceed-backend.h
jeremylt: Added `const` and specified that the pointers must be on the CPU.

PR #432, opened 2019-12-11
Minor Renaming for Consistency

jeremylt:
When working of the Python interface, I realized we had a couple of files named `vec` whereas most of the time we moved to `vector`. This fixes this.


jeremylt:
The random linux/gcc Travis failures are now showing up on this branch too. I have no clue what is causing this. Sometimes relaunching the build fixes it. Sometimes there are segfaults in the OCCA backends. Sometimes there are segfaults in the XSMM backends.

jedbrown:
I've had non-reproducible failures in tests run locally at least twice in recent months. I don't know if it's likely that occa (the backend or the library) has a race condition. It shouldn't block this PR, but is something to be aware of and hopefully solve.

PR #434, opened 2019-12-17
XSMM - release JiT kernels

jeremylt:
This releases the XSMM kernels upon destruction of the `TensorContract` object.


jeremylt:
This might have fixed our Linux/GCC random errors.

Issue #408, opened 2019-11-01
ARM continuous integration


jedbrown:
https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support

PR #440, opened 2020-01-08
Travis - Update to Bionic, Add ARM64 and POWER8 arch

jeremylt:
This fixes issue #408

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,7 +86,7 @@ script:
     fi
 
 after_success:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+  - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
         lcov --directory . --capture --output-file coverage.info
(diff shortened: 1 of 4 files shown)


jeremylt:
Side note: I am seeing the arm64 Travis builds taking longer on another project of mine. I think it's the nature of their current setup.

ppc64le (IBM Power (8 I think)) and s390x (IBM Z9) are also options, if desired. We might have to change the OPT flags for the Power 8 though (`-mcpu` in place of `-march`).

Any opinion on adding an IBM build as well @jedbrown?

jedbrown:
No value in zSeries, but POWER8 has some merit in helping to avoid any CPU-related hiccups whenever we run on Summit/Lassen. That part has been smooth thus far (to my knowledge, only GPU-related issues), but I wouldn't object to adding it to the build matrix.

jeremylt:
ARM and POWER8 are now working. I am trying to update to Bionic so we can use a newer gcc with better warnings. (The tricky part is Nek)

jeremylt:
All set. We're updated to Bionic with gcc 7.4 and I found a small bug in t504 thanks to the POWER8 testing.

I'll merge once Travis finishes.

PR #446, opened 2020-01-22
User manual

valeriabarra:
Our very first user manual! 

It uses sphinx, which is able to read both Markdown (~~although I still have to figure out some small things~~ see below) and ReStructuredText files. Math formulas are rendered nicely in html directly from ReStructuredText files using the mathjax extension (a bit of reformatting using the proper directives is necessary if you want to convert a latex file straight to an rst file). 

TO DOs:
~~- [ ] Figure out how to properly parse and render existing `\snippet` from md files (see, e.g., libCEEDapi.md:238).~~ this seems to not be possible, but I have converted the `md` files to `rst` files and used different directives for code snippet inclusions.
- [x] Include man pages from code with `breathe`.
- [x] Add release notes page
- [x] Figure out citations with `sphinxcontrib-bibtex` or other  available extensions.
- [x] Conclude description of all existing examples.
- [x] Add getting started page

PR #447, opened 2020-01-24
ElemRestriction Lmode in Create over Apply

jeremylt:
***Breaking Change***

This is the first of the two ElemRestriction changes we discussed earlier this week. I made `lmode` a property of the ElemRestriction when created rather than having it as a flag during application.

@YohannDudouit, this will break libCEED+MFEM integration.

@nbeams, this should be fully compatible with #424, provided that I got the changes to the CUDA backends correct. (I need to test) Edit: Testing looks good on my machine

### DIFF
--- a/examples/ceed/ex1-volume.c
+++ b/examples/ceed/ex1-volume.c
@@ -346,7 +346,7 @@ int BuildCartesianRestriction(Ceed ceed, int dim, int nxyz[dim], int order,
   // nnodes:   0   1    p-1  p  p+1       2*p             n*p
   CeedInt *el_nodes = malloc(sizeof(CeedInt)*num_elem*nnodes);
   for (CeedInt e = 0; e < num_elem; e++) {
-    CeedInt exyz[3], re = e;
+    CeedInt exyz[3] = {1, 1, 1}, re = e;
     for (int d = 0; d < dim; d++) { exyz[d] = re%nxyz[d]; re /= nxyz[d]; }
     CeedInt *loc_el_nodes = el_nodes + e*nnodes;
     for (int lnodes = 0; lnodes < nnodes; lnodes++) {
--- a/examples/ceed/ex1-volume.c
+++ b/examples/ceed/ex1-volume.c
@@ -330,7 +330,7 @@ int BuildCartesianRestriction(Ceed ceed, int dim, int nxyz[dim], int order,
                               int ncomp, CeedInt *size, CeedInt num_qpts,
                               CeedElemRestriction *restr,
                               CeedElemRestriction *restr_i) {
-  CeedTransposeMode lmode = CEED_NOTRANSPOSE;
+  CeedInterlaceMode imode = CEED_NONINTERLACED;
   CeedInt p = order, pp1 = p+1;
   CeedInt nnodes = CeedIntPow(pp1, dim); // number of scal. nodes per element
   CeedInt elem_qpts = CeedIntPow(num_qpts, dim); // number of qpts per element
@@ -359,10 +359,10 @@ int BuildCartesianRestriction(Ceed ceed, int dim, int nxyz[dim], int order,
       loc_el_nodes[lnodes] = gnodes;
     }
   }
-  CeedElemRestrictionCreate(ceed, lmode, num_elem, nnodes, scalar_size,
+  CeedElemRestrictionCreate(ceed, imode, num_elem, nnodes, scalar_size,
                             ncomp, CEED_MEM_HOST, CEED_COPY_VALUES, el_nodes,
                             restr);
-  CeedElemRestrictionCreateIdentity(ceed, lmode, num_elem, elem_qpts,
+  CeedElemRestrictionCreateIdentity(ceed, imode, num_elem, elem_qpts,
                                     elem_qpts*num_elem, ncomp, restr_i);
   free(el_nodes);
   return 0;
--- a/examples/ceed/ex1-volume.c
+++ b/examples/ceed/ex1-volume.c
@@ -202,12 +202,12 @@ int main(int argc, const char *argv[]) {
   CeedOperator build_oper;
   CeedOperatorCreate(ceed, build_qfunc, CEED_QFUNCTION_NONE,
                      CEED_QFUNCTION_NONE, &build_oper);
-  CeedOperatorSetField(build_oper, "dx", mesh_restr, CEED_NOTRANSPOSE,
-                       mesh_basis,CEED_VECTOR_ACTIVE);
-  CeedOperatorSetField(build_oper, "weights", mesh_restr_i, CEED_NOTRANSPOSE,
-                       mesh_basis, CEED_VECTOR_NONE);
-  CeedOperatorSetField(build_oper, "qdata", sol_restr_i, CEED_NOTRANSPOSE,
-                       CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(build_oper, "dx", mesh_restr, mesh_basis,
+                       CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(build_oper, "weights", mesh_restr_i, mesh_basis,
+                       CEED_VECTOR_NONE);
+  CeedOperatorSetField(build_oper, "qdata", sol_restr_i, CEED_BASIS_COLLOCATED,
+                       CEED_VECTOR_ACTIVE);
 
   // Compute the quadrature data for the mass operator.
   CeedVector qdata;
@@ -247,12 +247,10 @@ int main(int argc, const char *argv[]) {
   CeedOperator oper;
   CeedOperatorCreate(ceed, apply_qfunc, CEED_QFUNCTION_NONE,
                      CEED_QFUNCTION_NONE, &oper);
-  CeedOperatorSetField(oper, "u", sol_restr, CEED_NOTRANSPOSE,
-                       sol_basis, CEED_VECTOR_ACTIVE);
-  CeedOperatorSetField(oper, "qdata", sol_restr_i, CEED_NOTRANSPOSE,
-                       CEED_BASIS_COLLOCATED, qdata);
-  CeedOperatorSetField(oper, "v", sol_restr, CEED_NOTRANSPOSE,
-                       sol_basis, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(oper, "u", sol_restr, sol_basis, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(oper, "qdata", sol_restr_i, CEED_BASIS_COLLOCATED,
+                       qdata);
+  CeedOperatorSetField(oper, "v", sol_restr, sol_basis, CEED_VECTOR_ACTIVE);
 
   // Compute the mesh volume using the mass operator: vol = 1^T \cdot M \cdot 1
   if (!test) {
@@ -332,6 +330,7 @@ int BuildCartesianRestriction(Ceed ceed, int dim, int nxyz[dim], int order,
                               int ncomp, CeedInt *size, CeedInt num_qpts,
                               CeedElemRestriction *restr,
                               CeedElemRestriction *restr_i) {
+  CeedTransposeMode lmode = CEED_NOTRANSPOSE;
   CeedInt p = order, pp1 = p+1;
   CeedInt nnodes = CeedIntPow(pp1, dim); // number of scal. nodes per element
   CeedInt elem_qpts = CeedIntPow(num_qpts, dim); // number of qpts per element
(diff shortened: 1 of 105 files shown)


jeremylt:
Note to self: Add the following to merge commit message

[NEWS] Breaking change: ElemRestriction lmode parameter moved from ElemRestrictionApply to ElemRestrictionCreate*. This change requires updating previous calls to ElemRestrictionCreate*, ElemRestrictionApply, ElemRestrictionGetMultiplicity, and CeedOperatorSetField to move lmode exclusively into ElemRestrictionCreate*.

Further note to self: we're `make style` clean

examples/ceed/ex1-volume.c
```diff
@@ -202,12 +202,12 @@ int main(int argc, const char *argv[]) {
   CeedOperator build_oper;
   CeedOperatorCreate(ceed, build_qfunc, CEED_QFUNCTION_NONE,
                      CEED_QFUNCTION_NONE, &build_oper);
-  CeedOperatorSetField(build_oper, "dx", mesh_restr, CEED_NOTRANSPOSE,
-                       mesh_basis,CEED_VECTOR_ACTIVE);
-  CeedOperatorSetField(build_oper, "weights", mesh_restr_i, CEED_NOTRANSPOSE,
-                       mesh_basis, CEED_VECTOR_NONE);
-  CeedOperatorSetField(build_oper, "qdata", sol_restr_i, CEED_NOTRANSPOSE,
-                       CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(build_oper, "dx", mesh_restr, mesh_basis,
```
valeriabarra: So when the user sets the fields now, where is the possibility to have interlaced fields or not?

jeremylt: It is set when the element restriction is created. The same elem restriction will always use the same lmode.

valeriabarra: Maybe I'm confused now.. But I thought that the way the restriction is applied, i.e., from an L-vec to an E-vec (operator E), or from an E-vec to an L-vec (operator E^T), could have been a totally independent concept from the choice of interlacing the fields or not.

jeremylt: That is still the case. The transpose mode for E or E^T is still an option when you call `CeedElemRestrictionApply()`. Now the layout mode for interlaced fields or not will only be set once, when you call `CeedElemRestrictionCreate()`.

We found that it was very easy to create bugs by inconsistent use of the lmode field in each `CeedOperatorSetField()`.

valeriabarra: Ok. Thanks, sounds clear now.
Sorry, but have we ever considered calling it something else? I think the fact that both `lmode = CEED_TRANSPOSE` and `tmode = CEED_TRANSPOSE` have the same values for different concepts, can be confused, especially by new users.
Have we ever considered using something like `lmode = CEED_INTERLACED` instead of `TRANPOSE` for both concepts?

jeremylt: I don't have strong feelings about that either way. @jedbrown, do you have an opinion on the clarity of using the same enum two different ways for the ElemRestrictions?

If we want to change it, I can bundle it into this PR easily.

jedbrown: I think a new enum would be more clear. One cheap way is just to `#define CEED_INTERLACED CEED_TRANSPOSE`, which allows for more self-documenting code, but also doesn't prevent confusing misuse. A distinct enum would let the type system enforce that they are not interchangeable.

jeremylt: I have the dentist today, but I can crank this out tomorrow morning.

nbeams:
@jeremylt Thanks for taking this on.

This is not really a complaint -- I don't have strong feelings about it -- but would we want to go ahead and make all the backends change to no longer accept `lmode` as a parameter to `Apply`, and get it from the restriction instead?  Just wondering if at some point way down the line, when we've all forgotten the specifics of the development history or many new people are working with libCEED, it will seem strange that the backend `Apply` functions are supposed to accept `lmode` while the user-facing function does not, but it's the opposite for the `Create` functions.  I guess this would mostly affect future backend developers, since the average libCEED user won't deal with the backend interface, so maybe it won't matter even if it seems a bit odd.  (I appreciate what I assume is the intent of disrupting the backends as little as possible, though, for obvious reasons.)

jeremylt:
@nbeams, good point. I originally left `lmode` as an explicit argument for future backends as a reminder that it's important, but upon reflection I think that it should be removed as it will not be important for "Strided" ElemRestrictions and it makes sense to only have arguments that must be used.

nbeams:
Seems good to me.  I'll have to go back and add this change to the magma element restrictions re-introduced in #424, but that's simple.  It just depends on how soon this gets merged if I fold it into #424 or add it in a new PR, no big deal either way.

YohannDudouit:
Does that mean that the restriction has to use the same `lmode` for the restriction and its transpose?

jeremylt:
Yes @YohannDudouit

jeremylt:
(Finished with dentist early, yay!) I added the lmode change (need to test GPU), but Nek is failing on my machine. All Fortran tests pass, and Nek was passing before this change. Either I did something dumb, or Nek is doing something dumb, or both.

I'll investigate further tomorrow.

jeremylt:
It's passing locally for me.

We should now be set with

1) Setting `lmode` on `CeedElemRestrictionCreate*()`
2) Dropping `lmode` as a backend function argument
3) Creating a separate `lmode` enum

I still need to check 3) for CUDA/MAGMA

nbeams:
Just as a note, I tested this branch before you changed the backend function changes and created the enum, and it looked to me that the magma test failures were related to what I fixed in 49d6f15a340367bc26f3fde800bbdee1bb83bbb6 which is waiting to be merged in #424

jeremylt:
The GPU tests look good. Here is the updated commit message:

[NEWS] Breaking change: ElemRestriction lmode parameter moved from ElemRestrictionApply to ElemRestrictionCreate*. This change requires updating previous calls to CeedElemRestrictionCreate*, CeedElemRestrictionApply, CeedElemRestrictionGetMultiplicity, and CeedOperatorSetField to move lmode exclusively into CeedElemRestrictionCreate*.

[NEWS] Breaking change: ElemRestriction layout mode enum created. This change requires changing use of CEED_TRANSPOSE and CEED_NOTRANSPOSE to CEED_LAYOUT_NODE_COMPONENT and CEED_LAYOUT_COMPONENT_NODE, respectively, for CeedElemRestriction lmode in CeedElemRestrictionCreate*.

jedbrown:
Once #446 is merged (so we have a nicer web page), we should make a release notes/changes page with API changes such as these.

jeremylt:
Agreed. At least one API change snuck by without a message like above, so it would be nice to formalize our process for advertising these changes in a way that is easy to see.

examples/ceed/ex1-volume.c
jeremylt: Today I Learned: GCC will not warn about using the wrong enum because different enums are still just integer constants. OSX is the only Travis env giving us these enum warnings.

jedbrown: Clang has `-Wenum-conversion`, and can be used from Linux. Per this patch, it'll also be available in gcc-10. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78736#c14

jeremylt: For now I'm adding Clang locally and paying extra attention to the OSX Travis builds.

valeriabarra:
> Once #446 is merged (so we have a nicer web page), we should make a release notes/changes page with API changes such as these.

OK! I can add a release notes/changes page in #446 . Should I refer to the brief release notes [here](https://github.com/CEED/libCEED/releases) or write something more?

examples/navier-stokes/navierstokes.c
```diff
@@ -356,7 +356,9 @@ int main(int argc, char **argv) {
   const CeedInt dim = 3, ncompx = 3, ncompq = 5;
   CeedVector xcorners, xceed, qdata, q0ceed, mceed, onesvec, multlvec;
   CeedBasis basisx, basisxc, basisq;
-  CeedTransposeMode lmodeceed = CEED_NOTRANSPOSE, lmodepetsc = CEED_TRANSPOSE;
+
+  CeedLayoutMode lmodeceed = CEED_LAYOUT_COMPONENT_NODE,
+                 lmodepetsc = CEED_LAYOUT_NODE_COMPONENT;
```
valeriabarra: Thanks for doing this @jeremylt ! I'm not quite sure if this is the proper variable name. I know it's just a name, but petsc must not be the only package out there using this layout. Could we think of using something that is reminiscent of being "interlaced" or "consecutive/contiguous"?

jeremylt: I am happy to change it if there is a more conventional name. My hesitation with 'interlaced' is that it does not specify *what* is interlaced. Same thing with 'contiguous' or 'consecutive'. Nodes? Fields?

jedbrown: I use interlaced when explaining, and believe it's pretty unambiguous. This clip is from Gropp et al, 2001. Some people/projects use "blocked" as the converse, which I find highly ambiguous compared to "noninterlaced". In the hardware/vendor world, they tend to speak of SoA (structure of array) and AoS (array of structure) for this concept. (Vendors tend to recommend SoA for better vectorization, ignoring the cache locality consequences.)

Screenshot_20200129_103140

Screenshot_20200129_103033

jeremylt: `CEED_LAYOUT_NODE_INTERLACED` vs `CEED_LAYOUT_FIELD_INTERLACED`?

jeremylt: or `CEED_LAYOUT_INTERLACED` and `CEED_LAYOUT_NONINTERLACED`. (I'm just making sure we have a decision so I only change this once more)

jedbrown: I prefer interlaced and noninterlaced. I'd even consider `CeedInterlaceMode` and `CEED_INTERLACED` and `CEED_NONINTERLACED`. "Layout" is pretty ambiguous. The man page for `CeedInterlaceMode` can have a detailed description with examples.

valeriabarra:
> Once #446 is merged (so we have a nicer web page), we should make a release notes/changes page with API changes such as these.

I have added a Release Notes page. When we are ready with v0.6, I will add the new changes as well.

However, for now, I have not provided an out-of-band documentation for any (new or existing) enumeration. They are documented as in the software manpages and parsed through `breathe`.

examples/navier-stokes/navierstokes.c
```diff
@@ -356,7 +356,9 @@ int main(int argc, char **argv) {
   const CeedInt dim = 3, ncompx = 3, ncompq = 5;
   CeedVector xcorners, xceed, qdata, q0ceed, mceed, onesvec, multlvec;
   CeedBasis basisx, basisxc, basisq;
-  CeedTransposeMode lmodeceed = CEED_NOTRANSPOSE, lmodepetsc = CEED_TRANSPOSE;
+
+  CeedInterlaceMode imodeceed = CEED_NONINTERLACED,
+                    imodepetsc = CEED_INTERLACED;
```
valeriabarra: Sorry, I still wouldn't call these variables `imodeceed` and `imodepetsc`. It wouldn't say much to a new user who is learning from our examples as tutorials. What about something like `interlacenode` and `interlacecomp` (or `consecnode` and `conseccomp`)?

jeremylt: I think that would make it more confusing - we decided interlaced referred to one specific meaning and our examples shouldn't muddle that single meaning.

I could directly type `CEED_INTERLACED` or `CEED_NONINTERLACED` into every one of those fields if its more clear.

jeremylt:
Update commit message

[NEWS] Breaking change: ElemRestriction lmode parameter moved from ElemRestrictionApply to ElemRestrictionCreate*. This change requires updating previous calls to CeedElemRestrictionCreate*, CeedElemRestrictionApply, CeedElemRestrictionGetMultiplicity, and CeedOperatorSetField to move lmode exclusively into CeedElemRestrictionCreate*.

[NEWS] Breaking change: ElemRestriction layout mode renamed interlace mode and interlace mode enum created. This change requires changing use of CEED_TRANSPOSE and CEED_NOTRANSPOSE to CEED_INTERLACED and CEED_NONINTERLACED, respectively, for imode in CeedElemRestrictionCreate*.

examples/navier-stokes/navierstokes.c
jeremylt: Updated

jeremylt:
If there are no further objections, can we try to merge this soon so we have plenty of time to work on the new strided restriction interface?

PR #448, opened 2020-01-27
Fix t412 segfault

nbeams:
I noticed I was getting segfaults with all backends on t412-qfunction.c.  It seemed like the problem was from the declared size of the array `u`.  Is this change what was originally intended for the test, or was the loop setting `u` supposed to be changed instead?


nbeams:
Great, thanks everyone!

PR #450, opened 2020-01-30
[Tests] Removed testing padded entries

dmed256:
Hello,

I'm wrapping up on the OCCA backend and the implementation for applying the element restriction in a blocked fashion is a bit different than the other implementations. Rather than defaulting padded nodes to the last element nodes, I'm avoiding doing any computation for those nodes.

I updated the `t208-elemrestriction` to avoid this undeterministic behavior depending on the backend. 

Thanks,
David


jeremylt:
I think it would be preferable to test that the output vector contains the intended values for the non-padded entries over checking against an output file. I've been meaning to make the change, but it's been low priority.

dmed256:
I can make that change :+1:

jeremylt:
@dmed256 what do you think about merging this now? I can catch the output clean up in a future PR.

dmed256:
Sounds good!

PR #455, opened 2020-02-05
MAGMA - restore deleted file

jeremylt:
@nbeams, I don't know what this file is or does, but it was accidentally deleted when I rebased for #447 with the updated master changes from #424. The test suite passed fine for me without it.

### DIFF
--- a/backends/magma/ceed-magma-restriction_cuda.cu
+++ b/backends/magma/ceed-magma-restriction_cuda.cu
@@ -1,0 +1,8 @@
+#include <ceed.h>
+
+#include "atomics.cuh"        
+#include "magma_check_cudaerror.h"   
+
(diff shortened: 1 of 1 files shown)


nbeams:
Hmm, I'm not actually sure why that needed to be added as a separate file.  It looks like Stan added it originally, so I will check with him to be sure.

nbeams:
I realized this file is generated by `gccm.py` each time you build the MAGMA backend.  It doesn't need to be in the repository and was most likely added by accident.  I think we can cancel this PR.  Thanks for checking!

jedbrown:
Thanks! Should it be listed in a `.gitignore` to avoid this issue in the future? Or (better), can it be generated in `build/`?

PR #294, opened 2019-07-12
Adapt PETSc BPs to use CUDA

jeremylt:
This WIP PR modifies the PETSc BPs that we use to benchmark to use CUDA vectors, if the libCEED backend and PETSc both support it. I need to test it, but I believe it works.

Issue #250, opened 2019-05-18
Identity Restriction Handling


jeremylt:
We need to improve the `CeedElemRestrictionApply` interface to better handle identity restrictions and avoid unnecessary memory movement. See the discussion in PR #228.

jedbrown:
I think this was essentially addressed by #456. Please reopen if that is not accurate.

PR #456, opened 2020-02-06
Strided Restrictions

jeremylt:
This first commit is very WIP and will be updated, but I wanted to give everyone a look at my first take on the strided elem restriction that will replace identity elem restriction.

@dmed256, I'm working to get this in for v0.6. I hope this will be easier/clearer to impl than identity restr.

ToDo:
- [x] python interface
- [x] examples
- [x] OCCA rstr
- [x] CUDA rstr
- [x] MAGMA rstr

Specifically:
- [x] Add CEED_BACKEND_STRIDES to CUDA
- [ ] ~Optimize CUDA ref operator to use CEED_BACKEND_STRIDES~ deferred to follow-up PR
- [x] Finish Strided backends in cuda/gen


jeremylt:
Update: Tests should all be passing for CPU/OCCA. Once I see Travis is happy, I can start taking a crack at adding this to CUDA and MAGMA, though I'm not the fastest person for that job. Any time you have @nbeams and @YohannDudouit to help make these changes would be appreciated.

This is the written description of how these new kernels should use the `strides` data
```
  @param strides    Array for strides between [nodes, components, elements].
                      The data for node i, component j, element k in the
                      L-vector is given by
                        i*strides[0] + j*strides[1] + k*strides[2]
```

jeremylt:
@YohannDudouit, I forgot to mention, this is also a breaking change for libCEED + MFEM integration. This will require some changes over in MFEM for the libCEED and MFEM packages in CEED 3.0 spack to be compatible.

nbeams:
Thanks for all your work on this so far @jeremylt.  

I'm happy to make the necessary updates to the MAGMA backend, but I have a question and a concern.

First, the question:
The new StridedRestriction's stride parameters are only for the L-Vector (which is really an E-Vector or Q-Vector, and has a sense of elements), meaning the output of `tmode=CEED_NOTRANSPOSE` and input of `tmode=TRANSPOSE` mode are expected to be in "standard" E-Vector layout -- can this be the standard for the backend, or for interoperability reasons relating to your preconditioning, etc., would it have to be the [elem, comp, node] layout?  

Now, the concern:
I don't think we want to have the L-Vector strides set explicitly in, e.g., the operator tests and examples, because they should be determined by the backend, right?  

What about something like:

1. "Standard" ElementRestriction continues as now, with backend determining E-Vector layout.  This is the only restriction that deals with "pure" L-Vectors, meaning L-Vectors which are not also E-Vectors/Q-Vectors, but a non-overlapping representation of dofs on a process
2. Identity ElementRestriction is reintroduced as a "true" identity, meaning E-Vector=Q-Vector=L-Vector, but the layout of those vectors is still entirely up to the backend.  This is what gets used for `qdata` vectors produced by libCEED operators in the tests, examples, etc. and by most CEED users
3. StridedRestriction is a "super-user" feature, only for interacting with other libraries (e.g., if you have an external input Q-Vector of data created by something other than CEED, you have to know its layout to specify for the L-Vector strides) or for reordering between backends.

Alternatively, we could combine 2. and 3. into StridedRestriction as you've already done, but have a `strides` parameter option (CEED_BACKEND_DEFAULT or something) for letting the StridedRestriction know that it should just use its normal Q-Vector layout, and this is what's used in the examples and tests.

jeremylt:
Re E-vector layout:
  I think we want to get this working with backend specific E-vector layouts for now. I think we should discuss ways to fix the preconditioning fallback in a future PR.

Re L-vector layouts:
  The identity restriction needs to go. It is my intention to add a `CEED_BACKEND_STRIDES` option that can be used to indicate that the backend should use their own layout; however, I wanted to get the new interface working first and then add this small performance enhancement (small to code, big perf implications).

nbeams:
The magma strided restriction in 8dcdac61eac907a792ca02213269fa61e741f4ce is passing all tests for me except:
t534-6 (AssembleLinearDiagonal)
t540 (FDM element inverse)

It doesn't really surprise me that these tests aren't working yet (as mentioned above about the E-Vector layout being different).  However, what I currently don't understand is that while t534-6 fail because they produce the wrong output, t540 gives me the "CeedElemRestrictionGetIMode(): Strided ElemRestriction has no interlace mode" error.  I didn't dig too far into it yet, though.

jeremylt:
I added a WIP version of CUDA strided restrictions. I will need to debug them, but maybe I am close?

jeremylt:
The t540 failure could be related to the borked fallback.

jeremylt:
@YohannDudouit, @nbeams 

One we get all of the backends working in this PR, I will add `CEED_STRIDE_BACKEND`.

My intention is that this will only touch CeedOperatorApply code. Backends will be able to call `CeedElemRestrictionGetStridedStatus(rstr, &strided_flag)`. If `strided_flag` is true, then the backend can check that `strides = CEED_STRIDE_BACKEND`. If so, the l-vector can be used as a e/q-vector, as desired by the backend.

If desired, I can also add `CeedElemRestrictionGetBackendStrideStatus(rstr, &strided_flag)` (or something like that) to do both steps in one call.

jeremylt:
Update: @YohannDudouit, cuda/ref works with the new identity restrictions. cuda/reg and cuda/shared do not. I am not sure how those restriction kernels work. I have not attempted to work on cuda/gen yet

jeremylt:
Update: I have `/gpu/cuda/ref`, `/gpu/cuda/reg`, and `/gpu/cuda/shared` working. I've started on `/gpu/cuda/gen`. To do it properly, I'll probably need to add the optimization discussed above.

nbeams:
@jeremylt I assume the cuda/ref, shared, reg backends all pass the AssembleLinearDiagonal and FDM element inverse tests that MAGMA has trouble with?  Is it just down to the different E-Vector ordering?  (cuda/ref, at least, uses the same E-Vector layout as the CPU backends, I know)

jeremylt:
None of the GPU backends pass the tests involving delegation. That will require more work, but this PR will give us a framework to get there.

I just added `CEED_STRIDES_BACKEND` which operators can use to optimize to prevent data shuffling. I need to turn it on for more of the examples, but you can use ex1 to test it.

With this added, I hope to be able to work on `/gpu/cuda/gen` in a way that preserves the current performance. Hopefully once we can tidy up the rest of these changes we will have the same performance (with the opportunity for future improvements) but with an improved interface.

jeremylt:
All examples other than PETSc multigrid use CEED_STRIDES_BACKEND now.

nbeams:
For some reason the changes in ad7a69c05695652a73b50f35fd50e168f3fc2883 made t540 pass now.  (t534-6 still fail)

backends/magma/ceed-magma-restriction.c
```diff
@@ -34,27 +34,57 @@ static int CeedElemRestrictionApply_Magma(CeedElemRestriction r,
...
+    ierr = CeedElemRestrictionGetBackendStridesStatus(r, &backendstrides);
+
+    if (backendstrides) {
+
+      CeedScalar *du;
+
+      // Set pointer to same data; no re-ordering needed
+      ierr = CeedVectorGetArray(u, CEED_MEM_DEVICE, &du); CeedChk(ierr);
+      ierr = CeedVectorSetArray(v, CEED_MEM_DEVICE, CEED_USE_POINTER, du);
+      ierr = CeedVectorRestoreArray(u, &du); CeedChk(ierr);
```
jeremylt: I think it would be safer(?)/clearer to handle this optimization at the operator level. Applying a restriction and then modifying u should not result in a change in v, so I think alising u as v could result in bugs that might be hard for a user to trace.

The operator should instead realize that a strided restriction with the backend ordering is being used and use u in place of v. (Need to add in this PR for CUDA, but this would be a new performance improvement to the GPU backends).

For the CUDA/OCCA backends, I'm using the backend's preferred strides if this function is ever actually called with CEED_STRIDES_BACKEND.

nbeams: That makes sense.  I changed it in 98df188a45fac370d96950e2cefb88ef466f8462 to set the strides explicitly with the layout we're using for the MAGMA backend.

jeremylt: Thanks

jeremylt:
Update: `/gpu/cuda/gen` is pretty close.

The code below has a `libCEED/backends/cuda/ceed-cuda.c:81 in CeedCompileCuda(): CUDA_ERROR_ILLEGAL_ADDRESS`

```
# + extern "C" __global__ void oper(CeedInt nelem, void* ctx, CudaFieldsInt indices, CudaFieldsInt strides, CudaFields fields, CudaFields B, Cuda$
# + const CeedScalar* d_u0 = fields.in[0];
# + const CeedScalar* d_u1 = fields.in[1];
# + CeedScalar* d_v0 = fields.out[0];
# + const CeedInt Dim = 1;
# + const CeedInt Q1d = 8;
# + extern __shared__ CeedScalar slice[];
# + BackendData data;
# + data.tidx = threadIdx.x;
# + data.tidy = threadIdx.y;
# + data.tidz = threadIdx.z;
# + data.tid  = threadIdx.x + threadIdx.y*blockDim.x + threadIdx.z*blockDim.y*blockDim.x;
# + data.slice = slice+data.tidz*Q1d;
# + // Input field 0
# + const CeedInt P_in_0 = 8;
# + const CeedInt ncomp_in_0 = 1;
# + const CeedInt nnodes_in_0 = 120;
# + // Input field 1
# + const CeedInt P_in_1 = 5;
# + const CeedInt ncomp_in_1 = 1;
# + const CeedInt nnodes_in_1 = 61;
# + __shared__ double s_B_in_1[40];
# + loadMatrix<P_in_1,Q1d>(data, B.in[1], s_B_in_1);
# + // Output field 0
# + const CeedInt P_out_0 = 5;
# + const CeedInt ncomp_out_0 = 1;
# + const CeedInt nnodes_out_0 = 61;
# + __shared__ double s_B_out_0[40];
# + loadMatrix<P_out_0,Q1d>(data, B.out[0], s_B_out_0);
# +
# + __syncthreads();
# + for (CeedInt elem = blockIdx.x*blockDim.z + threadIdx.z; elem < nelem; elem += gridDim.x*blockDim.z) {
# + const CeedInt nnodes_in_0 = 120;
# + // Input field 1
# + const CeedInt P_in_1 = 5;
# + const CeedInt ncomp_in_1 = 1;
# + const CeedInt nnodes_in_1 = 61;
# + __shared__ double s_B_in_1[40];
# + loadMatrix<P_in_1,Q1d>(data, B.in[1], s_B_in_1);
# + // Output field 0
# + const CeedInt P_out_0 = 5;
# + const CeedInt ncomp_out_0 = 1;
# + const CeedInt nnodes_out_0 = 61;
# + __shared__ double s_B_out_0[40];
# + loadMatrix<P_out_0,Q1d>(data, B.out[0], s_B_out_0);
# +
# + __syncthreads();
# + for (CeedInt elem = blockIdx.x*blockDim.z + threadIdx.z; elem < nelem; elem += gridDim.x*blockDim.z) {
# + // Input field 0
# + CeedScalar r_u0[ncomp_in_0*P_in_0];
# + readDofsStrided1d<ncomp_in_0,P_in_0>(data, elem, strides.in[0], d_u0, r_u0);
# + CeedScalar* r_t0 = r_u0;
# + // Input field 1
# + CeedScalar r_u1[ncomp_in_1*P_in_1];
# + readDofs1d<ncomp_in_1,P_in_1>(data, nnodes_in_1, elem, indices.in[1], d_u1, r_u1);
# + CeedScalar r_t1[ncomp_in_1*Q1d];
# + interp1d<ncomp_in_1,P_in_1,Q1d>(data, r_u1, s_B_in_1, r_t1);
# + // Output field 0
# + CeedScalar r_tt0[ncomp_out_0*Q1d];
# + // QFunction
# + CeedScalar* r_q0 = r_t0;
# + CeedScalar* r_q1 = r_t1;
# + CeedScalar* r_qq0 = r_tt0;
# + CeedScalar* in[2];
# + in[0] = r_q0;
# + in[1] = r_q1;
# + readDofsStrided1d<ncomp_in_0,P_in_0>(data, elem, strides.in[0], d_u0, r_u0);
# + CeedScalar* r_t0 = r_u0;
# + // Input field 1
# + CeedScalar r_u1[ncomp_in_1*P_in_1];
# + readDofs1d<ncomp_in_1,P_in_1>(data, nnodes_in_1, elem, indices.in[1], d_u1, r_u1);
# + CeedScalar r_t1[ncomp_in_1*Q1d];
# + interp1d<ncomp_in_1,P_in_1,Q1d>(data, r_u1, s_B_in_1, r_t1);
# + // Output field 0
# + CeedScalar r_tt0[ncomp_out_0*Q1d];
# + // QFunction
# + CeedScalar* r_q0 = r_t0;
# + CeedScalar* r_q1 = r_t1;
# + CeedScalar* r_qq0 = r_tt0;
# + CeedScalar* in[2];
# + in[0] = r_q0;
# + in[1] = r_q1;
# + CeedScalar* out[1];
# + out[0] = r_qq0;
# + mass(ctx, 1 , in, out);
# + // Output field 0
# + CeedScalar r_v0[ncomp_out_0*P_out_0];
# + interpTranspose1d<ncomp_out_0,P_out_0,Q1d>(data, r_tt0, s_B_out_0, r_v0);
# + writeDofs1d<ncomp_out_0,P_out_0>(data, nnodes_out_0, elem, indices.out[0], r_v0, d_v0);
# + }
# + }
```

jeremylt:
Ready for review. Since the end of the month is coming soon, I propose we review and merge this now and do the performance optimizations from `CEED_STRIDES_BACKEND` in a follow-up.

The future performance changes will not affect user code but this PR will, so I want to make sure we get this in for v0.6.

jeremylt:
I thought I had it, but something isn't right with 2d/3d grad for cuda/gen.

jeremylt:
CI failures are due to some OCCA change.

jeremylt:
Ok, the CEED examples work. There is some strange behavior with PETSc + CUDA that is also showing on master, so I believe it to be an unrelated bug.

jeremylt:
Rebased to the new master. Once this is merged, I'll go back and adjust #294

jeremylt:
Note to me: I should go ahead and add 'CEED_RESTRICTION_NONE` for `CEED_EVAL_WEIGHT` so we don't require users to make useless dummy restrictions for that QFunction field. It would fit in this PR and its pretty small. It shouldn't really affect review of the meat of this PR.

jeremylt:
Ok, I added `CEED_ELEMRESTRICTION_NONE`.

This PR touches a lot, but it should be pretty straightforward to review (hopefully). I replaced `CeedElemRestrictionCreateIdentity()` with `CeedElemRestrictionCreateStrided()`. Data in the L-vector is given by `ind = node * STRIDE_NODES + comp * STRIDE_COMP + elem * STRIDE_ELEM`, with the user providing `[STRIDE_NODES, STRIDE_COMP, STRIDE_ELEM]`.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -887,9 +854,11 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
   code << "data.tidz = threadIdx.z;\n";
   code << "data.tid  = threadIdx.x + threadIdx.y*blockDim.x + threadIdx.z*blockDim.y*blockDim.x;\n";
   code << "data.slice = slice+data.tidz*Q1d"<<(dim>1?"*Q1d":"")<<";\n";
+
+  code << "\n// Input field constants and basis data\n";
```
jeremylt: @YohannDudouit, hopefully you don't mind. I added a bunch of these types of comments to the generated code to help me debug. I think it will help us with #460.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -485,32 +461,23 @@ inline __device__ void writeDofsTranspose3d(BackendData& data, const CeedInt nno
...
-    for (CeedInt comp = 0; comp < NCOMP; ++comp) {
-      d_v[ind * NCOMP + comp] = r_v[z+comp*Q1d];
+template <int NCOMP, int P1d>
+inline __device__ void writeDofsStrided3d(BackendData& data, const CeedInt elem, const CeedInt *strides, const CeedScalar* r_v, CeedScalar* d_v) {
+  if (data.tidx<P1d && data.tidy<P1d) {
+    for (CeedInt z = 0; z < P1d; ++z) {
+      const CeedInt node = data.tidx + data.tidy*P1d + z*P1d*P1d;
+      const CeedInt ind = node * strides[0] + elem * strides[2];
+      for (CeedInt comp = 0; comp < NCOMP; ++comp) {
+        d_v[ind + comp * strides[1]] += r_v[z+comp*P1d];
```
jeremylt: I used `+=` here instead of the previous `=` because per spec we currently sum into each field. I am unsure of the level of impact, so we may wish to revisit my choice here.

backends/blocked/ceed-blocked-operator.c
```diff
@@ -60,16 +60,25 @@ static int CeedOperatorSetupFields_Blocked(CeedQFunction qf,
...
+                                                data->indices,
+                                                &blkrestr[i+starte]);
+        CeedChk(ierr);
+      } else {
+        CeedInt strides[3];
+        ierr = CeedElemRestrictionGetStrides(r, &strides); CeedChk(ierr);
+        ierr = CeedElemRestrictionCreateBlockedStrided(ceed, nelem, elemsize,
+               blksize, nnodes, ncomp, strides, &blkrestr[i+starte]);
+        CeedChk(ierr);
+      }
```
jeremylt: For our blocked backends, I do not think we can us`CEED_STRIDES_BACKEND` to improve the amount of data movement at the moment. We may be able to reduce data movement for the serial backends (#250), but I don't think we would see a big impact unless we are recomputing qdata (or other inputs that use a strided restriction) frequently. For all CPU backends, we maintain a copy of the restricted input vector and use that appropriately permuted copy until the input changes.

backends/occa/ceed-occa-restrict.okl
```diff
@@ -113,9 +113,25 @@
...
+
+// *****************************************************************************
+@kernel void kRestrict7(const double* uu, double* vv) {
+  for (int e=0; e<nelem; e++; @tile(TILE_SIZE,@outer,@inner)){
+    for (int k=0; k<numcomp; k++) {
+      for (int n=0; n<elemsize; n++) {
+        vv[n*stridenode + k*stridecomp + e*strideelem] +=
+          uu[n + k*elemsize + e*elemsize*numcomp];
+      }
+    }
```
jeremylt: I didn't try to make this kernel especially fast since we will be replacing the entire OCCA backend at some point.

examples/navier-stokes/densitycurrent.h
```diff
@@ -77,7 +77,8 @@
 //   center          ,  Location of bubble center
 //   dc_axis         ,  Axis of density current cylindrical anomaly, or {0,0,0} for spherically symmetric
 // *****************************************************************************
-static int Exact_DC(CeedInt dim, CeedScalar time, const CeedScalar X[], CeedInt Nf, CeedScalar q[], void *ctx) {
+static int Exact_DC(CeedInt dim, CeedScalar time, const CeedScalar X[],
+                    CeedInt Nf, CeedScalar q[], void *ctx) {
```
jeremylt: `make style` had a bunch of changes for `densitycurrent.h`, and I rejected about half of them. I need to look into how to get astyle to stop trying to add strange spacing for some of the multiplication in this file.

Unrelated - we may wish to consider putting all the QFunction headers in a separate folder. We did this for the petsc folder and it seems to make it easier for new users to navigate.

include/ceedf.h
```diff
@@ -154,14 +161,17 @@
...
 
       integer ceed_vector_none
-      parameter(ceed_vector_none          = -2)
+      parameter(ceed_vector_none          = -6)
+
+      integer ceed_elemrestriction_none
+      parameter(ceed_elemrestriction_none = -7)
+
+      integer ceed_basis_collocated
+      parameter(ceed_basis_collocated     = -8)
```
jeremylt: I updated these numbers to be unique while I added these fields. It should help make less buggy Fortran code.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
YohannDudouit: The reason why it was an `=` was because it was for quadrature data, not for dofs, I think we want to continue differentiating these two cases because of the performance difference.

jeremylt: I don't think we need an atomic add here since you only write once in a strided restriction? Q-vector vs L-vector should be one-to-one.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -419,51 +395,51 @@ inline __device__ void readDofsTranspose3d(BackendData& data, const CeedInt nnod
...
+    r_u[comp] = d_u[ind * NCOMP + comp];
   }
 }
 
 template <int NCOMP, int Q1d>
-inline __device__ void readSliceQuadsTranspose3d(BackendData& data, const CeedInt nquads, const CeedInt elem, const CeedInt q, const CeedScalar* d_u, CeedScalar* r_u) {
+inline __device__ void readSliceQuadsStrided3d(BackendData& data, const CeedInt elem, const CeedInt q, const CeedInt *strides, const CeedScalar* d_u, CeedScalar* r_u) {
   const CeedInt node = data.tidx + data.tidy*Q1d + q*Q1d*Q1d;
-  const CeedInt ind = node + elem * Q1d*Q1d*Q1d;
+  const CeedInt ind = node * strides[0] + elem * strides[2];
```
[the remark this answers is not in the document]
jeremylt: I can add this later today.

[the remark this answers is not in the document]
jeremylt: Updated.

examples/navier-stokes/densitycurrent.h
```diff
@@ -102,7 +103,9 @@ static int Exact_DC(CeedInt dim, CeedScalar time, const CeedScalar X[], CeedInt
   // -- Potential temperature, density current
   CeedScalar rr[3] = {x - center[0], y - center[1], z - center[2]};
   // (I - q q^T) r: distance from dc_axis (or from center if dc_axis is the zero vector)
-  for (int i=0; i<3; i++) rr[i] -= dc_axis[i] * (dc_axis[0]*rr[0] + dc_axis[1]*rr[1] + dc_axis[2]*rr[2]);
+  for (int i=0; i<3;
+       i++) rr[i] -= dc_axis[i] * (dc_axis[0]*rr[0] + dc_axis[1]*rr[1] +
```
valeriabarra: probably this should have not been accepted this way, with the for loop statement broken on two lines

interface/ceed-fortran.c
```diff
@@ -343,7 +370,7 @@ void fCeedElemRestrictionApplyBlock(int *elemr, int *block, int *tmode,
     FORTRAN_NAME(ceedelemrestrictiongetmultiplicity,CEEDELEMRESTRICTIONGETMULTIPLICITY)
 void fCeedElemRestrictionGetMultiplicity(int *elemr, int *mult, int *err) {
   *err = CeedElemRestrictionGetMultiplicity(CeedElemRestriction_dict[*elemr],
-                                            CeedVector_dict[*mult]);
+         CeedVector_dict[*mult]);
```
valeriabarra: Is also this line supposed to be accepted according to make style or keep as was? (Asking because I'm going through the same changes in my branch)

jeremylt: I think we want it to be like this.

jeremylt: fixed

valeriabarra: Even though it is different from other function calls?

jeremylt: The goal is always to have make style return no changes on master. I have to figure out what is strange in the density current header, but that's not part of this PR (ideally).

I just ran make style to fix my changes consistent with our style. Sometimes some PRs get through without make style, and I am catching some of those issues here.

valeriabarra: I understand the goal. But this is not in the density current header and I'm trying to understand why we should accept this change, if the previous version was still in our style guidelines, no? The original line did not exceed 90 characters and the line as is now would be just different from all other function call styles, where we indent at the parenthesis.

jeremylt: Make style wouldn't have changed it if it was consistent. The function calls indent 4 spaces or at the parentheses, depending upon how deep the parentheses are. To see this, look at the QFunction portion of this header.

jeremylt:
I know that this touches a lot of code, but I think that it's ready to go. Do we need any other changes @YohannDudouit @jedbrown?

I'm working downstream of this for the Solids example and I'd like to open a PR to accept that code soon.

include/ceed.h
```diff
@@ -254,17 +259,26 @@ typedef enum {
 
 CEED_EXTERN const char *const CeedInterlaceModes[];
 
+/// Argument for CeedElemRestrictionCreateStrided that L-vector is in
+/// the Ceed backend's preferred layout. This argument should only be used
+/// with vectors created by a Ceed backend.
+/// @ingroup CeedElemRestriction
+CEED_EXTERN CeedInt CEED_STRIDES_BACKEND[3];
```
jedbrown: This is a flag value, and thus can be `const`, right?

include/ceed.h
```diff
@@ -254,17 +259,26 @@ typedef enum {
...
-    CeedInterlaceMode imode,CeedInt nelem, CeedInt elemsize, CeedInt blksize,
+    CeedInterlaceMode imode, CeedInt nelem, CeedInt elemsize, CeedInt blksize,
     CeedInt nnodes, CeedInt ncomp, CeedMemType mtype, CeedCopyMode cmode,
     const CeedInt *indices, CeedElemRestriction *rstr);
+CEED_EXTERN int CeedElemRestrictionCreateStrided(Ceed ceed,
+    CeedInt nelem, CeedInt elemsize, CeedInt nnodes, CeedInt ncomp,
+    CeedInt strides[3], CeedElemRestriction *rstr);
+CEED_EXTERN int CeedElemRestrictionCreateBlockedStrided(Ceed ceed,
+    CeedInt nelem, CeedInt elemsize, CeedInt blksize, CeedInt nnodes,
+    CeedInt ncomp, CeedInt strides[3], CeedElemRestriction *rstr);
```
jedbrown:
```suggestion
    CeedInt ncomp, const CeedInt strides[3], CeedElemRestriction *rstr);
```

include/ceed.h
```diff
@@ -254,17 +259,26 @@ typedef enum {
...
-    CeedInterlaceMode imode,CeedInt nelem, CeedInt elemsize, CeedInt nnodes,
-    CeedInt ncomp, CeedElemRestriction *rstr);
 CEED_EXTERN int CeedElemRestrictionCreateBlocked(Ceed ceed,
-    CeedInterlaceMode imode,CeedInt nelem, CeedInt elemsize, CeedInt blksize,
+    CeedInterlaceMode imode, CeedInt nelem, CeedInt elemsize, CeedInt blksize,
     CeedInt nnodes, CeedInt ncomp, CeedMemType mtype, CeedCopyMode cmode,
     const CeedInt *indices, CeedElemRestriction *rstr);
+CEED_EXTERN int CeedElemRestrictionCreateStrided(Ceed ceed,
+    CeedInt nelem, CeedInt elemsize, CeedInt nnodes, CeedInt ncomp,
+    CeedInt strides[3], CeedElemRestriction *rstr);
```
jedbrown:
```suggestion
    const CeedInt strides[3], CeedElemRestriction *rstr);
```

include/ceed.h
jeremylt: Right! I'll fix it

jeremylt:
@YohannDudouit, any update? I'd really like to be able to merge this in the next three hours, if we can.

YohannDudouit:
The benchmark script is bugged when using branches, I'm working with @v-dobrev to fix it.

YohannDudouit:
The runtime compilation in `cuda-gen` seems broken.
```
/g/g91/dudouit1/Software/benchmarks/builds/lassen_xlc/libceed/backends/cuda/ceed-cuda.c:71 in CeedCompileCuda(): NVRTC_ERROR_COMPILATION
default_program(298): error: no instance of function template "writeDofsStrided3d" matches the argument list
            argument types are: (BackendData, int, double *, double *)

default_program(234): warning: variable "P_out_0" was declared but never referenced

default_program(236): warning: variable "nnodes_out_0" was declared but never referenced

1 error detected in the compilation of "default_program".
```

jeremylt:
Better?

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -263,28 +245,22 @@ inline __device__ void writeDofsTranspose2d(BackendData& data, const CeedInt nno
...
-  for (CeedInt comp = 0; comp < NCOMP; ++comp) {
-    d_v[ind * NCOMP + comp] = r_v[comp];
+template <int NCOMP, int P1d, int STRIDES_NODE, int STRIDES_COMP, int STRIDES_ELEM>
+inline __device__ void writeDofsStrided2d(BackendData& data, const CeedInt elem, const CeedScalar* r_v, CeedScalar* d_v) {
+  if (data.tidx<P1d && data.tidy<P1d)
+  {
+    const CeedInt node = data.tidx + data.tidy*P1d;
+    const CeedInt ind = node * STRIDES_NODE + elem * STRIDES_ELEM;
+    for (CeedInt comp = 0; comp < NCOMP; ++comp) {
+      d_v[ind + comp * STRIDES_COMP] += r_v[comp];
```
YohannDudouit: `atomicAdd` too?

YohannDudouit:
Would there be a way to avoid all these new `atomicAdd`? Do we really need to accumulate values at quadrature points?

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -1105,7 +1105,7 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
         if (data->indices.in[i]) {
           ierr = CeedElemRestrictionGetIMode(Erestrict, &imode); CeedChk(ierr);
           code << "  // InterlaceMode: "<<CeedInterlaceModes[imode]<<"\n";
-          code << "  readSliceQuads"<<(imode==CEED_NONINTERLACED?"":"Transpose")<<"3d<ncomp_in_"<<i<<",Q1d>(data, nquads_in_"<<i<<", elem, q, indices.in["<<i<<"], d_u"<<i<<", r_q"<<i<<");\n";
+          code << "  readSliceQuads"<<(imode==CEED_NONINTERLACED?"":"Transpose")<<"3d<ncomp_in_"<<i<<",Q1d>(data, nquads_in_"<<i<<", elem, q, d_u"<<i<<", r_q"<<i<<");\n";
```
YohannDudouit: I don't think you mean to modify this line, you got confused with line 1119.

backends/cuda-reg/ceed-cuda-reg-restriction.c
```diff
@@ -256,17 +208,24 @@ static int CeedElemRestrictionApply_Cuda_reg(CeedElemRestriction r,
...
+      ierr = CeedElemRestrictionGetElementSize(r, &elemsize); CeedChk(ierr);
+      void *args[] = {&nelem, &impl->d_ind, &d_u, &d_v};
+      CeedInt blocksize = elemsize<1024?(elemsize>32?elemsize:32):1024;
+      ierr = CeedRunKernelCuda(ceed, kernel, CeedDivUpInt(nnodes, blocksize),
+                               blocksize, args); CeedChk(ierr);
     } else {
-      kernel = impl->noTrTr;
+      kernel = impl->noTrStrided;
+      void *args[] = {&nelem, &d_u, &d_v};
+      ierr = CeedRunKernelCuda(ceed, kernel, CeedDivUpInt(nnodes, blocksize),
```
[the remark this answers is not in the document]
jeremylt: set on line 204

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -119,28 +113,22 @@ inline __device__ void writeDofsTranspose1d(BackendData& data, const CeedInt nno
...
-  for (CeedInt comp = 0; comp < NCOMP; ++comp) {
-    d_v[ind * NCOMP + comp] = r_v[comp];
+template <int NCOMP, int P1d, int STRIDES_NODE, int STRIDES_COMP, int STRIDES_ELEM>
+inline __device__ void writeDofsStrided1d(BackendData& data, const CeedInt elem, const CeedScalar* r_v, CeedScalar* d_v) {
+  if (data.tidx<P1d)
+  {
+    const CeedInt node = data.tidx;
+    const CeedInt ind = node * STRIDES_NODE + elem * STRIDES_ELEM;
+    for (CeedInt comp = 0; comp < NCOMP; ++comp) {
+      d_v[ind + comp * STRIDES_COMP] += r_v[comp];
```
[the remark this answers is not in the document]
jeremylt: Strided restrictions have unique write access. The relationship between the L-vector and the E/Q-vector is one-to-one.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -1264,14 +1271,35 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
...
+    if (data->indices.out[i]) {
+      ierr = CeedElemRestrictionGetIMode(Erestrict, &imode); CeedChk(ierr);
+      code << "  // InterlaceMode: "<<CeedInterlaceModes[imode]<<"\n";
+      code << "  writeDofs"<<(imode==CEED_NONINTERLACED?"":"Transpose")<<dim<<"d<ncomp_out_"<<i<<",P_out_"<<i<<">(data, nnodes_out_"<<i<<", elem, indices.out["<<i<<"], r_v"<<i<<", d_v"<<i<<");\n";
+    } else {
+      bool backendstrides;
+      ierr = CeedElemRestrictionGetBackendStridesStatus(Erestrict,
+                                                        &backendstrides);
+      CeedChk(ierr);
+      CeedInt strides[3] = {1, elemsize, elemsize*ncomp};
```
YohannDudouit: I have doubt about the performance of this data layout on GPU, I'll check this during benchmarking.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -1028,30 +999,45 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
...
+        CeedChk(ierr);
+        CeedInt strides[3] = {1, elemsize, elemsize*ncomp};
+        if (!backendstrides) {
+          ierr = CeedElemRestrictionGetStrides(Erestrict, &strides);
+          CeedChk(ierr);
+        }
+        code << "  // Strides: {"<<strides[0]<<", "<<strides[1]<<", "<<strides[2]<<"}\n";
+        code << "  readDofsStrided"<<dim<<"d<ncomp_in_"<<i<<",P_in_"<<i<<","<<strides[0]<<","<<strides[1]<<","<<strides[2]<<">(data, elem, d_u"<<i<<", r_u"<<i<<");\n";
+      }
+    }
```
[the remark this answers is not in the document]
jeremylt: It's a question of code duplication (the switch had a lot) vs extra logic. I rearranged when I was debugging my switch to strided so I didn't have to change the restriction code in each switch statement (~3x as many places)

backends/cuda-reg/ceed-cuda-reg-restriction.c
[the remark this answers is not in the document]
YohannDudouit: I'm just curious about the reason for not reusing: `CeedInt blocksize = elemsize<1024?(elemsize>32?elemsize:32):1024;`

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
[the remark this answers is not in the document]
jeremylt: If its important to rearrange into the switch for some reason, I can give it a shot.

backends/cuda-reg/ceed-cuda-reg-restriction.c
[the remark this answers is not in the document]
jeremylt: Ah. I must have copied and pasted something poorly.

backends/cuda-reg/ceed-cuda-reg-restriction.c
```diff
@@ -461,8 +425,7 @@ int CeedElemRestrictionCreate_Cuda_reg(CeedMemType mtype,
 }
 
 int CeedElemRestrictionCreateBlocked_Cuda_reg(const CeedMemType mtype,
-    const CeedCopyMode cmode,
-    const CeedInt *indices,
+    const CeedCopyMode cmode, const CeedInt *indices,
```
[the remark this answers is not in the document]
jeremylt: yeah

YohannDudouit:
[cuda-gen.pdf](https://github.com/CEED/libCEED/files/4294681/cuda-gen.pdf)
[cuda-gen-strided.pdf](https://github.com/CEED/libCEED/files/4294682/cuda-gen-strided.pdf)
Benchmarking on lassen shows a small performance improvement for `cuda-gen` backend!

jedbrown:
Great! Thanks for the careful review and testing, @YohannDudouit.

@jeremylt You may want to `make style` before merging. There are a lot of `for(` versus `for (` lines that have been touched and may as well conform to style guidelines. But in any case, I think you're welcome to merge.

PR #461, opened 2020-02-19
Travis: use PETSc from master

jedbrown:
This gets the DMPlex features we've used recently, and will make further updates a 1-line change.


jeremylt:
Do we want this for Azure as well? (though, that's only a Azure side change)

jedbrown:
Yeah, I've been building the new image (also updated base to `ubuntu:rolling` and mpich-3.3.2), but no changes will be needed in this repository.

jedbrown:
The Docker image is updated, so Azure should work now. Please merge this when approved.

PR #463, opened 2020-02-25
PETSc BPs on the cubed-sphere

valeriabarra:
This PR expands our PETSc BPs suite to work on the cubed-sphere.  

TO DO:

- [x]  BP1
- [x]  BP2
- [x]  BP3
- [x]  BP4
- [x] Update with latest version of `CeedElemRestrictionCreate` (i.e., remove `CEED_TRANSPOSE` / `CEED_NOTRANSPOSE`) arguments from `CeedOperatorSetField()`


valeriabarra:
When I started this branch, the two initial commits `559c430` and `3a21148` slipped in from master at that time. Apparently git still hates me sometimes...

valeriabarra:
The PETSc BPs on the cubed-sphere are fully functional. This is ready for review. Thank you

examples/petsc/qfunctions/area/areacube.h
```diff
@@ -109,7 +109,7 @@ CEED_QFUNCTION(SetupMassGeoCube)(void *ctx, const CeedInt Q,
 // -----------------------------------------------------------------------------
 
 // *****************************************************************************
-// This QFunction applies the mass matrix for a scalar field.
+// This QFunction applies the mass operator for a scalar field.
```
jeremylt: Would we want to apply this change elsewhere?

examples/petsc/setup.h
```diff
@@ -565,10 +565,10 @@ static int SetupLibceedByDegree(DM dm, Ceed ceed, CeedInt degree, CeedInt dim,
   CeedOperatorApply(op_setupgeo, xcoord, qdata, CEED_REQUEST_IMMEDIATE);
 
   // Set up RHS if needed
+  CeedVectorCreate(ceed, nelem*nqpts*ncompu, target);
```
jeremylt: Moved?

examples/petsc/bpssphere.c
```diff
@@ -0,0 +1,366 @@
...
+  CeedQFunction qf_error;
+  CeedOperator op_error;
+  CeedVector rhsceed, target;
+  bpType bpChoice;
+
+  ierr = PetscInitialize(&argc, &argv, NULL, help);
+  if (ierr) return ierr;
+  comm = PETSC_COMM_WORLD;
+
+  // Read command line options
```
jeremylt: I think that we should do another refactoring pass to identify common code relatively soon in the future. There are a few more common pieces across area, areasphere, bps, bpssphere, multigrid, and setup (or in different subsets) that can probably be rorganized to reduce total number of lines and future mantaience burden while increasing readability and user friendliness. We should also break those big setup.h files into separate .c files eventually.

examples/ceed/ex2-surface.c
```diff
@@ -425,7 +425,6 @@ int SetCartesianMeshCoords(int dim, int nxyz[3], int mesh_order,
 
 #ifndef M_PI
 #define M_PI    3.14159265358979323846
-#define M_PI_2  1.57079632679489661923
```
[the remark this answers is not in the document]
valeriabarra: Thanks for your revision Jeremy. What do you mean exactly by the change using it?  I have just removed the declaration of the `M_PI_2` define because it was unused in the code.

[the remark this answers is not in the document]
jeremylt: I worded it poorly. I just meant that it looked unused.

examples/petsc/setup.h
valeriabarra: Yes, outside of the conditional block so that it may not be used uninitialized.

examples/petsc/qfunctions/area/areacube.h
valeriabarra: I'm updating it in all examples

examples/petsc/setup.h
jeremylt: Is there a case in which that occurred? The right hand side and target are only ever generated together.

jeremylt: This might be creating memory leak in the multigrid example because it's creating this vector multiple times. I really don't think we want to do this.

valeriabarra: Okay, I'll change it back for the memory leak. I had change it when I was debugging my code and was calling the `SetupLibceedByDegree()` with `false`, so that was a problem because that vector was never initialized. By I agree that this is not the case, since we always call it with `true` in this problem.

valeriabarra:
@jeremylt I have made the changes you requested. If you are satisfied with the changes, you can please approve them, otherwise GitHub is going to keep marking those are "Changes requested" and blocking the merge

valeriabarra:
> Lgtm

Thanks Jeremy!

If it looks good also for @jedbrown I can attempt to do a rebased version with fewer commits or just entirely squash this when merging.

examples/petsc/qfunctions/bps/bp1sphere.h
```diff
@@ -0,0 +1,195 @@
...
+
+    // Setup
+    // x = xx (xx^T xx)^{-1/2}
+    // dx/dxx = I (xx^T xx)^{-1/2} - xx xx^T (xx^T xx)^{-3/2}
+    const CeedScalar modxxsq = xx[0][0]*xx[0][0]+xx[1][0]*xx[1][0]+xx[2][0]*xx[2][0];
+    CeedScalar xxsq[3][3];
+    for (int j=0; j<3; j++)
+      for (int k=0; k<3; k++) {
+        xxsq[j][k] = 0;
+        for (int l=0; l<1; l++)
```
jedbrown: Why are there a number of vectors that are created and indexed as 3x1 matrices? Seems needlessly complicated.

valeriabarra:
Thanks for catching the typo! Oh wow, I had really copied & pasted that a lot of times. I didn't realize. I could have fixed it myself. Thanks

examples/petsc/qfunctions/bps/bp1sphere.h
valeriabarra: Sorry I don't know why I had started working that way. Simplified now.

jedbrown:
Looks good to me for a squash-merge or similar.

PR #469, opened 2020-03-09
Jed/doc examples

jedbrown:
I think this makes example documentation easier to navigate/extend. Generated docs here: https://libceed.readthedocs.io/en/jed-doc-examples

@LeilaGhaffari If you'd like to write some of your notes about stabilization and boundary conditions, I think it would make sense to start from this PR.


valeriabarra:
Thanks for starting this PR @jedbrown. I was going to do that as soon as I could have on Tuesday. I will add the description of my problems on the cubed-sphere as soon as I can.

LeilaGhaffari:
This documentation is very nice. I will add the descriptions as soon as I can.

jedbrown:
@valeriabarra Could you review my latest push, which simplifies the code and expresses the metric tensor in a simpler way. I think you mentioned Heinbockel's book at some point; see page 174 for the contravariant expression of the Laplacian -- that's effectively what we're using here.

valeriabarra:
Thanks @jedbrown for the simplification. It looks good to me and I have only made tiny doc changes.

jedbrown:
Great, thanks. With respect to section headers, it's hard to remember what the requisite number of characters for a constant width is, so I've just been matching the line above. I have no objections to a different convention, but think it may be hard to maintain an exact width.

Any objection to merging this now?

valeriabarra:
Wait for @LeilaGhaffari , she wants to update the NS solver example.

valeriabarra:
Regarding the section headers style, I had initially used a fixed length that was long enough to both contain all titles and show a clear section break when navigating long files. I think I had chosen 40 characters or something similar, and I could easily check for consistency because I use a multi-line cursor in my IDE, but I understand that not everybody does. But then files have been changed significantly and I tried to keep the consistent style. 

If we agree on a number, whether it be 40 or some other number, it'd be clearer and easier for future additions if it is documented somewhere. I'm all in favor to actually have all of this (together with other style options for variable names etc) written down in a style guide, that could be a section in our CONTRIBUTING file, so that there is no ambiguity and, if things change like for this particular case of the length of the sphinx section headers for instance, we can always update the style guide accordingly.

jedbrown:
Leila says she'll make a fresh PR and merging this will help #474 if you approve. I think it's hard to guarantee that section titles will never exceed 40 chars. The convention (or even lack of precise convention) doesn't bother me. We've sort of honored the convention of maintaining consistency with nearby code/docs, though I feel like `make style` is sometimes overly disruptive (mainly due to inaccuracy; evidently it doesn't understand function pointers and thus breaks them oddly, for example).

valeriabarra:
Sorry, but OCD... ^_^''

Do you think people will not respect that, not even if we write a style guide? I have always been in favor of style guides and thought that libCEED was missing one. For this particular case of the section header titles, we want to avoid titles that are too long anyways, because then they won't render nicely in the TOC sidebar on readthedocs. But I think a style guide is always useful, for instance when we had that initial misunderstanding about what convention to use, whether camelCase or not, at the beginning of the Python interface development.

jedbrown:
Sufficiently deeply nested headings could have longer names without affecting the sidebar. I'm okay with a recommendation, but it's hard to check in the browser and having a strict rule adds cognitive load. People are forgetful even if they've read the style guide, which won't always be the case. And people context switch between dozens of projects with subtly different style guides, so I think the best case is to attempt to match local style. If you want to mandate 40-char heading lines, I'll do my best to stick with it when writing and reviewing. Judging from other projects like SciPy and Scikit-Learn, as well as idempotency of `pandoc -t rst source-file.rst`, matching the title length is much more standard.

valeriabarra:
Yes I had seen that it is more standard to match the title lengths, but I also thought that it would help to discern section breaks in long files, since we had some pretty long files. For now, I just prefer to have them consistent. Since some files had a convention, and some had another, I think it doesn't help new contributors and just confuse them on what style they should adopt.

Anyhow if we want to add this and other standards for this project to a style guide, I think it would be useful, and we can always change and update these standards as we consider appropriate for contributors' need.

I have added a citation and it is ok if you want to merge this now.

PR #471, opened 2020-03-10
Interface - use __restrict__ for C++ compatibility

jeremylt:
@nbeams, does this fix the issue you were having with building the CUDA backends?


jedbrown:
If you do this, you can also revert 2517cf59045249613f75fff00c6f061afde59f66. Alternatively, do what most system headers do and use `__restrict`, which is supported by most C++ compilers.

jeremylt:
@jedbrown, I updated it to use `__restrict__`.

nbeams:
I added the change for the `t` parameter as well; now it's building properly.  Thanks!

jeremylt:
Ah yes. Thank you.

PR #474, opened 2020-03-11
Solid Mechanics Example

jeremylt:
Here is the long discussed initial work on solids. We have some documentation tidying to do, but I wanted to give everyone a chance to take a look.

ToDo:
- [x] Update README
- [x] Update documentation
- [x] VLA GPU support

@ArashMehraban


jedbrown:
Thanks! I'd rather not put binary mesh files here. We could have Cubit .jou or Gmsh .geo files.

Should we rename directories `examples/fluids/` and `examples/solid/`? It feels like inconsistency between `examples/navier-stokes/navierstokes` and `examples/solid-mechanics/elasticity`. Cc: @valeriabarra

jeremylt:
I like the renaming.

re meshs - that works for me

jeremylt:
Ok, I've cleaned up the code some more. The only thing that's left, I believe, is the documentation and README.

jeremylt:
I'll update this with VLA for the GPU after #481 merges

jeremylt:
@jedbrown, @valeriabarra we're ready for review

jedbrown:
Should be here soon: https://libceed.readthedocs.io/en/jeremy-solids-example

jeremylt:
Awesome, the local build looked good to me

jeremylt:
The pushes you are seeing are me trying to fix the GPU support - they are not substantive

jeremylt:
Ok, since GetLinearDiagonal isn't fully supported on the GPU side of the house, we can't really do this on the GPU yet.

jeremylt:
The test suite is now giving the correct results for everything except `petsc-bpsraw`, `petsc-area`, and `petsc-bpssphere`.

jedbrown:
I'm not entirely following the Neo-Hookean formulation. This (from Wriggers) is the form I'm familiar with for expressing the second Piola-Kirchoff stress in terms of the right Cauchy-Green tensor.

Screenshot_20200318_164826

jedbrown:
@ArashMehraban My preferred formulation in solid mechanics would be to define the Green-Lagrange strain `E`, the right Cauchy-Green tensor `C = I + 2 E`, and consider strain energy density as a function `W(E)`. Then the second Piola-Kirchoff tensor is `S = \partial W/\partial E`. This approach reduces naturally to
* small strain, geometrically nonlinear: linearize the constitutive relation, https://en.wikipedia.org/wiki/Hyperelastic_material#Saint_Venant%E2%80%93Kirchhoff_model
* hyperelastic, but geometrically linear: E reduces to infinitesimal strain
* linear elasticity: linearize constitutive relation and reduce to infinitesimal strain

I think it's clumsy to use slightly different notation starting from the simple case and building up. Also, discussing the strain energy function is important, IMO, since that is what users often want to change. I also don't think switching to index notation is really required. All computations are carried out in the initial configuration so the programmer only ever needs to deal with reference element and physical/initial configuration.

ArashMehraban:
@JedBrown I edited the docs file for the solids in the hyperelasticity FS and checked the rest of the doc in linear elasticity and hyperelasticity SS. The notation is consistent now and we do not have random switching to indicial notation.  Got rid of B_0 and all integrals \Omega and \partail \Omega

examples/solids/index.rst
```diff
@@ -316,27 +316,21 @@ Carrying through the differentiation :math:numref:`strain-energy-grad` for the m
...
 .. math::
 
-    \int_{B}{\boldsymbol{v} \cdot \left(\nabla_X \cdot \boldsymbol{P} + \rho \boldsymbol{g}\right)} JdV = \boldsymbol{0}
-
-and in indicial notation we have,
-
-.. math::
-
-   \int_{B_0}{v_i \left(\dfrac{\partial{P_{iI}}}{\partial{X_I}} + \rho_0 g_i \right)} dV = 0
+    \int_{\Omega}{\boldsymbol{v} \cdot \left(\nabla_X \cdot \boldsymbol{P} + J \rho \boldsymbol{g}\right)} dV = \boldsymbol{0}
```
jedbrown:
```suggestion
    \int_{\Omega}{\boldsymbol{v} \cdot \left(\nabla_X \cdot \boldsymbol{P} + \rho_0 \boldsymbol{g}\right)} dV = \boldsymbol{0}
```
We don't know `\rho` in the current configuration (because it depends on the solution), only `\rho_0`.

examples/solids/index.rst
```diff
@@ -316,27 +316,21 @@ Carrying through the differentiation :math:numref:`strain-energy-grad` for the m
...
 
 By Integration by part we arrive at the weak form:
```
jedbrown:
```suggestion
Integrating by parts, we arrive at the weak form:
```

examples/solids/index.rst
jedbrown: It warrants a comment that the body force depends on the current configuration and needs to be mapped back to the initial configuration. This is another form of nonlinearity.

examples/solids/index.rst
```diff
@@ -316,27 +316,21 @@ Carrying through the differentiation :math:numref:`strain-energy-grad` for the m
...
-
-where :math:`t_i` is a prescribed boundary written in terms of reference
+    \int_{\Omega}{\nabla_X \boldsymbol{v} \colon \boldsymbol{P}}dV =
+  - \int_{\Omega}{\boldsymbol{v} \cdot \rho_0 \boldsymbol{g}}dV
+  - \int_{\partial \Omega}{\boldsymbol{v} \cdot \boldsymbol{P} \cdot \hat{\boldsymbol{N}}}dA
+    
+where :math:`\boldsymbol{P} \cdot \hat{\boldsymbol{N}}` is a prescribed boundary written in terms of reference
 configuration.
 
 The constitutive law in indicial notation is given by:
```
jedbrown: The text below here should be revised to carry on the narrative from above.

examples/solids/index.rst
```diff
@@ -0,0 +1,348 @@
...
+The strong form of the static balance linear momentum at small strain for the three-dimensional linear elasticity problem is given by :cite:`hughes2012finite`:
+
+.. math::
+   :label: lin-elas
+   
+   \nabla \cdot \boldsymbol{\sigma} + \boldsymbol{g} = \boldsymbol{0} 
+
+
+where :math:`\boldsymbol{\sigma}`, :math:`\boldsymbol{\epsilon}`
+and :math:`\boldsymbol{g}` are stress, strain, and forcing functions,
```
valeriabarra: Here it reads "where sigma, epsilon, and g are stress, strain" etc, but strain hasn't been introduced in eq. (24). So the definition of the strain and relative symbol has to be pushed right after eq. (25), where it's introduced for the first time.

valeriabarra: Or just removed from this sentence, as I see now that it is actually defined on line 50

examples/solids/index.rst
```diff
@@ -0,0 +1,348 @@
...
+   \boldsymbol{\sigma} = \mathsf{C} \boldsymbol{\epsilon},
+
+where 
+
+.. math::
+
+   \boldsymbol{\epsilon} = \dfrac{1}{2}\left(\nabla \boldsymbol{u} + \nabla \boldsymbol{u}^T \right)
+
+is the (symmetric) strain tensor.
+For notational convenience, we suppose the symmetric second order tensors :math:`\bm \sigma` and :math:`\bm \epsilon` are expressed as vectors of length 6 using `Voigt notation <https://en.wikipedia.org/wiki/Voigt_notation>`_, in which case the fourth order material stiffness tensor :math:`\mathsf C` can be represented as a :math:`6\times 6` symmetric matrix
```
valeriabarra: FWIW maybe material stiffness tensor could be confused by someone for the element stiffness matrix? Hughes and other classical authors call this "material moduli matrix" or "elasticity matrix of moduli".

valeriabarra:
FWIW, just a minor style comment. As far as the documentation is concerned, except for math expressions, I had used maximum 90 chars per line (my bad for not following 80 as code style per  astyle—at some point I thought it was 90 and stuck to that). I'm just saying this here since we haven't written an explicit style guide, so people contributing to it could not know.

examples/solids/index.rst
```diff
@@ -0,0 +1,348 @@
...
+   
+   \nabla \cdot \boldsymbol{\sigma} + \boldsymbol{g} = \boldsymbol{0} 
+
+
+where :math:`\boldsymbol{\sigma}`, :math:`\boldsymbol{\epsilon}`
+and :math:`\boldsymbol{g}` are stress, strain, and forcing functions,
+respectively. Integrating by parts on the divergence term, we arrive at the weak form the of equation :math:numref:`lin-elas`:
+
+.. math::
+
```
valeriabarra: I would label this equation and reference it later, together with eq. (24) in the first paragraph of the Hyperelasticity at Small Strain section

examples/solids/index.rst
```diff
@@ -0,0 +1,348 @@
...
+   \nabla \cdot \boldsymbol{\sigma} + \boldsymbol{g} = \boldsymbol{0} 
+
+Integrating by parts on the divergence term, we arrive at the weak form:
+
+.. math::
+
+   \int_{\Omega}{ \nabla \boldsymbol{v} \colon \boldsymbol{\sigma}} dV - \int_{d\Omega}{\boldsymbol{v} \cdot \left(\boldsymbol{\sigma}_t \cdot \hat{\boldsymbol{n}}\right)} dS + \int_{\Omega}{\boldsymbol{v} \cdot \boldsymbol{g}} dV = 0
+
+where :math:`\boldsymbol{\sigma}_t \cdot \hat{\boldsymbol{n}}` is typically replaced with a boundary condition.
+
```
valeriabarra: I would avoid repeating these lines. The governing equation and weak form are the same. It is only the constitutive law that changes now. See my [comment above](https://github.com/CEED/libCEED/pull/474/files#r395913230)

examples/solids/index.rst
```diff
@@ -0,0 +1,348 @@
...
+   
+   \boldsymbol{\sigma} = \lambda \ln(1 + \boldsymbol{\epsilon_v)} \boldsymbol{I}_3 + 2\mu \boldsymbol{\epsilon}
+
+where :math:`\boldsymbol{\sigma}`, :math:`\boldsymbol{\epsilon}`, are stress and
+strain, respectively. Strain is defined by
+
+.. math::
+
+   \boldsymbol{\epsilon} = \dfrac{1}{2}\left(\nabla \boldsymbol{u} + \nabla \boldsymbol{u}^T \right)
+
```
valeriabarra: I would remove these lines too. The symbols and nomenclature have not changed since these symbols have been introduced above.

examples/solids/index.rst
```diff
@@ -0,0 +1,348 @@
...
+where :math:`\bm u` is the displacement and :math:`\mu` and  :math:`\lambda` are the Lamé parameters.
```
valeriabarra: Sorry, I deleted the comment and retyped it here, because I had accidentally dragged a wrong initial line. Here it is :
I would remove these lines too. The symbols and nomenclature have not changed since these symbols have been introduced above.

valeriabarra: Sorry, I deleted the comment and retyped it here, because I had accidentally dragged a wrong initial line. Here it is :
I would remove these lines too. The symbols and nomenclature have not changed since these symbols have been introduced above.

valeriabarra: If the multi-line comment doesn't render (sometimes it happens to me), I mean lines 123-130 here.

examples/solids/index.rst
```diff
@@ -0,0 +1,348 @@
...
+
+Integrating by parts on the divergence term, we arrive at the weak form:
+
+.. math::
+
+   \int_{\Omega}{ \nabla \boldsymbol{v} \colon \boldsymbol{\sigma}} dV - \int_{d\Omega}{\boldsymbol{v} \cdot \left(\boldsymbol{\sigma}_t \cdot \hat{\boldsymbol{n}}\right)} dS + \int_{\Omega}{\boldsymbol{v} \cdot \boldsymbol{g}} dV = 0
+
+where :math:`\boldsymbol{\sigma}_t \cdot \hat{\boldsymbol{n}}` is typically replaced with a boundary condition.
+
+The small strain version of a Neo-Hookean hyperelasticity material is given as
```
valeriabarra: Even the first time strain is introduced we should mention that it is small/infinitesimal strain, otherwise it sounds like peculiar to this Neo-Hookean constitutive relationship, and it is not.

jedbrown:
Arash Mehraban <notifications@github.com> writes:

> In regards to the "The text below here should be revised to carry on the narrative from above." comment, I was wondering if you are suggesting to add extra text to continue on the narrative earlier or if in addition to the extra text, the indicial notation formulation for the constitutive law and its material derivative should be replaced with non-indicial notation? When it comes to the finite strain regime (with the reference and current configuration), I find it confusing when non-indicial notation is used only. Holzapfel demonstrate both in his text too.

My point is that there should be a narrative arc.  At this stage, I see the index notation as supplemental, but it doesn't make sense to suddenly switch and drop in a new equation.  If you want to include index notation, I would probably put those into notes/admonitions.

examples/solids/index.rst
jedbrown: Ogden and Wriggers both use "elasticity tensor". "Stiffness" is used sometimes to distinguish from "compliance", though the latter is much less common. People also find this usage sometimes: https://en.wikipedia.org/wiki/Hooke%27s_law#Matrix_representation_.28stiffness_tensor.29

I don't think confusion with a finite element stiffness matrix is problematic. I don't like using "matrix" for the mathematical object ("tensor" is better). No strong feelings between "elasticity tensor", "material tensor", "elastic moduli tensor", or the wording in the PR.

examples/solids/index.rst
valeriabarra: I took care of that (I think).

examples/solids/index.rst
valeriabarra: I took care of that.

examples/solids/index.rst
```diff
@@ -322,27 +327,27 @@ It is crucial to distinguish between the current and reference element in the To
...
-The constitutive law in indicial notation is given by:
+For simplicity, we express here the constitutive law in indicial notation:
 
 .. math::
 
-   P_{iI} = F_{iB}S_{BI} 
+   P_{iI} = F_{iB}S_{BI}\, ,
 
-Therfore, its material derivative is given by
+so that we can define its Jacobian (or stiffness) fourth order tensor as
```
jedbrown: Hmm, I find this confusing terminology here. The analog of the linear elasticity tensor would be `\mathsf C = dS/dE`, e.g., as described here by Wriggers (the irritating factors of 2 evaporate if you use Green-Lagrange strain E instead of right Cauchy-Green tensor C).
Screenshot_20200320_174203

examples/solids/index.rst
```diff
@@ -322,27 +327,27 @@ It is crucial to distinguish between the current and reference element in the To
...
+where :math:`\boldsymbol{P} \cdot \hat{\boldsymbol{N}}` is a prescribed boundary
+condition written in terms of the reference configuration.
 
-The constitutive law in indicial notation is given by:
+For simplicity, we express here the constitutive law in indicial notation:
 
 .. math::
 
-   P_{iI} = F_{iB}S_{BI} 
+   P_{iI} = F_{iB}S_{BI}\, ,
```
jedbrown: I'm just curious, but is this a rendering issue on your end? It's easy to forget this convention and it makes the math noisier to always have to throw in the extra bit of space `\,`. A raw comma looks fine to my eye.

jedbrown:
Regarding 90 characters, there is a tradeoff with prose. If you wrap a paragraph, then small changes can rewrap the paragraph, resulting in excessively complicated diffs. My usual convention for editing papers is to use `M-x visual-line-mode` in Emacs to wrap text, and then put a hard line break after each sentence. That way the text looks fine on my screen and any edits to a single sentence will appear as one-line diffs (an appropriate amount of context for reviewing). If there is consensus that coding line length should be used in rst files, I can do that, but there are drawbacks.

examples/solids/index.rst
jedbrown: I stand by my initial recommendation that we introduce the fully nonlinear case first and linearize parts geometry and materials to derive the linear case as a simplified model for testing purposes. (I don't think there will be a big performance difference for a good implementation. Linearization might reduce 2 Newton iterations to 1 and make a modest difference in operator application time.) I don't know if @jeremylt and @ArashMehraban affirmatively didn't like that suggestion or just had this text and didn't want to spend time refactoring it.

jeremylt: It seemed like @ArashMehraban took the lead on the documentation, so I hadn't jumped in to avoid too many cooks. I have no strong opinions here.

jeremylt:
Re 80 characters, the only reason this file has breaks at 80 characters is because I reflexively hit a carriage return when I cross the 80 character line on my editor and I was the one that converted the tex to rst. I like @jedbrown's approach above.

examples/solids/index.rst
jedbrown: Yep, and he needs practice writing about this stuff for his paper. I'm happy for it to be in his hands. I (re)wrote the hyperelastic/constitutive modeling section because I felt like it was the most efficient way to communicate about that essential core structure, but I'd like him to help make the whole document fit together. And thanks @valeriabarra for your review and edits.

examples/solids/index.rst
valeriabarra: I agree the "(stiffness)" part could be confusing, in relation to what you called stiffness tensor above. I used the Hughes classical terminology of "material moduli matrix" or "elasticity matrix of moduli" in the past, for which this now would lead to no confusion regarding the stiffness (if this is what you meant). We can also use the incremental terminology as Wriggers does. (I have used algorithmic, too, when expressing discrete stress/strain relationships). But in any case, the existing terminology was not adequate in my opinion.

examples/solids/index.rst
valeriabarra: No it's just that especially when there are terms with subscripts in the equations, the comma or semicolon tend to be too close to them. Just a matter of aesthetics. Can be removed of course.

examples/solids/index.rst
jedbrown: I agree that rewording is needed. I'd prefer consistency with the linear terminology, so call this the incremental elasticity tensor or whatever. We can leave this for @ArashMehraban to think about.

examples/solids/index.rst
jedbrown: Thanks, seems fair to add them (when we remember) after equations that end in subscript. No need to spend more time on this point.

valeriabarra:
> Re 80 characters, the only reason this file has breaks at 80 characters is because I reflexively hit a carriage return when I cross the 80 character line on my editor and I was the one that converted the tex to rst. I like @jedbrown's approach above.

I have the same set up in my editor. I just had erroneously set it to 90 originally, instead of 80. I was also trying to be consistent with code style, but yes I agree with @jedbrown's remark on the diffs hard to read when it's a small addition that changes multiple line breaks.

@jedbrown I also have text wrapping enabled in my editor, but unfortunately it wraps to the window size instead of a fixed column count, so every time I maximize or move the window those lines are very hard to read. But it's not a problem of course. I was just noticing that now this file would look different from others in the documentation.

jedbrown:
@valeriabarra What is your opinion of the directory rename to `examples/fluids/`? Should we pluck that part out to merge before v0.6 (assuming it will take a few days to be happy with this entire PR) so we have naming continuity going forward?

jedbrown:
On text wrapping (gee, I wish GitHub had threading beyond code comments), I'm not concerned about strict rst formatting between modules. I'd just say just adopt conventions that make sense going forward, but we don't need mass content reformatting passes. That'll also help keep `git blame` accurate as to the most recent author of various sections.

valeriabarra:
> @valeriabarra What is your opinion of the directory rename to `examples/fluids/`? Should we pluck that part out to merge before v0.6 (assuming it will take a few days to be happy with this entire PR) so we have naming continuity going forward?

It sounds reasonable to me. If we really wanted to be pedantic we could use `examples/petsc/fluids` and `examples/petsc/solids` since these are PETSc examples.

But then if we add support for complex matter like gels and granular material we would have to create more of those folders to classify them, even though it's just the constitutive relationship that would change a lot, not the main file. 🤔 But anyhow... This is just speculative. It's ok for now!

jeremylt:
Personally, I prefer the convention of keeping our miniapps (fluids, solids, etc) in their own folders and the basic PETSc/MFEM/Nek BP based examples in different folders.

jedbrown:
Those more complicated materials often need more state variables. This is a domain we'll be hitting hard for the PSAAP, using a material-point method for tracking post-failure mechanics of visco-elasto-plastic composites.

ArashMehraban:
@jedbrown I just modified the doc for the HyperFS. I am insisting that the indicial notation must be there as the matrix notation does not quiet make sense to me for the HyperFS regime.

examples/solids/index.rst
```diff
@@ -338,16 +338,35 @@ Integrating by parts, we arrive at the weak form:
 where :math:`\boldsymbol{P} \cdot \hat{\boldsymbol{N}}` is a prescribed boundary
 condition written in terms of the reference configuration.
 
-For simplicity, we express here the constitutive law in indicial notation:
+Equation :math:numref:`1st2nd` devises a method to model a hyperelastic solid with a Neo-Hookean constitutive law. Evaluatinig the derivative of this model, yields a forth order tensor:
```
jedbrown: That equation is just a definition, `P = F S`. It isn't a method. It also isn't clear in this context why you are computing this derivative.

examples/solids/index.rst
```diff
@@ -338,16 +338,35 @@ Integrating by parts, we arrive at the weak form:
...
+
+.. math::
+   :label: mtfs
+
+   \dfrac{\partial \bm P}{\partial \bm F} = \bm S + \left[ \lambda \bm F^{-1} \bm F^{-1} \left(\lambda \log(J) - \mu \right) \left(\bm F^{-1}\bm F^{-1} + \bm C^{-1} \right) \right]
```
jedbrown: FWIW, I prefer to write linearizations in delta form. That is, I'd write `\frac{\partial P}{\partial F} \delta F` where `\delta F` becomes the Newton step you're solving for. For the sake of interpretability, it can be useful to split terms,
```
\frac{\partial P}{\partial F} \delta F = \delta F S + F \frac{\partial S}{\partial E} \delta E
```
where `\delta E = \frac{\partial E}{\partial F} \delta F` and then you can focus on writing down `\frac{\partial S}{\partial E} \delta E`, which is the important interpretable part. This is also the way I would write the code, BTW. (I expect there will be interest in adding different hyperelastic models, so factoring the code well is important.)

examples/solids/index.rst
```diff
@@ -338,16 +338,35 @@ Integrating by parts, we arrive at the weak form:
...
+
+with
+
+.. math::
+   
+   \bm C^{-1} = \bm F^{-1} \bm F^{-T}
```
jedbrown: We've already defined `C` and I'm not sure you ever need `F^{-1}` on its own, just `C^{-1}`

ArashMehraban: @jedbrown I have pushed the latest for the symbolic version of the material derivative for hyper FS (equation 35). According to Rich the two tensor product symbols in that equation should be different. Any preference on that?

valeriabarra:
Since some of my comments above were not addressed in the latest commits, I went ahead and made some improvements myself.

@ArashMehraban as a general recommendation when writing something like documentation or papers (not a note-to-self, but something for others to read), avoid the usage of colloquial language, such as "much like", and re-read everything thoroughly to make sure that it constitutes a coherent, fluent story, do not proof-read just single separate sentences here and there.

examples/solids/index.rst
```diff
@@ -244,12 +216,14 @@ We begin by defining two symmetric tensors in the reference configuration, the r
...
+the latter of which converges to the linear strain tensor :math:`\bm \epsilon` in the small-deformation limit.
+The constitutive models considered, appropriate for large deformations, express :math:`\bm S` as a function of :math:`\bm E`, similar to the linear case, showed in equation  :math:numref:`linear-stress-strain`, which  expresses the relationship between :math:`\bm\sigma` and :math:`\bm\epsilon`.
 
-This is a nonlinear tensor-valued function of tensor-valued input, but an arbitrary choice of such a function will generally not be invariant under orthogonal transformation, thus not admissible because a physical model must not depend on the choice of coordinate system chosen to express it.
+Equation :math:numref:`eq-non-linear-strain` represents a nonlinear tensor-valued function of a tensor-valued input, but an arbitrary choice of such a function will generally not be invariant under orthogonal transformations, thus will not admissible because a physical model must not depend on the choice of coordinate system chosen to express it.
```
jedbrown: "This" referred to `S(E)` in the previous sentence, not the definition of `E` (which isn't a choice). I'm editing now so no need for you to fix.

valeriabarra: Sorry. Yes, I misunderstood. In fact, it's better to avoid "this" when 2 things are introduced in the previous sentence.

jedbrown: Thanks; agreed.

examples/solids/index.rst
jedbrown: Please have a read through how I would write it
https://libceed.readthedocs.io/en/jeremy-solids-example/examples/solids/index.html#newton-linearization

Please also read my discussion of implementation. I'll add some review notes in `hyperFS.h` because I think this can be made much faster.

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,360 @@
...
+        for (int k = 0; k < 3; k++) {
+            C[j][k] = 0;
+            for(int m = 0; m < 3; m++)
+               C[j][k] += F[k][m] * F[j][m];
+        }
+
+
+    // Compute C^(-1) : C-Inverse
+    const CeedScalar A00 = C[1][1]*C[2][2] - C[1][2]*C[2][1];
+    const CeedScalar A01 = C[0][2]*F[2][1] - C[0][1]*C[2][2];
```
jedbrown: I'm pretty sure this `F` is a bug.

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,360 @@
...
+    const CeedScalar A20 = C[1][0]*C[2][1] - C[1][1]*C[2][0];
+    const CeedScalar A21 = C[0][1]*C[2][0] - C[0][0]*C[2][1];
+    const CeedScalar A22 = C[0][0]*C[1][1] - C[0][1]*C[1][0];
+    const CeedScalar detC = C[0][0]*A00 + C[1][0]*A01 + C[2][0]*A02;
+
+    // *INDENT-OFF*
+    const CeedScalar C_inv[3][3] = {{A00/detC, A01/detC, A02/detC},
+                                    {A10/detC, A11/detC, A12/detC},
+                                    {A20/detC, A21/detC, A22/detC}
+                                   };
```
jedbrown: Note that `C` and `C^{-1}` are symmetric and could be represented as such.

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,360 @@
...
+
+    // Compute the Second Piola-Kirchhoff (S)
+    const CeedScalar llnj_m = lambda *log(detF)-mu;
+    CeedScalar S[3][3] = {{mu + llnj_m* C_inv[0][0],      llnj_m* C_inv[0][1],      llnj_m* C_inv[0][2]},
+                          {     llnj_m* C_inv[1][0], mu + llnj_m* C_inv[1][1],      llnj_m* C_inv[1][2]},
+                          {     llnj_m* C_inv[2][0],      llnj_m* C_inv[2][1], mu + llnj_m* C_inv[2][2]}
+                         };
```
jedbrown: `S` is also symmetric and can be represented as such.

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,360 @@
...
+    const CeedScalar Fa21 = F[0][1]*F[2][0] - F[0][0]*F[2][1];
+    const CeedScalar Fa22 = F[0][0]*F[1][1] - F[0][1]*F[1][0];
+    const CeedScalar detF = F[0][0]*Fa00 + F[1][0]*Fa01 + F[2][0]*Fa02;
+
+    // *INDENT-OFF*
+    const CeedScalar F_inv[3][3] = {{Fa00/detF, Fa01/detF, Fa02/detF},
+                                    {Fa10/detF, Fa11/detF, Fa12/detF},
+                                    {Fa20/detF, Fa21/detF, Fa22/detF}
+                                   };
+    // *INDENT-ON*
```
jedbrown: `F_inv` is not needed in the formulation I would recommend (https://libceed.readthedocs.io/en/jeremy-solids-example/examples/solids/index.html#newton-linearization).

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,360 @@
...
+        for (int k = 0; k < 3; k++) {
+            C[j][k] = 0;
+            for(int m = 0; m < 3; m++)
+               C[j][k] += F[k][m] * F[j][m];
+        }
+
+
+    // Compute C^(-1) : C-Inverse
+    const CeedScalar A00 = C[1][1]*C[2][2] - C[1][2]*C[2][1];
+    const CeedScalar A01 = C[0][2]*F[2][1] - C[0][1]*C[2][2];
```
jedbrown: Bug here too.

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,360 @@
...
+
+    // C : right Cauchy-Green tensor
+    // C = F^T * F (^T  means Transpose)
+    CeedScalar C[3][3];
+    for (int j = 0; j < 3; j++)
+        for (int k = 0; k < 3; k++) {
+            C[j][k] = 0;
+            for(int m = 0; m < 3; m++)
+               C[j][k] += F[k][m] * F[j][m];
+        }
```
jedbrown: `C` is symmetric and can be treated as such.

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,360 @@
...
+    CeedScalar dPdF[81];
+    for (int j = 0; j < 3; j++)
+      for(int J = 0; J < 3; J++)
+        for(int a = 0; a < 3; a++)
+          for(int A = 0; A < 3; A++)
+            dPdF[27*j+9*J+3*a+A] = (a == j ? 1 : 0) * (S[A][J]) +
+                                   lambda*F_inv[A][a]*F_inv[J][j]
+                                   - (lambda*log(detF) - mu) *
+                                   (F_inv[A][j]*F_inv[J][a] +
+                                   (a == j ? 1 : 0) * C_inv[A][J]);
```
jedbrown: This is my biggest efficiency concern, and is very unnecessary. Removing this should also reduce register pressure (relevant to both CPU and GPU). Cc: @jeremylt

examples/solids/index.rst
ArashMehraban: @jedbrown @jeremylt 
BTW, the derivation of the material derivative in indicial notation is as follows:
fs-dirv-1
fs-dirv-2
fs-dirv-3

examples/solids/qfunctions/hyperFS.h
jeremylt: To clarify, do you mean that with your recommended formulation we can avoid constructing this full tensor or that there is a better way to fill this tensor?

jedbrown: No tensor. Did the explanation in my docs make sense?

jeremylt: Yeah. I think I swapped over, but now I've lost convergence, so something has to be wrong.

examples/solids/src/setupdm.c
```diff
@@ -166,22 +166,35 @@ PetscErrorCode SetupDMByDegree(DM dm, AppCtx appCtx, PetscInt order,
...
+      ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, NULL, "Face Sets", 0, 0, NULL,
+                           (void(*)(void))BCZero, 1,
+                           &appCtx->bcZeroFaces[faceSet], NULL); CHKERRQ(ierr);
+    }
+    // ---- Clamp BCs
+    for (PetscInt faceSet = 0; faceSet < appCtx->bcClampCount; faceSet++) {
+      ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, NULL, "Face Sets", 0, 0, NULL,
+                           (void(*)(void))BCClamp, 1,
+                           &appCtx->bcClampFaces[faceSet], NULL); CHKERRQ(ierr);
+    }
```
jeremylt: @jedbrown, this fixes our boundary issue

examples/solids/qfunctions/hyperFS.h
jeremylt: Now it seems to be fine.

examples/solids/index.rst
jedbrown: I added a derivation of this form in the bottom note here, along with explanation of why the `dS` form (41) is preferable in computation.
Screenshot_20200323_152545

examples/solids/src/setupdm.c
jedbrown: Why not call `DMAddBoundary` once with the whole list of sett numbers?

jeremylt: Temporary insanity from copy-paste euphoria

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,369 @@
...
+
+    // deltaS = dSdE:deltaE
+    CeedScalar deltaS[3][3];
+    for (CeedInt j = 0; j < 3; j++)
+      for (CeedInt k = 0; k < 3; k++) {
+        deltaS[j][k] = 0;
+        for (CeedInt m = 0; m < 3; m++)
+          for (CeedInt n = 0; n < 3; n++)
+            deltaS[j][k] += lambda*C_inv[m][n]*deltaE[m][n]*C_inv[j][k] -
+                            2.*llnj_m*C_inv[j][m]*deltaE[m][n]*C_inv[n][k];
```
jedbrown: `C^{-1} : dE` can be hoisted out of all these loops. I would also evaluate `C^{-1} dE C^{-1} = C^{-1} (dE C^{-1})` -- i.e., two matrix products instead of one. We know every matrix is symmetric so we could use Voigt notation for everything (but we can leave that optimization for later).

jeremylt: Thanks

jeremylt:
On my laptop, were looking at a change from ~112 sec to ~45 sec with these changes to run 10 increments of hyperFS with one end end at 0 and the other dropping 1 in the y direction by the last increment.

jedbrown:
Great, I'm glad to see that. Have you compared performance with linear and hyperSS for a small-strain experiment?

jeremylt:
linElas runs faster than hyperFS. hyperSS runs slower. The constitutive law in that case looks strange to me

examples/solids/qfunctions/hyperFS.h
```diff
@@ -0,0 +1,387 @@
...
+#ifndef PHYSICS_STRUCT
+#define PHYSICS_STRUCT
+typedef struct Physics_private *Physics;
+struct Physics_private {
+  CeedScalar   nu;      // Poisson's ratio
+  CeedScalar   E;       // Young's Modulus
+};
+#endif
+
+// -----------------------------------------------------------------------------
```
valeriabarra: I know that now we have an awesome extensive documentation, but shouldn't these functions at least have a short help docstring? Not all people might recollect at a first glance what your acronyms mean. All other files (like in the NS example) have the `@file`

jedbrown:
`hyperSS` is slower than `hyperFS` now? I'd hope the compiler has factored out your `log(1 + strain_vol)` in the residual, but don't think that explains the perf diference. I it slower per linear iteration or just converging more slowly? Can you just check how long one fine-grid Jacobian-vector multiply costs (to give us a handle on that part)?

jeremylt:
It's taking an obscene number of SNES and KSP iterations, so that's the problem. I can't seem to track down whats awry though.

ArashMehraban:
The code seems to follow the Matlab code

[userf_3d_hyperelas.txt](https://github.com/CEED/libCEED/files/4372481/userf_3d_hyperelas.txt)

And

[userdf_3d_hyperelas.txt](https://github.com/CEED/libCEED/files/4372484/userdf_3d_hyperelas.txt)

examples/solids/qfunctions/hyperSS.h
```diff
@@ -0,0 +1,295 @@
...
+    // Above Voigt Notation is placed in a 3x3 matrix:
+    const CeedScalar strain_vol = gradu[0][0][i] + gradu[1][1][i] + gradu[2][2][i];
+    const CeedScalar lambda_bar = lambda / (1 + strain_vol);
+
+    const CeedScalar dsigma00 = (TwoMu + lambda_bar)*de[0][0] + lambda_bar*de[1][1] + lambda_bar*de[2][2],
+                     dsigma11 = lambda_bar*de[0][0] + (TwoMu + lambda_bar)*de[1][1] + lambda_bar*de[2][2],
+                     dsigma22 = lambda_bar*de[0][0] + lambda_bar*de[1][1] + (TwoMu + lambda_bar)*de[2][2],
+                     dsigma12 = 0.5*TwoMu*de[1][2],
+                     dsigma02 = 0.5*TwoMu*de[0][2],
+                     dsigma01 = 0.5*TwoMu*de[0][1];
```
jedbrown: Is this Jacobian really correct (Newton converges quadratically)? I think we're getting confused with the factor of 2 that Voigt inserts in the shear components of `\epsilon` (but not `\sigma`). I think it should look like this (using Voigt's ordering convention, but not scaling).
```
                     dsigma12 = TwoMu*de[1][2],
                     dsigma02 = TwoMu*de[0][2],
                     dsigma01 = TwoMu*de[0][1];
```

examples/solids/elasticity.c
```diff
@@ -0,0 +1,713 @@
...
+//
+// Build with:
+//
+//     make elasticity [PETSC_DIR=</path/to/petsc>] [CEED_DIR=</path/to/libceed>]
+//
+// Sample runs:
+//
+//     ./elasticity -problem linElas -degree 2 -nu 0.3 -E 1 -forcing mms -mesh ./meshes/cylinder8_672e_4ss_us.exo
+//     ./elasticity -problem hyperSS -ceed /cpu/self -degree 2 -nu 0.3 -E 1 -forcing mms -mesh ./meshes/cylinder8_672e_4ss_us.exo
+//     ./elasticity -problem hyperFS -ceed /gpu/occa -degree 2 -nu 0.3 -E 1 -forcing mms -mesh ./meshes/cylinder8_672e_4ss_us.exo
```
jedbrown: Do these work for you? I'm getting `DIVERGED_FNORM_NAN`.

examples/solids/qfunctions/hyperSS.h
knepley: Note that you can do these checks automatically with DMSNESCheckFromOptions()

examples/solids/elasticity.c
jeremylt: That should be `-problem linElas`. My mistake.

examples/solids/qfunctions/hyperSS.h
ArashMehraban: @jeremylt This is the indicial notation of the Jacobian:
hyperSS_dirv

jedbrown: Thanks, @knepley.

@jeremylt @ArashMehraban I somewhat hate the matrix format for displaying elasticity tensors, at least as a basis for actually computing. (It's good for clarity when explaining the formulation.) The reason being that it encourages code like the above instead of writing the derivative incrementally:
```
  lambda_dtrace = lambda_bar * (de[0][0] + de[1][1] + de[2][2]);
  dsigma00 = lambda_dtrace + TwoMu * de[0][0];
  dsigma11 = lambda_dtrace + TwoMu * de[1][1];
  dsigma22 = lambda_dtrace + TwoMu * de[2][2];
  dsigma12 = TwoMu * de[1][2];
  dsigma02 = TwoMu * de[0][2];
  dsigma01 = TwoMu * de[1][2];
```
which is much closer to the residual evaluation code and what you'd prefer the compiler to do anyway.

jedbrown: Yep, @ArashMehraban, or in incremental notation
```
  dS = \bar\lambda \trace d\epsilon I + 2 \mu d\epsilon
```

ArashMehraban: Now that we learned how code the Jacobian in HyperFS,  I totally agree with you that the matrix format sucks. I learned it from Engineering dept. and havebeen coding everything like that so far. I am switching to the way we coded HyperFS. Just out of curiosity, do you see anything wrong (from mathematical point of view) in the derivation of Jacobian in HyperSS?

ArashMehraban: @jedbrown @jeremylt code it like this?
hyperSS-linearization

jeremylt: This was it.

jedbrown: Yeah, but you're mixing `\delta u` with `d u` and your equation (1) is mangled; just start with (2).

Note: I used `\operatorname{d} u` in the .rst, so it's typeset in upright and has suitable spacing on the left. I have often used `\delta u` in the context of Newton linearization (because the step isn't infinitesimal), but that sort of conflicts with the Kronecker delta already in the doc, so I switched to `d`.

examples/solids/index.rst
```diff
@@ -142,12 +142,12 @@ Consequently, equation :math:numref:`derss` can be written in matrix form as fol
...
-        \bar{\lambda} & \bar{\lambda} & 2\mu +\bar{\lambda} & & & \\
-        & & & \mu & & \\    
-        & & & &\mu & \\
-        & & & & & \mu \\   
+       2 \mu +\bar{\lambda} & \bar{\lambda} & \bar{\lambda} & & & \\
+       \bar{\lambda} & 2 \mu +\bar{\lambda} & \bar{\lambda} & & & \\
+       \bar{\lambda} & \bar{\lambda} & 2 \mu +\bar{\lambda} & & & \\
+       & & & 2 \mu & & \\    
+       & & & & 2 \mu & \\
+       & & & & & 2 \mu \\
```
jedbrown: Note that the original was correct if you use the scaling of Voigt notation. If you're just ordering without the scaling convention (which won't match textbooks), then the matrices in the linear elasticity section need to also have their scaling updated. This equation should also use `\begin{pmatrix}`. Cc: @ArashMehraban

examples/solids/qfunctions/hyperFS.h
jeremylt: Thanks for catching that

examples/solids/qfunctions/hyperSS.h
jeremylt: Thank you for catching that @jedbrown.

examples/solids/index.rst
ArashMehraban: @jeremylt  I think this order:
d\sigma_11 , d\sigma_22, d\sigma_33, d\sigma_12, d\sigma_13, d\sigma_23
should be changed to:
 d\sigma_11 , d\sigma_22, d\sigma_33, d\sigma_23, d\sigma_13, d\sigma_12

jedbrown: This comment and diagram may be useful, and the narrative could be adjusted to introduce it earlier.
Screenshot_20200323_222831

I'll let one of you repair the currently-inconsistent Voigt-like scaling between the linear and small-strain sections.

jedbrown: I expanded the numerical stability note, and this will allow removing some code in `hyperFS.h`.
Screenshot_20200324_103435

jedbrown: To make the point about numerical stability, consider this demo. If you choose single precision, neither of these quantities will have the first digit correct. `H` here is meant to represent `\nabla_X u`.
Screenshot_20200324_100501

jeremylt: Updated - on my machine this is faster for larger nu and slower for smaller nu, compared to the previous formulation.

jedbrown: "slower"? Perhaps a Newton bug? I'll comment on this patch.

jedbrown: Well, I deleted some unnecessary code. We're getting poor vectorization of the quadrature loop, but the code looks good to me. Can you give me two cases to compare to see slower for small nu?

jeremylt: It's better than it was, so something must have been going in the background on my machine.

There is now a small consistent performance difference on the 672 element mesh (~3-5 more seconds), but I think that's because on this mesh I'm getting the same number of SNES iterations and what looks to the about the same number of KSP iterations, but with slightly more computations in the QFunction. I don't seem to be in a region where the extra stability is helping cut iterations with higher accuracy.

jedbrown: This choice isn't about cutting iterations for these large-deformation problems, but rather being able to solve small-deformation problems to high relative accuracy despite using the FS formulation. Our vectorization is held up entirely because `log1p` from `libc` isn't vectorized. We get decently vectorized code if we replace it with a local approximation (and almost double the end-to-end performance).
```
00000000000131b9 <HyperFSdF+0xcc9> vfmadd231pd ymm9,ymm1,ymm3
00000000000131be <HyperFSdF+0xcce> vfmadd231pd ymm14,ymm1,ymm2
00000000000131c3 <HyperFSdF+0xcd3> vfmadd231pd ymm9,ymm15,ymm4
00000000000131c8 <HyperFSdF+0xcd8> vfmadd231pd ymm14,ymm15,ymm6
00000000000131cd <HyperFSdF+0xcdd> vmovapd YMMWORD PTR [rsp+0x288],ymm9
00000000000131d6 <HyperFSdF+0xce6> vmulpd ymm9,ymm13,ymm7
00000000000131da <HyperFSdF+0xcea> vmulpd ymm13,ymm13,ymm4
00000000000131de <HyperFSdF+0xcee> vfmadd231pd ymm9,ymm0,ymm3
00000000000131e3 <HyperFSdF+0xcf3> vfmadd231pd ymm13,ymm0,ymm2
00000000000131e8 <HyperFSdF+0xcf8> vmovapd ymm0,YMMWORD PTR [rsp+0x1e8]
00000000000131f1 <HyperFSdF+0xd01> vmulpd ymm0,ymm0,YMMWORD PTR [rsp-0x58]
00000000000131f7 <HyperFSdF+0xd07> vfmadd231pd ymm9,ymm12,ymm4
00000000000131fc <HyperFSdF+0xd0c> vfmadd231pd ymm13,ymm12,ymm6
0000000000013201 <HyperFSdF+0xd11> vmovapd YMMWORD PTR [rsp+0x228],ymm9
000000000001320a <HyperFSdF+0xd1a> vmulpd ymm9,ymm11,ymm7
000000000001320e <HyperFSdF+0xd1e> vmulpd ymm11,ymm11,ymm4
0000000000013212 <HyperFSdF+0xd22> vfmadd231pd ymm9,ymm5,ymm3
0000000000013217 <HyperFSdF+0xd27> vfmadd231pd ymm11,ymm5,ymm2
000000000001321c <HyperFSdF+0xd2c> vfmadd231pd ymm9,ymm10,ymm4
0000000000013221 <HyperFSdF+0xd31> vfmadd132pd ymm10,ymm11,ymm6
0000000000013226 <HyperFSdF+0xd36> vmovapd ymm11,YMMWORD PTR [rsp+0x408]
000000000001322f <HyperFSdF+0xd3f> vsubpd ymm11,ymm11,YMMWORD PTR [rsp+0x328]
```

We don't have to resolve this now, but [SVML](https://software.intel.com/en-us/cpp-compiler-developer-guide-and-reference-intrinsics-for-logarithmic-operations-512-bit) has a vectorized `log1p` that we could work out how to call. Alternatively, we can rejigger the portable code to help the compiler vectorize despite the fact that _that_ part can't vectorize. That would probably involve some manual tiling.
```diff
diff --git i/examples/solids/qfunctions/hyperFS.h w/examples/solids/qfunctions/hyperFS.h
index 48dd06da..8bb443bc 100644
--- i/examples/solids/qfunctions/hyperFS.h
+++ w/examples/solids/qfunctions/hyperFS.h
@@ -176,7 +176,7 @@ CEED_QFUNCTION(HyperFSF)(void *ctx, CeedInt Q, const CeedScalar *const *in,
     // *INDENT-ON*
 
     // Compute the Second Piola-Kirchhoff (S)
-    const CeedScalar llnj = lambda*log1p(detC_m1)/2.;
+    const CeedScalar llnj = lambda*detC_m1/2.;
     const CeedScalar S00 = llnj*Cinv[0][0] +
                            mu*(Cinv[0][0]*E2[0][0] + Cinv[0][1]*E2[1][0] +
                                Cinv[0][2]*E2[2][0]),
@@ -371,7 +371,7 @@ CEED_QFUNCTION(HyperFSdF)(void *ctx, CeedInt Q, const CeedScalar *const *in,
     // *INDENT-ON*
 
     // Compute the Second Piola-Kirchhoff (S)
-    const CeedScalar llnj = lambda*log1p(detC_m1)/2.;
+    const CeedScalar llnj = lambda*detC_m1/2.;
     const CeedScalar S00 = llnj*Cinv[0][0] +
                            mu*(Cinv[0][0]*E2[0][0] + Cinv[0][1]*E2[1][0] +
                                Cinv[0][2]*E2[2][0]),
```

jeremylt: Would a second or third order approximation provide sufficient accuracy and allow for vectorization?

jedbrown: Yeah, definitely possible. This series is probably the one to use. The modeling error for any real material will almost certainly be larger than the approximation error.

What's your feeling about factoring utility functions into static inline functions that we call from qfunctions? I don't think there's a portability problem with that and it would allow a lot less duplicated code between the HyperFSF and HyperFSdF.
Screenshot_20200324_202713

jeremylt: I unrolled the loop in that series and now we have a much faster hyperFS.

jeremylt:
> I think we should explain some example runs (including boundary conditions) in the rst docs ~(and make sure the ones in the file work).~

I think this is the biggest remaining thing before merging, correct @jedbrown @ArashMehraban?

jedbrown:
Yes, I feel it's ready to merge with that documentation. Thanks.

jeremylt:
Ok. The code is good now and we are ready for a final review of the documentation: https://libceed.readthedocs.io/en/jeremy-solids-example/examples/solids/index.html#

jeremylt:
Do we want to go ahead and use the series approximation for small strain lop1p as well?

examples/solids/index.rst
```diff
@@ -0,0 +1,479 @@
...
+To set the boundary conditions, use :code:`-bc_zero` or :code:`-bc_clamp` followed by a comma separated list of the constrained faces::
+
+   ./elasticity -mesh ./meshes/meshes/cylinder8_672e_4ss_us.exo -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998
+
+These command line options are the minimum requirements for the mini-app, but additional options may also be set.
+For example, the materiel model (:code:`-problem`), forcing term (:code:`-forcing`), libCEED backend resource (:code:`-ceed`) can be specified::
+
+   ./elasticity -mesh ./meshes/meshes/cylinder8_672e_4ss_us.exo -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998 -problem hyperFS -forcing none -ceed /cpu/self/opt/blocked
+
+The full list of command line options is detailed in the :file:`README` file in the :file:`examples/solids` subdirectory.
```
jedbrown: Can we convert the README to rst so we can link it here too?

jedbrown:
`log1p` in finite strain is more benign because it's only present in the residual, not in the Jacobian. We could use the series there too, but I don't much care either way.

examples/solids/index.rst
```diff
@@ -0,0 +1,479 @@
...
+      \text{Small Strain Hyperelastic} & \underset{\bm \sigma = \mathsf C \bm \epsilon}{\overset{\text{constitutive}}{\LARGE \longrightarrow}} & \text{Linear Elastic} \\
+      \end{matrix}
+
+.. _running-elasticity:
+
+Running the mini-app
+----------------------------------------
+
+There are five required command line options and a variety of additional command line options for this mini-app.
+The four required command line options are :code:`-mesh`, :code:`-degree`, :code:`-E`, and :code:`-nu`. Additionally, at least one boundary condition must be set, using :code:`-bc_zero` or :code:`-bc_clamp`.
```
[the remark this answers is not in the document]
jeremylt: I am incrementally building a single example, but I didn't explain that well. Updated.

examples/solids/elasticity.c
```diff
@@ -0,0 +1,714 @@
...
+//
+// Build with:
+//
+//     make elasticity [PETSC_DIR=</path/to/petsc>] [CEED_DIR=</path/to/libceed>]
+//
+// Sample runs:
+//
+//     ./elasticity -mesh ./meshes/cylinder8_672e_4ss_us.exo -degree 2 -E 1 -nu 0.3 -problem linElas -forcing mms
+//     ./elasticity -mesh ./meshes/cylinder8_672e_4ss_us.exo -degree 2 -E 1 -nu 0.3 -bc_zero 999 -bc_clamp 998 -problem hyperSS -forcing none -ceed /cpu/self
+//     ./elasticity -mesh ./meshes/cylinder8_672e_4ss_us.exo -degree 2 -E 1 -nu 0.3 -bc_zero 999 -bc_clamp 998 -problem hyperFS -forcing none -ceed /gpu/occa
```
valeriabarra: I think these sample runs need to be updated, as it appears that the mesh file has been renamed, so these runs will crash

jeremylt: see above discussion, this is the right name but the meshes are not included in our repo

examples/solids/Makefile
```diff
@@ -0,0 +1,94 @@
...
+LDLIBS = $(call pkgconf, --libs-only-l $(PETSc.pc) $(ceed.pc)) -lm
+
+OBJDIR := build
+SRCDIR := src
+
+src.c := $(sort $(wildcard $(SRCDIR)/*.c))
+src.o = $(src.c:$(SRCDIR)/%.c=$(OBJDIR)/%.o)
+
+elasticity.c := elasticity.c
+elasticity.o := $(elasticity.o:%.c=$(OBJDIR)/%.o)
```
jedbrown: `src.o` and `elasticity.o` are all used together below. Why not put them together?

examples/solids/elasticity.c
valeriabarra: Ah, I lost that comment. So what should I run?

examples/solids/Makefile
```diff
@@ -0,0 +1,94 @@
...
+print: $(PETSc.pc) $(ceed.pc)
+	$(info CC      : $(CC))
+	$(info CFLAGS  : $(CFLAGS))
+	$(info CPPFLAGS: $(CPPFLAGS))
+	$(info LDFLAGS : $(LDFLAGS))
+	$(info LDLIBS  : $(LDLIBS))
+	@true
+
+clean:
+	$(RM) -r $(OBJDIR) $(elasticity) *.d
```
jedbrown: `*.d` isn't enough to get subdirectories, but if you have `src.o`, then this could be `$(src.o:%.o=%.d)` like we have in the `include` statement below.

examples/solids/README.rst
```diff
@@ -0,0 +1,84 @@
...
++-----------------------+-------------------------------------------------------------------+
+| ``-test``             | Run in test mode                                                  |
++-----------------------+-------------------------------------------------------------------+
+| ``-degree``           | Polynomial degree of tensor product basis                         |
++-----------------------+-------------------------------------------------------------------+
+| ``-mesh``             | Filepath to Exodus-II file                                        |
++-----------------------+-------------------------------------------------------------------+
+| ``-problem``          | Problem to solve (``linElas``, ``hyperSS``, or ``hyperFS``)       |
++-----------------------+-------------------------------------------------------------------+
+| ``-forcing``          | Forcing term option (``none``, ``constant``, or ``mms``)          |
```
valeriabarra: How do we feel about explaining what `mms` stands for? I would have personally chosen another name for it, since the method of manufactured solutions is a method, not a type of forcing. But if you really like that name, at least, please explain the acronyms you choose for the people who did not follow the entire development and are going to read this later on some day.

examples/solids/index.rst
```diff
@@ -0,0 +1,479 @@
...
+
+To set the ExodusII mesh file, use the :code:`-mesh` option and the file path to the mesh file::
+
+   ./elasticity -mesh ./meshes/meshes/cylinder8_672e_4ss_us.exo
```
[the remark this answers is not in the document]
jeremylt: https://github.com/ArashMehraban/CeedPetscSolid/tree/master/meshes
or
https://github.com/jeremylt/ceedSampleMeshes
We have a bunch of meshes here. Would either of these work?

examples/solids/qfunctions/hyperSS.h
```diff
@@ -0,0 +1,338 @@
...
+  CeedScalar sum = 0;
+  if (1) { // Disable if the smaller range sqrt(2) < J < sqrt(2) is sufficient
+    if (x < left) { // Replace if with while for arbitrary range (may hurt vectorization)
+      sum -= log(2) / 2;
+      x = 1 + 2 * x;
+    } else if (right < x) {
+      sum += log(2) / 2;
+      x = (x - 1) / 2;
+    }
+  }
```
jedbrown: Could flip this to `if (0)` because the SS formulation isn't valid outside that range anyway (due to geometric nonlinearity). Could also get away with fewer iterations, but that's premature optimization in my book.

examples/solids/README.rst
```diff
@@ -0,0 +1,84 @@
...
+
+``./elasticity -mesh [.exo file] -degree [degree] -nu [nu] -E [E] [boundary options] -problem [problem type] -forcing [forcing] -ceed [ceed]``
+
+Runtime options
+---------------
+
+.. inclusion-marker-do-not-remove
+
+There are five required command line options and a variety of additional command line options for this mini-app.
+The four required command line options are :code:`-mesh`, :code:`-degree`, :code:`-E`, and :code:`-nu`. Additionally, at least one boundary condition must be set, using :code:`-bc_zero` or :code:`-bc_clamp`.
```
jedbrown: Are these "primary" or "principal" command-line options?

examples/solids/README.rst
```diff
@@ -0,0 +1,84 @@
...
+
+To set the ExodusII mesh file, use the :code:`-mesh` option and the file path to the mesh file.
```
jedbrown: Should this be a bulleted list?

examples/solids/README.rst
```diff
@@ -0,0 +1,84 @@
...
+
+To set the polynomial degree of the finite element basis, use the :code:`-degree` option.
+
+To set the materiel parameters, use the :code:`-E` and :code:`-nu` options, for Young's modulus and Poisson's ratio, respectively.
+
+To set the boundary conditions, use :code:`-bc_zero` or :code:`-bc_clamp` followed by a comma separated list of the constrained faces.
```
jedbrown: Bulleted list or table would be easier to read here.

examples/solids/README.rst
```diff
@@ -0,0 +1,84 @@
...
+.. inclusion-marker-do-not-remove
+
+There are five required command line options and a variety of additional command line options for this mini-app.
+The four required command line options are :code:`-mesh`, :code:`-degree`, :code:`-E`, and :code:`-nu`. Additionally, at least one boundary condition must be set, using :code:`-bc_zero` or :code:`-bc_clamp`.
+
+To set the ExodusII mesh file, use the :code:`-mesh` option and the file path to the mesh file.
+
+To set the polynomial degree of the finite element basis, use the :code:`-degree` option.
+
+To set the materiel parameters, use the :code:`-E` and :code:`-nu` options, for Young's modulus and Poisson's ratio, respectively.
```
jedbrown: Spell check "materiel".

examples/solids/README.rst
jeremylt: I added an explanation. "the forcing term to use when running the method of manufactured solutions to verify the convergence of the linear formulation" was a bit lengthy to make as an option which is why we used mms : )

examples/solids/README.rst
```diff
@@ -0,0 +1,90 @@
...
+A face set with :code:`-bc_zero` will remain fixed at zero displacement and a face set with :code:`-bc_clamp` will be displaced by :code:`-bc_clamp_max` in the y direction, or -1 in the y direction if this value is not set.
+
+The following is an example of a minimal set of command line options::
+
+   ./elasticity -mesh ./meshes/meshes/cylinder8_672e_4ss_us.exo -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998
+
+These command line options are the minimum requirements for the mini-app, but additional options may also be set.
+For example, the problem formulation (:code:`-problem`), forcing term (:code:`-forcing`), libCEED backend resource (:code:`-ceed`) can be specified::
+
+   ./elasticity -mesh ./meshes/meshes/cylinder8_672e_4ss_us.exo -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998 -problem hyperFS -forcing none -ceed /cpu/self/opt/blocked
```
jedbrown: Probably not literally `meshes/meshes/`

examples/solids/index.rst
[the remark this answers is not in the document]
valeriabarra: To even just run the code with a given mesh a user would need Cubit?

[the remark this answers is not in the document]
jeremylt: Cubit is only required to generate a mesh.

examples/solids/README.rst
jeremylt: What about mandatory? I am trying to communicate that the code won't run without these options set.

examples/solids/index.rst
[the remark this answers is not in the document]
jedbrown: Your repository is fine. Better not to mix up access to meshes with the legacy version of the code.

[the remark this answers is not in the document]
valeriabarra: That is what I thought. But then I'm not entirely following why not to push at least the meshes mentioned in the sample runs, or as Jed said to make them available for download somewhere and document that, specifying the proper instructions to run the example. Otherwise one would just copy and paste this sample run and it would crash

jeremylt:
@jedbrown, it looks like the Azure Pipelines container might need PETSc updated.

examples/solids/README.rst
jedbrown: Oh, sure. I thought they had defaults based on earlier runs. How about skip the prose and just say the following options are mandatory and have a [list-table](https://docutils.sourceforge.io/docs/ref/rst/directives.html#list-table)?

examples/solids/index.rst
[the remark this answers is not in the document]
jeremylt: Updated with a link.

examples/solids/README.rst
jeremylt: updated

examples/solids/elasticity.c
valeriabarra: Please @jeremylt or @ArashMehraban consider removing these lines now or update them to show where the meshes are, since they provide an instruction that won't run. There is no point in providing a wrong instruction.

jeremylt: updated

examples/solids/Makefile
jeremylt: The change makes sense to me - I'm just too tired to fight with make right now.

examples/solids/README.rst
valeriabarra: Hum... the other two options denote a type of forcing (`none` or `constant`, not a method). I would have tried to be consistent and just used a word that describes the type of forcing for this case, perhaps `nonlinear`? All other examples we have for our BP3s adopt the method of manufactured solutions, but we never say that those examples serve as a code-validation tool, to verify that our numerical approximation can adequately reproduce the continuous operator. Instead we say, we solve Poisson's problem. Do you see what I mean?

examples/solids/elasticity.c
valeriabarra: Thanks

examples/solids/README.rst
jeremylt: `mms` also denotes a forcing term. This option sets the forcing term to be the one analytically computed from the manufactured solution. `nonlinear` provides less information about the forcing term.

examples/solids/Makefile
jedbrown: Pushed.

examples/solids/README.rst
jedbrown: s/validation/verification/

I don't have a problem with `-forcing mms` as terminology. The fluids solver primarily solves the density current configuration (we need to make it more flexible). It would be really useful for the fluids solver to have good manufactured solutions, such as [MASA](https://github.com/manufactured-solutions/MASA), in which case something like `-forcing mms` or `-problem mms` would make sense. The other examples are currently toys so they _only_ do manufactured/analytic solutions.

jedbrown:
@jeremylt What do you think about modest squashing at this point? It'd be nice to roughly retain attribution, but I guess either extreme of squashing everything or nothing would also be acceptable.

jedbrown:
Oh, and we should remove the meshes directory and png. That can move to your meshes repository.

jedbrown:
Your removal of meshes should be squashed into the commit that introduced them so it doesn't take up space in the repository for eternity. 47fef2c6f82a22d4c7cdc600858078b378085f76

valeriabarra:
FYI: There is a problem in the `latexpdf` build of the documentation in this branch.

jeremylt:
What is the problem? Where do I look to find it?

valeriabarra:
> What is the problem? Where do I look to find it?

You can run `doc-latexpdf` from top level, or 
```
cd doc/sphinx/
make latexpdf
```

jeremylt:
I still am not sure what is the problem.
It looks ok here? https://libceed.readthedocs.io/_/downloads/en/jeremy-solids-example/pdf/

valeriabarra:
Let me clean my builds and try again. It was breaking for me locally.

jeremylt:
Ah, now I see. It's cutting off early?

valeriabarra:
The problem seems to be at line

```
l.2751 ...boldsymbol{\diff u}^T \right)\end{split}
```

jedbrown:
Pushed a fix.

valeriabarra:
I have removed the `\LARGE` before the `\longrightarrow` in the commutative linearization diagram because it did not render properly in the PDF version (see attached screenshot from previous version)
Screenshot from 2020-03-26 10-31-37

jedbrown:
Urgh, okay. There are good packages for commutative diagrams in real LaTeX, but we don't want to maintain two versions.

examples/solids/index.rst
```diff
@@ -0,0 +1,459 @@
...
+Running the mini-app
+----------------------------------------
+
+.. toctree::
+   :hidden:
+
+   README
+
+.. include:: README.rst
+   :start-after: inclusion-marker-do-not-remove
```
jeremylt: This is rendering correctly in HTML but not PDF.

valeriabarra: If something deosn't render properly in one build or another, and you want to exclude that directive from one specific build, you can use the `.. only:: html` for instance or `.. only:: latexf`, as we've done to exlude the mermaid graph in the API doc that was not rendering correctly in the PDF build

examples/petsc/index.rst
```diff
@@ -138,8 +138,8 @@ and thus the weak Laplacian may be expressed as
 .. math::
    :label: eq-weak-laplace-sphere
 
-   \int_S \frac{\partial v}{\partial \overset\circ{\mathbf x}} \left( \frac{\partial u}{\partial \overset\circ{\mathbf x}} \right)^T
-       = \int_S \frac{\partial v}{\partial \mathbf X} \underbrace{\frac{\partial \mathbf X}{\partial \overset\circ{\mathbf x}} \left( \frac{\partial \mathbf X}{\partial \overset\circ{\mathbf x}} \right)^T}_{\mathbf g_{(2\times 2)}}  \left(\frac{\partial u}{\partial \mathbf X} \right)^T
+   \int_{\Omega} \frac{\partial v}{\partial \overset\circ{\mathbf x}} \left( \frac{\partial u}{\partial \overset\circ{\mathbf x}} \right)^T
```
valeriabarra: I took the occasion to homogenize the volume/surface integrals notation in all our examples docs. We had mainly used `dV` and `dS` in all other examples, including the NS mini-app.

@jedbrown, here in the BPs on the cubed-sphere you had used `S`. I guess it was the domain `\Omega` in this case. Now this equation and the one below are the only ones that are missing the differentials. I wanted to make sure I understood your notation. We should add them for consistency

examples/solids/README.rst
```diff
@@ -0,0 +1,178 @@
...
+   * - :code:`-bc_zero [int list]`
+     - List of faces sets on which to enforce zero displacement
+
+   * - :code:`-bc_clamp [int list]`
+     - List of face sets on which to displace by :code:`-bc_clamp_max` in the :math:`y` direction
+
+(One can set only one of :code:`-bc_zero` or :code:`-bc_clamp`, but the result will likely not be interesting.)
+The following is an example of a minimal set of command line options::
+
+   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998
```
valeriabarra: Where can a reference to these numbers be found? How do I know what `999` or `998` mean?

jeremylt: in the sample meshes file. The numbers depend of the mesh file

valeriabarra: Yes, but it should be documented and explained. Perhaps we should add this [diagram](https://github.com/jeremylt/ceedSampleMeshes/blob/master/cylinderDiagram.png) to this repo docs as well. Otherwise those are numbers that for someone who wants to test the code are meaningless.

jeremylt: I don't think that diagram makes any sense here. This code is not specific to that mesh.

Personally, I think its fine. These are just sample face numbers. If you want, we can use
```
./elasticity -mesh [mesh] -degree [int] -E [real] -nu [real] -bc_zero [int list] -bc_clamp [int list]
```
to make it totally obvious that the user has to specify everything for the exact problem they want to solve. Or we can remove this sample to avoid the controversy altogether.

examples/petsc/index.rst
jedbrown: For the sphere, we should perhaps use `\int_{S^2}` in place of `\int_{\Omega}` to make clear that the domain is the surface of the 2-sphere.

The differential is always with respect to the volume of integration, and is thus redundant information if we always specify the domain of integration. One could use `\oint_{S^2}` as a more explicit alternative.

examples/solids/README.rst
jedbrown: FWIW, you can include it directly from Jeremy's repository.
```
.. image:: https://github.com/jeremylt/ceedSampleMeshes/raw/master/cylinderDiagram.png
```
Someone not familiar with mesh formats may be confused about what is meant. It would be nice to link an image marking the surfaces and a low-res visualization of a solution.

examples/petsc/index.rst
valeriabarra: I prefer keeping consistency with all other examples. I have clarified what `\Omega` means in this case.

jedbrown: Okay, thanks for the consistency. At least we're using `\diff` for upright `d` for functional differentials, versus standard `d` for volume measures.

jedbrown:
@jeremylt Good to merge IMO. If you're squashing, make sure you get everyone in `Co-authored-by`.

examples/solids/qfunctions/hyperSS.h
```diff
@@ -0,0 +1,325 @@
...
+#endif
+
+// -----------------------------------------------------------------------------
+// Series approximation of log1p()
+//  log1p() is not vectorized in libc
+//
+//  The series expansion is accurate to 1e-7 in the range sqrt(2)/2 < J < sqrt(2),
+//  with machine precision accuracy near J=1.  The initialization extends this range
+//  to 0.35 ~= sqrt(2)/4 < J < sqrt(2)*2 ~= 2.83, which should be sufficient for
+//  applications of the Neo-Hookean model.
```
jedbrown: Doh, when you deleted the range extension, you didn't update the comment explaining what it did. This version doesn't have the extended range.

PR #475, opened 2020-03-11
Fix Make Style (....Again)

jeremylt:
This is the latest fix to get `make style` working out of the box.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ addons:
     - libopenblas-dev
     - valgrind-dbg
     - lcov
+    - astyle
   homebrew:
     packages:
     - ccache
@@ -105,6 +106,9 @@ script:
   - make -j2 prove-all PROVE_OPTS=-v
   - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
         clang-tidy --version && make -j2 tidy;
+    fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
+        make style && git diff --exit-code;
     fi
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,7 @@ script:
   - make -j2
   - make -j2 prove-all PROVE_OPTS=-v
   - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
-        clang-tidy --version && make -j2 tidy;
+        clang-tidy --version && TIDY_OPTS="-fix-errors" make -j2 tidy && git diff --exit-code;
     fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
         make style && git diff --exit-code;
--- a/examples/navier-stokes/navierstokes.c
+++ b/examples/navier-stokes/navierstokes.c
@@ -513,21 +513,26 @@ PetscErrorCode SetUpDM(DM dm, problemData *problem, const char *prefix,
     ierr = DMAddField(dm,NULL,(PetscObject)fe); CHKERRQ(ierr);
     ierr = DMCreateDS(dm); CHKERRQ(ierr);
     /* Wall boundary conditions are zero velocity and zero flux for density and energy */
-    // *INDENT-OFF*
-    ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, "wall", "Face Sets", 0,
-                         3, (PetscInt[]) {1,2,3},
-                         (void(*)(void))problem->bc, bc->nwall, bc->walls,
-                         ctxSetup); CHKERRQ(ierr);
-    ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, "slipx", "Face Sets", 0,
-                         1, (PetscInt[]) {1}, (void(*)(void))NULL, bc->nslip[0],
-                         bc->slips[0], ctxSetup); CHKERRQ(ierr);
-    ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, "slipy", "Face Sets", 0,
-                         1, (PetscInt[]) {2}, (void(*)(void))NULL, bc->nslip[1],
-                         bc->slips[1], ctxSetup); CHKERRQ(ierr);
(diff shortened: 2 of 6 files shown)


examples/navier-stokes/navierstokes.c
```diff
@@ -514,20 +514,20 @@ PetscErrorCode SetUpDM(DM dm, problemData *problem, const char *prefix,
     ierr = DMCreateDS(dm); CHKERRQ(ierr);
     /* Wall boundary conditions are zero velocity and zero flux for density and energy */
     ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, "wall", "Face Sets", 0,
-                         3,(PetscInt[]) {1,2,3},
-                         (void(*)(void))problem->bc, bc->nwall, bc->walls, ctxSetup);
+    3,(PetscInt[]) {1,2,3},
+    (void(*)(void))problem->bc, bc->nwall, bc->walls, ctxSetup);
```
jedbrown: Is this what you want? I don't know any editors or style guides that behave this way. Sometimes I think we should abandon astyle due to its inconsistencies. clang-format is reliable, but draconian, especially with respect to error checking macros.

examples/petsc/setupsphere.h
```diff
@@ -103,8 +103,10 @@ static const char *const bpTypes[] = {"bp1","bp2","bp3","bp4","bp5","bp6",
 typedef struct {
   CeedInt ncompu, qdatasize, qextra;
   CeedQFunctionUser setupgeo, setuprhs, apply, error, ident;
+  // *INDENT-OFF*
   const char *setupgeofname, *setuprhsfname, *applyfname, *errorfname,
-        *identfname;
+             *identfname;
```
valeriabarra: Thanks @jeremylt, I think you may actually drop this. It appears not to be used

examples/navier-stokes/navierstokes.c
jeremylt: Yeah, I didn't like that either. I fixed it.

I'm happy with using astyle or something else, but whatever we do, I think we need to make `make style` part of the CI. Right now it gets left off frequently and then one of us tries to check the style of new code and gets hit with a bevy of old changes from another PR because we don't always `make style` our PRs.

examples/petsc/setupsphere.h
jeremylt: Right.

.travis.yml
```diff
@@ -106,6 +106,9 @@ script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
         clang-tidy --version && make -j2 tidy;
     fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
+        make style;
```
jedbrown: This won't commit the changes or error, it'll just make them locally. Perhaps you want `git diff --exit-code` after running the above so that it creates an error when changes are made?

examples/navier-stokes/navierstokes.c
```diff
@@ -513,22 +513,21 @@ PetscErrorCode SetUpDM(DM dm, problemData *problem, const char *prefix,
...
-    CHKERRQ(ierr);
+                         1, (PetscInt[]) {2}, (void(*)(void))NULL, bc->nslip[1],
+                         bc->slips[1], ctxSetup); CHKERRQ(ierr);
     ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, "slipz", "Face Sets", 0,
-                         1, (PetscInt[]) {3},
-                         (void(*)(void))NULL, bc->nslip[2], bc->slips[2], ctxSetup);
-    CHKERRQ(ierr);
+                         1, (PetscInt[]) {3}, (void(*)(void))NULL, bc->nslip[2],
+                         bc->slips[2], ctxSetup); CHKERRQ(ierr);
+    // *INDENT-ON*
```
jedbrown: The formatting here chokes because astyle doesn't understand compound literals `(PetscInt[]){1,2,3}`. I'd say better to toss them in named variables than to manually format them.

.travis.yml
jeremylt: updated

examples/navier-stokes/navierstokes.c
jeremylt: That fixed it.

jedbrown:
Irritating that astyle also can't tell when we're doing multiplication versus declaring a pointer. I re-checked what clang-format can do, and it'd serve our purposes if we converted all
```c
  ierr = CeedFunction(...); CeedChk(ierr);
```
to
```c
  CeedChk(CeedFunction(...));
```
I suppose we could consider that change. I'd adopted PETSc conventions when we got started, but either works and the latter at least gets the error messages on the same line. This isn't a problem for PETSc, which allows infinitely long lines, but we have an 80-char limit based on contributor preferences.

`clang-format` will also insist on spaces around binary operations, including indexing `a[i][j + 1][4]`, though this doesn't bother me much.

.travis.yml
```diff
@@ -104,7 +105,10 @@ script:
   - make -j2
   - make -j2 prove-all PROVE_OPTS=-v
   - if [[ "$TRAVIS_OS_NAME" == "linux" && "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
-        clang-tidy --version && make -j2 tidy;
+        clang-tidy --version && TIDY_OPTS="-fix-errors" make -j2 tidy && git diff --exit-code;
```
jeremylt: A few times we've have `clang-tidy` fail but `make tidy` not forward the error code. This should catch that.

jeremylt:
@jedbrown, there is also `CeedChk_Nvrtc` and `CeedChk_Cu` that would have to be updated, if we went that route, though I think the change would be roughly the same.

jeremylt:
deal-ii had a long [discussion](https://github.com/dealii/dealii/issues/6597) about clang-format vs astyle. 

I think it sounds like a better fit for us... I'm just not super excited about going digging through the codebase and changing all of the instances of `CeedChk()`... do you think we could reasonably grep and fix errors? We'd need to replace `ierr = ` with `CeedChk(` and `;*CeedChk(ierr);` with `);`.

jedbrown:
I believe I can automate that change with some Perl (assuming everyone is okay with it), but let's get all our loose ends merged first.

jeremylt:
For my own education, I'd love to see that Perl script. I've only heard secondhand horror stories about Perl.

jedbrown:
It'd just be a regex with some multiline shenanigans.

PR #478, opened 2020-03-13
Style - run make style

jeremylt:
One style change slipped by as the two branches were merged.


jeremylt:
Just double checked, currently the `make style` check on Travis is generating a failure. This fixes it.

PR #480, opened 2020-03-14
Navier-Stokes - fix indexing bug

jeremylt:
I made a bug. This fixes it. It turns out that C uses 0 based indexing.


examples/navier-stokes/navierstokes.c
```diff
@@ -524,11 +524,11 @@ PetscErrorCode SetUpDM(DM dm, problemData *problem, const char *prefix,
       ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, "slipx", "Face Sets", 0,
```
jedbrown: Benign, but the above is like honey spilled under the cabinets for bugs.

valeriabarra: Lol, thanks for shedding that image now that this neat-freak is even more concerned about hygiene than usual. I'm going to check under my cabinets now.

PR #481, opened 2020-03-14
Add VLA support for CUDA backends

jeremylt:
I need to fine tune this, but this is my rough idea for adding VLA support in the QFunctions for the backends that use CUDA/C++.


jeremylt:
Right. I was thinking of something similar, but converting my above `#define Q` into `#define Q_vla` and adding a `#if !define(Q_vla) ...` in the Qfunction. I just didn't finish the logic before bailing out for dinner. I think we have roughly the same idea once I fix this. My instinct is to focus on just the VLA arguments since `Q` is used in other parts of the QFunction body, like the loop bounds.

jedbrown:
Oh, that should work too. I was thinking we'd prefer literals everywhere, but the compiler should have that information from inlining and this gives an opportunity for a warning (if we check for those) in case of array bounds mismatch (i.e., we make a mistake).

YohannDudouit:
Can't we just use `const int Q = 1;`?

jedbrown:
@YohannDudouit  You'd have to put that inside the function, which I assume isn't a code transformation we want to be doing.

jeremylt:
There we go. I think I have it, but I need to run the tests on CUDA. To that, I need to rebuild my PETSc on CU Summit. I'll report back with results when they exist.

YohannDudouit:
@jedbrown I think if you put it outside the function it's just a global variable.

include/ceed.h
```diff
@@ -61,6 +61,10 @@
   static int name
 #endif
 
+#ifndef CEED_Q_VLA
+#  define CEED_Q_VLA Q
+#endif
```
jedbrown: LGTM, but please document this with a comment here and a note in the documentation (VLA is a C99 feature that we find useful, but is not compatible with the C++ dialect used by CUDA.)

jedbrown:
@YohannDudouit Sure, but if you also have `Q` as an argument to the qfunction, that one (the variable argument) will be in scope for the VLA-pointer declaration.

examples/navier-stokes/advection.h
```diff
@@ -20,11 +20,52 @@
...
 
-static int Exact_Advection(CeedInt dim, CeedScalar time, const CeedScalar X[],
-                           CeedInt Nf, CeedScalar q[], void *ctx) {
-  const SetupContext context = ctx;
+// *****************************************************************************
+// TODO: Annotate this function
+// *****************************************************************************
+static inline int Exact_Advection(CeedInt dim, CeedScalar time,
+                                  const CeedScalar X[], CeedInt Nf,
+                                  CeedScalar q[], void *ctx) {
```
jeremylt: The VLA stuff works fine now. But. These helper functions are causing NVRTC to choke
```
+default_program(124): error: A function without execution space annotations (__host__/__device__/__global__) is considered a host function, and host functions are not allowed in JIT mode. Consider using -default-device flag to process unannotated functions as __device__ functions in JIT mode
```
@jedbrown, thoughts on what we want to do here?
I think supporting helper functions for QFunctions fosters more maintainable code. I'm hesitant to go macro crazy, but do we need something like `CEED_QFUNCTION()` for helper functions at the library level? We could also put it here in the QFunction source if these will be uncommon, but that breaks the abstraction that QFunction source is backend agnostic.

jedbrown: If we can set default device, I think that would be ideal, otherwise the macro. These helpers are important.

YohannDudouit: You can also use the `QFunction` macro I think, which might be cleaner.

jedbrown: Yeah, though it'll also create an unused `_loc`.

YohannDudouit: We could consider adding some kind of similar `CEED_FUNCTION` that does not create the `_loc`.

include/ceed.h
jeremylt: I went ahead and added all of our macros to the documentation.

examples/navier-stokes/advection.h
YohannDudouit: I guess `-default-device` flag is also okay since it only concerns `nvrtc`.

jedbrown: Yeah, that's the fallback plan. I don't see a reason to do that if we can make `__device__` implicit during JIT. (I don't see a downside to that. I don't think the JIT scope ever needs to include host functions.)

jeremylt:
Now it compiles but we have
```
# +/projects/jeth8984/libCEEDDev/backends/cuda/ceed-cuda.c:100 in CeedRunKernelCuda(): CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES
```

examples/navier-stokes/advection.h
jeremylt: I added `-default-device` as a flag.

jeremylt:
Update - the new test shows all CUDA backends but cuda/gen are ok for my new test. The issue with NS passing is a question of size rather than the VLA macro.

jeremylt:
Update - CUDA backends now have VLA support, but our current Navier-Stokes test case does not work with these backends

examples/navier-stokes/densitycurrent.h
```diff
@@ -123,13 +160,16 @@ static int Exact_DC(CeedInt dim, CeedScalar time, const CeedScalar X[],
   return 0;
 }
 
+// *****************************************************************************
+// TODO: Annotate this function
+// *****************************************************************************
```
jeremylt: @LeilaGhaffari @valeriabarra For NS we are mostly pretty good about providing a documentation block for all functions in the QFunction header files, but a few slipped through in the big update. I'm less familiar with these newer functions - would one of you have a bit of time to add documentation blocks to the functions I've marked?

jeremylt:
@LeilaGhaffari, @valeriabarra, I'm soliciting a review from you two since I did some reshuffling of the content of the Navier-Stokes headers (to make the JiT work correctly) and changed up a bit of documentation.

examples/navier-stokes/densitycurrent.h
valeriabarra: Hey @jeremylt ! I have not authored the newer functions, so I would probably not be the best person to help with that. Perhaps @LeilaGhaffari and/or @jedbrown have a clearer idea than me.

jedbrown: If we could write the entire rst document interspersed in comments, it might be appropriate, but my inclination is to not try to explain everything in comments because keeping {code, comments, rst} in sync is harder than {code, rst}. So I'd rather just say that someone editing this code should open the sphinx docs and display them side-by-side. Note that the sphinx docs could have `literalinclude` snippets from the source to clarify coding techniques.

jeremylt: Maybe we want just a single description sentence for these then with the docs providing full details? The lack of a description meant I had to play a bit of hide and seek to figure out what was going on as I was trying to get these files working on CUDA.

LeilaGhaffari: I agree with @jedbrown . It's hard to sync code, comments and rst while the comments are not really as helpful as rst.

valeriabarra: I second @jeremylt on the fact that at least a concise sentence before these functions saying at least why now we have an "exact" version wouldn't hurt.

doc/sphinx/source/api/CeedQFunction.rst
```diff
@@ -11,3 +11,12 @@ Resolution/space-independent weak forms and quadrature-based operations
    :path: ../../../../xml
    :content-only:
    :members:
+
+Macros
+--------------------------------------
+
+.. doxygendefine:: CEED_QFUNCTION
```
valeriabarra: Thanks for adding these macros to the documentation @jeremylt. Do you think this PR could also fit the documentation for the QFunction Gallery that we were discussing on Slack?

jedbrown: Seems out of scope; I'd rather merge this to release #474 and anything else that might depend on it.

examples/navier-stokes/densitycurrent.h
```diff
@@ -24,12 +24,49 @@
...
+
+#ifndef advection_context_struct
+#define advection_context_struct
+typedef struct AdvectionContext_ *AdvectionContext;
+struct AdvectionContext_ {
+  CeedScalar CtauS;
+  CeedScalar strong_form;
+  int stabilization; // See StabilizationType: 0=none, 1=SU, 2=SUPG
+};
+#endif
```
jedbrown: Yuck, but I see why it's needed at present. I have sights on refactoring to make it easier to define new configurations (initial/boundary conditions and quantities of interest), but I think that'll have to be after v0.6, which Tzanio would like released this week.

examples/navier-stokes/densitycurrent.h
jeremylt: Yeah, I try to add a one line description to my functions to help future me and new collaborators. I've refractored stuff too many times to feel comfortable without something like that

doc/sphinx/source/api/CeedQFunction.rst
jeremylt: Agreed. Maybe a small separate PR?

examples/navier-stokes/densitycurrent.h
jeremylt: I'm not wild about the guarded multiple definitions either but nothing better comes to mind with our current infrastructure.

jeremylt:
I'll try to add some basic function descriptions in the morning so we can get this and solids merged.

jeremylt:
I'll merge once CI finishes.

examples/navier-stokes/advection.h
```diff
@@ -60,7 +60,29 @@ struct AdvectionContext_ {
...
+//      0.0 flux
+//    Momentum Density:
+//      0.0
+//    Energy Density:
+//      0.0 flux
+//
+// *****************************************************************************
+
+// *****************************************************************************
+// This helper function provides the current 3D advection IC formulation
```
valeriabarra: current?

jeremylt: Yeah. I'm guessing since I didn't write the code, but it seems like a big purpose of these helper functions is the ability to more easily switch out ICs with less boilerplate.

jeremylt: If there is a more natural description, anyone is welcome to replace it

PR #482, opened 2020-03-16
Leila/doc ns

LeilaGhaffari:
This PR includes documentation of `stabilization` techniques in our `navier-stokes` example and also a correction in the comments of `densitycurrent`.

I have not added a title for stabilization. I thought adding the `SU` and `SUPG` right after introducing the Galerkin form was appropriate. Let me know if you think a title is necessary.

Also, do you mind if I move the source terms `S(q)` to the left hand side of all equations prior to `SUPG`. Only for the sake of consistency! I had to keep all terms in the left hand side.

I have a question! I have seen in FEM literature that equation (15) (of the doc of Navier-Stokes example)  is called `strong form` while in this documentation, equation (16) is mentioned as the strong form. Are these two common?

### DIFF
--- a/examples/navier-stokes/index.rst
+++ b/examples/navier-stokes/index.rst
@@ -130,9 +130,8 @@ Integrating by parts on the divergence term, we arrive at the weak form,
    :label: eq-weak-vector-ns
 
    \begin{aligned}
-   \int_{\Omega} \boldsymbol v \cdot \frac{\partial \boldsymbol{q}_N}{\partial t}  \,dV
-   - \int_{\Omega} \nabla \boldsymbol v \!:\! \boldsymbol{F}(\boldsymbol{q}_N)\,dV
-   - \int_\Omega \boldsymbol v \cdot \mathbf{S}(\boldsymbol{q}_N) \, dV & \\
+   \int_{\Omega} \boldsymbol v \cdot \left( \frac{\partial \boldsymbol{q}_N}{\partial t} - \mathbf{S}(\boldsymbol{q}_N) \right)  \,dV
+   - \int_{\Omega} \nabla \boldsymbol v \!:\! \boldsymbol{F}(\boldsymbol{q}_N)\,dV & \\
    + \int_{\partial \Omega} \boldsymbol v \cdot \boldsymbol{F}(\boldsymbol q_N) \cdot \widehat{\mathbf{n}} \,dS
      &= 0 \, , \; \forall \boldsymbol v \in \mathcal{V}_p \,,
    \end{aligned}
@@ -157,9 +156,8 @@ Our formulation follows :cite:`hughesetal2010`, which offers a comprehensive rev
        :label: eq-weak-vector-ns-supg
 
        \begin{aligned}
-       \int_{\Omega} \boldsymbol v \cdot \frac{\partial \boldsymbol{q}_N}{\partial t}  \,dV
-       - \int_{\Omega} \nabla \boldsymbol v \!:\! \boldsymbol{F}(\boldsymbol{q}_N)\,dV
-       - \int_\Omega \boldsymbol v \cdot \mathbf{S}(\boldsymbol{q}_N) \, dV & \\
+       \int_{\Omega} \boldsymbol v \cdot \left( \frac{\partial \boldsymbol{q}_N}{\partial t} - \mathbf{S}(\boldsymbol{q}_N) \right)  \,dV
+       - \int_{\Omega} \nabla \boldsymbol v \!:\! \boldsymbol{F}(\boldsymbol{q}_N)\,dV & \\
        + \int_{\partial \Omega} \boldsymbol v \cdot \boldsymbol{F}(\boldsymbol{q}_N) \cdot \widehat{\mathbf{n}} \,dS & \\
        + \int_{\Omega} \boldsymbol{P}(\boldsymbol v)^T \, \left( \frac{\partial \boldsymbol{q}_N}{\partial t} \, + \,
        \nabla \cdot \boldsymbol{F} \, (\boldsymbol{q}_N) - \mathbf{S}(\boldsymbol{q}_N) \right) \,dV &= 0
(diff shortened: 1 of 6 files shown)


valeriabarra:
You are right about the strong/weak form. That was a slip on my end. Eq. (15) is the strong form. Eq. (16) is the weak form, and (17) is the weak form to which integration by parts has been applied.

valeriabarra:
On line 8 of `examples/navier-stokes/index.rst` we want to use `unstructured` now that we support unstructured meshes.

examples/navier-stokes/index.rst
```diff
@@ -132,13 +132,80 @@ And its weak form is:
...
+    In this method, the weighted residual of the original system of equations 
+    :math:numref:`eq-vector-ns` is added to the Galerkin formulation 
+    :math:numref:`eq-weak-vector-ns`. The weak form for this method is given as
+
+    .. math::
+       :label: eq-weak-vector-ns-supg
+
+       \begin{multline}
+          \int_{\Omega} v \, \frac{\partial \boldsymbol{q}_N}{\partial t}  \,dV + 
+          \int_{\Gamma} v \, \widehat{\mathbf{n}} \cdot \boldsymbol{F} \, (\boldsymbol{q}_N) \,dS -
```
valeriabarra: Suggested change to make it consistent with the standalone libCEED example's notation (not only in this equation, but also in eq. (17) and the following one):

`\Gamma` -> `\partial \Omega`

valeriabarra:
Thanks for adding this in your PR @LeilaGhaffari ! Now that we also support the implicit time-stepping formulation, we need to update also the time discretization description including at least the default implicit time-stepper we use, and show its command line options.

valeriabarra:
Two things to take care of:

- When building the doc from this branch there are warnings like:

```
/home/valeria/Research/CEED/libCEED/doc/sphinx/source/api/CeedElemRestriction.rst:9: WARNING: Duplicate declaration, int CeedElemRestrictionGetIMode(CeedElemRestriction rstr, CeedInterlaceMode  * imode)
```
and other similar ones. But @jedbrown had gotten rid of those in the branch merged to master.
- Resolve merge conflicts with master

jedbrown:
FYI, I just rebased onto `master`. It looks like maybe you rebased too far back because you had duplicates of the last week of `master` in your branch. The diff looks much cleaner now.

examples/navier-stokes/index.rst
jedbrown: Thanks; done.

LeilaGhaffari:
> FYI, I just rebased onto `master`. It looks like maybe you rebased too far back because you had duplicates of the last week of `master` in your branch. The diff looks much cleaner now.

Ugh! I guess I haven't understood `rebase` completely. Yeah, I think that is what happened.

jedbrown:
Should be live in a moment: https://libceed.readthedocs.io/en/leila-doc_ns

LeilaGhaffari:
> On line 8 of `examples/navier-stokes/index.rst` we want to use `unstructured` now that we support unstructured meshes.

Thank you @valeriabarra for your review. I will change them shortly.

LeilaGhaffari:
> Should be live in a moment: https://libceed.readthedocs.io/en/leila-doc_ns

@jedbrown , This is great. I am glad you added that blue note in the middle. I remember it was confusing for me when I first joined.
I only need to make few changes then I think it's ready to go.

examples/navier-stokes/index.rst
```diff
@@ -119,30 +119,32 @@ and with
...
-        \int_\Omega v \mathbf{S}(\boldsymbol{q}_N) \, dV \, , \; \forall v \in \mathcal{V}_p
+    \int_{\Omega} \boldsymbol v \cdot \frac{\partial \boldsymbol{q}_N}{\partial t}  \,dV + \int_{\partial \Omega} \boldsymbol v \cdot \boldsymbol{F}(\boldsymbol q_N) \cdot \widehat{\mathbf{n}} \,dS - \int_{\Omega} \nabla \boldsymbol v \!:\! \boldsymbol{F}(\boldsymbol{q}_N)\,dV  =
+        \int_\Omega \boldsymbol v \cdot \mathbf{S}(\boldsymbol{q}_N) \, dV \, , \; \forall \boldsymbol v \in \mathcal{V}_p \,,
    \end{multline}
 
-We solve equation :math:numref:`eq-weak-vector-ns` with Galerkin method discretization
-which is the default option in our Navier-Stokes example.
+where :math:`\boldsymbol{F}(\boldsymbol q_N) \cdot \widehat{\mathbf{n}}` is typically replaced with a boundary condition.
+
+.. note::
```
valeriabarra: Thanks @jedbrown for adding this note on the notation. I would add this also to the Theoretical Framework section where this symbol is first used.

jedbrown: Do you know a clean way to display the same admonition in multiple places, or does it need to be in its own file that gets included?

valeriabarra: I would say it has to be in the same file where it's displayed. Also, now that I can ask you this related question. Why the need for renaming the file in your commit [10423a](https://github.com/CEED/libCEED/commit/10423a34c0f42ed3ce2e01fda83fe72572c90501). Sorry if I had missed this detail before. Thanks.

jedbrown: The rename was so Sphinx doesn't complain about the file not appearing in a toctree. That's the pattern used in other projects. I adopted the naming convention used by SciPy.
https://stackoverflow.com/questions/15249340/warning-document-isnt-included-in-any-toctree-for-included-file
https://github.com/sphinx-doc/sphinx/issues/5281

valeriabarra: Ah! That was a good hack! I had bumped into the same thing earlier

LeilaGhaffari:
Could you review my last commit where I have added the implicit time stepping method?

LeilaGhaffari:
> Could you review my last commit where I have added the implicit time stepping method?

Ouch! something went wrong in my last commit! Sorry about that! I will correct it shortly!

jedbrown:
I revised the description of implicit time integration. With our move to KaTeX, we also have the freedom to use `\bm q` instead of `\boldsymbol q`. I'll approve now and wait for @valeriabarra's input.

LeilaGhaffari:
> I revised the description of implicit time integration. With our move to KaTeX, we also have the freedom to use `\bm q` instead of `\boldsymbol q`. I'll approve now and wait for @valeriabarra's input.

Thank you, @jedbrown , for improving my poor description. I wasn't even thinking of explaining alpha method.

examples/navier-stokes/index.rst
valeriabarra: @LeilaGhaffari would you mind adding the same note about the ":" notation in the FEMtheory.rst.inc file, right after it is used in eq. (1)? Thanks!

jedbrown:
FWIW, I didn't explain the alpha method. `\alpha` in my description is just an arbitrary shift, and the formulation applies to any diagonally implicit general linear method (a class that includes multistep and most Runge-Kutta methods). This formulation is discussed in the PETSc user manual (and a tech report that I should put an arxiv because we'll probably never submit to a journal). The PETSc user manual is in the process of converting to Sphinx, so I'll be able to direct-link it soon. ;-)

LeilaGhaffari:
> FWIW, I didn't explain the alpha method. `\alpha` in my description is just an arbitrary shift, and the formulation applies to any diagonally implicit general linear method (a class that includes multistep and most Runge-Kutta methods). This formulation is discussed in the PETSc user manual (and a tech report that I should put an arxiv because we'll probably never submit to a journal). The PETSc user manual is in the process of converting to Sphinx, so I'll be able to direct-link it soon. ;-)

Oh, you are right! Multistep methods are not included in alpha method. Thank you for the clarification.

jedbrown:
BDF and alpha methods are examples of linear multistep methods. All such methods, as well as diagonally implicit Runge-Kutta and General Linear (a class that includes multistep, Runge-Kutta, and hybrids), involve solving implicit equations with this structure.

PR #483, opened 2020-03-17
CUDA - enable multiple output modes for cuda/gen

### LINKED ISSUE #374: CUDA Gen Multiple Output Mode Error

jeremylt:
This is a patch. We will want to make a more efficient solution in the long term.

I'm running the test suite now.

Fixes #374

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator.c
+++ b/backends/cuda-gen/ceed-cuda-gen-operator.c
@@ -49,7 +49,7 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector invec,
   ierr = CeedQFunctionGetFields(qf, &qfinputfields, &qfoutputfields);
   CeedChk(ierr);
   CeedEvalMode emode;
-  CeedVector vec;
+  CeedVector vec, outvecs[16] = {};
 
   //Creation of the operator
   ierr = CeedCudaGenOperatorBuild(op); CeedChk(ierr);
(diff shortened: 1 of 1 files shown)


jedbrown:
Does the Navier-Stokes example work with this change?

jeremylt:
This fixes #374, I'm testing it for NS next

jeremylt:
Yeah, we will want to enable a more optimized solution in the future, but this fixes the behavior at least.

jeremylt:
@jedbrown, NS now runs on cuda/gen

PR #485, opened 2020-03-18
Some jupyter-notebook tutorials

valeriabarra:
Here is an initial PR containing a jupyter-notebook tutorial for `examples/ceed/ex1-volume` and `examples/ceed/ex2-surface`. 

TODO:

- [x]  Add simple python tutorials, which will (partly) take care of #459

PR #486, opened 2020-03-20
Ceed - add CeedView()

jeremylt:


python/ceed.py
```diff
@@ -44,6 +44,22 @@ def __init__(self, resource = "/cpu/self"):
...
+  # String conversion for print() to stdout
+  def __str__(self):
+    """View a Basis via print()."""
+
+    # libCEED call
+    with tempfile.NamedTemporaryFile() as key_file:
+      with open(key_file.name, 'r+') as stream_file:
+        stream = ffi.cast("FILE *", stream_file)
+
+        lib.CeedView(self._pointer[0], stream)
```
jedbrown: Yikes. Can we not do this with `io.BytesIO`? This should be like `fmemopen` in C.

jeremylt: io objects don't have a `fileno`, so I haven't been able to get them to cast as a `FILE` stream.

jedbrown: Hmm, sad, but makes sense if Python doesn't use libc `FILE*` under the hood. We could add support on the C side to return the output in a newly allocated buffer. The implementation would use:
```c
void *ptr;
size_t size;
FILE *f = open_memstream(&ptr, &size);
CeedView(ceed, f);
close(f);
// ptr is a buffer of length size, to be released with free(ptr)
```

jeremylt: Since we use this stream handling in Python for all of the other `Ceed*View()` functions, it would touch a lot of files. Do we want to do that change in a separate PR? It lets @valeriabarra wrap up #485.

valeriabarra: Yes, we had already encountered this issue when we worked on all `__str__` functions in the Python interface. 
I agree with @jeremylt and for now, I would say let's merge this so that I can also merge #485 (if it looks good on all other aspects) and we can leave this improvement for later.

PR #490, opened 2020-03-26
Solids - fix documentation for log1p series approx without shift

jeremylt:


examples/solids/qfunctions/hyperSS.h
```diff
@@ -36,11 +36,6 @@ struct Physics_private {
 // -----------------------------------------------------------------------------
 // Series approximation of log1p()
 //  log1p() is not vectorized in libc
-//
-//  The series expansion is accurate to 1e-7 in the range sqrt(2)/2 < J < sqrt(2),
-//  with machine precision accuracy near J=1.  The initialization extends this range
```
jedbrown: The first sentence is correct. It's the range extension that was wrong.

jeremylt: updated

PR #493, opened 2020-03-28
python: contextmanager for v.array() and v.array_read()

jedbrown:
I think this is cleaner syntax and more Pythonic.


tests/python/test-1-vector.py
```diff
@@ -60,11 +56,9 @@ def test_101(ceed_resource):
   a = np.arange(10, 10 + n, dtype="float64")
   x.set_array(a, cmode=libceed.USE_POINTER)
 
-  b = x.get_array_read()
-  for i in range(len(b)):
-    assert b[i] == 10 + i
-
-  x.restore_array_read()
+  with x.array() as b:
```
valeriabarra: Shouldn't this be `x.array_read()` ?

jedbrown:
```suggestion
  with x.array_read() as b:
```

valeriabarra:
Thanks for doing this @jedbrown. I have run the tests and they run fine. Instead, I had troubles with the tutorials. It's gotta be some configuration/environment issue? Although I have tried both with my user environment and a virtual env one.

This is for tutorial4-python-basis.ipynb:

```
AttributeError                            Traceback (most recent call last)
<ipython-input-4-dea7f38a727a> in <module>
     25   bxl.apply(1, libceed.EVAL_INTERP, X, Xq)
     26 
---> 27   with Xq.array_read() as xq:
     28     for i in range(Qdim):
     29       xx = np.empty(dim, dtype="float64")

AttributeError: 'Vector' object has no attribute 'array_read'
```

valeriabarra:
While you are changing some python files... would you like to rename the tutorials by starting their indexing at 0? So they would match with their corresponding tests. Thank you.

jedbrown:
I assume your Jupyter environment doesn't have the updated version of `libceed` installed.

Yeah, I can renumber the tutorials.

examples/tutorials/tutorial3-python-elemrestriction.ipynb
```diff
@@ -42,17 +42,9 @@
...
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "y = [10. 11. 11. 12. 12. 13.]\n"
```
valeriabarra: I thought it was useful and the norm in tutorials to show the expected output, at least of the cells relevant to the learning material (I had excluded the build ones). So that if users changed something in the cell before running it for whatever reason, they would know what they should have expected out of it, or if they'd messed something up and caused an unexpected behavior.

jedbrown: I wouldn't say it's a norm. It's possible to have rendered notebooks as part of a Sphinx site, though I'd expect these to be most useful for those who click through on Binder. I added figures, at which point checking in the graphics is poor form for a source repository. (Some people publish separate repositories with static rendered notebooks, but that's conceptually like putting them on Sphinx.) Pretty typical practice is to shift-click through the tutorial and edit a cell when curious, not to edit before first execution.

valeriabarra: I know it should not be a typical practice to edit a cell before running it, but it may happen. I was just saying this because, to me personally, seeing the expected outputs from say, the PETSc tutorials, helped me understand what I should have expected and question when I was seeing something different, for instance

jedbrown: Where do you see expected output from PETSc tutorials?

valeriabarra: There used to be. I reported broken links to them as my very first issue in the PETSc repo 
https://gitlab.com/petsc/petsc/-/issues/235

valeriabarra: And the webpage still shows the [Expected output] text, but with broken links
https://www.mcs.anl.gov/petsc/documentation/tutorials/HandsOnExercise.html

jedbrown: Thanks; those links will work again tomorrow (when master becomes `petsc-current` on the website). It's only been broken for a few days. (Maintaining legacy websites is a minefield.)

I often forget that page exists. In the repository, those are all text output for testing. It's quite a different environment from Jupyter.

If we include output in the notebooks,
1. we should test it with `nbval`
2. it should not include rendered figures

An alternative is to use [nbsphinx](https://nbsphinx.readthedocs.io/en/0.5.1/usage.html) or related tools to render output on the Sphinx site without storing that output in the repository (it's generated by sphinx).

jedbrown: I'll leave it to you to decide what we do here for the release. The Binder link will always be to master so I don't feel it's super important how these tutorials look as of the release tag.

valeriabarra: Ok, no problem. Proceed as you think it's more important for the release please.

PR #498, opened 2020-03-29
Jed/petsc bps avoid redundant zero

jedbrown:
New numbers for BP3 with (unstructrured) `bps` on one node of Lassen:
```
  Performance:
    Pointwise Error (max)              : 2.424969e-05
    CG Solve Time                      : 0.653718 (0.653667) sec
    DoFs/Sec in CG                     : 8016.16 (8016.79) million
```
Uses: https://gitlab.com/petsc/petsc/-/merge_requests/2656#note_313507355

### DIFF
--- a/examples/petsc/bps.c
+++ b/examples/petsc/bps.c
@@ -311,7 +311,6 @@ int main(int argc, char **argv) {
     ierr = KSPSetTolerances(ksp, 1e-10, PETSC_DEFAULT, PETSC_DEFAULT,
                             PETSC_DEFAULT); CHKERRQ(ierr);
   }
-  ierr = KSPSetFromOptions(ksp); CHKERRQ(ierr);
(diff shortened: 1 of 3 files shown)


jedbrown:
@jeremylt There are a couple more `VecZeroEntries` that I think might be removable for multigrid, but I don't trust myself to think through the implications right now. They won't affect BP reporting with this release, but if you're confident they can be removed, feel free to add it to this PR.

jeremylt:
I believe a few of them are needed since I reuse work arrays, but I'll take a look.

jedbrown:
I guess the point here is `CeedOperatorApply` versus `CeedOperatorApplyAdd`.

jeremylt:
Right. The should be a couple of spots where I need to zero to clear boundary conditions for certain operators.

PR #499, opened 2020-03-30
Small doc fixes

valeriabarra:
In `examples/notation.rst`: A small shift in notation. This is needed since in the new description of [Ex2-surface](https://libceed.readthedocs.io/en/latest/examples/ceed/index.html) coordinate indices starting from 0 have been used

### DIFF
--- a/doc/sphinx/source/FEMtheory.rst.inc
+++ b/doc/sphinx/source/FEMtheory.rst.inc
@@ -18,7 +18,7 @@ We start by considering the discrete residual :math:`F(u)=0` formulation
 in weak form. We first define the :math:`L^2` inner product between real-valued functions
 
 .. math::
-   \langle u, v \rangle = \int_\Omega u v d \bm{x},
+   \langle v, u \rangle = \int_\Omega v u d \bm{x},
 
 where :math:`\bm{x} \in \mathbb{R}^d \supset \Omega`.
 
--- a/doc/sphinx/source/FEMtheory.rst.inc
+++ b/doc/sphinx/source/FEMtheory.rst.inc
@@ -18,9 +18,9 @@ We start by considering the discrete residual :math:`F(u)=0` formulation
 in weak form. We first define the :math:`L^2` inner product
 
 .. math::
-   \langle u, v \rangle = \int_\Omega u v d \mathbf{x},
+   \langle u, v \rangle = \int_\Omega u v d \bm{x},
 
-where :math:`d \mathbf{x} \in \mathbb{R}^d \supset \Omega`.
+where :math:`d \bm{x} \in \mathbb{R}^d \supset \Omega`.
 
 We want to find :math:`u` in a suitable space :math:`V_D`,
 such that
--- a/doc/sphinx/source/FEMtheory.rst.inc
+++ b/doc/sphinx/source/FEMtheory.rst.inc
@@ -15,42 +15,42 @@ of computational device types (selectable at run time). We present here the nota
 and the mathematical formulation adopted in libCEED.
 
 We start by considering the discrete residual :math:`F(u)=0` formulation
-in weak form. We first define the :math:`L^2` inner product
+in weak form. We first define the :math:`L^2` inner product between real-valued functions
 
 .. math::
    \langle u, v \rangle = \int_\Omega u v d \bm{x},
 
-where :math:`d \bm{x} \in \mathbb{R}^d \supset \Omega`.
+where :math:`\bm{x} \in \mathbb{R}^d \supset \Omega`.
 
 We want to find :math:`u` in a suitable space :math:`V_D`,
 such that
 
 .. math::
-   \langle v, f(u) \rangle = \int_\Omega v \cdot f_0 (u, \nabla u) + \nabla v : f_1 (u, \nabla u) = 0
+   \langle  \bm v,  \bm f(u) \rangle = \int_\Omega  \bm v \cdot  \bm f_0 (u, \nabla u) + \nabla \bm v :  \bm f_1 (u, \nabla u) = 0
    :label: residual
 
-for all :math:`v` in the corresponding homogeneous space :math:`V_0`, where :math:`f_0`
-and :math:`f_1` contain all possible sources in the problem. We notice here that
-:math:`f_0` represents all terms in :math:numref:`residual` which multiply the test
-function :math:`v` and :math:`f_1` all terms which multiply its gradient :math:`\nabla v`.
-For an n-component problems in :math:`d` dimensions, :math:`f_0 \in \mathbb{R}^n` and
-:math:`f_1 \in \mathbb{R}^{nd}`.
+for all :math:`\bm v` in the corresponding homogeneous space :math:`V_0`, where :math:`\bm f_0`
+and :math:`\bm f_1` contain all possible sources in the problem. We notice here that
+:math:`\bm f_0` represents all terms in :math:numref:`residual` which multiply the (possibly vector-valued) test
+function :math:`\bm v` and :math:`\bm f_1` all terms which multiply its gradient :math:`\nabla \bm v`.
+For an n-component problems in :math:`d` dimensions, :math:`\bm f_0 \in \mathbb{R}^n` and
+:math:`\bm f_1 \in \mathbb{R}^{nd}`.
 
 .. note::
-   The notation :math:`\nabla \bm v \!:\! f_1` represents contraction over both
+   The notation :math:`\nabla \bm v \!:\! \bm f_1` represents contraction over both
    fields and spatial dimensions while a single dot represents contraction in just one,
-   which should be clear from context, e.g., :math:`v \cdot f_0` contracts only over
+   which should be clear from context, e.g., :math:`\bm v \cdot \bm f_0` contracts only over
    fields.
 
 .. note::
    In the code, the function that represents the weak form at quadrature
    points is called the :ref:`CeedQFunction`. In the :ref:`Examples` provided with the
-   library (in the :file:`examples/` directory), we store the term :math:`f_0` directly
-   into `v`, and the term :math:`f_1` directly into `dv` (which stands for
-   :math:`\nabla v`). If equation :math:numref:`residual` only presents a term of the
-   type :math:`f_0`, the :ref:`CeedQFunction` will only have one output argument,
+   library (in the :file:`examples/` directory), we store the term :math:`\bm f_0` directly
+   into `v`, and the term :math:`\bm f_1` directly into `dv` (which stands for
+   :math:`\nabla \bm v`). If equation :math:numref:`residual` only presents a term of the
+   type :math:`\bm f_0`, the :ref:`CeedQFunction` will only have one output argument,
    namely `v`. If equation :math:numref:`residual` also presents a term of the type
-   :math:`f_1`, then the :ref:`CeedQFunction` will have two output arguments, namely,
+   :math:`\bm f_1`, then the :ref:`CeedQFunction` will have two output arguments, namely,
    `v` and `dv`.
 
 
--- a/doc/sphinx/source/FEMtheory.rst.inc
+++ b/doc/sphinx/source/FEMtheory.rst.inc
@@ -37,7 +37,7 @@ For an n-component problems in :math:`d` dimensions, :math:`f_0 \in \mathbb{R}^n
 :math:`f_1 \in \mathbb{R}^{nd}`.
 
 .. note::
-   The notation :math:`\nabla \boldsymbol v \!:\! f_1` represents contraction over both
+   The notation :math:`\nabla \bm v \!:\! f_1` represents contraction over both
    fields and spatial dimensions while a single dot represents contraction in just one,
    which should be clear from context, e.g., :math:`v \cdot f_0` contracts only over
    fields.
--- a/doc/sphinx/source/libCEEDapi.rst
+++ b/doc/sphinx/source/libCEEDapi.rst
@@ -52,9 +52,9 @@ versions of :math:`\bm{P}`, :math:`\bm{G}` and :math:`\bm{B}`.
 Note that in the case of adaptive mesh refinement (AMR), the restrictions
 :math:`\bm{P}` and :math:`\bm{G}` will involve not just extracting sub-vectors,
 but evaluating values at constrained degrees of freedom through the AMR interpolation.
-There can also be several levels of subdomains (:math:`\bm{P1}`, :math:`\bm{P2}`,
+There can also be several levels of subdomains (:math:`\bm P_1`, :math:`\bm P_2`,
 etc.), and it may be convenient to split :math:`\bm{D}` as the product of several
-operators (:math:`\bm{D1}`, :math:`\bm{D2}`, etc.).
+operators (:math:`\bm D_1`, :math:`\bm D_2`, etc.).
 
 
 Terminology and Notation
@@ -78,7 +78,7 @@ Vector representation/storage categories:
      element containing :math:`i`
 
    - this is an overlapping vector decomposition with overlaps only across
-     different processors - there is no duplication of unknowns on a single
+     different processors---there is no duplication of unknowns on a single
      processor
 
    - the shared dofs/unknowns are the overlapping dofs, i.e. the ones that have
@@ -109,7 +109,7 @@ Vector representation/storage categories:
 
    - in other words, an entry in an **E-vector** is obtained by copying an entry
      from the corresponding **L-vector**, optionally switching the sign of the
-     entry (for :math:`H(\mathrm{div})`- and :math:`H(\mathrm{curl})`-conforming spaces).
+     entry (for :math:`H(\mathrm{div})`---and :math:`H(\mathrm{curl})`-conforming spaces).
 
    .. image:: ../../img/L-vector-AMR.svg
 
@@ -125,13 +125,13 @@ Vector representation/storage categories:
 
 - In many cases it is useful to distinguish two types of vectors:
 
-   - X-vector, or **primal** X-vector, and X'-vector, or **dual** X-vector
+   - **X-vector**, or **primal X-vector**, and **X'-vector**, or **dual X-vector**
 
    - here X can be any of the T, L, E, or Q categories
 
-   - for example, the mass matrix operator maps a **T-vector** to a T'-vector
+   - for example, the mass matrix operator maps a **T-vector** to a **T'-vector**
 
-   - the solutions vector is a **T-vector**, and the RHS vector is a T'-vector
+   - the solutions vector is a **T-vector**, and the RHS vector is a **T'-vector**
 
    - using the parallel prolongation operator, one can map the solution
      **T-vector** to a solution **L-vector**, etc.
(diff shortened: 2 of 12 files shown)


valeriabarra:
@jedbrown I tried to clarify a bit the description by using the common notation used in the ex1 section. Please let me know if it looks good.

valeriabarra:
@jedbrown  in [bc4ae5](https://github.com/CEED/libCEED/pull/499/commits/bc4ae51c1535ccb22daeb39615360081f7a4ac4f) I changed the conf.py because we have not pushed the `_static` folder in the doc/sphinx/source directory. Thus, it was giving me that warning message.

doc/sphinx/source/FEMtheory.rst.inc
```diff
@@ -18,9 +18,9 @@ We start by considering the discrete residual :math:`F(u)=0` formulation
 in weak form. We first define the :math:`L^2` inner product
 
 .. math::
-   \langle u, v \rangle = \int_\Omega u v d \mathbf{x},
+   \langle u, v \rangle = \int_\Omega u v d \bm{x},
 
-where :math:`d \mathbf{x} \in \mathbb{R}^d \supset \Omega`.
+where :math:`d \bm{x} \in \mathbb{R}^d \supset \Omega`.
```
jedbrown, line 23:
```suggestion
where :math:`\bm{x} \in \mathbb{R}^d \supset \Omega`.
```

doc/sphinx/source/FEMtheory.rst.inc
```diff
@@ -37,7 +37,7 @@ For an n-component problems in :math:`d` dimensions, :math:`f_0 \in \mathbb{R}^n
 :math:`f_1 \in \mathbb{R}^{nd}`.
 
 .. note::
-   The notation :math:`\nabla \boldsymbol v \!:\! f_1` represents contraction over both
+   The notation :math:`\nabla \bm v \!:\! f_1` represents contraction over both
    fields and spatial dimensions while a single dot represents contraction in just one,
    which should be clear from context, e.g., :math:`v \cdot f_0` contracts only over
```
jedbrown, line 42:
```suggestion
   which should be clear from context, e.g., :math:`\bm v \cdot f_0` contracts only over
```
Should `f` be boldened too?

doc/sphinx/source/libCEEDapi.rst
```diff
@@ -50,11 +50,11 @@ versions of :math:`\mathbf{P}`, :math:`\mathbf{G}` and :math:`\mathbf{B}`.
...
 Note that in the case of adaptive mesh refinement (AMR), the restrictions
-:math:`\mathbf{P}` and :math:`\mathbf{G}` will involve not just extracting sub-vectors,
+:math:`\bm{P}` and :math:`\bm{G}` will involve not just extracting sub-vectors,
 but evaluating values at constrained degrees of freedom through the AMR interpolation.
-There can also be several levels of subdomains (:math:`\mathbf{P1}`, :math:`\mathbf{P2}`,
-etc.), and it may be convenient to split :math:`\mathbf{D}` as the product of several
-operators (:math:`\mathbf{D1}`, :math:`\mathbf{D2}`, etc.).
+There can also be several levels of subdomains (:math:`\bm{P1}`, :math:`\bm{P2}`,
+etc.), and it may be convenient to split :math:`\bm{D}` as the product of several
+operators (:math:`\bm{D1}`, :math:`\bm{D2}`, etc.).
```
jedbrown, line 57:
```suggestion
There can also be several levels of subdomains (:math:`\bm P_1`, :math:`\bm P_2`,
etc.), and it may be convenient to split :math:`\bm{D}` as the product of several
operators (:math:`\bm D_1`, :math:`\bm D_2`, etc.).
```

doc/sphinx/source/releasenotes.rst
```diff
@@ -7,7 +7,7 @@ for each release of libCEED.
 
 .. _v0.6:
 
-v0.6 (Mar 29, 2020)
+v0.6 (Mar 30, 2020)
```
jedbrown, line 10: It was Mar 29 on the west coast. ;-)

examples/ceed/index.rst
```diff
@@ -18,20 +18,22 @@ are supported from the same code.
...
+:math:`\bm{1}`\s. It computes:
 
 .. math::
-   I = \int_{\Omega} \mathbf{1} \, dV .
+   I = \int_{\Omega} \bm{1} \, dV .
    :label: eq-ex1-volume
 
 Using the same notation as in :ref:`Theoretical Framework`, we write here the vector
-:math:`u(\mathbf{x})\equiv \mathbf{1}` in the Galerkin approximation,
+:math:`u(\bm{x})\equiv \bm{1}` in the Galerkin approximation,
```
jedbrown, line 28: Seems like `1` is not a vector in this context.

examples/ceed/index.rst
```diff
@@ -18,20 +18,22 @@ are supported from the same code.
...
 and find the volume of :math:`\Omega` as
 
 .. math::
-   \sum_e \int_{\Omega_e} v(x) \cdot \mathbf{1} \, dV
+   :label: volume-sum
 
-with :math:`v(x) \in \mathcal{V}_p = \{ v \in H^{1}(\Omega_e) \,|\, v \in P_p(\boldsymbol{I}), e=1,\ldots,N_e \}`,
+   \sum_e \int_{\Omega_e} v(x) \cdot \bm{1} \, dV
```
jedbrown, line 34: Aren't both terms scalars, so this is not an inner product?

examples/ceed/index.rst
```diff
@@ -47,7 +49,7 @@ Arbitrary mesh and solution orders in 1D, 2D and 3D are supported from the same
 Similarly to :ref:`Ex1-Volume`, it computes:
 
 .. math::
-   I = \int_{\partial \Omega} \mathbf{1} \, dS .
+   I = \int_{\partial \Omega} \bm{1} \, dS .
```
jedbrown, line 52: 1 here is a scalar

examples/notation.rst
```diff
@@ -15,7 +15,8 @@ typically) by letting :math:`\Omega = \bigcup_{e=1}^{N_e}\Omega_e`, with :math:`
 disjoint elements. For most examples we use unstructured meshes for which the elements
 are hexahedra (although this is not a requirement in libCEED).
 
-The physical coordinates are denoted by :math:`\mathbf{x}=(x,y,z)\in\Omega_e`,
+The physical coordinates are denoted by
+:math:`\bm{x}=(x,y,z) \equiv (x_0,x_1,x_2) \in\Omega_e`,
 while the reference coordinates are represented as
-:math:`\boldsymbol{X}=(X,Y,Z) \equiv (X_1,X_2,X_3) \in\mathbf{I}=[-1,1]^3`
+:math:`\bm{X}=(X,Y,Z) \equiv (X_0,X_1,X_2) \in\bm{I}=[-1,1]^3`
```
jedbrown, line 21: Do we need to dedicate `\bm I` for this? It seems like we most always write out `[-1,1]^d` anyway, in which case that named symbol is unnecessary.

examples/petsc/index.rst
```diff
@@ -11,9 +11,9 @@ demonstrates a simple usage of libCEED with PETSc to calculate
 the surface area of a closed surface. The code uses higher level
 communication protocols for mesh handling in PETSc's DMPlex. This example has the
 same mathematical formulation as :ref:`Ex1-Volume`, with the exception that the
-physical coordinates for this problem are :math:`\mathbf{x}=(x,y,z)\in \mathbb{R}^3`,
+physical coordinates for this problem are :math:`\bm{x}=(x,y,z)\in \mathbb{R}^3`,
 while the coordinates of the reference element are
-:math:`\boldsymbol{X}=(X,Y) \equiv (X_1,X_2) \in\mathbf{I}=[-1,1]^2`.
+:math:`\bm{X}=(X,Y) \equiv (X_1,X_2) \in\bm{I}=[-1,1]^2`.
```
jedbrown, line 16:
```suggestion
:math:`\bm{X}=(X,Y) \equiv (X_0,X_1) \in\bm{I}=[-1,1]^2`.
```

doc/sphinx/source/releasenotes.rst
valeriabarra: Ahah I know... but GitHub is not forgiving :sweat_smile: I thought we should be consistent with it
Screenshot from 2020-04-01 09-17-15

doc/sphinx/source/FEMtheory.rst.inc
valeriabarra: Yes, but then also in eq. (1) right above, no?

examples/notation.rst
valeriabarra: I will remove the bold font for this, but I like to denote intervals. I think it's neater, and we may have to re-refer to it later. I'll check if we actually do in this case.

examples/ceed/index.rst
valeriabarra: But then also in eq. (2) above, no?

valeriabarra: I will fix that, too

doc/sphinx/source/releasenotes.rst
jedbrown: That's the client's timezone, FWIW. Releases pretty much always look a day off in Japan, for example; it's impossible to be globally consistent.
Screenshot_20200401_093315

valeriabarra: Ah, I wasn't sure if it  was the client's one or not. Ok then, I can put it back

examples/notation.rst
jedbrown: It looked like it repeated `\bm I = [-1, 1]^3` any time the interval was used, in which case the `I` was superfluous.

doc/sphinx/source/FEMtheory.rst.inc
jedbrown: Yeah, if you change notation, please make it consistent. I didn't read through to see everywhere else that we may use it.

examples/notation.rst
valeriabarra: I thought it was redefined only when necessary because different, for instance in the area problem `I=[-1,1]^2`. But I may have missed some repetitions. I'll look for those

valeriabarra:
@jedbrown I'm just realizing now something. In `libCEEDapi.rst:168` `precompute and store wdet(J), or det(J) (in the case of mass matrix) at all quadrature points in all mesh elements` (sorry, I can't seem to be able to copy the permalink of a source rst file), why the need to specify only det(J) in case of the mass matrix, when we actually still use wdet(J) for that?

jedbrown:
I don't think I wrote that. That section may have been ported from an MFEM/hypre context. Feel free to modify/update at will.

valeriabarra:
@jedbrown I've touched a lot of files again. Hopefully this time I was able to catch all inconsistencies. You may want to give a final look. Thank you

examples/bps.rst
```diff
@@ -28,8 +28,8 @@ regular :math:`u`, :math:`v`, and :math:`f`, we have:
 
 .. math::
    \begin{aligned}
-   \langle v,u \rangle &:= \int_{\Omega} \, v \, u \, dV ,\\
-   \langle v,f \rangle &:= \int_{\Omega} \, v \, f \, dV .
+   \langle u,v \rangle &:= \int_{\Omega} \, u \, v \, dV ,\\
+   \langle f,v \rangle &:= \int_{\Omega} \, f \, v \, dV .
```
jedbrown, line 32: Is there a reason for this notation change? I prefer the `(v, u)` ordering, corresponding to the complex conjugation `\int v^* u`. This ordering is more common in physics and engineering, and relates bilinear forms to matrices as `a(v, u) ~ v^T A u`.

The other convention pops up quite a bit in the math literature, but I think it takes more effort to relate to discrete operators and track conjugation/transposition.

valeriabarra: I was just trying to be consistent with the definition given above on line 21 of the `FEMtheory.rst.inc` file, where I had specified that that was the definition of the inner product for real-valued functions. 
FWIW, I am not sure I've seen many times the notation you mention, of defining `(v,u) = \int v^* u `. I've seen more often the definition as `(u,v) = \int v^* u `, so with the functions in the integrand swapped, not the ones in the bilinear form operator. This also matches the notation for the complex conjugate of an inner product, i.e., `\langle x, y \rangle = \bar{ \langle y, x \rangle }`, or the complex conjugate of the product of matrices, i.e., `(AB)^* = B^* A^*`

jedbrown: Yeah, it's more common in physics/engineering convention. I typically use it specifically because it maps more directly to the matrix expressions.
> Some authors, especially in physics and matrix algebra, prefer to define the inner product and the sesquilinear form with linearity in the second argument rather than the first. ...
> https://en.wikipedia.org/wiki/Inner_product_space#Alternative_definitions,_notations_and_remarks

> The inner product on Hilbert space (   ,   ) {\displaystyle (\ ,\ )} {\displaystyle (\ ,\ )} (**with the first argument anti linear as preferred by physicists**) ...
> https://en.wikipedia.org/wiki/Bra%E2%80%93ket_notation

https://www.netlib.org/lapack/explore-html/d2/df9/group__complex16__blas__level1_ga45b00bad9285ff50cd86e97dfb04facd.html#ga45b00bad9285ff50cd86e97dfb04facd

It is true that PETSc chose the more common math convention for VecDot.
https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecDot.html

valeriabarra: I would personally stick to the most common mathematical notation of `(first, second)` with the second argument being conjugate linear, rather than the first one. This embraces both real-valued and complex-valued cases. In fact, in the case of real-valued vectors/functions the order of the operands on the RHS doesn't really matter. For the complex-valued case, it resembles the matrix form: `\langle x, y \rangle = y^+ M x`. 
In the eventuality we wanted to use the bra-ket notation, we should remark that 
⟨·, ·⟩ and ⟨· | ·⟩ are distinct notations. But honestly since we don't even have quantum mechanics examples for now, I would not adopt a niche notation in that field for general-purpose documentation that we have right now.

jedbrown: Are you saying that the resemblance of the first is more natural than the second?
* `(A x, B y) ~ y^* (B^* A) x`
* `(A x, B y) ~ x^* (A^* B) y`

FWIW, it isn't a quirk in quantum mechanics, but also widespread in engineering literature; e.g., Hughes and associates:
Screenshot_20200402_114422
Screenshot_20200402_114445

valeriabarra: > Are you saying that the resemblance of the first is more natural than the second?

> * (A x, B y) ~ y^* (B^* A) x

Honestly, to me, yes, because I know that in the mathematical definition of inner product for  complex-valued functions/vectors, the order needs to be inverted as also wiki reports for the [Conjugate symmetry property](https://en.wikipedia.org/wiki/Inner_product_space#Definition), or for the [Complex coordinate space example](https://en.wikipedia.org/wiki/Inner_product_space#Complex_coordinate_space). I believe students are more likely to find classical mathematical definitions than more field-specific ones. And they can relate to similar properties like the transpose of the product of two matrices, or the conjugate transpose of the product of two matrices and so on..

jedbrown: I won't argue the point further, though perhaps other people would like to comment. The notation in which order is preserved `(v, u) ~ v^T M u` is not niche, just a reflection of how people are introduced to finite elements. I honestly don't know which is more common, and more importantly, the integrated transition cost between this and other notational differences. I often eschew the inner product notation due to this ambiguity, preferring to write `\int \bar v u` or whatever. Of course dropping the `d V` helps make the `\int` notation lighter.

valeriabarra: I just answered your question on which one looks more natural to me personally, but of course you know what's best for the libCEED's community, and we can revert that.

jedbrown: Fair enough. I'd prefer to either leave it or get a few more opinions.

valeriabarra:
Converted all (if I haven't missed any) `(u,v)` to `(v,u)`

valeriabarra:
I already created the SVG version, but couldn't post that on GitHub as apparently is not supported

valeriabarra:
> Kind of. I don't care about filled or not, but I'd prefer them to be at Gauss points (which don't include endpoints). What do the dashed lines to q0 and q3 signify on this diagram?

Ah, yes, Gauss, not LGL! Those dashed lines were supposed to show the quad element. I was going to explain this. Should I remove them if not clear at first glance?

valeriabarra:
[SphereSketch.pdf](https://github.com/CEED/libCEED/files/4429510/SphereSketch.pdf)

jedbrown:
I think it's fine for the sketch to address the surface of the 1-sphere (i.e., circle), so the reference element is just a line segment. I think I see now that this would be just four line elements. You could extend the circle arc all the way to the edge of the diagram, but make it become dashed (like the lines) once outside the current element.

The q0 and q3 markers probably need to be ditched or brought inside the element. Should we write `x(X)` and `\overset{\circ}{x}(x)`? And maybe dotted (instead of dashed; or a different color/shade) with an arrow showing the map from the line element to the arc.

valeriabarra:
I think one linear element (with two nodes `p0, p1` and 2 Gauss points `q0,q1` is simpler and easier for me. Does this look ok?
[SphereSketch.pdf](https://github.com/CEED/libCEED/files/4429944/SphereSketch.pdf)

valeriabarra:
I'm going to work on the arrows to show the coordinates mapping now.

valeriabarra:
What about this one?
[SphereSketch.pdf](https://github.com/CEED/libCEED/files/4429984/SphereSketch.pdf)

jedbrown:
It looks like an oblique project that should be radial. With that and maybe mark both dots on the circle, I'm ready to merge.

valeriabarra:
[SphereSketch.pdf](https://github.com/CEED/libCEED/files/4430140/SphereSketch.pdf)
Et voila

valeriabarra:
I'm going to add it to the docs now

jedbrown:
Where is the center of this circle?

valeriabarra:
> Where is the center of this circle?

(-3,0) with radius 3. 
The point p1 is at (0,0)

jedbrown:
Yep, so the line you drew can't be radial.

valeriabarra:
> Yep, so the line you drew can't be radial.

Yes, sorry, I had the coordinates slightly off in my TikZ picture. Thanks for noticing. Here it is, with the proper radial projection.
[SphereSketch.pdf](https://github.com/CEED/libCEED/files/4430320/SphereSketch.pdf)

jedbrown:
Do you mind my converting to `dirhtml`? It results in cleaner links (no `index.html`).
https://libceed.readthedocs.io/en/valeria-some_small_doc_fixes/examples/petsc/#sphere
I think we should stick with the plain `html` builder for our `make doc` because then you can just open files in the browser instead of needing to run a local server, though we could add a `make serve` if you want them to match.

valeriabarra:
No, it's fine this way thank you. I didn't realize you pushed something while I was trying to reduce the number of commits in this PR, and now some of the commits are mangled by the rebase. I would just squash and merge if that's fine with you.

jedbrown:
Yeah, no problem. Sorry to interfere.

PR #502, opened 2020-04-02
Python: apply autopep8 and add to CI

jedbrown:
RFC @valeriabarra @jeremylt 
I skipped benchmarks/ and backends/magma/gccm.py because they may need to be kept in sync with other repository.s


python/ceed.py
```diff
@@ -27,337 +27,341 @@
 from .ceed_constants import *
 
 # ------------------------------------------------------------------------------
+
+
```
jeremylt, line 31: I'm not sure why we're getting these two blank lines after every single 
```
# ----------------
# description
# ----------------
```
comment. That feels like a linting error.

jedbrown: I believe it's because our commenting style is discouraged by the PEP (which would prefer that be a docstring). Is it E265? https://pypi.org/project/autopep8/#features

jeremylt: Block comments don't require two blank lines after them. I think this is what's causing the extra spacing "Surround top-level function and class definitions with two blank lines.".

jeremylt: Now that I understand, I am ok with this. There are a couple of instances that look a little awkward, but most of it is adding padding around lines that look like
```
# ---------------
```
and those lines really help me find stuff in these files.

jedbrown:
One popular alternative is [black](https://black.readthedocs.io/en/stable/the_black_code_style.html), which is more aggressive (similar to `clang-format`). The style is quirky in some ways, mostly in its attempt to make as many future modifications affect minimal lines. I don't have a strong opinion, but some formatter seemed useful.

jeremylt:
I like working towards reducing the size of future diffs since we've seen a lot of churn over the last 6-12 months that's not really representative of actual development but rather style or changes in adjacent lines spilling over. If black's changes are reasonable looking, I'd vote we go that way.

WRT `clang-format`, have we decided to change over to that after #305? I like the idea of getting the style tools settled upon and stable so we can stop having so many of these little pervasive changes in various PRs.

jedbrown:
We'll have to convert our error handling to `CeedChk(Function(args, ...));`, but I think we could teach ourselves to see through that the way we do with trailing checks. I'd prefer to increase max line length a little (to cut down on those frequent wraps). I think clang-format will look okay with those points. Exactly this same conversation is playing out on the Arrow mailing list; you could look at their comparison PRs.

> * black: https://github.com/apache/arrow/pull/6810
> * autopep8: https://github.com/apache/arrow/pull/6811

jeremylt:
It looks like in their case the biggest difference is that
black(
  wants,
  every,
  argument,
  on,
  its,
  own,
  line,
  no,
  matter,
  what,
  that,
  does,
  do,
  the,
  length,
  of,
  your,
  files
)

PR #504, opened 2020-04-04
Solids - Add Twist BCs and Strain Energy

jeremylt:
Adding a couple of features to the Elasticity example to show performance.

### DIFF
--- a/doc/sphinx/source/conf.py
+++ b/doc/sphinx/source/conf.py
@@ -39,7 +39,7 @@ import subprocess
 # ones.
 extensions = [
     'breathe',
-    #'hoverxref.extension',  # still in beta; not rendering well on RTD
+    # 'hoverxref.extension',  # still in beta; not rendering well on RTD
     'recommonmark',
     'sphinx_markdown_tables',
     'sphinx_rtd_theme',
--- a/examples/solids/README.rst
+++ b/examples/solids/README.rst
@@ -40,14 +40,10 @@ The elasticity min-app is controlled via command-line options, the following of 
    * - :code:`-nu [real]`
      - `Poisson's ratio <https://en.wikipedia.org/wiki/Poisson%27s_ratio>`_, :math:`\nu < 0.5`
 
-   * - :code:`-bc_zero [int list]`
-     - List of faces sets on which to enforce zero displacement
-
    * - :code:`-bc_clamp [int list]`
      - List of face sets on which to displace by :code:`-bc_clamp_[facenumber]_translate [x,y,z]` or :code:`bc_clamp_[facenumber]_rotate [rx,ry,rz,theta]`
 
-Note: A displacement or rotation vector must be set for each clamped face.
-(One can set only one of :code:`-bc_zero` or :code:`-bc_clamp`, but the result will likely not be interesting.)
+Note: The default for a clamped face is zero displacement.
 
 .. note::
 
@@ -64,7 +60,7 @@ Consider the specific example of the mesh seen below:
 
 With the sidesets defined in the figure, we provide here an example of a minimal set of command line options::
 
-   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998 -bc_clamp_998_translate 0,-0.5,1
+   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_clamp 998,999 -bc_clamp_998_translate 0,-0.5,1
 
 In this example, we set the left boundary, face set :math:`999`, to zero displacement and the right boundary, face set :math:`998`, to displace :math:`0` in the :math:`x` direction, :math:`-0.5` in the :math:`y`, and :math:`1` in the :math:`z`.
 
--- a/examples/solids/README.rst
+++ b/examples/solids/README.rst
@@ -99,6 +99,10 @@ The command line options just shown are the minimum requirements to run the mini
      -  Forcing term option (:code:`none`, :code:`constant`, or :code:`mms`)
      - :code:`none`
 
+   * - :code:`-forcing_vec`
+     -  Forcing vector
+     - :code:`0,-1,0`
+
    * - :code:`-num_steps`
      - Number of load increments for continuation method
      - :code:`1` if :code:`linElas` else :code:`10`
--- a/examples/solids/README.rst
+++ b/examples/solids/README.rst
@@ -44,8 +44,9 @@ The elasticity min-app is controlled via command-line options, the following of 
      - List of faces sets on which to enforce zero displacement
 
    * - :code:`-bc_clamp [int list]`
-     - List of face sets on which to displace by :code:`-bc_clamp_max` in the :math:`y` direction
+     - List of face sets on which to displace by :code:`-bc_clamp_facenumber_translate`
 
+Note: The displacement vector must be set for each clamped face.
 (One can set only one of :code:`-bc_zero` or :code:`-bc_clamp`, but the result will likely not be interesting.)
 
 .. note::
@@ -63,9 +64,11 @@ Consider the specific example of the mesh seen below:
 
 With the sidesets defined in the figure, we provide here an example of a minimal set of command line options::
 
-   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998
+   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp  998 -bc_clamp_998_translate 0,-0.5,1
 
-In this example, we set the left boundary, face set :math:`999`, to zero displacement and the right boundary, face set :math:`998`, to displace by the default value of :math:`-1.0` in the :math:`y` direction.
+In this example, we set the left boundary, face set :math:`999`, to zero displacement and the right boundary, face set :math:`998`, to displace :math:`0` in the :math:`x` direction, :math:`-0.5` in the :math:`y`, and :math:`1` in the :math:`z`.
+
+As an alternative to specifying a mesh with :code:`-mesh`, the user may use a DMPlex box mesh by specifying :code:`-dm_plex_box_faces [int list]`, :code:`-dm_plex_box_upper [real list]`, and :code:`-dm_plex_box_lower [real list]`.
 
 The command line options just shown are the minimum requirements to run the mini-app, but additional options may also be set as follows
 
@@ -95,10 +98,6 @@ The command line options just shown are the minimum requirements to run the mini
    * - :code:`-forcing`
      -  Forcing term option (:code:`none`, :code:`constant`, or :code:`mms`)
      - :code:`none`
-
-   * - :code:`-bc_clamp_max`
-     - Maximum value to displace clamped boundary
-     - :code:`-1.0`
 
    * - :code:`-num_steps`
      - Number of load increments for continuation method
--- a/examples/solids/README.rst
+++ b/examples/solids/README.rst
@@ -41,9 +41,9 @@ The elasticity min-app is controlled via command-line options, the following of 
      - `Poisson's ratio <https://en.wikipedia.org/wiki/Poisson%27s_ratio>`_, :math:`\nu < 0.5`
 
    * - :code:`-bc_clamp [int list]`
-     - List of face sets on which to displace by :code:`-bc_clamp_[facenumber]_translate [x,y,z]` or :code:`bc_clamp_[facenumber]_rotate [rx,ry,rz,theta]`
+     - List of face sets on which to displace by :code:`-bc_clamp_[facenumber]_translate [x,y,z]` and/or :code:`bc_clamp_[facenumber]_rotate [rx,ry,rz,theta]`
 
-Note: The default for a clamped face is zero displacement.
+Note: The default for a clamped face is zero displacement. All displacement is with respect to the initial configuration.
 
 .. note::
 
--- a/examples/solids/README.rst
+++ b/examples/solids/README.rst
@@ -44,9 +44,9 @@ The elasticity min-app is controlled via command-line options, the following of 
      - List of faces sets on which to enforce zero displacement
 
    * - :code:`-bc_clamp [int list]`
-     - List of face sets on which to displace by :code:`-bc_clamp_facenumber_translate`
+     - List of face sets on which to displace by :code:`-bc_clamp_[facenumber]_translate [x,y,z]` or :code:`bc_clamp_[facenumber]_rotate [rx,ry,rz,theta]`
 
-Note: The displacement vector must be set for each clamped face.
+Note: A displacement or rotation vector must be set for each clamped face.
 (One can set only one of :code:`-bc_zero` or :code:`-bc_clamp`, but the result will likely not be interesting.)
 
 .. note::
@@ -64,7 +64,7 @@ Consider the specific example of the mesh seen below:
 
 With the sidesets defined in the figure, we provide here an example of a minimal set of command line options::
 
-   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp  998 -bc_clamp_998_translate 0,-0.5,1
+   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998 -bc_clamp_998_translate 0,-0.5,1
 
 In this example, we set the left boundary, face set :math:`999`, to zero displacement and the right boundary, face set :math:`998`, to displace :math:`0` in the :math:`x` direction, :math:`-0.5` in the :math:`y`, and :math:`1` in the :math:`z`.
 
--- a/examples/solids/elasticity.c
+++ b/examples/solids/elasticity.c
@@ -28,8 +28,8 @@
 // Sample runs:
 //
 //     ./elasticity -mesh [.exo file] -degree 2 -E 1 -nu 0.3 -problem linElas -forcing mms
-//     ./elasticity -mesh [.exo file] -degree 2 -E 1 -nu 0.3 -bc_zero 999 -bc_clamp 998 -problem hyperSS -forcing none -ceed /cpu/self
-//     ./elasticity -mesh [.exo file] -degree 2 -E 1 -nu 0.3 -bc_zero 999 -bc_clamp 998 -problem hyperFS -forcing none -ceed /gpu/occa
+//     ./elasticity -mesh [.exo file] -degree 2 -E 1 -nu 0.3 -bc_clamp 998,999 -bc_clamp_998_translate 0.1,0.2,0.3 -problem hyperSS -forcing none -ceed /cpu/self
+//     ./elasticity -mesh [.exo file] -degree 2 -E 1 -nu 0.3 -bc_clamp 998,999 -bc_clamp_998 rotate 1,0,0,0.2 -problem hyperFS -forcing none -ceed /gpu/occa
 //
 // Sample meshes can be found at https://github.com/jeremylt/ceedSampleMeshes
 //
--- a/examples/solids/elasticity.c
+++ b/examples/solids/elasticity.c
@@ -619,7 +619,7 @@ int main(int argc, char **argv) {
                        "    PC Type                            : %s\n",
                        pcType); CHKERRQ(ierr);
 
-    if (appCtx->multigridChoice != MULTIGRID_NONE && appCtx->degree > 1) {
+    if (!strcmp(pcType, PCMG)) {
       PCMGType pcmgType;
       ierr = PCMGGetType(pc, &pcmgType); CHKERRQ(ierr);
       ierr = PetscPrintf(comm,
--- a/examples/solids/elasticity.c
+++ b/examples/solids/elasticity.c
@@ -420,7 +420,6 @@ int main(int argc, char **argv) {
       ierr = PCJacobiSetType(pc, PC_JACOBI_DIAGONAL); CHKERRQ(ierr);
     } else if (appCtx->degree == 1) {
       // ---- AMG for degree 1
-      ierr = KSPSetType(ksp, KSPPREONLY); CHKERRQ(ierr);
       ierr = PCSetType(pc, PCGAMG); CHKERRQ(ierr);
     } else {
       // ---- PCMG
--- a/examples/solids/elasticity.c
+++ b/examples/solids/elasticity.c
@@ -33,7 +33,7 @@
 //
 // Sample meshes can be found at https://github.com/jeremylt/ceedSampleMeshes
 //
-//TESTARGS -ceed {ceed_resource} -test -degree 2 -nu 0.3 -E 1  -dm_plex_box_faces 3,3,3
+//TESTARGS -ceed {ceed_resource} -test -degree 2 -nu 0.3 -E 1 -dm_plex_box_faces 3,3,3
 
 /// @file
 /// CEED elasticity example using PETSc with DMPlex
@@ -700,6 +700,23 @@ int main(int argc, char **argv) {
     // -- Cleanup
     ierr = VecDestroy(&errorVec); CHKERRQ(ierr);
     ierr = VecDestroy(&trueVec); CHKERRQ(ierr);
+  }
+
+  // ---------------------------------------------------------------------------
+  // Compute energy
+  // ---------------------------------------------------------------------------
+  if (!appCtx->testMode) {
+    // -- Compute L2 error
+    CeedScalar energy;
+    ierr = ComputeStrainEnergy(resCtx,
+                               ceedData[fineLevel]->opEnergy, U,
+                               ceedData[fineLevel]->energy, &energy);
+    CHKERRQ(ierr);
+
+    // -- Output
+    ierr = PetscPrintf(comm,
+                       "    Strain Energy                      : %e\n",
+                       energy); CHKERRQ(ierr);
   }
 
   // ---------------------------------------------------------------------------
--- a/examples/solids/qfunctions/constantForce.h
+++ b/examples/solids/qfunctions/constantForce.h
@@ -49,14 +49,14 @@ CEED_QFUNCTION(SetupConstantForce)(void *ctx, const CeedInt Q,
   CeedPragmaSIMD
   for (CeedInt i=0; i<Q; i++) {
     // Setup
-    CeedScalar rho = qdata[i];
+    CeedScalar wdetJ = qdata[i];
 
     // Forcing function
     // -- Component 1
     force[i+0*Q] = 0;
 
     // -- Component 2
-    force[i+1*Q] = -rho;
+    force[i+1*Q] = -wdetJ;
 
     // -- Component 3
     force[i+2*Q] = 0;
--- a/examples/solids/qfunctions/constantForce.h
+++ b/examples/solids/qfunctions/constantForce.h
@@ -34,7 +34,7 @@ struct Physics_private {
 #endif
 
 // -----------------------------------------------------------------------------
-// Constant forcing term, -1 in y direction only
+// Constant forcing term along specified vector
 // -----------------------------------------------------------------------------
 CEED_QFUNCTION(SetupConstantForce)(void *ctx, const CeedInt Q,
                                    const CeedScalar *const *in,
@@ -45,6 +45,9 @@ CEED_QFUNCTION(SetupConstantForce)(void *ctx, const CeedInt Q,
   // Outputs
   CeedScalar *force = out[0];
 
+  // Context
+  const CeedScalar *forcingVector = (CeedScalar(*))ctx;
+
   // Quadrature Point Loop
   CeedPragmaSIMD
   for (CeedInt i=0; i<Q; i++) {
@@ -53,13 +56,13 @@ CEED_QFUNCTION(SetupConstantForce)(void *ctx, const CeedInt Q,
 
     // Forcing function
     // -- Component 1
-    force[i+0*Q] = 0;
+    force[i+0*Q] = forcingVector[0]*wdetJ;
 
     // -- Component 2
-    force[i+1*Q] = -wdetJ;
+    force[i+1*Q] = forcingVector[1]*wdetJ;
 
     // -- Component 3
-    force[i+2*Q] = 0;
+    force[i+2*Q] = forcingVector[2]*wdetJ;
 
   } // End of Quadrature Point Loop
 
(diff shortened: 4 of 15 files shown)


examples/solids/src/cloptions.c
```diff
@@ -91,6 +109,37 @@ PetscErrorCode ProcessCommandLineOptions(MPI_Comm comm, AppCtx appCtx) {
                             NULL, appCtx->bcClampMax, &appCtx->bcClampMax,
                             NULL); CHKERRQ(ierr);
 
+  appCtx->bcTwistXCount = 16;
+  ierr = PetscOptionsIntArray("-bc_twist_x",
+                              "Face IDs to apply twisted Dirichlet BC",
+                              NULL, appCtx->bcTwistXFaces,
+                              &appCtx->bcTwistXCount, NULL); CHKERRQ(ierr);
+  appCtx->bcTwistXMax = 1;
```
jedbrown, line 117: So long as faces are numbered, you could process `-bc_twist_998_rotate .3,.6,.9` where the arguments are a (Rodrigues vector](https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula. This is a relatively compact and intuitive method. You could scale the vector by pi to make it easier to specify even fractions.

jeremylt:
@jedbrown @valeriabarra if I'm understanding this correctly, in `DMAddBoundary()`, using `"Face Sets"` for a box domaindoes not correspond to only the exterior faces. It looks like in `navierstokes.c` we are setting all faces of specified exterior elements to the specified BC. Is that the intended behavior for box meshes?

To see what I'm seeing, run
```
./elasticity -cells 10,10,10 -nu 0.3 -E 1 -bc_zero 5 -bc_twist_x 6 -degree 2 -outer_ksp_monitor -view_soln -lx 5
```
I'm printing out the value of x, which should be `-lx/2 = 2.5`, but we can see that the BC is being called for interior faces as well. Am I missing something about how the faces are labeled on a box?

knepley:
"Face Sets" should only mark exterior faces. Can you give me BoxMesh inputs to run with where Face Sets is wrong?

jedbrown:
Labels look fine at mesh creation.
```
DM Object: 1 MPI processes                 
  type: plex                                                                                   
DM_0x560cda6560c0_0 in 3 dimensions:         
  0-cells: 1331                            
  1-cells: 3630                                                                                
  2-cells: 3300                                                                                
  3-cells: 1000
Labels:
  marker: 1 strata with value/size (1 (1800))
  Face Sets: 6 strata with value/size (6 (100), 5 (100), 3 (100), 4 (100), 1 (100), 2 (100))
  depth: 4 strata with value/size (0 (1331), 1 (3630), 2 (3300), 3 (1000))
  celltype: 4 strata with value/size (7 (1000), 0 (1331), 4 (3300), 1 (3630))
```

examples/solids/src/cloptions.c
```diff
@@ -59,6 +59,24 @@ PetscErrorCode ProcessCommandLineOptions(MPI_Comm comm, AppCtx appCtx) {
...
+    ierr = PetscOptionsScalar("-lx", "Length scale in x direction",
+                              NULL, appCtx->lx, &appCtx->lx, NULL);
+    CHKERRQ(ierr);
+    ierr = PetscOptionsScalar("-ly", "Length scale in y direction",
+                              NULL, appCtx->ly, &appCtx->ly, NULL);
+    CHKERRQ(ierr);
+    ierr = PetscOptionsScalar("-lz", "Length scale in z direction",
+                              NULL, appCtx->lz, &appCtx->lz, NULL);
+    CHKERRQ(ierr);
+  }
```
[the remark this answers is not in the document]
jeremylt: Right, thanks for reminding me

jedbrown:
@knepley I think your label completion is sucking in more than the points in the closure of the faces in the face set. This with PETSc master.

Screenshot_20200403_220803

examples/solids/src/boundary.c
```diff
@@ -86,15 +86,11 @@ PetscErrorCode BCTwistX(PetscInt dim, PetscReal loadIncrement,
...
   u[0] = 0;
-  u[1] = (y*cos(theta) - z*sin(theta))*loadIncrement;
-  u[2] = (y*sin(theta) + z*cos(theta))*loadIncrement;
-
-// Testing
-PetscScalar x = coords[0];
-printf("%f\n", x);
-
+  u[1] = (c - 1.)*y - s*z;
+  u[2] = s*y + (c - 1.)*z;
```
jedbrown, line 93: @jeremylt Recall that we're setting the displacement of a rotation `(Q - I) x`, not the coordinates of the rotated point `Q x`.

I also changed the load increment to apply the twisting, otherwise you'd see no difference between zero and 2pi rotation, for example.

examples/solids/src/boundary.c
```diff
@@ -107,12 +103,11 @@ PetscErrorCode BCTwistY(PetscInt dim, PetscReal loadIncrement,
...
-
-  PetscScalar theta = twistMax*M_PI;
-  u[0] = (x*cos(theta) - z*sin(theta))*loadIncrement;
+  PetscScalar theta = twistMax*M_PI*loadIncrement;
+  PetscReal c = cos(theta), s = sin(theta);
+  u[0] = (c - 1.)*x + s*z;
   u[1] = 0;
-  u[2] = (x*sin(theta) + z*cos(theta))*loadIncrement;
-
+  u[2] = -s*x + (c - 1.)*z;
```
jedbrown, line 110: Note that signs are swapped here because (y,z,x) is the rotation that satisfies the right-hand rule.

knepley:
We add cells, but we use a different label. Hmm, maybe that is colliding with one of yours. Its in DMLabelAddCells()

jedbrown:
@knepley Never mind. It was just BS from doing a large-deformation problem with linear elasticity. Here's the proper solution (after load increments and Newton).

```
mpiexec -n 2 ./elasticity -^Coblem hyperfs -dm_plex_box_faces 10,10,10 -dm_plex_box_lower -1,-1,-1 -dm_plex_box_upper 1,1,1 -nu 0.3 -E 1 -bc_zero 5 -bc_twist_x 6 -bc_twist_max_x .2 -degree 2 -outer_ksp_monitor -snes_monitor -view_soln -dm_view
```
Screenshot_20200403_222836

examples/solids/src/boundary.c
jeremylt: Right. I had something closer ( without the (c - 1) ) but I horribly mangled it while trying to figure out what was going on.

jedbrown:
@jeremylt I think we need to check the Jacobian for these large deformation cases.
```console
mpiexec -n 2 ./elasticity -problem hyperfs -dm_plex_box_faces 16,4,4 -dm_plex_box_lower -4,-1,-1 -dm_plex_box_upper 4,1,1 -nu 0.3 -E 1 -bc_zero 5 -bc_twist_x 6 -bc_twist_max_x 1 -degree 2 -outer_ksp_converged_reason -snes_monitor -view_soln -dm_view -num_steps 30
```
Notice how Newton stops converging quadratically and the solves fail at the end.

twisting

jeremylt:
That does look better, but if I add
```
  if (fabs(x +1.0) > 1e-10)
    printf("%f %f %f\n",x,y,z);
```
I still see the BC function called for coordinates such as `-0.800000 -0.800000 -0.800000` which is what had me concerned in the first place.

jedbrown:
I don't have an explanation. I assume the same occurs for translation. I'd use a debugger to find out where it was called. You can also check the result after `DMPlexLabelComplete`.

jeremylt:
It does not appear to be in the vtu. Perhaps its some sort of overcomputation?

I can check with the cylinder for similar behavior. Those faces are set differently.

examples/solids/qfunctions/hyperSS.h
```diff
@@ -319,5 +319,132 @@ CEED_QFUNCTION(HyperSSdF)(void *ctx, CeedInt Q, const CeedScalar *const *in,
...
+    // mu = E / (2 * (1 + nu))
+    // bulk modulus = E / (2 * (1 - 2 * nu))
+    // lambda = (3 * bulk modulus - 2 * mu) / 3
+    // e_v = volumetric strain = e00 + e11 + e22
+    //
+    // sigma = lambda * log(1 + e_v) + 2 * mu * epsilon
+    //
+    // Above Voigt Notation is placed in a 3x3 matrix:
+    // Volumetric strain
+    const CeedScalar strain_vol = e[0][0] + e[1][1] + e[2][2];
```
valeriabarra, line 421: Thanks Jeremy. Do we want to add the calculation of the volumetric strain to the other material models as well? I think it could be a useful diagnostic variable.

jeremylt:
For a small twist, hyperFS diverges while hyperSS converges. I agree that there has to be an issue in the Jacobian.

jedbrown:
twist2
Thanks for fixing the Jacobian. I also added support for `-snes_mf_operator`, which allowed me to see the same convergence before you had fixed it.
As for computing strain energy, we want to implement this formula, not just compute incremental strain energy.
https://libceed.readthedocs.io/en/latest/examples/solids/#equation-neo-hookean-energy

jeremylt:
I think I fixed it. 
```
mpiexec -n 2 ./elasticity -problem hyperfs -dm_plex_box_faces 16,4,4 -dm_plex_box_lower -4,-1,-1 -dm_plex_box_upper 4,1,1 -nu 0.3 -E 1 -bc_zero 5 -bc_twist_x 6 -bc_twist_max_x 1 -degree 2 -snes_monitor -view_soln -num_steps 30
```
now does not show the solver failure above. I also put in a check to abort the load increments loop once a SNES solve diverges.

jedbrown:
Yep, I did a fill revolution in 10 steps, which seems quite reasonable.
```console
./elasticity -problem hyperfs -dm_plex_box_faces 16,4,4 -dm_plex_box_lower -4,-1,-1 -dm_plex_box_upper 4,1,1 -nu 0.3 -E 1 -bc_zero 5 -bc_twist_x 6 -bc_twist_max_x 2 -degree 2 -outer_ksp_converged_reason -snes_monitor -view_soln -dm_view -num_steps 10
```

examples/solids/src/setuplibceed.c
```diff
@@ -37,35 +37,35 @@ problemData problemOptions[3] = {
...
     .setupgeo = SetupGeo,
     .apply = HyperSSF,
     .jacob = HyperSSdF,
-    .energy = HyperSSEnergy,
+    .energy = StrainEnergy,
     .setupgeofname = SetupGeo_loc,
     .applyfname = HyperSSF_loc,
     .jacobfname = HyperSSdF_loc,
-    .energyfname = HyperSSEnergy_loc,
+    .energyfname = StrainEnergy_loc,
```
jedbrown: Note that these aren't the correct strain energies for the linear and small-strain models. Ideally we would report the strain energy for the model actually being used. It's just a deficiency of our notes that we only explicitly wrote down the energy for the Neo-Hookean model.

Although not critical for our scaling studies, our future trajectory will have alternatives to Neo-Hookean.
Computation of strain energy shouldn't be performance sensitive so it's okay to factor however is most convenient.

The repetition in certain `static inline` ingredients is a reason why it'd be useful for qfunction source files to be allowed to include headers. We can discuss that after #305 merges.

jeremylt: Agreed on all fronts. I am being very messy in my rush to get these convergence studies done. I anticipate rebuilding this branch after I distribute my proposal.

jedbrown: I'd like to get the bug fix into master sooner rather than later. If you report at "Neo-Hookean Strain Energy" then I'd be fine with merging it and we can make it match the model later.

jeremylt: I made a clean PR with just the fixes.

examples/solids/src/cloptions.c
jeremylt: The translation option was very quick to code up.

For rotation, do we want the users to provide `R = I + (sin theta) K + (1 - cos theta) K^2` (what I understand the Rodriques vector to be) or would it be more intuitive for them to specify an axis and angle?

jedbrown: Ah, it might not be clear in the article, but the usual approach is that you specify a single (non-unit) vector `k`. The orientation of `k` defines the axis and its length is `\theta` (how much to rotate around that axis). We could choose to scale by pi (or even pi/180) to make normalization more convenient for axis-aligned rotation. When off-axis, you generally need to do some math separately, but I suppose the command line could accept `kx,ky,kz,theta` which would mean to normalize the specified vector `k = (kx, ky, kz)` and scale it by `theta`. If using degrees, that would allow `-bc_clamp_998_rotate 3,4,5,45` (rotate by 45 degrees about the specified axis).

jeremylt: I think `kx,ky,kz,theta` would make it easier for users to tinker or adjust bc parameter values, at least for me.

jedbrown: Sounds good to me. You could allow `kx,ky,kz` (theta = norm(k)) as short-hand, or skip it for uniformity.

jeremylt:
Ok, I have cleaned up the messiness of this branch. I added
1) Strain energy for HyperFS
2) Improved clamp BCs with translation vector
3) Rotation clamp BCs
ToDo:
1) Strain energy for HyperSS and LinElas

jeremylt:
@jedbrown, thank you for the ideas to clean up and generalize the Clamp BCs. I think this will be ready to merge once I add the other strain energy formulations. @ArashMehraban, do you want to help with adding the other two strain energy formulations?

jeremylt:
The linear case really is behaving poorly.
```
[jeremy@pheonix solids]$ ./elasticity -nu 0.3 -E 1 -dm_plex_box_faces 10,10,10 -boundary mms -forcing mms -degree 1 -snes_monitor

-- Elastisticy Example - libCEED + PETSc --
  libCEED:
  Problem:
    Problem Name                       : Linear elasticity
    Forcing Function                   : Manufactured solution
  Mesh:
    File                               : Box Mesh
    Number of 1D Basis Nodes (p)       : 2
    Number of 1D Quadrature Points (q) : 2
    Global nodes                       : 729
    Owned nodes                        : 729
    DoF per node                       : 3
  Multigrid:
    Type                               : Algebraic multigrid
    Number of Levels                   : 0
    Level 0 (coarse):
      Number of 1D Basis Nodes (p)     : 2
      Global Nodes                     : 729
      Owned Nodes                      : 729
    Level 0 (fine):
      Number of 1D Basis Nodes (p)     : 2
      Global Nodes                     : 729
      Owned Nodes                      : 729
0 Load Increment
  0 SNES Function norm 3.003170393787e-07 
  1 SNES Function norm 5.047317986543e-08 
  2 SNES Function norm 1.177576804263e-08 
  3 SNES Function norm 2.933626247215e-09 
  4 SNES Function norm 7.524634106083e-10 
  5 SNES Function norm 1.962790888253e-10 
  6 SNES Function norm 5.177835159194e-11 
  7 SNES Function norm 1.377188580773e-11 
  8 SNES Function norm 3.686376998419e-12 
  9 SNES Function norm 9.917956484007e-13 
 10 SNES Function norm 2.679606362733e-13 
 11 SNES Function norm 7.265330268473e-14 
 12 SNES Function norm 1.975838917966e-14 
 13 SNES Function norm 5.387449427811e-15 
 14 SNES Function norm 1.472341946943e-15 
  SNES:
    SNES Type                          : newtonls
    SNES Convergence                   : CONVERGED_FNORM_RELATIVE
    Number of Load Increments          : 1
    Completed Load Increments          : 1
    Total SNES Iterations              : 14
    Final rnorm                        : 1.472342e-15
  Linear Solver:
    KSP Type                           : preonly
    PC Type                            : gamg
  Performance:
    SNES Solve Time                    : 8.45367 (8.45367) sec
    L2 Error                           : 8.073619e-03
    Strain Energy                      : 1.556976e-13
```
```
[jeremy@pheonix solids]$ ./elasticity -nu 0.3 -E 1 -dm_plex_box_faces 10,10,10 -boundary mms -forcing mms -degree 2 -snes_monitor

-- Elastisticy Example - libCEED + PETSc --
  libCEED:
  Problem:
    Problem Name                       : Linear elasticity
    Forcing Function                   : Manufactured solution
  Mesh:
    File                               : Box Mesh
    Number of 1D Quadrature Points (q) : 3
    Global nodes                       : 6859
    Owned nodes                        : 6859
    DoF per node                       : 3
  Multigrid:
    Type                               : P-multigrid, logarithmic coarsening
    Number of Levels                   : 2
    Level 0 (coarse):
      Number of 1D Basis Nodes (p)     : 2
      Global Nodes                     : 729
      Owned Nodes                      : 729
    Level 1 (fine):
      Number of 1D Basis Nodes (p)     : 3
      Global Nodes                     : 6859
      Owned Nodes                      : 6859
0 Load Increment
  0 SNES Function norm 5.886759069794e-07 
  1 SNES Function norm 1.518331439692e-17 
  SNES:
    SNES Type                          : newtonls
    SNES Convergence                   : CONVERGED_FNORM_RELATIVE
    Number of Load Increments          : 1
    Completed Load Increments          : 1
    Total SNES Iterations              : 1
    Final rnorm                        : 1.518331e-17
  Linear Solver:
    PC Type                            : mg
  P-Multigrid:
    PCMG Type                          : MULTIPLICATIVE
    PCMG Cycle Type                    : v
    Coarse Solve:
      KSP Type                         : preonly
      PC Type                          : gamg
  Performance:
    SNES Solve Time                    : 0.349257 (0.349257) sec
    L2 Error                           : 1.131716e-04
    Strain Energy                      : 1.592684e-13
```

jedbrown:
Fix the KSP type for linear elements.

jeremylt:
Duh, I should have caught that.

examples/solids/elasticity.c
```diff
@@ -473,7 +473,7 @@ int main(int argc, char **argv) {
 
       // -------- Coarse KSP
       ierr = PCMGGetCoarseSolve(pc, &kspCoarse); CHKERRQ(ierr);
-      ierr = KSPSetType(kspCoarse, KSPPREONLY); CHKERRQ(ierr);
+      ierr = KSPSetType(kspCoarse, KSPCG); CHKERRQ(ierr);
```
jedbrown, line 476: Curious to see perf, but we probably don't want this.

jeremylt: It significantly drops the number of v-cycles in a SNES iteration in my initial tests. 16-14 to 9-8

jedbrown: Yeah, at the expense of lots more global synchronization (so it'll scale much worse). The goal here will be to balance work happening on the coarse grid with that occurring in levels. It could be that richardson(2) on the GAMG part is a good balance. We can also make GAMG smoothers stronger. Alternatively, we could make our p-MG smoothers weaker so they don't over-solve.

jeremylt: True. I dropped it for now so we can merge this. I can come back to this sort of tuning on the next PR.

jedbrown: Cool. These knobs are also somewhat problem specific. We can write a tuning guide at some point, to teach users (and remind us) how to balance work through the MG cycle so don't waste flops on the fine grid and don't waste synchronization on the coarse grid.

examples/solids/src/setuplibceed.c
```diff
@@ -37,29 +37,35 @@ problemData problemOptions[3] = {
...
   [ELAS_HYPER_SS] = {
     .qdatasize = 10,
     .setupgeo = SetupGeo,
     .apply = HyperSSF,
     .jacob = HyperSSdF,
+    .energy = LinElasEnergy,
     .setupgeofname = SetupGeo_loc,
     .applyfname = HyperSSF_loc,
     .jacobfname = HyperSSdF_loc,
+    .energyfname = LinElasEnergy_loc,
```
jedbrown, line 56: These should use HyperSSEnergy and we should write the strain energy functions into the documentation as well. @ArashMehraban This would be a great contribution from you.

examples/solids/qfunctions/hyperSS.h
```diff
@@ -319,5 +319,95 @@ CEED_QFUNCTION(HyperSSdF)(void *ctx, CeedInt Q, const CeedScalar *const *in,
...
+                                 (gradu[2][1] + gradu[1][2])/2.,
+                                 (gradu[2][2] + gradu[2][2])/2.}
+                               };
+    // *INDENT-ON*
+
+    // Strain energy W = 
+    energy[i] = ((lambda/2.+mu)*(e[0][0]*e[0][0] + e[1][1]*e[1][1] + e[2][2]) +
+                 2*mu*(e[0][1]*e[0][1] + e[0][2]*e[0][2] + e[1][2]*e[1][2]))*wJ;
+
+  } // End of Quadrature Point Loop
```
jedbrown: This is missing the log term at least.

jeremylt: That was a formula mess on my part. I think I repaired it.

examples/solids/src/setuplibceed.c
ArashMehraban: @jedbrown Just sent it to Jeremy to upload. Look forward to comments

valeriabarra: I'm a bit confused on why not push it yourself? We can help you and revise it, but you will keep authorship of your contribution

examples/solids/qfunctions/constantForce.h
```diff
@@ -53,13 +56,13 @@ CEED_QFUNCTION(SetupConstantForce)(void *ctx, const CeedInt Q,
...
-    force[i+0*Q] = 0;
+    force[i+0*Q] = forcingVector[0]*wdetJ;
 
     // -- Component 2
-    force[i+1*Q] = -wdetJ;
+    force[i+1*Q] = forcingVector[1]*wdetJ;
 
     // -- Component 3
-    force[i+2*Q] = 0;
+    force[i+2*Q] = forcingVector[2]*wdetJ;
```
jedbrown: We should make sure this is only used for linear/small-strain. For finite strain, body forces (like gravity) act in the current configuration.

jeremylt: Do you mean that `[0, -1, 0]` should be the only valid forcing term for the finite strain case?

jedbrown: No, it wasn't valid either. The gravity vector needs to be pulled back to the reference frame.

jeremylt: That makes sense. I've added an error message.

jeremylt:
@ArashMehraban, feel free to commit directly to this branch with @jedbrown's requested changes.

jeremylt:
convergeHRefine

I used a 2x finer mesh for the true case and the p=3 p=4 cases are still off. There is either an issue in my calculation of strain energy or in the hyperFS qfunction.

examples/solids/src/cloptions.c
```diff
@@ -82,12 +97,46 @@ PetscErrorCode ProcessCommandLineOptions(MPI_Comm comm, AppCtx appCtx) {
...
+      PetscScalar norm = sqrt(appCtx->bcClampMax[i][0]*appCtx->bcClampMax[i][0]+
+                              appCtx->bcClampMax[i][1]*appCtx->bcClampMax[i][1]+
+                              appCtx->bcClampMax[i][2]*appCtx->bcClampMax[i][2]);
+      for (PetscInt j = 0; j < 3; j++)
+        appCtx->bcClampMax[i][j] /= norm;
+
+      // Error if neither were set
+      if (!bcFlag) {
+        SETERRQ(PETSC_COMM_SELF, PETSC_ERR_SUP,
+                "-bc_clamp additional option needed");
```
jedbrown, line 136: How about `-bc_clamp` defaults to zero translation and rotation. That way we just have one option (get rid of `-bc_zero`).

We can (in this PR or another) add `-bc_clamp_%d_slip 1,0,0` to specify a vector normal to the plane on which we allow free slip (no friction or contact). This would most physically be applied in the current configuration, but that creates a nonlinearity in the boundary condition and defining it in the reference configuration is sufficient in this case (due to the nature of a clamped planar boundary).

examples/solids/src/cloptions.c
```diff
@@ -82,12 +97,46 @@ PetscErrorCode ProcessCommandLineOptions(MPI_Comm comm, AppCtx appCtx) {
...
+
+    appCtx->bcClampTranslate[i] = PETSC_FALSE;
+    ierr = PetscOptionsScalarArray(optionName,
+                                   "Vector to translate clamped end by", NULL,
+                                   appCtx->bcClampMax[i], &maxn,
+                                   &appCtx->bcClampTranslate[i]);
+    CHKERRQ(ierr);
+
+    // Rotation vector
+    if (!appCtx->bcClampTranslate[i]) {
```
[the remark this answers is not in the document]
jedbrown: In particular, I'd rather have this options processing set up a general affine transformation (rotation matrix and translation, e.g., stored as a 3x4 matrix). The boundary condition function would unconditionally apply that general affine transformation (scaled by `loadIncrement`).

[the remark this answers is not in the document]
jeremylt: Agreed

examples/solids/src/cloptions.c
jeremylt: Also agreed. It might make sense to get these changes into master and then add no slip?

jedbrown: Sure, can you clean up the affine matrix formulation in this PR? Happy to merge.

examples/solids/src/cloptions.c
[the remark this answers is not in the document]
jeremylt: So, if we keep only a transformation matrix, then we lose 2n*pi as a rotation. I left it as building the incremental transformation each time, but we can change it if we're ok losing this one case.

jeremylt:
To see the new BCs, use
```
./elasticity -problem hyperFS -dm_plex_box_faces 16,4,4 -dm_plex_box_lower -4,-1,-1 -dm_plex_box_upper 4,1,1 -nu 0.3 -E 1 -bc_clamp 5,6 -bc_clamp_6_rotate 1,0,0,1 -bc_clamp_6_translate 0.1,-0.1,0 -degree 2 -snes_monitor -num_steps 30 -view_soln
```

doc/sphinx/source/conf.py
```diff
@@ -39,7 +39,7 @@
 # ones.
 extensions = [
     'breathe',
-    #'hoverxref.extension',  # still in beta; not rendering well on RTD
+    # 'hoverxref.extension',  # still in beta; not rendering well on RTD
```
jeremylt, line 42: I'm not sure why my `make style` wanted this.

examples/solids/src/cloptions.c
```diff
@@ -25,11 +25,8 @@
 // Process general command line options
 PetscErrorCode ProcessCommandLineOptions(MPI_Comm comm, AppCtx appCtx) {
   PetscErrorCode ierr;
-  PetscBool degreeFalg   = PETSC_FALSE;
+  PetscBool degreeFlag   = PETSC_FALSE;
```
jeremylt, line 28: Do we want to just set this to automatically use `degree = 2` like we do with some of our other examples?

doc/sphinx/source/conf.py
valeriabarra: Maybe it's the comment style that requires having at least a space after the comment symbol. I wouldn't worry too much about it now, since it most likely change once we finalize the use of the hoverxf extension.

examples/solids/src/cloptions.c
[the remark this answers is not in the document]
jedbrown: Yep, it's the generator that we should store, and that looks like what you've done. Great.

examples/solids/README.rst
```diff
@@ -64,7 +60,7 @@ Consider the specific example of the mesh seen below:
 
 With the sidesets defined in the figure, we provide here an example of a minimal set of command line options::
 
-   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_zero 999 -bc_clamp 998 -bc_clamp_998_translate 0,-0.5,1
+   ./elasticity -mesh [.exo file] -degree 4 -E 1e6 -nu 0.3 -bc_clamp 998,999 -bc_clamp_998_translate 0,-0.5,1
```
valeriabarra, line 63: Thanks. Perhaps you want to update the following line, too?

jeremylt: I'm not sure if its clear, but here I mean to be reinforcing that the option `-bc_clamp` without `-bc_clamp_[facenum]_[translate/rotate]`  sets as default displacement of zero displacement on that face.

jedbrown: You can remove the reference to `-bc_zero` in `elasticity.c`.

jeremylt: Got it

valeriabarra: Yes the part for which `-bc_clamp` without `[facenum]_[translate/rotate]` sets the default zero displacement condition on the face specified was clear. What was not entirely clear to me is if `-bc_clamp 998,999`  specifies this condition on both facesets labeled as 998 and 999. And then in the following sentence (on which I could not comment in the review, because was not modified) it reads "In this example, we set the left boundary, face set :math:`999`, to zero displacement"

jeremylt: `-bc_clamp_[facenum]_[translate/rotate]` always needs to be preceded by `-bc_clamp [int list that contains facenum]` so that I can build the strings for the command line options corresponding to the translation and/or rotation vectors on that face.

jedbrown:
We might want to be explicit in the docs that the translation is with respect to the initial configuration.

examples/solids/README.rst
jeremylt: Correct.

valeriabarra: Ok, if one option must necessarily precede the other, it is something that we should make sure it's specified somewhere either in the documentation or as a comment right after the sample runs in the `elasticity.c` file. Otherwise a user may accidentally use them in arbitrary order. 

The following line here could become
```suggestion
   In this example, we set the boundary conditions for the displacement field on the face sets denoted by math:`:998,999`. We set the right boundary, face set :math:`998`, to translate :math:`0` in the :math:`x` direction, :math:`-0.5` in the :math:`y`, and :math:`1` in the :math:`z`, and the left boundary, face set :math:`999`, with unspecified condition, is defaulted to to zero displacement.
```

As I said on Zulip, to me personally it would sound clearer if these were expressed as `-bc_displacement 998,999` or `-bc_disp 998,999` for short, and then `-bc_disp_998_translate 0,-0.5,1`. But it's totally up to you of course.

valeriabarra: Sorry my email notifications are lagged. I just received the one about the merge just now.

PR #505, opened 2020-04-05
Solids fixes

jeremylt:
I pulled out the fixes that are clean.

### DIFF
--- a/examples/solids/elasticity.c
+++ b/examples/solids/elasticity.c
@@ -483,6 +483,13 @@ int main(int argc, char **argv) {
     ierr = KSPSetFromOptions(ksp);
     ierr = PCSetFromOptions(pc);
   }
+  { // Default to critical-point (CP) line search (related to Wolfe's curvature condition)
+    SNESLineSearch linesearch;
+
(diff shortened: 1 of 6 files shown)


examples/solids/elasticity.c
```diff
@@ -472,6 +472,13 @@ int main(int argc, char **argv) {
     ierr = KSPSetFromOptions(ksp);
     ierr = PCSetFromOptions(pc);
   }
+  { // Default to critical-point (CP) line search (related to Wolfe's curvature condition)
+    SNESLineSearch linesearch;
+
+    ierr = SNESGetLineSearch(snes, &linesearch); CHKERRQ(ierr);
+    ierr = SNESLineSearchSetType(linesearch, SNESLINESEARCHCP); CHKERRQ(ierr);
+  }
```
jedbrown, line 480: This will need https://gitlab.com/petsc/petsc/-/merge_requests/2692 to have effect, but I think it's the right thing to have for this model, unless we make energy required, in which case we can SNESSetObjective to support "optimization" line searches (which strictly satisfy both Wolfe conditions). CP is good, cheap, and doesn't need that extra information. It's effectively one extra residual evaluation per Newton, but allows much larger load increments.

PR #508, opened 2020-04-07
Travis - only run lcov for linux + gcc + amd64

jeremylt:
As discussed on Slack, multiple codecov commits create in non-deterministic results. (It's also just a waste of time)
Also see discussion here: https://github.com/jiahao/LowRankModels.jl/pull/2#issuecomment-605697528

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,8 @@
 language: c
 
-arch: amd64
-
-os:
-  - linux
-  - osx
-
 dist: bionic
 
 osx_image: xcode11.2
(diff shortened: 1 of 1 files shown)


jeremylt:
I tossed in naming the jobs and only running `make style` and `make tidy` for one job as well.

jeremylt:
I left a note to future us in the comments, but sometimes setting an option in the build matrix via `include:` sets other options to the default value, even if they set to something else at a higher context in the yml.

PR #509, opened 2020-04-08
Pin breathe==4.14.2

jedbrown:
Workaround https://github.com/michaeljones/breathe/issues/495

Reported-by: Jeremy L. Thompson <jeremy.thompson@colorado.edu>

PR #510, opened 2020-04-08
doc: upgrade to sphinx-3.0

jedbrown:
There is a newer version of Sphinx; this is a better option than #509 
Preview: https://libceed.readthedocs.io/en/jed-sphinx-3.0/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,6 +14,6 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.7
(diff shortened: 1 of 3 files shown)


jeremylt:
In `make distclean` we have

```
distclean : clean
	$(RM) -r doc/html $(CONFIG)
```

Do we want instead

```
distclean : clean
	$(RM) -r doc/html doc/sphinx/build $(CONFIG)
```

jedbrown:
@jeremylt Thanks; pushed

PR #511, opened 2020-04-09
Add support for level of refinements for viz

valeriabarra:
Added support for multiple level of viz refinement in the fluids miniapp. Now `-viz_refine` is not  a PetscBool anymore, but a PetscInt.

### DIFF
--- a/examples/fluids/navierstokes.c
+++ b/examples/fluids/navierstokes.c
@@ -28,10 +28,10 @@
 //
 // Sample runs:
 //
-//     ./navierstokes -ceed /cpu/self -problem density_current -petscspace_degree 1
-//     ./navierstokes -ceed /gpu/occa -problem advection -petscspace_degree 1
+//     ./navierstokes -ceed /cpu/self -problem density_current -degree 1
+//     ./navierstokes -ceed /gpu/occa -problem advection -degree 1
 //
-//TESTARGS -ceed {ceed_resource} -test -petscspace_degree 1
+//TESTARGS -ceed {ceed_resource} -test -degree 1
 
 /// @file
 /// Navier-Stokes example using PETSc
@@ -532,8 +532,8 @@ static PetscErrorCode ComputeLumpedMassMatrix(Ceed ceed, DM dm,
   PetscFunctionReturn(0);
 }
 
-PetscErrorCode SetUpDM(DM dm, problemData *problem, const char *prefix,
-                       SimpleBC bc, void *ctxSetup, PetscInt *degree) {
+PetscErrorCode SetUpDM(DM dm, problemData *problem, PetscInt degree,
+                       SimpleBC bc, void *ctxSetup) {
   PetscErrorCode ierr;
 
   PetscFunctionBeginUser;
@@ -542,9 +542,9 @@ PetscErrorCode SetUpDM(DM dm, problemData *problem, const char *prefix,
     PetscFE fe;
     PetscSpace fespace;
     PetscInt ncompq = 5;
-    ierr = PetscFECreateDefault(PETSC_COMM_SELF,problem->dim, ncompq,
-                                PETSC_FALSE, prefix, PETSC_DETERMINE, &fe);
-    CHKERRQ(ierr);
+    ierr = PetscFECreateLagrange(PETSC_COMM_SELF, problem->dim, ncompq,
+                                 PETSC_FALSE, degree, PETSC_DECIDE,
+                                 &fe);
     ierr = PetscObjectSetName((PetscObject)fe, "Q"); CHKERRQ(ierr);
     ierr = DMAddField(dm,NULL,(PetscObject)fe); CHKERRQ(ierr);
     ierr = DMCreateDS(dm); CHKERRQ(ierr);
@@ -572,12 +572,6 @@ PetscErrorCode SetUpDM(DM dm, problemData *problem, const char *prefix,
     ierr = DMPlexSetClosurePermutationTensor(dm,PETSC_DETERMINE,NULL);
     CHKERRQ(ierr);
     ierr = PetscFEGetBasisSpace(fe, &fespace); CHKERRQ(ierr);
-    if (degree) {
-      ierr = PetscSpaceGetDegree(fespace, degree, NULL); CHKERRQ(ierr);
-      if (*degree < 1) SETERRQ1(PetscObjectComm((PetscObject)dm),
-                                  PETSC_ERR_ARG_OUTOFRANGE,
-                                  "Degree %D; must specify -petscspace_degree 1 (or greater)", *degree);
-    }
     ierr = PetscFEDestroy(&fe); CHKERRQ(ierr);
   }
   {
@@ -662,7 +656,7 @@ int main(int argc, char **argv) {
   PetscScalar resz      = 1000.;    // m (resolution in z)
   PetscInt outputfreq   = 10;       // -
   PetscInt contsteps    = 0;        // -
-  PetscInt degree;
+  PetscInt degree       = 1;        // -
   PetscInt qextra       = 2;        // -
   DMBoundaryType periodicity[] = {DM_BOUNDARY_NONE, DM_BOUNDARY_NONE,
                                   DM_BOUNDARY_NONE
@@ -801,6 +795,8 @@ int main(int argc, char **argv) {
                          NULL, outputfreq, &outputfreq, NULL); CHKERRQ(ierr);
   ierr = PetscOptionsInt("-continue", "Continue from previous solution",
                          NULL, contsteps, &contsteps, NULL); CHKERRQ(ierr);
+  ierr = PetscOptionsInt("-degree", "Polynomial degree of finite elements",
+                         NULL, degree, &degree, NULL); CHKERRQ(ierr);
   ierr = PetscOptionsInt("-qextra", "Number of extra quadrature points",
                          NULL, qextra, &qextra, NULL); CHKERRQ(ierr);
   PetscStrncpy(user->outputfolder, ".", 2);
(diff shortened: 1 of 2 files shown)


valeriabarra:
@jedbrown I added this support. It works fine for `-viz_refine 1` and `-viz_refine 2`, but it crashes for higher values. The problem is in the `TSMonitor_NS()` when calling `MatInterpolate()` between Q and Qrefined. I don't fully understand why though, only for -viz_refine > 2.

```
[0]PETSC ERROR: Nonconforming object sizes
[0]PETSC ERROR: Mat mat,Vec x: global dim 1447 213
[0]PETSC ERROR: See https://www.mcs.anl.gov/petsc/documentation/faq.html for trouble shooting.
[0]PETSC ERROR: Petsc Development GIT revision: v3.13-82-ga94d045231  GIT Date: 2020-04-07 10:13:46 -0500
[0]PETSC ERROR: ./navierstokes on a arch-linux2-debug named work by valeria Wed Apr  8 22:35:32 2020
[0]PETSC ERROR: Configure options --download-mumps --download-hypre --download-fblaslapack=1 --download-scalapack --download-metis --download-parmetis --download-cmake --download-ml --download-netcdf --with-cxx-dialect=C++11 --download-hdf5 --download-exodusii --download-pnetcdf --download-mpich --download-superlu_dist --with-zlib PETSC_ARCH=arch-linux2-debug --with-debugging=1 --with-x=1 --download-pnetcdf --download-zlib
[0]PETSC ERROR: #1 MatMult() line 2390 in /home/valeria/Research/CEED/petsc/src/mat/interface/matrix.c
[0]PETSC ERROR: #2 MatInterpolate() line 8092 in /home/valeria/Research/CEED/petsc/src/mat/interface/matrix.c
[0]PETSC ERROR: #3 TSMonitor_NS() line 406 in /home/valeria/Research/CEED/libCEED/examples/fluids/navierstokes.c
[0]PETSC ERROR: #4 main() line 1194 in /home/valeria/Research/CEED/libCEED/examples/fluids/navierstokes.c

```

knepley:
@valeriabarra With that kind of error, it looks to me like things are not updated correctly in the loop because there is a mismatch between the vector and interpolation matrix, and for > 1, you have to interpolate multiple times.

valeriabarra:
> @valeriabarra With that kind of error, it looks to me like things are not updated correctly in the loop because there is a mismatch between the vector and interpolation matrix, and for > 1, you have to interpolate multiple times.

Thanks @knepley. I thought I was interpolating multiple times when I called `DMCreateInterpolation(dmviz, dmvizfine, &interpviz, NULL);` in the loop, but I am surely missing something dumb. The problem arises when refining > 2, not >1. So at least it goes through the loop once.

examples/fluids/navierstokes.c
```diff
@@ -897,12 +898,25 @@ int main(int argc, char **argv) {
...
     ierr = SetUpDM(dmviz, problem, "viz_", &bc, &ctxSetup, NULL); CHKERRQ(ierr);
     ierr = DMCreateInterpolation(dm, dmviz, &interpviz, NULL); CHKERRQ(ierr);
+    for (PetscInt i = 1; i < viz_refine; i++) {
+      ierr = DMRefine(dmviz, MPI_COMM_NULL, &dmvizfine); CHKERRQ(ierr);
+      if (dmvizfine) {
+        ierr = DMSetCoarseDM(dmvizfine, dmviz); CHKERRQ(ierr);
+        ierr = PetscOptionsSetValue(NULL,"-viz_petscspace_degree","1");
+        CHKERRQ(ierr);
+        ierr = SetUpDM(dmvizfine, problem, "viz_", &bc, &ctxSetup, NULL); CHKERRQ(ierr);
+        ierr = DMCreateInterpolation(dmviz, dmvizfine, &interpviz, NULL); CHKERRQ(ierr);
```
valeriabarra, line 915: @knepley Shouldn't this `DmCreateInterpolation()` in the loop update the size of the interpolation matrix?

knepley: Right but it is just interpolating between two refined levels. You have to keep all the interpolators and apply them in sequence. Also, it looks like one DM is not destroyed.

valeriabarra: @knepley @jedbrown 
I've been thinking how to solve this dilemma.
If I use
```
ierr = DMCreateInterpolation(dmviz, dmvizfine, &interpviz[i], NULL); CHKERRQ(ierr);
```
here with an array of length `viz_refine` of interpolation matrices, then when we [use them](https://github.com/CEED/libCEED/blob/6489aa1c689af3398a30fb46563fad9af5aa8ebd/examples/fluids/navierstokes.c#L406) in the TSMonitor_NS, I would need to have stored all the number of `viz_refine` `dm` and `dmviz` DMs, to extract all the `Q` and `Qrefined`  vectors from them of the proper size. 
I don't think this is the route we want to go in terms of memory usage.

knepley: How much more memory do you think it is? Its probably just 2x.

valeriabarra: Only 2x? If I allow the user to specify, say max 10 levels of viz refinement, then it would be 10 `dm` and 10 `dmviz` DMs, from which we will extract 10 `Q` and `Qrefined` vectors, no? Am I seeing this wrong? 
I was never really concerned with memory consumption or performance before, but now I am afraid of almost anything I do in this project :sweat_smile:

valeriabarra: I will try it and inspect the memory usage. Thanks.

valeriabarra:
Thanks @jedbrown! I was going to tackle this on Monday (tomorrow)

examples/fluids/navierstokes.c
jedbrown: Fixed now in this branch. @valeriabarra I'll leave it to you to squash as you see fit.

@knepley Most of the interpolator setup time is in `MatPreallocatorPreallocate_Preallocator`.

Screenshot_20200412_204521

valeriabarra: @jedbrown just curious, how do you visualize this perf?

jedbrown: http://www.brendangregg.com/FlameGraphs/cpuflamegraphs.html
```
perf record -g ./navierstokes ...
perf script | stackcollapse-perf | flamegraph > flame.svg
firefox flame.svg
```
You can click around in the SVG.

knepley: @jedbrown It sounds like someone should speed up MatPreallocator since we are recommending people use it. I would have to do all the work it has to do in order to preallocate.

jedbrown: Yeah, it could be as simple as sorting the entries so entire (block) rows could be inserted at once.

Suppose we wanted to implement refinement of Lagrange bases to the Q_1 spaces on nodes? Then the interpolation matrix would be the identity. What interface would you prefer for such a thing, or is it too special-purpose for your interests. Seems like we could change `DMPlexSetRefinementUniform` to `DMPlexSetRefinementType` with `DMPLEX_REFINE_UNIFORM, DMPLEX_REFINE_COLLOCATE, DMPLEX_REFINE_?` where collocated refinement would be unavailable for non-Lagrange bases (and any Lagrange that doesn't have natural embedded element topology).

valeriabarra:
@jedbrown thanks. I tested this and it looks good to me. If someone approves it, I'd say it is ok to merge this branch (merge only, without squash, to preserve your commits).

examples/fluids/navierstokes.c
knepley: Let me see if I understand. You are saying that projecting P1 to Q1 is identity if the set of vertices is identical and identically numbered? That seems pretty special case. What is the generalization?

jedbrown: Q_k on a coarse mesh to Q_1 on a finer mesh. It's useful for visualization so you can see the non-uniform resolution of the high-order method. This would be moot if we had good visualization for high-order bases, but that work has been slow.

knepley: Okay, that will be a boolean matrix. I guess we could allow the user to tell us it would be boolean, which we could then use to set a floor on zeros in MatSetValues().

jedbrown: We'd still have to be able to define the coordinates of the vertices on the fine mesh. I'm not sure how this would fit into your vision regarding interface, but I know the numerical code that would need to be written. I could write the code (using `DMPlexGetClosureIndices`) to preserve the ordering of vertices (so the interpolation would be a literal identity).

knepley: I was thinking another way. We use the code exactly as it is, but tell MatSetValues() to ignore small values. Then you get the Boolean matrix. Were you thinking you could simplify the matrix construction? I was just trying to get a sparser projector.

jedbrown: I could make the construction fast, but perhaps it wouldn't be a bottleneck to do naive evaluation on the interpolation nodes and discard small values. We'd still need the dispatch to get to refining to Lagrange nodes. I'd be happy to write that code (and it would be useful for preconditioning, which is reason enough to make this a well-supported functionality).

examples/fluids/navierstokes.c
```diff
@@ -542,9 +542,9 @@ PetscErrorCode SetUpDM(DM dm, problemData *problem, const char *prefix,
     PetscFE fe;
     PetscSpace fespace;
     PetscInt ncompq = 5;
-    ierr = PetscFECreateDefault(PETSC_COMM_SELF,problem->dim, ncompq,
-                                PETSC_FALSE, prefix, PETSC_DETERMINE, &fe);
-    CHKERRQ(ierr);
+    ierr = PetscFECreateLagrange(PETSC_COMM_SELF, problem->dim, ncompq,
+                                 PETSC_FALSE, degree, PETSC_DECIDE,
+                                 &fe);
```
valeriabarra, line 547: Sorry, I have just noticed that the `CHKERRQ(ierr);` is missing here. Perhaps @jeremylt can add that to his active branch since he's already making some changes for the updated restrictions in this file?

jedbrown: Damn; sorry. I wish I could just always stick it on the end of the line. This isn't related to restrictions so I can just fix it.

valeriabarra: Yeah I know it would be OT for that PR, but I don't know how to make fixes to a closed PR. Is it possible without having to open a new PR?

jedbrown: You can push another commit to the branch and I can merge it to master without a PR. If these come all alone, I also don't mind committing an occasional one-liner directly on master.

PR #512, opened 2020-04-10
Small fixes and cleanups

valeriabarra:
These are some small fixes of things noticed during the live demo


examples/ceed/ex1-volume.h
```diff
@@ -14,6 +14,10 @@
 // software, applications, hardware, advanced system engineering and early
 // testbed platforms, in support of the nation's exascale computing imperative.
 
+#ifndef ex1_volume_h
+#define ex1_volume_h
+#include <ceed.h>
```
jeremylt, line 19: Do we want to include `ceed.h` in all of these? Is this to silence some IDEs?

jeremylt: While we're at it, do we want to guard the gallery QFs? Just in case I guess?

examples/nek/bps/bps.h
```diff
@@ -14,6 +14,10 @@
 // software, applications, hardware, advanced system engineering and early
 // testbed platforms, in support of the nation's exascale computing imperative.
 
+#ifndef bps_h
+#define bps_h
+#include <ceed.h>
```
jeremylt, line 19: Have you tested this inclusion on the GPU? I feel like this could cause NVRTCC some problems.

valeriabarra: I tested it on OLCF Summit just now. You are right. I didn't know it could be problematic for NVRTCC. I will remove these inclusions for the other files too.

jedbrown: You can likely tune your IDE to use `-include ceed.h`, which means to process as if that header was included. IMO, it would be worthwhile to extend the RTC backends (OCCA and cuda-gen) to be able to include headers instead of hand-coding some declarations into their environment.

examples/ceed/ex1-volume.h
valeriabarra: Yes, it is annoying in my IDE, because it gives all these warnings about unknown types, since there is not include directive in the file directly
Screenshot from 2020-04-10 20-35-59

valeriabarra: I will find a way to change this setting

valeriabarra: > While we're at it, do we want to guard the gallery QFs? Just in case I guess?

I added them. Just to be extra safe in case any of us accidentally includes any of those files more than once. Honestly, since those are not files that the users must use/include, I don't think the risk was high.

valeriabarra: > I will find a way to change this setting

I couldn't find a way to make my IDE happy about the missing include in the same file. Perhaps @v-dobrev knows, since he mentioned he uses the same IDE?

v-dobrev: Maybe this will work: try adding the flags `-include ceed.h` or `-include /path/to/ceed.h` to the project `*.cflags` file. This, however, will probably apply to all c files in the project.

examples/nek/bps/bps.h
jeremylt: The problem, currently, is that including ceed.h overrides the macros that the backends themselves overrode.

jedbrown: We would include a different header `ceed/qfunction.h` or some such, and it would use
```c
#ifndef MACRO
#  define MACRO default def
#endif
```

jeremylt: Ah, that should be easy.

Mildly related, there is actually no technical reason for the Qfunctions to be in a .h. The CUDA backendddoes not care about the file extension. I've thought about putting the Qfunctions in .c files for the solids code so that I don't have to `make - b` or similar nonsense.

jedbrown: The issue as I see it is that there may be stuff that the JIT (OCCA?) can't handle in our `.c` files.

Note that solids has dependencies so you don't need `make -B`. Try touching a header and re-running `make`.

jeremylt: I was going to just rename `linElas.h` -> `linElas.c`. Identical contents.

Ah, right. That fixed that. Solids is the first code where I *don't* have to force compilation after updating the qfunction files.

jedbrown: And #include it? That won't make any difference.

Just add the deps include like I did for solids.

jeremylt: I was going to make a separate .h file with the definitions and include that in the main source and qfunction source. It would just make our examples look a little closer to what people expect from header files. It hasn't been a super high priority of mine though.

jedbrown: In that case, the macro won't be able to provide the source file name. I'd rather have a single .c per model that initializes everything specific to that model.

jeremylt: Right. Duh. I really need to catch up on some sleep and start thinking clearly.

examples/ceed/ex1-volume.h
valeriabarra: Thanks @v-dobrev. Initially I tried to add it as a new line in the `*.cflags` file. Just for future reference: it doesn't work if you put the `-include` in a new line. But it works only if all flags are all the same line. For instance, I have
```
-std=c17 -include ceed.h
```
I was able to figure this out by looking at the [IDE bug report
](https://bugreports.qt.io/browse/QTCREATORBUG-19668)

valeriabarra:
@jedbrown I added a figure of the cubed-sphere. I thought it would be good to add, since some people may not be familiar with the term.

This PR is ready for review. Thanks.

PR #513, opened 2020-04-14
Convert restriction to offsets

jeremylt:
This PR replaces `CeedInterlaceMode` with a component stride. This encodes the exact same information, but it is more flexible and facilitates future work where `CeedElemRestriction`s are being used for more general state vectors, such as with mixed finite element methods.

ToDo:
- [x] convert t3 tests
- [x] convert t5 tests
- [x] convert examples
- [x] CPU impl
- [x] OCCA impl
- [x] CUDA impl
- [x] MAGMA impl
- [x] Python impl

This will let me finish #503 more easily.


jeremylt:
I'm halfway through converting here. Any thoughts on this @jedbrown?

include/ceed.h
```diff
@@ -277,53 +277,41 @@ typedef enum {
...
 /// with vectors created by a Ceed backend.
 /// @ingroup CeedElemRestriction
 CEED_EXTERN const CeedInt CEED_STRIDES_BACKEND[3];
 
-CEED_EXTERN int CeedElemRestrictionCreate(Ceed ceed, CeedInterlaceMode imode,
-    CeedInt nelem, CeedInt elemsize, CeedInt nnodes, CeedInt ncomp,
-    CeedMemType mtype, CeedCopyMode cmode, const CeedInt *indices,
+CEED_EXTERN int CeedElemRestrictionCreate(Ceed ceed, CeedInt nelem,
+    CeedInt elemsize, CeedInt ncomp, CeedInt compstride, CeedInt lsize,
+    CeedMemType mtype, CeedCopyMode cmode, const CeedInt *offsets,
```
jeremylt, line 288: I kept `nnodes` but renamed and rescaled it as `lsize`. It is useful for `CeedElemRestrictionGetMultiplicity()`.

jeremylt:
We are getting closer. `bpssphere` is the only example not working yet.

include/ceed.h
jedbrown: That seems fine to me.

valeriabarra:
> We are getting closer. `bpssphere` is the only example not working yet.

What is the problem you see with `bpssphere` their restriction declarations are very similar to the ones in the other `bps`. I can help

interface/ceed-elemrestriction.c
```diff
@@ -29,12 +29,12 @@
 /**
   @brief Permute and pad indices for a blocked restriction
 
-  @param indices    Array of shape [@a nelem, @a elemsize]. Row i holds the
-                      ordered list of the indices (into the input CeedVector)
+  @param offsets    Array of shape [@a nelem, @a elemsize]. Row i holds the
+                      ordered list of the offsets (into the input CeedVector)
                       for the unknowns corresponding to element i, where
                       0 <= i < @a nelem. All indices must be in the range
                       [0, @a nnodes).
```
jedbrown, line 36:
```suggestion
                      [0, @a lsize).
```

interface/ceed-elemrestriction.c
```diff
@@ -29,12 +29,12 @@
 /**
   @brief Permute and pad indices for a blocked restriction
 
-  @param indices    Array of shape [@a nelem, @a elemsize]. Row i holds the
-                      ordered list of the indices (into the input CeedVector)
+  @param offsets    Array of shape [@a nelem, @a elemsize]. Row i holds the
+                      ordered list of the offsets (into the input CeedVector)
                       for the unknowns corresponding to element i, where
                       0 <= i < @a nelem. All indices must be in the range
```
jedbrown, line 35:
```suggestion
                      0 <= i < @a nelem. All offsets must be in the range
```

jeremylt:
Huh, its working now.

@YohannDudouit and @nbeams, I will be starting on converting the CUDA and MAGMA restrictions tomorrow. It should be pretty straightforward, but I wanted to give you a heads up since you two are the main stewards of those backends.

interface/ceed-elemrestriction.c
jeremylt: Bundled into a larger set of related changes. Thanks.

examples/petsc/setup.h
```diff
@@ -461,7 +460,7 @@ static int CreateRestrictionPlex(Ceed ceed, CeedInterlaceMode imode, CeedInt P,
       }
       // Essential boundary conditions are encoded as -(loc+1)
       PetscInt loc = indices[i] >= 0 ? indices[i] : -(indices[i] + 1);
-      erestrict[eoffset++] = loc/ncomp;
+      erestrict[eoffset++] = loc;
```
jeremylt, line 463: As an extra win, this function starts looking more straightforward with this change.

jeremylt:
There is a single failure for `occa/omp` on AMD. It started some time in the last 17 hours and is likely unrelated to us. I have two CI runs, both with OCCA pinned at a previously good commit, and one run passes and one run fails, so it has to be something in the Travis env that changed for that job.

jeremylt:
Edit: And now the issue is gone.

jeremylt:
@YohannDudouit and @nbeams, I am testing now (as soon as I can get a node), but this should be pretty close.

@YohannDudouit, this replaces the non-templated `nnodes` with the new templated `compstride` for restrictions for the `cuda/gen` so I think this won't hurt performance.

backends/cuda-reg/ceed-cuda-reg-restriction.c
```diff
@@ -275,21 +204,23 @@ static int CeedElemRestrictionOffset_Cuda_reg(const CeedElemRestriction r,
   ierr = CeedElemRestrictionGetCeed(r, &ceed); CeedChk(ierr);
   CeedElemRestriction_Cuda_reg *impl;
   ierr = CeedElemRestrictionGetData(r, (void *)&impl); CeedChk(ierr);
-  CeedInt nelem, elemsize, nnodes;
+  CeedInt nelem, elemsize, lsize, ncomp;
```
jeremylt, line 207: @YohannDudouit, @jedbrown I think I've found an issue with the `cuda/reg` and `cuda/shared` backends for mixed finite elements support. `cuda/ref` and `cuda/gen` are fine with the changes in this PR. 

It looks like knowledge of `nnodes` is being used heavily in the `cuda/reg` restrictions. This will break with mixed finite elements because `nnodes` does not make sense with a vector that holds a elements of different geometries, sizes, number of components, etc.

I can bully these backends into working against the current test suite, but I don't think we should merge something that will break.

So my questions are
1) Is it easy to remove this dependency on `nnodes`? @YohannDudouit, you'd know best how easy this is.
2) If it isn't easy, do we want to retire these backends (or at least temporarily) since they are not our high performance CUDA backend or our reference CUDA backend?

For context, I'm trying to get mixed finite element support added for someone who needs this code within a couple of weeks.

jedbrown: At a glance, it looks like you already removed most/all of those references. Is that not working?

A stop-gap if you need it would be for the backend to convert offsets back into node indices (and fail with a helpful error message if that isn't possible).

jeremylt: I did remove those references by assuming that `nnodes` is given by `lsize / ncomp`, but I am going to violate that assumption next week. I'm actually pretty surprised that this works for our composite operator tests.

jedbrown: My suggestion (in lieu of input from others) would be to have the backend check that its assumptions are satisfied and error if not. That backend might not be usable for the mixed FEM spaces, but it can still pass the existing tests and serve as a platform if someone wants to revisit/optimize a backend with its strategy. I don't think we need to rush in deleting it.

jeremylt: I've been thinking on this. I don't think that my conversion is very robust and it can't really be used as a check. But, I think that this use of `nnodes` is possible equivalent to `(max(offsets) - min(offsets)) / a_factor_that_depends_upon_ncomp_and_compstride`. I could try a one time computation of `max(offsets)` and `min(offsets)` to fix this in a more robust way.

jeremylt: I fixed it, and hopefully its adequately robust for mixed finite elements.

examples/fluids/navierstokes.c
```diff
@@ -230,9 +230,9 @@ static PetscErrorCode CreateRestrictionFromPlex(Ceed ceed, DM dm, CeedInt P,
   ierr = DMGetLocalVector(dm, &Uloc); CHKERRQ(ierr);
   ierr = VecGetLocalSize(Uloc, &Ndof); CHKERRQ(ierr);
   ierr = DMRestoreLocalVector(dm, &Uloc); CHKERRQ(ierr);
-  CeedElemRestrictionCreate(ceed, CEED_INTERLACED, Nelem, PetscPowInt(P, dim),
-                            Ndof/fieldoff[nfields], fieldoff[nfields],
-                            CEED_MEM_HOST, CEED_COPY_VALUES, erestrict, Erestrict);
+  CeedElemRestrictionCreate(ceed, Nelem, PetscPowInt(P, dim), fieldoff[nfields],
+                            1, Ndof, CEED_MEM_HOST, CEED_COPY_VALUES, erestrict,
+                            Erestrict);
```
jedbrown, line 235: How can this be enough? Don't you have to make `erestrict` be an offset? Perhaps use
```c
erestrict[eoffset++] = loc;
```
up above?

interface/ceed-elemrestriction.c
```diff
@@ -178,39 +178,33 @@ const CeedElemRestriction CEED_ELEMRESTRICTION_NONE =
...
+                      the elements and fields given by this restriction.
+  @param mtype      Memory type of the @a offsets array, see CeedMemType
+  @param cmode      Copy mode for the @a offsets array, see CeedCopyMode
+  @param offsets    Array of shape [@a nelem, @a elemsize]. Row i holds the
+                      ordered list of the offsets (into the input CeedVector)
                       for the unknowns corresponding to element i, where
-                      0 <= i < @a nelem. All indices must be in the range
-                      [0, @a nnodes - 1].
+                      0 <= i < @a nelem. All offsets must be in the range
+                      [0, @a lsize - 1].
```
jedbrown, line 196: Do you have code that checks whether this criteria is satisfied?

examples/fluids/navierstokes.c
jeremylt: Fixed. Yet another reason to port this to a single place.

interface/ceed-elemrestriction.c
jeremylt: Do we? No. Can we? Yes.

It would be easy to check. It seems like this would be nice to put in the debugging mode we've been planning? It's only checking `nelem*elemsize` values, so I can put it in this PR if its desired and turn if off by default for the non debugging mode.

jedbrown: Just do it all the time for reference backends?

jeremylt: Right now the ref CPU backend restriction is used for all optimized CPU backends.  I guess I could turn on this check only when the backend resource name contains `ref`?

jedbrown: Yeah, or a private flag that you set to enable that sort of checking behavior (so you aren't checking resource names during execution).

jeremylt: What about the added check? It's only a strcmp on creation of a restriction and offsets verification if we have a ref or memcheck backend.

backends/ref/ceed-ref-restriction.c
```diff
@@ -272,34 +290,66 @@ static int CeedElemRestrictionDestroy_Ref(CeedElemRestriction r) {
...
+    ierr = CeedGetResource(parentCeed, &resource); CeedChk(ierr);
+    if (!strcmp(resource, "/cpu/self/ref/serial")
+        || !strcmp(resource, "/cpu/self/ref/blocked")
+        || !strcmp(resource, "/cpu/self/memcheck/serial")
+        || !strcmp(resource, "/cpu/self/memcheck/blocked")) {
+      CeedInt lsize;
+      ierr = CeedElemRestrictionGetLVectorSize(r, &lsize); CeedChk(ierr);
+
+      for (CeedInt i = 0; i < nelem*elemsize; i++)
+        if (offsets[i] > lsize - 1)
```
jedbrown, line 330: Is this stricter variant correct based on your current assumptions?
```suggestion
        if (offsets[i] < 0 || lsize <= offsets[i] + (ncomp - 1) * compstride)
```

jeremylt:
I'm working on the PR over with MFEM to update their codebase.

doc/sphinx/source/releasenotes.rst
```diff
@@ -4,6 +4,22 @@ Changes/Release Notes
 On this page we provide a summary of the main API changes, new features and examples
 for each release of libCEED.
 
+.. _master:
```
jeremylt, line 7: @jedbrown, @valeriabarra, what about maintaining a running log of big updates to master since the last release? It would make future release notes easier to generate and highlight API changes in master.

jedbrown: Yes, I meant to ask for us to start this practice. I would try to organize similarly to the v0.6 release notes section. Whatever you put there will show up on the website
https://libceed.readthedocs.io/en/latest/releasenotes/

nbeams:
I was just about to add some commits renaming indices --> offsets in the MAGMA backend and updating the comments for the restriction kernels to reflect the new nomenclature.  Should I open a new PR for that?

jedbrown:
Thanks @nbeams. With this just-merged, go ahead and put that in a new PR.

PR #516, opened 2020-04-22
Some fixes in NS miniapp

valeriabarra:
A few things noticed and fixed/improved.

Bug fixes: 
- [x]  A memory leak reported by @jedbrown 

Some improvements:
- [x]  Remove unused variables in main
- [x]  Some small style fixes
- [x]  Report unit of time summary
- [x]  QFunction descriptions
- [x] Make `tap.sh` process multiple test runs which print output


valeriabarra:
Question:
Currently the interface of the `Exact_**()` QFunctions has some unused parameters, namely `time` (for `advection` and `density_current` problems), and `dim` and Nf` for all of them. 

I can see that we provided the support for `time` (even though for `advection` and `density_current` problems currently it is not used) to be able to calculate the exact, time-dependent solution. But I am a bit unclear on what was the intent for `Nf` (supposedly "number of fields") and `dim` ones. Should we remove these unused parameters?

jedbrown:
Note how these `Exact_*` functions are also being used for boundary conditions (see `problem->bc`).

valeriabarra:
> Note how these `Exact_*` functions are also being used for boundary conditions (see `problem->bc`).

Right. I've realized only now that that signature is required by the `DMAddBoundary()` function. Although I have found some contrasting examples, one [with](https://www.mcs.anl.gov/petsc/petsc-current/src/snes/tutorials/ex56.c.html) `time` and one [without](https://www.mcs.anl.gov/petsc/petsc-current/src/dm/impls/plex/tutorials/ex2.c.html).

jedbrown:
https://gitlab.com/petsc/petsc/commit/05decbb5c660ae3e3e52410f97a476c0767a8d2e#note_328655817

examples/fluids/navierstokes.c
```diff
@@ -256,7 +256,8 @@ static int VectorPlacePetscVec(CeedVector c, Vec p) {
   ierr = CeedVectorGetLength(c, &mceed); CHKERRQ(ierr);
   ierr = VecGetLocalSize(p, &mpetsc); CHKERRQ(ierr);
   if (mceed != mpetsc) SETERRQ2(PETSC_COMM_SELF,PETSC_ERR_ARG_INCOMP,
-                                  "Cannot place PETSc Vec of length %D in CeedVector of length %D",mpetsc,mceed);
+                                "Cannot place PETSc Vec of length %D in CeedVector of length %D",
```
valeriabarra, line 259: Do we know why `make-style` would change this line back to having two spaces after the `(`, but it wouldn't for similar macro calls, such as `SETERRQ4` call on line 215?

examples/fluids/navierstokes.c
```diff
@@ -862,34 +863,40 @@ int main(int argc, char **argv) {
...
+  // Distribute the mesh over processes
+  DM               dmDist = NULL;
+  PetscPartitioner part;
+
+  ierr = DMPlexGetPartitioner(dm, &part); CHKERRQ(ierr);
+  ierr = PetscPartitionerSetFromOptions(part); CHKERRQ(ierr);
+  ierr = DMPlexDistribute(dm, 0, NULL, &dmDist); CHKERRQ(ierr);
+  if (dmDist) {
+    ierr = DMDestroy(&dm); CHKERRQ(ierr);
+    dm  = dmDist;
```
jedbrown, line 883: FWIW, the pattern on the left is handy to be explicit that the local variables aren't used outside this scope.

examples/fluids/navierstokes.c
jedbrown: I don't know.

examples/fluids/navierstokes.c
valeriabarra: Yes, I can add the scope back. That was a slip. I just thought the `if(1)` was unnecessary and I was unclear on why it was there. (it could leave the impression that we want an easy and fast switch to test some features on/off for debugging purposes, but I thought the parallel partitioning of the mesh should be an established feature not to be tested)

valeriabarra:
@jedbrown I have a question. If the wall faces are 6, why 
```
PetscInt walls[10], slips[3][10];
```
on line 169 and not
```
PetscInt walls[6], slips[3][6];
```
?

jedbrown:
Should be set dynamically. Not all meshes are a box

valeriabarra:
> Should be set dynamically. Not all meshes are a box

Perhaps we could think of how to implement the dynamic allocation for arbitrary meshes when we do the refactor and addition of the new boundary conditions (since we will have to determine which are the "walls" of those arbitrary-shaped meshes). For now, since in practice we use `DMPlexCreateBoxMesh()` in this code, should I reduce those arrays to 6? I was just wondering how 10 was chosen as a number.

jedbrown:
No need to do it now, but yes, it needs to be addressed in refactoring. It was just mean to be long enough for simple variations.

The most important thing to do before refactoring is to define some fast-running tests that can easily be identified as good or bad. Good in this sense means no worse than we have now.

valeriabarra:
I think I have addressed all the small patches needed for now. Ready for review. Thanks!

examples/fluids/navierstokes.c
```diff
@@ -549,12 +549,21 @@ PetscErrorCode SetUpDM(DM dm, problemData *problem, PetscInt degree,
...
+    // Wall boundary conditions are zero energy density and zero flux for
+    //   velocity in advection/advection2d, and zero velocity and zero flux
+    //   for mass density and energy density in density_current
     {
-      PetscInt comps[3] = {1, 2, 3};
-      ierr = DMAddBoundary(dm, DM_BC_ESSENTIAL, "wall", "Face Sets", 0,
-                           3, comps, (void(*)(void))problem->bc,
-                           bc->nwall, bc->walls, ctxSetup); CHKERRQ(ierr);
+      if (problem->bc == Exact_Advection || problem->bc == Exact_Advection2d) {
+        PetscInt comps[1] = {4};
```
LeilaGhaffari, line 557: What does `4` indicate? In other `comps`, I thought we were giving the set of directions that the BC was implemented on. Isn't it so?

valeriabarra: It's the component. So it's energy

examples/fluids/navierstokes.c
```diff
@@ -763,7 +763,8 @@ int main(int argc, char **argv) {
                             "Scale coefficient for tau (nondimensional)",
                             NULL, CtauS, &CtauS, NULL); CHKERRQ(ierr);
   if (stab == STAB_NONE && CtauS != 0) {
-    ierr = PetscPrintf(comm, "Warning! Use -CtauS only with -stab su or -stab supg\n");
+    ierr = PetscPrintf(comm,
+                       "Warning! Use -CtauS only with -stab su or -stab supg\n");
```
LeilaGhaffari, line 767: @valeriabarra , I guess we need to add space between the inputs in line `582`, right?

valeriabarra: Yup, thanks. I just did.

examples/fluids/navierstokes.c
```diff
@@ -533,8 +533,8 @@ static PetscErrorCode ComputeLumpedMassMatrix(Ceed ceed, DM dm,
   PetscFunctionReturn(0);
 }
 
-PetscErrorCode SetUpDM(DM dm, problemData *problem, PetscInt degree,
-                       SimpleBC bc, void *ctxSetup) {
+static PetscErrorCode SetUpDM(DM dm, problemData *problem, PetscInt degree,
```
jeremylt, line 537: I've thought about suggesting that we move these functions into their own file to make it a little bit easier to find things, like we did with the solids code. I'm not sure if 1) other people like that and 2) other people find it valuable.

valeriabarra: Definitely. This code needs some major refactoring and separation of modules. I was leaving that to subsequent PRs as we were discussing on Zulip. For now this PR was mainly focused with fixing impelling issues like broken pieces and functionalities.

valeriabarra:
> No need to do it now, but yes, it needs to be addressed in refactoring. It was just mean to be long enough for simple variations.
> 
> The most important thing to do before refactoring is to define some fast-running tests that can easily be identified as good or bad. Good in this sense means no worse than we have now.

@jedbrown How do we want to proceed with this PR? I have been trying to get "good" tests for us to  use as testbeds. For now, I can share the CL options for the ones with all slip faces (with either implicit temporal integrator C-N or alpha) I shared the results of on Zulip (which still develop the instabilities discussed). But perhaps we can consider merging this PR, which has a few fixes, while I keep investigating the nature of the instabilities.

```
mpiexec -n 4 ./navierstokes -problem density_current -degree 3 -qextra 4 -output_freq 100 -ts_adapt_monitor -lx 2000 -ly 125 -lz 1000 -rc 400. -dm_plex_box_faces 16,1,8 -mu 75. -thetaC -25. -units_meter 1e-2 -units_second 1e-2 -units_kilogram 1e-9 -ts_type cn -snes_mf -snes_monitor -ksp_monitor -ts_view -ksp_converged_reason -snes_converged_reason -viz_refine 2 -log_view
```

or 

```
mpiexec -n 4 ./navierstokes -problem density_current -degree 3 -qextra 2 -dm_plex_box_faces 16,1,8 -implicit -ksp_atol 1e-4 -ksp_monitor -ksp_rtol 1e-3 -ksp_type bcgs -log_view -lx 2000 -ly 125 -lz 1000 -mu 75. -output_freq 25 -rc 400. -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -snes_monitor -stab supg -thetaC -25. -ts_dt 5e-4 -ts_max_time 2.5 -ts_monitor -ts_type alpha -viz_refine 2 -log_view
```

valeriabarra:
@jedbrown I'm not sure if this is what you had envisioned. This currently works only in serial though. 

Next step would be making `tap.sh` process multiple tests that produce some numerical output.

valeriabarra:
@jedbrown do you have any recommendation for making the view that I set up work in parallel? Should I have used another viewer instead of from CL option? If so, which one? Thanks!

valeriabarra:
Thanks. But it doesn't work in parallel. It crashes. 😔 I have to figure out why

jedbrown:
What crashes in parallel?

valeriabarra:
It crashes when I use the same args as in TESTARGS, but under mpiexec with n different from 1. While it deosn't happen in serial. It seems to be an issue in the element restriction which I don't understand.
```
$ mpiexec -n 2 ./navierstokes  -test -degree 3 -dm_plex_box_faces 1,1,2 -lx 125 -ly 125 -lz 250 -center 62.5,62.5,187.5 -rc 100. -thetaC -35. -ksp_atol 1e-4 -ksp_rtol 1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3 -implicit -ts_type alpha -stab supg -vec_view
/home/valeria/Research/CEED/libCEED/interface/ceed-elemrestriction.c:179 in CeedElemRestrictionGetBackendStridesStatus(): ElemRestriction has no stride data

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 7667 RUNNING AT work
=   EXIT CODE: 134
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Aborted (signal 6)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```

valeriabarra:
And just to be clear, this is unrelated to the commit in which I introduced the VecView (`52eb2b6`). It happens even if I go back to [911f3b1](https://github.com/CEED/libCEED/pull/516/commits/911f3b1145c5236bad48c6bcaf7843603791acb8)

valeriabarra:
And I have tested that this crash in parallel happens also on master. So not related to this PR, but perhaps to the StridedRestrictions

jeremylt:
Stack trace? That function shouldn't be called for a non-strided restriction.

valeriabarra:
> Stack trace? That function shouldn't be called for a non-strided restriction.

There is none unfortunately, even if I compiled using the debug-mode PETSC_ARCH

jeremylt:
I have a funny suspicion that it might be a bizarre case where there is an empty MPI process and an empty restriction, so no restriction indices, which is confusing the backend. I can try something in the morning and see if it fixes it.

Basically I think I need to replace all of the `if (!offsets)` instances in the backend restriction init.

valeriabarra:
@jeremylt is it possible that since this test has very few elements (2) of relatively low degree (3), when the very little data is partitioned across multiple processes, something funny that was not considered happens in the strided restriction? 
You did provide support to handle empty L-vectors a while ago, when these were arising from an unbalanced data partition. But I'm wondering if something similar is happening

valeriabarra:
Ah, we wrote basically the same thing at the same time!

valeriabarra:
I have made `tap.sh` process the fluid example outputs. Now only one thing left to do here:
- [ ] Make the VecView output format something like `%10.7f`. I have noticed that some backends produce a slightly different output at the eight decimal digit—I think it's a tolerance that we can bear, can't we? Would this be a possible approach?

valeriabarra:
Update:

In [98cb202](https://github.com/CEED/libCEED/pull/516/commits/98cb20282e0ab631d2c34b1284529cdca72dc5e4) I tried to output a different number of decimals instead of the default one given by `VecViewFromOptions()`. I thought this way I could avoid the test failing on some backends, because they produce slightly different numerical values. But this was still not sufficient because we still have to deal with rounding errors. For instance `248.33739` vs `248.33740` as seen below:

```
# TESTARGS:  -ceed {ceed_resource} -test -degree 3 -dm_plex_box_faces 1,1,2 -units_kilogram 1e-9 -lx 125 -ly 125 -lz 250 -center 62.5,62.5,187.5 -rc 100. -thetaC -35. -ksp_atol 1e-4 -ksp_rtol 1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3 -implicit -ts_type alpha -stab supg -vec_view
ok 79 fluids-navierstokes /cpu/self/ref/serial
Suffices: -implicit-supgstab    ok 80 fluids-navierstokes /cpu/self/ref/serial stdout
ok 81 fluids-navierstokes /cpu/self/ref/serial stderr
ok 82 fluids-navierstokes /cpu/self/ref/blocked
Suffices: -implicit-supgstab    not ok 83 fluids-navierstokes /cpu/self/ref/blocked stdout
# --- examples/fluids/output/fluids-navierstokes-implicit-supgstab.out	2020-05-05 18:56:35.014227228 -0600
# +++ fluids-navierstokes.ruZz.out	2020-05-05 18:58:32.392860936 -0600
# @@ -408,7 +408,7 @@
# 249.30360
# 0.00115
# 0.00014
# -248.33739
# +248.33740
# 0.00115
# 0.00047
# 247.08747
ok 84 fluids-navierstokes /cpu/self/ref/blocked stderr

```

valeriabarra:
At this point I am not sure what is the best way to deal these types of checks for numerical values for testing purposes.

jedbrown:
There are a few options. One is to load the reference solution in and compute the relative difference (printing if large). In that case, the reference solution can be saved as PETSc binary (so you get full double precision and can `VecLoad`). Another is to use a fuzzy diff that only prints in case of large relative difference; this doesn't work for things like convergence history because residuals are almost all noise when they approach machine precision.

valeriabarra:
Thanks for the tips! 
I've been thinking about this and these are my considerations:

* I think option 1 is viable but since these were fast tests designed to run in a few seconds, they accounted only for a few time steps history anyway (it was set to 10). But even if I increased the number of steps to, say, 50 (a number of steps that still allows the test to be run under a minute, that you said was desired), the history won't be long enough to tell us much about possible divergence due to numerical instabilities or other factors, which usually appear in a much longer time frame. That's why I initially thought about 10 steps. If something is disastrously off (because of a bug in a boundary condition, IC, or scaling of any quantity) it should already show up in a relatively small number of steps. So, having more history that still matches the "fast-running" requirement I think won't help much in our di/convergence analysis in this case, but it will only give us more levels of discretion (other than setting up a desired tolerance for the error, we would also have to decide after how many steps we can allow differences to occur. Is it ok if they are equal for a few steps, say 5, and then they start to diverge? How do you decide this number?). So in summary, I think this option is doable, but it adds more levels of discretion on our side and doesn't really help us much in identifying development errors that would be disastrous anyway and would probably show up in easier ways (like simple inspection).

* Option 2 sounds the fastest to develop for me right now. I can revert the last commit and go back to [27086b0](https://github.com/CEED/libCEED/pull/516/commits/27086b0693f6c0d1e4de47c665a35891a475731e) but I wasn't able to find anything really useful about the fuzzy diff online. Do you know any specific tool or way to implement this, which I should look into?

Thanks a lot for your suggestions!

jedbrown:
The concerns you have with 1 are all still present for 2, and the reason we'd prefer to test with manufactured solutions. It doesn't make sense to have exact agreement for several time steps and then suddenly behave differently. We're usually not dealing with linearly unstable behavior (exponential growth), so it's hard to make a sharp test. So long as there are no right answers, I'd say just pick something that seems reasonable and check that it notices when you make small adjustments to parameters.

`$PETSC_DIR/lib/petsc/bin/petscdiff` does a comparison that ignores floating point values (i.e., for convergence testing), but shows proper diffs if no match. The MOOSE project wrote `exodiff`, which does fuzzy diffs on solutions written in Exodus format. I think the easiest will be `-compare_final_solution path/to/state.petsc`.

Are you still having an issue with parallel runs?

valeriabarra:
The parallel run is problematic only in `98cb202` now, not in the previous commits. That is because of my clumsy attempt with the parallel print for a specified format. If I go back using `VecViewFromOptions()` with its default format, there is not issue in parallel

jedbrown:
I would recommend viewing to a binary file and comparing `PetscAbsScalar(u[i] - uref[i]) < tol * norm_uref`.

tests/tap.sh
```diff
@@ -31,7 +32,12 @@ elif [ ${1::4} == "nek-" ]; then
       allargs+=("$(awk -v i="$i" '/C_TESTARGS/,/\n/{j++}j==i+1{print; exit}' examples/nek/bps/${1:4}.usr* | cut -d\  -f2- )")
     done
 elif [ ${1::7} == "fluids-" ]; then
-    allargs=$(grep -F //TESTARGS examples/fluids/${1:7}.c* | cut -d\  -f2- )
+    # get all test configurations
+    numconfig=$(grep -F //TESTARGS examples/fluids/${1:7}.c* | wc -l)
+    for ((i=0;i<${numconfig};++i)); do
+      suffices+=("$(awk -v i="$i" '/\/\/TESTARGS/,/\n/{j++}j==i+1{print; exit}' examples/fluids/${1:7}.c | cut -d\  -f2- | cut -d ";" -f 1 | cut -d "=" -f 2 )")
```
valeriabarra, line 38: Just to be clear, this was done because initially we said to to support CI processing for multiple test configurations (i.e., multiple `//TESTARGS` lines with different suffices) which produce some numerical output. But now, my current understanding of the latest suggestion is that the check for consistency with the reference solution is going to happen in the user code (I am on it), hence all this machinery in `tap.sh` will be gone, and `tap` will just check if the comparison has produced some output or not, as for all other examples

jedbrown: If this code is debugged, we'll probably use it eventually.

valeriabarra: Yes it is and I had tested it. How to preserve it for future reference?

valeriabarra: These edits to `tap.sh` to process multiple tests with output are all contained in one commit ([a0c2d49](https://github.com/CEED/libCEED/pull/516/commits/a0c2d496e17e245ba9197a78c69f3b0c81057b90)), so it will be easy to retrieve that for future reference if we don't squash and merge

valeriabarra:
Ready for final review. Thanks!!

examples/fluids/navierstokes.c
```diff
@@ -765,6 +765,10 @@ int main(int argc, char **argv) {
   ierr = PetscOptionsScalar("-CtauS",
                             "Scale coefficient for tau (nondimensional)",
                             NULL, CtauS, &CtauS, NULL); CHKERRQ(ierr);
+  if (stab == STAB_NONE && CtauS != 0) {
+    ierr = PetscPrintf(comm, "Warning! Use -CtauS only with -stab su or -stab supg\n");
+    CHKERRQ(ierr);
+  }
```
LeilaGhaffari: @valeriabarra , thank you for adding this warning. Actually, `CtauS` and `strong_form` are not used in `density densitycurrent`. Maybe adding warning for those too?!

LeilaGhaffari: I think we don't need line 64 and 65 of `densitycurrent.h`.

examples/fluids/advection.h
```diff
@@ -145,7 +146,7 @@ static inline int Exact_Advection(CeedInt dim, CeedScalar time,
 }
 
 // *****************************************************************************
-// Initial conditions for 3D advection
+// This QFunction sets the initial conditions for 3D advection
```
jeremylt, line 149: This change is fine - I just don't want to see us do another massive code sweep to change all comments to sentences

examples/fluids/navierstokes.c
```diff
@@ -1235,23 +1266,79 @@ int main(int argc, char **argv) {
...
+        ierr = PetscStrncpy(filepath,
+                            "examples/fluids/tests-output/fluids-navierstokes-implicit-supgstab.bin",
+                            sizeof filepath);
+        CHKERRQ(ierr);
+      }
+    }
+    // Read reference file
+    Vec Qref;
+    PetscReal error, Qrefnorm;
+    ierr = VecDuplicate(Q, &Qref);
```
jeremylt, line 1323: Missing CHKERRQ

examples/fluids/navierstokes.c
```diff
@@ -1235,23 +1266,79 @@ int main(int argc, char **argv) {
...
+                            sizeof filepath);
+        CHKERRQ(ierr);
+      }
+      else {
+        ierr = PetscStrncpy(filepath,
+                            "examples/fluids/tests-output/fluids-navierstokes-implicit-supgstab.bin",
+                            sizeof filepath);
+        CHKERRQ(ierr);
+      }
+    }
```
jeremylt, line 1319: Maybe we should make a testType enum, testTypes strings, testData struct, and testOptions global variable. I found myself working with similar if chains when I rewrote the PETSc BPs. These additions would remove the if chains with repeated code and make it easier to add future tests.

tests/tap.sh
```diff
@@ -175,6 +180,15 @@ for ((i=0;i<${#backends[@]};++i)); do
...
 
+    # grep to skip t506 for MAGMA, range of basis kernels limited for now
+    if [[ "$backend" = *"magma" ]] \
+            && [[ "$1" = "t506"* ]] ; then
+        printf "ok $i0 # SKIP - backend basis kernel not available $1 $backend\n"
+        printf "ok $i1 # SKIP - backend basis kernel not available $1 $backend stdout\n"
+        printf "ok $i2 # SKIP - backend basis kernel not available $1 $backend stderr\n"
+        continue
+    fi
+
```
jeremylt, line 191: I think this should be in master already?

examples/fluids/navierstokes.c
```diff
@@ -110,7 +112,7 @@ problemData problemOptions[] = {
     .apply_ifunction     = IFunction_Advection,
     .apply_ifunction_loc = IFunction_Advection_loc,
     .bc                  = Exact_Advection,
-    .non_zero_time       = true,
+    .non_zero_time       = false,
```
jeremylt, line 115: PETSC_FALSE?

examples/fluids/navierstokes.c
```diff
@@ -882,14 +911,19 @@ int main(int argc, char **argv) {
...
+  // Setup DM
   ierr = DMLocalizeCoordinates(dm); CHKERRQ(ierr);
   ierr = DMSetFromOptions(dm); CHKERRQ(ierr);
   ierr = SetUpDM(dm, problem, degree, &bc, &ctxSetup); CHKERRQ(ierr);
+
+  // Print FEM space information
   if (!test) {
     ierr = PetscPrintf(PETSC_COMM_WORLD,
-                       "Degree of FEM Space: %D\n",
+                       "Degree of FEM space: %D\n",
```
jeremylt, line 922: At some point it might be nice to make this output match the rest of the PETSc examples. (maybe this is that point?)

examples/fluids/navierstokes.c
```diff
@@ -1128,8 +1160,8 @@ int main(int argc, char **argv) {
   ierr = ComputeLumpedMassMatrix(ceed, dm, restrictq, basisq, restrictqdi, qdata,
                                  user->M); CHKERRQ(ierr);
 
-  ierr = ICs_PetscMultiplicity(op_ics, xcorners, q0ceed, dm, Qloc, Q, restrictq,
-                               &ctxSetup, 0.0);
+  ierr = ICs_FixMultiplicity(op_ics, xcorners, q0ceed, dm, Qloc, Q, restrictq,
```
jeremylt, line 1163: CHKERRQ

jeremylt:
Travis has make style complaints.

tests/tap.sh
valeriabarra: Yes, I didn't change this part. This change comes from merging with the master branch

jeremylt: Weird. In that case this branch should know that this doesn't belong to this branch and is already in master.

If I understood the discussion above and the plan is to merge all of these commits into master instead of squashing, then we might want to remove that merging of master into this branch and instead do a rebase.

examples/fluids/advection.h
valeriabarra: I just wanted to be consistent with line 174 below

tests/tap.sh
valeriabarra: We don't need to merge all commits in this PR, but yes, I would retain at least a few, including ([a0c2d49](https://github.com/CEED/libCEED/pull/516/commits/a0c2d496e17e245ba9197a78c69f3b0c81057b90)) for future reference. When I attempt complex rebases with merges with master in between, git always hates me :( Perhaps someone else can help me with that when we are to merge this branch and at least retain a few commits, including that important one `a0c2d49` at the very least

examples/fluids/advection.h
jeremylt: Like I said, I think that it looks fine here. I was just voicing a concern - specifically that I don't think we need to make a big code sweep to turn all comments into sentences.

examples/fluids/navierstokes.c
valeriabarra: Thanks Leila!

examples/fluids/navierstokes.c
valeriabarra: Thanks! Fixed

tests/tap.sh
jedbrown: That type of rebase is fundamentally hard. It's better to clean up the history before merging from 'master' (or rebasing your clean branch onto 'master'). When I have a dirty branch with merges, I usually make a squashed merge (yields a single commit on top of 'master'), then use `git citool` to extract out the parts I want in separate commits.

tests/tap.sh
jedbrown: That commit could be cherry-picked, but I don't mind if it just rides in with this branch.

tests/tap.sh
valeriabarra: Yes, I understand. I had cleaned up the history before the merge with master a little bit, but not extensively to one commit only, to help reviewers follow the changes since the review was still in process.

examples/fluids/navierstokes.c
valeriabarra: Done. Please let me know if it looks good to you

examples/fluids/navierstokes.c
valeriabarra: Done. Please let me know if it looks good to you

examples/fluids/navierstokes.c
```diff
@@ -677,8 +740,12 @@ int main(int argc, char **argv) {
   ierr = PetscOptionsString("-ceed", "CEED resource specifier",
                             NULL, ceedresource, ceedresource,
                             sizeof(ceedresource), NULL); CHKERRQ(ierr);
-  ierr = PetscOptionsBool("-test", "Run in test mode",
-                          NULL, test=PETSC_FALSE, &test, NULL); CHKERRQ(ierr);
+  testChoice = TEST_NONE;
```
jeremylt, line 743: Would it be possible to initialize this to NULL so that we don't need the special TEST_NONE?

examples/fluids/navierstokes.c
```diff
@@ -938,18 +1023,17 @@ int main(int argc, char **argv) {
     ierr = MPI_Comm_size(comm, &comm_size); CHKERRQ(ierr);
     ierr = PetscOptionsGetString(NULL, NULL, "-dm_plex_box_faces", box_faces_str,
                                  sizeof(box_faces_str), NULL); CHKERRQ(ierr);
-    if (!test) {
-      ierr = PetscPrintf(comm, "Global FEM dofs: %D (%D owned) on %d ranks\n",
+    if (testChoice == TEST_NONE) {
```
jeremylt, line 1026: This would become !testChoice if setting to NULL initially works.

examples/fluids/navierstokes.c
jeremylt: I had a couple of comments, but overall I think this is much cleaner. Thanks.

examples/fluids/navierstokes.c
valeriabarra: `testChoice` is of type `testType`, which is an enum. So we could not assign `NULL` to an enum type. It would cause an assignment mismatch warning:
```
warning: incompatible pointer to integer conversion assigning to 'testType' from 'void *'
```

examples/fluids/navierstokes.c
valeriabarra: Done. Let me know if it looks good to you. Thank you

valeriabarra:
> Travis has make style complaints.

Fixed them. Thanks

valeriabarra:
The Travis failure here it's not related to this PR (It's the `not ok 26 petsc-bpsraw /omp/occa stdout`)

valeriabarra:
If there are no further comments, this is ready for git magic and merge :)

examples/fluids/navierstokes.c
```diff
@@ -1013,33 +1029,68 @@ int main(int argc, char **argv) {
...
-      CHKERRQ(ierr);
-    }
+    const char *usedresource;
+    CeedGetResource(ceed, &usedresource);
+
+    ierr = PetscPrintf(comm,
+                       "\n-- libCEED Navier-Stokes solver + PETSc --\n"
+                       "  rank(s):                             : %d\n"
+                       "  PETSc:\n"
+                       "    dm_plex_box_faces:                 : %s\n"
```
jeremylt, line 1067: A few small thoughts,
- The title is a bit different in style from the solilds example, though that's no big deal to me
"\n-- Elasticity Example - libCEED + PETSc --\n"

- Is there a better name for dm_plex_box_faces? Box Faces?

- Do we want to include the problem type and stabilization type?

Overall I like this a lot better.

valeriabarra: Done in [6f55dfd](https://github.com/CEED/libCEED/pull/516/commits/6f55dfd5a730b78d8f668c0a3f4f8ca678245335) which unfortunately also includes some changes from master after I tried to reduce the number of commits (my bad). 

Please let me know if the summary print statement looks good now. Thanks

examples/fluids/navierstokes.c
```diff
@@ -1077,7 +1079,8 @@ int main(int argc, char **argv) {
                        "    DoFs per node                      : %D\n"
                        "    Global nodes                       : %D\n"
                        "    Owned nodes                        : %D\n",
-                       comm_size, box_faces_str, usedresource,
+                       comm_size, problemTypes[problemChoice],
```
jeremylt, line 1082: Should be under the heading
                       "  Problem:\n"
maybe?

valeriabarra: I don't think it needs a dedicate heading since there is not other option in this case. 
To me having 
```
                       "  Problem:\n"
                       "    Problem Name                       : %s\n"
```
only with the name, it would look a bit awkward. But if you really like it that way, I can change it of course

jeremylt: It would be problem name + stabilization type under the heading though?

valeriabarra: Ah ok, now I see what you meant

jeremylt: Looking at my original comment, I probably should have used more words. It's time for some tea.

PR #519, opened 2020-04-23
Ref - drop unneeded malloc and memcpy for strided restrictions

jeremylt:


backends/ref/ceed-ref-restriction.c
```diff
@@ -337,23 +338,23 @@ int CeedElemRestrictionCreate_Ref(CeedMemType mtype, CeedCopyMode cmode,
...
-    impl->offsets = impl->offsets_allocated;
-    break;
-  case CEED_OWN_POINTER:
-    impl->offsets_allocated = (CeedInt *)offsets;
-    impl->offsets = impl->offsets_allocated;
-    break;
-  case CEED_USE_POINTER:
-    impl->offsets = offsets;
+    // Copy data
+    switch (cmode) {
```
jeremylt, line 343: I brought this inside the `if(offsets)` so we don't pointlessly `malloc` for strided restrictions. I overlooked it before.

jeremylt:
I'm seeing this issue in Travis as well. Our `make tidy` check on Travis seems to still not be aggressive enough.

PR #520, opened 2020-04-24
CeedElemRestrictionGetOffsets

jeremylt:
This cleans up the backend interface a bit and makes it easier to create related restrictions.

### DIFF
--- a/backends/cuda-reg/ceed-cuda-reg-restriction.c
+++ b/backends/cuda-reg/ceed-cuda-reg-restriction.c
@@ -212,6 +212,26 @@ static int CeedElemRestrictionApply_Cuda_reg(CeedElemRestriction r,
 }
 
 //------------------------------------------------------------------------------
+// Get offsets
+//------------------------------------------------------------------------------
+static int CeedElemRestrictionGetOffsets_Cuda_reg(CeedElemRestriction rstr,
+    CeedMemType mtype, const CeedInt **offsets) {
+  int ierr;
+  CeedElemRestriction_Cuda_reg *impl;
+  ierr = CeedElemRestrictionGetData(rstr, (void *)&impl); CeedChk(ierr);
+
+  switch (mtype) {
+  case CEED_MEM_HOST:
+    *offsets = impl->h_ind;
+    break;
+  case CEED_MEM_DEVICE:
+    *offsets = impl->d_ind;
+    break;
+  }
+  return 0;
+}
+
+//------------------------------------------------------------------------------
 // Destroy restriction
 //------------------------------------------------------------------------------
 static int CeedElemRestrictionDestroy_Cuda_reg(CeedElemRestriction r) {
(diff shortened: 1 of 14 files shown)


backends/cuda-reg/ceed-cuda-reg-restriction.c
```diff
@@ -211,6 +211,26 @@ static int CeedElemRestrictionApply_Cuda_reg(CeedElemRestriction r,
...
+    CeedMemType mtype, const CeedInt **offsets) {
+  int ierr;
+  CeedElemRestriction_Cuda *impl;
+  ierr = CeedElemRestrictionGetData(rstr, (void *)&impl); CeedChk(ierr);
+
+  switch (memtype) {
+  case (CEED_MEM_HOST)
+      *offsets = impl->h_ind;
+    break;
+  case (CEED_MEM_DEVICE)
```
jedbrown, line 227: Odd style; no parens needed in the case statement. Also, why does this need to be different from `CeedElemRestrictionGetOffsets_Cuda` (which looks like a copy).

interface/ceed-elemrestriction.c
```diff
@@ -102,6 +102,32 @@ int CeedElemRestrictionGetStrides(CeedElemRestriction rstr,
...
+
+  if (!rstr->GetOffsets)
+    // LCOV_EXCL_START
+    return CeedError(rstr->ceed, 1, "Backend does not support GetOffsets");
+  // LCOV_EXCL_STOP
+
+  ierr = rstr->GetOffsets(rstr, mtype, offsets); CeedChk(ierr);
+  return 0;
+}
+
```
jedbrown, line 130: These being substantial data, should we have to return the pointers when we're done?

backends/cuda-reg/ceed-cuda-reg-restriction.c
jeremylt: Typo there on my part. The two backends have different backend data structs.

interface/ceed-elemrestriction.c
jeremylt: Since this is access should be read only, is there an intended outcome if a user/backend does not return the pointer before destruction? We currently do not error for a `CeedVector` if read only access is never returned by the time the vector is destroyed. Is this a bug?

jeremylt: I added a readers counter. I realized that it's possible to create a second restriction with a original restriction's offsets with `CEED_USE_POINTER`, destroy the first restriction, and break the second restriction.

interface/ceed-elemrestriction.c
```diff
@@ -146,6 +146,22 @@ int CeedElemRestrictionRestoreOffsets(CeedElemRestriction rstr,
...
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedElemRestrictionGetStridedStatus(CeedElemRestriction rstr,
+    bool *status) {
+  *status = rstr->strides ? 1 : 0;
+  return 0;
+}
+
```
jeremylt, line 164: @nbeams, is this what you were thinking of?

nbeams: Yes, thanks!  I think this will be very useful

jeremylt:
Ready for review. If desired, I can make a separate PR to error on destroying a vector if there are any readers.

PR #521, opened 2020-04-27
MAGMA: simplify build

nbeams:
This PR removes the script `gccm.py`, which is no longer used in the current MAGMA backend, and simplifies the Makefile commands related to the MAGMA backend.


nbeams:
Bringing over my comment from Slack:
t506 is a known failure for the MAGMA backend, unrelated to this PR.  It's because the MAGMA backend uses templated kernels for tensor basis actions, and the templates are currently only instantiated up to order 10 (for either p or q).  t506 uses a q of 24, so the kernel launch fails.  We had discussed adding some sort of fallback kernels or increasing the ranges of instantiation, but hadn't decided on exactly what to do yet.

Since it's a separate issue, am I okay to merge this PR to streamline the build?

nbeams:
Great, thanks for the fast reviews, everyone!

PR #522, opened 2020-05-01
Soldis - fix small bug in energy restriction

jeremylt:


valeriabarra:
Hey Jeremy, while at it, perhaps you can integrate this [suggested documentation change](https://github.com/CEED/libCEED/pull/504#discussion_r409904630) that was left when revising #504, if you agree it would make it clearer. Thanks

jeremylt:
I have no strong feelings. For me, documentation with more words can be harder to parse than smaller smaller, more direct comments. But I can see how more verbose can be clearer for others.

valeriabarra:
> I have no strong feelings. For me, documentation with more words can be harder to parse than smaller smaller, more direct comments. But I can see how more verbose can be clearer for others.

The important part of that comment was that if one option must precedes the other, this should be specified somewhere. The rest was just for clarity.

jeremylt:
Just checked - One precedes the other in our CL option processing, but PETSc processes the same way no matter the given order.

valeriabarra:
I was looking at the `SetupLibceedByDegree()` function in other setup files (like `petsc/setup.h` or `petsc/setupsphere.h`), which I think is similar to the `SetupLibceedFineLevel()` function here, and I'm not sure why we are getting the section (in `setuplibceed.c:252`)

```
ierr = DMGetSection(dmcoord, &section); CHKERRQ(ierr);
```
if then we are not using it, are we?

jeremylt:
It's a stray piece of CreateRestrictionPlex(). We should do a PR to pull that stray piece from all of the PETSc examples.

valeriabarra:
Okay, that's what I thought

PR #523, opened 2020-05-01
Jed/breathe 4.17

jedbrown:
This upgrades breathe and switches to C domain, because libCEED is a C interface. It also avoids needing the `restrict` hack (though we took `restrict` out of the public interface).

Rendered: https://libceed.readthedocs.io/en/jed-breathe-4.17/

Cc: https://github.com/michaeljones/breathe/issues/477#issuecomment-622525692


jeremylt:
Do we want to take this opportunity to eliminate the unused ``* :ref:`modindex` ``?

PR #525, opened 2020-05-04
Fix make tidy for CUDA/MAGMA

jeremylt:
I fixed `make tidy` for CUDA/MAGMA and added a few fixes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -565,7 +565,7 @@ style : style-c style-py
 
 CLANG_TIDY ?= clang-tidy
 %.c.tidy : %.c
-	$(CLANG_TIDY) $(TIDY_OPTS) $^ -- $(CPPFLAGS) --std=c99
+	$(CLANG_TIDY) $(TIDY_OPTS) $^ -- $(CPPFLAGS) --std=c99 -I$(CUDA_DIR)/include
 
 tidy : $(libceed.c:%=%.tidy)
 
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -565,7 +565,7 @@ style : style-c style-py
 
 CLANG_TIDY ?= clang-tidy
 %.c.tidy : %.c
-	$(CLANG_TIDY) $(TIDY_OPTS) $^ -- $(CPPFLAGS) --std=c99
+	$(CLANG_TIDY) $(TIDY_OPTS) $^ -- $(CPPFLAGS) --std=c99 -I$(CUDA_DIR)/include
```
jeremylt, line 568: When `CUDA_DIR` is not defined, this becomes `-I/include` which is perfectly valid. It works fine in Travis and locally.

backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -113,7 +113,7 @@ static int CeedCudaLoadQFunction(CeedQFunction qf, char *c_src_file) {
   if (!last_dot)
     return CeedError(ceed, 1, "Cannot find file's extension!");
   const size_t cuda_path_len = last_dot - cuda_file;
-  strcpy(&cuda_file[cuda_path_len], ".h");
+  strncpy(&cuda_file[cuda_path_len], ".h", 3);
```
jeremylt, line 116: This technically imposes the requirement that the code is in a `.h`. Do we want to relax this and allow any extension?

This was originally added to the OCCA backend to convert `foo.c` to `foo.okl`. With the `CeedQFunction` macro we get the actual source path of the QFunction, so I think removing this would be perfectly fine.

PR #531, opened 2020-05-05
XSMM - fix q=1 index computation

jeremylt:
This fixes the underintegration case of `q = 1` for the libXSMM backends so that we can use this backend for the new solids formulation. I was accidentally dividing by zero.

When we use Q2 elements, we're actually in the regime where our AVX can do better than LIBXSMM, but I get nice results with Q3s.

### DIFF
--- a/backends/xsmm/ceed-xsmm-tensor.c
+++ b/backends/xsmm/ceed-xsmm-tensor.c
@@ -20,9 +20,10 @@
 // Get Kernel Index
 //------------------------------------------------------------------------------
 int CeedGetXsmmInd_NonTensor(CeedInt add, CeedInt P, CeedInt Q, CeedInt B,
-                             CeedInt C, CeedInt J) {
-  return (C == 8 ? 1:0)*4*2 + (add ? 1:0)*4 +
+                             CeedInt C, CeedInt J, CeedInt *ind) {
+  *ind = (C == 8 ? 1:0)*4*2 + (add ? 1:0)*4 +
          (B == P ? (J == Q ? 0:1) : (B == Q ? 2:3));
+  return 0;
 }
 
 //------------------------------------------------------------------------------
@@ -69,12 +70,12 @@ static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A,
 
   // Get kernel index
   if (impl->tensorbasis) {
-    CeedInt key = ((B*impl->indScale + C)*impl->indScale + J)*10 +
-                  2*(tmode == CEED_TRANSPOSE) + add;
+    CeedHashIJKLMKey key = {B, C, J, tmode, add};
     khint_t k = kh_get(m32, impl->lookup, key);
-    ind = kh_value(impl->lookup, k);
+    CeedHashGetValue(impl->lookup, k, ind);
   } else {
-    ind = CeedGetXsmmInd_NonTensor(add, impl->P, impl->Q, B, C, J);
+    ierr = CeedGetXsmmInd_NonTensor(add, impl->P, impl->Q, B, C, J, &ind);
+    CeedChk(ierr);
   }
 
   // Run kernel or fallback to default implementation
@@ -127,9 +128,6 @@ int CeedTensorContractCreate_Xsmm(CeedBasis basis,
     ierr = CeedCalloc(impl->numkernels, &impl->kernels); CeedChk(ierr);
     // Setup indices hash table
     CeedInt ind = 0;
-    impl->indScale = 10;
-    while ((impl->P > impl->Q ? impl->P : impl->Q)/impl->indScale > 0)
-      impl->indScale *= 10;
     impl->lookup = kh_init(m32);
     // Build all required kernels
     for (CeedInt nelem = 1; nelem <= 8; nelem+=7)
@@ -142,10 +140,9 @@ int CeedTensorContractCreate_Xsmm(CeedBasis basis,
                       J = grad ? impl->Q : (tmode ? impl->P : impl->Q),
                       C = nelem*CeedIntPow(J, dim);
               // Add key, ind pair to hash table
-              CeedInt key = ((B*impl->indScale + C)*impl->indScale + J)*10 +
-                            2*(tmode) + add;
+              CeedHashIJKLMKey key = {B, C, J, tmode, add};
               khint_t k = kh_get(m32, impl->lookup, key);
-              CeedInt keyMissing = (k == kh_end(impl->lookup));
+              bool keyMissing = CeedHashMissing(impl->lookup, k);
               // Skip if duplicate key
               if (keyMissing) {
                 CeedInt absent;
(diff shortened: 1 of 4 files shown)


backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -75,8 +75,10 @@ static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A,
   ierr = CeedTensorContractGetData(contract, (void *)&impl); CeedChk(ierr);
 
   // Get nelem and current dim
-  CeedScalar currdim = log(C/blksize) / log(J);
-  if (!(C % blksize) && currdim - (int)currdim < 1e-15) {
+  CeedScalar currdim = log(C/blksize);
+  if (fabs(log(J)) > CEED_EPSILON)
+    currdim /= log(J);
+  if (!(C % blksize) && currdim - (int)currdim < CEED_EPSILON) {
```
jedbrown, line 81: Surely this calculation can be done purely with integer arithmetic.

jeremylt: Maybe. I need to get `currdim` out of the 4 numbers `(ncomp*P**(dim-1-currdim), P, nelem*Q**currdim, Q)` without knowing what `ncomp` or `nelem` are.

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -73,22 +73,33 @@ static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A,
...
+      nelem = blksize;
+    } else {
+      nelem = 1;
+      isThisC = nelem;
+      currdim = 0;
+      while (isThisC < C && currdim < dim) {
+        currdim++;
+        isThisC *= J;
+      }
+    }
```
jedbrown, line 97: At a minimum, we need a comment to describe what this code is doing. But as you described it in the previous version (effectively 2 equations in 3 unknowns), it seems like this has non-unique solutions for at least some values of the inputs, and that means we should be more explicit instead of getting too clever here.

jeremylt: Context: The kernels need to be identified by `C`, `J`, `B`, `tmode`, and `add`. I convert `C` and `J` to `nelem` and `currdim` to make it easier to create an index of kernels. So long as the calculation of `nelem` and `currdim` maps to the same 5 values above, its ok if we find a different solution to those equations.

jedbrown: This seems like a mild maintenance liability that would vanish if we used a hash table or even just a sorted list of triplets.

jeremylt: I've never made a hash table before, but I figured there had to be a better way to do this. Wasn't sure if there would be any performance cost to a sorted list of triples vs a few calculations.

jedbrown: We use khash.h in PETSc. It's fast, header-only, and used internally by a number of other packages. https://github.com/attractivechaos/klib/blob/master/khash.h

jeremylt: Any preference between khash or a sorted list?

jeremylt: I added hkash. I've got some bizarre behavior from xsmm where I'm freeing an already freed kernel that I need to fix though.

jeremylt:
Alrighty. @jedbrown I think this is good to go now.

backends/xsmm/ceed-xsmm.h
```diff
@@ -17,12 +17,17 @@
 #include <ceed-backend.h>
 #include <libxsmm.h>
 #include <string.h>
+#include <khash.h>
 #include <math.h>
 
+// Instantiate khash structs and methods
+KHASH_MAP_INIT_INT(m32, char)
```
jedbrown, line 24: You can make a map from tuple (as a struct) to kernel, and thus get rid of a lot of bookkeeping that you're doing manually. You'd call `KHASH_INIT(hashx, KeyTuple, ValueType, 1, KeyHash, KeyEqual)` where you write `KeyHash` and `KeyEqual`. There are examples of such functions in PETSc, e.g., https://gitlab.com/petsc/petsc/-/blob/master/src/dm/impls/plex/plexinterpolate.c#L9-33

jeremylt: How about this? I largely adapted the PETSc macros (and called out the fact in the ceed-hash.h)

include/ceed-hash.h
```diff
@@ -0,0 +1,149 @@
...
+ */
+#define kh_reset(name, h) {                                     \
+        if (h) {                                                \
+                kfree((h)->keys); kfree((h)->flags);            \
+                kfree((h)->vals);                               \
+                memset((h), 0x00, sizeof(*(h)));                \
+        } }
+#endif /*kh_reset*/
+
+#if !defined(kh_foreach)
```
jedbrown, line 55: Why are these here? They're already defined in `khash.h`?

include/ceed-hash.h
```diff
@@ -0,0 +1,149 @@
...
+  CeedHashCombine(CeedHashCombine(CeedHashInt((key).i),CeedHashInt((key).j)), \
+                  CeedHashCombine(CeedHashInt((key).k),CeedHashInt((key).l))), \
+                  CeedHashInt((key).m))
+
+#define CeedHashIJKLMKeyEqual(k1,k2) \
+  (((k1).i==(k2).i) ? ((k1).j==(k2).j) ? ((k1).k==(k2).k) ? ((k1).l==(k2).l) ? \
+   ((k1).m==(k2).m) : 0 : 0 : 0 : 0)
+
+#define CeedHashIJKLMInit(name, value)										\
+	KHASH_INIT(name,CeedHashIJKLMKey,char,1,CeedHashIJKLMKeyHash,CeedHashIJKLMKeyEqual)
```
jedbrown, line 147:
```suggestion
#define CeedHashIJKLMInit(name, value)										\
	KHASH_INIT(name,CeedHashIJKLMKey,value,1,CeedHashIJKLMKeyHash,CeedHashIJKLMKeyEqual)
```

include/ceed-hash.h
```diff
@@ -0,0 +1,149 @@
...
+  return CeedHash_UInt32((CeedHash32_t)key);
+}
+
+static inline CeedHash_t CeedHashCombine(CeedHash_t seed, CeedHash_t hash) {
+  /* https://doi.org/10.1002/asi.10170 */
+  /* https://dl.acm.org/citation.cfm?id=759509 */
+  return seed ^ (hash + (seed << 6) + (seed >> 2));
+}
+
+#define CeedHashEqual(a,b) ((a) == (b))
```
jedbrown, line 133: Never used?

backends/xsmm/ceed-xsmm.h
```diff
@@ -15,14 +15,19 @@
 // testbed platforms, in support of the nation's exascale computing imperative.
 
 #include <ceed-backend.h>
+#include <ceed-hash.h>
 #include <libxsmm.h>
 #include <string.h>
 #include <math.h>
 
+// Instantiate khash structs and methods
+CeedHashIJKLMInit(m32, char)
```
jedbrown, line 24: Why map to char instead of to the kernel itself? Then you can avoid shuffling indexes.

jeremylt: updated

include/ceed-hash.h
jeremylt: updated

include/ceed-hash.h
jeremylt: I cut out everything we aren't currently using. No clue why they were in the PETSc version of this file.

include/ceed-hash.h
jeremylt: removed

valeriabarra: Ah, I had seen there were some unused definitions. I thought you put them for future reference. I wasn't sure indeed

jeremylt:
I really hope the OCCA revamp fixes this spurious ARM failure we keep getting on Travis.

jedbrown:
@jeremylt Please have a look at my last commit and let me know if you have concerns.

jeremylt:
I only had a `make style` issue on Travis. Makes sense to me. Thanks for the patience as I learned how to hash things. I'll merge after CI wraps up.

PR #532, opened 2020-05-05
Small GPU Test Changes

jeremylt:
This PR adds some small fixes from GPU testing.

We still have issues with `petsc-area` and `petsc-bpssphere` issue *only* with `/gpu/cuda/shared`.

### DIFF
--- a/examples/petsc/multigrid.c
+++ b/examples/petsc/multigrid.c
@@ -300,7 +300,7 @@ int main(int argc, char **argv) {
     ierr = VecCUDAGetArray(rhsloc, &r); CHKERRQ(ierr);
   }
   CeedVectorCreate(ceed, xlsize[fineLevel], &rhsceed);
-  CeedVectorSetArray(rhsceed, CEED_MEM_HOST, CEED_USE_POINTER, r);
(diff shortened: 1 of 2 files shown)


examples/petsc/bpsraw.c
```diff
@@ -35,7 +35,7 @@
 //     ./bpsraw -problem bp5 -ceed /omp/occa
 //     ./bpsraw -problem bp6 -ceed /ocl/occa
 //
-//TESTARGS -ceed {ceed_resource} -test -problem bp2 -degree 5 -qextra 5
+//TESTARGS -ceed {ceed_resource} -test -problem bp2 -degree 4 -qextra 5
```
jeremylt, line 38: The issue wasn't that `p == q` or `q > p` but that the problem was too large for `/gpu/cuda/ref`.

jeremylt: Dropping this drops an AVX case.

jeremylt: I'm dropping this commit for now since it appears that something deeper is going on with /cuda/ref and I don't want to reduce the coverage of the AVX backends.

PR #534, opened 2020-05-11
Solids - additional command line options

jeremylt:


examples/solids/elasticity.c
```diff
@@ -555,7 +555,8 @@ int main(int argc, char **argv) {
     ierr = SNESSolve(snes, F, U); CHKERRQ(ierr);
 
     // -- View solution
-    if (appCtx->viewSoln) {
+    if (appCtx->viewSoln ||
+        (increment == appCtx->numIncrements && appCtx->viewFinalSoln)) {
       ierr = ViewSolution(comm, U, increment, loadIncrement); CHKERRQ(ierr);
     }
```
jedbrown, line 561: If this options is turned on for performance runs, it would be better for the visualization to occur after the `PetscLogStagePop` (a few lines below). Conditional there would be `if (appCtx->viewFinalSoln && !appCtx->viewSoln)`.

jeremylt: Right, We'd made that change before but forgot about it. The joys of developing in multiple places.
Updated.

PR #536, opened 2020-05-14
Refactor Petsc area example

valeriabarra:
This PR refactors the `area` example to make it consistent with the other PETSc examples.

Major usage change:
* It now accepts the command line option `-degree`, as all other PETSc examples, instead of `-petscspace_degree`

Bug fixes:
* It eliminates a double call to `DMSetFromOptions()` which was causing an erroneous double refinement of the mesh

### DIFF
--- a/examples/petsc/area.c
+++ b/examples/petsc/area.c
@@ -29,15 +29,15 @@
 // Sample runs:
 //   Sequential:
 //
-//     ./area -problem cube -petscspace_degree 3 -dm_refine 2
+//     ./area -problem cube -degree 3 -dm_refine 2
 //
-//     ./area -problem sphere -petscspace_degree 3 -dm_refine 2
+//     ./area -problem sphere -degree 3 -dm_refine 2
 //
 //   In parallel:
 //
-//     mpiexec -n 4 ./area -probelm cube -petscspace_degree 3 -dm_refine 2
+//     mpiexec -n 4 ./area -problem cube -degree 3 -dm_refine 2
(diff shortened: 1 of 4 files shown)


examples/petsc/area.c
```diff
@@ -29,15 +29,15 @@
...
-//     ./area -problem cube -petscspace_degree 3 -dm_refine 2
+//     ./area -problem cube -degree 3 -dm_refine 2
 //
-//     ./area -problem sphere -petscspace_degree 3 -dm_refine 2
+//     ./area -problem sphere -degree 3 -dm_refine 2
 //
 //   In parallel:
 //
-//     mpiexec -n 4 ./area -probelm cube -petscspace_degree 3 -dm_refine 2
+//     mpiexec -n 4 ./area -probelm cube -degree 3 -dm_refine 2
```
jeremylt, line 38: small typo

valeriabarra: Thanks for catching that!

valeriabarra: Fixed

valeriabarra:
> Looks good to me. Eventually I want to go into the PETSc examples (and fluids) and separate these huge headers into source files, like in solids. We might be able to trim some common code.

Yes, I agree. Honestly I have always thought that our use of `.h` files in those examples was weird. As they are conventionally used only for function prototypes. 

I have used the conventional way also in the `valeria/shallow-water` branch, on which I was working on the side.

PR #538, opened 2020-05-14
Jed/petsc 3.14 gcc 10

jedbrown:
Add support for petsc-3.14 and fix Fortran bugs reported by gfortran-10.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ build/*
 *.DIR
 ceed.pc
 config.mk
+.prove
 
 # Mac specific
 .DS_Store
@@ -30,6 +31,8 @@ dist
 .pytest_cache
 *cffi.so
 *cffi.dylib
+__pycache__/
+.ipynb_checkpoints/
 
 # Documentation
 xml/*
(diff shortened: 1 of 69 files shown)


valeriabarra:
I think [travis](https://github.com/CEED/libCEED/blob/master/.travis.yml) will need to be updated as well

valeriabarra:
I don't see the [13.4 tag](https://gitlab.com/petsc/petsc/-/tags) though. Has it been released yet?

jedbrown:
I guess you're asking about 3.14, which isn't expected until September.

valeriabarra:
> I guess you're asking about 3.14, which isn't expected until September.

Lol! I was misreading.. haha yes... my low vision always bites me

valeriabarra:
There is a `make style` change in Travis which you can run to avoid, and a failure in `petsc-bpsraw /omp/occa` which happens very often but strangely not always. I saw it in #516 but not in #536

jedbrown:
Seems to just be the `bpsraw` with `/omp/occa` on non-x86 issue, which will hopefully be resolved for good in #305.

PR #539, opened 2020-05-15
benchmarks: Use PETSc unstructured bps

valeriabarra:
Here is a quick PR to be able to also use the unstructured PETSc bps in our benchmark script.

### DIFF
--- a/benchmarks/petsc-bps.sh
+++ b/benchmarks/petsc-bps.sh
@@ -37,10 +37,8 @@ function run_tests()
    for ((sol_p = 1; sol_p <= max_p; sol_p++)); do
       local loc_el=
       for ((loc_el = 1; loc_el*sol_p**3 <= max_loc_dofs; loc_el = 2*loc_el)); do
-         local loc_dofs=$((loc_el*sol_p**3))
-         local loc_cell=$(echo "e(l($loc_dofs)/3)" | bc -l)
-         loc_cell=${loc_cell%.*}
-         local all_args=("${common_args[@]}" -degree $sol_p -cells $loc_cell,$loc_cell,$loc_cell -problem $bp)
+         local loc_nodes=$((loc_el*sol_p**3))
+         local all_args=("${common_args[@]}" -degree $sol_p -local_nodes $loc_nodes -problem $bp)
          if [ -z "$dry_run" ]; then
             echo
             echo "Running test:"
--- a/benchmarks/petsc-bps.sh
+++ b/benchmarks/petsc-bps.sh
@@ -1,0 +1,57 @@
+# Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
+# Produced at the Lawrence Livermore National Laboratory. LLNL-CODE-734707.
+# All Rights reserved. See files LICENSE and NOTICE for details.
+#
+# This file is part of CEED, a collection of benchmarks, miniapps, software
+# libraries and APIs for efficient high-order finite element and spectral
+# element discretizations for exascale applications. For more information and
+# source code availability see http://github.com/ceed.
+#
+# The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+# a collaborative effort of two U.S. Department of Energy organizations (Office
+# of Science and the National Nuclear Security Administration) responsible for
+# the planning and preparation of a capable exascale ecosystem, including
+# software, applications, hardware, advanced system engineering and early
+# testbed platforms, in support of the nation's exascale computing imperative.
+
+
+function run_tests()
+{
+   $dry_run cd "$test_exe_dir"
+
+   # Some of the available options are:
+   # -degree <1>: Polynomial degree of tensor product basis
+   # -qextra <2>: Number of extra quadrature points
+   # -ceed </cpu/self>: CEED resource specifier
+   # -cells <10,10,10>: Target number of locally (per rank) owned mesh cells
+
+   # The variables 'max_dofs_node', and 'max_p' can be set on the command line
+   # invoking the 'benchmark.sh' script.
+   local ceed="${ceed:-/cpu/self}"
+   local common_args=(-ceed $ceed -pc_type none -benchmark)
+   local max_dofs_node_def=$((3*2**20))
+   local max_dofs_node=${max_dofs_node:-$max_dofs_node_def}
+   local max_loc_dofs=$((max_dofs_node/num_proc_node))
+   local max_p=${max_p:-8}
+   local sol_p=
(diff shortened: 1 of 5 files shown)


valeriabarra:
@jedbrown unfortunately for small number of cells, for which the partitioning creates empty entities, there is the problem in PETSc I already encountered earlier (see output file)
[petsc-bps-bp1-cpuself-output.txt](https://github.com/CEED/libCEED/files/4637342/petsc-bps-bp1-cpuself-output.txt)

benchmarks/petsc-bps.sh
```diff
@@ -0,0 +1,57 @@
...
+   local max_loc_dofs=$((max_dofs_node/num_proc_node))
+   local max_p=${max_p:-8}
+   local sol_p=
+   for ((sol_p = 1; sol_p <= max_p; sol_p++)); do
+      local loc_el=
+      for ((loc_el = 1; loc_el*sol_p**3 <= max_loc_dofs; loc_el = 2*loc_el)); do
+         local loc_dofs=$((loc_el*sol_p**3))
+         local loc_cell=$(echo "e(l($loc_dofs)/3)" | bc -l)
+         loc_cell=${loc_cell%.*}
+         local all_args=("${common_args[@]}" -degree $sol_p -cells $loc_cell,$loc_cell,$loc_cell -problem $bp)
```
jedbrown, line 43: `-cells` is global, but this looks like it's used locally. Perhaps just do this processing in the C executable? E.g., accept `-dofs_local` and compute a problem size using `MPI_Comm_size(comm, &size)` and `total_cells = dofs_local * size / (ncomp * CeedPowInt(degree, 3))`, then compute suitable box sizes?

jedbrown:
It's still a bug in ParMETIS (which has lots of bugs and is abandonware). You can use `-petscpartitioner_type simple` to get it running when you have more ranks than elements, but the quality will deteriorate as you scale. PETSc would accept a PR to bypass ParMETIS in the degenerate case.

benchmarks/petsc-bps.sh
valeriabarra: I have updated it using the same strategy that is in `bpsraw.c`

benchmarks/petsc-bps.sh
```diff
@@ -0,0 +1,55 @@
...
+   local ceed="${ceed:-/cpu/self}"
+   local common_args=(-ceed $ceed -pc_type none -benchmark)
+   local max_dofs_node_def=$((3*2**20))
+   local max_dofs_node=${max_dofs_node:-$max_dofs_node_def}
+   local max_loc_dofs=$((max_dofs_node/num_proc_node))
+   local max_p=${max_p:-8}
+   local sol_p=
+   for ((sol_p = 1; sol_p <= max_p; sol_p++)); do
+      local loc_el=
+      for ((loc_el = 1; loc_el*sol_p**3 <= max_loc_dofs; loc_el = 2*loc_el)); do
```
jedbrown, line 39: Seems like `max_loc_dofs` here is actually counting nodes. (Relevant for BP2 and BP4.)

benchmarks/petsc-bps.sh
```diff
@@ -0,0 +1,55 @@
...
+
+function run_tests()
+{
+   $dry_run cd "$test_exe_dir"
+
+   # Some of the available options are:
+   # -degree <1>: Polynomial degree of tensor product basis
+   # -qextra <2>: Number of extra quadrature points
+   # -ceed </cpu/self>: CEED resource specifier
+   # -cells <10,10,10>: Target number of locally (per rank) owned mesh cells
```
jedbrown, line 26: This line seems obsolete now.

benchmarks/petsc-bps.sh
valeriabarra: Should I rename it? I left its original name

valeriabarra: Renamed it

jedbrown:
I pushed a change to make `-local_nodes` actually refer to local sizes. We still have an inconsistency between bps.c reporting "Owned elements" while bpsraw.c reports "Local Elements". We should make that consistent. It appears that the latter is used by `postprocess-base.py`. (Our benchmark postprocessing scripts are quite weird. IMO, they should have callable functions that return dataframes, rather than using `exec` to create globals.

valeriabarra:
> I pushed a change to make `-local_nodes` actually refer to local sizes. We still have an inconsistency between bps.c reporting "Owned elements" while bpsraw.c reports "Local Elements". We should make that consistent. It appears that the latter is used by `postprocess-base.py`. (Our benchmark postprocessing scripts are quite weird. IMO, they should have callable functions that return dataframes, rather than using `exec` to create globals.

Okay, I've changed that to print "Local Elements". I had understood your suggestion about the `-local_nodes` but then I thought we wanted to be consistent with the way it's done in the `bpsraw.c`, do we?

jedbrown:
> but then I thought we wanted to be consistent with the way it's done in the bpsraw.c, do we?
In what sense? Is there a specific conflict?

valeriabarra:
This is ready for final review and merge or squash+merge if there are no other comments. Thanks!

valeriabarra:
Besides a small `make style` thing that I can take care of

benchmarks/petsc-bpsraw.sh
```diff
@@ -23,20 +23,20 @@ function run_tests()
    # -degree <1>: Polynomial degree of tensor product basis
    # -qextra <2>: Number of extra quadrature points
```
jedbrown, line 24: `-qextra 2` is not the default, FWIW.

examples/petsc/bps.c
```diff
@@ -141,6 +168,16 @@ int main(int argc, char **argv) {
     ierr = DMPlexCreateFromFile(PETSC_COMM_WORLD, filename, PETSC_TRUE, &dm);
     CHKERRQ(ierr);
   } else {
+    if (userlnodes) {
+      // Find a nicely composite number of elements no less than lnodes
```
jedbrown, line 172: `lnodes` -> `localnodes`?

valeriabarra: We have `localnodes`. This `userlnodes` (which refers to user local nodes for short) was just a flag to determine whether the user had set at CL option for nodes or not (otherwise the mesh is created using the global `-cells` parameter)

jedbrown: 'twas a poorly worded phone comment. There is no `lnodes` in the code and the comment isn't correct anyway. I'm happy with any wording that makes the comment accurate (or delete the comment).

valeriabarra: Ah, gotcha!

PR #540, opened 2020-05-16
WIP: Fluids - BCs

LeilaGhaffari:
This is a WIP PR for adding Outflow/Inflow Boundary Conditions to the fluids solver.
So far, I have only implemented outflow BC for all examples but there are still some problems with the restrictions. 
I am perplexed with the gradients of the components in the QFunction that is implementing the outflow BC for density current example (DC_Sur). I have defined an operator for this purpose but I don't think it is doing what I am expecting.


examples/fluids/navierstokes.c
```diff
@@ -1200,64 +1204,90 @@ int main(int argc, char **argv) {
...
+                         basisxSur, CEED_VECTOR_NONE);
+    CeedOperatorSetField(op_setupSur[i], "qdataSur", restrictqdiSur[i],
+                         CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
+
+    // Utility operator that builds the quadrature data for computing viscous terms
+    CeedOperatorCreate(ceed, qf_setupVol, NULL, NULL, &op_setupSur_[i]);
+    CeedOperatorSetField(op_setupSur_[i], "dx", restrictxSur[i], basisxVol, CEED_VECTOR_ACTIVE);
+    CeedOperatorSetField(op_setupSur_[i], "weight", CEED_ELEMRESTRICTION_NONE,
+                         basisxVol, CEED_VECTOR_NONE);
+    CeedOperatorSetField(op_setupSur_[i], "qdata", restrictqdiSur[i],
```
LeilaGhaffari: @jedbrown , I remember you were suggesting assuming the surface elements as 3d (in our 3d examples) and dropping the quadrature data on the surface. I was thinking of creating a separate restriction for the faces with `height = 0` for this block of code but the compiler complained (is it even a feasible think to do with `CreateRestrictionFromPlex()`?) so I decided to use the surface restrictions (which I don't think it's correct!).

knepley: Is there a Plex limitation?

LeilaGhaffari: @knepley , do you mean Plex limitation for creating 3d restrictions on the faces? I have to build the 3d restriction again to answer positively (right now I am having trouble building PETSc) but if I remember correctly, the limitation was in libCEED and the compatibility with the operators.

jedbrown:
@knepley This is an edge element.
```
$ gdb -ex r --args ./navierstokes -problem advection2d -degree 2 -ts_adapt_monitor -ts_max_time 6 -ksp_monitor -ts_view -petscspace_degree 2  -bc_outflow 1,2,3,4 -dm_plex_box_faces 20,20
[...]
(gdb) up
#1  0x0000555555575525 in CreateRestrictionFromPlex (ceed=0x5555558655a0, dm=0x55555579cd70, P=2, height=1, domainLabel=0x555555f72a50, value=1, Erestrict=0x7fffffffb120) at navierstokes.c:315
315                 SETERRQ4(PETSC_COMM_SELF,PETSC_ERR_ARG_INCOMP,
(gdb) p *indices @numindices 
$1 = {400, 0, 401, 0, 4094, 2000, 2001, -2003, -2004, 2004, 4095, 4096, 4097, -4099, 4099}
(gdb) call DMView(dm,0)
DM Object: 1 MPI processes
  type: plex
DM_0x84000000_0 in 2 dimensions:
  0-cells: 441
  1-cells: 840
  2-cells: 400
Labels:
  marker: 1 strata with value/size (1 (160))
  Face Sets: 4 strata with value/size (4 (41), 2 (41), 1 (41), 3 (41))
  depth: 3 strata with value/size (0 (441), 1 (840), 2 (400))
  celltype: 3 strata with value/size (4 (400), 0 (441), 1 (840))
Field Q:
  adjacency FEM
```

Evidently this is not in closure permutation. How can we make closure permutation apply to all strata? Also, can you let us know what is happening to the edge node (first five indices here)?

jedbrown:
Here are the indices with all the essential BCs deactivated. (@LeilaGhaffari We need to ensure that a given boundary is not both a slip and outflow condition.)
```
(gdb) p *indices @numindices 
$1 = {400, 0, 401, 0, 4094, 2000, 2001, 2002, 2003, 2004, 4095, 4096, 4097, 4098, 4099}
```
I would expect the first five entries to be `{400, 401, 402, 403, 404, ...}`, not jumping around like this.

knepley:
@jedbrown I have to understand what that means. Right now closure permutations are just that, permutations of a certain closure. There is no overarching structure. I think the easiest thing to do would be to have a closure permutation for each depth.

jedbrown:
What is the right way to ask for tensor permutation on all strata?

jedbrown:
@knepley The bug is here:
```
if (useClPerm) {ierr = PetscSectionGetClosureInversePermutation_Internal(section, (PetscObject) dm, NULL, &clperm);CHKERRQ(ierr);}
```
This function only works when every element is homogeneous and you never ask for a closure permutation of a different type of point. See also:
>  Note: This strategy only works when all cells have the same size dof closure, and no closures are retrieved for other points (like faces).

How would you like this fixed? Closure permutation indexed by `(depth, coneSize)`?

knepley:
Right now, it is not possible, since every point gets the same permutation. I think the thing to do is to assign a permutation by depth. It should not be hard to do.

jedbrown:
The issue is I don't want to have to refactor again to support mixed topology meshes or variable degree p. It seems like all the current call sites know the closure size at the time they request the permutation, so perhaps we can index by `(depth, clSize)`?

jedbrown:
I.e., change from the current
```c
PetscErrorCode PetscSectionGetClosureInversePermutation_Internal(PetscSection section, PetscObject obj, PetscInt *size, const PetscInt *perm[])
```
to
```c
PetscErrorCode PetscSectionGetClosureInversePermutation_Internal(PetscSection section, PetscObject obj, PetscInt depth, PetscInt size, const PetscInt *perm[])
```

knepley:
It sounds reasonable. Are you doing variable p with mortars?

jedbrown:
I'm primarily interested in conforming p-adaptivity.

examples/fluids/navierstokes.c
```diff
@@ -1298,7 +1297,7 @@ int main(int argc, char **argv) {
     CeedOperatorSetField(op, "x", restrictx, basisx, xcorners);
     CeedOperatorSetField(op, "v", restrictq, basisq, CEED_VECTOR_ACTIVE);
     CeedOperatorSetField(op, "dv", restrictq, basisq, CEED_VECTOR_ACTIVE);
-    user->op_ifunction = op;
+    user->op_ifunction_vol = op;
```
LeilaGhaffari, line 1300: @jedbrown, yes, this is what it should have been. Thank you for catching it.

examples/fluids/navierstokes.c
```diff
@@ -823,7 +823,6 @@ int main(int argc, char **argv) {
   PetscInt contsteps     = 0;        // -
   PetscInt degree        = 1;        // -
   PetscInt qextra        = 2;        // -
-  PetscInt degreeSur     = 1;        // -
```
LeilaGhaffari: @jedbrown , why do the surface and the volume need to have the same degree?

jedbrown: If you have a tensor Lagrange finite element, then the restriction to a face has the same degree. We're viewing the same set of dofs.

LeilaGhaffari: Very true. Thanks.

knepley:
> I'm primarily interested in conforming p-adaptivity.

Does that means imposing hanging-node-type constraints on faces between different p cells?

jedbrown:
@knepley Yes, which will require making anchors fast, including on GPU.

LeilaGhaffari:
@jedbrown, could you take a look at the refactoring? Thanks.

examples/fluids/navierstokes.c
```diff
@@ -294,6 +341,91 @@ static PetscErrorCode CreateRestrictionFromPlex(Ceed ceed, DM dm, CeedInt P,
...
+        CeedElemRestrictionGetNumElements(restrictqSur[i], &localNelemSur[i]);
+        CeedVectorCreate(ceed, qdatasizeSur*localNelemSur[i]*NqptsSur, &qdataSur[i]);
+
+        // Create the operator that builds the quadrature data for the NS operator
+        CeedOperatorCreate(ceed, qf_setupSur, NULL, NULL, &op_setupSur[i]);
+        CeedOperatorSetField(op_setupSur[i], "dx", restrictxSur[i], basisxSur, CEED_VECTOR_ACTIVE);
+        CeedOperatorSetField(op_setupSur[i], "weight", CEED_ELEMRESTRICTION_NONE,
+                             basisxSur, CEED_VECTOR_NONE);
+        CeedOperatorSetField(op_setupSur[i], "qdataSur", restrictqdiSur[i],
+                             CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
```
jedbrown, line 413: It looks like `op_setupSur` is never applied, and this is why the setup data is missing/zero when used in `Advection2d_Sur`.

LeilaGhaffari: Hmm! So, the problem that you previously mentioned still exists.
For applying composite operators, I had followed `t521`. Only, I didn't create any composite operators for the setup  which I think makes sense because the setup QFunctions are not communicating with each other. That is not the issue, or is it?

jeremylt: You don't need to make a composite setup operator (I mainly made a composite setup operator for t521 because I wanted to test composite operators with passive outputs). One way or another though, every setup operator needs to be run once. In the case of t521, I bundled them into a singe operator that was applied once, but if there is no reason to compose the operators, then they'll need to be applied individually, like in line 1271 of navierstokes.c on master.

LeilaGhaffari: Ah, I see now. Then, `op_setupSur` is truly never applied :D 
Thanks, @jedbrown and @jeremylt.

examples/fluids/navierstokes.c
```diff
@@ -526,6 +526,7 @@ static PetscErrorCode TSMonitor_NS(TS ts, PetscInt stepno, PetscReal time,
   ierr = PetscViewerVTKOpen(PetscObjectComm((PetscObject)Q), filepath,
                             FILE_MODE_WRITE, &viewer); CHKERRQ(ierr);
   ierr = VecView(Qloc, viewer); CHKERRQ(ierr);
+  ierr = PetscViewerDestroy(&viewer); CHKERRQ(ierr);
```
valeriabarra, line 596: This is already [in master](https://github.com/CEED/libCEED/blob/4ec104bbb9931294739dc47481cf59de6af2df1d/examples/fluids/navierstokes.c#L454). Wasn't this branch updated?

LeilaGhaffari: Hmm! Yes, it is updated but I wasn't able to rebase this file with master (especially after #516) so I had to resolve the conflicts manually. I might have missed this one (or maybe more!). I will compare it with master again to see if I have missed other changes or not. Thanks!

jedbrown: Would it help to squash/rebase this PR onto current 'master', or do you want to keep individual commits distinct.

LeilaGhaffari: But, I have been rebasing this branch onto `master` almost after each merge. Do you mean if I squash all changes and then rebase onto master, then I can catch other misses? 
I am not intending to keep all commits distinct but I thought it could be helpful to keep them while we are still reviewing.

jedbrown: I'd assumed you had some conflicts if changes from `master` are reproduced here, but in this case, I think @valeriabarra is mistaken; this line fixes a memory leak that still exists in 'master'.

valeriabarra: It is very possible that I am mistaken, but I see this [in master](https://github.com/CEED/libCEED/blob/master/examples/fluids/navierstokes.c) line 454, before the `if (user->dmviz) {` block. 

Just curious @LeilaGhaffari, were you manually integrating the changes or by using `git merge` or `git rebase`?

jedbrown: Line 454 is
```c
  ierr = VecView(Qloc, viewer); CHKERRQ(ierr);
```
That `viewer` isn't destroyed. The variable next appears on line 485:
```c
  ierr = PetscViewerBinaryOpen(user->comm, filepath, FILE_MODE_WRITE, &viewer);
```
which overwrites the pointer, thereby leaking the memory.

LeilaGhaffari: Yes, it is fixed in master.
@valeriabarra , I wasn't able to rebase it (I think because both of us had touched most parts of the file) so I thought I could skip this only file while I was rebasing and then, integrating the changes manually. I thought, this way, I could keep git history clean which apparently wasn't a good idea.

valeriabarra: S0rry, I was referring to the wrong line. I meant line 488 on master, where the viewer is destroyed after viewing the global vector, not the local one. 
Anyways, good that's fixed here! Sorry about the confusion!

LeilaGhaffari:
This PR is ready for review. 
I have implemented outflow (consistent) and inflow BCs on advection(2d) problem.
The original advection(2d) problem runs with `-problem_advection_wind rotation` which is the default option. 
`-problem_advection_wind uniform` invokes channel flow with the default background wind vector `(1,0,0)` while `-problem_advection_wind_uniform 1,.5,0` can change it to a diagonal wind.
`-P_left` and `-rho_left` are also for changing the value of incoming pressure and density, respectively.
@jedbrown , I see some instabilities/shocks in uniform wind. Do you have any suggestions for fixing it?

valeriabarra:
Thanks Leila for your work! 
Have we considered calling the `uniform` problem `laminar` instead? Technically unfirom could be also circular uniform with constant rotational velocity

jedbrown:
Laminar flow:
<p><a href="https://commons.wikimedia.org/wiki/File:Laminar_flow_profile.gif#/media/File:Laminar_flow_profile.gif"></a><br>By R Nave
Georgia State University - HyperPhysics
<a rel="nofollow" class="external free" href="http://hyperphysics.phy-astr.gsu.edu/hbase/pfric.html">http://hyperphysics.phy-astr.gsu.edu/hbase/pfric.html</a>, Public Domain, <a href="https://commons.wikimedia.org/w/index.php?curid=12134056">Link</a></p>

valeriabarra:
> Laminar flow:
> 
> [](https://commons.wikimedia.org/wiki/File:Laminar_flow_profile.gif#/media/File:Laminar_flow_profile.gif)
> By R Nave Georgia State University - HyperPhysics http://hyperphysics.phy-astr.gsu.edu/hbase/pfric.html, Public Domain, [Link](https://commons.wikimedia.org/w/index.php?curid=12134056)

I k ow laminar flow it's characterized by high diffusion and low convective momentum (which we could have in case we expand this to the full NS code instead of adection only). I was just trying to think of a possible other name that would not be ambiguous because uniform can mean anything, even uniform circular, with constant angular velocity.
I was just trying to think of another possible name. Maybe we can call one `parallel`and one `circular`?

jedbrown:
I just think "laminar" is far more ambiguous. Laminar flows can be pretty complicated. "Circular" could represent a vortex, not a simple rigid rotation. I think if you said a flow had "uniform velocity", most people would have the right idea. (Note that in any rotation, the velocity depends on location.)

valeriabarra:
https://en.wikipedia.org/wiki/Circular_motion#Uniform_circular_motion

jedbrown:
Yeah, though "circular" is doing a lot of work there. We can change terms, but this feels a bit like bikeshedding.

valeriabarra:
https://en.m.wikipedia.org/wiki/Linear_motion
Perhaps one linear or rectilinear and one circular or rotational? I'm just trying to help. 
Sorry, but I really think uniform is ambiguous as it can mean with just uniform, constant velocity (either rectilinear or circular).
To me, ensuring clarity and helping users with their overall experience with the app is not really bikeshedding.

jedbrown:
"linear" sounds like a linear gradient. Rigid rotation and rigid translation?

valeriabarra:
Okay, what does @LeilaGhaffari think about calling one case `ADVECTION_WIND_ROTATION` and one `ADVECTION_WIND_TRASLATION` ? This way, the CL options would become `-problem_advection_wind_rotation` `-problem_advection_wind_translation` respectively.

LeilaGhaffari:
> Okay, what does @LeilaGhaffari think about calling one case `ADVECTION_WIND_ROTATION` and one `ADVECTION_WIND_TRASLATION` ? This way, the CL options would become `-problem_advection_wind_rotation` `-problem_advection_wind_translation` respectively.

Actually, I don't have strong opinions about any of them, that is why I have been silent. 
However, I think `TRANSLATION` is a good choice.

LeilaGhaffari:
May I go ahead and change `uniform` -> `translation`?

jedbrown:
Sounds good to me.

examples/fluids/README.md
```diff
@@ -26,6 +26,7 @@ Available runtime options are:
 | `-stab`                  | Stabilization method                                                                            |
 | `-implicit`              | Use implicit time integartor formulation                                                        |
 | `-bc_wall`               | Use wall boundary conditions on this list of faces                                              |
+| `-bc_outflow`            | Use outflow boundary conditions on this list of faces                                              |
```
valeriabarra, line 29: Should also include `-bc_inflow` and the new command line options to run with the different wind settings

examples/fluids/navierstokes.c
```diff
@@ -727,6 +934,8 @@ int main(int argc, char **argv) {
   CeedScalar theta0      = 300.;     // K
   CeedScalar thetaC      = -15.;     // K
   CeedScalar P0          = 1.e5;     // Pa
+  CeedScalar P_left      = 1.5e5;    // Pa
+  CeedScalar rho_left    = 1.2;      // Kg/m^3
```
valeriabarra, line 938: Should we document why these new parameters are needed in the `fluids/index.rst` file with some mathematical formulation of the new boundary conditions?

valeriabarra:
I am testing this branch, but I'm not quite sure what is a sample run to test the new features. Should I just use 
```
./navierstokes -problem advection -problem_advection_wind_translation
```
without specifying any BCs? What is the default BC now? 
Does it work the same in 2D?

LeilaGhaffari:
> I am testing this branch, but I'm not quite sure what is a sample run to test the new features. Should I just use
> 
> ```
> ./navierstokes -problem advection -problem_advection_wind_translation
> ```
> 
> without specifying any BCs? What is the default BC now?
> Does it work the same in 2D?

@valeriabarra , the default option is `-problem_advection_wind rotation` with the same default BCs that is set up on master.
`-problem_advection_wind translation` will invoke channel flow.  The default wind (1,0,0) works well but I see some instabilities when I make it diagonal with `-problem_advection_wind_translation 1,.5,0`.

LeilaGhaffari:
Also, you can use the same command line arguments for both 2d and 3d advection.

examples/fluids/navierstokes.c
```diff
@@ -1153,6 +1173,47 @@ int main(int argc, char **argv) {
...
+      bc.inflow[0] = 6; bc.outflow[0] = 5;
+      // Faces 3 and 4 are slip
+      bc.nslip[0] = bc.nslip[2] = 0; bc.nslip[1] = 2;
+      bc.slips[1][0] = 3; bc.slips[1][1] = 4;
+      // Faces 1 and 2 are wall
+      bc.nwall = 2;
+      bc.walls[0] = 1; bc.walls[1] = 2;
+      break;
+    }
+  }
```
LeilaGhaffari: @valeriabarra , default BCs are here.

examples/fluids/navierstokes.c
```diff
@@ -965,6 +1007,17 @@ int main(int argc, char **argv) {
...
+                              "Use inflow boundary conditions on this list of faces",
+                              NULL, bc.inflow,
+                              (len = sizeof(bc.inflow) / sizeof(bc.inflow[0]),
+                              &len), &flg); CHKERRQ(ierr);
+    if (flg) {
+      bc.ninflow = len;
+      // Using inflow boundaries disables automatic wall/slip boundaries (they must be set explicitly)
+      bc.nwall = 0;
+      bc.nslip[0] = bc.nslip[1] = bc.nslip[2] = 0;
+    }
```
LeilaGhaffari, line 1030: I think I need to remove these options since inflow and outflow BCs are invoked automatically by calling `-problem_advection_wind translation`.

examples/fluids/navierstokes.c
LeilaGhaffari: So, you don't need to specify BCs.

examples/fluids/navierstokes.c
LeilaGhaffari: Yes, you are right. I will add them. Thanks.

examples/fluids/README.md
LeilaGhaffari: Well, I think I need to remove this but I need to add other command line options.

examples/fluids/navierstokes.c
jedbrown: If you make the flow diagonal, there will be two inflow faces and two outflow faces. Have you been doing that or just running with default parameters?

LeilaGhaffari: Ah, so that makes it strange. No, I haven't. I thought I was only changing the velocity vector. I will fix it. Thanks.

examples/fluids/advection.h
```diff
@@ -121,27 +124,36 @@ static inline int Exact_Advection(CeedInt dim, CeedScalar time,
...
+      q[4] = ((r <= rc)
+              && (y<center[1])) ? (1.-r/rc)*fmin(1.0,(center[1]-y)/1.25) : 0.;
+    } break;
+    }
+    case 1:
+      q[0] = 1.;
+      q[1] = wind[0];
+      q[2] = wind[1];
+      q[3] = wind[2];
+      q[4] = 0.;
```
LeilaGhaffari, line 155: @jedbrown , is this what you had in mind or do we need a bubble in the domain?

jedbrown: How about the same blobs as for rotation?

LeilaGhaffari: Yeah. For some reason I thought we would prefer an empty tunnel. In fact, I have found that presence of those blobs helps with spotting bugs and visualization.

examples/fluids/navierstokes.c
```diff
@@ -1165,20 +1165,50 @@ int main(int argc, char **argv) {
...
+          bc.outflow[0] = 1; bc.outflow[1] = 2;
+        break;} else if (wind[0] < 0 && wind[1] > 0) {
+          bc.ninflow = bc.noutflow = 2; // Faces 1 and 2 are inflow and Faces 3 and 4 are outflow
+          bc.inflow[0]  = 1; bc.inflow[1]  = 2;
+          bc.outflow[0] = 3; bc.outflow[1] = 4;
+        break;} else if (wind[0] < 0 && wind[1] < 0) {
+          bc.ninflow = bc.noutflow = 2; // Faces 2 and 3 are inflow and Faces 1 and 4 are outflow
+          bc.inflow[0]  = 2; bc.inflow[1]  = 3;
+          bc.outflow[0] = 1; bc.outflow[1] = 4;
+        break;}
```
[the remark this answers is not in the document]
LeilaGhaffari: Yeah, it could be even cumbersome to implement it on our 3d problems. I thought we were only concerned about this mesh. But, of course, considering a more general case is wise.

we could keep only one QFunction for both `inflow` and `outflow` and after computing `sign(dot(velocity, normal))`, let a `switch-case` decide which operation is required. My only concern is how to update  `SetUpDM` when `dot(velocity, normal) = 0`. Is it possible to update a struct inside the QFunction?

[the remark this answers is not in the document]
jedbrown: For advection, it doesn't matter because flux is multiplied by velocity, so if the velocity is zero, the flux (integrand) is zero. For Euler/NS, we have only the incoming/outgoing acoustic wave, but we normally identify inflow vs outflow anyway. Zero velocity can happen in low-speed recirculating flows, but we'd have to think through the details and the viscous contribution is likely significant.

So yes, I would go ahead with a single in/outflow QFunction.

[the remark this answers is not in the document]
LeilaGhaffari: @jedbrown , could you please take a look at my latest changes? Is this what you meant? I think I am missing something because I can't reproduce the correct behavior of the blob with `translation` mode. I have been trying to find the bug for hours now but I can't see what goes wrong.

examples/fluids/advection2d.h
```diff
@@ -381,7 +410,116 @@ CEED_QFUNCTION(IFunction_Advection2d)(void *ctx, CeedInt Q,
...
+    for (CeedInt j=0; j<4; j++) {
+      v[j][i] = 0;
+    }
+    // Implementing in/outflow BCs
+    switch (implicit) {
+    case 0: {  //explicit
+      if (windDotNorm >  1E-5) { // outflow
+        v[4][i] = -(1-strong_form) *wdetJb *E *u_n;
+      } else if (windDotNorm < -1E-5) { // inflow
+        v[4][i] = (1-strong_form) *wdetJb *E_star *u_star;
```
jedbrown, line 505: I would expect the inflow `E_star` to be whatever inflow value is prescribed. With this, we get a wave coming in from the left.
```suggestion
        v[4][i] = (1-strong_form) *wdetJb * 1.0 *u_star;
```
I ran with
```
 ./navierstokes -problem advection2d -problem_advection_wind translation -problem_advection_wind_translation 1,0,0 -ts_max_time 50 -dm_plex_box_faces 32,32 -ts_adapt_monitor
```

LeilaGhaffari: Thanks, @jedbrown . Yep, It is working, even in 3D.
I am very surprised though. This formulation used to work before this latest change in the setup. Could it be for the fact that we are not implementing slip/wall BCs anymore?!

Well, since it was working previously, I never doubted the correctness or compatibility of the inflow formulations with our problem. But, now I see that the value of `E_star` is negative because the initial `E` is Zero. I have updated `index.rst`, if that is easier to read. Could it be fundamentally wrong?

Currently, `P_wind` and `rho_wind` are prescribed by the user and `E_star` is computed accordingly. Do you think I should go ahead and prescribe `E_star` instead or do you have other suggestions?

valeriabarra:
This looks great. Except, that I ran this with

```./navierstokes -problem advection2d -problem_advection_wind translation -problem_advection_wind translation 0,-1 -degree 1 -dm_plex_box_faces 20,20 -lx 1000 -ly 1000  -rc 200 -viz_refine 1 -ts_max_time 10
```
So the wind should have gone downward, towards the negative y-axis. No?
Translation

valeriabarra:
Same question for a 3D run. And also, in 2D the default IC is a spherical 2D-ball, which I think it's perfectly fine. But in 3D is the half cilinder, and it's sideways relative to the wind direction, so perhaps it's not the best start due to the sharp interface. I think it would be good to default to spherical bubble (3D-ball) for the 3D case, too

```
./navierstokes -problem advection -problem_advection_wind translation -problem_advection_wind translation 0,0,-1 -degree 1 -dm_plex_box_faces 20,20,20 -lx 1000 -ly 1000 -lz 1000  -rc 200 -viz_refine 1 -ts_max_time 10
```
Transaltion3D

LeilaGhaffari:
Thanks, @valeriabarra  for your comments. I agree with you about changing the default IC to spherical bubble. Regarding the direction of the wind, I need to do some corrections. I will let you know when I update it.

valeriabarra:
Great, @LeilaGhaffari thanks! Let me know if you need help with that.

Also, just a general comment about the workflow for merging this PR. My recollection was that this PR is dependent on this [MR](https://gitlab.com/petsc/petsc/-/merge_requests/2802) in the PETSc's repo, which I see is not merged yet. @jedbrown is that so?

LeilaGhaffari:
> This looks great. Except, that I ran this with
> 
> ```
> ./navierstokes -problem advection2d -problem_advection_wind translation -problem_advection_wind translation 0,-1 -degree 1 -dm_plex_box_faces 20,20 -lx 1000 -ly 1000  -rc 200 -viz_refine 1 -ts_max_time 10
> ```
> 
> So the wind should have gone downward, towards the negative y-axis. No?
> Translation

Oh, you have used `-problem_advection_wind translation 0,-1` instead of `-problem_advection_wind_translation 0,-1`. That is why the default velocity is applied.

valeriabarra:
Oooh my bad! Thanks for spotting that typo. I will run it again

LeilaGhaffari:
> Great, @LeilaGhaffari thanks! Let me know if you need help with that.
> 
> Also, just a general comment about the workflow for merging this PR. My recollection was that this PR is dependent on this [MR](https://gitlab.com/petsc/petsc/-/merge_requests/2802) in the PETSc's repo, which I see is not merged yet. @jedbrown is that so?

Yes, that is true. Thank you for reminding.

examples/fluids/index.rst
```diff
@@ -272,16 +272,41 @@ of total energy, is given by
...
-The :math:`3D` version of this problem can be run with::
+blob of total energy (defined by a characteristic radius :math:`r_c`) is transported by two
+different wind types.
 
-   ./navierstokes -problem advection
+- **Rotation**
 
-while the :math:`2D` version with::
+   In this case, a uniform circular velocity field transports the blob of total energy. We have
+   solved :math:numref:`eq-advection` with zero energy density :math:`E`, and no-flux for :math:`\bm{u}`.
```
valeriabarra, line 281: When you say "with zero energy density" here, do you mean for the BCs?

examples/fluids/index.rst
```diff
@@ -272,16 +272,41 @@ of total energy, is given by
...
+      ./navierstokes -problem advection -problem_advection_wind rotation
+
+   while the :math:`2D` version with::
+
+      ./navierstokes -problem advection2d -problem_advection_wind rotation
+
+- **Translation**
+
+   In this case, a background wind with a constant velocity field, enters the domain and compels
+   the blob of total energy to leave the domain.
```
valeriabarra, line 294:
```suggestion
   In this case, a background wind with a constant rectilinear velocity field, enters the domain and transports
   the blob of total energy out of the domain.
```

examples/fluids/index.rst
```diff
@@ -272,16 +272,41 @@ of total energy, is given by
...
+
+   For the inflow boundary conditions, a prescribed :math:`E_wind` is applied weakly on the inflow boundaries.
```
valeriabarra, line 296:
```suggestion
   For the inflow boundary conditions, a prescribed :math:`E_{wind}` is applied weakly on the inflow boundaries.
```

examples/fluids/index.rst
LeilaGhaffari: Hmm! I believe so! Actually, I didn't write this part. It was already there. Maybe adding `on the boundaries` or `for BCs`  at the end of the sentence?

examples/fluids/index.rst
LeilaGhaffari: Ouch! Thank you for catching this.

examples/fluids/index.rst
valeriabarra: I suggested `rectilinear` just to be extra clear, since constant alone does not specify this

LeilaGhaffari: Oh, sorry I missed it.

examples/fluids/index.rst
```diff
@@ -293,27 +293,12 @@ different wind types.
...
-   the domain, and incoming wind, respectively. The computed variables :math:`P_*` and :math:`u_*` and,
-   a prescribed :math:`E_L` are applied weakly on the inflow boundaries.
-
-   For the outflow boundary conditions, we have
-   applied values from previous iteration, following :cite:`papanastasiou1992outflow` which extends the
-   validity of the weak form of the governing equations to the synthetic outflow instead of replacing them
-   with unknown essential or natural boundary conditions.
+   For the inflow boundary conditions, a prescribed :math:`E_wind` is applied weakly on the inflow boundaries.
+
+   For the outflow boundary conditions, we have applied values from previous iteration, following
```
valeriabarra, line 303: Thanks Leila for adding this! It looks great! I'm having a bit of a hard time to grasp this sentence, where we refer to previous interations to calculate a spatial flux. Is this what this sentence means?

```suggestion
   In order to calculate :math:`\bm{F}(\bm q_N^{n+1})`, for the outflow boundary condition, we have used values of :math:`E` from the previous time iteration, namely  :math:`E^n`, following
```

jedbrown: Unless I'm missing context, I don't think it's true that we use values from the previous time step.

LeilaGhaffari: Sorry for the confusion.
I guess I wrote this because I was only thinking of explicit time stepping. 

How about: 
"For the outflow boundary conditions, we have used the current values of `E`, following"
?

jedbrown: I think it's all consistent with the semi-discrete form `g(t, q, qdot) = 0` or `qdot = f(t, q)` in that all state variables appear at the same instant. (Stages in the time integrator correspond to quadrature points in time.) So we're quite literally performing the integral that appears in the weak form.

examples/fluids/densitycurrent.h
```diff
@@ -53,7 +53,9 @@ struct SetupContext_ {
   CeedScalar periodicity2;
   CeedScalar center[3];
   CeedScalar dc_axis[3];
+  CeedScalar wind[3];
```
valeriabarra, line 56: I know it's not relevant to this PR, but should these struct variables 
```
  CeedScalar periodicity0;
  CeedScalar periodicity1;
  CeedScalar periodicity2;
```
be removed now that we currently not support periodic BCs in this code? They are never used in neither `densitycurrent.h`, `advection2d.h` nor `advection.h`

jedbrown: Yeah, we should delete unused variables like these. Periodicity will be supported at the mesh level, hopefully with no changes necessary in this example.

LeilaGhaffari: Done. Thanks, @valeriabarra for catching it.

examples/fluids/advection.h
```diff
@@ -59,6 +58,16 @@ struct DCContext_ {
 };
 #endif
 
+#ifndef surface_context_struct
+#define surface_context_struct
+typedef struct SurfaceContext_ *SurfaceContext;
+struct SurfaceContext_ {
+  CeedScalar E_wind;
+  CeedScalar strong_form;
+  PetscBool implicit;
```
valeriabarra, line 67:
```suggestion
  bool implicit;
```
It's cleaner not to use Petsc's typedefs in QFunction code, which should be separate and independent from the user-code infrastructure

examples/fluids/advection.h
```diff
@@ -437,6 +458,88 @@ CEED_QFUNCTION(IFunction_Advection)(void *ctx, CeedInt Q,
...
+// This QFunction implements consistent outflow and inflow BCs
+//      for 3D advection
+//
+//  Inflow and outflow faces are determined based on sign(dot(wind, normal)):
+//    sign(dot(wind, normal)) > 0 : outflow BCs
+//    sign(dot(wind, normal)) < 0 : inflow BCs
+//
+//  Outflow BCs:
+//    The validity of the weak form of the governing equations is extended
+//    to the outflow and the values obtained from the previous iteration
```
valeriabarra, line 471: I think it would be better if `from the previous iteration` could be updated to match the description given in the user manual

valeriabarra: Also, I think we need to update the description at the beginning of this file, lines 83-89:
```
//  Boundary Conditions:
//    Mass Density:
//      0.0 flux
//    Momentum Density:
//      0.0
//    Energy Density:
//      0.0 flux
```
with whichever default BCs we have now, and maybe mention the optional ones (not default ones) as well

valeriabarra: The same comment/description appears on lines 87-93 of `advection2d.h`

examples/fluids/advection.h
LeilaGhaffari: Done. Thanks.

examples/fluids/advection.h
LeilaGhaffari: Oh, it completely slipped my mind to fix these too. Thanks.

examples/fluids/setup-boundary.h
```diff
@@ -0,0 +1,153 @@
...
+#ifndef setup_boundary_h
+#define setup_boundary_h
+
+#ifndef __CUDACC__
+#  include <math.h>
+#endif
+
+// *****************************************************************************
+// This QFunction sets up the geometric factor required for integration when
+//   reference coordinates is 2D and the physical coordinates is 3D
```
valeriabarra, line 29:
```suggestion
//   reference coordinates are in 2D and the physical coordinates are in 3D
```

examples/fluids/advection.h
```diff
@@ -121,10 +131,21 @@ static inline int Exact_Advection(CeedInt dim, CeedScalar time,
...
+    break;
+  case 1:    // Translation
+    q[0] = 1.;
+    q[1] = wind[0];
+    q[2] = wind[1];
+    q[3] = wind[2];
+    break;
+  }
+
   CeedInt continuityBubble = -1;
```
valeriabarra, line 149: I think it would be better to have `continuityBubble = 0;` as default.

examples/fluids/setup-boundary.h
```diff
@@ -55,24 +55,31 @@
...
+      },
+      {
+        J[0][1][i],
+        J[1][1][i]
+      },
+      {
+        J[0][2][i],
+        J[1][2][i]
+      }
+    };
```
LeilaGhaffari, line 82: Does this seem fine?

jeremylt: That sort of stupidness from `make style` can be prevented by using `// *INDENT-OFF*` and `// *INDENT-ON*`.

LeilaGhaffari: Ah, okay. Thanks.

valeriabarra: Thanks @LeilaGhaffari  for fixing that!

valeriabarra:
> Thanks, @valeriabarra for fixing these. Did you fix them with `make style`?

No, sadly `make style` does not always catch all things consistently. I modified these by hand and the result is still `make style`-clean

LeilaGhaffari:
> > Thanks, @valeriabarra for fixing these. Did you fix them with `make style`?
> 
> No, sadly `make style` does not always catch all things consistently. I modified these by hand and the result is still `make style`-clean

Hmm! Then, I need to learn more about our style rules to be more careful about them from now on.

valeriabarra:
> > > Thanks, @valeriabarra for fixing these. Did you fix them with `make style`?
> > 
> > 
> > No, sadly `make style` does not always catch all things consistently. I modified these by hand and the result is still `make style`-clean
> 
> Hmm! Then, I need to learn more about our style rules to be more careful about them from now on.

I wish we had written a style guide to help you with that so that it could be easily accessed and checked as a reference. Generally, we tend to not exceed 80 characters per line, except when this would break strings.

LeilaGhaffari:
I see. Thanks. This really helps.

examples/fluids/advection.h
```diff
@@ -59,25 +58,59 @@ struct DCContext_ {
...
+//        Mass Density:
+//          0.0 flux
+//        Momentum Density:
+//          0.0
+//        Energy Density:
+//          Inflow BCs:
+//            E = 1E6 is applied weakly in QFunction "Advection2d_Sur"
+//          Outflow BCs:
+//            The current values of E are applied weakly in
+//              QFunction "Advection2d_Sur"
```
valeriabarra, line 113: Thanks @LeilaGhaffari Leila for updating these descriptions. What do you think of the following suggestion? Not sure if the outflow one is correct this way. But I thought the description of the QFunction could have been combined for both and put afterwards.
```suggestion
//          Inflow BCs:
//            E = E_wind
//          Outflow BCs:
//            E = E(boundary)
//           Both In/Outflow BCs for E are applied weakly in the QFunction "Advection2d_Sur"
```

valeriabarra: Same thing would apply to `advection_2d.h`.

Overall everything works and looks good to me. Great job!

LeilaGhaffari: @valeriabarra , it is correct and much cleaner. Thanks. It's fixed.

LeilaGhaffari:
I think I have addressed all the comments. I have one question though. Do we want to keep [this commit](https://github.com/CEED/libCEED/commit/60adbd4b95dc120bf6b0aede89646e993ea8d65c)?

PR #542, opened 2020-05-19
CUDA - fix empty restriction support

jeremylt:
I did an incomplete fix of empty restriction support. This should be the last of it.

@valeriabarra, this branch works fine locally with the benchmarks script.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -1003,15 +1003,18 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
     if (emode != CEED_EVAL_WEIGHT &&
         !((emode == CEED_EVAL_NONE) && basis_data->d_collograd1d)) {
       code << "  CeedScalar r_u"<<i<<"[ncomp_in_"<<i<<"*P_in_"<<i<<"];\n";
-      ierr = CeedElemRestrictionGetData(Erestrict, (void **)&restr_data); CeedChk(ierr);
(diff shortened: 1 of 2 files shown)


jeremylt:
I'll merge when CI finishes.

valeriabarra:
Thanks @jeremylt . I'm testing this branch on OLCF Summit, but the `bps` example is still failing. If it works for you then something must be wrong on my environment, even though I do't know what it might be right now, since I have done everything as usual to compile the library, the examples and run the `benchmark.sh` script.

jeremylt:
Well, could be a completely different bug. I'm testing with `make BACKENDS=/gpu/cuda/gen  benchmarks`

jeremylt:
Yeah, locally I'm seeing out of memory errors for our standard script test sizes for the non /gpu/cuda/gen backeds, which is error code 134.

valeriabarra:
This is the (messy) output of the test run on OLCF Summit
[petsc-bps-bp1-gpucudagen-output.txt](https://github.com/CEED/libCEED/files/4653751/petsc-bps-bp1-gpucudagen-output.txt)

valeriabarra:
But I was testing the `/gpu/cuda/gen` backend

jeremylt:
It's a little strange that the runs are on top of each other. We should only have one test running at at time. Is there corresponding output from the bash?

valeriabarra:
Yes, that looked off. Hold on a sec. I think I see a clearer error in the batch output file

valeriabarra:
Fixed it. I am able to run fine with the` /gpu/cuda/gen` backend now on OLCF Summit

PR #543, opened 2020-05-19
Docs - pin python package markdown at 3.2.1

jeremylt:
@jedbrown, can you verify that this fixes the issue on readthedocs? Locally, I have v3.0.1, so I believe that you're correct and this is the problem.

### DIFF
--- a/doc/sphinx/requirements.txt
+++ b/doc/sphinx/requirements.txt
@@ -1,5 +1,6 @@
 sphinx>=3.0
 breathe>=4.17
+markdown==3.2.1
 recommonmark
 sphinx_rtd_theme
 sphinxcontrib-bibtex


doc/sphinx/requirements.txt
```diff
@@ -1,5 +1,6 @@
 sphinx>=3.0
 breathe>=4.17
+markdown==3.0.1
```
valeriabarra, line 3: Was `3.2.1` the one that passed?

jeremylt: Good catch, thanks. 3.0.1 is what I had locally. 3.2.1 is working locally as well.

jedbrown:
Building here: https://readthedocs.org/projects/libceed/builds/11077011/

jeremylt:
@jedbrown thanks. It looks like readthedocs is happy. I'll merge when CI is finished.

PR #544, opened 2020-05-20
Python - add support for getting/setting CUDA arrays

jeremylt:
This adds the ability to use device memory in the Python interface. As other Python packages add bi-directional [CUDA array interface](https://numba.pydata.org/numba-doc/latest/cuda/cuda_array_interface.html) support, we can modify this, but for now it's based upon Numba DeviceNDArray objects. We will also have to think carefully if we want to extend support to AMD devices, but Numba ect aren't quite there yet with an HIP array interface.

I think this is a nice add-on for the SciPy paper that lets us say we can provide integration with the family of modules that fully support Numba's CUDA array interface.


python/ceed_vector.py
```diff
@@ -17,6 +17,7 @@
 from _ceed_cffi import ffi, lib
 import tempfile
 import numpy as np
+import numba.cuda as nbcuda
```
jedbrown, line 20: Can we make this optional, such that one need not have Numba (or CUDA) to use libCEED, but if they use Numba, it will work?

jeremylt: sure, I'll look into that

valeriabarra:
Thanks Jeremy! We could also think if/how we want to add/extend the tutorials to include one usage example

tests/python/test-1-vector.py
```diff
@@ -198,3 +212,19 @@ def test_108(ceed_resource, capsys):
     assert abs(norm - 9.) < 1e-14
 
 # -------------------------------------------------------------------------------
+# Test modification of reshaped array
+# -------------------------------------------------------------------------------
+
+
+def test_199(ceed_resource):
```
valeriabarra, line 219: Was this supposed to be following the sequential local numbering here, i.e., `109` and it has a typo, or you decided that it won't follow the existing format because it does not reflect a C test?

jeremylt: Since this does not correspond to a C test and it was numbered the same as the new test that does correspond to a C test, I renamed it.

python/ceed_vector.py
jeremylt: How about the adjustment I made?

tests/python/test-1-vector.py
valeriabarra: Yeah that's what I thought. Makes sense, thanks

python/ceed_vector.py
```diff
@@ -106,13 +115,26 @@ def get_array(self, memtype=MEM_HOST):
...
+        if memtype == MEM_HOST:
+            # Create buffer object from returned pointer
+            buff = ffi.buffer(
+                array_pointer[0],
+                ffi.sizeof("CeedScalar") *
+                length_pointer[0])
+            # return Numpy array
+            return np.frombuffer(buff, dtype="float64")
+        else:
+            # CUDA array interface
```
jedbrown, line 128: Maybe just import here? The error message would be more transparent than `NameError: name 'nbcuda' is not defined`.

jeremylt: agreed

jeremylt:
I fixed a testing bug. Should be good to go now.

valeriabarra:
I am not able to test 
```
ceed = libceed.Ceed('/gpu/cuda/gen')
```
on our Binder. 

Do we want to mention any of this in the vector's class tutorial?

jedbrown:
Does it work on https://colab.research.google.com ?

valeriabarra:
> Does it work on https://colab.research.google.com ?

Nope, it crashes in there too

valeriabarra:
```
/tmp/pip-req-build-qakpy6hp/backends/avx/ceed-avx-blocked.c:27 in CeedInit_Avx(): AVX backend cannot use resource: /gpu/cuda/gen
```

jedbrown:
Did you do Runtime -> Change runtime type -> add GPU?

valeriabarra:
Ah, I din't know about that. But even after doing that, it still crashes, with same message

jeremylt:
It looks like the version of libCEED in that notebook was not built with GPU backends.

valeriabarra:
I have tried to test this also on OLCF Summit, where I have the build with GPU backends:

```
[valeria@login5.summit libCEED]$ python          
Python 3.7.0 (default, Oct 12 2018, 15:17:16) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-28)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import libceed
>>> ceed = libceed.Ceed()
>>> ceed = libceed.Ceed('gpu/cuda/gen')
/gpfs/alpine/csc262/proj-shared/valeria/CEED/libCEED/interface/ceed.c:623 in CeedInit(): No suitable backend: gpu/cuda/gen
Aborted (core dumped)
```

valeriabarra:
should I have specified something in the python build of libCEED? I have simply used ` pip install libceed --user `

jeremylt:
This
```
>>> ceed = libceed.Ceed('gpu/cuda/gen')
```
should be
```
>>> ceed = libceed.Ceed('/gpu/cuda/gen')
```

valeriabarra:
Ah my bad! I missed the leading slash. Thanks

tests/python/test-1-vector.py
```diff
@@ -136,17 +136,31 @@ def test_104(ceed_resource):
...
+# -------------------------------------------------------------------------------
+# Test creation, setting, reading, restoring, and destroying of a vector using
+#   CEED_MEM_DEVICE
+# -------------------------------------------------------------------------------
+
 
 def test_105(ceed_resource):
-    """Modification of reshaped array"""
-    ceed = libceed.Ceed(ceed_resource)
+    # Skip test for non-GPU backend
```
valeriabarra, line 146: @jeremylt  I'm not able to test this even on Summit, but if you are and it works fine for you, what would you think to add it to the vector class' tutorial, specifying that that would work only if one has built libCEED with GPU support?

jedbrown:
Seems like `pip install libceed` would install the binary wheel from PyPI, which won't have CUDA support.

valeriabarra:
For me now when I tested it on OLCF Summit, the libCEED install phase went OK, so I was able to test `ceed = libceed.Ceed('/gpu/cuda/gen')`,  but is the installation of the numba package that I can't seem to get to work. So I can't have `test_105` work now.

tests/python/test-1-vector.py
jeremylt: Sometimes installing Numba with CUDA support can be touchy. I can work on adding that. Would we want to add it in this PR or separately after we update PyPi?

jedbrown: Seems out of scope for libCEED itself, unless you're talking about CI.

jeremylt: Bad spacing/layout on my part. Two thoughts.
1) Numba is touchy - explaining why there might be issues in @valeriabarra's Summit env
2) Do we want to add it in this PR - referencing adding GPU vectors to the tutorial

jedbrown: 2. It should be in some tutorial by the time we submit to SciPy.

jeremylt: Sure. And I know the cell I want to add. I just wasn't sure if anyone cared about if it would be before or after updating PyPi.

jeremylt: Added. We can cherry pick the commit if we want to wait until after PyPi is updated.

valeriabarra: Thanks Jeremy!

valeriabarra: What is it needed for the PyPi update?

jedbrown: I've done it in the past. Run `make wheel` to build the wheel in a docker container. I can give you access to upload to PyPI, but it does not allow any modifications after upload so you need to be confident it's the way you want. There is test.pypi.org that can be used to check.

valeriabarra: I'm not sure I could be extremely confident on which way would be the way we want the wheel to be uploaded. Not to mention I've never even created a docker container myself. Sorry. I wanted to help, but perhaps it is better and faster if I don't embark in this.

jeremylt:
> How much work would it be to support qfunctions in Numba (even though they won't work with all backends)?

I missed this. I'm not quite sure how long it will take. On the libCEED side we need to make a /gpu/cuda/python backend that accepts a pointer to a CUDA kernel. Shouldn't be a problem. On the Python side we need to figure out the best way to make that kernel. Unknown time required. Thoug,  figuring it out will probably also give us a path to do the same thing for the CPU.

jedbrown:
I'd hope this doesn't need to be a special backend, but really any unfused backend. And it wouldn't be Python-specific, just that we only have a function pointer (which need not _be_ a CUDA kernel; it can accept device pointers and launch its kernel internally). If it's going to reside in a shared object, it could just be a string to be used with `dlsym()`, otherwise it'd be a new argument or new function to set that supplementary information. For CPU kernels, I think the existing qfunction interface is sufficient.

jeremylt:
All of the CUDA backends only use the QFunction source and ignore passed in function pointers.

jedbrown:
My thought was that fused backends (like `cuda-gen`) would error and the others (when provided with the supplementary function pointer and no source from which to generate their version) would call it instead of `CeedRunKernelCuda` in `CeedQFunctionApply_Cuda`.

jeremylt:
Ah. That would probably be about the same amount of work as making a new backend but would be easier to use. I'm less sure of the volume of work required if we aren't handing in a kernel.

jedbrown:
I think the work is in defining and documenting the API. The function the user provides should do the same thing ours does: launch a kernel and do the same operations on the provided data. I suspect interop (CUDA/HIP/SYCL and from packages like JAX) if we were to insist on being handed a raw kernel.

jeremylt:
> I think the work is in defining and documenting the API. The function the user provides should do the same thing ours does: launch a kernel and do the same operations on the provided data. I suspect interop (CUDA/HIP/SYCL and from packages like JAX) if we were to insist on being handed a raw kernel.

Since this will take a while, do we want to go ahead and merge this branch as is? It extends our GPU Python support and users can still use the singe source or gallery QFs like we have on the CPU side for Python.

examples/tutorials/tutorial-1-vector.ipynb
```diff
@@ -200,6 +200,31 @@
     "norm_max = x.norm(normtype=libceed.NORM_MAX)\n",
     "print(norm_max)"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "* If you have installed libCEED with CUDA support and Numba, you can use device memory in your `CeedVector`s. In the following example, we create a `CeedVector` with a libCEED contex that supports CUDA, associate the data stored in a CeedVector with a `numpy.array`, and get a Numba `DeviceNDArray` containing the data on the device."
```
jedbrown, line 208: I see there was prior use of this type, but `CeedVector` doesn't have specific meaning to a Python user, which may think of `Ceed.Vector` or `libceed.Vector`.

optional-requirements.txt
```diff
@@ -0,0 +1 @@
+numba
```
jedbrown, line 1: Probably `requirements-optional.txt` or `requirements-gpu.txt`?

jeremylt: Changed.

examples/tutorials/tutorial-1-vector.ipynb
jeremylt: Changed across all notebooks.

PR #546, opened 2020-05-21
Add ICC to CI

jeremylt:
I added ICC, modifying the example here: https://github.com/rscohn2/oneapi-ci/blob/master/.gitlab-ci.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: c
 
+os: linux
+dist: bionic
+
 addons:
   apt:
     update: true
@@ -17,7 +20,7 @@ addons:
     - lapack
     - openblas
 
-matrix:
+jobs:
   include:
 # Linux GCC + Codecov
    - name: "Linux GCC + CodeCov"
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -17,7 +20,7 @@ addons:
     - lapack
     - openblas
 
-matrix:
+jobs:
```
jeremylt, line 23: I switched to 'jobs' and added 'os' and 'dist' default values to silence some warnings in Travis's yaml validator.

backends/occa/ceed-occa-restrict.c
```diff
@@ -57,7 +57,7 @@ int CeedElemRestrictionApply_Occa(CeedElemRestriction r,
   if (!strided) {
     ierr = CeedElemRestrictionGetCompStride(r, &compstride); CeedChk(ierr);
   }
-  const CeedTransposeMode restriction = (tmode == CEED_NOTRANSPOSE);
+  bool CeedTransposeMode restriction = (tmode == CEED_NOTRANSPOSE);
```
jeremylt, line 60: These two OCCA changes were to silence some warnings that ICC threw.

Makefile
```diff
@@ -407,30 +407,30 @@ $(OBJDIR)/% : examples/ceed/%.f | $$(@D)/.DIR
...
 $(OBJDIR)/nek-bps : examples/nek/bps/bps.usr examples/nek/nek-examples.sh $(libceed) | $$(@D)/.DIR
-	+$(MAKE) -C examples MPI=$(MPI) CEED_DIR=`pwd` NEK5K_DIR="$(abspath $(NEK5K_DIR))" nek
+	+$(MAKE) -C examples MPI=$(MPI) FC=$(FC) CEED_DIR=`pwd` NEK5K_DIR="$(abspath $(NEK5K_DIR))" nek
 	mv examples/nek/build/bps $(OBJDIR)/bps
 	cp examples/nek/nek-examples.sh $(OBJDIR)/nek-bps
 
 $(OBJDIR)/petsc-% : examples/petsc/%.c $(libceed) $(ceed.pc) | $$(@D)/.DIR
 	+$(MAKE) -C examples/petsc CEED_DIR=`pwd` \
-	  PETSC_DIR="$(abspath $(PETSC_DIR))" OPT="$(OPT)" $*
+	  PETSC_DIR="$(abspath $(PETSC_DIR))" CC=$(CC) OPT="$(OPT)" $*
```
jeremylt, line 423: These examples were using the default env CC instead of the one we set. This was a silent problem for GCC/Clang but caused conflicting opt flags vs compilers for ICC.

backends/occa/ceed-occa-restrict.c
jedbrown: This can't possibly be right. Perhaps `bool restriction = ...`?

jeremylt: Right. Fixed in one of the various fixup I added.

Makefile
jeremylt: I changed my approach here. I switched to Travis building PETSc with the env CC.

PR #551, opened 2020-06-05
PETSc Bps: allow looping over degrees, problems and backends

valeriabarra:
This PR aims at building a loop of different runs inside the BPs executable. This way, if we want to run a battery of benchmarking runs, we don't need to use the `benchmark.sh` script, but can call the executable once.

The parameters that were selected at runtime in the `becnhmark.sh` script, now are translated as CL options of the program, for instance `-max_degree`, `-n` for total number of processors across all compute nodes, and `-p` for number of processors per node.


valeriabarra:
Note: for now I have tested this only locally. Needs testing on a multi-node machine.

jedbrown:
This passed, modulo the unrelated OCCA-on-ARM issue. @valeriabarra Let me know what you think or go ahead and merge. I made it so `-degree` takes a list, but `-local_nodes` takes either a single value or a range. I think the ability to loop over ranges can be useful even in non-benchmark mode.

I'm collecting a new set of results on Noether using this variant. With luck, it'll be less noisy than the previous.

valeriabarra:
Thanks @jedbrown for your additions to this branch. 
I was testing it and, currently, the `postprocess_plot.py` script errors because there is no `DoFs/Sec in CG` string anymore in the bps printout, so there is a `KeyError: 'cg_iteration_dps'` when using that value for plotting. 
Did you mean to remove that from the summary and the postprocess script needs to be updated, or you didn't mean to change that in the perf summary?

jedbrown:
Did you leave `-benchmark` off when you ran? (We should make it print that by default.)

valeriabarra:
Ah, I see what's happened then! Yes, without the `-benchmark` CL option. Will add the print by defualt then. It's better

valeriabarra:
Looks good now and the plotting script ran smoothly. 
I can go on and merge this if there are no further comments/suggestions. Thanks for your help!

jedbrown:
Should we do the pre-solve by default, or otherwise convey when it had been skipped? I'm worried about people using the slower/noisy results.

jeremylt:
That seems reasonable to me.

There is also a benchmark mode in `multigrid.c` fwiw.

valeriabarra:
Okay, I removed the if statement. Now we are always doing the pre-solve.

@jedbrown you said you were testing this on Noether. How do the new results look like compared to the ones from master? A bit less noisy?

jedbrown:
Yeah, see the notebook on the altair branch.

valeriabarra:
It would be nice if we could specify a range of degrees so that we can run the whole suite (say degree 1-16) without specifying all integers in the list `1,2,3, ... , 16`. But I don't know if that's doable with the PETSc parsing option functions

jedbrown:
Isn't that -max_degree?

valeriabarra:
> Isn't that -max_degree?

Ah, I have just noticed that you had created the whole array with that. It wasn't in my original version. I have updated its help string, too. Not sure if more clear now

valeriabarra:
> Yeah, see the notebook on the altair branch.

I see the old results with the multiple calls to `mpirun`

jedbrown:
I think the rendered results are from the new data, but I might have forgotten to git add it. Will fix when I get home.

valeriabarra:
I've reduced the number of commits here a little bit. I am ok to merge it now

jedbrown:
It looks like you force-pushed over my 04f0c8f310c74ffa3268d96b7ef2c5022ffadef9. I just rebased and pushed. It cuts mesh setup cost significantly by reusing a given `DMPlex` across each CEED resource. Erm, I'll also make it do all the BPs while it has that mesh.

valeriabarra:
Sorry, we must have pushed at the same time. I was just reducing number of commits.
Thanks so much for your help!

I am ok to merge this branch, if you are

jedbrown:
I'm fine with it now. I'd ideally like to adjust `SetupDMByDegree` to accept a `PetscFE` that is passed in. This would help a lot on the larger problem sizes by allowing us to create just one `PetscFE` for all the sizes. (The actual solves still take less than 2% of total run-time. Most of the run-time is mesh partition/distribution for the low-degree meshes and the PetscFE setup for high degree.) @knepley, we'd still like a convenient way to skip that FE setup. Perhaps it could be done lazily the first time it is used?

View cleaner perf here.
https://nbviewer.jupyter.org/github/CEED/libCEED/blob/altair/benchmarks/postprocess_altair.ipynb?flush_cache=true

knepley:
You will have to help me understand your perf output. I think there is a problem with Toby's new PetscDualSpaceSetup() for Lagrange. Everything else is completely trivial and happens only on the reference cell. I have performance problems in PyLith also for the new dual space, but cannot track it down since it appears to stem from a bunch of recursive calls.

jedbrown:
@knepley The time is all in GEMM (PetscFECreateTabulation_Basic) and PetscSpaceEvaluate_Polynomial, which does a tabulation for each basis function (thereby stifling vectorization).
```c
  ierr = PetscDualSpaceGetDimension(fem->dualSpace, &pdim);CHKERRQ(ierr);
  ierr = PetscMalloc1(pdim*pdim,&fem->invV);CHKERRQ(ierr);
  for (j = 0; j < pdim; ++j) {
    PetscReal       *Bf;
    PetscQuadrature  f;
    const PetscReal *points, *weights;
    PetscInt         Nc, Nq, q, k, c;

    ierr = PetscDualSpaceGetFunctional(fem->dualSpace, j, &f);CHKERRQ(ierr);
    ierr = PetscQuadratureGetData(f, NULL, &Nc, &Nq, &points, &weights);CHKERRQ(ierr);
    ierr = PetscMalloc1(Nc*Nq*pdim,&Bf);CHKERRQ(ierr);
    ierr = PetscSpaceEvaluate(fem->basisSpace, Nq, points, Bf, NULL, NULL);CHKERRQ(ierr);
    for (k = 0; k < pdim; ++k) {
      /* V_{jk} = n_j(\phi_k) = \int \phi_k(x) n_j(x) dx */
      fem->invV[j*pdim+k] = 0.0;

      for (q = 0; q < Nq; ++q) {
        for (c = 0; c < Nc; ++c) fem->invV[j*pdim+k] += Bf[(q*pdim + k)*Nc + c]*weights[q*Nc + c];
      }
    }
    ierr = PetscFree(Bf);CHKERRQ(ierr);
  }
```
I think we could get a significant boost by batching this. This setup cost really hurts around degree 7 and above. This hack runs cleanly (for our use case) and reduces time from several seconds to a fraction of a second.
```diff
diff --git i/src/dm/dt/fe/interface/fe.c w/src/dm/dt/fe/interface/fe.c
index 746ce1e931..b131f2d7ff 100644
--- i/src/dm/dt/fe/interface/fe.c
+++ w/src/dm/dt/fe/interface/fe.c
@@ -285,7 +285,7 @@ PetscErrorCode PetscFESetUp(PetscFE fem)
   PetscValidHeaderSpecific(fem, PETSCFE_CLASSID, 1);
   if (fem->setupcalled) PetscFunctionReturn(0);
   fem->setupcalled = PETSC_TRUE;
-  if (fem->ops->setup) {ierr = (*fem->ops->setup)(fem);CHKERRQ(ierr);}
+  if (fem->ops->setup && 0) {ierr = (*fem->ops->setup)(fem);CHKERRQ(ierr);}
   PetscFunctionReturn(0);
 }
 
diff --git i/src/dm/interface/dm.c w/src/dm/interface/dm.c
index c3ac1f0d27..3e15a2c02d 100644
--- i/src/dm/interface/dm.c
+++ w/src/dm/interface/dm.c
@@ -5460,7 +5460,7 @@ PetscErrorCode DMCreateDS(DM dm)
     ierr = ISRestoreIndices(fields, &fld);CHKERRQ(ierr);
   }
   /* Setup DSes */
-  if (doSetup) {
+  if (doSetup && 0) {
     for (s = 0; s < dm->Nds; ++s) {ierr = PetscDSSetUp(dm->probs[s].ds);CHKERRQ(ierr);}
   }
   PetscFunctionReturn(0);
```

valeriabarra:
> 
> View cleaner perf here.
> https://nbviewer.jupyter.org/github/CEED/libCEED/blob/altair/benchmarks/postprocess_altair.ipynb?flush_cache=true
Thanks Jed for the update. Interesting that there seem to be two data points for BP1, but when I hover the mouse on them, they show the same number of unknowns. Was this a repeated run for some reason or a double count in the data analysis?
Screenshot from 2020-06-09 10-32-09

jedbrown:
Yes, those are two independent runs in different data files. The difference gives an indication of system noise. I'd like to enable better plotting of repeat observations (they're super cheap once the problem has been set up).

valeriabarra:
Yup, I've seen now the two different output files

jeremylt:
Did we want to make the same benchmark flag changes in `multigrid.c` before merging so the three files agree?

jedbrown:
Sure (but not necessary because multigrid BPs are not standardized). If we do that, I'm also inclined to hoist creation of the PetscFE out.

valeriabarra:
I would leave those to separate PRs

valeriabarra:
Since the CI failures in this PR are only with the occa backend, which is currently being updated, I think we can proceed with the merge here.

jedbrown:
Sounds good; go ahead.

PR #552, opened 2020-06-08
Composite Operator support for AssembleLinearDiag/PBDiag

jeremylt:
I'm working on the tests and a sensible error message when the fields are dissimilar.

EDIT: Ready for review

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -322,6 +322,22 @@ int CeedOperatorGetQFunction(CeedOperator op, CeedQFunction *qf) {
 }
 
 /**
+  @brief Get a boolean value indicating if the CeedOperator is composite
+
+  @param op                CeedOperator
+  @param[out] isComposite  Variable to store composite status
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+
+int CeedOperatorGetCompositeStatus(CeedOperator op, bool *isComposite) {
+  *isComposite = op->composite;
+  return 0;
+}
+
+/**
   @brief Get the number of suboperators associated with a CeedOperator
 
   @param op              CeedOperator
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -799,8 +799,8 @@ int CeedOperatorAssembleLinearQFunction(CeedOperator op, CeedVector *assembled,
 
   This returns a CeedVector containing the diagonal of a linear CeedOperator.
 
-  Note: Currently only non-composite CeedOperators with a single field are
-          supported.
+  Note: Currently only non-composite CeedOperators with a single field and
+          composite CeedOperators with single field sub-operators are supported.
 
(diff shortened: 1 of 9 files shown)


jeremylt:
Update: Tested and working. Codecov should be happy once Travis runs.

The ToDo for the PR is a message with multiple fields.

jeremylt:
Should be good to go. Now composite operators are supported on the CPU side. I'll use this as the basis for my GPU side work.

I used the gallery in this new test, which means I adjusted the test cases for ex1/ex2 so we are testing/using more of that code.

interface/ceed-operator.c
```diff
@@ -321,6 +321,22 @@ int CeedOperatorGetQFunction(CeedOperator op, CeedQFunction *qf) {
...
+
+  @param op                CeedOperator
+  @param[out] isComposite  Variable to store composite status
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+
+int CeedOperatorGetCompositeStatus(CeedOperator op, bool *isComposite) {
```
valeriabarra, line 335: Thanks Jeremy!
Sometimes getter functions that return a boolean are called `IsProperty()`. In this case it would be  `CeedOperatorIsComposite`; a bit shorter. This could be a minor preference for someone. It doesn't really matter to me --- overall it looks good!

jeremylt: I like `Is*` better than `Get*Status`, but we use `Get*Status` pretty consistently (just an unfortunate choice of mine at some point in the past that's spread a fair bit now). I'd be open to changing it in a separate PR.

valeriabarra: Sounds good. 
In that case, I know this comment is also late, but perhaps we could consider changing `CeedOperatorAssembleLinearDiagonal` -> `CeedOperatorLinearAssembleDiagonal` and so on for relative functions?
It could be a minor wording preference for someone. Again, it doesn't really matter to me. But it just occurred to me that the way it is now, it sounds like the diagonal is linear (which may be confusing to someone reading it for the first time) rather than the operator

jedbrown: That'd be fine with me, and best to get any such breaking changes out of the way in 0.7.

jedbrown: It's possible to deprecate the old names so they continue working with a warning for the next release.

jeremylt: The `Get*Status` functions are really only used internally, so the changes would be pretty easy.

jeremylt: For `CeedOperator*Diagonal`, I don't really care, but now is the time to decide if we want to change the name before I tie this into MFEM.

jedbrown: Valeria's naming seems more consistent if cost to change is still low.

jeremylt: Cool, I can whip up a name change PR.

PR #553, opened 2020-06-09
CUDA/MAGMA - add AssembleLinearQFunction

jeremylt:
I've added AssembleLinearQFunction() for CUDA/MAGMA. With this and #552 I can start adding GPU diagonal assembly support.

### DIFF
--- a/backends/cuda/ceed-cuda-operator.c
+++ b/backends/cuda/ceed-cuda-operator.c
@@ -211,36 +211,27 @@ static int CeedOperatorSetup_Cuda(CeedOperator op) {
 }
 
 //------------------------------------------------------------------------------
-// Apply and add to output
+// Setup Operator Inputs
 //------------------------------------------------------------------------------
-static int CeedOperatorApplyAdd_Cuda(CeedOperator op, CeedVector invec,
-                                     CeedVector outvec, CeedRequest *request) {
-  int ierr;
-  CeedOperator_Cuda *impl;
-  ierr = CeedOperatorGetData(op, (void *)&impl); CeedChk(ierr);
-  CeedQFunction qf;
-  ierr = CeedOperatorGetQFunction(op, &qf); CeedChk(ierr);
-  CeedInt Q, numelements, elemsize, numinputfields, numoutputfields, size;
-  ierr = CeedOperatorGetNumQuadraturePoints(op, &Q); CeedChk(ierr);
-  ierr = CeedOperatorGetNumElements(op, &numelements); CeedChk(ierr);
-  ierr = CeedQFunctionGetNumArgs(qf, &numinputfields, &numoutputfields);
-  CeedChk(ierr);
-  CeedOperatorField *opinputfields, *opoutputfields;
-  ierr = CeedOperatorGetFields(op, &opinputfields, &opoutputfields);
-  CeedChk(ierr);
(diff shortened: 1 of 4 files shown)


jeremylt:
The active/passive vector difference doesn't say anything about layout of the vectors.The only distinctions are
1) The passive vectors are set at the time of creating the operator field
2) The active vectors can be changed on every operator apply

A user can (and we have a test for this somewhere) make a operator completely using passive vectors.

So the only time a restriction can be omitted is when the vector (input or output, active or passive) has the backend strides.

nbeams:
Okay, instead of speaking of only "passive" input I mean specifically with input basis action `NONE`.

I think the problem still stands, specific to having different E- and Q-Vector layouts with basis `NONE`. 

Consider a passive input with `NONE` input basis action, which should be an L-Vector with a layout corresponding to its restriction (as specified by the Operator interface, everything in or out is an L-Vector), but this restriction does not have the backend strides.  The input vector goes through its input element restriction which does L-Vector --> E-Vector.  But then the input basis action for `NONE` passes this E-Vector directly through as a Q-Vector for the QFunction. If E- and Q-Vectors have different layouts, like in cuda-ref, this is the wrong input for the QFunction.

To me it looks like for cuda-ref, the `AssembleLinearQFunction` code will output in E-Vector layout instead of L-(Q)-Vector layout (it doesn't matter for magma because they are the same layout). Then when the assembled vector is used as an input-with-`NONE`-basis in the tests, it's okay because it was actually an E-Vector and its input restriction "reverses" it back to L-Vector.  But that won't work in general for any L-Vector described by any general restriction.  The most general case would be described above, where you have an L-Vector, its input restriction turns it into an E-Vector, and the E-Vector gets passed directly to the QFunction in the wrong format.

backends/cuda/ceed-cuda-operator.c
```diff
@@ -403,34 +491,185 @@ static int CeedOperatorApplyAdd_Cuda(CeedOperator op, CeedVector invec,
...
+  ierr = CeedOperatorRestoreInputs_Cuda(numinputfields, qfinputfields,
+                                        opinputfields, true, impl);
+  CeedChk(ierr);
+
+  // Restore output
+  if (impl->eandqdiffer) {
+    ierr = CeedVectorRestoreArray(assembledE, &a); CeedChk(ierr);
+    ierr = CeedElemRestrictionApply(*rstr, CEED_TRANSPOSE, assembledE,
+                                    *assembled, request); CeedChk(ierr);
+    ierr = CeedVectorDestroy(&assembledE); CeedChk(ierr);
```
jeremylt, line 661: Now I see what you're saying. That would explain why these lines are here when they are not in the CPU impl. CPU does not have this issue because our Evector and Qvector layouts are the same.

@YohannDudouit thoughts? I feel like rearanging the cuda ref/reg/shared backends to have the same Lvector and Qvector layout would take a fair bit of effort but would make things a lot cleaner and provide correct output when using user provided qdata with a non backend strides restriction.

jeremylt:
I've started #554 to address this issue.

jeremylt:
#554 is ready - when merged, I need to rebase this branch to include those changes so I can clean up per the discussion above

jeremylt:
Alrighty. @nbeams I believe I've fixed your concerns above.

As an added benefit, fallback of this now works for `/gpu/cuda/gen` as well.

PR #554, opened 2020-06-11
Convert CUDA ref/reg/shared E-Layout

jeremylt:
I have modified the test suite to work for any [strided] E-Layout. Next I'll adapt CUDA ref/reg/shared to use the same E-Layout as Q-Layout.


interface/ceed-elemrestriction.c
```diff
@@ -185,6 +185,55 @@ int CeedElemRestrictionGetBackendStridesStatus(CeedElemRestriction rstr,
...
+                            stored as [nodes, components, elements].
+                            The data for node i, component j, element k in the
+                            E-vector is given by
+                            i*strides[0] + j*strides[1] + k*strides[2]
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedElemRestrictionGetELayout(CeedElemRestriction rstr,
```
jeremylt, line 203: At some point, it may make sense to have `CeedOperatorGetELayout` and `CeedOperatorGetQLayout` to better manage checking if the base and fallback operators are compatible.

jeremylt:
Ok, ready for review. This took far less time than I expected since the code made good sense. Thanks @YohannDudouit  for that.

The t5*, PETSc, MFEM, and NEK tests pass just fine. I'm curious about performance impacts. I think this might be a small boost. (Although, `cuda/ref`, `cuda/reg`, and `cuda/shared` aren`t as fast as `cuda/gen`.)

backends/cuda-reg/ceed-cuda-reg-basis.c
```diff
@@ -31,12 +31,14 @@ typedef CeedScalar real;
 //------------------------------------------------------------------------------
 // Read non interleaved dofs
 //------------------------------------------------------------------------------
+/*
```
jeremylt, line 34: I highlighted code that is not currently used (some from my changes, some from older changes). We should decide what to do with these functions. Untested/unused code gets stale fast. I'd recommend deleting.

nbeams:
Do we want to go ahead and 
1) Remove `eandqdiffer` from the `CeedOperator_Cuda` struct in ceed-cuda.h, and 
2) Remove ceed-magma-operator.c entirely (and the `OperatorCreate` backend function set in ceed-magma.c) 
?  
It seems it would fit with this PR.

jeremylt:
Good catch, thank you.

tests/tap.sh
```diff
@@ -171,15 +171,6 @@ for ((i=0;i<${#backends[@]};++i)); do
...
 
-    # grep to skip t204-7 for MAGMA (different layout used for E-vectors)
-    if [[ "$backend" = *"magma" ]] \
-            && [[ "$1" = "t20"[4-7]* ]] ; then
-        printf "ok $i0 # SKIP - backend uses different E-vector layout $1 $backend\n"
-        printf "ok $i1 # SKIP - backend uses different E-vector layout $1 $backend stdout\n"
-        printf "ok $i2 # SKIP - backend uses different E-vector layout $1 $backend stderr\n"
-        continue
-    fi
-
```
nbeams, line 182: I'm really happy to be able to take advantage of the element restriction tests for the MAGMA backend again!

backends/cuda/ceed-cuda-operator.c
```diff
@@ -311,7 +311,7 @@ static int CeedOperatorApplyAdd_Cuda(CeedOperator op, CeedVector invec,
   for (CeedInt i = 0; i < numoutputfields; i++) {
     ierr = CeedQFunctionFieldGetEvalMode(qfoutputfields[i], &emode);
     CeedChk(ierr);
-    if (emode == CEED_EVAL_NONE && !impl->eandqdiffer) {
+    if (emode == CEED_EVAL_NONE) {
       // If the E- and Q-Vector layouts are not the same, we will perform
       // an element restriction during the output basis apply, and don't
       // need to set these pointers.  Otherwise, set the output Q-Vector
```
nbeams, line 317: First sentence of this comment no longer applies and should be removed.

jeremylt: done

backends/cuda-reg/ceed-cuda-reg-basis.c
YohannDudouit: I'm okay with deleting.

interface/ceed-elemrestriction.c
YohannDudouit: One of the problem I see here is that now the E-layout must be strided, if you look at the unused code for `cuda-reg`, the data layout expected was not strided.

backends/cuda-reg/ceed-cuda-reg-basis.c
```diff
@@ -123,7 +127,7 @@ inline __device__ void interp1d(const CeedInt nelem, const int transpose,
     for(int comp = 0; comp < BASIS_NCOMP; comp++) {
       if (!transpose) {
         const int sizeU = P1D;
-        readDofs(bid, tid, comp, sizeU, nelem, d_U, r_V);
+        readQuads(bid, tid, comp, 0, sizeU, nelem, d_U, r_V);
```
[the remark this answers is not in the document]
jeremylt: It's the same layout for e and q now, so it would the same function with different names, as far as I could tell.

jeremylt:
I wouldn't say that it must be strided in general, only that it most be strided to use the test suite.

backends/cuda-reg/ceed-cuda-reg-basis.c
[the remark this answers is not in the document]
YohannDudouit: From what I can see they don't take the same number of arguments, I think it would be cleaner to have:
```c++
inline void readDofs(const int bid, const int tid, const int comp, const int sizeU, const int nelem, const double *U, double *V)
{
   readQuads(bid, tid, comp, 0, sizeU, nelem, U, V);
}
```
and the same for `writeDofs`.

backends/cuda-reg/ceed-cuda-reg-basis.c
jeremylt: done

backends/cuda-reg/ceed-cuda-reg-basis.c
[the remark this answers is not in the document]
jeremylt: agreed. done

PR #556, opened 2020-06-12
Improve Developer Documentation

jeremylt:
I added some improved documentation for developers to better give us a shared language about a few things and to write down some 'corporate knowledege' that was undocumented.

### DIFF
--- a/doc/sphinx/source/libCEEDdev.rst
+++ b/doc/sphinx/source/libCEEDdev.rst
@@ -58,7 +58,7 @@ There are several common layouts for **L-vectors**, **E-vectors**, and **Q-vecto
 
 * **E-vector** layouts
 
-  * If possible, backends should use :ref:`CeedElemRestrictionSetELayout` to use the ``t2**`` tests.
+  * If possible, backends should use :c:func:`CeedElemRestrictionSetELayout()` to use the ``t2**`` tests.
     If the backend uses a strided **E-vector** layout, then the data for node ``i``, component ``j``, element ``k`` in the **E-vector** is given by ``i*layout[0] + j*layout[1] + k*layout[2]``.
 
   * Backends may choose to use a non-strided **E-vector** layout; however, the ``t2**`` tests will not function correctly in this case and the tests will need to be whitelisted for the backend to pass the test suite.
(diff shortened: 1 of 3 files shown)


doc/sphinx/source/libCEEDdev.rst
```diff
@@ -42,6 +42,51 @@ are purely implicit -- one just indexes the same array using the
...
+    If the backend uses a strided **E-vector** layout, then the data for node ``i``, component ``j``, element ``k`` in the **E-vector** is given by ``i*layout[0] + j*layout[1] + k*layout[2]``.
+
+  * Backends may choose to use a non-strided **E-vector** layout; however, the ``t2**`` tests will not function correctly in this case and the tests will need to be whitelisted for the backend to pass the test suite.
+
+* **Q-vector** layouts
+
+  * When the size of a :ref:`CeedQFunction` field is greater than ``1``, data for quadrature point ``i`` component ``j`` can be found in the **Q-vector** at index ``i + Q*j``.
+    Backends are free to provide the quadrature points in any order.
+
+  * When the :ref:`CeedQFunction` field has ``emode`` ``CEED_EVAL_GRAD``, data for quadrature point ``i``, component ``j``, derivative ``k`` can be found in the **Q-vector** at index ``i + Q*j + Q*size*k``.
```
valeriabarra, line 71: Thanks @jeremylt for these additions. These really clarify a lot!

Question: Do we want to mention the case of `CEED_EVAL_NONE` that  determines the fact that the Q-Vector uses the E-Vector data layout? At least this was my understanding after PR #554, but correct me if I'm wrong

jeremylt: That's up to the backend. A backend may choose to add a permutation to shuffle from e to q in the case of CEED_EVAL_NONE. I agree that backends *shouldn't*, in general, use separate e and, q layouts, but there may be some GPU use cases where this is appropriate.

Other possible fixes for 554 could have been to make a separate permutation for CEED_EVAL_NONE or to change the CEED_EVAL_NONE restrictions to write directly to the q layout.

valeriabarra: Ok, cool. Do we want to mention something like you said here, i.e., "A backend may choose to add a permutation to shuffle from e to q in the case of CEED_EVAL_NONE"? Or is it redundant at this point?

jeremylt: Yeah, I can add a new bullet under q warning about different e and q layouts

jeremylt: Added. Thanks

jeremylt:
@dmed256, this new documentation might be useful to you.

doc/sphinx/source/libCEEDdev.rst
```diff
@@ -42,6 +42,53 @@ are purely implicit -- one just indexes the same array using the
...
+----------------------------------------
+
+Ceed backends are free to use any **E-vector** and **Q-vector** data layout, to include never fully forming these vectors, so long as the backend passes the ``t5**`` series tests and all examples.
+There are several common layouts for **L-vectors**, **E-vectors**, and **Q-vectors**, detailed below:
+
+* **L-vector** layouts
+
+  * **L-vectors** described by a :ref:`CeedElemRestriction` have a layout described by the ``offsets`` array and ``compstride`` parameter.
+    Data for node ``i``, component ``j``, element ``k`` can be found in the **L-vector** at index ``offsets[i + j*elemsize] + k*compstride``.
+
```
nbeams, line 55: I don't think this is correct.  Per the comments in `ceed-elemrestriction.c`, 

```
offsets    Array of shape [@a nelem, @a elemsize]. Row i holds the
                      ordered list of the offsets (into the input CeedVector)
                      for the unknowns corresponding to element i, where
                      0 <= i < @a nelem. All offsets must be in the range
                      [0, @a lsize - 1].
```
So this should be `offsets[i + k*elemsize] + j*compstride`, no?

jeremylt: Crap. Good catch, bad copy-paste from me

dmed256:
@jeremylt This helps a lot, thank you :)

PR #561, opened 2020-06-17
Ceed - add IsDeterministic

jeremylt:
This PR starts the process of documenting which backends are deterministic.

1) Where do we want to keep a list of which backends are deterministic?
2) How can we verify? Run an operator twice and check that the output matches?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -207,7 +207,7 @@ directory, or set ``MAGMA_DIR`` to the proper location.  MAGMA version 2.5.0 or 
 
 Bit-for-bit reproducibility is important in some applications.
 However, some libCEED backends use non-deterministic operations, such as ``atomicAdd`` for increased performance.
-The backends which are capable of generating reproducable results, with the proper compilation options, are highlighted in the list above.
+The backends which are capable of generating reproducible results, with the proper compilation options, are highlighted in the list above.
 
 Examples
 ----------------------------------------
--- a/README.rst
+++ b/README.rst
@@ -130,45 +130,45 @@ Backends
 
 There are multiple supported backends, which can be selected at runtime in the examples:
 
-+----------------------------+---------------------------------------------------+
-| CEED resource              | Backend                                           |
-+----------------------------+---------------------------------------------------+
-| ``/cpu/self/ref/serial``   | Serial reference implementation                   |
-+----------------------------+---------------------------------------------------+
-| ``/cpu/self/ref/blocked``  | Blocked reference implementation                  |
-+----------------------------+---------------------------------------------------+
-| ``/cpu/self/ref/memcheck`` | Memcheck backend, undefined value checks          |
-+----------------------------+---------------------------------------------------+
-| ``/cpu/self/opt/serial``   | Serial optimized C implementation                 |
-+----------------------------+---------------------------------------------------+
-| ``/cpu/self/opt/blocked``  | Blocked optimized C implementation                |
-+----------------------------+---------------------------------------------------+
-| ``/cpu/self/avx/serial``   | Serial AVX implementation                         |
-+----------------------------+---------------------------------------------------+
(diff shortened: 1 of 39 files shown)


jeremylt:
Updated `CeedIsDeterministic` to require an explicit use of `true` or `false`.

I think that determinism is tied to the Ceed backend. I could (in theory) make a backend that delegates to `/cuda/ref` for everything except a new deterministic restriction. Then `/cuda/ref` operators would be non-deterministic while `/cuda/newfangled` would be deterministic, but `/cuda/newfangled` operators would delegate through the `/cuda/ref` operator, so we need to ID the backends as different.

jedbrown:
Perhaps this interface should be saying whether the Ceed is *capable* of being deterministic and we could have a public interface to request that it stick to deterministic implementations. In that case, the query should perhaps be able to work with a resource string so the caller can automatically choose a deterministic one if that's what they want. But it's okay to keep it simple if this is what fulfills your current need.

jeremylt:
I don't have a current need. There were just some discussions on the MFEM team with some testing (@YohannDudouit and @camierjs have better context) so I thought it would be helpful to at least start documenting which backends are/can be deterministic.

jeremylt:
I added to the list of backends a column indicating if the backends are capable of deterministic results.

YohannDudouit:
We could also have `Cuda-ref` use the deterministic Restriction/Prolongation from `Cuda-reg`, then only `cuda-gen` would be non-deterministic (and cannot be as far as I know).

YohannDudouit:
The best would be to benchmark the perf difference between the Restriction of Ref and Reg, if there is no benefit in the Ref Restriction, then we move the Restriction from Ref into Ref. If there is a difference, we could imagine a mechanism to choose between deterministic and non-deterministic.

jeremylt:
Sure, I don't have any strong feelings one way or another, but I do believe that `ref` is the slowest CUDA backend.

nbeams:
I got an email notification about commit d8c2b9c that would change MAGMA's delegate from ref to reg, but now I don't see it (or that change) here.  Is that because it was changed back?  We currently only use the Operator and QFunction from cuda-ref, so changing the delegate wouldn't change the deterministic status (which is no).

jeremylt:
@nbeams yeah, that was a mistake on my part. I caught my mistake and reverted the change.

jeremylt:
Is the MAGMA basis apply deterministic? We could make a deterministic MAGMA backend if that is the case by making a backend that delegates the restriction to cuda/reg

abdelfattah83:
> Is the MAGMA basis apply deterministic? We could make a deterministic MAGMA backend if that is the case by making a backend that delegates the restriction to cuda/reg

The basis action kernels in MAGMA should be deterministic. If the restrictions in cuda/reg are deterministic, then the magma backend can guarantee reproducible numerical output by using those restrictions. This has to be verified, though, by running actual benchmarks.

jeremylt:
The plan is to have a future benchmark that verifies the reproducibility of backends in a follow-up PR. I can make a small `gpu/magma/det` PR as a follow-up as well.

README.rst
```diff
@@ -205,6 +205,10 @@ By default, ``MAGMA_DIR`` is set to ``../magma``; to build the MAGMA backend
 with a MAGMA installation located elsewhere, create a link to ``magma/`` in libCEED's parent
 directory, or set ``MAGMA_DIR`` to the proper location.  MAGMA version 2.5.0 or newer is required.
 
+Bit-for-bit reproducibility is important in some applications.
+However, some libCEED backends use non-deterministic operations, such as ``atomicAdd`` for increased performance.
+The backends which are capable of generating reproducable results, with the proper compilation options, are highlighted in the list above.
```
jedbrown, line 210:
```suggestion
The backends which are capable of generating reproducible results, with the proper compilation options, are highlighted in the list above.
```
Perhaps check spelling more broadly.

PR #562, opened 2020-06-18
python: use precise ext suffix instead of globs in tests

jedbrown:
Otherwise we get multiple matching globs any time we've built for multiple versions of Python.

If approved, I'll do the merge manually so it's consistent with flake8 style in master.

### DIFF
--- a/tests/python/test-4-qfunction.py
+++ b/tests/python/test-4-qfunction.py
@@ -18,7 +18,6 @@
 # Test Ceed QFunction functionality
 
 import os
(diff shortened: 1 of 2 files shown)


jedbrown:
OCCA failure because the version wasn't pinned. This branch intentionally started from the v0.6 tag.

jedbrown:
Here's a tentative merge result and pipeline: https://travis-ci.com/github/CEED/libCEED/builds/171962299

PR #563, opened 2020-06-19
XSMM - use v1.16

jeremylt:
This updates our CI to the latest LIBXSMM release and fixes the warnings generated by LIBXSMM's macro to refefine `_Bool` to `int`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -115,10 +115,10 @@ install:
         && cd occa && git reset --hard 327badb && cd ..
         && make -C occa -j2
         && export OCCA_DIR=$PWD/occa;
-# LIBXSMM v1.15 (need to use specific commit, XSMM develops on master)
+# LIBXSMM v1.16 (need to use specific commit, XSMM develops on master)
   - if [[ "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
         git clone https://github.com/hfp/libxsmm.git
-        && cd libxsmm && git reset --hard 724205c && cd ..
+        && cd libxsmm && git reset --hard 6d6a0f6 && cd ..
         && make -C libxsmm -j2
         && export XSMM_DIR=$PWD/libxsmm;
     fi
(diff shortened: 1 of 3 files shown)


jeremylt:
Note - CI will fail due to the Nek5000 bug that I have a PR open for: https://github.com/Nek5000/Nek5000/pull/667

jeremylt:
Verified: v1.16 works locally and v1.16 clones and installs correctly on Travis.

.travis.yml
```diff
@@ -115,10 +115,10 @@ install:
         && cd occa && git reset --hard 327badb && cd ..
         && make -C occa -j2
         && export OCCA_DIR=$PWD/occa;
-# LIBXSMM v1.15 (need to use specific commit, XSMM develops on master)
+# LIBXSMM v1.16 (need to use specific commit, XSMM develops on master)
   - if [[ "$TRAVIS_CPU_ARCH" == "amd64" ]]; then
         git clone https://github.com/hfp/libxsmm.git
-        && cd libxsmm && git reset --hard 724205c && cd ..
+        && cd libxsmm && git reset --hard 6d6a0f6 && cd ..
```
valeriabarra, line 121: Just curious. Would it be the same if we used `git checkout tags/1.16` here instead of  `git reset --hard <commit>`, or we wouldn't want to do a `checkout`?

jedbrown: `git checkout -f <commit>` would be good.

jeremylt: It would have the same effect, I believe. We've just been using commit numbers because sometimes we want a commit between releases (like with OCCA and PETSc).

jeremylt: Updated. If there are no objections, I can merge now. It annoys me that our CI fails due to two external issues, but I don't suppose we can do more about that than we are doing right now.

jedbrown: Difference is that `checkout` (or `switch`) changes `HEAD` while `reset` changes the index/working tree. The difference arises if you were to commit something or otherwise care about that detail.

In the interest of reproducible CI and ability to maintain a `stable` branch (post-1.0 if not before), we should probably move toward pinning versions of each dependency.

PR #564, opened 2020-06-19
CUDA - LinearAssemble*Diagonal

jeremylt:
~~Initial work for~~ diagonal assembly for CUDA/MAGMA backends.

Closes #425

Issue #451, opened 2020-01-31
PETSc BP4 and BP6 with OCCA backend


valeriabarra:
As per our discussion in PR #444 , we need to make sure that after the revamping of the OCCA backend, the bug in the PETSc multi-component BP4&BP6 examples with `cpu/occa` is fixed. 

This is just for future reference.

PR #566, opened 2020-06-23
Solids - enable use of GPU vectors

jeremylt:
Once #564 is merged, I will update and test this PR.

### DIFF
--- a/examples/petsc/bpsraw.c
+++ b/examples/petsc/bpsraw.c
@@ -35,7 +35,7 @@
 //     ./bpsraw -problem bp5 -ceed /omp/occa
 //     ./bpsraw -problem bp6 -ceed /ocl/occa
 //
-//TESTARGS -ceed {ceed_resource} -test -problem bp2 -degree 5 -qextra 1 -ksp_max_it_clip 20,20
+//TESTARGS -ceed {ceed_resource} -test -problem bp2 -degree 5 -qextra 1 -ksp_max_it_clip 15,15
 
 /// @file
 /// CEED BPs example using PETSc
@@ -232,7 +232,7 @@ bpData bpOptions[6] = {
     .error = Error3,
     .setupgeofname = SetupDiffGeo_loc,
     .setuprhsfname = SetupDiffRhs3_loc,
-    .applyfname = Diff_loc,
+    .applyfname = Diff3_loc,
     .errorfname = Error3_loc,
     .inmode = CEED_EVAL_GRAD,
     .outmode = CEED_EVAL_GRAD,
@@ -264,7 +264,7 @@ bpData bpOptions[6] = {
     .error = Error3,
     .setupgeofname = SetupDiffGeo_loc,
     .setuprhsfname = SetupDiffRhs3_loc,
-    .applyfname = Diff_loc,
+    .applyfname = Diff3_loc,
     .errorfname = Error3_loc,
     .inmode = CEED_EVAL_GRAD,
     .outmode = CEED_EVAL_GRAD,
--- a/examples/solids/src/matops.c
+++ b/examples/solids/src/matops.c
@@ -41,6 +41,10 @@ PetscErrorCode ApplyLocalCeedOp(Vec X, Vec Y, UserMult user) {
   CeedVectorSetArray(user->Yceed, user->memType, CEED_USE_POINTER, y);
 
   // Apply CEED operator
+  // Note: CeedVectorSyncArray is not required when the backend memtype and
+  //         the PETSc vector agree. We could in that case use
+  //         VecGetArrayInPlace. Instead, we use SetArray/SyncArray so we can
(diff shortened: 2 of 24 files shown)


examples/solids/src/matops.c
```diff
@@ -34,19 +34,20 @@ PetscErrorCode ApplyLocalCeedOp(Vec X, Vec Y, UserMult user) {
...
   CeedOperatorApply(user->op, user->Xceed, user->Yceed, CEED_REQUEST_IMMEDIATE);
-  CeedVectorSyncArray(user->Yceed, CEED_MEM_HOST);
+  CeedVectorSyncArray(user->Yceed, user->memType);
 
   // Restore PETSc vectors
-  ierr = VecRestoreArrayRead(user->Xloc, (const PetscScalar **)&x);
+  ierr = user->VecRestoreArrayRead(user->Xloc, (const PetscScalar **)&x);
   CHKERRQ(ierr);
-  ierr = VecRestoreArray(user->Yloc, &y); CHKERRQ(ierr);
+  ierr = user->VecRestoreArray(user->Yloc, &y); CHKERRQ(ierr);
```
jeremylt, line 50: Debugging note - something in these 14 lines is breaking our multi-component PETSc examples.

jeremylt: If I transfer host pointers from PETSc to libCEED, then the code works correctly. If I pass device pointers, then the code does not. We have the same issue with `petsc-bps`.

jeremylt: Note - this problem doesn't present for single component PETSc examples, like BP1 or BP3.

jeremylt: Solids works, still tracking the bp/multigrid bug

jeremylt:
Got the bug (or more accurately, a whole slew of bugs).

Ready for review.

examples/petsc/bpsraw.c
```diff
@@ -878,7 +878,7 @@ int main(int argc, char **argv) {
     ierr = KSPSetTolerances(ksp, 1e-10, PETSC_DEFAULT, PETSC_DEFAULT,
                             PETSC_DEFAULT); CHKERRQ(ierr);
   }
-  ierr = KSPSetOperators(ksp, mat, mat); CHKERRQ(ierr);
+  ierr = KSPSetOperators(ksp, mat, mat);
```
jedbrown, line 881: Why?

jeremylt: woops

jeremylt: fixed

examples/petsc/bpsraw.c
```diff
@@ -232,7 +232,7 @@ bpData bpOptions[6] = {
     .error = Error3,
     .setupgeofname = SetupDiffGeo_loc,
     .setuprhsfname = SetupDiffRhs3_loc,
-    .applyfname = Diff_loc,
+    .applyfname = Diff3_loc,
```
valeriabarra, line 235: Oh wow, thanks for spotting this!

jeremylt: It might be the cause of #451. Let me check.

jeremylt: Nope

valeriabarra: :/

jeremylt:
Grr, found a different bug, squashed it, and updated test parameters to help catch it.

jeremylt:
I thought about ditching the `VecGetArray` ect, but I ended up retaining them because they were useful for debugging purposes.

jeremylt:
I can update the manpage, but I think it should say that users either use [ `GetArray` and `RestoreArray`] or [`SetArray` and `SyncArray`].

jedbrown:
Yeah, perhaps explain that in the manual. Also that it's a no-op if the preferred memtype of your backend matches the memtype that you've Set. So if one always matches CUDA backends with PETSc CUDA Vecs and host with host, then the call could be dropped. (We can recommend not doing that because it's useful for debugging.)

Should we add an interface to PETSc where the desired memory space is an explicit argument, instead of implicit as in `VecGetArrayInPlace`?

jeremylt:
Updated

jeremylt:
We perhaps want to consider a `CeedVectorResetArray(CeedVector vec)` at some point (which would just be a clearer way to call `CeedVectorSetArary(vec, CEED_MEM_HOST, NULL)` to detach an external pointer from a libCEED vector.

jedbrown:
`CeedVectorTakeArray` to reclaim it? I agree it's dirty and error-prone to leave the stale pointer in there.

jeremylt:
I was thinking of basically: https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecResetArray.html

We could also do a version that does Sync + Reset.

I'll make an issue

PR #568, opened 2020-06-23
Small CUDA Tidying

jeremylt:
This fixes some small memory leaks in the CUDA backend and tidies a few other little things.


jeremylt:
I plan on double checking the code coverage before I merge. I might just make a t3 basis test for the specific AVX case and except the GPU backends.

PR #572, opened 2020-06-27
Name the kernels in cuda-gen.

YohannDudouit:
This PR names the kernels generated by the `cuda-gen` backend, this allows to differentiate them when profiling.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -776,6 +776,9 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
   code << devFunctions;
 
   string qFunction(qf_data->qFunctionSource);
+  string qFunctionName(qf_data->qFunctionName);
+  string oper;
+  oper = "kernel_" + qFunctionName;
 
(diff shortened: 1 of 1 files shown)


backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -776,6 +776,9 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
   code << devFunctions;
 
   string qFunction(qf_data->qFunctionSource);
+  string qFunctionName(qf_data->qFunctionName);
+  string oper;
+  oper = "kernel_" + qFunctionName;
```
jedbrown, line 781: Should this be more namespaced? Seems like there could be lots of places within a production application that would have as much claim to `kernel_mass` as we do. Perhaps something like this?
```suggestion
  oper = "CeedKernel_Cuda_gen_" + qFunctionName;
```

jedbrown:
I wish GitHub had something like GitLab Merge Trains. I think this is ready to merge after confirming that pipelines pass.

YohannDudouit:
@jedbrown @jeremylt If I understand correctly Travis is failing because of Occa, is it okay to merge?

PR #574, opened 2020-06-29
Use Released for CI

jeremylt:
With XSMM releasing a new version, I went ahead and made this PR to switch Travis to use specific releases of updates. I hope this reduces failures due to other repos


jeremylt:
I'll let Travis catch up before merging

PR #577, opened 2020-06-30
Vec - add CeedVectorTakeArray

### LINKED ISSUE #575: VecResetArray
We should create

1) `CeedVectorResetArray(vec)` to replace use of `CeedVectorSetArray(vec, CEED_MEM_HOST, NULL)`

2) `CeedVectorTakeArray(vec)` to do `CeedVectorSyncArray(vec, memtype)` + `CeedVectorResetArray(vec)`

For 1) this is just a clearer renaming so its obvious what the effect is.

In the case of 2) the sync call would be against the last `memtype` used in `CeedVectorSetArray()` with `CEED_USE_POINTER`.

See discussion in #566

jeremylt:
Closes #575

### DIFF
--- a/backends/cuda/ceed-cuda-vector.c
+++ b/backends/cuda/ceed-cuda-vector.c
@@ -154,6 +154,39 @@ static int CeedVectorSetArray_Cuda(const CeedVector vec,
 }
 
 //------------------------------------------------------------------------------
+// Vector Take Array
+//------------------------------------------------------------------------------
+static int CeedVectorTakeArray_Cuda(CeedVector vec, CeedMemType mtype,
+                                    CeedScalar **array) {
+  int ierr;
+  CeedVector_Cuda *impl;
+  ierr = CeedVectorGetData(vec, (void *)&impl); CeedChk(ierr);
+
+  switch(mtype) {
+  case CEED_MEM_HOST:
+    if (impl->memState == CEED_CUDA_DEVICE_SYNC) {
+      ierr = CeedSyncD2H_Cuda(vec); CeedChk(ierr);
+    }
+    (*array) = impl->h_array;
+    impl->h_array = NULL;
+    impl->h_array_allocated = NULL;
+    impl->memState = CEED_CUDA_HOST_SYNC;
+    break;
+  case CEED_MEM_DEVICE:
+    if (impl->memState == CEED_CUDA_HOST_SYNC) {
+      ierr = CeedSyncH2D_Cuda(vec); CeedChk(ierr);
+    }
+    (*array) = impl->d_array;
+    impl->d_array = NULL;
+    impl->d_array_allocated = NULL;
+    impl->memState = CEED_CUDA_DEVICE_SYNC;
+    break;
+  }
+
+  return 0;
+}
+
+//------------------------------------------------------------------------------
 // Set host array to value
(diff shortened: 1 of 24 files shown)


interface/ceed-vector.c
```diff
@@ -266,6 +265,41 @@ int CeedVectorSyncArray(CeedVector vec, CeedMemType mtype) {
...
+  if (vec->state % 2 == 1)
+    // LCOV_EXCL_START
+    return CeedError(vec->ceed, 1, "Cannot take CeedVector array, the access "
+                     "lock is already in use");
+  // LCOV_EXCL_STOP
+  if (vec->numreaders > 0)
+    // LCOV_EXCL_START
+    return CeedError(vec->ceed, 1, "Cannot take CeedVector array, a process "
+                     "has read access");
+  // LCOV_EXCL_STOP
```
jeremylt, line 294: It feels silly to test these when we already test access lock, but I can add a pair of tests for these if desired.

backends/cuda/ceed-cuda-vector.c
```diff
@@ -153,6 +153,39 @@ static int CeedVectorSetArray_Cuda(const CeedVector vec,
...
+  ierr = CeedVectorGetData(vec, (void *)&impl); CeedChk(ierr);
+
+  switch(mtype) {
+  case CEED_MEM_HOST:
+    if (impl->memState == CEED_CUDA_DEVICE_SYNC) {
+      ierr = CeedSyncD2H_Cuda(vec); CeedChk(ierr);
+    }
+    (*array) = impl->h_array;
+    impl->h_array = NULL;
+    impl->h_array_allocated = NULL;
```
jedbrown, line 172: Not a memory leak why?

jedbrown: Should it be invalid (and documented+enforced in the interface) to Take an array that is owned by the Vector?

jeremylt: I'm OK with the user taking ownership of (and management of) memory we allocated. Should I not be? Using the interface is declaring that you wish to take responsibility for the array. (probably should be added to the documentation)

I'm not sure where the leak would be if we assumed the user asked for responsibility and we transfer control over to them.

jedbrown: How does a user free device memory that they didn't allocate? Direct calls to `cudaFree` or `hipFree` (or ...)?

jeremylt: That is what I would expect. I expect this interface to be used when

1) the user called SetArray

Or

2) the user wants to take control of the array

1 would be the typical use. I could set it to be the only possible use pretty easily if we want to be more restrictive. It would not impact current usage.

I think we implicitly assume that the user knows what sort of device array is being used in the interface currently. If we are in the case where CEED_MEM_DEVICE could be allocated and freed differently on mixed device hardware we will only segfault at runtime if the user makes a mess. Otherwise the user should know how to free memory on their device.

jedbrown: Sounds fair. Can you just make sure the documentation is clear about the responsibility in that case?

valeriabarra:
Late comment: I have just realized that it could have been useful to update right away the release notes with this addition.

PR #579, opened 2020-07-07
Op - add interface for multigrid level creation

jeremylt:
This is a WIP to simplify the creation of multigrid operators (prolong, restrict, coarse) from a fine grid operator. Implementation should be straightforward. Refactoring the PETSc code might be a (mild) pain in the neck.


jeremylt:
I need to add tests and add this functionality to our multigrid and solids examples, but this should be pretty close to what I'm picturing for the interface.

I am expecting the user to provide
1) Coarse grid restriction (100% not libCEED's business)
2) Coarse to fine grid interpolation matrix (or enough info for me to make it)
3) Coarse grid basis (or enough info for me to make it)

From there I provide
1) Coarse grid operator (same quadrature space as fine grid)
2) Prolongation operator
3) Restriction operator

jeremylt:
The open ToDo is adding tests for the other 2 constructors and adding this to the Python interface.

jeremylt:
Ok, all new code is tested. The two remaining pieces are
- Decide the best way to scale for multiplicity
- Add functions to python interface

jeremylt:
ToDo Update:
- [ ] Clean up unused code in solids example
- [ ] Add functions to Python interface

@jedbrown, in my mind the new function interface is stable on the C/Fortran side. I'm not going to tag 'In-Review' until I finish the above ToDo, but if you have any comments on the interface, please let me know if you get a moment.

gallery/multigrid/ceed-multigrid1.h
```diff
@@ -0,0 +1,43 @@
...
+**/
+
+#ifndef multigrid1_h
+#define multigrid1_h
+
+CEED_QFUNCTION(Multigrid1)(void *ctx, const CeedInt Q,
+                           const CeedScalar *const *in,
+                           CeedScalar *const *out) {
+  // in[0] is input, size (Q*1)
+  // in[1] is multiplicity, size (Q)
```
valeriabarra, line 28: slightly inconsistent comment

tests/python/test-5-operator.py
```diff
@@ -1551,6 +1551,342 @@ def test_533(ceed_resource):
...
+    # Restrict state to coarse grid
+    op_restrict.apply(v_fine, v_coarse)
+
+    # Check
+    with v_coarse.array_read() as v_array:
+        total = 0.0
+        for i in range(nu_coarse * ncomp):
+            total = total + v_array[i]
+        assert abs(total - 2.0) < 1E-13
+
```
valeriabarra, line 1821: I'm a bit puzzled... what is the difference between this and the previous test? They have the same name and description

valeriabarra:
LGTM overall. One question: I guess the case `size = 2` was not relevant for the gallery?

gallery/multigrid/ceed-multigrid1.h
jeremylt: Updated

tests/python/test-5-operator.py
jeremylt: Weird, removed

valeriabarra: No problem! I just wanted to make sure I was not seeing double on my screen! 😅😂

jeremylt:
I didn't add size=2 because we don't have any current two field multigrid examples. I was thinking that we can easily add as needed but I didn't see a good reason to add it before we needed it

valeriabarra:
I think for completeness sake it would be better and more consistent to have it, since we have already in the gallery `mass1d`, `mass2d`, `mass3d`, and so on for the Poisson operator.

jeremylt:
Sure. I added the case and (hopefully) expanded the test coverage to handle this new case.

interface/ceed-qfunction.c
```diff
@@ -743,7 +743,7 @@ int CeedQFunctionDestroy(CeedQFunction *qf) {
   ierr = CeedFree(&(*qf)->inputfields); CeedChk(ierr);
   ierr = CeedFree(&(*qf)->outputfields); CeedChk(ierr);
   // Free ctx if identity
-  if ((*qf)->identity) {
+  if ((*qf)->ownctx) {
```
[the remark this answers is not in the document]
jedbrown: Should we use a similar pattern to `CeedVector`, where we hold the ownership pointer separately, instead of a `bool`?

interface/ceed-operator.c
```diff
@@ -983,6 +1194,183 @@ int CeedOperatorLinearAssembleAddPointBlockDiagonal(CeedOperator op,
...
+  @param[out] opProlong   Coarse to fine operator
+  @param[out] opRestrict  Fine to coarse operator
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedOperatorMultigridLevelCreateTensorH1(CeedVector PMultFine,
+    CeedElemRestriction rstrCoarse, CeedBasis basisCoarse,
+    const CeedScalar *interpCtoF, CeedOperator opFine,
```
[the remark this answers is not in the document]
jeremylt: No reason. I agree its more natural to have opFine first.

interface/ceed-operator.c
```diff
@@ -197,6 +197,211 @@ int CeedOperatorSingleView(CeedOperator op, bool sub, FILE *stream) {
...
+  @brief Find the active vector basis for a CeedOperator
+
+  @param[in] op            CeedOperator to find active basis for
+  @param[out] activeBasis  Basis for active input vector
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ ref Developer
+**/
+static inline int CeedOperatorGetActiveBasis(CeedOperator op,
```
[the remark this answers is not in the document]
jeremylt: Good point. Updated.

gallery/multigrid/ceed-multigrid2.h
```diff
@@ -0,0 +1,47 @@
...
+CEED_QFUNCTION(Multigrid2)(void *ctx, const CeedInt Q,
+                           const CeedScalar *const *in,
+                           CeedScalar *const *out) {
+  const CeedInt size = 2;
+
+  // in[0] is input, size (Q*2)
+  // in[1] is multiplicity, size (Q*2)
+  const CeedScalar *input = in[0];
+  const CeedScalar *mult = in[1];
+  // out[0] is output, size (Q*1)
```
jedbrown, line 33:
```suggestion
  // out[0] is output, size (Q*2)
```

jeremylt:
I can't force an operator to call the QFunction with Q*size instead of Q (in our current interface), and copying the context slows down the cuda backends, which is why I made the 123 versions.

I agree about the naming.

I considered adding a CeedVectorReciprocal to make the QFunction simpler. That would be easy to add.

jedbrown:
Hmm, could probably abuse the `void *` to pass the integer, but I guess that's okay. It's a one-time cost, right? I'd do the reciprocal and then just call these `gallery/scale/` or something like that (because it's all they do).

jeremylt:
Because context can change between operator applications, we have to pay the cost to transfer `scale` on every application

interface/ceed-operator.c
```diff
@@ -983,6 +1194,183 @@ int CeedOperatorLinearAssembleAddPointBlockDiagonal(CeedOperator op,
   return 0;
 }
 
+/**
+  @brief Create a multigrid coarse operator and level transfer operators
+           for a CeedOperator with a Lagrange tensor basis for the active basis
+
+  @param[in] PMultFine    L-vector multiplicity in parallel gather/scatter
+  @param[in] rstrCoarse   Coarse grid restriction
+  @param[in] degreeCoarse Coarse grid basis polynomial order
```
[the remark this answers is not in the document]
jeremylt: I am happy with the other changes - this should be the last open question, I believe.

For the MFEM integration, I'll want to use their 'transfer matrix', so I think that `CeedOperatorMultigridLevelCreateTensorH1` and `CeedOperatorMultigridLevelCreateH1` are what I will use for that.

That leaves replacing `CeedOperatorMultigridLevelCreateTensorH1Lagrange` with `CeedOperatorMultigridLevelCreate` where the transfer basis is automatically created out of the fine and coarse bases. It would be pretty flexible and would cover non-tensor and tensor bases. I think I should be able to use some of the same code in GetCollocatedGrad.

jeremylt:
Note - Travis currently failing due to some problem I can't reproduce locally in t202 with destroying the vectors. Will investigate more tomorrow.

jedbrown:
Surely we could track whether the context has actually changed. Maybe hash it on the CPU so we have something easy to compare with, or require that the user notify after changes.

jeremylt:
We have an issue discussing this, which I was alluding to above.

In theory the user or the QFunction can update the context, so context requires two copies, host->device and device->host. (not sure if the second one is actually in the GPU backends) 

In practice our context has been largely static, so we discussed having two types of context, static and dynamic. I could remove the 123 variants and followup with a separate PR to finally update our context like we discussed in Dec 2018.

interface/ceed-operator.c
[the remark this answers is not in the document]
jeremylt: Update - should be ready to go now

jeremylt:
Note: need to fix reciprocal, none sync makes no sense

jedbrown:
I'm not sure I follow about the sync. I'd expect it should not need to eagerly sync because the next usage is usually still on the device.

jeremylt:
That was a note to future me to fix the cuda/hip backend reciprocal function because taking the reciprocal of a vector that has a vector in state CEED_CUDA/HIP_SYNC_NONE will result in gibberish.

jeremylt:
Ok, all fixed and ready to merge.

I added the new interface that creates the prolongation basis from the fine and coarse grid interpolation, and I fixed the couple of small issues I made a note about above.

jeremylt:
Yeah, I was thinking of leaving the 'raw' example to expose the details more easily for users.

jeremylt:
Merge button currently blocked by ReadTheDocs glitch

jeremylt:
@jedbrown I added the allowed ARM failure, this this branch is ready to merge.

CodeCov claims "A common practice in the modern developer's workflow is to test multiple build systems, split up tests in different containers, and group tests based on test focus (unittests vs integration tests). Codecov flawlessly supports all these setups. Therefore, no setup is required."
but that isn't actually working right, so there is a false decrease in coverage blocking the mergability of the branch now. I'm investigating but I don't know if/when I'll be able to fix.

jedbrown:
I've also had issues with that and the docs are useless for debugging. Is there a way to report bugs to codecov?

jeremylt:
They have a user forum. I'm planning on posting there asking for help when I get a chance.

jeremylt:
I rebased for the changes on main

jedbrown:
Is this branch going to be squashed?

jeremylt:
I was planning on doing one big squash.

jedbrown:
Go ahead as far as I'm concerned.

jeremylt:
Intel changed something and Travis is broken. Will investigate tomorrow

jeremylt:
I had a realization and was able to fix it.

jeremylt:
GCOV made big changes in the reports for GCC 8 (and then again for GCC 9). I'm trying to run Travis GCOV with GCC 8 like Noether has.

jeremylt:
OK, so it looks like isse with CodeCov is that GCC 8 is making wrong reports in our current configuration.

jedbrown:
Any reason not to use gcc-9 or gcc-10 or clang-10 for the C compiler?

jeremylt:
G Fortran 9\10 have the compiler bug.

jedbrown:
Ok, bummer that gcov can't use gfortran-8 with more recent gcc. So I guess we should just stick with everything from gcc-8 suite until the bug is fixed. I can do an admin merge when you're ready.

jeremylt:
I think I have a workaround for the Fortran issue, but I feel like I should play with this in a separate branch. I stripped those experimental commits. This is ready to squash+merge and I'll tinker on a different branch.

PR #580, opened 2020-07-08
Rename master branch to main

jedbrown:
This may be a small gesture, but it's easy and I haven't seen any names rise to become contenders with `main`.

https://www.zdnet.com/article/github-to-replace-master-with-alternative-term-to-avoid-slavery-references/

If approved, I'll set the default branch here and update codecov and readthedocs. It'll require current developers to do the following once. (Fresh clones will require no extra steps.)

    git fetch
    git branch -m main
    git branch --set-upstream-to=origin/main

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,8 +2,8 @@
 
 Contributions to libCEED are encouraged.
 <!---
-Please use a pull request to the appropriate branch ('maint' for
-backward-compatible bug fixes for the last stable release, 'master' for
+Please use a pull request to the appropriate branch ('stable' for
+backward-compatible bug fixes for the last stable release, main' for
 new features and everything else).
 -->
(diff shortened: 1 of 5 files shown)


valeriabarra:
Cool. I like the new naming convention. 
So existing branches that were already forked off the old master don't need any update? Even if PRs are already open?

jedbrown:
Target branches in open PRs will likely have to be updated, but I think that's it. The branches themselves don't need any changes.

valeriabarra:
Okay, I'm not sure how to do that in the PR web interface. The only thing I can do is click on the target branch, which will just send me to the current main page. Perhaps something in the PR interface will change later when we deploy this change?

jedbrown:
You just click "Edit" right of the PR title and select the target branch. I'll merge now and handle readthedocs and codecov.

PR #583, opened 2020-07-13
CUDA Vector Norm on Device

jeremylt:
This is a PR to indicate that I'd like to include a CUDA implementation of vector norms in the v0.7 release. This will impact performance for diagonal assembly in GPU backends.

The commit so far is just some initial efforts on learning reduction kernels in CUDA - nothing works yet.


jeremylt:
Calling directly to cublas was much simpler that my first attempt at rolling my own reduction kernels. This passes the tests and make style/tidy locally.

backends/cuda/ceed-cuda-vector.c
```diff
@@ -348,6 +348,55 @@ static int CeedVectorRestoreArray_Cuda(const CeedVector vec) {
...
+  Ceed_Cuda *ceed_Cuda;
+  ierr = CeedGetData(ceed, (void *) &ceed_Cuda); CeedChk(ierr);
+  CeedVector_Cuda *data;
+  ierr = CeedVectorGetData(vec, (void *)&data); CeedChk(ierr);
+  CeedInt length;
+  ierr = CeedVectorGetLength(vec, &length); CeedChk(ierr);
+  if (!ceed_Cuda->cublasHandle) {
+    ierr = cublasCreate(&ceed_Cuda->cublasHandle); CeedChk_Cublas(ceed, ierr);
+  }
+  cublasHandle_t handle = ceed_Cuda->cublasHandle;
```
jedbrown, line 368: Could we capture this pattern?
```suggestion
  cublasHandle_t handle;
  ierr = CeedCudaGetCublasHandle(ceed, &handle); CeedChk(ierr);
```

jeremylt: argeed

backends/cuda/ceed-cuda.h
```diff
@@ -40,13 +41,46 @@ do { \
...
+
 #define QUOTE(...) #__VA_ARGS__
 
-typedef enum {CEED_CUDA_HOST_SYNC,
-              CEED_CUDA_DEVICE_SYNC,
-              CEED_CUDA_BOTH_SYNC,
-              CEED_CUDA_NONE_SYNC
-             } CeedCudaSyncState;
+static const char *cublasGetErrorName(cublasStatus_t error) {
+  switch (error) {
```
jeremylt, line 56: Its annoying that this error code -> char utility doesn't ship with CUBLAS

jedbrown: Agreed. You can use a macro like
```c
#define CASE(name) case name: return #name
```
and use like
```c
CASE(CUBLAS_STATUS_SUCCESS);
CASE(CUBLAS_STATUS_NOT_INITIALIZED);
...
```
but it still sucks. The same thing can go in an array with
```
const char *const cublas_status[] = {
  [CUBLAS_STATUS_SUCCESS] = "CUBLAS_STATUS_SUCCESS",
  ...
};
```
which may be close to what the compiler does with your switch statement. But in any case, it'd be much better provided by cuBLAS.

jeremylt: I like that macro better. At least it doesn't eat as much space. Thanks

PR #584, opened 2020-07-14
Python - add operator diagonal assembly to the python interface

### LINKED ISSUE #581: Python Diagonal Assembly
I forgot to add `CeedOperatorLinearAssemble*Diagonal()` to the Python interface. It should be quick to add these two methods + two tests.

Anyone is welcome to jump on this if they have some time and I haven't done it yet.

jeremylt:
closes #581

### DIFF
--- a/python/ceed_operator.py
+++ b/python/ceed_operator.py
@@ -54,6 +54,58 @@ class _OperatorBase(ABC):
 
         return out_string
 
+    # Assemble linear diagonal
+    def linear_assemble_diagonal(self, d, request=REQUEST_IMMEDIATE):
+        """Assemble the diagonal of a square linear Operator
+
+           Args:
+             d: Vector to store assembled Operator diagonal
+             **request: Ceed request, default CEED_REQUEST_IMMEDIATE"""
+
+        # libCEED call
+        lib.CeedOperatorLinearAssembleDiagonal(self._pointer[0],
+            d._pointer[0], request)
+
+    # Assemble add linear diagonal
+    def linear_assemble_add_diagonal(self, d, request=REQUEST_IMMEDIATE):
+        """Sum the diagonal of a square linear Operator into a Vector
+
+           Args:
+             d: Vector to store assembled Operator diagonal
+             **request: Ceed request, default CEED_REQUEST_IMMEDIATE"""
+
+        # libCEED call
+        lib.CeedOperatorLinearAssembleAddDiagonal(self._pointer[0],
+            d._pointer[0], request)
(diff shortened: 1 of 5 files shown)


valeriabarra:
Thanks for these additions Jeremy! Would it be worth to expand our operator tutorial with a snippet from the test you added?

python/ceed_operator.py
```diff
@@ -54,6 +54,58 @@ def __str__(self):
...
+                  at each node
+             **request: Ceed request, default CEED_REQUEST_IMMEDIATE"""
+
+        # libCEED call
+        lib.CeedOperatorLinearAssemblePointBlockDiagonal(self._pointer[0],
+            d._pointer[0], request)
+
+    # Assemble linear point block diagonal
+    def linear_assemble_add_point_block_diagonal(self, d, request=REQUEST_IMMEDIATE):
+        """Sum the diagonal of a square linear Operator into a Vector
```
valeriabarra, line 97: This has the same docstring of the `linear_assemble_add_diagonal` above

valeriabarra: And the same thing for `linear_assemble_point_block_diagonal`

jeremylt: Fixed that, thank you

jeremylt:
Since the tutorial only has a single application of the operator to compute volume of the domain, I am not sure where adding the diagonal assembly would fit in well. Do you have thoughts on that?

valeriabarra:
No, not any particular thoughts. Maybe it's just worth to mention these new functions as available in a textbox instead of a code block example?

tests/python/test-5-operator.py
```diff
@@ -1498,3 +1498,100 @@ def test_524(ceed_resource):
         assert abs(total - 1.0) < 1E-10
 
 # -------------------------------------------------------------------------------
+# Test assembly of mass matrix operator diagonal
```
valeriabarra, line 1501: Not concerning this comment in particular, but while at it, would you please double check all others comments from earlier tests in this series? Some of them appear to be the same string for tests that do slightly different things.

jeremylt: I updated some to better descriptions.

valeriabarra: Thanks! Some of them still have the same descriptions, but I can't spot at a first glance now what are the differences in those tests

jeremylt: They are the same test, but with 0 vs 1 as the input vector

PR #585, opened 2020-07-14
Fortran - guard double calls to ceed*destroy in Fortran interface

jeremylt:
Unlike with the C interface, multiple calls to `Ceed*Destroy()` can cause crashes. This fixes this small issue.

### DIFF
--- a/interface/ceed-fortran.c
+++ b/interface/ceed-fortran.c
@@ -102,7 +102,7 @@ void fCeedView(int *ceed, int *err) {
 
 #define fCeedDestroy FORTRAN_NAME(ceeddestroy,CEEDDESTROY)
 void fCeedDestroy(int *ceed, int *err) {
-  if (Ceed_n == 0) return;
+  if (Ceed_n == 0 || !Ceed_dict[*ceed]) return;
   *err = CeedDestroy(&Ceed_dict[*ceed]);
 
   if (*err == 0) {
@@ -211,7 +211,7 @@ void fCeedVectorView(int *vec, int *err) {
 
 #define fCeedVectorDestroy FORTRAN_NAME(ceedvectordestroy,CEEDVECTORDESTROY)
 void fCeedVectorDestroy(int *vec, int *err) {
-  if (CeedVector_n == 0) return;
+  if (CeedVector_n == 0 || !CeedVector_dict[*vec]) return;
   *err = CeedVectorDestroy(&CeedVector_dict[*vec]);
 
   if (*err == 0) {
@@ -433,7 +433,7 @@ void fCeedRequestWait(int *rqst, int *err) {
 #define fCeedElemRestrictionDestroy \
     FORTRAN_NAME(ceedelemrestrictiondestroy,CEEDELEMRESTRICTIONDESTROY)
 void fCeedElemRestrictionDestroy(int *elem, int *err) {
-  if (CeedElemRestriction_n == 0) return;
+  if (CeedElemRestriction_n == 0 || !CeedElemRestriction_dict[*elem]) return;
   *err = CeedElemRestrictionDestroy(&CeedElemRestriction_dict[*elem]);
 
   if (*err == 0) {
@@ -573,7 +573,7 @@ void fCeedBasisGetNumQuadraturePoints(int *basis, int *Q, int *err) {
 
 #define fCeedBasisDestroy FORTRAN_NAME(ceedbasisdestroy,CEEDBASISDESTROY)
 void fCeedBasisDestroy(int *basis, int *err) {
-  if (CeedBasis_n == 0) return;
+  if (CeedBasis_n == 0 || ! CeedBasis_dict[*basis]) return;
   *err = CeedBasisDestroy(&CeedBasis_dict[*basis]);
 
   if (*err == 0) {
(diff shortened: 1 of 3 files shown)


tests/t000-ceed-f.f90
```diff
@@ -10,6 +10,7 @@ program test
 
       call ceedinit(trim(arg)//char(0),ceed,err)
       call ceeddestroy(ceed,err)
+      call ceeddestroy(ceed,err)
```
jeremylt, line 13: I added this to test the double destroy case. I'm happy having it stay or go.

valeriabarra: Maybe just leave a comment, so that a reader in the future does not interpret this as a possible copy-paste mistake?

jeremylt: Good idea. Updated

jeremylt:
Maybe I missed something, but that `destroy` for `b2` should occur and run fine on this branch.

```Fortran
integer b1, b2;
! create both b1 and b2
call ceedbasisdestroy(b1, e) ! This works fine
call ceedbasisdestroy(b1, e) ! This immediately returns, nothing happens
call ceedbasisdestroy(b2, e) ! This works fine
```

I put this PR together because I noticed a discrepancy between the C and Fortran API wrt duplicate destroys.

On the C side of the house, the following code works without any errors
```C
CeedBasis b1, b2;
// Create both b1 and b2
CeedBasisDestroy(&b1);
CeedBasisDestroy(&b1);
CeedBasisDestroy(&b2);
```
but on main the Fortran equivalent fails 
```
# +Program received signal SIGSEGV: Segmentation fault - invalid memory reference.
# +
# +Backtrace for this error:
# +#0  0x7f5c933093df in ???
# +#1  0x7f5c9394a961 in CeedBasisDestroy
# +at /home/jeremy/Dev/libCEED/interface/ceed-basis.c:950
# +#2  0x7f5c9394f3cc in ceedbasisdestroy_
# +at /home/jeremy/Dev/libCEED/interface/ceed-fortran.c:573
# +#3  0x5603afcfc3c2 in test
# +at /home/jeremy/Dev/libCEED/tests/t306-basis-f.f90:37
# +#4  0x5603afcfc515 in main
# +at /home/jeremy/Dev/libCEED/tests/t306-basis-f.f90:40
# +tests/tap.sh: line 118: 60823 Segmentation fault      (core dumped) build/$1 ${args/{ceed_resource}/$backend}
```
With this PR, I don't get that segfault. The test runs clean
```
==61538== Memcheck, a memory error detector
==61538== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==61538== Using Valgrind-3.16.0.GIT and LibVEX; rerun with -h for copyright info
==61538== Command: ./build/t306-basis-f /cpu/self/ref/serial
==61538== 
==61538== 
==61538== HEAP SUMMARY:
==61538==     in use at exit: 8 bytes in 1 blocks
==61538==   total heap usage: 64 allocs, 63 frees, 94,251 bytes allocated
==61538== 
==61538== LEAK SUMMARY:
==61538==    definitely lost: 0 bytes in 0 blocks
==61538==    indirectly lost: 0 bytes in 0 blocks
==61538==      possibly lost: 0 bytes in 0 blocks
==61538==    still reachable: 8 bytes in 1 blocks
==61538==         suppressed: 0 bytes in 0 blocks
==61538== Rerun with --leak-check=full to see details of leaked memory
==61538== 
==61538== For lists of detected and suppressed errors, rerun with: -s
==61538== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

interface/ceed-fortran.c
```diff
@@ -102,7 +102,7 @@ void fCeedView(int *ceed, int *err) {
 
 #define fCeedDestroy FORTRAN_NAME(ceeddestroy,CEEDDESTROY)
 void fCeedDestroy(int *ceed, int *err) {
-  if (Ceed_n == 0) return;
+  if (Ceed_n == 0 || !Ceed_dict[*ceed]) return;
```
jeremylt, line 108: Ah, before this commit I would have had exactly the issue you described.

jedbrown: Oh, I see. There are still inconsistencies, but hard to fix unless we're allowed to ask Fortran for an address-sized argument.
```fortran
ceed c1,c2
call ceedinit(resource, c1, err)
call ceeddestroy(c1, err)
call ceedinit(resource, c2, err) ! reuses old index value
call ceeddestroy(c1, err) ! actually destroys c2
```

The type safety is so much better in recent versions of Fortran, it makes me sad we're stuck with this.

jeremylt: I was puzzling on how to prevent that case. I think I can add a small fix to prevent that case!

interface/ceed-fortran.c
```diff
@@ -105,10 +105,11 @@ void fCeedView(int *ceed, int *err) {
 
 #define fCeedDestroy FORTRAN_NAME(ceeddestroy,CEEDDESTROY)
 void fCeedDestroy(int *ceed, int *err) {
-  if (Ceed_n == 0 || !Ceed_dict[*ceed]) return;
+  if (Ceed_n == 0 || *ceed == FORTRAN_NULL || !Ceed_dict[*ceed]) return;
```
jedbrown: Could we simplify to this?
```suggestion
  if (*ceed == FORTRAN_NULL || !Ceed_dict[*ceed]) return;
```

jeremylt: I also cut the third condition in my latest commit. I think `*ceed == FORTRAN_NULL` is the only test needed

jeremylt:
Will merge after CI

PR #586, opened 2020-07-16
CUDA - Fix Backend Setting LinearAssemble*

jeremylt:
Small bug I added; wasn't caught because the function was accidentally generating correct results.

@nbeams, this should fix the issue you were seeing if you add the same fix to `hip-ref`.

### DIFF
--- a/backends/cuda/ceed-cuda-operator.c
+++ b/backends/cuda/ceed-cuda-operator.c
@@ -1132,7 +1132,7 @@ static inline int CeedOperatorAssembleDiagonalCore_Cuda(CeedOperator op,
 //------------------------------------------------------------------------------
 // Assemble composite diagonal common code
 //------------------------------------------------------------------------------
-static inline int CeedOperatorLinearAssembleDiagonalCompositeCore_Cuda(
+static inline int CeedOperatorLinearAssembleAddDiagonalCompositeCore_Cuda(
   CeedOperator op, CeedVector assembled, CeedRequest *request,
   const bool pointBlock) {
   int ierr;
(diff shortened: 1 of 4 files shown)


backends/cuda/ceed-cuda-operator.c
```diff
@@ -1204,11 +1204,11 @@ int CeedOperatorCreate_Cuda(CeedOperator op) {
...
                                 CeedOperatorLinearAssembleQFunction_Cuda);
   CeedChk(ierr);
-  ierr = CeedSetBackendFunction(ceed, "Operator", op, "LinearAssembleDiagonal",
+  ierr = CeedSetBackendFunction(ceed, "Operator", op, "LinearAssembleAddDiagonal",
                                 CeedOperatorLinearAssembleDiagonal_Cuda);
   CeedChk(ierr);
   ierr = CeedSetBackendFunction(ceed, "Operator", op,
-                                "LinearAssemblePointBlockDiagonal",
+                                "LinearAssembleAddPointBlockDiagonal",
                                 CeedOperatorLinearAssemblePointBlockDiagonal_Cuda);
```
jedbrown, line 1212: Should the function names be chosen to match so this doesn't cause confusion in the future?

jeremylt: Reasonable. We should really do the same with Apply as well.

jeremylt: Updated. It wasn't there before because I was being sloppy when I updated Apply->ApplyAdd and Assemble -> AssembleAdd in the interface.

PR #593, opened 2020-07-23
Travis - skip OCCA on non-AMD64 hardware to avoid random failures

jeremylt:
Any objection to dropping the OCCA testing for the ARM and IBM chips to prevent the frequent but inconsistent failures? This has been making our CI less useful.

We would drop this change when the new OCCA backend is merged.


jeremylt:
Travis is finishing - this reliably passes now

jeremylt:
Agreed

PR #594, opened 2020-07-24
backends/hip: exclude testing of unimplemented code

jedbrown:


jeremylt:
Following our usual conventions, we'd exclude:
operator
  465-470
  1186-1190
qfunction
  115
  125
  137-139
restriction
  287
header
  47-59

I'm about to stop for the weekend, but I can contribute on Monday and chase down the equivalent parts of the cuda backends as well

jedbrown:
I shot myself in the foot by transferring the repository from gitlab.com/jedbrown to gitlab.com/libceed and thereby losing the free CI/CD until we can get open source project status.

valeriabarra:
What do you mean by "until we can get open source project status"?
Also, was this the reason to preferring gitlab.com instead of github.com? Probably this was discussed a year ago and I missed it

jedbrown:
Working here now: https://gitlab.com/libceed/libCEED/-/pipelines

The check name includes the branch and is thus different for each PR, so we can't set it as "required".

@valeriabarra This GitLab is just for running CI on Noether (and perhaps eventually other hosts). GitHub has an "actions" feature now, but my understand is that it is closed-source and has security issues for public repositories. GitLab-CI seems a better fit for us, and is what we use for PETSc.

Makefile
```diff
@@ -201,19 +201,19 @@ avx.c          := $(sort $(wildcard backends/avx/*.c))
 xsmm.c         := $(sort $(wildcard backends/xsmm/*.c))
 cuda.c         := $(sort $(wildcard backends/cuda/*.c))
 cuda.cpp       := $(sort $(wildcard backends/cuda/*.cpp))
-cuda.cu        := $(sort $(wildcard backends/cuda/*.cu))
+cuda.cu        := $(sort $(wildcard backends/cuda/kernels/*.cu))
```
jeremylt, line 204: @nbeams and I talked about this idea in Slack. It looks like the new OCCA backend will be similarly organized.

jeremylt:
Travis needs to catch up, but it looks good on Noether

jedbrown:
Cool. Should we add testing with PETSc on Noether? (This PR or another?)

jeremylt:
I was thinking a separate PR because fixing the PETSc examples to work with hip vectors will take a bit of work.

PR #596, opened 2020-07-27
QFunction Context Data Object

### LINKED ISSUE #183: CeedQFunctionContext synchronisation logic
I have issues with the Fortran QFunction/Operator tests because a context magically appears in the QFunction. I think that the `CeedQFunctionSetContext` function should call a backend function that sets/updates the `ctx`, otherwise it's difficult to avoid unnecessary systematic `memCpy` for a GPU backend. This drives us to the question of knowing if we would want to have the same memory access logic for the ctx as for the Vectors, with `GetCtx()`, `GetCtxRead()`, `RestoreCtx()` and `RestoreCtxRead()`.
Besides, I feel it's wrong that the Fortran interface is using the ctx as it's meant to be used by the backend (If I understood correctly). If a backend wants to use the context for some reason, it is currently impossible as the Fortran interface is already using it.
In my case the tests fail because a context has been created by the Fortran interface after the creation of the QFunction, I imagine I could memCpy the data on the GPU, but this data is actually useless for my backend, which makes me really think that there is a design flaw here.

jeremylt:
This adds a libCEED context data object. This should offer a mild performance improvement for the GPU backends as it will reduce unneeded memcopies.

Closes #183


jeremylt:
ToDo:
 - [ ] OCCA impl
 - [ ] Python impl

jeremylt:
My biggest question is - Do we want to call this `CeedUserContext` or `CeedQFContext` or something else? Names are hard.

jeremylt:
Cuda test pass locally, other backends covered by CI

jedbrown:
Probably `CeedQFunctionContext`. I wonder if we should allow a user-defined function to manage copying to/from a device, in particular, if the user wants to put a function pointer or other dynamic structure in there.

jeremylt:
I also like `CeedQFunctionContext` better. Renamed.

CI failing due to the #600 - I'll rebase once that merges.

Do we want to add user functions to manage copying once we have a use case?

backends/cuda/ceed-cuda-qfunctioncontext.c
```diff
@@ -0,0 +1,242 @@
...
+  case CEED_MEM_DEVICE:
+    return CeedQFunctionContextSetDataDevice_Cuda(ctx, cmode, data);
+  }
+  return 1;
+}
+
+//------------------------------------------------------------------------------
+// Get array
+//------------------------------------------------------------------------------
+static int CeedQFunctionContextGetData_Cuda(const CeedQFunctionContext ctx,
```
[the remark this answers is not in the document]
jeremylt: Context data is writeable in the QFunction so I only provided a writeable interface. I don't see a read-only use case internally at the moment.

jedbrown:
I think it's fine to wait on a use case. There are basically two approaches to modularity. One is to allow multiple qfunctions, forming a DAG from the inputs (state variables plus aux) to the outputs (residual terms, perhaps cached terms for future use). The other is to allow the qfunction to make arbitrary function calls. Only the latter is feasible for things like modular equations of state, where it might be called in a loop (e.g., within a Riemann solver). In the current model, it can be done by templating, which either means you have to recompile to select the equation of state (in this example) or instantiate all the variants. Better for usability is to put the function pointers in the context.

jeremylt:
I found the leak - thank you @nbeams!

backends/cuda/ceed-cuda-qfunctioncontext.c
```diff
@@ -0,0 +1,242 @@
...
+
+  ierr = cudaMemcpy(impl->d_data, impl->h_data, bytes(ctx),
+                    cudaMemcpyHostToDevice); CeedChk_Cu(ceed, ierr);
+  return 0;
+}
+
+//------------------------------------------------------------------------------
+// Sync device to host
+//------------------------------------------------------------------------------
+static inline int CeedSyncD2H_Cuda(const CeedQFunctionContext ctx) {
```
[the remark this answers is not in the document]
jeremylt: The functions are scoped only to those files, but I'll go ahead and change the name.

jeremylt:
And yes, this will require another MFEM update. I have a MFEM + libCEED refactor PR planned to facilitate another PR, so I will catch this change in that PR.

jeremylt:
I made a couple of small fixed ID'd by CoveCov.

python/ceed.py
```diff
@@ -340,6 +341,14 @@ def IdentityQFunction(self, size, inmode, outmode):
 
         return IdentityQFunction(self, size, inmode, outmode)
 
+    def QFunctionContext(self):
+        """Ceed User Context: stores Ceed QFunction user context data.
+
+           Returns:
+             userContext: Ceed Identity QFunction"""
```
valeriabarra, line 348: Why `Identity QFunction` here?

valeriabarra:
This overall looks good to me. I only have a question. Was there a specific need to have `ceed-qfunctioncontext.c` separate from `ceed-qfunction.c` (and so on for the other interfaces)? To me that would semantically follow under the same class, as a sort of derived object.

python/ceed.py
jeremylt: fixed

jeremylt:
It could go in the same file - I put it in a separate file like we did with the tensor contraction object because I find one class per file clearer, personally.

valeriabarra:
At least for the Python interface, I would put it under `ceed_qfunction.py` (moving also relative tests and tutorials). I'd see it semantically as a derived object, and for the Python interface it is nice that we have the main 6 classes of the API (ceed, vector, basis, restriction, qfunction, opeartor; plus some build and costant files)

jeremylt:
For testing and tutorials, the context object is grouped with the qfunction. I only separated the classes in the implementation files.

valeriabarra:
> For testing and tutorials, the context object is grouped with the qfunction. I only separated the classes in the implementation files.

Yeah, sorry, and I meant that I think that's good, and it would make sense to homogenize the implementation file to follow that, too :)

jeremylt:
So - what is the general consensus for this branch? I'd like to build the Rust interface off of this branch as a baseline

Also - I think it would be nice to have this merged before Ceed4AM if possible so we can talk about the new interface in the tutorial

python/ceed_qfunction.py
```diff
@@ -112,15 +112,12 @@ def set_context(self, ctx):
...
-             *ctx: Numpy array holding context data to set"""
-
-        # Setup the numpy array for the libCEED call
-        ctx_pointer = ffi.new("CeedScalar *")
-        ctx_pointer = ffi.cast("void *", ctx.__array_interface__['data'][0])
+             ctx: Ceed User Context object holding context data"""
 
         # libCEED call
         err_code = lib.CeedQFunctionSetContext(
-            self._pointer[0], ctx_pointer, len(ctx))
```
jeremylt, line 123: Using `len(ctx)` was actually a bug that was missed with the GPU testing somehow.

interface/ceed-fortran.c
```diff
@@ -763,12 +860,18 @@ void fCeedQFunctionAddOutput(int *qf, const char *fieldname,
...
-  fctx->innerctx = ctx;
-  fctx->innerctxsize = ((size_t) *n)*sizeof(CeedScalar);
+  CeedUserContext fctx;
+  *err = CeedQFunctionGetContext(qf_, &fctx);
+  if (*err) return;
+  CeedFortranContext fctxdata;
+  *err = CeedUserContextGetData(fctx, CEED_MEM_HOST, (void **)&fctxdata);
+  if (*err) return;
+  fctxdata->innerctx = ctx_;
+  *err = CeedUserContextRestoreData(fctx, (void **)&fctxdata);
```
jeremylt: I'm still not super happy about Fortran hijacking the context to make the interface work, but I couldn't think of a better way to do this given the limitations of Fortran77

interface/ceed-fortran.c
```diff
@@ -631,6 +631,85 @@ void fCeedLobattoQuadrature(int *Q, CeedScalar *qref1d, CeedScalar *qweight1d,
...
+#define fCeedQFunctionContextSetData \
+    FORTRAN_NAME(ceedqfunctioncontextsetdata,CEEDQFUNCTIONCONTEXTSETDATA)
+void fCeedQFunctionContextSetData(int *ctx, int *memtype, int *copymode,
+                                  CeedInt *n,
+                                  CeedScalar *data, int64_t *offset, int *err) {
+  size_t ctxsize = ((size_t) *n)*sizeof(CeedScalar);
+  *err = CeedQFunctionContextSetData(CeedQFunctionContext_dict[*ctx],
+                                     (CeedMemType)*memtype,
+                                     (CeedCopyMode)*copymode, ctxsize,
+                                     (CeedScalar *)(data + *offset));
```
jedbrown, line 668:
```suggestion
                                     data + *offset);
```
Doesn't need a cast.

jedbrown: Is it necessary for `offset` to be part of this interface? I guess it's for symmetry with `CeedVectorSetArray` because it might have been obtained from a different place. But the user can always pass the offset value `data(offset)` in Fortran.

interface/ceed-qfunctioncontext.c
```diff
@@ -0,0 +1,306 @@
...
+  @note The CeedQFunctionContextGetData() and @ref CeedQFunctionContextRestoreData() functions
+    provide access to array pointers in the desired memory space. Pairing
+    get/restore allows the Context to track access.
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedQFunctionContextGetData(CeedQFunctionContext ctx, CeedMemType mtype,
+                                void **data) {
```
jedbrown, line 206: Needs Fortran usage notes. Also, consider using `void *` (like we do in `CeedMallocArray`) instead of `void **`.

interface/ceed-fortran.c
jeremylt: I'm not exactly convinced it needs to be in the CeedVector either, but you are correct, I added it for symmetry.

interface/ceed-qfunctioncontext.c
jeremylt: I added better usage notes to `CeedQFunctionGetContext` and `CeedQFunctionGetInnerContext`

Re `void *` vs `void **`, we use `void **` quite a bit across the library. I'm not sure I understand why we do so

jedbrown: So `void **` has the right arity: we pass the address of a pointer to (unknown to us) data. However, `void **` always needs a cast because `Thing **` does not implicitly convert to `void *` so it *always* needs an explicit cast. So other libraries, including MPI (cf. `MPI_Comm_get_attr`) and PETSc, just use `void *` and document that one should pass the address of the pointer. C and C++ implicitly converts to `void *`, so no cast is necessary. (C implicitly converts _from_ `void *` while C++ does not.)

jeremylt: That's what I thought. So should we just pick one and stick with it for the whole library? It would be tedious but easy to change over.

jedbrown: I refer to avoid casts except in unusual circumstances so I'd be in favor of using `void *` for this purpose despite it having the "wrong" arity.

jeremylt: Ok, should be good now.

jeremylt:
Any further comments?

examples/fluids/Makefile
```diff
@@ -31,9 +31,26 @@ navierstokes := $(navierstokes.c:%.c=%)
...
+# if TERM=dumb, use it, otherwise switch to the term one
+output = $(if $(TERM:dumb=),$(call color_out,$1,$2),$(call emacs_out,$1,$2))
+
+# if V is set to non-nil, turn the verbose mode
+quiet = $(if $(V),$($(1)),$(call output,$1,$@);$($(1)))
+
 # Rules for building the example
 %: %.c
-	$(CC) $(CFLAGS) $(CPPFLAGS) $(LDFLAGS) $(CEED_LDFLAGS) $(abspath $<) -o $@ $(LDLIBS)
+	$(call quiet,CC) $(CFLAGS) $(CPPFLAGS) $(LDFLAGS) $(CEED_LDFLAGS) $(abspath $<) -o $@ $(LDLIBS)
```
jedbrown, line 53: I fear this makes the example makefiles scary. Perhaps just
```makefile
-include common.mk
quiet ?= $($(1))
```
So this would still work if the user relocated it (albeit not being quiet), but this way we don't have multiple variants to maintain and nothing is too scary.

jeremylt: Agreed. I'll fix that in the morning

jeremylt: Fixed

jedbrown:
Let me know if you want to do any squashing or wait for anyone else to review.

jeremylt:
There are a fair number of little intermediate commits, so I'm fine just squash merging the whole thing now.

PR #599, opened 2020-07-30
Cuda - reorder loops in diagonal assembly for better memory access

jeremylt:
@YohannDudouit, this is a small tweak, but I think it might offer a little bit of help for performance

### DIFF
--- a/backends/cuda/ceed-cuda-operator.c
+++ b/backends/cuda/ceed-cuda-operator.c
@@ -729,51 +729,53 @@ __device__ void diagonalCore(const CeedInt nelem,
   const CeedScalar qfvaluebound = maxnorm*1e-12;
 
   // Compute the diagonal of B^T D B
-  // Each element
-  for (CeedInt e = blockIdx.x*blockDim.z + threadIdx.z; e < nelem;
-       e += gridDim.x*blockDim.z) {
-    CeedInt dout = -1;
-    // Each basis eval mode pair
-    for (CeedInt eout = 0; eout < NUMEMODEOUT; eout++) {
-      const CeedScalar *bt = NULL;
-      if (emodeout[eout] == CEED_EVAL_GRAD)
-        dout += 1;
-      CeedOperatorGetBasisPointer_Cuda(&bt, emodeout[eout], identity, interpout,
-                                      &gradout[dout*NQPTS*NNODES]);
-      CeedInt din = -1;
-      for (CeedInt ein = 0; ein < NUMEMODEIN; ein++) {
-        const CeedScalar *b = NULL;
-        if (emodein[ein] == CEED_EVAL_GRAD)
-          din += 1;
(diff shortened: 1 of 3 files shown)


jeremylt:
I'm moving to performance testing with MFEM.

jeremylt:
This offers a mild performance improvement (cuts off about 1/4 of the time, but its still 5ish x slower than the version in MFEM (which is not at all general)). I think this is ready to merge as an intermediate effort until we can make a cuda/gen version of the diagonal assembly

backends/cuda/ceed-cuda-operator.c
```diff
@@ -729,51 +729,53 @@ __device__ void diagonalCore(const CeedInt nelem,
...
+            const CeedScalar bt_q = bt[q*NNODES+tid], b_q = b[q*NNODES+tid];
+            // Each element
+            for (CeedInt e = blockIdx.x*blockDim.z + threadIdx.z; e < nelem;
+                e += gridDim.x*blockDim.z) {
               const CeedScalar qfvalue =
                 assembledqfarray[((((ein*NCOMP+compOut)*NUMEMODEOUT+eout)*
                                    NCOMP+compOut)*nelem+e)*NQPTS+q];
               if (abs(qfvalue) > qfvaluebound)
-                evalue += bt[q*NNODES+tid] * qfvalue * b[q*NNODES+tid];
+                elemdiagarray[(compOut*nelem+e)*NNODES+tid] += bt_q * qfvalue * b_q;
```
jedbrown, line 777: I'm not sure why this inner loop over elements is good memory locality. And do we need the conditional on `qfvaluebound`? I'd expect it actually prevents the read/store from being executed on all lanes, which would require higher occupancy to hide. But each thread processes all the quadrature points so higher occupancy probably isn't possible. (Do you have profiler/occupancy stats?)

I'd sort of expect that a good implementation would load `assembledqfarray` into shared memory (single element or small batch per thread block) and have one thread per (component, node). Tuning such a thing might require specializing on sizes, which we'd rather not do here.

jeremylt: I made the switch pulling 2 basis entries from memory is more costly than pulling 1 entry from the assembledqfarray.

I could try to drop the conditional. I originally added that check because repeatedly adding nearly zero values added a ton of error.

Ultimately this whole thing needs to be rebuilt in the style of cuda-gen to actually work well, where we build assembledqfarray as we need it, but that's a bigger PR and I only want to use this approach in the cuda-gen backend.

jedbrown: If it's an accuracy issue, I think it would be better to preprocess inputs so values that should be exactly zero actually are.

jeremylt:
Upon further testing, this does not seem to actually consistently make a difference. I had noise in an unlucky direction giving me a false positive.

PR #600, opened 2020-08-04
PETSc examples: Update DMPlexCreateSphereMesh() and DMAddBoundary() calls after changes in PETSc's API

valeriabarra:
I have updated the couple of instances where we were calling `DMPlexCreateSphereMesh` after changes in PETSc.

### DIFF
--- a/examples/petsc/area.c
+++ b/examples/petsc/area.c
@@ -116,7 +116,7 @@ int main(int argc, char **argv) {
     CHKERRQ(ierr);
   } else {
     // Create the mesh as a 0-refined sphere. This will create a cubic surface, not a box
-    ierr = DMPlexCreateSphereMesh(PETSC_COMM_WORLD, topodim, simplex, &dm);
+    ierr = DMPlexCreateSphereMesh(PETSC_COMM_WORLD, topodim, simplex, 1., &dm);
     CHKERRQ(ierr);
     // Set the object name
     ierr = PetscObjectSetName((PetscObject)dm, problemTypes[problemChoice]);
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,7 +150,7 @@ install:
         && export NEK5K_DIR=$PWD/Nek5000 PATH=$PWD/Nek5000/bin:$PATH MPI=0;
     fi
 # PETSc
-  - PETSC_COMMIT=afc36f2e62dc81f9d8cae345c2c65c61f89b0500 # master 2020-05-13
+  - PETSC_COMMIT=caa22a8a6959549f3567c51cb8028520e3252e52 # master 2020-08-04
   - export PETSC_INSTALL=$HOME/install/petsc-$PETSC_COMMIT
   - test -s "$PETSC_INSTALL/lib/pkgconfig/PETSc.pc"
         || (  curl -O https://gitlab.com/petsc/petsc/-/archive/$PETSC_COMMIT/petsc-$PETSC_COMMIT.tar.gz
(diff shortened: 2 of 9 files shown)


valeriabarra:
Need to update `junit`

examples/petsc/area.c
```diff
@@ -116,7 +116,7 @@ int main(int argc, char **argv) {
     CHKERRQ(ierr);
   } else {
     // Create the mesh as a 0-refined sphere. This will create a cubic surface, not a box
-    ierr = DMPlexCreateSphereMesh(PETSC_COMM_WORLD, topodim, simplex, &dm);
+    ierr = DMPlexCreateSphereMesh(PETSC_COMM_WORLD, topodim, simplex, 1., &dm);
```
jeremylt, line 119: Question - do we want to update this with a macro, like we did for DMPlexGetClosureIndices?

jedbrown:
Have we tested it with 3.13?  If it works at all there, then we could use the compatibility macro, otherwise we can remove them and say the example is for >=3.14.

valeriabarra:
> Have we tested it with 3.13? If it works at all there, then we could use the compatibility macro, otherwise we can remove them and say the example is for >=3.14.

I have just switched to 3.13 to test this and tried to simply reconfigure and compile as usual with 
```
./$PETSC_ARCH/lib/petsc/conf/reconfigure-$PETSC_ARCH.py && make all
```
but now I'm fighting a configure error I'm getting with `MUMPS`

jedbrown:
You don't need MUMPS at all for these tests.  I would use a separate PETSC_ARCH so you don't have to reconfigure.

examples/petsc/area.c
valeriabarra: Added them. Thanks

valeriabarra:
Done

jeremylt:
If I recall correctly, Azure is failing because the version checking macros don't work when you're not on a released version (which the PETSc build for Azure isn't).

Other than that, this seems good to me.

valeriabarra:
Thanks Jeremy. Can I go ahead with the merge?

jedbrown:
What's going on here? It's a fresh image.

https://dev.azure.com/CEED-ECP/libCEED/_build/results?buildId=3014&view=logs&j=6b902995-b73d-5f5c-66fd-a7f66c857d2c&t=a1c457be-0fbe-5525-b6b0-ac3cf58f8d6c

jedbrown:
DMAddBoundary isn't in this PR?

valeriabarra:
In [beae6a5](https://github.com/CEED/libCEED/pull/600/commits/beae6a5ce4cd87fc6ef6384d5f79ffdc2f341d81) I updated the call to `DMAddBoundary()` and in [23469ac](https://github.com/CEED/libCEED/pull/600/commits/23469acc68d9f2a98f561c7b5de6ee2a3745201a) I aded the relative compatibility macros. I tested with both 3.13 and master on my local machine and it was fine.

jeremylt:
I think the fluids example also needs the new compatibility macro (and maybe solids too iirc?)

valeriabarra:
> I think the fluids example also needs the new compatibility macro (and maybe solids too iirc?)

Thanks for the reminder Jeremy! Update those, too.

`prove-all` passes clean locally for me.

jedbrown:
It passed on Azure too.

jeremylt:
Travis needs the new commit hash for PETSc to be able to pass.

valeriabarra:
Should I just update [this line](https://github.com/CEED/libCEED/blob/9f2752f144770f23a090e84e243e59bcab3f0c80/.travis.yml#L153) with the latest hash?

jeremylt:
Yes, that should be all that's needed

PR #602, opened 2020-08-06
Rust interface

jeremylt:


jeremylt:
Currently: 
```
cd rust
cargo build
```
builds the libCEED + Rust bindings

rust/src/basis.rs
```diff
@@ -0,0 +1,211 @@
...
+    Hex,
+}
+pub enum EvalMode {
+    None,
+    Interp,
+    Grad,
+    Div,
+    Curl,
+    Weight,
+}
```
jeremylt, line 29: Do we want to just put all enums in `ceed` like was done with the Python interface? Then the user always uses `ceed::EnumType::EnumValue` like `ceed::MemType::Host`

jeremylt:
The big ToDo is QFunctions and Operators.

The question there is what's the Rust-y way to handle our different subclasses of QFunctions/Operators. There are some common functions between our two variants of QFunctions (or Operators), but there some functions unique to each. I need to figure out the best way to express this with Rust.

QFunctions
- Add Fields
- Add Context
- Apply
QFunctionByName
- Apply

Operator
- Add Fields
- Apply
- Assemble Diag
- Multigrid Create

CompositeOperator
- Add Sub Operators
- Apply
- Assemble Diag

rust/src/vector.rs
```diff
@@ -0,0 +1,398 @@
...
+    /// let vec = ceed.vector_from_slice(&[10., 11., 12., 13.]);
+    /// let v = vec.view();
+    /// assert_eq!(v[0..2], [10., 11.]);
+    /// // It is valid to have multiple immutable views
+    /// let w = vec.view();
+    /// assert_eq!(v[1..], w[1..]);
+    /// ```
+    pub fn view(&self) -> VectorView {
+        VectorView::new(self)
+    }
```
jedbrown, line 321: I'd like to get rid of the functions like `get_array`, which leak memory if the caller does not remember to `restore_array`. I don't have strong opinions about whether `VectorView` coerces (via `Deref`) to a slice or ndarray. One possible point in favor of slices is minimizing dependencies and encouraging users to use `zip` patterns for manipulating vectors in-place versus the common allocating semantics in `ndarray`.

rust/src/vector.rs
```diff
@@ -0,0 +1,398 @@
...
+            bind_ceed::CeedVectorGetArrayRead(vec.ptr, crate::MemType::Host as bind_ceed::CeedMemType, &mut ptr);
+        }
+        let array = std::rc::Rc::new(ptr);
+        vec.array_weak.replace(Rc::downgrade(&array));
+        Self {
+            vec: vec,
+            array: array,
+        }
+    }
+}
```
jedbrown, line 373: An alternate way to do this would be to always call `CeedVectorGetArrayRead` when creating a view. I used reference counting on the Rust side (`Vector` holds a `Weak` reference while each `VectorView` holds a strong (`Rc`)) because it seems to let the compiler be a little stronger about enforcing the mutable reference rules. I'm not attached to it, however.

jeremylt:
Unfortunately, our `const` CeedObjects like `CEED_VECTOR_ACTIVE`, etc are not `const` in the bindings but instead `static` which means we can't make a const rust `vector_active`, ect out of them. I'm not sure if this is a bindgen bug or not, but if it's not then we will have workaround in a slightly annoying way.

In Python we were able to create the analogos const objects and just call through to C
```
    def set_field(self, fieldname, restriction, basis, vector):
        """Provide a field to an Operator for use by its QFunction.
           Args:
             fieldname: name of the field (to be matched with the same name used
                          by QFunction)
             restriction: ElemRestriction
             basis: Basis in which the field resides or CEED_BASIS_COLLOCATED
                      if collocated with quadrature points
             vector: Vector to be used by Operator or CEED_VECTOR_ACTIVE
                       if field is active or CEED_VECTOR_NONE if using
                       CEED_EVAL_WEIGHT in the QFunction"""

        # libCEED call
        fieldnameAscii = ffi.new("char[]", fieldname.encode('ascii'))
        err_code = lib.CeedOperatorSetField(self._pointer[0], fieldnameAscii,
                                            restriction._pointer[0], basis._pointer[0],
                                            vector._pointer[0])
        self._ceed._check_error(err_code)
```

jedbrown:
How did you try to implement this? Why can't we have a static object on the Rust side?

jeremylt:
If I try to make these constants as static I get: 
```
`*mut prelude::bind_ceed::Ceed_private` cannot be shared between threads safely
```

jeremylt:
Alright, I think that we have the CEED_VECTOR_NONE and such now

jeremylt:
We are now in a basic functional state :tada:

I think think the biggest question now is the array interface

jeremylt:
Updated ToDo (Feel free to edit):

- [ ] array ownership and viewing
- [ ] lots of casting between usize and i32 right now for arguments
- [ ] multigrid functions should return [op_coarse, op_prolong, op_restrict] (don't think rust likes passing in uninitalized args)
- [ ] expand documentation tests (lots of op stubs)
- [ ] ...

rust/src/vector.rs
jeremylt: As thoroughly discussed on Slack, I dropped  `get_array` and `restore_array` and added a mutable analogue of this interface.

jeremylt:
The current ToDos:

Impl
- [ ] Implement User QFunctions, I think we need a trampoline
- [ ] Implement QFunctionContext, need to pass around arbitrary structs

Testing
- [ ] Add examples of point block assembly for testing
- [ ] Add composite operator examples for testing

Refactor
- [ ] Refactor multigrid assembly stuff to return new Rust operator objects

jedbrown:
I don't think contexts are needed for QFunctions written in Rust since Rust has closures and [Fn](https://doc.rust-lang.org/std/ops/trait.Fn.html) to carry that state. We can deal with GPUs later.

jeremylt:
That will make things easier

jeremylt:
User QFunction are now needed to proceed with the other ToDos.

I'm looking into both a way to use native Rust functions and a way to continue to use our single source files.

jedbrown:
I think a Rust qfunction should implement a trait. There could be a trait for scalar functions and vectorized functions, with a default implementation lifting scalar to vector. There could be a `From` implementation from appropriately-typed functions so users could write their qfunction in a closure. We might be able to use a macro to provide an interface similar to what Will has done in #632.

rust/src/qfunction.rs
```diff
@@ -140,22 +191,85 @@ impl QFunction {
...
+        let input_sizes = [0; crate::MAX_QFUNCTION_FIELDS];
+        let output_sizes = [0; crate::MAX_QFUNCTION_FIELDS];
+        let trampoline_data = QFunctionTrampolineData {
+            number_inputs,
+            number_outputs,
+            input_sizes,
+            output_sizes,
+        };
+
+        let qf_closure = |q: bind_ceed::CeedInt,
```
jeremylt, line 211: @jedbrown @YohannDudouit 

This isn't quite right but should be close to what we need for (finally) finish this. Thoughts?

jeremylt:
We now have support for user closures.

ToDo:
- Composite operator examples
- Documentation auto deployment

I should be done tomorrow

jeremylt:
Documentation live link: http://ceed.exascaleproject.org/libCEED-rust-docs/ceed/index.html

jeremylt:
Updated for the Julia PR merge. I think this should be ready to merge and then we're ready to submit to JOSS.

.travis.yml
```diff
@@ -109,6 +109,49 @@ jobs:
...
+       - cd ..
+     deploy:
+       provider: pages
+       skip_cleanup: true
+       token: $RUST_DOCS_TOKEN
+       local_dir: rust/target/doc
+       strategy: git
+       repo: CEED/libCEED-rust-docs
+       on:
+         branch: jeremy/rust
```
jeremylt, line 154: We'll similarly need to shift this target branch when merging

jeremylt:
I'll update for #658 once merged

jeremylt:
ToDo: set up doc deployment for GitHub Actions.

rust/src/lib.rs
```diff
@@ -588,7 +588,7 @@ mod tests {
         op_build.apply(&x, &mut qdata);
 
         // Mass operator
-        let qf_mass = ceed.q_function_interior_by_name("MassApply".to_string());
+        let qf_mass = ceed.q_function_interior_by_name("MassApply");
```
jeremylt, line 591: This is much cleaner for the users now

jeremylt:
I think this is pretty close. Sure there are things that could be improved, but I think this is in a good functional yet experimental state.

I just need to fix the doc deployment and I think its ok to merge (and then JOSS).

.github/workflows/rust-documentation.yml
```diff
@@ -0,0 +1,37 @@
+name: Rust Documentation
+
+on:
+  push:
+    branches:
+      - main
+      - jeremy/rust
```
jeremylt, line 7: Note: drop extra branch when merged

jeremylt:
Doc deployment all set

rust/src/qfunction.rs
```diff
@@ -0,0 +1,532 @@
...
+            u_c[i] = u[i].ptr;
+        }
+        let mut v_c = [std::ptr::null_mut(); crate::MAX_QFUNCTION_FIELDS];
+        for i in 0..std::cmp::min(crate::MAX_QFUNCTION_FIELDS, v.len()) {
+            v_c[i] = v[i].ptr;
+        }
+        unsafe {
+            bind_ceed::CeedQFunctionApply(self.ptr, Q, u_c.as_mut_ptr(), v_c.as_mut_ptr());
+        }
+    }
```
YohannDudouit, line 190: I'm really not a big fan of having to copy all the pointers for each apply, I feel like we should be able to avoid it, but I don't really see how.

rust/src/qfunction.rs
```diff
@@ -0,0 +1,532 @@
...
+}
+
+// -----------------------------------------------------------------------------
+// User QFunction Closure
+// -----------------------------------------------------------------------------
+pub type QFunctionUserClosure = dyn FnMut(usize, &[&[f64]], &mut [&mut [f64]]) -> i32;
```
YohannDudouit, line 196: I'm wondering if we could instead directly have arrays of slices of size `MAX_QFUNCTION_FIELDS` instead of slices of slices, could that prevent having to copy all the pointers in `apply`?

rust/src/qfunction.rs
```diff
@@ -0,0 +1,532 @@
...
+    inputs: *const *const bind_ceed::CeedScalar,
+    outputs: *const *mut bind_ceed::CeedScalar,
+) -> ::std::os::raw::c_int {
+    let context = &mut *(ctx as *mut QFunction);
+    let trampoline_data = &context.trampoline_data;
+
+    // Inputs
+    let inputs_slice: &[*const bind_ceed::CeedScalar] =
+        std::slice::from_raw_parts(inputs, crate::MAX_QFUNCTION_FIELDS);
+    let mut inputs_vec: Vec<&[f64]> = Vec::new();
```
YohannDudouit, line 210: Could we have arrays of slices here to prevent transforming `Vec` into arrays?

rust/src/qfunction.rs
```diff
@@ -0,0 +1,532 @@
...
+        inputs_vec.push(&std::slice::from_raw_parts(
+            inputs_slice[i],
+            (trampoline_data.input_sizes[i] * q) as usize,
+        ) as &[f64]);
+    }
+
+    // Outputs
+    let outputs_slice: &[*mut bind_ceed::CeedScalar] =
+        std::slice::from_raw_parts(outputs, crate::MAX_QFUNCTION_FIELDS);
+    let mut outputs_vec: Vec<&mut [f64]> = Vec::new();
```
YohannDudouit, line 221: Use `[&mut [f64], MAX_QFUNCTION_FIELDS]`?

YohannDudouit (commented):
Overall I think this is almost ready to be merged, I just have small suggestions/questions about the way we handle qFunctions. The qFunction code feels more complex than it should be. Otherwise very good work @jeremylt .

rust/src/qfunction.rs
jedbrown: [arrayvec](https://docs.rs/arrayvec/0.5.2/arrayvec/struct.ArrayVec.html) gives the `Vec` interface for stack-allocated arrays, though a raw array also seems fine here.

Somewhat related to this review, I don't know how to provide interior mutability without a much more complicated type signature (e.g., using `RefCell`).
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b33c4708959797511baf72a8be6fc965

rust/src/qfunction.rs
jeremylt: I picked vec with a runtime determined length so the user couldn't reach into the inputs/outputs array into an uninitalized entry, but I don't feel strongly about it.

rust/src/qfunction.rs
YohannDudouit: Is it what you want?
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=112b28c2b8892bd520bc7725553eef44

jeremylt: That will not allow assignment to a mutable variable, which seems desirable
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=112b28c2b8892bd520bc7725553eef44

rust/src/qfunction.rs
jeremylt: We would still need to make slices from the raw pointers, since I'd like the users not to have to do `unsafe` things and using raw pointers is `unsafe`

rust/src/qfunction.rs
jeremylt: I agree that the copies are ugly, but realistically, the user should never call this function (should use OperatorApply), so I don't know that the overhead here is a big deal.

rust/src/qfunction.rs
YohannDudouit: I thought this was what @jedbrown wanted

jeremylt: Well, I think its good to allow
```
mut v = &mut outputs[0];
```
and disallow
```
outputs[0] = v
```
if we can, but I feel more strongly about allowing the first.

jedbrown: I wanted the interior (array entry) mutability without ownership (i.e., slice, not array). But I suppose that as a function argument, it'd be okay to move the array, as in
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=38ba82d5e1bf110ea4fe5075eabf1fe2

Trouble is that the length of the array is now part of the function signature (maybe this is also a good thing). I'm not sure the best way to express it, but it'd be better if we could just make them all function arguments. It seems like macro should be able to determine the number of arguments and mutability of those arguments in order to do the transformation, but I think we could work out that refinement in a subsequent PR.

jedbrown: And the move semantics produce good code (no copies, just a bounds check).
```
example::f:
        push    rax
        cmp     rsi, 2
        jb      .LBB6_2
        add     dword ptr [rdi + 4], 50
        pop     rax
        ret

...
        mov     rdi, qword ptr [rsp + 80]
        mov     rsi, qword ptr [rsp + 88]
        call    example::f
```
for
```rust
#[inline(never)]
fn f(a: [&mut [i32]; 2]) {
    a[0][1] += 50;
}

pub fn main() {
    let mut r0 = [1, 2];
    let mut r1 = [3, 4];
    let z: [&mut [i32]; 2] = [&mut r0, &mut r1];
    z[0][1] = 5;
    println!("z {:?}", z);
    f(z);         // z moved here, but r0 and r1 are still valid
    println!("z {:?}", [r0, r1])
}
```
https://gcc.godbolt.org/z/GoKzGx

jeremylt: I think when I tried this before I had to set the array length to the max possible, which has us back at 
```
[&mut [f64], MAX_QFUNCTION_FIELDS]
```

jedbrown: A good use case for const generics?
https://without.boats/blog/shipping-const-generics/

rust/src/qfunction.rs
YohannDudouit: I think what @jedbrown proposed with `ArrayVec` solves this issue.

rust/src/qfunction.rs
YohannDudouit: I also think that we should be able to do with Rust macros (`qfunction!`) something with the same flavor as what @pazner has done for the Julia interface. DSL are also extremely easy to write with Rust which should help us write very elegantly `CeedOperator`. I agree that none of this needs to be in the first release of the Rust interface.

YohannDudouit (approved):

jeremylt:
I'm using `q` here
```
let mut mass_2_comp = |
  q: usize,
  [u, qdata, ..]: [&[f64]; MAX_QFUNCTION_FIELDS],
  [v, ..]: [&mut [f64]; MAX_QFUNCTION_FIELDS],
 | -> i32
 {
   // Iterate over quadrature points
   for i in 0..q {
    v[i + 0*q] = u[i + 1*q] * qdata[i];
    v[i + 1*q] = u[i + 0*q] * qdata[i];
  }

  // Return clean error code
  0
};
```

jeremylt:
Alrighty, we're down to
```
let mut user_f = |
  [u, weights, ..]: [&[f64]; MAX_QFUNCTION_FIELDS], // inputs
  [v, ..]: [&mut [f64]; MAX_QFUNCTION_FIELDS], // outputs
| -> i32
{
  // Iterate over quadrature points
  v
    .iter_mut()
    .zip(u.iter().zip(weights.iter()))
    .for_each(|(v, (u, w))| *v = u * w);

  // Return clean error code
  0
};
```

I think this should work for now and give us something good to work with for a future macro.

jedbrown:
I think my push is a modest improvement.

I think it would be more idiomatic to return a `Result<(), E>` so the `?` operator can be used inside qfunctions.

jedbrown:
I wouldn't let it block merging this branch, but perhaps we can use an iteration idiom like this.
https://github.com/lumol-org/soa-derive

jedbrown (approved):
I'm fine with merging this now and we can refine in future PRs. One thing that I think will be useful is to have a way to build an internal libceed.so (or libceed.a). That seems common among other packages, to avoid installation complexities that the user needs to build and set `LD_LIBRARY_PATH` separately.
https://github.com/gyscos/zstd-rs
https://github.com/blas-lapack-rs/blas-lapack-rs.github.io/wiki

rust/src/lib.rs
```diff
@@ -18,6 +18,12 @@
 //!
 //! This is the documentation for the high level libCEED Rust interface.
 //! See the full libCEED user manual [here](https://libceed.readthedocs.io).
+//! 
+//! ## Installing
+//! 
+//! First, libCEED needs to be built and `LD_LIBRARY_PATH` updated to contain
+//! the filepath to `libceed.so`. Then `cargo` commands may be used as usual in
+//! the `libCEED/rust` folder, such as `cargo build` and `cargo test`.
```
jeremylt, line 26: @jedbrown, install instructions added.

rust/src/qfunction.rs
```diff
@@ -13,6 +13,10 @@
 // the planning and preparation of a capable exascale ecosystem, including
 // software, applications, hardware, advanced system engineering and early
 // testbed platforms, in support of the nation's exascale computing imperative
+
+//! A Ceed QFunction represents the spatial terms of the point-wise functions
+//! describing the physics at the quadrature points.
```
jeremylt, line 18: I also added brief descriptions of each module to make the docs easier to read.

Issue #603, opened 2020-08-08
PETSc examples refinement


valeriabarra:
I don't know what's changed on the PETSc side lately, but currently on `main` our problems that use uniform refinement, e.g., `examples/petsc/area` and `examples/petsc/bpssphere` are broken. Our CI didn't catch this because it doesn't test the refinement. It should have. I'm going to add it. 

I'm investigating this issue, but I don't know exactly what has changed to cause it. It was fine until a few days ago. This also affects PR #550 .

To reproduce:
```
./bpssphere -dm_refine 1
```

Stack trace:
```
[0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
[0]PETSC ERROR: Nonconforming object sizes
[0]PETSC ERROR: The section point (0) closure size 12 != dual space dimension 9
[0]PETSC ERROR: See https://www.mcs.anl.gov/petsc/documentation/faq.html for trouble shooting.
[0]PETSC ERROR: Petsc Development GIT revision: v3.13.4-637-gcaa22a8a69  GIT Date: 2020-08-04 04:43:23 +0000
[0]PETSC ERROR: ./bpssphere on a arch-linux2-debug named work by valeria Fri Aug  7 18:24:16 2020
[0]PETSC ERROR: Configure options --download-cmake --download-exodusii --download-fblaslapack=1 --download-hdf5 --download-hypre --download-metis --download-ml --download-mpich --download-mumps --download-netcdf --download-parmetis --download-pnetcdf --download-scalapack --download-superlu --download-superlu_dist --download-zlib --with-cxx-dialect=C++11 --with-debugging=1 --with-x=1 --with-zlib PETSC_ARCH=arch-linux2-debug
[0]PETSC ERROR: #1 DMProjectLocal_Generic_Plex() line 700 in /home/valeria/Research/CEED/petsc/src/dm/impls/plex/plexproject.c
[0]PETSC ERROR: #2 DMProjectFieldLocal_Plex() line 867 in /home/valeria/Research/CEED/petsc/src/dm/impls/plex/plexproject.c
[0]PETSC ERROR: #3 DMProjectFieldLocal() line 8291 in /home/valeria/Research/CEED/petsc/src/dm/interface/dm.c
[0]PETSC ERROR: #4 DMPlexRemapGeometry() line 3124 in /home/valeria/Research/CEED/petsc/src/dm/impls/plex/plexgeometry.c
[...]

PR #604, opened 2020-08-08
PETSc area and bpssphere examples: Improve testing

valeriabarra:
This PR adds the command line argument `-dm_refine` to our `petsc/examples/bpssphere` and `petsc/examples/area` problem `TESTARGS`. This could have caught the bug in PETSc that was introduced lately. And while at it, I realized that the area problem test was not checking for any tolerance, so I added that, and I tightened the tolerance of the other test, too.

This PR is downstream of a PETSc's [MR](https://gitlab.com/petsc/petsc/-/merge_requests/3034?commit_id=68317524b39380e49bbfefbbf59f91b5b48300b3) for which I reported a bug. The PETSc MR and this PR will close issue #603

As soon as that one merges, this one can follow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,7 +150,7 @@ install:
         && export NEK5K_DIR=$PWD/Nek5000 PATH=$PWD/Nek5000/bin:$PATH MPI=0;
     fi
 # PETSc
-  - PETSC_COMMIT=caa22a8a6959549f3567c51cb8028520e3252e52 # master 2020-08-04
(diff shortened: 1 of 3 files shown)


valeriabarra:
I will update Travis to point to the right commit, once it is merged in PETSc

valeriabarra:
I'm not sure how to fix the Azure pipeline failing here. Do we have to have @jedbrown 's container updated?

PR #610, opened 2020-08-21
QF - add interface to set device user function

jeremylt:
@YohannDudouit @pazner @jedbrown 

This is half of the interface discussed on Slack. The other half would be the actual backend impl.

ToDo:
- [x] Add backend impl for CUDA backends (and HIP if we can)
- [ ] Decide how to test (HIP CI easier than CUDA right now)

### DIFF
--- a/backends/cuda/ceed-cuda-qfunction-load.cpp
+++ b/backends/cuda/ceed-cuda-qfunction-load.cpp
@@ -48,11 +48,15 @@ extern "C" int CeedCudaBuildQFunction(CeedQFunction qf) {
   CeedInt ierr;
   using std::ostringstream;
   using std::string;
+  Ceed ceed;
+  CeedQFunctionGetCeed(qf, &ceed);
   CeedQFunction_Cuda *data;
   ierr = CeedQFunctionGetData(qf, (void **)&data); CeedChk(ierr);
   // QFunction is built
   if (data->qFunction)
     return 0;
+  if (!data->qFunctionSource)
+    return CeedError(ceed, 1, "No QFunction source or CUfunction provided.");
 
   // QFunction kernel generation
   CeedInt numinputfields, numoutputfields, size;
@@ -125,8 +129,6 @@ extern "C" int CeedCudaBuildQFunction(CeedQFunction qf) {
   code << "}\n";
 
   // View kernel for debugging
-  Ceed ceed;
-  CeedQFunctionGetCeed(qf, &ceed);
   CeedDebug(code.str().c_str());
 
   // Compile kernel
--- a/backends/cuda/ceed-cuda-qfunction-load.cpp
+++ b/backends/cuda/ceed-cuda-qfunction-load.cpp
@@ -51,9 +51,9 @@ extern "C" int CeedCudaBuildQFunction(CeedQFunction qf) {
   CeedQFunction_Cuda *data;
   ierr = CeedQFunctionGetData(qf, (void **)&data); CeedChk(ierr);
   // QFunction is built
-  if (!data->qFunctionSource)
+  if (data->qFunction)
     return 0;
-  
+
   // QFunction kernel generation
   CeedInt numinputfields, numoutputfields, size;
   ierr = CeedQFunctionGetNumArgs(qf, &numinputfields, &numoutputfields);
@@ -77,7 +77,7 @@ extern "C" int CeedCudaBuildQFunction(CeedQFunction qf) {
   code << qReadWriteS;
   code << qFunction;
   code << "extern \"C\" __global__ void " << kernelName << "(void *ctx, CeedInt Q, Fields_Cuda fields) {\n";
-  
+
   // Inputs
   for (CeedInt i = 0; i < numinputfields; i++) {
     code << "// Input field "<<i<<"\n";
--- a/backends/cuda/ceed-cuda-qfunction.c
+++ b/backends/cuda/ceed-cuda-qfunction.c
@@ -97,6 +97,18 @@ static int CeedQFunctionDestroy_Cuda(CeedQFunction qf) {
 }
 
 //------------------------------------------------------------------------------
+// Set User QFunction
+//------------------------------------------------------------------------------
+static int CeedQFunctionSetCUDAUserFunction_Cuda(CeedQFunction qf,
+                                                 CUfunction f) {
+  int ierr;
+  CeedQFunction_Cuda *data;
+  ierr = CeedQFunctionGetData(qf, &data); CeedChk(ierr);
+  data->qFunction = f;
+  return 0;
+}
+
+//------------------------------------------------------------------------------
 // Load QFunction source file
 //------------------------------------------------------------------------------
 static int CeedCudaLoadQFunction(CeedQFunction qf, char *c_src_file) {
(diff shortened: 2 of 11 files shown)


include/ceed-cuda.h
```diff
@@ -0,0 +1,25 @@
...
+/// software, applications, hardware, advanced system engineering and early
+/// testbed platforms, in support of the nation's exascale computing imperative.
+
+#ifndef _ceed_cuda_h
+#define _ceed_cuda_h
+
+#include <cuda.h>
+
+CEED_EXTERN int CeedQFunctionSetCUDAUserFunction(CeedQFunction qf,
+    CUfunction *f);
```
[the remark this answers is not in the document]
jeremylt: You're right on both counts. I'll get it in the morning with other edits.

jeremylt:
Updated based upon feedback. What do we think now?

jeremylt:
@pazner if you're happy with your modification to the CUDA backends to accept CUFunctions, we can add that to this PR (do you are the right permissions or do we need to ask Jed?)

interface/ceed-qfunction.c
```diff
@@ -622,6 +628,56 @@ int CeedQFunctionAddOutput(CeedQFunction qf, const char *fieldname,
...
+
+  @ref User
+**/
+int CeedQFunctionSetCUDAUserFunction(CeedQFunction qf, CUfunction f) {
+  int ierr;
+  if (!qf->SetCUDAUserFunction) {
+    Ceed ceed;
+    ierr = CeedQFunctionGetCeed(qf, &ceed); CeedChk(ierr);
+    CeedDebug(ceed,
+              "Backend does not support CUfunction pointers for QFunctions.");
```
pazner, line 648:
```suggestion
    CeedDebug("Backend does not support CUfunction pointers for QFunctions.");
```
Is this right?

interface/ceed-qfunction.c
```diff
@@ -622,6 +628,56 @@ int CeedQFunctionAddOutput(CeedQFunction qf, const char *fieldname,
...
+
+  @ref User
+**/
+int CeedQFunctionSetHIPUserFunction(CeedQFunction qf, hipFunction_t f) {
+  int ierr;
+  if (!qf->SetHIPUserFunction) {
+    Ceed ceed;
+    ierr = CeedQFunctionGetCeed(qf, &ceed); CeedChk(ierr);
+    CeedDebug(ceed,
+              "Backend does not support hipFunction_t pointers for QFunctions.");
```
pazner, line 673: Same as above?

backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -164,26 +177,34 @@ int CeedQFunctionCreate_Cuda(CeedQFunction qf) {
   CeedQFunction_Cuda *data;
   ierr = CeedCalloc(1, &data); CeedChk(ierr);
   ierr = CeedQFunctionSetData(qf, data); CeedChk(ierr);
-  CeedInt numinputfields, numoutputfields;
-  ierr = CeedQFunctionGetNumArgs(qf, &numinputfields, &numoutputfields);
-  CeedChk(ierr);
```
pazner, line 169: Did these lines serve a purpose?

interface/ceed-qfunction.c
jeremylt: yes, I believe that you are correct

backends/cuda/ceed-cuda-qfunction.c
jeremylt: I think those predate single source Qfunctions and were missed in the transition

pazner:
> @pazner if you're happy with your modification to the CUDA backends to accept CUFunctions, we can add that to this PR (do you are the right permissions or do we need to ask Jed?)

Sure, I just pushed my changes. It uses an empty string for the source path to indicate that the user should supply the Q-function themselves. There is no real error checking if they give an empty string and don't supply the CUfunction (it will just crash).

backends/cuda/ceed-cuda-qfunction-load.cpp
```diff
@@ -51,9 +51,9 @@ extern "C" int CeedCudaBuildQFunction(CeedQFunction qf) {
   CeedQFunction_Cuda *data;
   ierr = CeedQFunctionGetData(qf, (void **)&data); CeedChk(ierr);
   // QFunction is built
-  if (!data->qFunctionSource)
+  if (data->qFunction)
     return 0;
-  
+
```
jeremylt, line 56:
```suggestion
  if (!data->QFunctionSource)
    return CeedError(ceed, "No QFunction source or CUfunction provided.");
```

would require moving up lines 128,129

jeremylt:
Thanks. I'll fix the Debug in the morning and then I think we're set.

jeremylt:
@jedbrown @pazner how should we test this so we can merge for 0.7? We don't have CUDA CI yet, but I think we should have multiple people test locally to verify before merging

pazner:
> @jedbrown @pazner how should we test this so we can merge for 0.7? We don't have CUDA CI yet, but I think we should have multiple people test locally to verify before merging

I can test locally. What exactly are the tests that should be run (is `make prove-all` enough)?

jeremylt:
Well, how can we test that this interface is correctly handing a CUfunction off to the CUDA backend and that backend can use the CUfunction while applying the action of a QFunction?

pazner:
It seems a bit tricky since as far as I know there is nothing in libCEED currently that generates a CUfunction that we can pass in...

jedbrown:
Can we have a CUDA test (only run it when CUDA is available) that generates one and uses it?

interface/ceed-qfunction.c
```diff
@@ -622,6 +628,54 @@ int CeedQFunctionAddOutput(CeedQFunction qf, const char *fieldname,
...
+  if (!qf->SetHIPUserFunction) {
+    Ceed ceed;
+    ierr = CeedQFunctionGetCeed(qf, &ceed); CeedChk(ierr);
+    CeedDebug("Backend does not support hipFunction_t pointers for QFunctions.");
+  } else {
+    ierr = qf->SetHIPUserFunction(qf, f); CeedChk(ierr);
+  }
+  return 0;
+}
+#endif
```
jedbrown, line 677: Rather than having globally-defined macros `CEED_CUDA_ENABLED`, could we just move these functions to separate files that are only built whin CUDA (HIP, respectively) are available? That saves anti-patterns of checking those macros and avoids needing to force rebuild if adding/removing them from the configuration.

Make doesn't know when variables change, so if you
```console
make      # builds without CUDA
make CUDA_DIR=/path/to/cuda # builds CUDA sources
```
with the macros, this file won't be recompiled, leading to an inconsistent build. There are ways to ensure that make rebuilds in such conditions, but they are a blunt instrument and simpler is better when possible.

jeremylt: I added a quick fix but didn't test. Headed to appointment, but can check later if CI issue.

jeremylt: note to self, need to drop the defs of CEED_*_ENABLED

jeremylt: Ok, rebased off main and fixed it up. It builds fine locally for CUDA and HIP

jedbrown (approved):

jedbrown:
"Decide how to test" is still "ToDo" at the top of this page so I'll let you decide if it's done.

I don't understand the GitHub policy change that one apparently needs administrator privileges to merge when a branch is "out-of-date" in a conflict-free way. It feels like this was a behavior change not too long ago, but perhaps it's a changed setting. Anyway, I can do an admin merge when ready.

jeremylt:
You can merge now. I plan on doing a testing cleanup after 0.7

PR #616, opened 2020-08-25
CUDA - clean up logic for collograd in cuda/gen

jeremylt:
@YohannDudouit as we discussed

This PR fixes the logic for using the collocated gradient for cuda-gen based upon if all bases  performing a grad support this operation instead of only if the last output basis supports a collocated gradient.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -785,15 +785,17 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
   code << "\n#define CeedPragmaSIMD\n";
 
   // Find dim and Q1d
-  bool collograd = false;
+  bool collograd = true;
   for (CeedInt i = 0; i < numinputfields; i++) {
     ierr = CeedOperatorFieldGetBasis(opinputfields[i], &basis); CeedChk(ierr);
     if (basis != CEED_BASIS_COLLOCATED) {
       ierr = CeedBasisGetData(basis, &basis_data); CeedChk(ierr);
+      ierr = CeedQFunctionFieldGetEvalMode(qfinputfields[i], &emode);
+      CeedChk(ierr);
 
       // Check for collocated gradient
-      if (basis_data->d_collograd1d)
-        collograd = true; 
+      if (emode == CEED_EVAL_GRAD)
+        collograd = collograd && !!basis_data->d_collograd1d; 
 
       // Collect dim and Q1d
       ierr = CeedBasisGetDimension(basis, &dim); CeedChk(ierr);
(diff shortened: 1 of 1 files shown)


backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -785,15 +785,17 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
...
     if (basis != CEED_BASIS_COLLOCATED) {
       ierr = CeedBasisGetData(basis, &basis_data); CeedChk(ierr);
+      ierr = CeedQFunctionFieldGetEvalMode(qfinputfields[i], &emode);
+      CeedChk(ierr);
 
       // Check for collocated gradient
-      if (basis_data->d_collograd1d)
-        collograd = true; 
+      if (emode == CEED_EVAL_GRAD)
+        collograd = collograd && !!basis_data->d_collograd1d;
```
YohannDudouit, line 798: Why do you need the double negation here?

jeremylt: I put it in there to explicitly cast as a bool. Its a style thing, so we can drop it if preferred.

YohannDudouit: I find it a bit confusing.

YohannDudouit: Do you really to cast if you don't do this?

jeremylt: It doesn't change the output at all, so we can drop it

YohannDudouit: Isn't it an `||` instead of `&&`?

YohannDudouit: like this you can use `bool collograd = false`.

jeremylt: Previously it was, but I believe that was not fully correct. We can only use the collograd approach if it's supported by all bases applying a grad, I believe

YohannDudouit: `collograd` will be `true` for a mass operator the way it is defined.

jeremylt: Same with the current code too. The big question is when do we actually want to use this mode.

Note, while we're here, I should probably generalize this for future nontensor elements. This approach won't work for nontensor and tensor elements both

YohannDudouit: (I'm not sure it's really a bad thing)

YohannDudouit: What `collograd` really means is, do we treat call the qFunction on 1 quadrature point, or on Q1d quadrature points.

PR #618, opened 2020-08-30
Some refactoring ideas.

YohannDudouit:
- Remove lifetimes for `Basis`, `ElemRestriction`, `QFunction`, `Operator`,
- Define Option types for `Basis`, `ElemRestriction`, `QFunction`, `Operator` instead of using static values,
- Refactor Display functions
co-author: @AdrienChampion


rust/src/operator.rs
```diff
@@ -14,29 +14,26 @@
...
 use crate::prelude::*;
-use std::ffi::CString;
-use std::fmt;
 
 // -----------------------------------------------------------------------------
 // CeedOperator context wrapper
 // -----------------------------------------------------------------------------
-pub struct OperatorCore<'a> {
-    ceed: &'a crate::Ceed,
+pub struct OperatorCore {
```
jeremylt, line 21: should be pub(crate) [my mistake]

rust/src/basis.rs
```diff
@@ -94,11 +113,11 @@ impl<'a> Basis<'a> {
                 &mut ptr,
             )
         };
-        Self { ceed, ptr }
+        Self { ptr }
```
jeremylt, line 116: Why drop the ref to the ceed?

YohannDudouit: I'm trying to reintroduce it with an `Rc`, I think it makes more sense that the objects created by the backend tells the backend to not self destroy rather than the backend telling the objects to not self destroy. But this is a questionnable choice. This makes the borrow checking become runtime instead of compile time.

jeremylt:
I think this is good and we should pull it over to the main rust branch

rust/src/basis.rs
```diff
@@ -14,21 +14,41 @@
...
 use crate::prelude::*;
-use std::ffi::CString;
-use std::fmt;
+
+// -----------------------------------------------------------------------------
+// CeedBasis option
+// -----------------------------------------------------------------------------
+#[derive(Clone, Copy)]
+pub enum BasisOpt<'a> {
+    Some(&'a Basis),
```
jedbrown, line 23: Maybe the enum should be called `Basis` and use `BasisRaw` or some such for the (not visible to most users) type we now call `Basis`?

jeremylt: It looks like the users still make bases as before to me? This looks like its an emun that returns the pointer from the user's basis object or `CEED_BASIS_COLLOCATED`

YohannDudouit: I think `BasisOpt` is more standard Rust for this kind of things, and I agree with @jeremylt that the user will still manipulate `Basis` most of time, the only time he manipulates `BasisOpt` is to access the special values.

rust/src/operator.rs
YohannDudouit: done.

rust/src/basis.rs
jedbrown: Sounds good if that's more idiomatic. I was thinking we could simplify the arg types for methods like `Operator::set_field`, but I'm very much still learning the idioms, and the `impl Into<BasisOpt>` technique is fine with me.

jeremylt:
@YohannDudouit any objection to merging now to the main Rust branch?

YohannDudouit:
No objection, sorry for not testing and documenting everything.

Issue #621, opened 2020-09-10
Tests failing for cuda-gen backend.


YohannDudouit:
t502, t506, t507, t537, t550, t552 are currently failing for the `cuda-gen` backend on the `main` branch.

YohannDudouit:
```
0f54b25e60fd522f688efb036ba1cac9e3fcde63 is the first bad commit
commit 0f54b25e60fd522f688efb036ba1cac9e3fcde63
Author: jeremylt <thompson.jeremy.luke@gmail.com>
Date:   Tue Aug 25 10:28:49 2020 -0600

    CUDA - clean up logic for collograd in cuda/gen, should be based on all bases with interp/grad
```
My bad I approved this.

jeremylt:
Huh, I had tested it locally on my machine

jeremylt:
Ah, its dropping the double negation
75c7b2080cf69f285fbdb78ddf9afe348969eb95
reverting that alone will fix the problem.

jeremylt:
nope, not quite

jeremylt:
If I go into your branch, use f7eb965ac7e108a7c57610ce6b3f3fccec58b347, and comment out lines 815 and 856 of `ceed-cuda-gen-operator-build.cpp`, then everything works.

YohannDudouit:
Lines 815 and 856?
`// Collect dim and Q1d`
`code << "\n#define CEED_Q_VLA "<<Q1d<<"\n\n";`

jeremylt:
Lines 815 and 856 on that commit:
```
      if (emode == CEED_EVAL_GRAD)
```

YohannDudouit:
Maybe that we want to fix this before #620 gets merged, it's already affecting people on the LLNL side.

PR #623, opened 2020-09-10
Consolidate CUDA backends

nbeams:
This PR removes the cuda-reg backend, moving its deterministic element restrictions to cuda-ref and its basis initialization kernels to cuda-shared (which also used them).  All backends that previously had a delegation to cuda-reg for element restrictions have been switched to cuda-ref.  The hip-ref backend restrictions have been updated to match cuda-ref.

I'm wondering how best to add this to the release notes.  "Deprecated features" doesn't seem right, as we don't really lose any important features with this change.  But it's not really an "interface change" either...


jeremylt:
I should be able to test locally tomorrow morning.

Perhaps for the release notes:
Removed `/gpu/cuda/reg` backend and transferred core features into `/gpu/cuda/ref` and `/gpu/cuda/shared`

valeriabarra:
> I'm wondering how best to add this to the release notes. "Deprecated features" doesn't seem right, as we don't really lose any important features with this change. But it's not really an "interface change" either...

"Deprecated backends"?

jeremylt:
Ah, I see now. I agree that its totally fair to add a section

nbeams:
> I should be able to test locally tomorrow morning.

I do see failures for cuda-gen, but they are the same ones as in #621, so I'm pretty sure they are unrelated to this PR.

nbeams:
@valeriabarra  and @jeremylt, I tried combining your suggestions in 48a65224c9c55c0983064b1cb736437fa872f5aa.  Will change further as requested, though!

jeremylt (approved):
LGTM. The only backed I wasn't able to test was MAGMA. Trying to reinstall MAGMA from the master branch is always an adventure, and I'm not currently successfully completing that adventure.

nbeams:
> Trying to reinstall MAGMA from the master branch is always an adventure, and I'm not currently successfully completing that adventure.

Luckily I always test MAGMA before opening a PR (as you would expect, considering) :)

nbeams:
The previous `make style` failure was my fault -- turns out I had the wrong version of `astyle` on the machine where I was doing HIP developtment -- but I'm not sure what's going on with the Travis failures now.  It says `make: icc: Command not found`.

jeremylt:
It appears that Intel might have broke the oneapi installer again. I'm looking into it.

jeremylt:
The Travis failure is due to Intel incorrectly rolling out oneapi 2021.1-beta09

In the Apt package, icc and ipp are 2021.1-beta08 while ifort is 2021.1-beta09, which means only the 'most up to date' package (ifort) is sourced and available after Travis installs oneapi.

nbeams:
Thanks @jeremylt for investigating the Intel issue.

Can someone help me understand the issue with the codecov failure?  When I look at the details, I don't see anything explaining why it would fail.  Under "coverage changes," it says, "Everything is accounted for!"  So how can I figure out what caused the problem?  Am I not looking at the right check?

jeremylt:
I think that's just because you cut the total number of lines in the codebase

valeriabarra:
I don't know.. I'm honestly confused by codecov in #530 too

jeremylt:
I think the issue on that PR might be more CodeCov getting confused because of the age of the branch and the rebases/merges. That branch started before we changed our CodeCov a little bit, which is probably not helping.

nbeams:
Um, I'm not sure what's going on with 443bad0b95654aa4c29e96e8951f664065060564 -- after rebasing, GitHub was still telling me the branch was out of date with main (it wasn't), so I tried clicking on the "update" option to see what would happen, and it made a commit of zero changes to say it merged `main` into the branch?  
If I try to undo this, will it go back to telling me the branch is out of date?

jedbrown:
Just getting to this now. There are two commits in 'main' that are not yet in this branch, so it is "out of date" (but not in a problematic way).

nbeams:
I rebased onto `main` to include the new commits...  🤷‍♀️  I see now that it has a different message than when there are conflicts, though, so I guess I shouldn't have worried about it!

nbeams:
If libCEED v0.7 is going to be released tomorrow, do we want to go ahead and merge this before, or wait until after the release?  (Or close without merging, if we decide not to go through with the consolidation?)

jeremylt:
The plan is to merge this today.

nbeams:
I'm not allowed to merge -- I guess because of the codecov failure?  But I don't know how to fix it since I can't see any files causing the failure...

jeremylt:
Yeah, main is protected and we can't merge with 'Required' tests failing. Jed can override.

jedbrown (approved):
LGTM. Let me know when you want it merged.

jedbrown:
And just confirming that you want squash-merge?

nbeams:
Sure, squash merge is fine.

nbeams:
I don't think there should be any conflicts with #620, but should I merge from main to make sure (if we will squash this branch anyway)?

jeremylt:
@nbeams should `magma/det` have been removed now that the deterministic restriction is in `cuda/ref`?

nbeams:
No, the standard `magma` backend uses its own element restrictions, which include atomic adds and are not deterministic.  So `magma/det` is still necessary; it just changes which CUDA backend it uses for the element restriction object delegation. 

I suppose in the future we could do testing to compare whether the deterministic `cuda-ref` restrictions actually perform better (or basically the same) and then we could remove the special `det` version and use them all the time...

PR #625, opened 2020-09-14
Travis - allow icc failure for now

jeremylt:


nbeams (approved):
Seems like a reasonable temporary measure to me.  Thanks @jeremylt

PR #626, opened 2020-09-17
Small doc fixes

valeriabarra:
I noticed a typo and I added some small edits before our release.

### DIFF
--- a/doc/sphinx/source/libCEEDapi.rst
+++ b/doc/sphinx/source/libCEEDapi.rst
@@ -341,7 +341,7 @@ constructors take a string representation specifying where the source for the
 implementation is found. This is used by backends that support Just-In-Time
 (JIT) compilation (i.e., CUDA and OCCA) to compile for coprocessors.
 For full support across all backends, these :ref:`CeedQFunction` source files must only contain constructs mutually supported by C99, C++11, and CUDA.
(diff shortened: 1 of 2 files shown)


jeremylt (approved):

valeriabarra:
Posting this question here because it's related to code doc. Why don't I see anything here https://codedocs.xyz/CEED/libCEED/?

jedbrown:
I'm not sure, but should we get rid of that site since I think all the functionality is available in the Sphinx site?

valeriabarra:
If I remember correctly, we said that we may have wanted to keep it because there were no hyperlinks to functions source code in Sphinx.

jedbrown:
Is anyone using it?  It's been empty since we renamed 'master' -> 'main', but I don't see any configuration to be able to change the branch.

https://codedocs.xyz/CEED/libCEED.log

This issue suggests it should be using the GitHub default branch (which is 'main' for libCEED), but it continues to build 'master': https://github.com/CodeDocs/CodeDocs/issues/4

(The project seems rather defunct.)

Patrick Sanan implemented a way to auto-link source code listings for PETSc (would require a little work to extract for use by libCEED and I'm not sure it's worth the effort right now).

valeriabarra:
It's ok to remove it now. I was just trying to remember why we said it was useful. But one thing is keeping a functionality that we thought someone may use and another thing is keeping something that now obviously does not work as we intended it to work. Fine with me if you want to go ahead and remove it. We should also update THE top-level README by removing its token.

jedbrown:
https://github.com/CEED/libCEED/pull/627

valeriabarra:
Thanks. I'm going to merge this now.

PR #628, opened 2020-09-22
Make - install hashing headers, used in MFEM

jeremylt:


nbeams (approved):
LGTM

include/ceed-hash.h
```diff
@@ -20,7 +20,7 @@
 #define _ceed_hash_h
 
 #include <ceed.h>
-#include <khash.h>
+#include <ceed-khash.h>
```
jedbrown, line 23: Renaming the file isn't enough to enable a user to include their own `khash.h` separately. Could you update the header guard in `ceed-khash.h`?

jedbrown (approved):

include/ceed-hash.h
jeremylt: updated

jeremylt:
It looks like rocm 3.8 + libCEED is broken?

nbeams:
Unfortunately, I don't currently have access to a machine with ROCm 3.8 (only 3.7), so I can't test it.  From the output of the failure, it seems like something's going wrong with `hipconfig`?

jedbrown:
I was just in the midst of upgrading/fixing a recurring packaging issue and rebooting. It passed now.

PR #632, opened 2020-09-23
Julia interface

pazner:
Add Julia interface for libCEED.

Features include:
- Low-level "C" interface. Access libCEED C API through Julia's FFI. Objects are not garbage collected, not type safe, etc.
- High-level "idiomatic Julia" interface. Wraps the low-level C interface with a higher-level more Julia-like interface. Objects are garbage collected, there is some interop between native Julia types and libCEED types (e.g. Julia's `Array` and libCEED's `CeedVector`).
- Define user Q-functions in Julia. These work both on CPU and with CUDA backends (not cuda/gen though).

There are Julia versions of examples 1 and 2.

Feedback would be appreciated, including API design decisions, missing features, etc.

Julia tests are run through Travis, and the coverage is reported on Codecov. There is also documentation, which is not currently built automatically, and could be improved for completeness.

Any help improving testing/coverage would be greatly appreciated!

### To do:
- [x] Build Julia documentation together with libCEED's readthedocs documentation
- [x] Add missing functionality (I think most critical stuff is there, but e.g. some functions relating to `CeedElemRestriction` are missing, among others)
- [x] Improve test coverage (**help wanted!**). Should the Julia codecov results have an associated "flag"?
- [x] Should the package attempt to build libCEED from source? Or automatically install a binary version? Right now the package requires the user to have already built libCEED, and uses an environment variable at package build-time to find the library.
### After merging:
- [ ] Set up Binder configuration
- [ ] Register the package with Julia's package registry
- [ ] Bump the version on the libCEED_jll binary package


.travis.yml
```diff
@@ -109,6 +109,20 @@ jobs:
...
+   - name: "Julia"
+     language: julia
+     os: linux
+     julia: 1
+     install:
+       - make info && make -j4
+       - export JULIA_LIBCEED_LIB=$(find $PWD/lib -name "libceed*")
+       - pushd julia/libCEED.jl
+     script:
+       - julia --project -e 'import Pkg; Pkg.build(); Pkg.test("libCEED"; coverage=true)'
```
jeremylt, line 122: We should also test formatting here.
```

        - julia --project -e 'using Pkg; Pkg.add("JuliaFormatter")'
        - julia --project -e 'using JuliaFormatter; format("src")'
        - git diff --exit-code src/
```

jeremylt:
I think the easiest way to hit full coverage is to include `jldoctest` in each case.

There are some functions that I don't think are very valuable to initially provide in the Julia and Rust interfaces. Blocked restrictions are intended for internal library use, so I wouldn't prioritize adding them. I would add generating vectors from an elem restriction and getting the multiplicity - those are handy.

I can look more in depth tomorrow (I hope...)

.travis.yml
pazner: Good call. I don't love the choices of JuliaFormatter (e.g. I find `a*x + b` to be much better than `a * x + b`), but it's not a big deal (and some of these things could be customized).

jeremylt: Agreed about the default style choices. If you know of good documentation/examples on modifying the style, it would be helpful in my other project. Once we have it dialed in, I like how formatters reduce code churn.

jedbrown (commented):
This is super cool. In the high-level QFunction interface, what would you think if the user were to write code for one quadrature point and have it broadcast/vectorize automatically (with macro help as needed)? Getting that loop out is helpful if the language can support it with equal performance.

Have you compared performance of the examples yet?

pazner:
> In the high-level QFunction interface, what would you think if the user were to write code for one quadrature point and have it broadcast/vectorize automatically (with macro help as needed)? Getting that loop out is helpful if the language can support it with equal performance.

Yes, that is fairly trivial to do. The reason I left the loop explicit is so the user could potentially do some preprocessing before looping over all Q-points.

> Have you compared performance of the examples yet?

Not counting compiler latency, performance is in the same ballpark as the C examples (maybe about 10% slower). CUDA performance is similar too (the Julia kernels tend to use more registers).

The first run of an example will unfortunately be pretty slow because of compile times, but subsequent runs are faster.

pazner:
I implemented a prototype of the automatic loop in user Q-functions, and I think it makes the design better.
The build and apply diffusion Q-functions (ex2) now become:
```julia
@interior_qf build_qfunc = (
    ceed,
    dim=dim,
    (J, :in, EVAL_GRAD, dim, dim),
    (w, :in, EVAL_WEIGHT),
    (qdata, :out, EVAL_NONE, dim * (dim + 1) ÷ 2),
    begin
        Jinv = inv(J)
        qdata .= setvoigt(w[] * det(J) * Jinv * Jinv')
    end,
)
@interior_qf apply_qfunc = (
    ceed,
    dim=dim,
    (du, :in, EVAL_GRAD, dim),
    (qdata, :in, EVAL_NONE, dim * (dim + 1) ÷ 2),
    (dv, :out, EVAL_GRAD, dim),
    begin
        dXdxdXdxT = getvoigt(qdata)
        dv .= dXdxdXdxT * du
    end,
)
```

In addition to getting rid of the Q-point loop (and associated `@inbounds` and `@simd` annotations), it also reduces the noise from the indexing and extracting views. The per-Q-point input arrays are statically sized, which means that operations like `inv(J)`, `det(J)`, `setvoigt`, `getvoigt`, and the matrix-vector products automatically dispatch to optimized implementations that are specialized on the dimension.

jedbrown:
This is great!

jedbrown:
@pazner You can add your name to `AUTHORS` in this PR.

Regarding build process, are there exemplary Julia packages we should compare with?

Sphinx has support for multilingual APIs, but I sort of doubt it's worth using given the quality of Julia and Rust documentation platforms and the fact that the interfaces are not precisely 1-1. I think I'd just link from the Sphinx site and have the Julia/Rust docs link to the Sphinx site for higher level concepts and the like.

We should make sure we have sufficient test/release process so we can publish Python/Julia/Rust packages as part of future releases.

Would you consider adding a Julia demo notebook for use in Binder?

pazner:
> Regarding build process, are there exemplary Julia packages we should compare with?

Some examples are [MPI.jl](https://github.com/JuliaParallel/MPI.jl), [PETSc.jl](https://github.com/JuliaParallel/PETSc.jl), [FFTW.jl](https://github.com/JuliaMath/FFTW.jl).

Many packages use something called [BinaryBuilder](https://binarybuilder.org), which can be used for cross-compilation to generate and host binaries for a bunch of platforms. This should allow the package to work "out of the box," but the user may have to compile libCEED on their own if they want GPU support, AVX, etc.

> Sphinx has support for multilingual APIs, but I sort of doubt it's worth using given the quality of Julia and Rust documentation platforms and the fact that the interfaces are not precisely 1-1. I think I'd just link from the Sphinx site and have the Julia/Rust docs link to the Sphinx site for higher level concepts and the like.

That sounds good. Do you think the Julia docs should be hosted using GitHub Pages?

> Would you consider adding a Julia demo notebook for use in Binder?

Yes, absolutely. I'll do that.

include/ceed.h
```diff
@@ -178,17 +178,17 @@ CEED_EXTERN int CeedErrorImpl(Ceed, const char *, int, const char *, int,
...
-                               const char *, va_list);
+                               const char *, va_list *);
 CEED_EXTERN int CeedErrorExit(Ceed, const char *, int, const char *, int,
-                              const char *, va_list);
+                              const char *, va_list *);
 CEED_EXTERN int CeedSetErrorHandler(Ceed ceed,
                                     int (*eh)(Ceed, const char *, int,
                                         const char *, int, const char *,
-                                        va_list));
+                                        va_list *));
```
pazner, line 191: For the purposes of handling errors in Julia, I tentatively changed this interface to use `va_list *` instead of `va_list`. My understanding is that `va_list` is implementation defined, and so is hard to interface with using FFI (this will probably come up in the Rust interface, cc @jeremylt). This allows libCEED errors to be thrown and caught as Julia exceptions.

If people don't like this idea I'm happy to revert it.

(On a related note, the behavior of `CeedErrorStore` is a bit weird because it will be called first on `parent`, which will invalidate the `va_list`, so one of the `Ceed` objects will have a garbage message. This seems like a bug.)

jeremylt: Did you see the latest fix in master where the error message is only set on the parent?

The message should be stored in the parent ceed and no other ceed. The user only has access to the parent ceed. and not the delegate ceeds created by the resource they specified, so those ceeds don't need to store the message.

.travis.yml
pazner: It's a bit tricky to customize the styles, it requires manipulating an AST-like object called FST. It is very powerful though. You can check out the script in `julia/libCEED.jl/.CeedStyle.jl`. It's a modification of the default style, but operators with the same precedence as `*` are not automatically surrounded with spaces (e.g. so fractions appear as `1/2` rather than `1 / 2`. Definitely agree about code churn, this should be useful.

include/ceed.h
pazner: Good to know, thanks! I hadn't seen the fix. Just merged latest main into this branch.

Let me know what you think of the `va_list` pointer.

jedbrown:
GH pages for docs seems fine. I could buy `libceed.org` and then we could have `julia.libceed.org` for those generated docs (with the Sphinx docs at the top level). We don't need to decide now.

julia/libCEED.jl/src/Basis.jl
```diff
@@ -0,0 +1,314 @@
...
+
+    apply!(b, nelem, tmode, emode, u_vec, v_vec)
+    Vector(v_vec)
+end
+
+"""
+    getdimension(b::Basis)
+
+Return the spatial dimension of the given [`Basis`](@ref).
+"""
```
jeremylt, line 233: I still feel like the tests would be better in the docstrings, like in the Rust interface. It would make the documentation better for the users and make it easier to add tests as we add functionality, in my opinion. I think it would be preferable to do the same for the Python interface as well. (Really, I should do that when I get the chance. Then the Python, Julia, and Rust interfaces would have example code in the auto documentation)

```suggestion
```jldoctest
# create basis
c = Ceed()
dim = 3
b = create_tensor_h1_lagrange_basis(c, dim, 1, 4, 6, GAUSS_LOBATTO)

# verify dimension
@assert getdimension(b) == dim
# output

``` # extra comment to force GitHub code suggestion to show this
"""

Comment (jeremylt) [reply julia/libCEED.jl/src/Basis.jl:233]: You can add 
```
DocMeta.setdocmeta!(libCEED, :DocTestSetup, :(using libCEED); recursive = true)
```
to avoid some repeated code in each doctest

Comment (pazner) [reply julia/libCEED.jl/src/Basis.jl:233]: This is a good point, and having the tests in the docstrings could be helpful. I think they could be useful to supplement the more traditional types of tests. In particular, some of tests have somewhat lengthy setup that would require duplication in the docstrings, and don't really lend themselves so well to that kind of testing.

Comment (jeremylt) [reply julia/libCEED.jl/src/Basis.jl:233]: Yeah, I'm not wild about showing the full setup some of the functions need. I think individual lines that don't help much can be hidden:
```
```jldoctest
ceed = Ceed() # hide
# create basis
dim = 3
basis = create_tensor_h1_lagrange_basis(ceed, dim, 1, 2, 3, GAUSS)

# verify dimension
@assert getdimension(basis) == dim

# output

``` # extra comment for markdown
```

the user will then only see
```
# create basis
dim = 3
basis = create_tensor_h1_lagrange_basis(ceed, dim, 1, 2, 3, GAUSS)

# verify dimension
@assert getdimension(basis) == dim

# output

``` # extra comment for markdown
```

jeremylt: Related - it seem like the Julia community is a fan of
```
function Base.getproperty(basis::Basis, foo::Symbol)
    if foo == :dim
        return getdimension(basis)
    elseif ....
```
so that
```
@assert basis.dim == 3
```
works.

Thoughts?

pazner: Yep, we can add those if we want. It hides what is really going on a little bit, which has the potential to be confusing, but I'm fine with it.

Re: `dim` vs. `dimension`, I think this particular abbreviation is fine (see e.g. `Base.ndims`, `Base.permutedims`, etc.)

pazner: For the doctests, I think it makes sense to identify functions for which the documentation itself will be improved by adding an example that serves as a test (your choice of `getdimension` is one of them). In this case, it's helpful to have a doctest.

However, in the case of a test which is useful for testing the code, but which is not so helpful from an explanatory perspective, I think a traditional test is a better match.

pazner:
I think this is pretty much ready to go:

* The test coverage is pretty thorough (I haven't done the doctests yet, but that is something that would be nice to do).
* Code style is enforced on Travis.
* The documentation is built and deployed automatically by Travis. It is available through GitHub Pages at https://ceed.exascaleproject.org/libCEED-julia-docs/dev/. When this is merged, the development branch in the docs script will need to be changed from `will/julia` to `main`.
* By default, a prebuilt binary version of libCEED will be used. The user can configure the Julia interface to use a custom-built version if needed.
* Almost all of the public interface has been wrapped. Some linear algebra routines (e.g. QR) are not wrapped, since equivalent functionality is available through the Julia standard library. Based on Jeremy's comment, I didn't do blocked restrictions. Otherwise, all that is missing is the Operator assembly and multigrid. These should be straightforward to add if there is a need.

In principle, this can work well with Binder, but there are some issues with placing the Binder configuration files in the `binder` subdirectory (looks like a bug in Binder to me, but I could also be doing something wrong). Perhaps a Binder tutorial with notebooks can be in a separate repo.

jeremylt:
I think we should add a new page to the current documentation (perhaps `doc/sphinx/source/ffi`) that links to the Julia (and soon Rust and Python) interface documentation.

julia/LibCEED.jl/src/Operator.jl
```diff
@@ -123,3 +123,46 @@ function apply_add!(
 )
     C.CeedOperatorApplyAdd(op[], vin[], vout[], request[])
 end
+
+"""
+    assemble(op::Operator; request=RequestImmediate())
```
jeremylt, line 128: I don't think it's intuitive for `assemble(operator)` to return the assembled representation of the QFunction. QFunction assembly isn't really intended for users anyway, so it might be preferable to omit this.

interface/ceed-operator.c
```diff
@@ -57,6 +57,7 @@ int CeedOperatorCreateFallback(CeedOperator op) {
   if (!op->ceed->opfallbackceed) {
     ierr = CeedInit(fallbackresource, &ceedref); CeedChk(ierr);
     ceedref->opfallbackparent = op->ceed;
+    ceedref->Error = op->ceed->Error;
```
jeremylt, line 60: I agree this is needed, but I think this might not be enough to ensure that the error message propagates back to the true parent ceed.

interface/ceed-operator.c
```diff
@@ -1067,7 +1067,6 @@ int CeedOperatorLinearAssembleAddDiagonal(CeedOperator op, CeedVector assembled,
       ierr = CeedOperatorCreateFallback(op); CeedChk(ierr);
     }
     // Assemble
-    ierr = CeedVectorSetValue(assembled, 0.0); CeedChk(ierr);
```
jeremylt, line 1070: Thanks for catching this. I'm hoping we can get this branch and these various small Bugfixes into master soon

julia/LibCEED.jl/src/Operator.jl
pazner: Ok, I'll get rid of it.

interface/ceed-operator.c
pazner: Yeah, I'm not sure if this is sufficient. It worked in the particular case I was working with, but I'm not sure about the general case.

julia/LibCEED.jl/docs/make.jl
```diff
@@ -23,5 +23,6 @@ makedocs(sitename="LibCEED.jl Docs",
          ])
 
 deploydocs(
-    repo = "github.com/CEED/libCEED-julia-docs.get"
+    repo="github.com/CEED/libCEED-julia-docs.get",
```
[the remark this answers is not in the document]
pazner: Yes, it ended up working pretty well with the deploy key. It's a private repo for now — I just added you, and I suppose I should either make it public or add the entire "CEED/CEED" team?

.travis.yml
```diff
@@ -109,6 +109,36 @@ jobs:
...
+       - julia --project -e 'import Pkg; Pkg.build(); Pkg.test("LibCEED"; coverage=true)'
+       - unset JULIA_LIBCEED_LIB && julia --project -e 'import Pkg; Pkg.build(); Pkg.test("LibCEED")'
+       - >
+         julia --project=.style/ -e 'import Pkg; Pkg.instantiate()'
+         && julia --project=.style/ .style/ceed_style.jl
+         && git diff --exit-code src test examples
+     after_success:
+       - popd
+       - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder("julia/LibCEED.jl/src"))'
+   - stage: "Documentation"
```
jeremylt, line 131: Two questions

a) Can we name this documentation job something like "Julia Documentation"

b) Why Julia 1 instead of 1.5? 1.5 seems to run faster on my other project and it seems reasonable to test with the latest released Julia.

jeremylt: Also, can we call the stage "documentation" with a lower case d? The default testing stage is named "test" with a lower case t.

pazner: >a) Can we name this documentation job something like "Julia Documentation"

Yes, sure, I'll change it.

> b) Why Julia 1 instead of 1.5? 1.5 seems to run faster on my other project and it seems reasonable to test with the latest released Julia.

1 means the latest stable 1.x.y release, so I believe it is running 1.5.2 automatically.

doc/sphinx/source/ffi.rst
```diff
@@ -0,0 +1,12 @@
...
+
+libCEED provides high-level interfaces using the Julia and Python programming
+languages.
+
+More information about the Julia interface can be found at the `LibCEED.jl
+documentation <http://ceed.exascaleproject.org/libCEED-julia-docs/dev/>`_.
+
+Usage of the Python interface is illustrated through a sequence of `Jupyter
+Notebook
+tutorials <https://github.com/CEED/libCEED/tree/main/examples/tutorials>`_.
```
valeriabarra, line 12: Thanks for adding this. I don't know if it's any useful to reference libCEED's SciPy paper here or not. It has some details, but not even much in depth.

julia/LibCEED.jl/src/Misc.jl
```diff
@@ -0,0 +1,119 @@
...
+    J[3]*(J[4]*J[8] - J[5]*J[7])
+)
+#! format: on
+
+"""
+    setvoigt(J::StaticArray{Tuple{D,D},T,2})
+    setvoigt(J, ::CeedDim{dim})
+
+Given a symmetric matrix `J`, return a `SVector` that encodes `J` using the [Voigt
+convention](https://en.wikipedia.org/wiki/Voigt_notation).
```
valeriabarra, line 32: Thanks for adding the reference here.

Just curious. I've seen this question come up somewhere else. Have you ever found in your experience a way to insert citations from a bibtex file and have them rendered nicely in Julia's docstrings?

doc/sphinx/source/ffi.rst
pazner: Good idea, I added a link to the SciPy paper.

julia/LibCEED.jl/src/Misc.jl
pazner: I haven't found anything good for bibtex references. I think there are a few "homegrown" solutions, but nothing official. There is a proposal to implement something in Documenter.jl, so I expect that it will happen at some point.

valeriabarra: Good to know. Thanks!

doc/sphinx/source/ffi.rst
jeremylt: Do we want to make an auto-documentation deployment for the Python interface and link it here? (Probably best in separate PR). It doesn't need to be fancy, but it would provide a reference for the function signatures.

valeriabarra: I think that sounds like a great idea. Yes, probably better in a separate PR

interface/ceed-operator.c
jeremylt: I think we just need to change `CeedErrorStore`
```
...
  if (ceed->parent)
    return CeedErrorStore(ceed->parent, filename, lineno, func, ecode, format,
                          args);

  if (ceed->opfallbackparent)
    return CeedErrorStore(ceed->parent, opfallbackparent, lineno, func, ecode,
                          format, args);
...
```
(total guess on line break for formatting)

pazner: Yes, that is a good point. Another option would be to change all the relevant `CeedError` calls to use `ceedparent` (obtained from `CeedGetOperatorFallbackParentCeed`) rather than `ceed`. But this route seems better to me.

julia/LibCEED.jl/docs/make.jl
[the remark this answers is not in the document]
jedbrown: Seems there's no harm in making it public.

include/ceed.h
jedbrown: We were mirroring `vprintf` conventions, but I'm not aware of any reason that passing a pointer would be problematic.

jedbrown (approved):
Looks good overall; just a couple threads to resolve. Would you mind squashing commits that go together on the new functionality, but keeping changes to existing components, such as `va_list*`, in their own commits?

jeremylt:
My only open request on this is the fix for CeedErrorStore for the fallback ceed.

julia/LibCEED.jl/docs/make.jl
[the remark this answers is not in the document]
pazner: Ok, it's public now.

interface/ceed-operator.c
pazner: What do you think of the `ceedparent` option?

pazner:
> Looks good overall; just a couple threads to resolve. Would you mind squashing commits that go together on the new functionality, but keeping changes to existing components, such as `va_list*`, in their own commits?

Great, sounds good! I'll tie up the last few loose ends and then squash the commits.

interface/ceed-operator.c
jeremylt: I do not think that it's a good option.

It scatters error handler logic into multiple places, makes the logic different for the two types of parents, and it will be easy to forget this special change only for fallback delegation in future code.

julia/LibCEED.jl/docs/make.jl
```diff
@@ -0,0 +1,28 @@
...
+                "Quadrature.md",
+             ],
+             "C.md",
+             "UserQFunctions.md",
+             "Examples.md",
+         ])
+
+deploydocs(
+    repo="github.com/CEED/libCEED-julia-docs.git",
+    devbranch="will/julia",
```
pazner, line 27: The `devbranch` here will need to be changed to main when we merge.

julia/LibCEED.jl/src/Misc.jl
valeriabarra: Just to follow up. I believe this would do it
https://github.com/ali-ramadhan/DocumenterCitations.jl

pazner: Very cool, this could be useful. Thanks!

jeremylt:
Are we good to merge (with a swap on the target branch for doc deployment)?

pazner:
> Are we good to merge (with a swap on the target branch for doc deployment and temp workoround to drop magma ci)?

I believe everything is ready on my end.

jedbrown:
Sounds good. Are there any logistics you need me to do prior to merging?

pazner:
> Sounds good. Are there any logistics you need me to do prior to merging?

Just changing the devbranch for the documentation: https://github.com/CEED/libCEED/pull/632#discussion_r511653616

After it is merged, I will update the JLL binary package and register the LibCEED.jl package with the Julia package registry.

julia/LibCEED.jl/docs/make.jl
jedbrown: Done in the merge.

jedbrown:
Merged now. Thanks!

include/ceed.h
jeremylt: It turns out that `clang-tidy` hates this

pazner: Apparently there are [some bugs](https://bugs.llvm.org/show_bug.cgi?id=41311) in clang-tidy relating to `va_list`. I don't know if we're seeing the same thing. Could we use a typedef or something to make clang-tidy happy?

jeremylt: Yeah, I saw those bugs. I'm personally fine with `// NOLINT`ing those lines since there is no critical logic in those 3 lines.

PR #634, opened 2020-09-28
OCCA - update backend resource names to match other GPU backends

jeremylt:


jedbrown (approved):

jeremylt:
Issue: should the OCCA backends respond to `/gpu/cuda` and `/gpu/hip`? If not, I should raise the prioritization of these backends so they don't accidentally match against those strings. If so, how should the prioritization of the OCCA backends mesh with the other backends?

The same question for MAGMA.

I think the crux of this is that `/cuda` is both a backend family and a hardware specification.

jedbrown:
> I think the crux of this is that /cuda [and now /hip] is both a backend family and a hardware specification.

In these cases, I think only the "native" API/backend should match. (In the case of HIP, the hardware part is ROCm and HIP is just the vendor-recommended API to reach it. In that sense, `/gpu/rocm` could specialize to `/gpu/rocm/hip`, though this much pedantry may not be helpful.

With CUDA, it refers both to the device/driver level and to the API.

jeremylt:
Updated. I gave OCCA the 200s in the priority so it won't collide with the native backends accidentally. We could give MAGMA the 100s.

jedbrown:
Hmm, there are still lots of references to `/gpu/occa`. Want to fix those now or deal with later?

jeremylt:
I think we can go one of two ways. We could have only

`/*/occa` (best available OCCA backend, or specifiers passed)
`/cpu/occa/serial`
`/cpu/occa/openmp`
`/gpu/cuda/cuda`
`/gpu/hip/occa`

or we could have

`/*/occa` (best available OCCA backend, or specifiers passed)
`/cpu/occa` (best available OCCA backend on host)
`/cpu/occa/serial`
`/cpu/occa/openmp`
`/gpu/occa` (best available OCCA backend on any device)
`/gpu/cuda/cuda`
`/gpu/hip/occa`

I don't think it makes a huge difference for most users since the OCCA backend is still slower than the others, so I guess we just go with whichever we feel is more consistent with our overall convention we want to establish.

jedbrown:
I'm reluctant for `/cpu/self` vs `/cpu/openmp` (meaning use the OpenMP thread pool/runtime, which would respond to `OMP_*` environment variables) to ever be implicit.

If the device is ever to be implicit, I think I'd rather `/gpu/*/occa` than `/gpu/occa` (which projects symlink semantics rather than just wildcards to resolve to `/gpu/cuda/occa` or whatever). I suppose I could see both sides and wouldn't have a problem so long as we have a way to visualize the convenience symlinks and there is a stable resolution algorithm so people can have reproducibility. I'm just not sure the value is sufficient.

jeremylt:
@jedbrown, this should be all set now

jeremylt:
@nbeams @abdelfattah83 I added `cuda` to the MAGMA backend resource identifier in this PR in preparation for v0.7 releasing so all backends strictly follow the convention

nbeams:
Sounds good, I'll update the backend priority numbers in my PR for adding `/gpu/hip/magma` as well.

PR #635, opened 2020-09-28
Python - pin pytest to 6.0.2

jeremylt:


jedbrown:
Damn, I have the same bug on Noether, with python-3.8.5 and pytest-6.0.2. This doesn't fix it (passed on my first run and failed every time after).

jedbrown:
Shall we close this because it doesn't actually fix this issue?

PR #636, opened 2020-09-28
Jed/python debugging

jedbrown:
A typo fix and some debugging assistance (should be irrelevant to users).


jeremylt (approved):
looks good to me

PR #637, opened 2020-09-29
Jeremy/py debugging

jeremylt:
This doesn't fix the problem, but I've found a few other things that we may want to fix


jeremylt:
Ready to squash + merge if it passes for you @jedbrown

PR #638, opened 2020-09-29
Make - add cpp files to code coverage

### LINKED ISSUE #611: OCCA and HIP Disapeared from CodeCov
Our CodeCov jumped to 98% because the OCCA and HIP backends disappeared from the coverage statistics.
image

jeremylt:
This might fix the fact that OCCA is missing from CodeCov.

Closes #611

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,6 +123,7 @@ FFLAGS ?= $(OPT) $(FFLAGS.$(FC_VENDOR))
 
 ifeq ($(COVERAGE), 1)
   CFLAGS += --coverage
+  CXXFLAGS += --coverage
   LDFLAGS += --coverage
 endif
 


Makefile
```diff
@@ -123,6 +123,7 @@ FFLAGS ?= $(OPT) $(FFLAGS.$(FC_VENDOR))
 
 ifeq ($(COVERAGE), 1)
   CFLAGS += --coverage
+  CXXFLAGS += --coverage
```
jeremylt, line 126: We could instead change `CFLAGS` to `CPPFLAGS` here instead.

Unrelated: It seems strange to have both `CPPFLAGS` and `CXXFLAGS` running around.

jeremylt:
image

Confirmed, this restores OCCA backend CodeCov tracking

Makefile
jedbrown: `CPPFLAGS` is for the C preprocessor (things like `-I`).
See `make -p -f /dev/null |grep CPPFLAGS`, for example.

jedbrown (approved):

PR #641, opened 2020-10-05
CI - add hipMAGMA to CI

jeremylt:
This should enable testing of hipMAGMA in CI

side note - at least when I tested it manually, the MAGMA backend builds remarkably slowly

### DIFF
--- a/backends/magma/ceed-magma-basis.c
+++ b/backends/magma/ceed-magma-basis.c
@@ -372,6 +372,22 @@ int CeedBasisCreateTensorH1_Magma(CeedInt dim, CeedInt P1d, CeedInt Q1d,
   Ceed ceed;
   ierr = CeedBasisGetCeed(basis, &ceed); CeedChk(ierr);
 
+  // Check for supported parameters
+  CeedInt ncomp = 0;
+  ierr = CeedBasisGetNumComponents(basis, &ncomp); CeedChk(ierr);
+  if (ncomp > 3)
+    // LCOV_EXCL_START
+    return CeedError(ceed, 1, "Magma backend does not support tensor bases with more than 3 components");
+  // LCOV_EXCL_STOP
+  if (P1d > 10)
+    // LCOV_EXCL_START
+    return CeedError(ceed, 1, "Magma backend does not support tensor bases with more than 10 nodes in each dimension");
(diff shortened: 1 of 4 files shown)


jeremylt:
Unfortunately, it looks like the Gitlab job didn't fire.

jeremylt:
There were some failures in the job: https://gitlab.com/libceed/libCEED/-/pipelines/198308411/test_report

nbeams:
I also don't see the failures when running the tests manually.  Both t506 and t318 should be known failures for MAGMA -- they use values of Q1d higher than the templates are instantiated for (which is only up to 10).  Is the job on GitLab somehow not correctly setting the "skip status" for these tests?

nbeams:
> side note - at least when I tested it manually, the MAGMA backend builds remarkably slowly

I have also noticed very slow compilation for the HIP kernels (much slower than the CUDA version).  Unfortunately, I'm not really sure if there's anything we can do about this, short of abandoning the templates and using `hiprtc`...however if anyone has ideas, I would be happy to hear them!  It drives me crazy to have to wait so long to re-compile and test the HIP backend.
(With new versions of HIP coming out at a fairly rapid pace, though, it's hard to say for sure what will be changed about the compiler in the future.)

backends/magma/ceed-magma-basis.c
```diff
@@ -372,6 +372,22 @@ int CeedBasisCreateTensorH1_Magma(CeedInt dim, CeedInt P1d, CeedInt Q1d,
   Ceed ceed;
   ierr = CeedBasisGetCeed(basis, &ceed); CeedChk(ierr);
 
+  // Check for supported parameters
```
jeremylt, line 375: @nbeams feel free to modify, I just added this to move the error message up to basis creation rather than application

jeremylt:
Giltab pipeline passed now. I'll switch the target back to the hipMAGMA branch. Feel free to merge this into your branch if you're satisfied with it.

jeremylt:
86% code coverage for the MAGMA backend

backends/magma/ceed-magma-basis.c
nbeams: LGTM, thanks

nbeams (approved):

PR #655, opened 2020-11-02
Solids - new BCs

jeremylt:


examples/solids/src/boundary.c
```diff
@@ -66,9 +66,9 @@ PetscErrorCode BCClamp(PetscInt dim, PetscReal loadIncrement,
               kx = clampMax[3], ky = clampMax[4], kz = clampMax[5];
   PetscScalar c = cos(theta), s = sin(theta);
 
-  u[0] = lx + s*(-kz*y + ky*z) + (1-c)*(-ky*ky+kz*kz*x + kx*ky*y + kx*kz*z);
-  u[1] = ly + s*(kz*x + -kx*z) + (1-c)*(kx*ky*x - (kx*kx+kz*kz)*y + ky*kz*z);
-  u[2] = lz + s*(-ky*x + kx*y) + (1-c)*(kx*kz*x + ky*kz*y - (kx*kx+ky*ky)*z);
+  u[0] = lx + s*(-kz*y + ky*z) + (1-c)*(-(ky*ky+kz*kz)*x + kx*ky*y + kx*kz*z);
```
jeremylt, line 69: This typo was not helping our convergence

jedbrown: Uff.

examples/solids/README.rst
```diff
@@ -46,6 +46,9 @@ The elasticity min-app is controlled via command-line options, the following of
 
 Note: The default for a clamped face is zero displacement. All displacement is with respect to the initial configuration.
 
+   * - :code:`-bc_plane [int list]`
+     - List of face sets on which affix to the plane :code:`-bc_plane_[facenumber] [x,y,z,n_x,n_y,n_z]`
```
jedbrown, line 50: Can these docs explain the arguments? Is it a translation with specified normal?

Should the name indicate it's slip, like `-bc_plane_slip 2,6,9 -bc_plane_slip_2_translate 1,.5,0 -bc_plane_slip_2_normal 0,1,0`?

jeremylt: The six fields are the point (x, y, z) and normal (nx, ny, nz) describing the plane that the face must adhere to. I can add a more detailed explanation.

jeremylt: I would like to keep the 6 arguments together to reduce the odds that someone will half define the boundary plane, but I can add `slip` to the name.

jedbrown: So the plane isn't translated or rotated? I was thinking we'd have equivalent semantics to `clamp`, but where the plane is free-slip (but stays in contact even under tension). If you don't move the plane from the initial configuration, why do you need to know its position versus just the normal?

jeremylt: I can also add translation/rotation of the boundary plane easily. 

I'd still need 1 pt and 1 vector to uniquely ID the plane in space and I think its better for the user to provide those two quantities together.

jeremylt: Well, rotation would be silly to add for a plane in this case, but I can add translation

jedbrown: Rotation can tilt the plane, which is relevant.

Every quadrature point of a face in the initial configuration is on the plane prior to any affine transformation that may be applied to the plane. Seems error-prone to ask for that to be specified independently.

jeremylt: -bc_plane_slip 2,6,9 

I suppose I don't understand what these proposed values would mean then.

The way I was picturing this, the user specifies a plane (which may be moving) and we project the face onto this plane.

jeremylt: Wait, I was thinking of u as position, not displacement. Give me a sec to untwist my thoughts

jedbrown: I figured we would take the surface point from initial configuration, apply the affine transformation, then apply Dirichlet normal/free slip in the current configuration.

jeremylt: I now agree that initially the user should only supply a normal vector
```
-bc_slip_2 1,2,5
```
though, the user can still create some physically weird behaviour if the normal and the face don't actually agree, right?

It seems like to fully specify movement (which I hadn't really considered at first), we'd want to use the affine transformation to describe how the point moves, and separately describe how the normal vector moves.

jedbrown: We could infer conservative vertex normals as in https://onlinelibrary.wiley.com/doi/pdf/10.1002/fld.663?casa_token=ukH103stMxcAAAAA:td75fmDTXm4Pe8nlNiafHeshBRUfBYFB9BQaaemBVohSwnRHxEAFDxYz810j7wy7sRJyR5iTGZB8Qaw

This would be exact for planar surfaces and locally correct for small displacement on curved surfaces. It would need a bonified geometric model for large deformation on curved surfaces.

jeremylt: I think for now it would be reasonable to have something like

```
# user given normal
-bc_slip_plane_2_normal_initial 0,0,1

# adjust normal vector in 3D
-bc_slip_plane_2_normal_final 1,0,0

# move plane in 3D like with BC clamp
-bc_slip_plane_2_point_translate 0.1,0,0.2
-bc_slip_plane_2_point_rotate 1,2,0,1
```
I'm not fully in love with the names.

That's far more than what I need for the p-multigrid paper

jedbrown: Why specify both normals? They transform under the provided affine transformation (expressed as translation and rotation).

jeremylt: Yeah, I think that's right. I just need to start this fresh tomorrow. I started with a completely different (and bugged) mental picture.

jeremylt: Alright, what I have in mind:

The boundary plane is defined by a translating and rotating point, as with BCClamp, and a rotating normal. I then project the current displacement onto the plane.

In order to do this, I need to use `DM_BC_ESSENTIAL_FIELD` or `DM_BC_ESSENTIAL_BD_FIELD` since `DM_BC_ESSENTIAL` does not provide the current displacement values to project onto the plane, but DMPlex does not seem to use the user ctx for those BC functions, unless I'm mistaken (in `DMProjectPoint_Field_Private`). I'm hoping that's just a small change though.

jeremylt: I need to test, but the latest version + [this MR](https://gitlab.com/petsc/petsc/-/merge_requests/3393) will work, I think

jeremylt:
traction
```
$PETSC_DIR/$PETSC_ARCH/bin/mpiexec -n 8 ./elasticity -problem hyperFS -dm_plex_box_faces 16,4,4 -dm_plex_box_lower -400,-100,-100 -dm_plex_box_upper 400,100,100 -nu 0.3 -E 1 -bc_clamp 5 -bc_traction 6 -bc_traction_6 0,0,-0.005 -snes_monitor -num_steps 30 -view_soln -snes_ksp_ew -snes_ksp_ew_alpha 2 -snes_rtol 1e-6
```

jeremylt:
sign_flip
Sign fixed now

jedbrown:
Cool! Is Newton converging properly? Do you really need all those load increments for the traction boundary?

jeremylt:
It converges in a single Newton step, I just wanted 30 for a nice gif.

valeriabarra:
Very cool Jeremy! Good job!
Questions:
- is there interest in making this BC time-dependent? It's cool to see the elastic behavior when the traction is taken off (i.e., you should see the solid bounce back to its original equilibrium configuration).
- do you have a diagnostic measure of the displacement, so that you can confirm that the material behaves as it should by the constitutive model (i.e., linearly in this case)?

jeremylt:
I have no plans to incorporate time at this time, as we don't have a TS in this example yet. I agree it would be interesting, but it's not on the straight line path between me and my dissertation.

I don't yet have a diagnostic measure or manufactured solution involving the traction BCs.

jedbrown:
An elastodynamics solver would be cool (even if just linear elastodynamics for seismic simulation, and perhaps inversion) but is way out of scope for this PR. Such a solver would be really valuable to our colleagues who work on PyLith (and all we'd need to implement is the volumetric terms).

valeriabarra:
Fair.
The slowness in convergence you see at the beginning might have something to do with the discontinuity of the stress tensor caused by the applied BC? 
In my experience with dynamic problems, when using time-dependent traction BCs, it's always better not to have the BC act like an "on/off" switch, but to use a ramp. In fact, people use different types of ramps, like linear or nonlinear, avoid this.

jedbrown (approved):

jeremylt:
Oh, that makes sense. The slow convergence is definitely better when I am using more load increments, though the total number of KSP its is also higher with more load increments.

PR #661, opened 2020-11-10
Some suggestions for the Rust interface

YohannDudouit:
- Avoid anything that can panic inside unsafe blocks (e.g. casting),
- Use slice in favor of Vec,
- Add Basis, ElemRestriction, Operator, QFunction, Vector to the prelude,
- Reorganize some of the unsafe blocks.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -36,11 +36,14 @@ pub mod prelude {
         #![allow(dead_code)]
         include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
     }
-    pub use crate::basis::{self, BasisOpt};
-    pub use crate::elem_restriction::{self, ElemRestrictionOpt};
-    pub use crate::qfunction::{self, QFunctionOpt};
-    pub use crate::vector::{self, VectorOpt};
-    pub use crate::{ElemTopology, EvalMode, MemType, NormType, QuadMode, TransposeMode};
+    pub use crate::{
+        basis::{self, Basis, BasisOpt},
+        elem_restriction::{self, ElemRestriction, ElemRestrictionOpt},
+        operator::{self, Operator, CompositeOperator},
+        qfunction::{self, QFunction, QFunctionOpt, QFunctionByName},
+        vector::{self, Vector, VectorOpt},
+        ElemTopology, EvalMode, MemType, NormType, QuadMode, TransposeMode,
+    };
     pub(crate) use std::ffi::CString;
     pub(crate) use std::fmt;
 }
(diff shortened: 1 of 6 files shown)


rust/src/lib.rs
```diff
@@ -486,8 +490,8 @@ impl Ceed {
     /// # let ceed = libceed::Ceed::default_init();
     /// let qf = ceed.q_function_interior_by_name("Mass1DBuild".to_string());
     /// ```
-    pub fn q_function_interior_by_name(&self, name: String) -> crate::qfunction::QFunctionByName {
-        crate::qfunction::QFunctionByName::create(self, name)
+    pub fn q_function_interior_by_name(&self, name: impl Into<String>) -> QFunctionByName {
```
jedbrown, line 493: Can we just use `&str`?

jedbrown (approved):

rust/src/operator.rs
```diff
@@ -1393,15 +1384,11 @@ impl Operator {
     /// ```
     pub fn create_multigrid_level_H1(
         &self,
-        p_mult_fine: &crate::vector::Vector,
-        rstr_coarse: &crate::elem_restriction::ElemRestriction,
-        basis_coarse: &crate::basis::Basis,
+        p_mult_fine: &Vector,
+        rstr_coarse: &ElemRestriction,
+        basis_coarse: &Basis,
         interpCtoF: &Vec<f64>,
```
jeremylt, line 1390: This would also need to be switched to `&[f64]`

rust/src/qfunction.rs
```diff
@@ -168,25 +170,25 @@ impl fmt::Display for QFunctionByName {
 // -----------------------------------------------------------------------------
 impl QFunctionCore {
     // Common implementation
-    pub fn apply(&self, Q: i32, u: &Vec<crate::vector::Vector>, v: &Vec<crate::vector::Vector>) {
+    pub fn apply(&self, Q: i32, u: &[Vector], v: &[Vector]) {
+        let mut u_c = [std::ptr::null_mut(); 16];
```
jeremylt, line 174: I specifically used `Vec` as the input/output of QFunctions so that the user would not need to know the magic number 16.

jeremylt: Woops, wrong place

rust/src/lib.rs
YohannDudouit: Do you mean everywhere?

jedbrown: We don't need a `String` unless we want ownership inside the function. I think `&str` is preferred unless it's specifically un-ergonomic.

jeremylt (approved):

jeremylt:
I went ahead and merged these changes since they look good

PR #663, opened 2020-11-19
README - update ffi information

jeremylt:
Minor update to the discussion of the additional interfaces in the README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -126,7 +126,7 @@ in the Julia package manager or in a clone of the repository via::
     julia> # press ] to enter package manager
     (env) pkg> build LibCEED
 
-Rust users can install using::
+Rust users can build using::
 
     LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/libceed.so
     cd rust
--- a/README.rst
+++ b/README.rst
@@ -79,7 +79,7 @@ Building
 ----------------------------------------
 
 The CEED library, ``libceed``, is a C99 library with no required dependencies, and
-with Fortran and Python interfaces.  It can be built using::
+with Fortran, Python, Julia, and Rust interfaces.  It can be built using::
 
     make
 
@@ -89,11 +89,6 @@ or, with optimization flags::
 
 These optimization flags are used by all languages (C, C++, Fortran) and this
 makefile variable can also be set for testing and examples (below).
-Python users can install using::
-
-    pip install libceed
-
-or in a clone of the repository via ``pip install .``.
 
 The library attempts to automatically detect support for the AVX
 instruction set using gcc-style compiler options for the host.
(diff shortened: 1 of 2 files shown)


jeremylt:
I tossed an update to the release notes in here too.

doc/sphinx/source/releasenotes.rst
```diff
@@ -47,7 +49,7 @@ Interface changes
 New features
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 * New HIP backend: ``/gpu/hip/ref``.
-* CeedQFunction support for user ``CUfunction``s in some backends
+* CeedQFunction support for user ``CUfunction``\s in some backends
```
jeremylt, line 52: If we don't escape this s, the inline code display fails and we are left with ``

README.rst
```diff
@@ -108,6 +103,32 @@ or::
...
+
+Python users can install using::
+
+    pip install libceed
+
+or in a clone of the repository via ``pip install .``.
+
+Julia users can install using::
+
+    ] add LibCEED
```
valeriabarra, line 119: I would add the prompts to avoid confusion:
```suggestion
   $ julia
   julia> ]
   pkg> add LibCEED
```

README.rst
```diff
@@ -108,6 +103,32 @@ or::
...
+
+    JULIA_LIBCEED_LIB=/path/to/libceed.so julia
+    julia> # press ] to enter package manager
+    (env) pkg> build LibCEED
+
+Rust users can install using::
+
+    LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/libceed.so
+    cd rust
+    cargo build
```
jedbrown, line 131: This doesn't tell a user how to use it, just how to build it. We'd need to tell users what to put in their `Cargo.toml`. (Publishing to crates.io is still a todo.)

jedbrown (approved):

README.rst
jeremylt: How about the updated language?

README.rst
```diff
@@ -108,6 +103,36 @@ or::
...
+
+   $ julia
+   julia> ]
+   pkg> add LibCEED
+
+in the Julia package manager or in a clone of the repository via::
+
+    JULIA_LIBCEED_LIB=/path/to/libceed.so julia
+    julia> # press ] to enter package manager
+    (env) pkg> build LibCEED
```
valeriabarra, line 127: I know this does not pertain this PR in particular, but I feel that the implicit dependencies needed for the `LibCEED.jl` Julia interface libCEED should be documented. 
For instance, when I tried to build it, I got the following 
```
(@v1.5) pkg> add LibCEED
  Resolving package versions...
ERROR: Unsatisfiable requirements detected for package CEnum [fa961155]:
 CEnum [fa961155] log:
 ├─possible versions are: [0.1.0-0.1.2, 0.2.0, 0.3.0, 0.4.0-0.4.1] or uninstalled
 ├─restricted to versions * by an explicit requirement, leaving only versions [0.1.0-0.1.2, 0.2.0, 0.3.0, 0.4.0-0.4.1]
 ├─restricted by compatibility requirements with LibCEED [2cd74e05] to versions: 0.4.0-0.4.1
 │ └─LibCEED [2cd74e05] log:
 │   ├─possible versions are: 0.1.0 or uninstalled
 │   └─restricted to versions * by an explicit requirement, leaving only versions 0.1.0
 └─restricted by compatibility requirements with CuArrays [3a865a2d] to versions: 0.2.0 — no versions left
```

I do see the compatibility requirements in [Project.toml](https://github.com/CEED/libCEED/blob/main/julia/LibCEED.jl/Project.toml) file, but I couldn't find this anywhere else in the docs. Should we rely on users having to go read the file themselves?

jeremylt: @pazner is best suited to answer this

PR #664, opened 2020-11-20
Actions - update for change in ICC CI example repo

jeremylt:


jeremylt:
I was hoping to directly use their example repo for better stability, but it seems the opposite might be happening.

https://github.com/oneapi-src/oneapi-ci

It's bizarre to me that they don't just have an action set up that others can use.

valeriabarra (approved):

PR #668, opened 2020-12-09
Add hip-shared and hip-gen backends

nbeams:
This PR adds ported versions of the `cuda-shared` and `cuda-gen` backends for AMD GPUs.

On an MI50 GPU, this version of `hip-gen` offers up to 4x speedup over `hip-ref` for BP3 (though many more cases are in the 2x-3x range).  This initial version of `hip-shared` has a more modest advantage over `hip-ref`, and seemingly only for higher orders of basis functions (> 4).

1ac155d3f1d39377a1304752acd8a47f26513007 contains a somewhat noticeable departure from a direct port of `cuda-shared`.  This is a proposed idea to deal with potential failures if the minimum required threads for the basis kernels is more than 256 (before making this change, `hip-shared` would fail for a 3D diffusion problem with P1d=8/Q1d = 10, for example). It could also be used to do further performance optimizations in terms of optimal thread configurations.

I also made some minor changes to flags related to the HIP backends, and changed the default HIP backend (with `/gpu/hip`) to be `hip-gen`.

~~I'm really not sure what's causing the GitLab failures on Noether.  I just built this branch manually on Noether and ran all the tests, and everything passed...~~ The failures have been fixed -- MAGMA needed updates with ROCm 3.10. (Thanks @jedbrown)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,11 +91,11 @@ OMP_SIMD_FLAG.clang     := $(OMP_SIMD_FLAG.gcc)
 OMP_SIMD_FLAG.icc       := -qopenmp-simd
 OPT.gcc                 := -ffp-contract=fast
 OPT.clang               := $(OPT.gcc)
-CFLAGS.gcc              := -fPIC -std=c99 -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -MMD -MP
+CFLAGS.gcc              := -fPIC -std=c99 -Wall -Wextra -Wno-unused-parameter -MMD -MP
 CFLAGS.clang            := $(CFLAGS.gcc)
 CFLAGS.icc              := $(CFLAGS.gcc)
 CFLAGS.XL               := -qpic -MMD
-CXXFLAGS.gcc            := -fPIC -std=c++11 -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -MMD -MP
+CXXFLAGS.gcc            := -fPIC -std=c++11 -Wall -Wextra -Wno-unused-parameter -MMD -MP
 CXXFLAGS.clang          := $(CXXFLAGS.gcc)
 CXXFLAGS.icc            := $(CXXFLAGS.gcc)
 CXXFLAGS.XL             := -qpic -std=c++11 -MMD
--- a/Makefile
+++ b/Makefile
@@ -92,11 +92,11 @@ OMP_SIMD_FLAG.icc       := -qopenmp-simd
 OPT.gcc                 := -ffp-contract=fast
 OPT.clang               := $(OPT.gcc)
 CFLAGS.gcc              := -fPIC -std=c99 -Wall -Wextra -Wno-unused-parameter -MMD -MP
-CFLAGS.clang            := $(CFLAGS.gcc)
+CFLAGS.clang            := $(CFLAGS.gcc) -Wno-unused-function
 CFLAGS.icc              := $(CFLAGS.gcc)
 CFLAGS.XL               := -qpic -MMD
 CXXFLAGS.gcc            := -fPIC -std=c++11 -Wall -Wextra -Wno-unused-parameter -MMD -MP
-CXXFLAGS.clang          := $(CXXFLAGS.gcc)
+CXXFLAGS.clang          := $(CXXFLAGS.gcc) -Wno-unused-function
 CXXFLAGS.icc            := $(CXXFLAGS.gcc)
 CXXFLAGS.XL             := -qpic -std=c++11 -MMD
 FFLAGS.GNU              := -fPIC -cpp -Wall -Wextra -Wno-unused-parameter -Wno-unused-dummy-argument -MMD -MP
@@ -217,6 +217,10 @@ magma.hip      := $(sort $(wildcard backends/magma/kernels/hip/*.hip.cpp))
 hip.c          := $(sort $(wildcard backends/hip/*.c))
 hip.cpp        := $(sort $(wildcard backends/hip/*.cpp))
 hip.hip        := $(sort $(wildcard backends/hip/kernels/*.hip.cpp))
+hip-shared.c   := $(sort $(wildcard backends/hip-shared/*.c))
+hip-shared.hip := $(sort $(wildcard backends/hip-shared/kernels/*.hip.cpp))
+hip-gen.c      := $(sort $(wildcard backends/hip-gen/*.c))
+hip-gen.cpp    := $(sort $(wildcard backends/hip-gen/*.cpp))
 
 # Quiet, color output
 quiet ?= $($(1))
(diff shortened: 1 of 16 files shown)


backends/hip-shared/kernels/hip-shared-basis.hip.cpp
```diff
@@ -0,0 +1,49 @@
...
+  hipMemcpyToSymbol(HIP_SYMBOL(c_B), d_B, Bsize, 0, hipMemcpyDeviceToDevice);
+  hipGetSymbolAddress((void **)c_B_ptr, HIP_SYMBOL(c_B));
+
+  return 0;
+}
+
+//------------------------------------------------------------------------------
+// Grad device initalization
+//------------------------------------------------------------------------------
+extern "C" int CeedHipInitInterpGrad(CeedScalar *d_B, CeedScalar *d_G,
```
jedbrown, line 39: `CEED_INTERN` should be sufficient here, though if you declare these in a local header, you wouldn't need to spell it out here and the compiler will check that they are compatible. I think this would be better than the manual declaration you currently have in `ceed-hip-shared-basis.c`.

Makefile
```diff
@@ -91,11 +91,11 @@ OMP_SIMD_FLAG.clang     := $(OMP_SIMD_FLAG.gcc)
 OMP_SIMD_FLAG.icc       := -qopenmp-simd
 OPT.gcc                 := -ffp-contract=fast
 OPT.clang               := $(OPT.gcc)
-CFLAGS.gcc              := -fPIC -std=c99 -Wall -Wextra -Wno-unused-parameter -MMD -MP
+CFLAGS.gcc              := -fPIC -std=c99 -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -MMD -MP
 CFLAGS.clang            := $(CFLAGS.gcc)
 CFLAGS.icc              := $(CFLAGS.gcc)
 CFLAGS.XL               := -qpic -MMD
-CXXFLAGS.gcc            := -fPIC -std=c++11 -Wall -Wextra -Wno-unused-parameter -MMD -MP
+CXXFLAGS.gcc            := -fPIC -std=c++11 -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -MMD -MP
```
jedbrown, line 98: I'd rather drop this and be more precise about where to ignore this warning.
```c
#define CEED_UNUSED __attribute__((unused))
```

jedbrown (approved):
Looks great in general. A couple points that I think will improve maintainability. A trivial rebase onto `main` should get clean pipelines.

Makefile
nbeams: Any preference on where the definition of `CEED_UNUSED` should go? 
Maybe `ceed-backend.h`, with `CEED_INTERN` (also uses `__attribute__`)?

jedbrown: That sounds good.

nbeams: Done in a52c4238f28ab92146d7f2c90434fd06ae9b8454

backends/hip-shared/kernels/hip-shared-basis.hip.cpp
nbeams: I moved these functions to the `ceed-hip-shared.h` header, and marked with `CEED_INTERN` (see 526015bc301544c98d597d81cf7f1f3611bf5ead).
This seemed in line with what other backends do (as we currently do in the MAGMA backend, for example).
I can also make a separate header just for these two functions...if we do that, we might want to do the same for the other backends, though (at least the CUDA backends, to mirror the HIP backends, since they are so closely related)...

jedbrown: Ah, cool. It's not necessary to repeat the `extern "C"` part here, so long as it's visible in the header. Looks like there are probably a lot of places where `extern "C"` can be replaced by `CEED_INTERN`, though it needn't be done in this PR. I don't mind it going in the common header at this point.

nbeams: Oh yeah, whoops.  I'll remove the extra `extern "C"`s

nbeams: You're right that we have a lot of `extern "C"`s floating around various backends where we should use `CEED_INTERN`, but for now, I at least changed them in these two new backends - the only `extern "C"`s left should be in the code strings that get fed to `hiprtc`.

jedbrown: Thanks!

jedbrown (approved):
Looks great. Could you just write a release note in `doc/sphinx/source/releasenotes.rst` since this is a major new feature?

I'm okay if this merges before #674, in which case I'll rebase it and apply the registration idiom. Could go the other way too, but I think that's slightly more work for you.

nbeams:
I added the backends to the release notes in 602cc54da33f8971b275df2b9b5f2ae2793f4bc5 -- did I need more information there?  Or should I move it to a different spot?

jedbrown:
Oops, I'm just blind.

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -0,0 +1,1382 @@
...
+  string devFunctions(deviceFunctions);
+
+  // Add atomicAdd function for old NVidia architectures
+  struct hipDeviceProp_t prop;
+  Ceed_Hip *ceed_data;
+  ierr = CeedGetData(ceed, &ceed_data); CeedChk(ierr);
+  ierr = hipGetDeviceProperties(&prop, ceed_data->deviceId); CeedChk(ierr);
+  if (prop.major<6){
+    code << atomicAdd;
+  }
```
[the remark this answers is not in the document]
YohannDudouit: (As far as I know there isn't hardware atomics...)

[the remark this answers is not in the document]
nbeams: You're right, I meant to delete that.  It's not getting used at the moment.  I'll open another PR.

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -0,0 +1,1382 @@
...
+                __double_as_longlong(val +
+                                     __longlong_as_double(assumed)));
+    // Note: uses integer comparison to avoid hang in case of NaN
+    // (since NaN != NaN)
+  } while (assumed != old);
+  return __longlong_as_double(old);
+}
+);
+
+static const char *deviceFunctions = QUOTE(
```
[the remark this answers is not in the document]
nbeams: I do have some ideas I was playing around with for this, but in the interest of going ahead and making the backends available to potential HIP users, I decided to push it off to a future PR.

PR #671, opened 2020-12-15
CI: move GitHub Actions to ubuntu 20.04

jedbrown:
This avoids warnings when using `ubuntu-latest`.

PR #672, opened 2020-12-15
Update configuration for sphinxcontrib-bibtex 2.0

nbeams:
sphinxcontrib-bibtex 2.0 had some changes that are backward incompatible: https://sphinxcontrib-bibtex.readthedocs.io/en/latest/changes.html

This PR is for updating the Sphinx configurations.

### DIFF
--- a/doc/sphinx/source/conf.py
+++ b/doc/sphinx/source/conf.py
@@ -57,6 +57,8 @@ extensions = [
     'sphinxcontrib.rsvgconverter',
 ]
 
+bibtex_bibfiles = ['references.bib']
+
 # The following, if true, allows figures, tables and code-blocks to be
 # automatically numbered if they have a caption.
 numfig = True


jedbrown:
Oh, thanks for this. I just put it into #671 with some comments and specifying the newer version in requirements.txt.

nbeams:
Oh never mind then, should we close this without merging?

jedbrown:
Yeah, I'm still working on the need to run Sphinx twice to actually process the references. (This seems to be a major usability regression in this extension.)

jedbrown:
Thanks, I've reverted to pre-2.0 sphinxcontrib-bibtex in 094d5ef252637cd3f4f34376e27d91a7a93ca84f, until the multiple runs of sphinx is resolved or has a good workaround. I'll close this.

PR #674, opened 2020-12-16
static link

jedbrown:
Add ability to link statically (requires refactoring registration pattern) and update the Rust crate to be installable and not need `LD_LIBRARY_PATH` hacks.

### DIFF
--- a/rust/build.rs
+++ b/rust/build.rs
@@ -1,16 +1,45 @@
 extern crate bindgen;
+extern crate pkg_config;
 
-use std::env;
 use std::path::PathBuf;
+use std::process::Command;
 
 fn main() {
-    // Tell cargo to tell rustc to link the system bzip2
-    // shared library.
-    println!("cargo:rustc-link-search=native={}", "../lib");
-    println!("cargo:rustc-link-lib=ceed");
+    let out_dir = PathBuf::from(env("OUT_DIR").unwrap());
+    let statik = env("CARGO_FEATURE_STATIC").is_some();
+    let system = env("CARGO_FEATURE_SYSTEM").is_some();
+
+    let ceed_pc = if system {
+        "ceed".to_string()
+    } else {
+        // Install libceed.a or libceed.so to $OUT_DIR/lib
+        let makeflags = env("CARGO_MAKEFLAGS").unwrap();
+        let mut make = Command::new("make");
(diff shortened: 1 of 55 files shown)


rust/build.rs
```diff
@@ -1,16 +1,41 @@
...
 fn main() {
-    // Tell cargo to tell rustc to link the system bzip2
-    // shared library.
-    println!("cargo:rustc-link-search=native={}", "../lib");
-    println!("cargo:rustc-link-lib=ceed");
+    let out_dir = PathBuf::from(env("OUT_DIR").unwrap());
+    let statik = env("CARGO_FEATURE_STATIC").is_some();
+
+    let mut pkgconf = pkg_config::Config::new();
+    if env("CARGO_FEATURE_SYSTEM").is_none() {
```
jeremylt, line 12: I think somewhere in here the GitHub action is reading an extra newline character

jedbrown: I don't know what exactly you were commenting on, but the action is clean now and I think cleaner.

jedbrown:
All checks pass so I think this is ready for review. We've gotten rid of the `LD_LIBRARY_PATH` hack for Rust testing and now use weak symbols instead of constructors to find the backend and gallery implementations.

I think the makefile changes are sane and understandable -- I had to use `-lstdc++` directly for the pkg-config file when linking statically (a .pc can't specify that `$(CXX)` be used for linking).

jeremylt:
@jedbrown, I get a few of these warnings:
```
/home/jeremy/Dev/libCEED/backends/cuda/ceed-cuda.c: In function ‘CeedRegister_Cuda’:
/home/jeremy/Dev/libCEED/backends/cuda/ceed-cuda.c:240:1: warning: control reaches end of non-void function [-Wreturn-type]
  240 | }
```

jedbrown:
Oops; fixed now.

jeremylt (approved):
Works locally for me, looks good

Maybe at some group meeting it would be interesting to have a discussion about how library packaging and linking works, with some of the cross-language packaging we've done

jedbrown:
Today? Great way to celebrate the end of the semester!

valeriabarra:
Too bad I missed today's discussion! It would have been useful to me too

PR #676, opened 2020-12-18
Remove unused CUDA atomicAdd code from hip-gen

nbeams:
@YohannDudouit pointed out that this CUDA-specific code was incorrectly brought over from the cuda-gen backend during the port.  

(It wasn't getting used, anyway, so this shouldn't affect anything in the operation of the current hip-gen backend.)


jedbrown (approved):

README.rst
```diff
@@ -186,7 +186,7 @@ There are multiple supported backends, which can be selected at runtime in the e
 +----------------------------+---------------------------------------------------+-----------------------+
 | ``/gpu/cuda/gen``          | Optimized pure CUDA kernels using code generation | No                    |
 +----------------------------+---------------------------------------------------+-----------------------+
-| HIP Native Backend                                                                                     |
+| HIP Native Backends                                                                                    |
```
jedbrown, line 189: Some projects have a policy to always use plural for headings that may become lists. It reads fine when there's only one and doesn't require someone to notice it later.

PR #679, opened 2020-12-23
examples/petsc: use VecGetArrayAndMemType() to support CUDA/HIP/Kokkos

### LINKED ISSUE #219: CI on GPUs

### LINKED ISSUE #660: PETSc Code Coverage
The PETSc examples cover important cases in our code coverage. We should add PETSc testing to Noether as soon PETSc + HIP is ready and/or add CodeCov to the Azure pipeline testing.

jedbrown:
We resolve default VECCUDA or VECHIP based on /gpu/cuda or /gpu/hip. For
others, one should pass -dm_vec_type kokkos or -dm_vec_type standard.

There is no longer a "libCEED User Requested MemType" field in the
output.

Enable CI testing using PETSc built locally with HIP. This should be
converted to building a specified version once some bugs are fixed and
the branch merges.

Exclude automatic selection of VECHIP for /gpu/hip/occa due to SEGV:
https://github.com/CEED/libCEED/issues/678

* Resolve #219 (a pretty complete test suite on GPUs)
* Resolve #660 (PETSc tests on Noether)

### DIFF
--- a/backends/hip/ceed-hip.c
+++ b/backends/hip/ceed-hip.c
@@ -33,11 +33,11 @@ int CeedHipInit(Ceed ceed, const char *resource, int nrc) {
   int ierr;
   const int rlen = strlen(resource);
   const bool slash = (rlen>nrc) ? (resource[nrc] == '/') : false;
-  const int deviceID = (slash && rlen > nrc + 1) ? atoi(&resource[nrc + 1]) : 0;
+  const int deviceID = (slash && rlen > nrc + 1) ? atoi(&resource[nrc + 1]) : -1;
 
   int currentDeviceID;
   ierr = hipGetDevice(&currentDeviceID); CeedChk_Hip(ceed,ierr);
-  if (currentDeviceID!=deviceID) {
+  if (deviceID >= 0 && currentDeviceID != deviceID) {
     ierr = hipSetDevice(deviceID); CeedChk_Hip(ceed,ierr);
   }
 
--- a/examples/petsc/multigrid.c
+++ b/examples/petsc/multigrid.c
@@ -58,11 +58,11 @@ int main(int argc, char **argv) {
   PC pc;
   Mat *matO, *matPR, matcoarse;
   Vec *X, *Xloc, *mult, rhs, rhsloc;
+  PetscMemType memtype;
   UserO *userO;
   UserProlongRestr *userPR;
   Ceed ceed;
   CeedData *ceeddata;
-  CeedMemType memtyperequested;
   CeedVector rhsceed, target;
   CeedQFunction qferror, qfrestrict, qfprolong;
   CeedOperator operror;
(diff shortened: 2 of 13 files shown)


jedbrown:
Pipeline is clean now so this is ready for review.

backends/hip/ceed-hip.c
```diff
@@ -33,11 +33,11 @@ int CeedHipInit(Ceed ceed, const char *resource, int nrc) {
...
   const bool slash = (rlen>nrc) ? (resource[nrc] == '/') : false;
-  const int deviceID = (slash && rlen > nrc + 1) ? atoi(&resource[nrc + 1]) : 0;
+  const int deviceID = (slash && rlen > nrc + 1) ? atoi(&resource[nrc + 1]) : -1;
 
   int currentDeviceID;
   ierr = hipGetDevice(&currentDeviceID); CeedChk_Hip(ceed,ierr);
-  if (currentDeviceID!=deviceID) {
+  if (deviceID >= 0 && currentDeviceID != deviceID) {
     ierr = hipSetDevice(deviceID); CeedChk_Hip(ceed,ierr);
   }
```
jedbrown, line 42: Does this logic make sense to you, @nbeams @YohannDudouit ?

I think we could alternatively defer it and provide an interface like `CeedSetDevice(Ceed c, int id)` so that all operations could pin themselves back to that device and restore the calling state when done, though I'm unsure that would be useful.

examples/solids/elasticity.c
```diff
@@ -140,14 +134,26 @@ int main(int argc, char **argv) {
...
   // -- Create distributed DM from mesh file
   ierr = CreateDistributedDM(comm, appCtx, &dmOrig); CHKERRQ(ierr);
+  VecType vectype;
+  switch (memTypeBackend) {
+  case CEED_MEM_HOST: vectype = VECSTANDARD; break;
+  case CEED_MEM_DEVICE: {
+    const char *resolved;
+    CeedGetResource(ceed, &resolved);
+    if (strstr(resolved, "/gpu/cuda")) vectype = VECCUDA;
+    else if (strstr(resolved, "/gpu/hip")) vectype = VECHIP;
```
jeremylt, line 144: Do we need to add the OCCA bypass here as well?

examples/petsc/multigrid.c
```diff
@@ -173,6 +149,20 @@ int main(int argc, char **argv) {
...
   }
 
+  VecType vectype;
+  switch (memtypebackend) {
+  case CEED_MEM_HOST: vectype = VECSTANDARD; break;
+  case CEED_MEM_DEVICE: {
+    const char *resolved;
+    CeedGetResource(ceed, &resolved);
+    if (strstr(resolved, "/gpu/cuda")) vectype = VECCUDA;
+    else if (strstr(resolved, "/gpu/hip")) vectype = VECHIP;
```
jeremylt, line 159: Same

examples/solids/elasticity.c
jedbrown: It's currently screened out in `tap.sh` and `junit.py`. I presume that #678 can be fixed before OCCA adds support for VLA (and thus is usable for our solids and fluids examples).

jeremylt: That's fair. 

One of these days we need to refactor all of these helper functions into a single place so that we don't have to keep track of these sorts of differences between examples.

examples/petsc/multigrid.c
jedbrown: Hmm, this is skipped in `junit.py`, but was removed from `tap.sh` in #305. I'll enable it in `junit.py` and mask it here.

examples/solids/elasticity.c
jedbrown: Yes, I actually want that to go into PETSc, which has enough context (when used in a mode that would call libCEED) to make such decisions.

examples/petsc/multigrid.c
jedbrown: Done.

jeremylt (approved):
I haven't been able to update ROCm and get it working on my machine yet (ROCm updates are always unpleasant for me for some reason), so I haven't built your PETSc branch sucessfully yet, but with the pipeline running clean, I don't see my local issues as a reason not to merge.

PR #680, opened 2020-12-28
doc: update to sphinxcontrib-bibtex 2.1

jedbrown:
Now the source files are declared in bibtex_bibfiles in conf.py and the
`.. bibliography` directive does not need the path.

Thanks to https://github.com/mcmtroffaes/sphinxcontrib-bibtex/pull/219


jeremylt (approved):
works locally for me

PR #683, opened 2020-12-29
Rust - add examples

### LINKED ISSUE #667: Rust code coverage
The Arrow project has some experienced Rust developers and some C FFI interop (though Rust Arrow doesn't call C++ Arrow). This is their new workflow, using Tarpaulin, which is the Rust coverage tool they chose.
https://github.com/apache/arrow/blob/master/.github/workflows/rust.yml#L268-L306

jeremylt:
All of the below issues are fixed

Note: this PR closes #667

---

Right now this is a straightforward transcription of C ex1 into Rust.

Issue 1:
```
User CeedQFunction 
  2 Input Fields:
    Input Field [0]:
      Name: "dx"
      Size: 9
      EvalMode: "gradient"
    Input Field [1]:
      Name: "weights"
      Size: 1
      EvalMode: "(null)"
  1 Output Field:
    Output Field [0]:
      Name: "qdata"
      Size: 1
      EvalMode: "none"
```
`EvalMode::Weight` is not being passed to libCEED correctly.

Issue 2:
```
error[E0597]: `dim` does not live long enough
   --> src/main.rs:92:19
    |
86  |         |[jacobian, weights, ..]: QFunctionInputs, [qdata, ..]: QFunctionOutputs| -> i32 {
    |         -------------------------------------------------------------------------------- value captured here
...
92  |             match dim {
    |                   ^^^ borrowed value does not live long enough
...
131 |         let mut qfunction = ceed.q_function_interior(1, Box::new(build_mass));
    |                                                         -------------------- cast requires that `dim` is borrowed for `'static`
...
251 | }
    | - `dim` dropped here while still borrowed

error: aborting due to previous error

For more information about this error, try `rustc --explain E0597`.
error: could not compile `ex1-volume`
```
Boxing the closure for the QFunction is leading to a strange borrow issue with a non-static env variable.... which defeats the whole purpose a boxed closure.


examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+    // Apply a transformation to the mesh coordinates
+    let exact_volume = transform_mesh_coordinates(dim, mesh_size, &mut mesh_coords);
+
+    // QFunction that builds the quadrature data for the mass operator
+    // -- QFunction from user closure
+    let build_mass =
+        |[jacobian, weights, ..]: QFunctionInputs, [qdata, ..]: QFunctionOutputs| -> i32 {
+// ---------------------------------------------------------------------------------------
+// ISSUE: dim IS NOT LIVING LONG ENOUGH FOR BORROW CHECKER
+let dim = 1;
```
jeremylt, line 89: See above mentioned Issue 2

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+    let qf_build_closure = {
+        let mut qfunction = ceed.q_function_interior(1, Box::new(build_mass));
+        qfunction.add_input("dx", (ncomp_x * dim) as i32, EvalMode::Grad);
+        qfunction.add_input("weights", 1, EvalMode::Weight);
+        qfunction.add_output("qdata", 1, EvalMode::None);
+        qfunction
+    };
+// ---------------------------------------------------------------------------------------
+// ISSUE: EvalMode::Weight is not displaying correctly
+println!("{}", qf_build_closure);
```
jeremylt, line 139: See above mentioned Issue 1

examples/rust/ex1-volume/src/cli.yml
```diff
@@ -0,0 +1,52 @@
...
+version: "1.0"
+author: Jeremy L Thompson <thompson.jeremy.luke@gmail.com>
+about: This example uses the mass matrix to compute the length, area, or volume of a region, depending upon runtime parameters.
+args:
+    - ceed:
+        short: -c
+        long: --ceed
+        help: Set libCEED backend resource to use
+        takes_value: true
+        default_value: /cpu/self
```
jedbrown, line 11: Do you prefer this over [structopt](https://docs.rs/structopt/0.3.21/structopt/)? It lets the compiler check that you've spelled and typed your options consistently and saves the unwrap.

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+    // Create a Vector with the mesh coordinates
+    let mut mesh_coords = set_cartesian_mesh_coords(&ceed, dim, num_xyz, mesh_order, mesh_size);
+
+    // Apply a transformation to the mesh coordinates
+    let exact_volume = transform_mesh_coordinates(dim, mesh_size, &mut mesh_coords);
+
+    // QFunction that builds the quadrature data for the mass operator
+    // -- QFunction from user closure
+    let build_mass =
+        |[jacobian, weights, ..]: QFunctionInputs, [qdata, ..]: QFunctionOutputs| -> i32 {
```
jedbrown, line 86:
```suggestion
        move |[jacobian, weights, ..]: QFunctionInputs, [qdata, ..]: QFunctionOutputs| -> i32 {
```
This is enough because `dim` is `Copy`. https://doc.rust-lang.org/std/keyword.move.html

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+        "qdata",
+        &restr_qdata,
+        BasisOpt::Collocated,
+        VectorOpt::Active,
+    );
+
+    // Compute the quadrature data for the mass operator
+    let elem_qpts = num_qpts.pow(dim as u32);
+    let mut num_elem = 1;
+    (0..dim).for_each(|d| num_elem *= num_xyz[d]);
```
jedbrown, line 172: More typical to write
```rust
for d in 0..dim {num_elem *= num_xyz[d]}
```
but here I would use
```suggestion
    let num_elem: usize = num_xyz.iter().product();
```

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+    let mut qdata = ceed.vector(num_elem * elem_qpts);
+    if !test {
+        print!("Computing the quadrature data for the mass operator ...");
+        std::io::stdout().flush().unwrap();
```
jedbrown, line 176: I'd expect this doesn't need to be flushed if you use `println!` above (because output to a terminal is line-buffered by default).

examples/rust/ex1-volume/src/cli.yml
jeremylt: I have no strong feelings either way. I picked clap only because it seems a bit more widely used. Having the command line options in a separate yaml seemed like a nice way to keep main tidy.

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+
+    // QFunction that applies the mass operator
+    // -- QFunction from user closure
+    let apply_mass = |[u, qdata, ..]: QFunctionInputs, [v, ..]: QFunctionOutputs| -> i32 {
+        // Apply mass operator
+        v.iter_mut()
+            .zip(u.iter().zip(qdata.iter()))
+            .for_each(|(v, (u, w))| *v = u * w);
+        // Return clean error code
+        0
```
jedbrown, line 191: We should eventually make these return `Result<(), ceed::Error>` where `ceed::Error` can be converted to `i32`.

examples/rust/ex1-volume/src/main.rs
jedbrown: You can delete this now.

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,429 @@
...
+        ElemRestrictionOpt::None,
+        &basis_mesh,
+        VectorOpt::None,
+    );
+    op_build.set_field(
+        "qdata",
+        &restr_qdata,
+        BasisOpt::Collocated,
+        VectorOpt::Active,
+    );
```
jedbrown, line 167: We should make these builders so this non-conditional configuration can be chained.
https://rust-lang.github.io/api-guidelines/type-safety.html#builders-enable-construction-of-complex-values-c-builder

I suppose the examples also suggest this should be `op.field(...)` because `get_` and `set_` prefixes are discouraged in most cases. https://rust-lang.github.io/api-guidelines/naming.html?highlight=get_#getter-names-follow-rust-convention-c-getter

examples/rust/ex1-volume/src/cli.yml
jedbrown: Ah, `structopt` uses `clap`, but the result of parsing is a struct instead of an object you query by string. No big deal right now.

examples/rust/ex1-volume/src/main.rs
jeremylt: Really, I'm not the biggest fan of this `"Message to the user..."` *basically instant to a human operation* `"done"` thing, so I dropped it

examples/rust/ex1-volume/src/main.rs
jeremylt: yeah, I was noticing that the `get_*` syntax isn't very idiomatic. I can do a pass to remove those in the morning in a separate PR

jedbrown:
I pushed a top-level workspace, so we can avoid repetitive building and can run all the tests in one shot. Feel free to drop or modify.

examples/rust/ex1-volume/src/cli.yml
jeremylt: refactoring to use structopt made testing easier

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,584 @@
+// Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
```
jedbrown, line 1: If we're using these copyright headers, we may as well get the year right on new code.

Note that they are not necessary from a licensing or IP standpoint and I'd have no problem skipping them entirely in source files.

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,584 @@
...
+
+// ----------------------------------------------------------------------------
+// Example 1
+// ----------------------------------------------------------------------------
+fn main() {
+    let opt = Opt::from_args();
+    let _ = example_1(opt);
+}
+
+fn example_1(opt: Opt) -> Result<(), &'static str> {
```
jedbrown, line 76: `&'static str` is limiting because you can't format output based on run-time state. I think `String` would make more sense here, or define an actual error type. Also note that `main` can return `Result` so you don't have to drop it `let _ = ` above.

jeremylt:
I was getting this failure locally, and it seems like it is showing in CI now too:
```
/home/runner/work/libCEED/libCEED/interface/ceed.c:632 in CeedInit(): No suitable backend: /cpu/self/ref/serial
```

jeremylt:
I'll update for the changes in #686 before creating ex2

jeremylt:
huh, 
```
--no-default-features
```
does not help, but
```
--test-threads=1
```
does

jeremylt:
Could we use https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html
Not sure what the equivalent for icc is.

jedbrown:
Oh, lovely.
```
running 4 tests
/home/jed/src/libceed/interface/ceed.c:632 in CeedInit(): No suitable backend: /cpu/self/ref/serial
/home/jed/src/libceed/interface/ceed-vector.c:583 in CeedVectorDestroy(): Cannot destroy CeedVector, a process has read access
```

jeremylt:
Yep, that's the one I've been seeing. The vector error seems to pop up as a spurious error sometimes when we have other issues

jedbrown:
I fixed that race, but there's still a panic.
<details>

```
$ gdb -ex r --args /home/jed/src/libceed/target/debug/deps/ex1_volume-4c3ec90d956f16b0 --test-threads=1
No symbol table is loaded.  Use the "file" command.
Breakpoint 1 (PetscError) pending.
No symbol table is loaded.  Use the "file" command.
Breakpoint 2 (rust_panic) pending.
Reading symbols from /home/jed/src/libceed/target/debug/deps/ex1_volume-4c3ec90d956f16b0...
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /home/jed/src/libceed/target/debug/deps/ex1_volume-4c3ec90d956f16b0.
Use `info auto-load python-scripts [REGEXP]' to list them.
Starting program: /home/jed/src/libceed/target/debug/deps/ex1_volume-4c3ec90d956f16b0 --test-threads=1
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

running 4 tests
test tests::example_1_1d ...
Breakpoint 2, std::panicking::rust_panic () at /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/std/src/panicking.rs:640
640     /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/std/src/panicking.rs: No such file or directory.
(gdb) bt
#0  std::panicking::rust_panic () at /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/std/src/panicking.rs:640
#1  0x00005555556785a6 in std::panicking::rust_panic_with_hook () at /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/std/src/panicking.rs:610
#2  0x0000555555678047 in std::panicking::begin_panic_handler::{{closure}} () at /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/std/src/panicking.rs:495
#3  0x0000555555675fbc in std::sys_common::backtrace::__rust_end_short_backtrace<closure-0,!> () at /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/std/src/sys_common/backtrace.rs:141
#4  0x0000555555677fd9 in std::panicking::begin_panic_handler () at /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/std/src/panicking.rs:493
#5  0x00005555556945b1 in core::panicking::panic_fmt () at /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/core/src/panicking.rs:92
#6  0x00005555556944fd in core::panicking::panic () at /rustc/158f8d034b15e65ba8dc0d066358dd0632bfcd6e//library/core/src/panicking.rs:50
#7  0x00005555556259c7 in libceed::qfunction::trampoline::{{closure}} () at /home/jed/src/libceed/rust/src/qfunction.rs:224
#8  0x0000555555625544 in core::iter::adapters::map::{{impl}}::__iterator_get_unchecked<&[f64],core::iter::adapters::enumerate::Enumerate<core::slice::iter::Iter<*const f64>>,closure-0> (self=0x7ffffff
f97e0, idx=0) at /home/jed/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/adapters/map.rs:129
#9  0x0000555555626338 in core::iter::adapters::zip::{{impl}}::next<core::iter::adapters::map::Map<core::iter::adapters::enumerate::Enumerate<core::slice::iter::Iter<*const f64>>, closure-0>,core::slic
e::iter::IterMut<&[f64]>> (self=0x7fffffff97e0) at /home/jed/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/adapters/zip.rs:198
#10 0x0000555555626047 in core::iter::adapters::zip::{{impl}}::next<core::iter::adapters::map::Map<core::iter::adapters::enumerate::Enumerate<core::slice::iter::Iter<*const f64>>, closure-0>,core::slic
e::iter::IterMut<&[f64]>> (self=0x7fffffff97e0) at /home/jed/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/adapters/zip.rs:45
#11 0x000055555562675d in core::iter::traits::iterator::Iterator::fold<core::iter::adapters::zip::Zip<core::iter::adapters::map::Map<core::iter::adapters::enumerate::Enumerate<core::slice::iter::Iter<*
const f64>>, closure-0>, core::slice::iter::IterMut<&[f64]>>,(),closure-0> (self=..., init=(), f=...) at /home/jed/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/
src/iter/traits/iterator.rs:2022
#12 0x0000555555626893 in core::iter::traits::iterator::Iterator::for_each<core::iter::adapters::zip::Zip<core::iter::adapters::map::Map<core::iter::adapters::enumerate::Enumerate<core::slice::iter::It
er<*const f64>>, closure-0>, core::slice::iter::IterMut<&[f64]>>,closure-1> (self=..., f=...) at /home/jed/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter
/traits/iterator.rs:678
#13 0x00005555555c6b06 in libceed::qfunction::trampoline (ctx=0x7fffffffa150, q=6, inputs=0x555555ace9b0, outputs=0x555555aceb90) at /home/jed/src/libceed/rust/src/qfunction.rs:220
#14 0x000055555562c148 in CeedQFunctionApply_Ref (qf=<optimized out>, Q=6, U=0x555555ada8b0, V=0x555555ada940) at /home/jed/src/libceed/backends/ref/ceed-ref-qfunction.c:51
#15 0x000055555562b8cb in CeedOperatorApplyAdd_Ref (op=0x555555acad60, invec=<optimized out>, outvec=0x555555ada1a0, request=0x5555558857c8 <ceed_request_immediate>) at /home/jed/src/libceed/backends/r
ef/ceed-ref-operator.c:446
#16 0x00005555555c27fb in libceed::operator::OperatorCore::apply (self=0x7fffffffad60, input=0x7fffffffab28, output=0x7fffffffade8) at /home/jed/src/libceed/rust/src/operator.rs:156
#17 0x00005555555c2817 in libceed::operator::Operator::apply (self=0x7fffffffad60, input=0x7fffffffab28, output=0x7fffffffade8) at /home/jed/src/libceed/rust/src/operator.rs:318
#18 0x0000555555578ff0 in ex1_volume::example_1 (options=...) at /home/jed/src/libceed/examples/rust/ex1-volume/src/main.rs:211
#19 0x000055555557353c in ex1_volume::tests::example_1_1d () at /home/jed/src/libceed/examples/rust/ex1-volume/src/main.rs:313
#20 0x0000555555573aea in ex1_volume::tests::example_1_1d::{{closure}} () at /home/jed/src/libceed/examples/rust/ex1-volume/src/main.rs:294
```
</details>

jeremylt:
I was having a weird issue with a multiply overflow in that part of the trampoline function yesterday. I ended up renistalling rustup and the error went away

jedbrown:
Huh, `cargo +stable test` passes, so maybe this is just a nightly issue.

jedbrown:
Do you know why this always gets repeated despite no code changes?
```
$ cargo +stable test
   Compiling libceed v0.7.0 (/home/jeka2967/libceed/rust)
   Compiling ex1-volume v0.1.0 (/home/jeka2967/libceed/examples/rust/ex1-volume)
   Compiling ex2-surface v0.1.0 (/home/jeka2967/libceed/examples/rust/ex2-surface)
    Finished test [unoptimized + debuginfo] target(s) in 3.22s
```

I'm concerned about the failures with nightly (1.51), in that we may be doing something unsafe that just happens to not break with the optimizations in stable (1.49).

jeremylt:
> Do you know why this always gets repeated despite no code changes?

I'm not sure, but I saw it on my small Advent of Code crate as well.

This is the part that seems to be having the panic, but it looks ok to me
```
    // Inputs
    let inputs_slice: &[*const bind_ceed::CeedScalar] =
        std::slice::from_raw_parts(inputs, MAX_QFUNCTION_FIELDS);
    let mut inputs_array: [&[f64]; MAX_QFUNCTION_FIELDS] = [&[0.0]; MAX_QFUNCTION_FIELDS];
    inputs_slice
        .iter()
        .enumerate()
        .map(|(i, &x)| {
            std::slice::from_raw_parts(x, (trampoline_data.input_sizes[i] * q) as usize) as &[f64]
        })
        .zip(inputs_array.iter_mut())
        .for_each(|(x, a)| *a = x);

    // Outputs
    let outputs_slice: &[*mut bind_ceed::CeedScalar] =
        std::slice::from_raw_parts(outputs, MAX_QFUNCTION_FIELDS);
    let mut outputs_array: [&mut [f64]; MAX_QFUNCTION_FIELDS] = mut_max_fields!(&mut [0.0]);
    outputs_slice
        .iter()
        .enumerate()
        .map(|(i, &x)| {
            std::slice::from_raw_parts_mut(x, (trampoline_data.output_sizes[i] * q) as usize)
                as &mut [f64]
        })
        .zip(outputs_array.iter_mut())
        .for_each(|(x, a)| *a = x);
```

jeremylt:
The examples now run and I mention them in the documentation.

jeremylt:
@jedbrown, I went ahead and started the field error checking in a separate branch. My plan is to defer the improved Rust error checks to a third PR after that one.

.github/workflows/rust-test-with-style.yml
```diff
@@ -19,14 +19,16 @@ jobs:
...
         FC: gfortran-9
-      run: |
-        make info
-        make -j2
-        cargo test
-    - name: Rust style
+    - name: Codecov upload
+      uses: codecov/codecov-action@v1.0.2
+      with:
+        token: ${{secrets.CODECOV_TOKEN}}
```
jeremylt, line 31: @jedbrown, when you get a chance, can you please put a `CODECOV_TOKEN` in the repo secrets?

rust/src/elem_restriction.rs
```diff
@@ -298,12 +298,12 @@ impl ElemRestriction {
...
     /// let c = r.comp_stride();
-    /// assert_eq!(c, compstride as i32, "Incorrect component stride");
+    /// assert_eq!(c, compstride, "Incorrect component stride");
     /// ```
-    pub fn comp_stride(&self) -> i32 {
+    pub fn comp_stride(&self) -> usize {
         let mut compstride = 0;
         unsafe { bind_ceed::CeedElemRestrictionGetCompStride(self.ptr, &mut compstride) };
-        compstride
+        compstride as usize
```
YohannDudouit, line 306: why don't you need to do `usize::try_from(compstride).unwrap()` here?

rust/src/qfunction.rs
```diff
@@ -392,14 +392,14 @@ impl QFunction {
...
-    pub fn input(mut self, fieldname: &str, size: i32, emode: crate::EvalMode) -> Self {
+    pub fn input(mut self, fieldname: &str, size: usize, emode: crate::EvalMode) -> Self {
         let name_c = CString::new(fieldname).expect("CString::new failed");
         let idx = self.trampoline_data.number_inputs;
         self.trampoline_data.input_sizes[idx] = size;
         self.trampoline_data.number_inputs += 1;
         let emode = emode as bind_ceed::CeedEvalMode;
         unsafe {
-            bind_ceed::CeedQFunctionAddInput(self.qf_core.ptr, name_c.as_ptr(), size, emode);
+            bind_ceed::CeedQFunctionAddInput(self.qf_core.ptr, name_c.as_ptr(), size as i32, emode);
```
YohannDudouit, line 402: You shouldn't cast `size` to `i32` inside an `unsafe` block.

jeremylt: why not?

rust/src/elem_restriction.rs
jeremylt: Yeah, the compiler should have caught that. That's a typo on my part

rust/src/basis.rs
```diff
@@ -234,7 +234,7 @@ impl Basis {
             tmode as bind_ceed::CeedTransposeMode,
             emode as bind_ceed::CeedEvalMode,
         );
-        unsafe { bind_ceed::CeedBasisApply(self.ptr, nelem, tmode, emode, u.ptr, v.ptr) };
+        unsafe { bind_ceed::CeedBasisApply(self.ptr, nelem as i32, tmode, emode, u.ptr, v.ptr) };
```
YohannDudouit, line 237: Don't cast `nelem` inside `unsafe` block.

rust/src/qfunction.rs
YohannDudouit: because casting can fail

jeremylt: Casting between a usize and an i32 shouldn't be able to fail

YohannDudouit: `usize` is "bigger" than `i32` no?

YohannDudouit: (the max is not the same)

jeremylt: The cast will truncate and will always be successful, but we can put in a check that the value is less than the max i32 I guess

rust/src/elem_restriction.rs
YohannDudouit: Why should the compiler have caught it?

jeremylt: Because 1) compstride needs to be an i32 for the CeedElemRestrictionGetCompStride call to actually be vaild and 2) a return value cannot use `as xxx`

rust/src/qfunction.rs
jedbrown: We want it to panic if it's out of range, do it unsafely and check manually. Presumably you can bind it outside.

We could consider updating the libCEED interface to use `size_t` in more places. The only place where there is a performance impact of size is indexing in `ElemRestriction`. Otherwise, we used signed ints mainly because they avoid possible bugs in reverse `for (i=n; i>=0; i--)` loops, though an equally valid point is that negative values should not exist in certain circumstances and the type system doesn't help us.

jeremylt: Integer overflow only panics in debugging mode, not in release, so we'd need to add the check manually if we want it to always occur (which would be my preference)

YohannDudouit: Isn't it what `try_from` does?

YohannDudouit: I don't think we can consider that it is a success that our value gets truncated.

jeremylt: I think we just agreed to check for overflow

rust/src/elem_restriction.rs
YohannDudouit: I think the compiler sees 1) through the type inference, I wasn't aware of 2). Thank you.

.github/workflows/rust-test-with-style.yml
jedbrown: Should work now.

rust/src/qfunction.rs
YohannDudouit: Sorry, I didn't update the page before sending my messages.

jeremylt: GitHub usually automatically refreshes for me, so I'm not sure why it isn't today, but I'm having to manually refresh as well

YohannDudouit: Again, isn't it what `try_from` does?

YohannDudouit: @jedbrown What's the benefit of doing it unsafely and checking manually?

YohannDudouit: Something like:
```rust
let size = i32::try_from(size)?;
unsafe { ... }
```

rust/src/basis.rs
```diff
@@ -106,7 +106,12 @@ impl Basis {
         qweight1d: &[f64],
     ) -> Self {
         let mut ptr = std::ptr::null_mut();
-        let (dim, ncomp, P1d, Q1d) = (dim as i32, ncomp as i32, P1d as i32, Q1d as i32);
+        let (dim, ncomp, P1d, Q1d) = (
+            i32::try_from(dim).unwrap(),
+            i32::try_from(ncomp).unwrap(),
+            i32::try_from(P1d).unwrap(),
+            i32::try_from(Q1d).unwrap(),
+        );
```
jeremylt, line 114: These unwraps will be eventually folded into our future error handling strategy w/Results

rust/src/elem_restriction.rs
```diff
@@ -301,7 +306,7 @@ impl ElemRestriction {
     /// assert_eq!(c, compstride, "Incorrect component stride");
     /// ```
     pub fn comp_stride(&self) -> usize {
-        let mut compstride = 0;
+        let mut compstride: i32 = 0;
```
jeremylt, line 309: I went ahead and explicitly declared these as i32 since the compiler didn't catch that previous issue - I wonder if the unsafe block is interfering with the type checking?

YohannDudouit: Why??

jeremylt:
Huh, I will have to investigate because it looks like the doctests are not counting in the Rust code coverage.

rust/src/elem_restriction.rs
YohannDudouit: Just let the compiler do its job

jeremylt: Because the compiler failed to do its job above, as you noted...

YohannDudouit: I wrote: "I think the compiler sees 1) through the type inference", meaning that the compiler does what he is supposed to do...

rust/src/basis.rs
YohannDudouit: Are we sure we want to use `unwrap` instead of `?`?

rust/src/qfunction.rs
jedbrown: `try_from` is good. I didn't see the context of whether we were in a `Result`. We want it checked (necessarily in debug mode, but this is sufficiently coarse granularity that checking always won't be perceptible in the performance profile.

rust/src/elem_restriction.rs
jeremylt: Right, but the `as usize` should have been a compilation failure because you can't typecast like that in a return, making me think the compiler is making some sort of issue there. It's a short chunk of code, so my suspicion is that the unsafe block is interfering with the compiler in some way.

rust/src/basis.rs
jeremylt: ? only works in functions that return Results

rust/src/elem_restriction.rs
YohannDudouit: Do you have a reference for this `as` cannot be used on return values? I can't find this anywhere.

jeremylt: My reference would be the compiler. I tried it at first and got an error.

rust/src/basis.rs
YohannDudouit: Okay, that makes sense, thank you.

rust/src/elem_restriction.rs
jeremylt: Huh, it is not erroring on a simple example on the Rust playground, so :shrug:

YohannDudouit: Well, if the reference is the compiler, the compiler didn't throw an error...
The way I understand it is `compstride` has type `i32` inferred from `bind_ceed::CeedElemRestrictionGetCompStride`, and then you cast an `i32` to a `usize`.
What should have gone wrong according to you?

jeremylt: The error said something roughly equivalent to 'can't use 'as' in a return statement'
I have zero clues as to where the error went and why it's not showing now

jeremylt: I'm not particularly interested in figuring out what happened, so I just removed all of the `: type` stuff. It's probably fine.

YohannDudouit: I'm trying to understand why you say that you suspect the compiler to do something wrong. Is it because you previously had an error than you didn't understand?

jeremylt:
CI should we working now. Coverage looks a little low considering almost every function has a doctest.

jeremylt:
Most of the missed lines are due to this bug: https://github.com/xd009642/tarpaulin/issues/496

tests/junit.py
```diff
@@ -127,7 +127,7 @@ def run(test, backends):
                 elif proc.stdout and test[:4] not in 't003':
                     case.add_failure_info('stdout', output=proc.stdout)
             testcases.append(case)
-        return TestSuite(test, testcases)
+    return TestSuite(test, testcases)
```
jeremylt, line 130: I think this will boost our CodeCov somewhat

jeremylt:
ToDo: Get the Rust Action to only upload coverage on the Rust files.

tests/junit.py
jedbrown: :eyes:

jeremylt:
Well, even when I dynamically link libCEED for the Rust bindings, tarpaulin still captures coverage info for the function calls inside of libceed.so.

jedbrown:
Did you build `libceed.so` without `--coverage`?

.github/workflows/rust-test-with-style.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         CC: ${{ matrix.compiler }}
         FC: gfortran-9
       with:
-        args: '--run-types Doctests Tests'
+        args: '--run-types Doctests Tests --exclude-files backends/* gallery/* include/* interface/*'
```
jeremylt, line 30: I found the knob to manually exclude. I wish there was a knob to manually include instead.

jeremylt:
@jedbrown, did something change on Noether or is the Gitlab failure spurious?

.github/workflows/rust-test-with-style.yml
jeremylt: According to CodeCov, this is working correctly. The only thing left with this PR should be whatever happened on Noether

jeremylt:
Noether is passing now, so it must have been spurious. This PR is good to go now.

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,434 @@
...
+            volume - exact_volume
+        );
+    }
+    let tolerance = if dim == 1 {
+        1E-14
+    } else if dim == 2 {
+        1E-7
+    } else {
+        1E-5
+    };
```
jedbrown, line 264:
```suggestion
    let tolerance = match dim {
        1 => 1e-14,
        2 => 1e-7,
        _ => 1e-5,
    };
```

examples/rust/ex1-volume/src/main.rs
```diff
@@ -0,0 +1,434 @@
...
+            ceed_spec,
+            dim,
+            mesh_degree,
+            solution_degree,
+            num_qpts,
+            problem_size_requested,
+            test,
+            quiet,
+            gallery,
+        };
```
jedbrown, line 304:
```suggestion
        let options = opt::Opt {
            ceed_spec: "/cpu/self/ref/serial".to_string(),
            dim: 1,
            mesh_degree: 4,
            solution_degree: 4,
            num_qpts: 6,
            problem_size_requested: -1,
            test: true,
            quiet: true,
            gallery: false,
        };
```

jedbrown (approved):
Should ex1 and ex2 share the utility that creates their mesh?

jeremylt:
I agree those are obvious (now that you point them out) improvements and I'll get them in the morning 

I thought about sharing the mesh code, but the mesh transformations are different between the two examples, so I wasn't sure if putting half of the mesh code in a common utility would make the examples more confusing for new users or not

jedbrown:
I think it's very reasonable to have the mesh transformation separate (or a closure that's passed in) while the initial mesh construction and topology is in the shared module.

jeremylt:
Strangely, our tests that check for required errors are not marking the lines that generate those error messages as reached in CodeCov. Not a big deal, but annoying

examples/rust/mesh/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "mesh"
+version = "0.1.0"
+authors = [
+    "Jeremy L Thompson <thompson.jeremy.luke@gmail.com>",
+]
+edition = "2018"
+
+[dependencies]
+structopt = { version = "0.3", default-features = false }
```
jedbrown, line 10: Isn't used in this crate.

PR #684, opened 2020-12-30
Rust - use enum values from bindgen

jeremylt:


jeremylt:
This builds fine, but I need to check if it fixes the issue tomorrow

jeremylt:
Fixes the issue seen in the other PR

jedbrown (approved):

PR #686, opened 2020-12-30
Rust - drop get_/set_/add_ syntax

jeremylt:
I removed `get_`, `add_`, and `set_`:
* `*.get_*()`
* `qfunction.add_input(...)`
* `qfunction.add_output(...)`
* `operator.set_field(...)`
* `operator.add_sub_operator(...)`

We still have
* `vector.set_value(42.)`
but it seems clearer to me than `vector.value(42.)`

### DIFF
--- a/rust/src/basis.rs
+++ b/rust/src/basis.rs
@@ -205,10 +205,7 @@ impl Basis {
     ///
     /// // Create function x^3 + 1 on Gauss Lobatto points
     /// let mut u_arr = [0.; Q];
-    /// let x_nodes_arr = x_nodes.view();
-    /// for i in 0..Q {
-    ///   u_arr[i] = x_nodes_arr[i]*x_nodes_arr[i]*x_nodes_arr[i] + 1.;
-    /// }
+    /// u_arr.iter_mut().zip(x_nodes.view().iter()).for_each(|(u, x)| *u = x * x * x + 1.);
     /// let u = ceed.vector_from_slice(&u_arr);
     ///
     /// // Map function to Gauss points
@@ -217,12 +214,13 @@ impl Basis {
     /// bu.apply(1, TransposeMode::NoTranspose, EvalMode::Interp, &u, &mut v);
     ///
     /// // Verify results
-    /// let v_arr = v.view();
-    /// let x_qpts_arr = x_qpts.view();
-    /// for i in 0..Q {
-    ///   let true_value = x_qpts_arr[i]*x_qpts_arr[i]*x_qpts_arr[i] + 1.;
-    ///   assert_eq!(v_arr[i], true_value, "Incorrect basis application");
-    /// }
+    /// v.view()
+    ///     .iter()
+    ///     .zip(x_qpts.view().iter())
+    ///     .for_each(|(v, x)| {
+    ///         let true_value = x * x * x + 1.;
+    ///         assert_eq!(*v, true_value, "Incorrect basis application");
+    ///     });
     /// ```
     pub fn apply(
         &self,
@@ -282,7 +280,7 @@ impl Basis {
     /// let b = ceed.basis_tensor_H1_Lagrange(2, 1, p, 4, QuadMode::Gauss);
     ///
     /// let nnodes = b.num_nodes();
-    /// assert_eq!(nnodes, (p*p) as i32, "Incorrect number of nodes");
+    /// assert_eq!(nnodes, (p * p) as i32, "Incorrect number of nodes");
     /// ```
     pub fn num_nodes(&self) -> i32 {
         let mut nnodes = 0;
@@ -300,7 +298,7 @@ impl Basis {
     /// let b = ceed.basis_tensor_H1_Lagrange(2, 1, 3, q, QuadMode::Gauss);
     ///
     /// let nqpts = b.num_quadrature_points();
-    /// assert_eq!(nqpts, (q*q) as i32, "Incorrect number of quadrature points");
+    /// assert_eq!(nqpts, (q * q) as i32, "Incorrect number of quadrature points");
     /// ```
     pub fn num_quadrature_points(&self) -> i32 {
         let mut Q = 0;
--- a/rust/src/basis.rs
+++ b/rust/src/basis.rs
@@ -247,10 +247,10 @@ impl Basis {
     /// let dim = 2;
     /// let b = ceed.basis_tensor_H1_Lagrange(dim, 1, 3, 4, QuadMode::Gauss);
     ///
-    /// let d = b.get_dimension();
+    /// let d = b.dimension();
     /// assert_eq!(d, dim as i32, "Incorrect dimension");
     /// ```
-    pub fn get_dimension(&self) -> i32 {
+    pub fn dimension(&self) -> i32 {
         let mut dim = 0;
         unsafe { bind_ceed::CeedBasisGetDimension(self.ptr, &mut dim) };
         dim
@@ -264,10 +264,10 @@ impl Basis {
     /// let ncomp = 2;
     /// let b = ceed.basis_tensor_H1_Lagrange(1, ncomp, 3, 4, QuadMode::Gauss);
     ///
-    /// let n = b.get_num_components();
+    /// let n = b.num_components();
     /// assert_eq!(n, ncomp as i32, "Incorrect number of components");
     /// ```
-    pub fn get_num_components(&self) -> i32 {
+    pub fn num_components(&self) -> i32 {
         let mut ncomp = 0;
         unsafe { bind_ceed::CeedBasisGetNumComponents(self.ptr, &mut ncomp) };
         ncomp
@@ -281,10 +281,10 @@ impl Basis {
     /// let p = 3;
     /// let b = ceed.basis_tensor_H1_Lagrange(2, 1, p, 4, QuadMode::Gauss);
     ///
-    /// let nnodes = b.get_num_nodes();
+    /// let nnodes = b.num_nodes();
     /// assert_eq!(nnodes, (p*p) as i32, "Incorrect number of nodes");
     /// ```
-    pub fn get_num_nodes(&self) -> i32 {
+    pub fn num_nodes(&self) -> i32 {
         let mut nnodes = 0;
         unsafe { bind_ceed::CeedBasisGetNumNodes(self.ptr, &mut nnodes) };
         nnodes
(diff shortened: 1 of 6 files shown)


jedbrown:
This looks reasonable.

Should we set up chaining in this PR, perhaps like the sample below?
```diff
diff --git i/rust/src/operator.rs w/rust/src/operator.rs
index e67ff10a..c3264221 100644
--- i/rust/src/operator.rs
+++ w/rust/src/operator.rs
@@ -296,15 +296,14 @@ impl Operator {
     ///
     /// op_build.apply(&x, &mut qdata);
     ///
+    /// v.set_value(0.0);
     /// // Mass operator
     /// let qf_mass = ceed.q_function_interior_by_name("MassApply");
-    /// let mut op_mass = ceed.operator(&qf_mass, QFunctionOpt::None, QFunctionOpt::None);
-    /// op_mass.field("u", &ru, &bu, VectorOpt::Active);
-    /// op_mass.field("qdata", &rq, BasisOpt::Collocated, &qdata);
-    /// op_mass.field("v", &ru, &bu, VectorOpt::Active);
-    ///
-    /// v.set_value(0.0);
-    /// op_mass.apply(&u, &mut v);
+    /// ceed.operator(&qf_mass, QFunctionOpt::None, QFunctionOpt::None)
+    ///     .field("u", &ru, &bu, VectorOpt::Active)
+    ///     .field("qdata", &rq, BasisOpt::Collocated, &qdata)
+    ///     .field("v", &ru, &bu, VectorOpt::Active)
+    ///     .apply(&u, &mut v);
     ///
     /// // Check
     /// let array = v.view();
@@ -433,7 +432,7 @@ impl Operator {
         r: impl Into<ElemRestrictionOpt<'b>>,
         b: impl Into<BasisOpt<'b>>,
         v: impl Into<VectorOpt<'b>>,
-    ) {
+    ) -> &mut Self {
         let fieldname = CString::new(fieldname).expect("CString::new failed");
         let fieldname = fieldname.as_ptr() as *const i8;
         unsafe {
@@ -445,6 +444,7 @@ impl Operator {
                 v.into().to_raw(),
             )
         };
+        self
     }
 
     /// Assemble the diagonal of a square linear Operator
```

Note that in this model, if you want to retain the Operator in a variable, you need to split the construction
```rust
let mut op = ceed.operator(...);
op.field(...).field(...);
op.apply(...);
```

An alternative is to thread ownership through configuration (so those methods take and return ownership), in which case you could
```rust
let op = ceed.operator(...).field(...).field(...);
op.apply(...);
```
but conditional configuration would need to look like
```rust
let mut op = ceed.operator(...).field(...);
if wild {
    op = op.field(...); // need reassignment
} else {
    op = op.field(...);
}
op.apply(...);
```

jeremylt:
Adding builder syntax for the operator is pretty straightforward.

Adding the syntax for QFunctions seems to be a know hard thing in rust due to ownership and cloning issues

jeremylt:
This is the current (failed) attempt to make a builder for a QF
```
    pub fn build(&mut self) -> Self {
        let dummy_f = |_: QFunctionInputs, _: QFunctionOutputs| 0;
        Self {
            qf_core: QFunctionCore {
                ptr: std::mem::replace(&mut self.qf_core.ptr, std::ptr::null_mut()),
            },
            qf_ctx_ptr: self.qf_ctx_ptr,
            trampoline_data: self.trampoline_data,
            user_f: std::mem::replace(&mut self.user_f, Box::new(dummy_f)),
        }
    }
```

Somehow, the `user_f` field does not actually contain the user closure

jedbrown (commented):
I'm concerned about the pattern in your `build` here.
```rust
let mut b = ceed.operator(...);
b.field(...).field(...);
let op = b.build();
// b is a confusing carcass
```
Given the principle of avoiding implicit cloning, I think we need the `build` to take ownership. Alternatively, just skip having a `build` method and live with the fact that configuration needs to come in a separate call.

> When the terminal methods of the builder require ownership, there is a basic tradeoff:
>
> *    If the other builder methods take/return a mutable borrow, the complex configuration case will work well, but one-liner configuration becomes impossible.
>
> *    If the other builder methods take/return an owned self, one-liners continue to work well but complex configuration is less convenient.
>
> Under the rubric of making easy things easy and hard things possible, all builder methods for a consuming builder should take and return an owned self.

So one option is to say that `apply()` serves the role of "terminal method(s)". It only takes a reference so we can just pass a reference through. You can still do a one-liner terminating with `.apply()` if you don't need to reuse the `op` (common for setup).

The other is to thread ownership through all the configuration methods (so they would take `mut self` instead of `&mut self` and return `Self` instead of `&mut Self`). This is what [clap does](https://docs.rs/clap/2.33.3/src/clap/app/mod.rs.html#793-796), for example. You can still write a one-liner, but you can stop it any time and have ownership. Having looked at a couple examples, I lean toward this option, despite slightly awkward conditional configuration.

jeremylt:
In the multiline case, we should be able to just do
```
let mut op = ceed.operator(...);
op.field(...).field(...);
```
without adding the `op.build()`

rust/src/operator.rs
```diff
@@ -425,15 +435,15 @@ impl Operator {
     /// let b = ceed.basis_tensor_H1_Lagrange(1, 1, 2, q, QuadMode::Gauss);
     ///
     /// // Operator field
-    /// op.set_field("dx", &r, &b, VectorOpt::Active);
+    /// op.field("dx", &r, &b, VectorOpt::Active);
     /// ```
-    pub fn set_field<'b>(
-        &mut self,
+    pub fn field<'a, 'b>(
```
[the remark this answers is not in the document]
jeremylt: I don't remember exactly why I added those, but I agree that they are unneeded

jedbrown:
You can do that (passing references or ownership), but you can't do this passing references:
```rust
let op = ceed.operator().field().field();
op.apply();
```
Meanwhile, you can't do this passing ownership:
```rust
let op = ceed.operator().field();
if cond {
    op.field(); // consumes op and drops it
}
op.apply();
```
Instead, you must
```rust
let mut op = ceed.operator().field();
if cond {
    op = op.field();
}
op.apply();
```
I don't think this is a big deal. and I'd rather support the one-line configuration at the top without creating a `mut` for the user. That's why my preference is to pass ownership rather than a reference.

jeremylt:
Sure, I can change it. I was just trying to avoid the `op = op.field()` syntax because I find it ugly.

jedbrown:
I agree that it's clumsy, but I don't think any of our examples would need it. Meanwhile, I find that not being able to chain the constructor with configuration is at least as irritating (and it shows up everywhere), thus lean toward making simple things simple.

jeremylt:
This is broken in some new way now
```
let qf = ceed.q_function_interior(1, Box::new(user_f))
    .input("u", 1, EvalMode::Interp)
    .input("weights", 1, EvalMode::Weight)
    .output("v", 1, EvalMode::Interp);
```
The contents of the box holding the user context are changing after the application of `input` and `output` even though those methods do not touch that data.

rust/src/qfunction.rs
```diff
@@ -404,6 +404,16 @@ impl QFunction {
...
+        /* Note: adding an input can change the pointer to self, resulting in bad ctx data */
+        unsafe {
+            bind_ceed::CeedQFunctionContextSetData(
+                self.qf_ctx_ptr,
+                crate::MemType::Host as bind_ceed::CeedMemType,
+                crate::CopyMode::UsePointer as bind_ceed::CeedCopyMode,
+                10, /* Note: size not relevant - CPU only approach */
+                &mut self as *mut _ as *mut ::std::os::raw::c_void,
+            );
+        }
```
jeremylt, line 416: This was annoying to find

rust/src/qfunction.rs
```diff
@@ -142,12 +142,10 @@ impl fmt::Display for QFunctionCore {
...
-/// let mut qf = ceed.q_function_interior(1, Box::new(user_f));
-///
-/// qf.add_input("u", 1, EvalMode::Interp);
-/// qf.add_input("weights", 1, EvalMode::Weight);
-///
-/// qf.add_output("v", 1, EvalMode::Interp);
+/// let qf = ceed.q_function_interior(1, Box::new(user_f))
+///     .input("u", 1, EvalMode::Interp)
+///     .input("weights", 1, EvalMode::Weight)
+///     .output("v", 1, EvalMode::Interp);
```
jeremylt, line 148: Note to self: We should add a check in the C code for when we add a Operator field that the basis dim and ncomp match the QFunction field size and eval mode provided. I've generated segfaults in our C (and Rust!) examples due to a mismatch. (separate PR)

rust/src/qfunction.rs
jedbrown: Good catch, but I think this will be a recurring nightmare. We're creating a self-referential data structure so the context should be pinned. https://doc.rust-lang.org/nightly/core/pin/index.html

jeremylt: Will this be an issue though
> Instead, Pin<P> prevents certain values (pointed to by pointers wrapped in Pin<P>) from being moved by making it impossible to call methods that require &mut T on them (like mem::swap).

Will revisit tomorrow

jeremylt: Ah, you beat me to it. That's what I was thinking of doing too

jedbrown: I think this is resolved now.

jedbrown (approved):

rust/src/basis.rs
```diff
@@ -217,12 +214,13 @@ impl Basis {
...
     /// // Verify results
-    /// let v_arr = v.view();
-    /// let x_qpts_arr = x_qpts.view();
-    /// for i in 0..Q {
-    ///   let true_value = x_qpts_arr[i]*x_qpts_arr[i]*x_qpts_arr[i] + 1.;
-    ///   assert_eq!(v_arr[i], true_value, "Incorrect basis application");
-    /// }
+    /// v.view()
+    ///     .iter()
+    ///     .zip(x_qpts.view().iter())
```
jedbrown, line 219: We could impl [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) for views so that an extra `iter()` isn't needed. Or probably implement `Iterator` for our vectors.

jedbrown (commented):
Do you know if it's possible to use cargo fmt inside doctests?

rust/src/basis.rs
jeremylt: Yeah, I'd like to impl iter for the vectors

jeremylt:
I've gotten cargo fmt to work in doctests once, but I don't remember how I did it, and I remember some sort of funny issue with the output

PR #687, opened 2020-12-31
Fluids - More Regression Tests

LeilaGhaffari:
This PR adds some tests to the fluids examples in order to check Advection(2d). I think this is important for the refactoring since Nemann BCs are not tested by CI.


examples/fluids/navierstokes.c
```diff
@@ -31,9 +31,15 @@
...
+//TESTARGS -ceed {ceed_resource} -test dc_implicit_stab_none -degree 3 -dm_plex_box_faces 1,1,2 -units_kilogram 1e-9 -lx 125 -ly 125 -lz 250 -center 62.5,62.5,187.5 -rc 100. -thetaC -35. -ksp_atol 1e-4 -ksp_rtol 1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3 -implicit -ts_type alpha
+//TESTARGS -ceed {ceed_resource} -test dc_implicit_stab_supg -degree 3 -dm_plex_box_faces 1,1,2 -units_kilogram 1e-9 -lx 125 -ly 125 -lz 250 -center 62.5,62.5,187.5 -rc 100. -thetaC -35. -ksp_atol 1e-4 -ksp_rtol 1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3 -implicit -ts_type alpha -stab supg
+//TESTARGS -ceed {ceed_resource} -test adv_rotation_explicit_strong -problem advection -strong_form 1 -degree 3 -dm_plex_box_faces 1,1,2 -units_kilogram 1e-9 -lx 125 -ly 125 -lz 250 -center 62.5,62.5,187.5 -rc 100. -ksp_atol 1e-4 -ksp_rtol 1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3
```
jedbrown, line 37: A few of these tests are failing for me with small errors and presumably just need to compare at a nonzero tolerance.
```
Test failed with error norm 2.11153e-12
```

These tests that use explicit integrators can remove all the solver controls (snes, ksp, pc) since they are unused.

examples/fluids/navierstokes.c
```diff
@@ -129,17 +150,41 @@ testData testOptions[] = {
...
+  [TEST_DC_IMPLICIT_STAB_NONE] = {
     .testtol = 5E-4,
-    .filepath = "examples/fluids/tests-output/fluids-navierstokes-implicit-stab-none.bin"
+    .filepath = "examples/fluids/tests-output/fluids-navierstokes-dc-implicit-stab-none.bin"
   },
-  [TEST_IMPLICIT_STAB_SUPG] = {
+  [TEST_DC_IMPLICIT_STAB_SUPG] = {
     .testtol = 5E-4,
-    .filepath = "examples/fluids/tests-output/fluids-navierstokes-implicit-stab-supg.bin"
+    .filepath = "examples/fluids/tests-output/fluids-navierstokes-dc-implicit-stab-supg.bin"
```
jedbrown, line 163: What do you think of making the tolerance and output filepath a command-line option and then you wouldn't need this enum and initialization that only sets these two values? (I see the value if you have "cases" that set initial and boundary conditions, but here it looks like the command-line is already needed for everything else.)

jedbrown (commented):
This is looking good. I wonder if we should extend the `//TESTARGS` syntax to include a human-readable name (to appear in tap and junit output). Perhaps
```c
//TESTARGS(name="test_dc_explicit") -ceed ...
```
and we could imagine adding additional keywords for things like parallel jobs.

examples/fluids/navierstokes.c
LeilaGhaffari: > These tests that use explicit integrators can remove all the solver controls (snes, ksp, pc) since they are unused.

Ah, right. I will remove those arguments. Thanks!

LeilaGhaffari: > A few of these tests are failing for me with small errors and presumably just need to compare at a nonzero tolerance.
> 
> ```
> Test failed with error norm 2.11153e-12
> ```
Are only the explicit ones failing? I wonder why CI passed them?!

examples/fluids/navierstokes.c
LeilaGhaffari: Yeah, I figured it was getting really messy but then I thought maybe we could create an enum for each problem after we are done with the refactoring! Even in that case though, the more capabilities we add to the mini-app, the more tests would be needed and it would get messy anyway!

examples/fluids/navierstokes.c
jedbrown: Likely passed in CI because the compiler version and optimization flags matched the environment you ran in.

examples/fluids/navierstokes.c
jedbrown: How about drop the enums and pass `-compare_final_state_atol 1e-11 -compare_final_state_filename tests-output/fluids-navierstokes-dc-implicit-stab-supg.bin`? Then the specification would be self-contained in the runtime options.

jedbrown:
You can see the test report here (click View Details). Can you reproduce the relative path issue locally? 
https://gitlab.com/libceed/libCEED/-/pipelines/236258244/test_report

LeilaGhaffari:
> You can see the test report here (click View Details). Can you reproduce the relative path issue locally?
> https://gitlab.com/libceed/libCEED/-/pipelines/236258244/test_report

No! I can't reproduce the error locally but I see the problem.

LeilaGhaffari:
> You can see the test report here (click View Details). Can you reproduce the relative path issue locally?
> https://gitlab.com/libceed/libCEED/-/pipelines/236258244/test_report

@jedbrown , do you have any suggestions on how to resolve this issue? I tried `/tests-output/*.bin` as well and it failed.

LeilaGhaffari:
> and we could imagine adding additional keywords for things like parallel jobs.

I have added `name` for each test (only for fluids so far). What else could we add?

jeremylt:
Can we add the `Test Name:` only when there is a `(name="foo_baz")`? I think that would tidy up the output.

I think we should try to add names for all of the tests with multiple arguments.
For the `examples/ceed` runs I'd propose `(name="nD Gallery QFunction)` and `(name="nD user QFunction").
For `examples/nek`,`(name="BP 1")` and `(name="BP 3")`
I think those are the only other tests with multiple arguments?

LeilaGhaffari:
> Can we add the `Test Name:` only when there is a `(name="foo_baz")`? I think that would tidy up the output.
> 
Good catch! I never tested other examples after this change to see how the output would look like. 
I'll change it. Thanks.

LeilaGhaffari:
> I think we should try to add names for all of the tests with multiple arguments.
> For the `examples/ceed` runs I'd propose `(name="nD Gallery QFunction)` and `(name="nD user QFunction")`.
> For `examples/nek`,`(name="BP 1")` and `(name="BP 3")`
> I think those are the only other tests with multiple arguments?

They look good to me. I will add those too. Thanks.

LeilaGhaffari:
Can we merge this PR if there isn't anything else we need to add? I need these tests for two other branches.
@jeremylt , I had to drop the spaces from the names you proposed since the strings after the first space would be included in `TESTARGS`.

jedbrown:
We should be able to do the same thing in `junit.py`.

My recollection is that the main reason we started with `tap.sh` in bash instead of python was that python startup time is significant. However, we routinely run with so many backends that maybe it isn't significant any more. In that case, we could make `junit.py` capable of outputting TAP and not be maintaining these things in multiple languages.

An advantage of the Python is that it's easy to match quoted strings, which I interpret to be the hang-up for names that contain spaces. (Porting the TAP-style output to Python can be out of scope for this PR, but our CI uses `junit.py` because it parses the xml output.)

LeilaGhaffari:
> We should be able to do the same thing in `junit.py`.
> 
Thanks, @jedbrown .
I am not seeing the outputs but I think [this](https://github.com/CEED/libCEED/pull/687/commits/53f7577344a9c73cb421ebb337b1a19caf62ae6b) will print out the names. Does it look good? I don't know how to make `parse_testargs()` more compact!

tests/junit.py
```diff
@@ -6,12 +6,21 @@
...
 def parse_testargs(file):
+    allargs = []
     if os.path.splitext(file)[1] in ['.c', '.cpp']:
-        return sum([[line.split()[1:]] for line in open(file).readlines()
-                    if line.startswith('//TESTARGS')], [])
+        for line in open(file).readlines():
+            if line.startswith('//TESTARGS'):
+                args = [line.split()[1:]]
+                args.append([line.split()[0].strip('//TESTARGS(name=').strip(')')])
+                allargs.append(args)
```
jeremylt, line 15: What about 
```
return sum([[line.split()[1:] + line.split()[0].strip('C_TESTARGS(name=').strip(')')
            for line in open(file).readlines()
            if line.startswith('//TESTARGS')], [])
```
(I think I got the syntax right)

examples/ceed/ex1-volume.c
```diff
@@ -39,12 +39,12 @@
...
-//TESTARGS -ceed {ceed_resource} -d 3 -t
-//TESTARGS -ceed {ceed_resource} -d 1 -t -g
-//TESTARGS -ceed {ceed_resource} -d 2 -t -g
-//TESTARGS -ceed {ceed_resource} -d 3 -t -g
+//TESTARGS(name="1D_user_QFunction") -ceed {ceed_resource} -d 1 -t
+//TESTARGS(name="2D_user_QFunction") -ceed {ceed_resource} -d 2 -t
+//TESTARGS(name="3D_user_QFunction") -ceed {ceed_resource} -d 3 -t
+//TESTARGS(name="1D_Gallery_QFunction") -ceed {ceed_resource} -d 1 -t -g
+//TESTARGS(name="2D_Gallery_QFunction") -ceed {ceed_resource} -d 2 -t -g
+//TESTARGS(name="3D_Gallery_QFunction") -ceed {ceed_resource} -d 3 -t -g
```
jeremylt, line 47: The capitalization difference bugs me. What do you think about `_User_`?

jeremylt (commented):
I'm not as sure about the fluids part, but the changes to the test suite look good to me with the changes I suggested.

examples/ceed/ex1-volume.c
LeilaGhaffari: Agreed! Thanks! Done!

tests/junit.py
LeilaGhaffari: Hmm! So, I have stored names in a separate sublist for each test to be able to retrieve them [here](https://github.com/CEED/libCEED/blob/4cc1b4280947a15993ad46cc05fb6c1cbb44106c/tests/junit.py#L72) and print them [here](https://github.com/CEED/libCEED/blob/4cc1b4280947a15993ad46cc05fb6c1cbb44106c/tests/junit.py#L92) if there is a name. Are you thinking of a better approach?

jeremylt:
```
return sum([[[line.split()[1:]] + [line.split()[0].strip('C_TESTARGS(name=').strip(')')]
            for line in open(file).readlines()
            if line.startswith('//TESTARGS')], [])
```
No, I was just trying to make the syntax more compact. Maybe that syntax?

LeilaGhaffari: Thank you! Unfortunately, none of them work.

jeremylt:
```
def parse_testargs(file):
    if os.path.splitext(file)[1] in ['.c', '.cpp']:
        return sum([[[line.split()[1:], [line.split()[0].strip('//TESTARGS(name=').strip(')')]]]
                    for line in open(file).readlines()
                    if line.startswith('//TESTARGS')], [])
        return allargs
    elif os.path.splitext(file)[1] == '.usr':
        return sum([[[line.split()[1:], [line.split()[0].strip('C_TESTARGS(name=').strip(')')]]]
                    for line in open(file).readlines()
                    if line.startswith('C_TESTARGS')], [])
    raise RuntimeError('Unrecognized extension for file: {}'.format(file))
```
This one?

LeilaGhaffari: Beautiful! It worked, thanks!

jeremylt (approved):
I think this is ready

LeilaGhaffari:
Thanks. I will rebase it unto main and merge it. `squash-merge`?

jeremylt:
Sounds good to me

PR #691, opened 2021-01-04
Error Handling Improvement [fix #696]

### LINKED ISSUE #696: Error/Exception Handling
I am currently reviewing libCEED as part of https://github.com/openjournals/joss-reviews/issues/2945.

One issue I've identified is around how errors are handled.  Specifically, standard operating procedure at the moment is along the lines of:
```
ierr = MightFail(); CeedChk(ierr);
```
where `CeedChk` expands to a return statement should `ierr` be non-zero.  This paradigm is problematic as it typically leaves the entity which the function is operating on in an undefined state.  This makes it virtually impossible for a user of libCEED do any meaningful error handling.  Moreover, it appears as if there several situations where this can result in memory being leaked; if a method allocates some memory (which succeeds) but then calls another internal method (which fails) the current code will immediately return without first freeing the temporary object it allocated.

There are also a few oddities.  For example:
https://github.com/CEED/libCEED/blob/06988bf74cc6ac18eacafe7930f080803395ba29/interface/ceed-operator.c#L67-L68
where the second `ierr` check appears to be redundant.

jeremylt:
I added compatibility checks between the QFunction and Operator fields.

Also, I made QFunctions immutable once they are associated with an Operator that has Apply or Assemble called. We have acted as though QFunctions are immutable once the Operator has been called, but we have not enforced this assumption.


interface/ceed-operator.c
```diff
@@ -876,7 +970,7 @@ int CeedOperatorSetField(CeedOperator op, const char *fieldname,
   }
   for (CeedInt i=0; i<op->qf->numoutputfields; i++) {
     if (!strcmp(fieldname, (*op->qf->outputfields[i]).fieldname)) {
-      qfield = op->qf->inputfields[i];
+      qfield = op->qf->outputfields[i];
```
jeremylt, line 973: Luckily, this didn't cause bugs yet

jeremylt:
OCCA Hip failures seem to be leaking into other branches now

jeremylt:
I'm converting this PR into a more broad error handing PR.

include/ceed.h
```diff
@@ -192,6 +192,29 @@ CEED_EXTERN int CeedSetErrorHandler(Ceed ceed,
...
+  CEED_ERROR_DIMENSION     = 4,
+  /// Recoverable error, incomplete object setup
+  CEED_ERROR_INCOMPLETE    = 6,
+  /// Unrecoverable error, generic
+  CEED_ERROR_UNRECOVERABLE = 1,
+  /// Unrecoverable error, internal backend error
+  CEED_ERROR_BACKEND       = 3,
+  /// Unrecoverable error, operation unsupported by current backend
+  CEED_ERROR_UNSUPPORTED   = 5,
+} CeedErrorType;
```
jeremylt, line 216: @jedbrown, any opinions on the error codes?

jedbrown: Any reason not to use positive and negative numbers instead of even/odd?

I'm not sure "recoverable" is the best term. It means that some object state might be inconsistent and/or resources may have leaked. "Recoverable" at this point should probably mean that the library and object states have not been modified since the call. We'll eventually want to distinguish incomplete modification of an output parameter, in which case the user could try again with different inputs (e.g., shorten a time step and retry).

jeremylt: The only reason to use even/odd over positive/negative is to provide human readable error code strings like we do with our other enums.

Agreed re recoverable, but I'm not sure what is a good replacement word? How about terminal and nonterminal?

jedbrown: The pointer can be shifted so there's no problem with an array of error strings.

jeremylt:
Changes so far:
* Add field checking logic
* Add error enum
  * 0 - success
  * negative - terminal error
  * positive - nonterminal error
* set all backend errors to terminal
* add `CeedChkBackend()` which promotes all other errors found in backend to terminal

ToDo:
* Check nonterminal errors to make sure they don't leak
* Use terminal/non-terminal errors to improve Python/Rust error handling
* Add basis mode + dimension checking logic (my second most common errors after field mismatch, biggest cause of accidental segfaults)

jeremylt:
I'm breaking for the day. I have classified all of the errors as terminal or nonterminal in my opinion, but it should probably have other people's opinions.

If we like this, I can start making Python and Rust panic vs return an error with these codes

I also added some dimension checking to the basis evaluations - that was my 2nd biggest segfault causing issue after field mismatches, so I'd like to have a check if we think it's worth the cost. Should be a minor perf impact for blocked backends and no impact for JiT backends.

jeremylt:
The python interface now will raise an exception on positive error codes and exit on negative error codes. With the error checking logic improvements in this PR, the majority of errors should be positive error codes, and the user can try to catch and fix the exception if desired.

jeremylt:
Ok @jedbrown, I think this is ready for review again

jeremylt:
@pazner, I think it would be good to add error handling to the Julia interface in a follow-up PR. Since Julia can be run interactively, I think we should model the error handling in the Julia interface after the error handling in the Python interface.

pazner:
> @pazner, I think it would be good to add error handling to the Julia interface in a follow-up PR. Since Julia can be run interactively, I think we should model the error handling in the Julia interface after the error handling in the Python interface.

Good idea, I'll look into this.

jeremylt:
I wrote error handling for the Python interface, but I'm not super familiar with the Julia interface. Just let me know if/how I can help

pazner:
I think we should already be good to go on the Julia interface without any changes.

Julia registers a custom error handler than raises an exception, and prints the error code and message. In non-interactive use, this will terminate the program unless the the exception is caught. In interactive use, this allows the user to recover from non-fatal errors.

jedbrown (approved):
Sorry this has been lingering. It's ready to squash-merge as far as I'm concerned.

jeremylt:
@jedbrown, I rebased to merge, but there is some error with the Python job on main that's in this branch too now

jedbrown:
Sorry, should be fixed in bd4df46207a63b48a485389a63848eb7a2477f6b.

jeremylt:
Awesome, I'll squash+merge once CI finishes

jedbrown:
Cool, and sorry about skipping the fresh pipeline I should have run, and that would have caught that hiccup.

PR #693, opened 2021-01-06
julia: pin JuliaFormatter to v0.12.0 because subminor versions are changing

### LINKED ISSUE #690: Julia formatting stability for CI
JuliaFormatter.jl has made a couple releases recently that change our code rendering. I applied changes in v0.12.0 (97769a3d2c45e9067c16da7ef18281dfae6857b7) on Friday, but there are more changes now and the diff below seems odd. I believe we can pin the versions in `.style/Project.toml` so we stop having surprising CI failures, but what should we do with the case below?

```diff
diff --git i/julia/LibCEED.jl/src/Cuda.jl w/julia/LibCEED.jl/src/Cuda.jl
index 372871ad..4a13c1d7 100644
--- i/julia/LibCEED.jl/src/Cuda.jl
+++ w/julia/LibCEED.jl/src/Cuda.jl
@@ -86,8 +86,8 @@ function generate_kernel(qf_name, kf, dims_in, dims_out)
     end

     def_outs = [
-        :($(f_outs[i]) = LibCEED.MArray{Tuple{$(dims_out[i]...)},Float64}(undef))
-        for i = 1:noutputs
+        :($(f_outs[i]) = LibCEED.MArray{Tuple{$(dims_out[i]...)},Float64}(undef)) for
+        i = 1:noutputs
     ]

     device_ptr_type = Core.LLVMPtr{CeedScalar,LibCEED.AS.Global}
```

jedbrown:
v0.12.0 to v0.12.2 changes some formatting, one in a peculiar way that
I expect will be reversed in a future version of JuliaFormatter. Since
subminor versions are causing a lot of noise, we'll pin for now.

Fix #690

### DIFF
--- a/julia/LibCEED.jl/.style/Project.toml
+++ b/julia/LibCEED.jl/.style/Project.toml
@@ -2,3 +2,6 @@
 CSTParser = "00ebfdb7-1f24-5e51-bd34-a7502290713f"
 JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 Tokenize = "0796e94c-ce3b-5d07-9a54-7f471281c624"
+
+[compat]
+JuliaFormatter = "= 0.12.0"


pazner (approved):
Looks good to me 👍 I made the same changes locally too.

PR #698, opened 2021-01-24
Add version macros and CeedGetVersion() [fix #697]

### LINKED ISSUE #697: Obtaining libCEED version at compile time/run time
I am currently reviewing libCEED as part of https://github.com/openjournals/joss-reviews/issues/2945.

Looking through the documentation/header files I am wondering how one can obtain the libCEED version at both compile time and run time (the latter being useful for those who are interfacing at the ABI level).

jedbrown:
If this is deemed sufficient, I can add support for `CeedGetVersion()` from the other language bindings.

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -1015,4 +1015,29 @@ int CeedResetErrorMessage(Ceed ceed, const char **errmsg) {
   return 0;
 }
 
+/**
+  @brief Get libCEED library version info
+
+  libCEED version numbers have the form major.minor.patch. Non-release versions
+  may contain unstable interfaces.
+
+  @param[out] major    Major version of the library
+  @param[out] minor    Minor version of the library
+  @param[out] patch    Patch (subminor) version of the library
+  @param[out] release  True for releases; false for development branches.
+
+  The caller may pass NULL for any arguments that are not needed.
+
+  @sa CEED_VERSION_GE()
+
+  @ref Developer
+*/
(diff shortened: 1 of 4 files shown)


interface/ceed.c
```diff
@@ -15,6 +15,7 @@
 // testbed platforms, in support of the nation's exascale computing imperative.
 
 #define _POSIX_C_SOURCE 200112
+#include <ceed.h>
```
jeremylt, line 18: Really, `ceed.h` is included in `ceed-backed.h`, which is included in `ceed-impl.h`, so perhaps everything in `interface/*` should just include `ceed-impl.h` and everything in `backends/*` should just include `ceed-backend.h`?

jedbrown: That's the PETSc style and I think would be fine here. I moved this because feature test macros need to be defined first. Or actually, heh, my editor/clangd automatically added `ceed.h` up above the feature test macro and I moved it down without realizing it just doesn't need to be there. (A lot of style guides say you should directly include what you use, rather than relying on transitive dependencies.)

jeremylt: I don't have strong feelings one way or another. I'm fine with including all used headers for clarity if that is a more typical style convention - I just object to the current inconsistency even though it has no effect on the code itself and this change happened to bring that inconsistency to my attention

I can spin up a short PR to make the includes all match in style

jeremylt:
This looks reasonable to me

PR #699, opened 2021-01-26
Improve hip-shared memory usage

nbeams:
This PR improves the performance of the hip-shared backend: it gets between 30-80% improvement over the original hip-shared port for MFEM BP3 (ROCm 4.0 / MI50 GPU).  Previously, hip-ref actually outperformed hip-shared for lower orders of basis functions; with these changes, this is no longer the case.

cuda-shared copies the 1D basis arrays into constant memory.  There is no benefit to doing this for AMD, as constant memory is not a separate memory block (per @moes1).  Furthermore, profiling tests showed that `hipMemcpyToSymbol` was a bottleneck in the code.  Thus, this PR removes the `CeedHipInitInterp*` routines from 
`hip-shared/kernels/hip-shared-basis.hip.cpp` - which means removing the `kernels` folder from hip-shared entirely. 
Instead, the 1D interp and grad arrays are read into shared memory through a `loadMatrix` routine like the one in cuda/hip-gen.


backends/hip-shared/ceed-hip-shared-basis.c
```diff
@@ -704,10 +713,15 @@ __device__ void weight3d(const CeedInt nelem, const CeedScalar *qweight1d,
...
 extern "C" __launch_bounds__(INTERP_BLKSIZE) __global__ void interp(
                                   const CeedInt nelem, const int transpose,
-                                  const CeedScalar *c_B,
+                                  CeedScalar *d_interp1d,
                                   const CeedScalar *__restrict__ d_U,
                                   CeedScalar *__restrict__ d_V) {
+
   HIP_DYNAMIC_SHARED( double, slice)
+  // load interp1d into shared memory
+  __shared__ double c_B[P1D*Q1D];
```
YohannDudouit, line 722: I would call it `s_B`, the semantic I tried to use is: `s_*` for shared memory, `r_*` for registers, `d_*` for device memory, and `c_*` for constant memory.

backends/hip-shared/ceed-hip-shared-basis.c
```diff
@@ -722,10 +736,16 @@ extern "C" __launch_bounds__(INTERP_BLKSIZE) __global__ void interp(
...
-                                const CeedScalar *c_B, const CeedScalar *c_G,
+                                CeedScalar *d_interp1d, CeedScalar *d_grad1d,
                                 const CeedScalar *__restrict__ d_U,
                                 CeedScalar *__restrict__ d_V) {
   HIP_DYNAMIC_SHARED( double, slice)
+  // load interp1d and grad1d into shared memory
+  __shared__ double c_B[P1D*Q1D];
+  loadMatrix(d_interp1d, c_B); 
+  __shared__ double c_G[P1D*Q1D];
+  loadMatrix(d_grad1d, c_G);
```
YohannDudouit, line 747: ditto

YohannDudouit (approved):
Nice work @nbeams , it could be interesting to look into `cuda-shared` too to see if we really get a performance improvement by using constant memory.

backends/hip-shared/ceed-hip-shared-basis.c
nbeams: Good catch, I fixed this in 4e6ce5a03b715256484190a896dd2a02cda2ad8b

nbeams:
I'm confused by the GitLab failures.  From the error message it seems there's a problem with `bpsraw`, but from the log, I don't understand what.  When I build the PETSc examples manually on Noether (using the PETsc in `/projects/petsc`, if that matters), `bpsraw` builds and runs with no issue (with `-ceed /gpu/hip/shared`).

jeremylt:
I'm seeing 
```
  KSP:
    KSP Convergence                    : DIVERGED_ITS
    Total KSP Iterations               : 15
    Final rnorm                        : 5.803150e-09
```
So it looks like its a little bit less accurate?

knepley:
@jeremylt Delayed convergence in CG is a well-known effect of finite precision jitter. For example, see this paper: https://royalsocietypublishing.org/doi/pdf/10.1098/rsta.2019.0050?casa_token=fIwmF0HtJEEAAAAA:hGqzC_eWS6aLqdTynAmk9-mqNsFSPiCf02L-m4Om5AkgxggKl3j32gTymxkzprOH3tbdppVcevfyDw

jeremylt:
As I understand it though, this PR just changes where the basis data is stored rather than how the computation is executed, so I'm not sure why that would result in a >5.8x increase in error? Perhaps I misunderstand the effect of this PR?

nbeams:
Trying my manually-built `bpsraw`, I get this convergence info for `./petsc-bpsraw -problem bp3 -ceed /gpu/hip/shared`:
```
  KSP:
    Total KSP Iterations               : 2
    Final rnorm                        : 8.650038e-16
  Performance:
    Pointwise Error (max)              : 1.112075e-01
 ```
 
 And this for hip-ref:
 ```
  KSP:
    Total KSP Iterations               : 2
    Final rnorm                        : 8.835249e-16
  Performance:
    Pointwise Error (max)              : 1.112075e-01
```

jeremylt:
What about with these args?
```
//TESTARGS -ceed {ceed_resource} -test -problem bp2 -degree 5 -qextra 1 -ksp_max_it_clip 15,15
```

nbeams:
Hmm, yes, now I see.  But you're correct that the actual computation wasn't changed at all, so I'm not sure what's going on here...

nbeams:
Because this branch already had conflicts with `main` at the time I fixed my bug that caused the PETSc test error, I was originally waiting until after the other JOSS-review-related PRs (#698, #691) were merged to do rebase/conflict resolution, assuming that that was the preferred order -- is that correct?  Or should I go ahead and work on updating this branch?

nbeams:
I decided to go ahead and do a rebase to resolve the conflicts with main, but I'm surprised/confused by the failures on GitLab.  This time, it looks like the failures are all coming from the Fortran tests, but I'm unable to reproduce the issue on Noether manually.  The Fortran tests all pass for me.

jeremylt:
It looks like a bunch of HIPBLAS issues? Things like: `HIPBLAS_STATUS_ALLOC_FAILED`

nbeams:
Wait, where did you see that?  I don't see it in the GitLab interface -- where should I be looking?  Or were you able to reproduce Fortran failures locally?

jeremylt:
I found some details on this page: https://gitlab.com/libceed/libCEED/-/pipelines/258422511/test_report

nbeams:
Oh, I see, I needed to go to the "Tests" tab and then "View details." 
What I don't understand is why I cannot recreate these failures when testing manually on Noether.  Shouldn't a hipBLAS issue affect that, too?  I built with gcc/gfortran, which is what the test uses.

jeremylt:
It might have been the env on Noether at that time? I'm about to merge a small change to master, so updating this branch will give us another shot to see if that was just a temporary issue.

nbeams:
I did a local rebase on Noether (haven't pushed yet) and was able to see hipBLAS errors, though I still can't figure out a pattern as to when they happen.  It looks like the same issue was affecting https://gitlab.com/libceed/libCEED/-/pipelines/260805581/test_report from #702, so I don't think it's anything related to this branch specifically, not that that really helps narrow it down...

jeremylt:
I'm having some success replicating failures in parallel with
```
make prove -j16
```
so maybe this is some sort of ROCm issue?

jeremylt:
Nope, parallelism doesn't have anything to do with it. I'm seeing it with `-j1` as well.

nbeams:
I tried updating and rebuilding my local hipMAGMA (/projects/nbeams/magma) just to see what happened.  When I tried running a test DGEMM script afterwards, I got `failed at testing/testing_dgemm.cpp:128: error -113: cannot allocate memory on GPU device` 

And currently every time I check `rocm-smi` on Noether, the VRAM% is showing as 100%, even if GPU % is low (2%, 10%, 0%...)

jeremylt:
Running the tests slower with `make prove PROVE_OPTS=-v` is not helping either. Is it possible that there is some sort of ROCm leak that filled up the device?

nbeams:
Just got a clean build with my own hipMAGMA build.  Now we'll see what the pipeline does with the one in /projects/hipMAGMA.  The VRAM% in rocm-smi is back to a reasonable number now whenever I check.

jeremylt:
@nbeams we think we figured it out. Another user was doing something that required a fair amount of device memory, which didn't leave enough memory for CI. Can we build an estimate of how much device memory we'd like to have on reserve for CI?

DiffeoInvariant:
Hey @nbeams --- as @jeremylt alluded to above, I was running a process that preallocates 90% of the GPU memory (and can get pretty greedy with the rest), but doesn't actually need all of that memory. If y'all have an estimate of how much VRAM you need for the tests though, I can ensure I always leave at least that much available

nbeams:
Thanks @DiffeoInvariant.  Is there a limit to how much extra can be allocated on top of the "baseline" preallocated amount?  I would have thought 10% would be plenty for the CI runs, but it seems 10% was the maximum available under the old settings, but not a guaranteed amount.  Of course it's hard to say for sure, since you never know exactly which tests will be running at the same time (and which backends).  I suppose we can always lower the `-j` parameter for the CI as well if it becomes an issue.

jeremylt:
Yeah, Jed and I were thinking of perhaps setting separate nproc for the cpu and gpu backend testing.

PR #701, opened 2021-01-28
headers - clearify includes to not rely on transitive includes

jeremylt:
This is just a minor (but touches a lot of files) style edit - no actual code changes

### DIFF
--- a/backends/avx/ceed-avx-blocked.c
+++ b/backends/avx/ceed-avx-blocked.c
@@ -16,6 +16,7 @@
 
 #include <ceed.h>
 #include <ceed-backend.h>
+#include <stdbool.h>
 #include <string.h>
 #include "ceed-avx.h"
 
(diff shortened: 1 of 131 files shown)


jeremylt:
tragically, not that I have found, but I'm still looking

jeremylt:
Google seems to think that it's a known hard thing to check for and there aren't really any tools to help.

jedbrown:
https://include-what-you-use.org/

jeremylt:
Ready now @jedbrown for squash+merge. There is a lot of manual intervention, so that tool is not good for CI, but it was useful for this PR.

jedbrown (approved):
LGTM. Is there a convention about ordering and spacing between headers (sometimes there's  newline and sometimes not)? And could you add a developer note about the tool?

jeremylt:
I'll add a note. Most of the files used
* one line of whitespace before and after
* installed headers before local headers
* headers alphabetized but with ceed headers listed first
so I tried to stick to that

PR #649, opened 2020-10-30
adding equations for MR and GP

karenlstengel:

PR #703, opened 2021-02-04
doc: replace recommonmark with MyST

jedbrown:
```
In addition to offering more natural $\pi$ math (versus :math:`\pi`) and
markdown syntax for all ReST features, this also adds support for
running and rendering Jupyter notebooks within the Sphinx site.
```


jedbrown:
Some recent updates in the MyST ecosystem have made it possible for us to automatically migrate all the ReST sources to MyST. My inclination is to wait until after we release v0.9 and merge #649 (which has nontrivial edits to solids documentation), then update this PR to implement that change. There will be much rejoicing in my book when we can stop needing to teach and remember ReST.

jedbrown:
Okay, all the docs are converted to MyST and the formatting looks good to me.

I finally figured out why the Altair plot was not showing up on the RTD website and fixed it (path idiosyncrasy).
image
 https://libceed--703.org.readthedocs.build/en/703/examples/solids/#equation-log1p

This isn't rendered in the PDF, but it never was and I don't think we should omit interactive features of the website in homage to the static-PDF-with-DOI gods.

I'm ready for review and merging in this state. I'd like to follow up by promoting `conf.py` to the top level so we can avoid the copies, which break the "Edit on GitHub" links and auto-rebuild workflows.

jeremylt (approved):
Rendered output looks good to me. Do we want to close #729 with this as well?

PR #705, opened 2021-02-11
Modify errmsg handling related to CeedErrorStore

barker29:
Currently, `CeedErrorStore` puts its message in the top-level `Ceed` object by following parents up the tree. However, `CeedGetErrorMessage` simply gets its own error message without going up the tree, so it may return a different error message from the one put with `CeedErrorStore` even if they are both called with the same `Ceed` object.

This PR modifies `CeedErrorFormat`, `CeedGetErrorMessage`, `CeedResetErrorMessage` to act more consistently with `CeedErrorStore`.

I discovered this behavior when trying to write a error handling wrapper in MFEM, see https://github.com/mfem/mfem/pull/1763 .

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -850,6 +850,10 @@ int CeedDestroy(Ceed *ceed) {
 
 // LCOV_EXCL_START
 const char *CeedErrorFormat(Ceed ceed, const char *format, va_list *args) {
+  if (ceed->parent)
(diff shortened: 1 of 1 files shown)


jeremylt (approved):
Looks fine to me, though @jedbrown will have to review+merge since GitHub actions don't run on branches of forks.

jedbrown (approved):

PR #706, opened 2021-02-24
gitlab - display number processes avaliable for tests

jeremylt:
I added this so we can see a bit more info.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,6 +9,7 @@ noether-rocm:
   script:
 # Compilers
     - export COVERAGE=1 CC=gcc CXX=g++ FC=gfortran HIPCC=hipcc
+    - echo "---------- nproc-----------" && echo $(nproc)
     - echo "---------- CC -------------" && $CC --version
     - echo "---------- CXX ------------" && $CXX --version
     - echo "---------- FC -------------" && $FC --version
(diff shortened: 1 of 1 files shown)


jeremylt:
It looks like the default value for `nproc` in our CI is 64 - is this what we want?

jeremylt:
Ok, now this PR lets us know how many processes we're using while testing and reduces the number of processes for the GPU tests.

.gitlab-ci.yml
```diff
@@ -9,6 +9,7 @@ noether-rocm:
   script:
 # Compilers
     - export COVERAGE=1 CC=gcc CXX=g++ FC=gfortran HIPCC=hipcc
+    - echo "---------- nproc-----------" && NPROC_CPU=$(nproc) && NPROC_GPU=8
```
jedbrown, line 12: Should `NPROC_GPU` be a minimum of `$(nproc)` and some static (or determined) maximum?

jedbrown (approved):
I'm okay with this if it's needed, though I want there to be a better way, such a resource manager that limits the number of concurrent jobs on CPU and GPU. https://github.com/flux-framework/flux-core/pull/3567#issuecomment-808372257

.gitlab-ci.yml
jeremylt: agreed and updated

jeremylt:
I feel like this is a good stopgap for now and we can make an issue to remind us to move to something more sophisticated in the future

jedbrown (approved):
Thanks, looks good to squash-merge.

PR #707, opened 2021-03-03
add CUDA_ARCH and HIP_ARCH options

nbeams:
Allow user to set GPU architecture variables for nvcc/hipcc through `CUDA_ARCH`/`HIP_ARCH` environment variables. 

Default values are left blank, and the flags are only added if the user specifies something; nothing will change for people who don't need this option in their current build setup.

I wasn't sure if/where I should add information about these options to the documentation.


jedbrown (commented):
Can you make this save to the file (`CONFIG_VARS`)?

jeremylt (approved):
Looks good to me

nbeams:
It it okay to go ahead and merge this now?

jedbrown (approved):

PR #708, opened 2021-03-09
Add ftr cur config

ArashMehraban:
I have implemented  anew QFunction based on tensor b and Tau. This code outperforms our current implementation with the 2nd Piola Kirchhoff code for the hyperelasticity at finite strain.
I have some suggestions for improvement:
- log1p function should be moved to common.h from hyperFS to be used in hyperFS and hyperFSCUR
- energy function needs to be done using b tensor. It is based on E (using C) right now

How to run on Noether:

`mpirun -n 16 ./elasticity -problem hyperFSCUR -degree 1 -mesh ./meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view`


jeremylt:
> log1p function should be moved to common.h from hyperFS to be used in hyperFS and hyperFSCUR

This will break the ability to run this code on CUDA/HIP, so we can't do this.

---

Do you have numbers on how this compares in terms of performance?

---

I'm super happy with calling this `_CUR` - it is a bit awkward, but a better name doesn't come to mind right now.

---

Also on the ToDo list - documentation updates

jedbrown:
If I'm following your intent, strain energy is like the residual and there's no specific reason to prefer current configuration for integrating strain energy.

rezgarshakeri:
> If I'm following your intent, strain energy is like the residual and there's no specific reason to prefer current configuration for integrating strain energy.

@jedbrown , right now for computing energy we use C, but we were thinking that we can write the energy function in terms of `b`  since trace(C) = trace(b). 
Just to make hyperFSCUR.h all in terms of `b` and `tau`

ArashMehraban:
Yes, we know that trace(C) = trace(b), so we are trying to get rid of `computeDetCM1()` function to make this implementation even faster. 

In addition, as Simo did, for isotropic materials most researchers work directly with the Kirchhoff stress while second Piola Kirchhoff stress formulation is general for anisotropic materials (which requires mapping to `tau`). Therefore, we expect this residual evaluation and action of Jacobian be faster using tau and b

jedbrown:
So C and b have all the same invariants, not just trace. det(C) - 1 is used in place of J because it's readily available and computing J (which then becomes log(J) or (J-1)) is less stable.

ArashMehraban:
@jeremylt about your comment: 
"I'm super happy with calling this `_CUR` - it is a bit awkward, but a better name doesn't come to mind right now."

I let you guys decide what you want to call it, but I'd be careful about terminology. If you are referring to the indicial notation, yes, `tau` and `b` use lowercase `i` and `j` while reference config uses upper case  `I` and `J`. So it is tempting to insinuate that `b` and `tau` implementation refer to the current config and use `_CUR`. Actually for a lack of better word, I did add `CUR` to the end the QFunction I just implemented. However, I believe, it is commonly understood among continuum mechanic experts that current config formulation requires updating coordinates and that is something that we are not doing that per se.

examples/solids/qfunctions/hyperFSCUR.h
```diff
@@ -0,0 +1,684 @@
...
+    const CeedScalar F[3][3] =  {{gradu[0][0][i] + 1,
+                                  gradu[0][1][i],
+                                  gradu[0][2][i]},
+                                 {gradu[1][0][i],
+                                  gradu[1][1][i] + 1,
+                                  gradu[1][2][i]},
+                                 {gradu[2][0][i],
+                                  gradu[2][1][i],
+                                  gradu[2][2][i] + 1}
+                                };
```
jedbrown, line 383: Jacobian action should not require building F. This algorithm appears to be accessing more data than Algorithm 2 of Davydov. What happened here?

jedbrown: I may have missed something, but per our Jan 26 email:
```
The issue here is we don't want to store F when evaluating the action of the Jacobian. We have

x_current = x_initial + u

We also have a constitutive relation \tau(\bm b, J) from Eq 14, and we're discretizing Eq 7 of Davydov, which can be written

\int_{B_current} \sigma : symgrad(du) = \int_{B_initial} J \sigma : symgrad(du)

where \tau = J \sigma and symgrad is in the current configuration. Their notation is clumsy, IMO, since we're really doing the integral in the current configuration so I'd just use \sigma (Cauchy stress). We know that

dF = \partial du/\partial x_initial
   = (\partial du/\partial x_current) (\partial x_current/\partial x_initial)
   = (\partial du/\partial x_current) F
   = grad(du) F

So you need to be able to evaluate db(du), which is

  db = dF F^T + F dF^T 
     = grad(du) F F^T + F F^T grad(du)^T
     = grad(du) b + b grad(du)^T

if I haven't messed something up in typing this out.
```

rezgarshakeri: This code right now answer this question: Is it better to implement hyper-elasticity with constitutive equation in terms of S and C, or tau and b?!

jedbrown: So I'm proposing that we should store (during residual evaluation or Jacobian "assembly") the derivative of reference with respect to current so that the Jacobian need only access that, not derivative of reference with respect to initial *and* derivative of initial with respect to current. It cuts down memory access and flops, and should give us a speed boost.

rezgarshakeri: If I understood correctly, you propose to save
dx_ref/dx_cur
right now dXdx which is dx_ref/dx_initial provided as input for residual and Jacobian. So we need to do
dx_ref/dx_cur = dx_ref/dx_initial * dx_initial/dx_cur = dx_ref/dx_initial * F^{-1}
(I think this is what you wrote in email)

rezgarshakeri:
> ## I was wondering if logp could be removed. Repeating it cause C guard problems, so I simply changed its name in the new Qfunction.
> Yes this code outperforms the current implementation by far except for order 2:
> degree 1
> tensor b:
> SNES Solve Time                    : 158.432 (158.424) sec
> tensor C:
> SNES Solve Time                    : 162.111 (162.111) sec
> ## degree 2
> :tensor b:
> SNES Solve Time                    : 64.7685 (64.7683) sec
> tensor C: SNES Solve Time                    : 53.3971 (53.393) sec
> ## degree 3:
> tensor b:
> SNES Solve Time                    : 521.59 (521.59) sec
> tensor C:
> SNES Solve Time                    : 548.046 (548.046) sec
> 
> degree 4:
> tensor b:
> SNES Solve Time                    : 772.913 (772.897) sec
> tensor C:
> SNES Solve Time                    : 1126.11 (1126.1) sec.

would you please run again?
I got these
Screenshot from 2021-03-10 15-39-16

examples/solids/qfunctions/hyperFSCUR.h
jedbrown: Yes, exactly. See also the derivation in Wriggers
image
and Davydov
image

rezgarshakeri: One more question, in Jacobian we have to compute tau, so we need Grad(u) to compute F then left Cauchy tenser b, then tau.
I think now we need to store Grad(u) and also dx_ref/dx_cur (or let say F^{-1}), am I right?
and if we want to store 2 variables in current configuration, why not do the same thing reference configuration by saving Grad(u) and C^{-1}
Screenshot from 2021-03-10 15-51-30

ArashMehraban:
I ran bending the tube problem and my numbers come from there. What is the problem you are running? Many things changed since my original commit too. I was waiting to see if the conclusion of this thread leads to only one storage or multiple storage as in Davydov's algorithm 2

ArashMehraban:
`mpirun -n 16 ./elasticity -problem hyperFSCUR -degree 1 -mesh ./meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view
`

SNES:
    SNES Type                          : newtonls
    SNES Convergence                   : CONVERGED_FNORM_RELATIVE
    Number of Load Increments          : 25
    Completed Load Increments          : 25
    Total SNES Iterations              : 75
    Final rnorm                        : 8.291981e-04
  Linear Solver:
    Total KSP Iterations               : 2618
    PC Type                            : gamg
  Performance:
    **SNES Solve Time                    : 50.6121 (50.6001) sec**
    DoFs/Sec in SNES                   : 0.938531 (0.938752) million
    Strain Energy                      : 1.133625855188e+07

AND

`mpirun -n 16 ./elasticity -problem hyperFS -degree 1 -mesh ./meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view`


 SNES:
    SNES Type                          : newtonls
    SNES Convergence                   : CONVERGED_FNORM_RELATIVE
    Number of Load Increments          : 25
    Completed Load Increments          : 25
    Total SNES Iterations              : 75
    Final rnorm                        : 8.292006e-04
  Linear Solver:
    Total KSP Iterations               : 2618
    PC Type                            : gamg
  Performance:
    **SNES Solve Time                    : 59.3996 (59.3995) sec**
    DoFs/Sec in SNES                   : 0.799685 (0.799686) million
    Strain Energy                      : 1.133625855188e+07

ArashMehraban:
I tried what Jed had in mind when I first implemented this in MATLAB. I still have the contents of the email above in my MATLAB implementation as comments: https://github.com/ArashMehraban/MatlabTestsPetsc/blob/617d21b542f3ab82273d045e1d4db0e82ebb488a/HyperFSF_dF_ref_dav.m#L95

It appears to me that if we go with what we have implemented already: db = du * F^{T} + F * dF^{T}
we need F stored while computing F_inv

But to go with  db = grad(du) b + b grad(du)^T, we could have stored F_inv for faster operations for the rest of formulation, but b needs F since  b = F * F^{T}

I guess the fastest as in Davydov's algorithm 2 is to store multiple things.

ArashMehraban:
And one more note: if we store multiple things for what have called "current config", we should try storing multiple things for the ref config too to determine how things change in that case.

ArashMehraban:
Important Update:

We have 4 different version of the QFunction now:
HyperFS that stores Gradu only
HyperFS that stores Gradu and C-inverse (HyperFS2)
HyperFSCUR that stores Gradu only
HyperFSCUR that stores Gradu  and F-inverse (HyperFSCUR2).

In order to compare apples to apples, in what we called current config implementation, (hyperFSCUR and hyperFSCUR2) , we used  `log1p` function instead of `log` from C. Here are the results:

degree 3: 
HyperFS that stores Gradu only : SNES time 201.37 sec
HyperFS that stores Gradu and C-inverse (HyperFS2) : SNES time 186.97 sec
HyperFSCUR that stores Gradu only : SNES time 233.44 sec
HyperFSCUR that stores Gradu  and F-inverse (HyperFSCUR2) : SNES time 193.61 sec

We update with degree 1-4 later

jeremylt:
Is the plan to reduce to only 1-2 versions in this PR once we fully understand the performance implications?

ArashMehraban:
I would recommend so

ArashMehraban:
I am not sure what we should do about the two implementations with tensors b and tau. Should we just completely throw them out or merge them? However, I feel like, after fixing the log1p guard, all these versions or whatever we want to keep are ready to be merged.

jeremylt:
I feel like we should only keep one or two versions of these QFunctions.

ArashMehraban:
I agree.

jedbrown:
Please have a look at my latest push, in which I derive the more efficient current configuration formulation in our notation and with what I hope is more clear than Davydov's paper. It should be possible to complete the commutative diagram (formulating the nonlinear problem in current configuration and linearizing to arrive at the same form). Note that this approach does not use a constitutive relation expressed in terms of the left Cauchy-Green tensor b, it just pushes the Newton linearization forward to the current configuration. It does not store a quantity like `gradu` in our present code.

Note that your implementations now are not taking advantage of symmetry in quantities like `Cinv`, but even if you did, the storage is higher by virtue of needing to store `dXdx` both (unpacked from `qdata`) and `gradu`. Can you please try applying this technique?

ArashMehraban:
@jedbrown We made an attempt at the derivation you provided:

jed-derv

The good thing about the above derivation is that we no longer need to compute `b^{-1}`. However, as you defined in your derivation,  `\nabla_x du = dF * F^{-1}`, and the qdata is computing `dF`.  Therefore, it appears to us that we do need `F^{-1}` in the action of Jcobian.

We compare this new implementation against the fastest we had with two storage:

they were both `make`-ed with `make OPT="-O3 -march=native" `

degree 3: Jed's derivation:

```mpirun -n 16 ./elasticity -problem hyperFSdav -degree 1 -mesh ./meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view```

``` SNES Type                          : newtonls
    SNES Convergence                   : CONVERGED_FNORM_RELATIVE
    Number of Load Increments          : 25
    Completed Load Increments          : 25
    Total SNES Iterations              : 100
    Final rnorm                        : 1.974894e-03
  Linear Solver:
    Total KSP Iterations               : 3112
    PC Type                            : mg
  P-Multigrid:
    PCMG Type                          : MULTIPLICATIVE
    PCMG Cycle Type                    : v
    Coarse Solve:
      KSP Type                         : preonly
      PC Type                          : gamg
  Performance:
    SNES Solve Time                    : 184.85 (184.85) sec
    DoFs/Sec in SNES                   : 7.00737 (7.00737) million
    Strain Energy                      : 1.132316029480e+07``` 

degree 3: 

`mpirun -n 16 ./elasticity -problem hyperFS2 -degree 3 -mesh ./meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view`

(this has two storage: Gradu and 6 element of C-inv matrix)

 ``` SNES:
    SNES Type                          : newtonls
    SNES Convergence                   : CONVERGED_FNORM_RELATIVE
    Number of Load Increments          : 25
    Completed Load Increments          : 25
    Total SNES Iterations              : 100
    Final rnorm                        : 1.974880e-03
  Linear Solver:
    Total KSP Iterations               : 3112
    PC Type                            : mg
  P-Multigrid:
    PCMG Type                          : MULTIPLICATIVE
    PCMG Cycle Type                    : v
    Coarse Solve:
      KSP Type                         : preonly
      PC Type                          : gamg
  Performance:
    SNES Solve Time                    : 159.612 (159.612) sec
    DoFs/Sec in SNES                   : 8.11539 (8.11539) million
    Strain Energy                      : 1.132316029480e+07```


So two important notes:
0) More storage is the fastest at the cost of memory I guess

1) computation of F_inv seems inevitable in Davydov's alg. 

2) we can attempt at storing F_inv for Davydov computaion also.

**Comment by jeremylt** [comment]
If I'm not mistaken, that comparison still uses log instead of log1p, yes?

**Comment by jeremylt** [comment]
Maybe not. I'm not sure.

**Comment by ArashMehraban** [comment]
Everything is `log1p` now.

**Comment by rezgarshakeri** [comment]
`hyperFS.h`: store 1 variable, Grad(u)
`hyperFSdav.h`: store 1 variable, Grad(u)

`hyperFS3.h`: store 3 variables, Grad(u), Cinv, constant=lambda*logJ
`hyperFSdav3.h`: store 3 variables, dXdx, tau, constant=mu - lambda*logJ

Just in case, we have 
`hyperFS2.h`: store 2 variables, Grad(u), Cinv
to see the effect of saving 2 variables vs 3 variables 

we will update with full camparison

**Comment by rezgarshakeri** [comment]
Results are as follow (note we used `make`). 
Note: hyperFS2 is there to show difference when `log1p` is computed in Jacobian as opposed to hyperFS3 that stores it.

results

**Comment by ArashMehraban** [comment]
If we review this as the final version in terms of what needs to be computed and stored, I can add the C guards for strain energy, diagnostics and `log1p` functions too

**Comment by ArashMehraban** [comment]
Note HyperFSCUR and HyperFSCUR2 will be removed!

**Comment by jedbrown** [comment]
Let's count scalar values. So `hyperFS.h` stores `dXdx` to initial configuration (9) + `gradu` (9) + quadrature weight (1) = 19.
What I think you're calling `hyperFSdav3` should be storing `dXdx` to current configuration (0) + `tau` (6) + the constant (1) + quadrature weight (1) = 17. Its advantage is mostly in doing fewer flops.

How would I repeat your test independently?

**Comment by ArashMehraban** [comment]
The tests were performed naively by just running things (not optimal). We just wanted to see how it differs.  I am gonna repeat implementing a Python code that writes bash script for reproducibility. This is what is in the abstract that depicts the storage strategy:

storage

**Comment by ArashMehraban** [comment]
Of course, When we say Cin is stored, we mean its symmetric part.

**Comment by rezgarshakeri** [comment]
> Let's count scalar values. So `hyperFS.h` stores `dXdx` to initial configuration (9) + `gradu` (9) + quadrature weight (1) = 19.
> What I think you're calling `hyperFSdav3` should be storing `dXdx` to current configuration (0) + `tau` (6) + the constant (1) + quadrature weight (1) = 17. Its advantage is mostly in doing fewer flops.
> 
> How would I repeat your test independently?

using
mpirun -n 16 ./elasticity -problem hyperFS -degree 1 -mesh ./meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view

and change hyperFS to hyperFS2, hyperFS3, hyperFSdav, hyperFSdav3 for different cases

**Comment by ArashMehraban** [comment]
@rezgarshakeri We did that for initial demonstration purposes only. I will provide a script later today to automate things like that.

**Comment by ArashMehraban** [comment]
@jedbrown out of curiosity, is it possible to use nohup on Noether, btw?

**Comment by jedbrown** [comment]
I'm not following the rationale for using `nohup`. It just keeps background jobs running after you log out or your connection is interrupted. You can run inside tmux (or screen) for better behavior after interrupts. Perhaps your intent is to get higher priority (lower `nice` level) to reduce noise from other other users. That doesn't really work in practice. We could use a resource manager (like SLURM) to avoid over-subscription and to make it possible to have exclusive jobs (though this wouldn't avoid interference from compilation and similar interactive use).


It's important to set affinity for reproducibility, especially when using fewer than all the cores. And I don't know what you've been reporting for time, but I would use `SNESSolve`.

**Comment by ArashMehraban** [comment]
Solve Time is reported through `SNESSolve` for sure. 

The last time my `ssh` pipe to Noether broke, the script had stopped running too. That's why I was asking about `nohup`.

**Comment by ArashMehraban** [comment]
@jedbrown It is not that important. These runs are not as many this time.

**Comment by jedbrown** [comment]
I just spot-checked two cases on Noether.
```
mpiexec -bind-to core -map-by socket -n 16 build/solids-elasticity -problem hyperFS -degree 3 -mesh ~arash/arash/cur-config/libCEED/examples/solids/meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view
SNESSolve             25 1.0 7.2424e+01
KSPSolve             100 1.0 5.3001e+01
```
versus new
```
mpiexec -bind-to core -map-by socket -n 16 build/solids-elasticity -problem hyperFSdav3 -degree 3 -mesh ~arash/arash/cur-config/libCEED/examples/solids/meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view
SNESSolve             25 1.0 5.9892e+01
KSPSolve             100 1.0 4.2087e+01
```

This is the sort of benefit I expected. You can spot-check some other operations, such as `VecAYPX`, `VecTDot`, `VecNorm`, and `MatSOR` to get a sense of whether a run is slow due to system noise. Those should track across all the variants.

**Comment by ArashMehraban** [comment]
That is super cool. Thanks :)

**Comment by knepley** [comment]
@jedbrown Is 'cp' really better than 'nleqerr'? It would be surprising because Deufelhard sells that as the best you can do if you have the Jacobian.

**Comment by jedbrown** [comment]
@knepley I hadn't tried, but I just did and see it needs more iterations and an additional linear solve, thus is much more expensive than CP. `diff -y` output:
```
SNESSolve             25 1.0 5.9892e+01 1.0 1.25e+10 1.5 1.0e | SNESSolve             25 1.0 1.3433e+02 1.0 3.04e+10 1.5 2.3e
SNESSetUp              1 1.0 1.3351e-05 2.8 0.00e+00 0.0 0.0e | SNESSetUp              1 1.0 1.3351e-05 3.1 0.00e+00 0.0 0.0e
SNESFunctionEval   11025 1.0 6.3268e+00 1.0 2.30e+08467647.0  | SNESFunctionEval   14325 1.0 9.1644e+00 1.0 2.91e+08592458.0
SNESJacobianEval     100 1.0 5.5277e+00 1.0 3.79e+07 2.1 9.3e | SNESJacobianEval     130 1.0 8.1158e+00 1.0 4.93e+07 2.1 1.2e
SNESLineSearch       100 1.0 8.6077e-01 1.0 2.48e+08 5.9 1.7e | SNESLineSearch       130 1.0 5.6610e+01 1.0 1.42e+10 1.5 1.0e
KSPSetUp            1200 1.0 8.0526e+00 1.0 9.47e+08 1.4 3.6e | KSPSetUp            1560 1.0 1.0476e+01 1.0 1.23e+09 1.4 4.7e
KSPSolve             100 1.0 4.2087e+01 1.0 1.08e+10 1.5 7.7e | KSPSolve             260 1.0 1.1036e+02 1.0 2.82e+10 1.5 2.0e
KSPGMRESOrthog      6000 1.0 2.4830e-01 5.4 6.64e+07 1.7 0.0e | KSPGMRESOrthog      7800 1.0 3.3019e-01 5.4 8.63e+07 1.7 0.0e
```

I recommend CP for problems like this because there really is an energy (which we could provide using `SNESSetObjective`, but haven't bothered). In other cases, an energy exists (or almost does) but might not be readily computable (due to modeling choices or regularization). CP is good for such things, though perhaps 'nleqerr' would be more robust (haven't tested).

**Review by jedbrown — COMMENTED** [review]
This is looking good to me. Could you please give a sanity check to my derivation (which now fully commutes and I think offers some insight)? I think there's still an open question of whether CRs should be formulated in initial or current configuration, but I think it's clear how to do both and what structure one might want to exploit.

I think we should rename this current configuration (the header file and its function names). It's a compact representation of Neo-Hookean in the currrrent configuration. With that and squashing of the early commits in this branch, I'd say we're ready to merge. Great work!

**Comment by ArashMehraban** [comment]
Sorry about the delay. My internet connections has been shady lately. It dies off all of a sudden.

**Comment by ArashMehraban** [comment]
@jedbrown To be clear on what to keep:

As of right now, we have 5 different implementation of HyperFS:
ref implementation:
1) original hyperFS.h
2) hyperFS2.h with two storage: Gradu and Cinv
3) hyperFS3.h with three storage: Gradu, Cinv and lamda_logJ
cur implementation: 
4) hyperFSdav.h with 1 storage (Gradu)
5) hyperFSdav3.h with 3 storage (dXdx, tau, mu-lambda_logJ)

Questions:
1) Which ones to keep? (please specify numbers from the list above) 
2) Assuming that we are keeping algorithm 3) and  5) what names do we use?
hyperFSref.h and hyperFScur.h respectively?

**Comment by rezgarshakeri** [comment]
I just changed reference to initial in `Solid mechanics mini-app` documentation

**Comment by jedbrown** [comment]
Thanks, though I think there were places where "reference frame" (as in reference frame invariance) were changed to "initial frame", which has a different meaning. Still needed:

- [ ] Rename impls
- [ ] squash to have consistent history
- [ ] comment here with updated performance table

Regarding the implementations:
> ### initial configuration
>  1. original hyperFS.h
>  2.  hyperFS2.h with two storage: Gradu and Cinv
>  3. hyperFS3.h with three storage: Gradu, Cinv and lamda_logJ
>  ### current configuration:
>  4. hyperFSdav.h with 1 storage (Gradu)
>  5. hyperFSdav3.h with 3 storage (dXdx, tau, mu-lambda_logJ)

* The accounting for storage above is not accurate. Variants 1-4 all store dXref/dxinitial and yet it just isn't counted. It doesn't exist for 5, replaced by dXref/dxcurrent. Nonsymmetric terms like Gradu and these dXdx have 9 independent values while symmetric matrices like Cinv and tau have only 6.
* I think "hyper" can be trimmed since everything we do here is hyperelastic (though Cauchy elasticity is also interesting) and it's the linear and geometrically small-strani cases that are notably different.
* I think we can drop case 2 since it's an odd middle ground and the single constant is cheap to store.
* The names should include the constitutive model. Perhaps "FSInitial-NH.h" and "FSCurrent-NH.h".

@karenlstengel has a Mooney-Rivlin model (#649 ) that we'd like to incorporate. (That branch needs a rebase/Git debugging, but the main bits are there, including some refactoring to make it easier to swap out CRs without copy/pasting the entire headers.) This branch can merge first, but it's of significant importance to efficiency how well other CRs can use the compact representation. (Cc: @diffeoinvariant)

**Comment by jedbrown** [comment]
- [ ] Fix the testing error that appeared in the rebase. 
image

**Comment by ArashMehraban** [comment]
@jedbrown 

>     * [ ]  Fix the testing error that appeared in the rebase.
>       image

@jedbrown We made all the changes in terms of name-change requests (and committed). The C-guards are still remaining to be done.  

I am not sure where the error is originating in regards to the `ncomp` from basis evaluation and element restriction since the last time we committed as of two weeks ago. Valgrind does not give any memory violation in our code either. This problem occurs with running linear elasticity code too. 


lin-elas

**Comment by jedbrown** [comment]
I see stale references to `hyperFS` still in the source and docs -- these should be made consistent.

These warnings are probably the source of the incompatibility.
```
home/jed/src/libceed/examples/solids/src/setuplibceed.c:684:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
  if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
      ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jed/src/libceed/examples/solids/src/setuplibceed.c:659:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
  if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
      ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jed/src/libceed/examples/solids/src/setuplibceed.c:623:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
  if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
      ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jed/src/libceed/examples/solids/src/setuplibceed.c:599:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
  if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
      ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jed/src/libceed/examples/solids/src/setuplibceed.c:537:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
  if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
      ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jed/src/libceed/examples/solids/src/setuplibceed.c:456:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
  if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
      ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
6 warnings generated.
```

The incompatibly sizes come here.
```c
  if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
    CeedOperatorSetField(opApply, "gradu", data[fineLevel]->ErestrictGradui,
                         data[fineLevel]->basisu, data[fineLevel]->gradu);
```
Compare this with the old code (on `main`):
```c
  if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
    CeedOperatorSetField(opApply, "gradu", data[fineLevel]->ErestrictGradui,
                         data[fineLevel]->basisu, data[fineLevel]->gradu);
```

ArashMehraban:
> I see stale references to `hyperFS` still in the source and docs -- these should be made consistent.
> 
> These warnings are probably the source of the incompatibility.
> 
> ```
> home/jed/src/libceed/examples/solids/src/setuplibceed.c:684:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:659:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:623:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:599:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:537:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:456:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 6 warnings generated.
> ```
> 
> The incompatibly sizes come here.
> 
> ```c
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>     CeedOperatorSetField(opApply, "gradu", data[fineLevel]->ErestrictGradui,
>                          data[fineLevel]->basisu, data[fineLevel]->gradu);
> ```
> 
> Compare this with the old code (on `main`):
> 
> ```c
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>     CeedOperatorSetField(opApply, "gradu", data[fineLevel]->ErestrictGradui,
>                          data[fineLevel]->basisu, data[fineLevel]->gradu);
> ```

I look into it today. What is really bothering me is that I have a libCEED build on Noether as of two weeks ago and I just copied all the modified files we touched into its `solids` folder and it compiles and runs fine!!

libceed

rezgarshakeri:
> I see stale references to `hyperFS` still in the source and docs -- these should be made consistent.
> 
> These warnings are probably the source of the incompatibility.
> 
> ```
> home/jed/src/libceed/examples/solids/src/setuplibceed.c:684:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:659:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:623:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:599:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:537:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/jed/src/libceed/examples/solids/src/setuplibceed.c:456:33: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>       ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 6 warnings generated.
> ```
> 
> The incompatibly sizes come here.
> 
> ```c
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>     CeedOperatorSetField(opApply, "gradu", data[fineLevel]->ErestrictGradui,
>                          data[fineLevel]->basisu, data[fineLevel]->gradu);
> ```
> 
> Compare this with the old code (on `main`):
> 
> ```c
>   if (problemChoice != ELAS_LIN || problemChoice != ELAS_HYPER_FSCurrent_NH2)
>     CeedOperatorSetField(opApply, "gradu", data[fineLevel]->ErestrictGradui,
>                          data[fineLevel]->basisu, data[fineLevel]->gradu);
> ```

we didn't touch those sizes, it is the same as 2 weeks ago, only changed the name `ELAS_HYPER_FSdav3` to `ELAS_HYPER_FSCurrent_NH2`. I did clone libCEED again and tried with original hyperFS and meshes in Jeremy's github and got the same error. (linearelasticity works fine)
Screenshot from 2021-03-28 11-24-38

jeremylt:
Your or statement is always true. This is causing size problems.

ArashMehraban:
> Your or statement is always true. This is causing size problems.

In case there was a logic error in our if-statements truth table, we redesigned storage strategy based on `problemChoice`. This way, based on the the `problemChoice`, we store the necessary constituents. This is more readable as apposed to the old design where we could exclude linear elasticity storage with an `if not`.   Now linear elasticity is functional again, but everything else has an issue with sizes. Also, `hyperFS` on main is not functional either. Note that the error showed up after with did 
`git pull --rebase origin main`

jeremylt:
Why duplicate those if statements? That makes the problem worse. 

Am i correct in understanding that literal doesn't have those restrictions and the size of those restrictions is otherwise different based on the problem case? Why not a switch with fall through on common cases? Or a switch to set the size in conjunction with the ifs

rezgarshakeri:
> Why duplicate those if statements? That makes the problem worse.
> 
> Am I correct in understanding that linElas doesn't have those elem restrictions and the size of those restrictions is otherwise different based on the problem case? Why not a switch with fall through on common cases? Or a switch to set the size in conjunction with the ifs.
> 
> The code on main is passing CI fine. The problem is with the logic in this PR.

yes, linElas doesn't need those restrictions. We just want to make sure the error is not because of if statements.
Would you please check with hyperFS and one of meshes on your github and run it in main. I shared screenshot above and got the same error.
(The default problem is linElas and for us also works fine)

jeremylt:
The error *is* because of the if statements. There might be other issues, but that if statement is always true.

As I said above, why not just rewrite those ifs to use a switch or something like that to use the correct size of each restriction for each nonlinear case?

jeremylt:
```
switch (problem) {
  case LinElas:
    // Nothing to do
    break;
  case HyperSS:
  case OldHyperFS:
    // Old sizes
    break;
  case NewHyperFS:
    // New sizes
    break
}
```

ArashMehraban:
> ```
> switch (problem) {
>   case LinElas:
>     // Nothing to do
>     break;
>   case HyperSS:
>   case OldHyperFS:
>     // Old sizes
>     break;
>   case NewHyperFS:
>     // New sizes
>     break
> }
> ```

Our latest commit does this idea using if statements and calling functions (duplicate in essense), but it no longer has a logic issue. The logic reads, 
if (this problem) 
then  (store so and so, etc). 
With that commit we wanted to represent that we do not a have size problem anymore.  I think what you are suggesting can be done once the compilation problem gets resolved. I am gonna focus on the version that is functional to have draft ready. We can work through this afterward. But please check what Rezgaar suggested. Do a rebase on the code in main and rerun `hyperFS`. It will not work. At least that is what we experienced.

jeremylt:
> Do a rebase on the code in main and rerun hyperFS.

Rebasing *this branch* on main and trying to build and run only shows an issue with *this branch*. Nothing changed in main that involves sizes in the solids code since the last time I ran hyperFS on main, so I still suspect a logical problem with your if statements, which is why I think they need to be rebuilt before you can blame other people.

ArashMehraban:
> > Do a rebase on the code in main and rerun hyperFS.
> 
> Rebasing _this branch_ on main and trying to build and run only shows an issue with _this branch_. Nothing changed in main that involves sizes in the solids code since the last time I ran hyperFS on main, so I still suspect a logical problem with your if statements, which is why I think they need to be rebuilt before you can blame other people.

I don't think I am blaming anyone here. I am simply asking what has changed since two weeks ago that the same code does not work on a fresh `pull` from libCEED while I have demonstrated that I can run this branch's code with libCEED from two weeks ago. Rezgar is seeing the same issue on his end too: https://user-images.githubusercontent.com/10523693/112755759-03b59e80-8f97-11eb-92cf-a291a7704669.png . Again, what you are suggesting is what we have done using if-statements. Every problem has its own sizes. Rezgar demonstrated that the `hyperFS` code from main does not work on a fresh pull from libCEED: https://user-images.githubusercontent.com/42816410/112761432-8dc22f00-8fb8-11eb-8a84-2c113df9e869.png.  
Give me a day or two. I see what I can do. Switching `if-else` to `switch` is not  a big deal. I'll do it once I get the paper going to a reasonable state.

jeremylt:
It is possible that the error checking PR found some long standing small bug from when we first wrote it. The fix is still exactly the same - carefully refactor the if logic to ensure that every single restriction, basis, and QFunction field size agree in every case.

jedbrown:
Oh my, I misquoted (pasted twice from the branch rather than once from `main`) in https://github.com/CEED/libCEED/pull/708#issuecomment-808840000 and I fear that may have caused some of the confusion here. The actual quote from `main` is
```c
 if (problemChoice != ELAS_LIN)
    CeedOperatorSetField(opJacob, "gradu", data[fineLevel]->ErestrictGradui,
                         CEED_BASIS_COLLOCATED, data[fineLevel]->gradu);
```
see a9375e40db51a39b05e4c102ed1987ef5cf70ccb for how I fixed this branch (just one case), with a working example run. It's the same issue and the mistake was inserted at some point when refactoring this branch.

ArashMehraban:
> `gradu` was previously defined to have `CEED_EVAL_NONE` so it can't be evaluated using `basisu`. We ccould have a better error message.

Thanks Jed! He is gonna commit it now. It fixed the issue.

ArashMehraban:
> Oh my, I misquoted (pasted twice from the branch rather than once from `main`) in [#708 (comment)](https://github.com/CEED/libCEED/pull/708#issuecomment-808840000) and I fear that may have caused some of the confusion here. The actual quote from `main` is
> 
> ```c
>  if (problemChoice != ELAS_LIN)
>     CeedOperatorSetField(opJacob, "gradu", data[fineLevel]->ErestrictGradui,
>                          CEED_BASIS_COLLOCATED, data[fineLevel]->gradu);
> ```
> 
> see [a9375e4](https://github.com/CEED/libCEED/commit/a9375e40db51a39b05e4c102ed1987ef5cf70ccb) for how I fixed this branch (just one case), with a working example run. It's the same issue and the mistake was inserted at some point when refactoring this branch.

I was wondering about that :) I am glad this is resolved.

ArashMehraban:
Test run on Noether with the latest commit:

`mpirun -n 16 ./elasticity -problem FSCurrent-NH2 -degree 2 -mesh ./meshes/Tube8.exo -nu 0.3 -E 1e8 -bc_clamp 998,999 -bc_clamp_998_translate 0,0,-1 -num_steps 25 -snes_linesearch_type cp -snes_rtol 1e-7 -snes_monitor -log_view`

```-- Elasticity Example - libCEED + PETSc --
  libCEED:
  PETSc:
    PETSc Vec Type                     : mpi
  Problem:
    Problem Name                       : Hyperelasticity finite strain Current config w/ dXref_dxcurr, tau, constant storage
    Forcing Function                   : None
  Mesh:
    File                               : ./meshes/Tube8.exo
    Number of 1D Quadrature Points (q) : 3
    Global nodes                       : 42946
    Owned nodes                        : 2289
    DoF per node                       : 3
  Multigrid:
    Type                               : P-multigrid, logarithmic coarsening
    Number of Levels                   : 2
    Level 0 (coarse):
      Number of 1D Basis Nodes (p)     : 2
      Global Nodes                     : 6048
      Owned Nodes                      : 281
    Level 1 (fine):
      Number of 1D Basis Nodes (p)     : 3
      Global Nodes                     : 42946
      Owned Nodes                      : 2289
0 Load Increment
  0 SNES Function norm 8.887954317074e+06 
  1 SNES Function norm 1.784419878717e+06 
  2 SNES Function norm 3.524986694396e+04 
  3 SNES Function norm 1.836014145202e+01 
  4 SNES Function norm 4.210081628912e-06 
1 Load Increment
  0 SNES Function norm 8.844365468773e+06 
  1 SNES Function norm 1.757548987603e+06 
  2 SNES Function norm 3.412597284404e+04 
  3 SNES Function norm 1.728520744365e+01 
  4 SNES Function norm 3.713044715662e-06

**Comment by ArashMehraban** [comment]
I can work on the paper now. This week, I will provide:

0) C-guards
1) `switch` statement per @jeremylt 's request.  

Will do this week.

Thanks everyone @rezgarshakeri @jedbrown @jeremylt

**Comment by ArashMehraban** [comment]
BTW, if I do not `make` the `Add_ftr_cur_config` branch and run with `-problem FSCurrent-NH2` in example bove, it does not recognize the problem type and PETSc throws this error with a suggestion which I do not think is in `cloptions.c`:

petsc-error

The suggestions should change to `linElas`, `hyperSS`, `FSInitial-NH1`, `FSInitial-NH2`, `FSCurrent-NH1`, `FSCurrent-NH2`

**Comment by jedbrown** [comment]
> BTW, if I do not `make` the `Add_ftr_cur_config` branch and run with `-problem FSCurrent-NH2` in example bove, it does not recognize the problem type and PETSc throws this error with a suggestion which I do not think is in `cloptions.c`:
> 
> petsc-error
> 
> The suggestions should change to `linElas`, `hyperSS`, `FSInitial-NH1`, `FSInitial-NH2`, `FSCurrent-NH1`, `FSCurrent-NH2`

I'm not sure what you're asking for. If you run with a stale version of the code, then of course it doesn't recognize current options. I think we should change `linElas` to `Linear` and `HyperSS` to `SS-NH` for "small strain neo-Hookean". Any documentation referencing the old names needs to be updated.

**Comment by jedbrown** [comment]
Note: I pushed some better checks so if you were to run with an erroneous version of the code, you get a more informative error.
```
$ ./elasticity -nu 0.3 -E 10 -degree 1 -dm_plex_box_faces 3,3,3 -problem FSInitial-NH2 -num_steps 5 -bc_clamp 3 -bc_traction 4 -bc_traction_4 0,0,-1 -snes_monitor -view_soln
/home/jed/src/libceed/interface/ceed-operator.c:123 in CeedOperatorCheckField(): Field 'gradu' configured with CEED_EVAL_NONE must be used with CEED_BASIS_COLLOCATED
[joule:227912] *** Process received signal ***
[joule:227912] Signal: Aborted (6)
[joule:227912] Signal code:  (-6)
[joule:227912] [ 0] /usr/lib64/libpthread.so.0(+0x13960)[0x7f690b0b5960]
[joule:227912] [ 1] /usr/lib64/libc.so.6(gsignal+0x145)[0x7f690af0fef5]
[joule:227912] [ 2] /usr/lib64/libc.so.6(abort+0x116)[0x7f690aef9862]
[joule:227912] [ 3] /home/jed/src/libceed/lib/libceed.so(+0x1318b)[0x7f690b34318b]
[joule:227912] [ 4] /home/jed/src/libceed/lib/libceed.so(CeedErrorImpl+0x9d)[0x7f690b3438fd]
[joule:227912] [ 5] /home/jed/src/libceed/lib/libceed.so(CeedOperatorSetField+0x2c1)[0x7f690b34a241]
[joule:227912] [ 6] ./elasticity(+0x219e3)[0x557f93ce29e3]
[joule:227912] [ 7] ./elasticity(+0x871c)[0x557f93cc971c]
[joule:227912] [ 8] /usr/lib64/libc.so.6(__libc_start_main+0xd5)[0x7f690aefab25]
[joule:227912] [ 9] ./elasticity(+0xda8e)[0x557f93ccea8e]
[joule:227912] *** End of error message ***
```

ArashMehraban:
> > BTW, if I do not `make` the `Add_ftr_cur_config` branch and run with `-problem FSCurrent-NH2` in example bove, it does not recognize the problem type and PETSc throws this error with a suggestion which I do not think is in `cloptions.c`:
> > petsc-error
> > The suggestions should change to `linElas`, `hyperSS`, `FSInitial-NH1`, `FSInitial-NH2`, `FSCurrent-NH1`, `FSCurrent-NH2`
> 
> I'm not sure what you're asking for. If you run with a stale version of the code, then of course it doesn't recognize current options. I think we should change `linElas` to `Linear` and `HyperSS` to `SS-NH` for "small strain neo-Hookean". Any documentation referencing the old names needs to be updated.

I was just trying to be informative about the problem options list now that we have changed the problem names. If the problem options message in PETSc error message gets read from our `problemlTypes` array, then this issue gets fixed once we merge. But if the error message is hand-written, then that should be updated also. Since it caught my eye, I thought I communicate it for completeness.

jedbrown:
Okay, yes. That is just written documentation that needs to be updated as the code is updated.

jedbrown:
Regarding the `switch` statement, it would be much better to make the details of collocated fields passed between residual and Jacobian declarative in `problemOptions` (or even factor these into independent compilation units). Replacing the chained `if` statements with a `switch` will only make a small difference to maintainability.

jeremylt:
I was thinking about that. I agree that replacing the if chain directly with a switch wouldn't be good. I was thinking more of using the switch to determine sizes [and emodes] (or store them in the big problems options struct) and clean up the logic to use the same control flow for all problems. It shouldn't hurt linElas performance to have an extra size zero field.

jeremylt (approved):
Ready after one more pass through the documentation

ArashMehraban:
I think the code can drop `FSInitial-NH2` implementation too. I have not `perf`-ed but all the data I am getting from the runs say, the tradeoff between storage and time for `FSInitial-NH2` is not worth it. Like I said, all of our original tests were in the realm of small strain simulations where I had bent a 100mm tube by 2mm instead of 20mm.. The following are runs with a 100mm tube mesh bent by 20mm  in 25 load increments while one side is clamped and the other end is pulled down in the -Y direction (both Dirichlet BC). The same mesh shows  38K, 277K, 906K and and 2.1M dofs in PETSc for poly order 1-4:

Algorithm 1: `FSInitial-NH1` 
Algorithm 2:  `FSInitial-NH2` 
Algorithm 3: `FSCurrent-NH1` 
Algorithm 4:  `FSCurrent-NH2`

bend

ArashMehraban:
BTW, every problem above was run 3 times and average was computed for the table

Issue #700, opened 2021-01-28
CeedRequestWait is documented but not implemented


FreddieWitherden:
I am currently reviewing libCEED as part of https://github.com/openjournals/joss-reviews/issues/2945.

The method CeedRequestWait is documented and seems extremely interesting.  However, looking through the source it does not appear to be actually implemented.  It might be prudent to relegate this method and associated paraphernalia to a feature branch until it is ready for mainline.  If nothing else it may be that when implementing that that a blocker raises its head or a simpler approach shows itself.

jedbrown:
This is a good point. We envisioned this as a consistent way to handle CPU threads and GPU streams, but now I'm less confident that this is the right interface (at least for GPUs). We've recently been experimenting in PETSc with offloading the scalar math for Krylov methods, such that the host doesn't need to block. We might want to switch to a stream model, something like `CeedOperatorApplyWithStream(CeedOperator op, CeedVector in, CeedVector out, Stream s)`, which would return as soon as the operation is enqueued, and thus support hiding kernel launch latency. Would you prefer this model? Other ideas?

Cc: @YohannDudouit @nbeams

FreddieWitherden:
I think the community is slowly moving away from streams/queues and towards more declarative/graph based APIs.  Good examples here are the (relatively recent) CUDA graph management APIs and the command list approach taken by Intel's Level Zero (which is probably the gold standard in terms of HPC API design at the moment, albeit being somewhat low level).

The difficulty I've always had with designing these kind of interfaces (and we have our own queues in PyFR) is how to handle CPUs.  If we execute the kernel immediately from the API then the call is not asynchronous (whereas it would be on a GPU if you launched it in a stream).  This confuses users who expect you to return immediately so they can go off and do some I/O while waiting for the kernel to finish.  If you defer execution until a later wait/sync call then you may catch out some users who expect forward progress to be made. (MPI libraries often run into this problem with non-blocking communication where if you're not probing/waiting then nothing happens.)  Finally, if you spawn a thread you may and do the work there you're increasing overhead and may be considered a bad citizen.

The problem is even more fiddly when you're a library and need to integrate with the models of a large number of stakeholders.

nbeams:
I agree this is a tricky issue for a library like libCEED, which may need to use various task/execution queue models (through backends targeting different architecture), while also potentially being used by other libraries/users. 

[...]

PR #712, opened 2021-03-11
Full assembly of CeedOperator

barker29:
This PR introduces `CeedOperatorLinearFullAssemble()`, which is able to produce the full assembled matrix for fairly general `CeedOperator` objects. The output format is `(row, column, value)` triples which are intended to be added (not set) into whatever (probably sparse) matrix format the client code would like. This implementation is not intended to be very fast, and is a bad idea in general for high-order operators, but it may be useful in p-multigrid type algorithms where you may want to assemble and use traditional AMG when you get to low-order.

This also introduces tests `t560-operator.c` through `t565`, which correspond respectively to `t533` through `t538`. Note I have made no attempt to duplicate the Fortran versions of these tests, however.

I expect this will need some polishing and updating, but I've gotten about as far as I can with it by myself - I'm happy to do more but I may need some guidance from the libCEED experts.

### DIFF
--- a/include/ceed.h
+++ b/include/ceed.h
@@ -594,6 +594,8 @@ CEED_EXTERN int CeedOperatorLinearAssemblePointBlockDiagonal(CeedOperator op,
     CeedVector assembled, CeedRequest *request);
 CEED_EXTERN int CeedOperatorLinearAssembleAddPointBlockDiagonal(CeedOperator op,
     CeedVector assembled, CeedRequest *request);
+CEED_EXTERN int CeedOperatorLinearFullAssemble(CeedOperator op,
+    CeedInt *nentries, CeedInt **rows, CeedInt **cols, CeedScalar **vals);
 CEED_EXTERN int CeedOperatorMultigridLevelCreate(CeedOperator opFine,
     CeedVector PMultFine, CeedElemRestriction rstrCoarse, CeedBasis basisCoarse,
     CeedOperator *opCoarse, CeedOperator *opProlong, CeedOperator *opRestrict);
--- a/include/ceed.h
+++ b/include/ceed.h
@@ -594,8 +594,9 @@ CEED_EXTERN int CeedOperatorLinearAssemblePointBlockDiagonal(CeedOperator op,
     CeedVector assembled, CeedRequest *request);
 CEED_EXTERN int CeedOperatorLinearAssembleAddPointBlockDiagonal(CeedOperator op,
     CeedVector assembled, CeedRequest *request);
-CEED_EXTERN int CeedOperatorLinearFullAssemble(CeedOperator op,
-    CeedInt *nentries, CeedInt **rows, CeedInt **cols, CeedScalar **vals);
+CEED_EXTERN int CeedOperatorLinearAssembleSymbolic(CeedOperator op,
+    CeedInt *nentries, CeedInt **rows, CeedInt **cols);
+CEED_EXTERN int CeedOperatorLinearAssemble(CeedOperator op, CeedVector values);
 CEED_EXTERN int CeedOperatorMultigridLevelCreate(CeedOperator opFine,
     CeedVector PMultFine, CeedElemRestriction rstrCoarse, CeedBasis basisCoarse,
     CeedOperator *opCoarse, CeedOperator *opProlong, CeedOperator *opRestrict);
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -296,6 +296,37 @@ int CeedOperatorSingleView(CeedOperator op, bool sub, FILE *stream) {
 }
 
 /**
+  @brief Find the active vector ElemRestriction for a CeedOperator
+
+  @param[in] op            CeedOperator to find active basis for
+  @param[out] activerstr   ElemRestriction for active input vector
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Utility
+**/
+static int CeedOperatorGetActiveElemRestriction(CeedOperator op,
+    CeedElemRestriction *activerstr) {
+  *activerstr = NULL;
+  for (int i = 0; i < op->qf->numinputfields; i++)
+    if (op->inputfields[i]->vec == CEED_VECTOR_ACTIVE) {
+      *activerstr = op->inputfields[i]->Erestrict;
+      break;
+    }
+
+  if (!*activerstr) {
+    // LCOV_EXCL_START
+    int ierr;
+    Ceed ceed;
+    ierr = CeedOperatorGetCeed(op, &ceed); CeedChk(ierr);
+    return CeedError(ceed, 1,
+                     "No active ElemRestriction found!");
+    // LCOV_EXCL_STOP
+  }
+  return 0;
+}
+
+/**
   @brief Find the active vector basis for a CeedOperator
 
   @param[in] op            CeedOperator to find active basis for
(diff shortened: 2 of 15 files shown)


jeremylt:
We should chat. I'd actually push this into the backend code right next to the diagonal assembly code. There'll be less duplication that way.

Setting up the Fortran interface isn't too bad and I can help with that too.

include/ceed.h
```diff
@@ -530,6 +530,8 @@ CEED_EXTERN int CeedOperatorLinearAssemblePointBlockDiagonal(CeedOperator op,
     CeedVector assembled, CeedRequest *request);
 CEED_EXTERN int CeedOperatorLinearAssembleAddPointBlockDiagonal(CeedOperator op,
     CeedVector assembled, CeedRequest *request);
+CEED_EXTERN int CeedOperatorLinearFullAssemble(CeedOperator op,
+    CeedInt *nentries, CeedInt **rows, CeedInt **cols, CeedScalar **vals);
```
jedbrown, line 534: Would you be willing to split this into `CeedOperatorLinearAssembleSymbolic(CeedOperator, size_t *nentries, CeedInt **rows, CeedInt **cols)` and `CeedOperatorLinearAssemble(CeedOperator, CeedVector values)`?

My rationale is that we often assemble multiple times using the same nonzero pattern. Many backends can use the same Symbolic code. The indices can live on the preferred `CeedMemType` or we can have another argument to specify where they go. We support this split phase COO assembly on GPUs in PETSc and one can always call both functions if they need both.

If we're going to store the entries in a `CeedVector` (which seems sensible because it conveniently works with both host and device) then its constructor `CeedVectorCreate` needs to use `size_t` instead of `CeedInt`.

jeremylt: I was going to chat with @barker29 on Monday about possible directions for this code. It sounds like maybe all three of us should chat?

jedbrown: Sure, though I suspect the interface can be fleshed out here. @stefanozampini added the split-phase assembly to PETSc, and is familiar with MFEM.

jeremylt: My concern is less on the interface is and more in the internals. As is, this PR introduces code duplication from our (albeit complicated) code for diagonal assembly. We pushed that code into the backed 'families' for performance and differences in qdata layout. I'm trying to understand this intended use case to see if the performance concerns apply here. I think the diagonal assembly code can be extended to do this without the duplication, but I'm still fuzzy on how this data will be consumed.

barker29: I can certainly split the function into two as @jedbrown suggests, though that will require some significant changes to the current implementation. Using a `CeedVector` for the values is easy.

@jeremylt 's concerns about code duplication with the diagonal assembly are a bit harder for me to understand what's actually wanted but we can hash that out in our meeting today.

barker29:
I pushed this branch to the main repo (not a fork) but I can't find a way to transfer the discussion here to a new PR (the first couple google hits suggest this is not possible). Should I just open a new PR with the non-fork branch?

jedbrown:
Ah, yeah. The interface won't let you change the source branch, only the destination. (Looks like CI ran here on this branch, maybe because you're a member now? So maybe you don't need a new PR.)

barker29:
Based on our meeting yesterday, my understanding of remaining tasks is this:
- Change interface to separate "symbolic" and numeric assembly [done]
- Change `CeedVector` to use a `size_t` for its size instead of `CeedInt` (not in this PR)
- Add hook so backends that have their own assembly routines can use them [TODO]
- Add Fortran versions of tests `t560` through `t565` [TODO]
- Add some kind of `CeedIntArray` object to manage memory for ints like `CeedVector` does for scalars (not in this PR)
- Reduce code duplication between diagonal assembly and full assembly (not in this PR?)

Am I missing anything?

jeremylt:
Re the backend function - three steps
1) add function pointers to CeedOperator_private struct just below line 300 in `ceed-impl.h`
2) add function names to CEED_FTABLE just below line 697 in `ceed.c`
3) This is the part I'm going to have to think about. Will this work?
```
  // Use backend version, if available
  if (op->LinearAssemble) {
    return op->LinearAssemblel(op, assembled, request);
  } else {
    // Check for valid fallback resource
    const char *resource, *fallbackresource;
    ierr = CeedGetResource(op->ceed, &resource); CeedChk(ierr);
    ierr = CeedGetOperatorFallbackResource(op->ceed, &fallbackresource);
    if (strcmp(fallbackresource, "") && strcmp(resource, fallbackresource)) {
      // Fallback to reference Ceed
      if (!op->opfallback) {
        ierr = CeedOperatorCreateFallback(op); CeedChk(ierr);
      }
      // Assemble
     return op->opfallback->LinearAssemble(op->opfallback, assembled,
             request);
    }
  }
// Otherwise this function wasn't written for the fallback and we're using the interface level code you wrote
```
I might be missing some things we should check?

jeremylt:
For the Fortran interface, lines 1115-1142 of `ceed-fortran.c` can be replicated for the new function.

jeremylt:
Since I'll be on vacation next week, there may not be enough time for me to refactor the diagonal assembly code before the release. I'd be fine with pushing that to a follow-up PR.

jedbrown:
Sounds good to me. I'm honestly not that concerned about full coverage from Fortran tests. (The Fortran interface is crippled to make it usable from Nek5000 (Fortran 77 only), but that team is moving to NekRS (C++) anyway, so our Fortran should probably be modernized. I see it as being in limbo until a prospective user comes along, and don't think it's worth our spending a lot of time on it.)

I'm okay with raw index arrays in this PR, which we can refactor later. I don't think it'll be a big deal, especially before we have a backend that does symbolic assembly on the device.

jeremylt:
Do we want to add this to Python, Julia, and Rust here or in a follow-up?

jedbrown:
I'd expect the language bindings will change a bit once we abstract integer arrays so I'd be okay to leave it. The C interface will be enough for us to use it in the examples (like coarse grid assembly in multigrid.c and solids). That exact assembly should be more efficient and accurate than finite differencing.

barker29:
I added code to allow potential backends for full assembly as Jeremy suggested - without actual backends, I'm not sure how to test that this is working properly. And based on Jed's comments I am going to ignore Fortran in this PR. 

That means everything on my list is either done or delayed - let me know if there's more to do to check the backends, or if you do want the Fortran interface put in in this PR, or anything else.

interface/ceed-operator.c
```diff
@@ -1180,6 +1211,543 @@ int CeedOperatorLinearAssembleAddPointBlockDiagonal(CeedOperator op,
...
+   Given m by n matrix A and n by p matrix B, add the matrix-matrix
+   product AB into the existing matrix C.
+
+   All matrices are stored in column-major order.
+
+   This could, probably should, be a GPU kernel.
+
+   @ref Utility
+**/
+static int MatrixMatrixMultiply(CeedInt m, CeedInt n, CeedInt p,
```
jeremylt, line 1226: See `CeedMatrixMultiply` in `ceed-basis.c`. I think if that function signature is added to `ceed-impl.h`, then we don't have to repeat this function here.

interface/ceed-operator.c
```diff
@@ -1180,6 +1211,543 @@ int CeedOperatorLinearAssembleAddPointBlockDiagonal(CeedOperator op,
...
+                                           (compout)*layout_er[1] + e*layout_er[2];
+            const CeedInt edof_index_col = (j)*layout_er[0] +
+                                           (compin)*layout_er[1] + e*layout_er[2];
+
+            const CeedInt row = elem_dof_a[edof_index_row];
+            const CeedInt col = elem_dof_a[edof_index_col];
+
+            rows[offset + count] = row;
+            cols[offset + count] = col;
+            // vals[offset + count] = val;
```
jeremylt, line 1307: dead code?

interface/ceed-operator.c
barker29: I've done this, reducing code duplication (which is good) but somehow I'm not sure the function signature really belongs in `ceed-impl.h` which only contains objects and not functions. Let me know if there's a better place for it.

jeremylt: Yeah, I think you're right. We could put it in `ceed-backend.h`. That has a lot of utility functions we don't want users accessing.

barker29: So apparently it was already in `ceed-backend.h` but I didn't notice it...

jeremylt:
This change will not work. A fallback ceed needs to be created if the backend does not have its own impl and is not itself a fallback resource. This change just bypasses the fallback if it hasn't yet been created for another reason.

The logic I wrote for detecting when a fallback does not have its own implementation and the interface implimentation needs to be used has to have some flaw.

The purpose of the fallback is to provide a single backend implementation for a common set of backends, such as all CUDA backends, while still allowing each backend to still have independent logic for the majority of the operator methods.

barker29:
I figured as much, but I was getting a segfault in the `CeedOperatorCreateFallback()` call with CUDA, and was having trouble debugging it.

jedbrown:
Is there a specific failure you'd like help debugging?

barker29:
Yes, please, the `codecov/project/backends`. It's probably missing `LCOV_EXCLUDE` around a `CeedError`, but I get an `Unauthorized` error on codecov when I try to investigate. (I think the rust failure is just a one-off and re-running it will be fine?)

barker29:
Regarding the fallback stuff, I may misunderstand but I think there are more cases than your sketch above takes care of:
1. backend implements `LinearAssemble`: call `op->LinearAssemble()`.
2. backend doesn't, fallback resource is `""`: use interface-level code
3. backend doesn't, fallback resource is whatever, say `"/cpu/self/ref/serial"`, and that fallback implements `LinearAssemble`: call `op->opfallback->LinearAssemble()`
4. backend doesn't, fallback resource is whatever, fallback does not implement `LinearAssemble`: I think your suggestion is missing this case, and in fact if I implement the changes you suggest this is where we end up (the resource isn't going to end up as `""` in practice even if that's the default).

The fix I see is to add an `if (op->opfallback->LinearAssemble)` to your logic [above](https://github.com/CEED/libCEED/pull/712#issuecomment-800461780). What I don't like about this is you end up creating a fallback resource that you don't use, but maybe that's not a big deal.

Regarding `CEED_ERROR_MAJOR` and friends, I need to merge or rebase `main` in here to make that compile. The contributing guidelines seem to discourage *both* merging and rebasing, which is preferred?

jeremylt:
Case 4 should work fine - it's what's happening with the CPU side right now and the CPU tests pass

jeremylt:
I almost always rebase off main unless the PR is too complex and old to support that easily. I don't read the contributing guide as discouraging that workflow, but maybe I'm missing something?

jedbrown:
It's best to do neither if there are no known interactions with 'main', any tests you may have run on individual commits are more likely to work (no spurious bisection failures) if you just leave them be than if you rebase. If there is a known interaction with 'main', then you can either rebase or merge. It's usually better to rebase if your PR is still in draft mode. If the interaction is complicated over several commits or if there are already users of your in-progress branch, then merging can be better.

The main thing we're trying to avoid is confusing history in which there are in-progress commits followed by merges from 'main' followed by more commits that repair the old stuff. Making anything coherent and testable/bisectable is hard in that setting. Reducing cognitive load on reviewers and future debuggers is the goal.

barker29:
@jeremylt One of us is confused. The code you suggest [here](https://github.com/CEED/libCEED/pull/712#issuecomment-800461780) does not work on CPU, it segfaults as I just verified locally. There is no place in the current code where any backend has `LinearAssembleSymbolic` or `LinearAssemble` defined, so calling them doesn't make sense. 

I think you're suggesting that `/cpu/self/ref/serial` should define `op->LinearAssemble` to point to `CeedOperatorLinearAssemble`, ie by adding a line in `CeedOperatorCreate_Ref`. I tried this, but then once you get to that resource the first check `if (op->LinearAssemble)` is true, so the function just calls itself infinitely becoming effectively: `int CeedOperatorLinearAssemble { return CeedOperatorLinearAssemble(); }`.

I'm not trying to be difficult here, I really don't know what you have in mind for the fallback logic.

jeremylt:
> I think you're suggesting that /cpu/self/ref/serial should define op->LinearAssemble to point to CeedOperatorLinearAssemble, ie by adding a line in CeedOperatorCreate_Ref

Not at all. I'll play around with forking your fork and show you what I mean.

jeremylt:
I need to actually test it and debug, but this branch has what I'm thinking of:
https://github.com/jeremylt/libCEED/tree/barker29/operator-assembly

jeremylt:
The fallback works on CUDA locally for me now.

interface/ceed-operator.c
```diff
@@ -1264,22 +1275,525 @@ int CeedOperatorLinearAssemblePointBlockDiagonal(CeedOperator op,
...
+}
+
+/**
+   @brief Build nonzero pattern for non-composite operator.
+
+   Users should generally use CeedOperatorLinearAssembleSymbolic()
+
+   @ref Developer
+**/
+int CeedSingleOperatorAssembleSymbolic(CeedOperator op, CeedInt offset,
```
jeremylt, line 1301: I think it would help if we clearly indicate in the function name if we are summing into a target vector or overwriting. This will help when I refactor the diagonal assembly to have an interface level impl with some reduced code duplication.

See the `/cpu/self/ref` impl of diagonal assembly for context:

<details><summary>Click to Show/Hide</summary>

```
// Assemble composite diagonal common code
static inline int CeedOperatorLinearAssembleAddDiagonalCompositeCore_Ref(
  CeedOperator op, CeedVector assembled, CeedRequest *request,
  const bool pointBlock) {
  int ierr;
  CeedInt numSub;
  CeedOperator *subOperators;
  ierr = CeedOperatorGetNumSub(op, &numSub); CeedChkBackend(ierr);
  ierr = CeedOperatorGetSubList(op, &subOperators); CeedChkBackend(ierr);
  for (CeedInt i = 0; i < numSub; i++) {
    ierr = CeedOperatorAssembleAddDiagonalCore_Ref(subOperators[i], assembled,
           request, pointBlock); CeedChkBackend(ierr);
  }
  return CEED_ERROR_SUCCESS;
}

// Assemble Linear Diagonal
static int CeedOperatorLinearAssembleAddDiagonal_Ref(CeedOperator op,
    CeedVector assembled, CeedRequest *request) {
  int ierr;
  bool isComposite;
  ierr = CeedOperatorIsComposite(op, &isComposite); CeedChkBackend(ierr);
  if (isComposite) {
    return CeedOperatorLinearAssembleAddDiagonalCompositeCore_Ref(op, assembled,
           request, false);
  } else {
    return CeedOperatorAssembleAddDiagonalCore_Ref(op, assembled, request, false);
  }
}

// Assemble Linear Point Block Diagonal
static int CeedOperatorLinearAssembleAddPointBlockDiagonal_Ref(CeedOperator op,
    CeedVector assembled, CeedRequest *request) {
  int ierr;
  bool isComposite;
  ierr = CeedOperatorIsComposite(op, &isComposite); CeedChkBackend(ierr);
  if (isComposite) {
    return CeedOperatorLinearAssembleAddDiagonalCompositeCore_Ref(op, assembled,
           request, true);
  } else {
    return CeedOperatorAssembleAddDiagonalCore_Ref(op, assembled, request, true);
  }
}
```

<\details>

jeremylt:
I thought that I was pushing to a copy of your branch on the libCEED repo, but somehow I was pushing to your fork? I'm not sure how that happened, but it's making a mess of CI. Sorry about that.

I kind of want to break out the interface level implementations of these advanced operator functions (QF assembly, operator/diagonal assembly) into their own file so they don't overwhelm the code in this file for basic use and management of operator objects, but that would be in a follow-up PR.

interface/ceed-operator.c
barker29: We are kind of "appending" to the target vector, so for a composite operator you get entries in the `values` vector for all the nonzero entries in each of the sub-operators. I'll try to update the documentation to make this clearer - are there names for the functions that would also make this clear?

jeremylt: With diagonal assembly, we use `Assemble` for functions that overwrite the target vector and `AssembleAdd` for functions that sum into the target vector. I set it up this way to support MFEM's ability to compose diagonal assembly for multiple integrators implemented in libCEED.

I think we should
1) add `Add` to the name of helper functions that sum into their vectors
2) make `Apply` and `ApplyAdd` versions of full assembly function

jeremylt:
CI failing on Noether it looks like. Investigating.

jeremylt:
@jedbrown, should we add a `request` to the args for these assembly functions?

interface/ceed-operator.c
barker29: I understand that. This is a different case from diagonal assembly and I don't think those conventions apply here. `Add` variants do not make sense here because there is no way to ensure that you are adding into the same nonzero structure, so what we do is *append* the values to a coordinate-format-like data structure, and then the client code (eg MFEM) can assemble or add or do whatever it wants.

jeremylt: So down the line MFEM will be obligated to add together the separately assembled mass and diffusion operators? I'm fine with making that MFEM's problem rather than libCEED's.

I still think it would be valuable to indicate in the name when you overwrite vs *append*. If you prefer the word *append*, go for it. I'll figure out how that fit with the future refactoring of assembly/diagonal assembly at the interface level when I get there.

jeremylt:
CI issue on Noether should be fixed now.

jedbrown:
Regarding "request", I think there is some question of what the model should be. https://github.com/CEED/libCEED/issues/700
Streams are the easy way, though a more general graph structure would be better. Perhaps our functions should take a l list of requests (operations on which it depends) and return a request. If all the input requests are on the same stream, we just enqueue on that stream. If they're on different streams (or the backend has no concept of streams) then it would use a graph abstraction or (if no such abstraction exists) block on the input requests. These semantics seem really noisy in C.

So a request is logically correct in the hypothetical model where it was conceived, but that concept may need to be revised in light of current thinking on async execution. I don't feel the need to add it in this PR, and discussion+refactoring post-release.

jedbrown (approved):
This looks good to me. I think it's useful and usable now, and we can improve internals in future PRs post-release.

jeremylt:
This is in that funky state where only you can merge @jedbrown without rebasing off main again

PR #714, opened 2021-03-17
Fluids & Solids | add/modify cl options for viz output folder

LeilaGhaffari:


examples/fluids/navierstokes.c
```diff
@@ -1149,7 +1149,7 @@ int main(int argc, char **argv) {
     CHKERRQ(ierr);
   }
   ierr = PetscStrncpy(user->outputfolder, ".", 2); CHKERRQ(ierr);
```
LeilaGhaffari, line 1151: What is this line for? I can't understand the purpose here!

jedbrown: It's setting the default output to the current directory. A file name formatted with `"%s/%s.vtu"` thus becomes `./*.vtu` instead of `some-dir/*.vtu`.

An alternative would be to initialize it blank and ensure that if a directory is specified, it ends with `/`. Then format with `"%s%s.vtu"` and you'll get functionally the same behavior with `*.vtu` and `some-dir/*.vtu`.

examples/fluids/navierstokes.c
```diff
@@ -1149,7 +1149,7 @@ int main(int argc, char **argv) {
...
-  ierr = PetscOptionsString("-of", "Output folder",
+  ierr = PetscOptionsString("-output_dir", "Output folder",
```
jedbrown, line 1152:
```suggestion
  ierr = PetscOptionsString("-output_dir", "Output directory",
```

examples/solids/src/cloptions.c
```diff
@@ -38,6 +38,10 @@ PetscErrorCode ProcessCommandLineOptions(MPI_Comm comm, AppCtx appCtx) {
                             sizeof(appCtx->ceedResource), &ceedFlag);
   CHKERRQ(ierr);
 
+  ierr = PetscOptionsString("-output_dir", "Output folder",
```
jedbrown, line 41:
```suggestion
  ierr = PetscOptionsString("-output_dir", "Output directory",
```

examples/fluids/navierstokes.c
LeilaGhaffari: Ah, it makes sense! Thanks. What we have right now looks good. I added the same line to the solids example as well.

LeilaGhaffari:
Thanks, @jedbrown for the suggestions. Have I addressed all of them? I kept push forcing and I am not sure if I noticed everything.

examples/solids/elasticity.h
```diff
@@ -108,6 +108,7 @@ typedef struct AppCtx_private *AppCtx;
 struct AppCtx_private {
   char          ceedResource[PETSC_MAX_PATH_LEN];     // libCEED backend
   char          meshFile[PETSC_MAX_PATH_LEN];         // exodusII mesh file
+  char          outputfolder[PETSC_MAX_PATH_LEN];
```
valeriabarra, line 111: Thanks Leila. 
Just for consistency, you could rename this variable `outputdir` or `outputdirectory` now, but it's really up to you

jeremylt (approved):
LGTM

jeremylt:
Yeah, squash+merge makes sense to me here. Sometimes with bigger PRs I will try to manually squash into a couple of commits that represent distinct changes, but for smaller ones I tend to squash it all together in the merge.

jedbrown (approved):

PR #715, opened 2021-03-19
Rey/kershaw

reykoki:


examples/petsc/multigrid.c
```diff
@@ -148,6 +173,44 @@ int main(int argc, char **argv) {
       dmorig = dmDist;
     }
   }
+  {
```
jeremylt, line 176: Should this be activated via a command line option?

examples/petsc/multigrid.c
```diff
@@ -578,7 +641,7 @@ int main(int argc, char **argv) {
 
     ierr = PetscViewerCreate(comm, &vtkviewersoln); CHKERRQ(ierr);
     ierr = PetscViewerSetType(vtkviewersoln, PETSCVIEWERVTK); CHKERRQ(ierr);
-    ierr = PetscViewerFileSetName(vtkviewersoln, "solution.vtk"); CHKERRQ(ierr);
+    ierr = PetscViewerFileSetName(vtkviewersoln, "solution.vtu"); CHKERRQ(ierr);
```
jeremylt, line 644: Should we go ahead and do this in every PETSc example?

jedbrown: Yeah, we never should have written `.vtk` (a slow and limited format what was deprecated almost 20 years ago).

examples/petsc/multigrid.c
jedbrown: The transform with `eps = 1` is the identity so it seems not worth deciding whether to apply. The default makes no transforamtion.

examples/petsc/multigrid.c
```diff
@@ -148,6 +173,44 @@ int main(int argc, char **argv) {
...
+        case 3:
+          c[i+1] = step(right(eps, y), left(eps, y), (1+lambda)/2);
+          c[i+2] = step(right(eps, z), left(eps, z), (1+lambda)/2);
+          break;
+        default:
+          c[i+1] = right(eps, y);
+          c[i+2] = right(eps, z);
+      }
+    }
+  }
```
jedbrown, line 213: Could you please move this into a function up above rather than this block of code. (`main` is already really long and I don't think the implementation of this transform is essential to the narrative.)

jedbrown (commented):
This looks good. The style check can be fixed by running `make style` locally (which will do what the diff shows if you click through).

examples/petsc/multigrid.c
jeremylt: good point

jedbrown (approved):
@reykoki I adjusted error handling and used `make style`. You can take a look at my last commit. I think it's ready to merge.

PR #721, opened 2021-03-31
mailmap - update jeremy's email

jeremylt:
I actually rolled over to a new email address (I think I edited this thing correctly)


jedbrown:
Got the own-domain swagger! :rocket:

PR #722, opened 2021-03-31
Disable autodetection of HIP and CUDA

jedbrown:
Some machines, especially HPC machines with shared filesystems (e.g., at
LLNL) have CUDA and/or HIP compilation stacks despite no such devices.
Autodetection produces confusing errors in such environments, so we
disable them and require explicitly setting CUDA_DIR and/or HIP_DIR to
locate them (picked up automatically from the environment if present).
Document this behavior.

Reported-by: Natalie Beams <nbeams@icl.utk.edu>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,11 +71,13 @@ OCCA_DIR ?= ../occa
 
 # env variable MAGMA_DIR can be used too
 MAGMA_DIR ?= ../magma
-# If CUDA_DIR is not set, check for nvcc, or resort to /usr/local/cuda
-CUDA_DIR  ?= $(or $(patsubst %/,%,$(dir $(patsubst %/,%,$(dir \
-               $(shell which nvcc 2> /dev/null))))),/usr/local/cuda)
+
+# Often /opt/cuda or /usr/local/cuda, but sometimes present on machines that don't support HIP
+CUDA_DIR  ?=
 CUDA_ARCH ?= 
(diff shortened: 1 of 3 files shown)


YohannDudouit:
What's the logic for doing this for CUDA? I think the solution is worst than the problem...

jeremylt:
I'm not sure I'd say it is worse. You can set your config to use the correct CUDA_DIR once and then everything works identically

Makefile
```diff
@@ -71,11 +71,13 @@ OCCA_DIR ?= ../occa
 
 # env variable MAGMA_DIR can be used too
 MAGMA_DIR ?= ../magma
-# If CUDA_DIR is not set, check for nvcc, or resort to /usr/local/cuda
-CUDA_DIR  ?= $(or $(patsubst %/,%,$(dir $(patsubst %/,%,$(dir \
-               $(shell which nvcc 2> /dev/null))))),/usr/local/cuda)
+
+# Often /opt/cuda or /usr/local/cuda, but sometimes present on machines that don't support HIP
```
YohannDudouit, line 75: HIP -> CUDA?

jedbrown: Fixed, thanks.

As for the solution, this isn't the first time it came up, and I think some symmetry between CUDA and HIP autodetection is desirable. We could try `CUDA_HOME` and `CUDA_PATH` as alternatives for autodetection.

PR #724, opened 2021-04-01
Update LibCEED.jl version number

pazner:
New LibCEED.jl version number (v0.1.1) after v0.8 release of libCEED. Will register the new version of the package once this is merged.

### DIFF
--- a/julia/LibCEED.jl/Project.toml
+++ b/julia/LibCEED.jl/Project.toml
@@ -1,7 +1,6 @@
 name = "LibCEED"
 uuid = "2cd74e05-b976-4426-91fa-5f1011f8952b"
-authors = ["Will Pazner <will.e.p@gmail.com>"]
-version = "0.1.0"
+version = "0.1.1"
 
 [deps]
 CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"


jeremylt:
The Rust and Python interfaces use the same version number as libCEED. Should the Julia interface do the same?

pazner:
That is a possibility, but perhaps having them uncoupled is more flexible. That way a breaking change in the Julia interface doesn't necessitate a new version number for all of libCEED, and vice versa.

jeremylt:
The way I was conceptualizing this, libCEED version X.Y and an interface tagged with the same release version would be compatible. To me, that would make it easier for users than having to maintain a compatibility matrix of libCEED versions + interface versions.

pazner:
That's fine by me, we should just keep in mind that it does preclude making a breaking change to one of the interfaces without incrementing the libCEED version number.

jedbrown:
After thinking more about it, I think independent versions are ultimately better for downstream users. It'll make it easier to upgrade libceed.so in a breaking way without breaking semver dependencies from Julia packages, and will allow the Julia interface to evolve rapidly even if libceed.so stays stable. If a breaking change on any one interface results in semver increments, then it'll be harder for us to ship upgrades to users.

jedbrown (approved):

PR #726, opened 2021-04-06
Refactor petsc examples to reduce redundant code

jeremylt:


jeremylt:
There is a small bug with BP3-6 in this that I need to track down

jeremylt:
Now there is just a bug with OCCA/OMP for bpssphere

jeremylt:
That example runs fine with the memcheck backends, so I suspect that we somehow have found another OCCA backend bug. I'll dig more tomorrow.

jeremylt:
Nope, I was wrong. A missed 'takearray' in the original bpssphere

jeremylt:
And we trimmed 781 lines of redundant code. We really should get around to doing this to the fluids miniapp one of these days.

I'll use this as the base to build the new BDDC example.

jeremylt:
Mildly related - do we want to finally kill that meshes directory in this PR?

jeremylt:
@LeilaGhaffari, just making sure you see this. I think we should do this with the fluids example too. I won't have time to get to it this month, but if you get a chance and/or are interested and hit any snags, let me know and I'll help.

jeremylt:
@valeriabarra, I'll start the style guide PR after this PR is approved.

LeilaGhaffari:
> @LeilaGhaffari, just making sure you see this. I think we should do this with the fluids example too. I won't have time to get to it this month, but if you get a chance and/or are interested and hit any snags, let me know and I'll help.

Yes, I am following this PR. I have used `PetscFunctionList()` for the fluids example and I was just able to get it work with the new release. I am hoping that the rest of the refactoring would be straightforward. Thanks for offering to help. I will let you know.

jeremylt:
I went ahead and tried the new style in this PR so we could see if we like it.

examples/petsc/area.c
```diff
@@ -60,24 +60,24 @@ int main(int argc, char **argv) {
...
-  PetscInt lsize, gsize, xlsize,
-           qextra  = 1, // default number of extra quadrature points
-           ncompx  = 3, // number of components of 3D physical coordinates
-           ncompu  = 1, // dimension of field to which apply mass operator
-           topodim = 2, // topological dimension of manifold
-           degree  = 3; // default degree for finite element bases
+       ceed_resource[PETSC_MAX_PATH_LEN] = "/cpu/self";
+  PetscInt l_size, g_size, xl_size,
+           q_extra  = 1, // default number of extra quadrature points
+           ncomp_x  = 3, // number of components of 3D physical coordinates
```
LeilaGhaffari, line 66: How about `n_comp_x`?

jeremylt: I'd be fine using `num_comp_x` (I think that's how we have it in Rust), but I'd prefer to avoid `n_comp_x`. The later feels harder to read to me.

LeilaGhaffari: I see your point. Yes, I like `num_comp_x`.

jeremylt: Thanks for asking - I went ahead and make some changes in that vein

jeremylt:
I plan on squashing this to 3 commits before merging
* code refactor
* code style update
* doc style update

examples/petsc/area.c
LeilaGhaffari: Thank you for taking care of that. I will make the same changes in the fluids example.

examples/petsc/bps.c
```diff
@@ -78,7 +78,7 @@ struct RunParams_ {
   MPI_Comm comm;
   PetscBool test_mode, read_mesh, user_l_nodes, write_solution;
   char *filename, *hostname;
-  PetscInt local_nodes, degree, q_extra, dim, num_comp_u, *m_elem;
+  PetscInt local_nodes, degree, q_extra, dim, num_comp_u, *mesh_elem;
```
jeremylt, line 81: Part of my philosophy with these name changes is that the PETSc (and solid mechanics) examples seem to be the entry point for people looking to understand and use libCEED (especially for new grad students), so I'm ok with slightly more verbose names that help ease that learning curve.

LeilaGhaffari: Very good point. This verbosity could have saved me a lot of time when I first started reading the fluids example.

jeremylt: Some of the pain with the fluids example comes from the fact I was the one who started it and it was my first attempt at making an mini-app :) hopefully this wave of style + modularity refactors will make things easier for everyone down the road

jedbrown (approved):
Tests are passing and changes look reasonable at a glance, so I'm happy to merge.

PR #727, opened 2021-04-06
Tiny fluid docs rewording

valeriabarra:
Just noticed this

### DIFF
--- a/examples/fluids/index.rst
+++ b/examples/fluids/index.rst
@@ -254,8 +254,8 @@ as
 where parameter :math:`\bm{\tau} \in \mathbb R^{3\times 3}` is an intrinsic time/space scale matrix.
 
 Currently, this demo provides two types of problems/physical models that can be selected
-at run time via the option ``-problem``. One is the problem of transport of energy in a
-uniform vector velocity field, called the :ref:`problem-advection` problem, and is the
+at run time via the option ``-problem``. One is the problem of the transport of energy in a
+uniform vector velocity field, called the :ref:`problem-advection` problem, and the other one is the
 so called :ref:`problem-density-current` problem.
 
 


jeremylt (approved):
Works for me.

examples/fluids/index.rst
```diff
@@ -254,8 +254,8 @@ as
 where parameter :math:`\bm{\tau} \in \mathbb R^{3\times 3}` is an intrinsic time/space scale matrix.
 
 Currently, this demo provides two types of problems/physical models that can be selected
-at run time via the option ``-problem``. One is the problem of transport of energy in a
-uniform vector velocity field, called the :ref:`problem-advection` problem, and is the
+at run time via the option ``-problem``. One is the problem of the transport of energy in a
```
jeremylt, line 257: I thought we were doing one line per sentence, or am I mixing up this documentation with another project?

valeriabarra: I just followed the current format of this period and just changed those couple of words. If we want to change it and make it one line per sentence, then we'll have to make sure the entire docs is consistent and will be a bigger change that I am not sure I can take at the moment :/ 
(Also having our project style guide might help -- I tried to advocate for it in the past, but didn't feel like it was well received)

jeremylt: I know we talked about that convention. I can go ahead and make that change and add a note in the contributors docs after this PR. Personally, I find (change in one sentence) => (one line diff) to be very handy when looking through commits.

valeriabarra: If you want to, please go ahead. Then, I think I should rename this PR since it won't be a tiny fluid docs rewording anymore XD

For a more comprehensive style guide, I would make sure we recommend the use of accessible choices for variable or function names (for instance by using `camelCase`, `PascalCase` or `snake_case` (whatever convention we like as a project, I think anything is more accessible than `nocase`)).

jeremylt: Oh, I'm thinking we should merge this now and fix the style later. Then I can start a more comprehensive style guide PR

PR #730, opened 2021-04-06
interface - improved resource matching check

jeremylt:
As discussed in Slack

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -595,7 +595,39 @@ int CeedSetData(Ceed ceed, void *data) {
 /// @{
 
 /**
+  @brief Get the list of avaliable resource names for Ceed contexts
+  Note: The caller is responsible for freeing the resources and priorities arrays,
+          but should not free the contents of the resources array.
+
+  @param[out] n          Number of avaliable resources
+  @param[out] resources  List of avaliable resource names
+  @param[out] priorities Resource name prioritization values, lower is better
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+// LCOV_EXCL_START
+int CeedRegistryGetList(size_t *n, char ***const resources,
+                        CeedInt **priorities) {
+  int ierr;
+
+  *n = num_backends;
+  ierr = CeedMalloc(*n, resources); CeedChk(ierr);
+  ierr = CeedMalloc(*n, priorities); CeedChk(ierr);
+  for (size_t i=0; i<num_backends; i++) {
+    *resources[i] = backends[i].prefix;
+    *priorities[i] = backends[i].priority;
+  }
+  return CEED_ERROR_SUCCESS;
+};
+// LCOV_EXCL_STOP
+
+/**
   @brief Initialize a \ref Ceed context to use the specified resource.
+  Note: Prefixing the resource with "help:" (e.g. "help:/cpu/self")
+    will result in CeedInt printing the current libCEED version number
+    and a list of current avaliable backend resources to stderr.
 
   @param resource  Resource to use, e.g., "/cpu/self"
   @param ceed      The library context
@@ -616,13 +648,34 @@ int CeedInit(const char *resource, Ceed *ceed) {
   // LCOV_EXCL_STOP
   ierr = CeedRegisterAll(); CeedChk(ierr);
 
+  // Check for help request
+  const char *help_prefix = "help";
+  size_t match_help;
+  for (match_help=0; match_help<4
+       && resource[match_help] == help_prefix[match_help]; match_help++) {}
+  if (match_help == 4) {
+    fprintf(stderr, "libCEED version: %d.%d%d%s\n", CEED_VERSION_MAJOR,
+            CEED_VERSION_MINOR, CEED_VERSION_PATCH,
+            CEED_VERSION_RELEASE ? "" : "+development");
+    fprintf(stderr, "Avaliable backend resources:\n");
+    for (size_t i=0; i<num_backends; i++) {
+      fprintf(stderr, "  %s\n", backends[i].prefix);
+    }
+    fflush(stderr);
+    match_help = 5; // Delineating character expected
+  } else {
+    match_help = 0;
+  }
+
+  // Find best match, currently computed as number of matching characters
+  //   from requested resource stem but may use Levenshtein in future
   size_t stem_length;
-  for (stem_length=0; resource[stem_length]
-       && resource[stem_length] != ':'; stem_length++) {}
+  for (stem_length=0; resource[stem_length+match_help]
+       && resource[stem_length+match_help] != ':'; stem_length++) {}
   for (size_t i=0; i<num_backends; i++) {
     size_t n;
     const char *prefix = backends[i].prefix;
-    for (n=0; prefix[n] && prefix[n] == resource[n]; n++) {}
+    for (n=0; prefix[n] && prefix[n] == resource[n+match_help]; n++) {}
     priority = backends[i].priority;
     if (n > matchlen || (n == matchlen && matchpriority > priority)) {
       matchlen = n;
@@ -727,9 +780,9 @@ int CeedInit(const char *resource, Ceed *ceed) {
   (*ceed)->debug = !!getenv("CEED_DEBUG") || !!getenv("DBG");
 
   // Backend specific setup
-  ierr = backends[matchidx].init(resource, *ceed); CeedChk(ierr);
+  ierr = backends[matchidx].init(&resource[match_help], *ceed); CeedChk(ierr);
 
-  // Copy resource prefix, if backend setup sucessful
+  // Copy resource prefix, if backend setup successful
   size_t len = strlen(backends[matchidx].prefix);
   char *tmp;
   ierr = CeedCalloc(len+1, &tmp); CeedChk(ierr);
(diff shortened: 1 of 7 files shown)


interface/ceed.c
```diff
@@ -616,22 +616,31 @@ int CeedInit(const char *resource, Ceed *ceed) {
   // LCOV_EXCL_STOP
   ierr = CeedRegisterAll(); CeedChk(ierr);
 
+  size_t stem_length;
+  for (stem_length=0; resource[stem_length]
+       && resource[stem_length] != ':'; stem_length++) {}
```
jeremylt, line 621: This assumes the delinator for device information is a `:`

nbeams: That's what I was planning in my branch (I'll push it soon -- I'm running into some technical issues testing the HIP version to make sure it works, but the CUDA version is working to set the device to something other than 0).

jeremylt:
Ah, codecov wants excelsion markers in the tests. I'll add soon

nbeams (approved):
LGTM.

Two minor comments for potential discussion but which I don't see as real problems that must be addressed:
 -  The CUDA (and HIP) backends will behave differently than the rest due to the current setup, which is meant to allow `/gpu/cuda` and `/gpu/hip` work (they currently will default to `gen`, though it's currently broken and should be fixed by my upcoming PR).  But because of this, for example, `/gpu/cuda/share` will match to `/gpu/cuda/shared` without giving an error, but `gpu/cuda/magm` will fail and suggest `/gpu/cuda/magma` (which the user was surely wanting).  
  - If the user has a typo like `/gpu/cuda/magam/det`, libCEED will suggest the closest match is `/gpu/cuda/magma` rather than `/gpu/cuda/magma/det`.  However, again, I don't think this is really a problem, since the message is clear enough that the user should realize the typo was the problem.

YohannDudouit (commented):
Since this discussion is now open, I would prefer to do something more in line with what @jedbrown suggested: "Closest prefix match and Levenshtein matches".
Ideally, I think the backend matching errors should look like the following examples:
- When `CUDA_DIR` omitted, asking for "/gpu/cuda/shared" should return an error of the type "/gpu/cuda/shared is unavailable, libCEED was compiled without CUDA support, please recompile libCEED and specify the CUDA_DIR variable corresponding to the location of the CUDA libraries on your machine."
- When Levenshtein distance is inferior to 1 (or 2, or more?) to an existing backend, an error message of the type "/gpu/cudq/shared does not name a backend, did you mean /gpu/cuda/shared?".
- When "/gpu/cuda/zzzz" is asked, an error message of the type "/gpu/cuda/zzzz does not name a backend, please use /gpu/cuda to let libCEED automatically select a CUDA backend, or use one of the following backend: /gpu/cuda/ref, /gpu/cuda/shared, /gpu/cuda/gen, /gpu/cuda/magma, /gpu/cuda/magma/det, /gpu/cuda/occa."

Same kind of behavior for HIP, AVX, libXSMM, MAGMA, OCCA...

jeremylt:
That is a bunch of special case string checking for C. I think those asks should be trimmed back.

jedbrown:
I wouldn't bother with Levenshtein distances at this time. `CUDA_DIR` is meant to be of compile-time significance, but not necessary at run time. I also think we don't have to be smart at this time; just printing a list of known backends won't swamp the user at this time. We can make it smarter when we learn of users being overwhelmed by the full list.

jeremylt:
Gap of my knowledege here - two questions

Can we currently do something like `ceed -v` (like `gcc -v`) and get the version information (I thought we had discussed it at some point)

If so, could we make `ceed --backends` work similarly?

jedbrown:
We don't have an executable named `ceed` and even if we did, it'd be trouble for a user of an application that calls libCEED (via MFEM or PETSc or ...) to track down some corresponding helper application. What I had in mind was that `CeedInit("help", &ceed)` would print useful things to stderr. We could extend it to `"help:/cpu/self"` to allow printing the help but also continuing without error. This could include version information, and would be useful to debug even within a GUI application that uses libCEED.

jedbrown:
Environment variables are an alternative way to procure such diagnostics, but an external user can't change them on a running application.

jeremylt:
We could make a new function that provides a list of available backends instead of smuggling the functionality in `CeedInit`

jedbrown:
We should do that, but standard debugging/diagnostics would still be useful, especially for one of us working with an unfamiliar application via an RFC2822 terminal.

jeremylt:
Perhaps we call through to a backend listing function `CeedListBackends(char *list)` and dump the result to stderr if we call `CeedInit("help", &ceed)`, such as via setting an ceed resource command line arg `-ceed help`?

jedbrown:
Yes, though perhaps `CeedRegistryGetList(size_t *n, const char *const *list)`. I don't know if it should also return priorities. Cheap access to sorted lists are used for things like bash autocomplete, GUIs, and smart editor modes for input files (e.g., [MOOSE autocomplete](https://github.com/dschwen/autocomplete-moose)). LibCEED is not that high level, but we should do our best to provide the tools to facilitate.

jeremylt:
If we output priorities too, that's sounding like a good place to output a JSON

jeremylt:
Perhaps a human readable output for `-ceed help` and a JSON for `CeedRegistryGetList(size_t *n, const char *const *list)`

jedbrown:
Why JSON? That would require us to allocate and copy into a JSON buffer, and the caller to parse the JSON (another copy). It's not that much data, but I don't see the benefit over simple native data structures (like an array of strings). The only place my feeble mind sees actually wanting JSON is in client-side JavaScript (none of which exists now or is in scope for libCEED).

jeremylt:
I was thinking of the case where we are providing the string resource name and integer priority number. That's a more complex structure, unless you were thinking of just providing two separate arrays?

jedbrown:
That's what I had in mind, yes.

jeremylt:
CI is happy. I think we're on the same page now.

interface/ceed.c
```diff
@@ -594,8 +594,40 @@ int CeedSetData(Ceed ceed, void *data) {
...
+  @ref User
+**/
+// LCOV_EXCL_START
+int CeedRegistryGetList(size_t *n, char ***const resources,
+                        CeedInt **priorities) {
+  int ierr;
+
+  *n = num_backends;
+  ierr = CeedMalloc(*n, resources); CeedChk(ierr);
+  ierr = CeedMalloc(*n, priorities); CeedChk(ierr);
```
jedbrown, line 617: `priorities` could be optional so you don't have to declare the variable and then free it if you won't use it.

interface/ceed.c
```diff
@@ -594,8 +594,40 @@ int CeedSetData(Ceed ceed, void *data) {
 /// @addtogroup CeedUser
 /// @{
 
+/**
+  @brief Get the list of avaliable resource names for Ceed contexts
+  Note: The caller is responsible for freeing the resources and priorities arrays,
+          but should not free the contents of the resources array.
```
jedbrown, line 600: We should explain how to free. `CeedFree` is in `ceed/backend.h` and I don't think we want to require that `CeedMalloc` be freeable with `free()`. So we should probably use raw `malloc` here.

jedbrown (approved):

YohannDudouit (changes requested):
I still think it would be valuable to recognize requests for existing but unavailable backends. Don't you think that each CUDA/HIP/libXSMM/MAGMA/OCCA backend's `CeedInit_*` could define its own error message to inform the user how to make available an existing unavailable backend? I think it shouldn't be too hard to fit in the design.

interface/ceed.c
```diff
@@ -616,22 +652,52 @@ int CeedInit(const char *resource, Ceed *ceed) {
...
+  // Check for help request
+  const char *help_prefix = "help";
+  size_t match_help;
+  for (match_help=0; match_help<4
+       && resource[match_help] == help_prefix[match_help]; match_help++) {}
+  if (match_help == 4) {
+    fprintf(stderr, "libCEED version: %d.%d%d%s\n", CEED_VERSION_MAJOR,
+            CEED_VERSION_MINOR, CEED_VERSION_PATCH,
+            CEED_VERSION_RELEASE ? "" : "+development");
+    fprintf(stderr, "Avaliable backend resources:\n");
```
YohannDudouit, line 664: typo

tests/junit.py
```diff
@@ -98,6 +98,10 @@ def run(test, backends):
                     case.add_skipped_info('device memory not supported {} {}'.format(test, ceed_resource))
 
             if not case.is_skipped():
+                if test[:4] in 't006 t007'.split():
+                    check_required_failure(case, proc.stderr, 'No suitable backend:')
+                if test[:4] in 't008'.split():
+                    check_required_failure(case, proc.stderr, 'Avaliable backend resources:')
```
YohannDudouit, line 104: typo

interface/ceed.c
YohannDudouit:
```suggestion
    fprintf(stderr, "Available backend resources:\n");
```

tests/junit.py
YohannDudouit:
```suggestion
                    check_required_failure(case, proc.stderr, 'Available backend resources:')
```

jeremylt:
Each backend's `CeedInit` is not compiled if they are not supported, so that backend can't generate an error message stating that it is not compiled.

We could create some new function to do that with the weak registration perhaps? But we'd have to be careful to prevent errant matches against the weakly registered backends, perhaps by setting the priority value to MAX_INT.

PR #725, opened 2021-04-01
Document release procedures

pazner:
Add some documentation for the release procedure. For now I've added documentation for the Julia interface, @jedbrown can add notes for C, Python, and Rust.

PR #732, opened 2021-04-07
header - update minor version in header definition

jeremylt:


jeremylt:
This should be added to #725

jedbrown (approved):
Damn, thanks for catching. Making this authoritative would be nice.

PR #733, opened 2021-04-07
Extra error message for uncompiled backends

jeremylt:
One of these days I'll figure out how to get  spellcheck to stop turning itself off.


YohannDudouit (approved):

jeremylt:
Added perhaps a better error message for existing but uncompiled backends:

> $ ./build/t000-ceed /cpu/self/xsmm/serial
/home/jeremy/Dev/libCEED/backends/ceed-backend-weak.c:9 in CeedInit_Weak(): Backend not currently compiled: /cpu/self/xsmm/serial
Aborted (core dumped)

backends/ceed-backend-weak.c
```diff
@@ -1,17 +1,30 @@
...
+                   "Backend not currently compiled: %s", resource);
+}
+
+// This function provides a debug target for weak symbols
+static int CeedRegister_Weak(const char *name, int num_prefixes, ...) {
   if (getenv("CEED_DEBUG")) fprintf(stderr, "Weak %s\n", name);
+  va_list prefixes;
+  va_start(prefixes, num_prefixes);
+  for (int i=0; i<num_prefixes; i++) {
+    CeedRegister(va_arg(prefixes, const char*), CeedInit_Weak, UINT_MAX);
```
jeremylt, line 19: Registered with largest possible priority value so that `CeedInit_Weak` only runs on an exact character for character match

backends/ceed-backend-list.h
```diff
@@ -5,24 +5,24 @@
...
+MACRO(CeedRegister_Memcheck_Serial, 1, "/cpu/self/memcheck/serial")
+MACRO(CeedRegister_Occa, 4, "/cpu/self/occa", "/cpu/openmp/occa", "/gpu/hip/occa", "/gpu/cuda/occa")
+MACRO(CeedRegister_Opt_Blocked, 1, "/cpu/self/opt/blocked")
+MACRO(CeedRegister_Opt_Serial, 1, "/cpu/self/opt/serial")
+MACRO(CeedRegister_Ref, 1, "/cpu/self/ref/serial")
+MACRO(CeedRegister_Ref_Blocked, 1, "/cpu/self/ref/blocked")
+MACRO(CeedRegister_Tmpl, 1, "/cpu/self/tmpl")
+MACRO(CeedRegister_Tmpl_Sub, 1, "/cpu/self/tmpl/sub")
+MACRO(CeedRegister_Xsmm_Blocked, 1, "/cpu/self/xsmm/blocked")
+MACRO(CeedRegister_Xsmm_Serial, 1, "/cpu/self/xsmm/serial")
```
jedbrown, line 28: This is a pretty clean solution, though I really don't love the redundancy and don't know how this can extend to programmatically generated resource strings (I think we've discussed this for multiple devices and maybe other contexts). One of my thoughts was that `ceed-backend-list.h` could some day be automatically generated by the build system, from scanning the source directories for matching declarations.

When we load plugins, there is definitely no centralized list. That would put them on unequal footing from in-package backends. I would have said that a simpler, and more generic error message would be sufficient, but now that you've put this together, I'm not going to object. I would like to hear other opinions.

jeremylt: I agree that the redundancy felt bad while I was building this. It would have been easier/cleaner if we didn't have a backend that responded to substantially different names (OCCA).

I'm definitely ok omitting this code if we prefer, I mostly put it together to see 1) if I could somewhat cleanly and 2) if we like it.

interface/ceed.c
```diff
@@ -661,9 +674,11 @@ int CeedInit(const char *resource, Ceed *ceed) {
     fprintf(stderr, "libCEED version: %d.%d%d%s\n", CEED_VERSION_MAJOR,
             CEED_VERSION_MINOR, CEED_VERSION_PATCH,
             CEED_VERSION_RELEASE ? "" : "+development");
-    fprintf(stderr, "Avaliable backend resources:\n");
+    fprintf(stderr, "Available backend resources:\n");
     for (size_t i=0; i<num_backends; i++) {
-      fprintf(stderr, "  %s\n", backends[i].prefix);
+      // Only report compiled backends
+      if (backends[i].priority < CEED_MAX_BACKEND_PRIORITY)
+        fprintf(stderr, "  %s\n", backends[i].prefix);
```
YohannDudouit, line 681: That's cool!

backends/ceed-backend-weak.c
```diff
@@ -2,16 +2,31 @@
...
 #include <stdio.h>
 
-// This function provides a debug target for weak symbols
 // LCOV_EXCL_START
-static int CeedRegister_Weak(const char *name) {
+// This function provides improved error messages for uncompiled backends
+static int CeedInit_Weak(const char *resource, Ceed ceed) {
+  return CeedError(ceed, CEED_ERROR_UNSUPPORTED,
+                   "Backend not currently compiled: %s", resource);
+}
```
YohannDudouit, line 10: That's great, how difficult would it be to let backends personalize the error message? But I think this is great already.

backends/ceed-backend-list.h
YohannDudouit: I think the feature it brings is nice, I'm not too concerned about the style, but I might not see the issue you foresee.

YohannDudouit (approved):
Thank you for doing this @jeremylt , I think this would be a nice addition.

backends/ceed-backend-weak.c
jeremylt: I mean, technically we could hack it in as extra information in the string holding the resource name in the MACRO. Not sure if I like it, but it'll *work*.

jedbrown: It could take pairs of strings, one stating the resource name and the other explaining which make options are needed (`make CUDA_DIR=/path/to/cuda`), but I think this is hard to test and wrong/outdated information is worse than less detailed information that accurately identifies the problem.

jeremylt: We could just change the message to
```
$ ./build/t000-ceed /cpu/self/xsmm/serial
/home/jeremy/Dev/libCEED/backends/ceed-backend-weak.c:9 in CeedInit_Weak(): Backend not currently compiled: /cpu/self/xsmm/serial
Consult the installation instructions to compile this backend
Aborted (core dumped)
```

jeremylt: done

jedbrown (approved):

PR #736, opened 2021-04-10
Solids - Current configuration update 

rezgarshakeri:
Recently we used `computeDetF` function to compute `J=det(F)`. However, we can compute it using `det(C)`.

### DIFF
--- a/examples/solids/qfunctions/FSCurrent-NH1.h
+++ b/examples/solids/qfunctions/FSCurrent-NH1.h
@@ -70,21 +70,6 @@ static inline CeedScalar log1p_series_shifted(CeedScalar x) {
 #endif
 
 // -----------------------------------------------------------------------------
-// Compute det F
-// -----------------------------------------------------------------------------
-static inline CeedScalar computeDetFdav(CeedScalar F[3][3]) {
-  CeedScalar a1 = F[0][0] * F[1][1] * F[2][2];
-  CeedScalar a2 = F[0][1] * F[1][2] * F[2][0];
-  CeedScalar a3 = F[0][2] * F[1][0] * F[2][1];
-  CeedScalar a4 = F[0][2] * F[1][1] * F[2][0];
-  CeedScalar a5 = F[0][0] * F[1][2] * F[2][1];
-  CeedScalar a6 = F[0][1] * F[1][0] * F[2][2];
-
-  return (a1+a2+a3) - (a4+a5+a6);
-};
-
-
-// -----------------------------------------------------------------------------
 // Compute det C - 1
 // -----------------------------------------------------------------------------
 #ifndef DETCM1
@@ -148,7 +133,7 @@ static inline int commonFtau(const CeedScalar lambda, const CeedScalar mu,
   CeedScalar logJ = log1p_series_shifted(detC_m1)/2.;
 
 
-  CeedScalar J = computeDetFdav(F);
+  const CeedScalar J = sqrt(detC_m1+1);
 
   // *INDENT-OFF*
   //Computer F^(-1)
--- a/examples/solids/qfunctions/FSCurrent-NH2.h
+++ b/examples/solids/qfunctions/FSCurrent-NH2.h
@@ -70,23 +70,6 @@ static inline CeedScalar log1p_series_shifted(CeedScalar x) {
 #endif
 
 // -----------------------------------------------------------------------------
-// Compute det F
-// -----------------------------------------------------------------------------
-#ifndef DETF
-#define DETF
-static inline CeedScalar computeDetF(CeedScalar F[3][3]) {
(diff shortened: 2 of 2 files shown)


rezgarshakeri:
It seems `sqrt` function is much more expansive than using `computeDetF`

jeremylt:
Which optimization flags? I'd be curious to play around and understand whats going on.

jedbrown:
In the constitutive model, we often only need `log(J) = log1p(detCm1) / 2`. Note that when we're in the current configuration, it's unnatural to compute C, but detC = detb.

For numerical stability, I'd prefer to formulate in terms of what I might call a "left Green-Lagrange strain", `e = (b - I)/2`. It serves the same role as the Green-Lagrange strain, simplifies CRs, and avoids numerical stability issues that arise when working directly with `b`. For example, Neo-Hookean in the current configuration is `\tau = \lambda \log J I + 2 \mu e`. (I'm not aware of a name for this tensor. The Almansi tensor plays a similar role, but needs `b^{-1}`, which is inconvenient for constitutive modeling.)

rezgarshakeri:
> Which optimization flags? I'd be curious to play around and understand whats going on.

Arsh checked in Noether with one of his meshes. Would you please try with dmplex:
`mpiexec -n 2 ./elasticity -nu 0.3 -E 20 -degree 1 -dm_plex_box_faces 10,10,10 -problem FSCurrent-NH1 -num_steps 15 -bc_clamp 3 -bc_traction 4 -bc_traction_4 0,0,-1 -snes_linesearch_type cp`

right now in main we use `J=computeDetF` function and in this branch we use `sqrt`

rezgarshakeri:
> In the constitutive model, we often only need `log(J) = log1p(detCm1) / 2`. Note that when we're in the current configuration, it's unnatural to compute C, but detC = detb.
> 
> For numerical stability, I'd prefer to formulate in terms of what I might call a "left Green-Lagrange strain", `e = (b - I)/2`. It serves the same role as the Green-Lagrange strain, simplifies CRs, and avoids numerical stability issues that arise when working directly with `b`. For example, Neo-Hookean in the current configuration is `\tau = \lambda \log J I + 2 \mu e`. (I'm not aware of a name for this tensor. The Almansi tensor plays a similar role, but needs `b^{-1}`, which is inconvenient for constitutive modeling.)

yes, we used `log(J) = log1p(detCm1) / 2`, but in current configuration we need `F^{-1}` to map qdata from initial to current. And we need `det(F)` to compute `F{-1}`

ArashMehraban:
I'd appreciate it if nothing is checked on Noether now. I am running a bunch of stuff using 32 and 64 procs.

ArashMehraban:
BTW, FSCurrent-NH2 is not 50% faster than other algorithms! That is what I am checking now. It might be in the realm of 2%-15% with large deformation runs. I realized all of our tests were on a  small strain scale. For an actual large deformation problem  2%-15% appear to be achievable.

rezgarshakeri:
> Which optimization flags? I'd be curious to play around and understand whats going on.

sorry I forgot to say we used `make OPT="-O3 -march=native"`

jeremylt:
Ah. Are the current runs using `-funsafe-math-optimizations`? We saw a performance impact with that flag the other day.

rezgarshakeri:
> Ah. Are the current runs using `-funsafe-math-optimizations`? We saw a performance impact with that flag the other day.

yeah, we used `PETSc_DIR=/projects/petsc` on noether which uses `-funsafe-math-optimizations`.

jeremylt:
Yea, but was the solids example also compiled with those flags? The QFunction has a lot of flops in it

rezgarshakeri:
> Yea, but was the solids example also compiled with those flags? The QFunction has a lot of flops in it

no just, `make OPT="-O3 -march=native"`

jeremylt:
Hmm. I wonder if there is any performance difference when compiling with everything with those math ops. I suppose the only way to know is to check.

rezgarshakeri:
yeah, lets Arash finishes his run, we will check this week hopefully

jedbrown:
`make V=1` shows the complete compile command, which I believe contains all the flags from PETSc plus the extra flags you pass in `OPT='-O3 -march=native'`. I would not expect an issue with compiler flags in what we've just discussed.

jedbrown:
Ideally, we would compute `Jm1 = det(F) - 1` stably from `\nabla_X u` and then _not_ need to compute the determinant of `C` or `b`.

jeremylt:
Huh, so was a lot of the speed change we say the other day in the vector-vector ops in PETSc?

jedbrown:
No, basically none there. The big differences were in qfunction vectorization. Now it's problem size dependent memory consequences.

PR #739, opened 2021-04-12
Copy Ceed Object References

jeremylt:
This adds reference copying to libCEED. This creates a second, separate reference to the same Ceed object that can (and should) be independently destroyed.

@knepley, I think this will address your segfault. @YohannDudouit, this might also be of interest for MFEM.

### DIFF
--- a/interface/ceed-elemrestriction.c
+++ b/interface/ceed-elemrestriction.c
@@ -280,7 +280,7 @@ int CeedElemRestrictionSetData(CeedElemRestriction rstr, void *data) {
 
   @ref Backend
 **/
-int CeedElemRestrictionIncrementRefCounter(CeedElemRestriction rstr) {
+int CeedElemRestrictionReference(CeedElemRestriction rstr) {
   rstr->ref_count++;
   return CEED_ERROR_SUCCESS;
 }
@@ -358,7 +358,7 @@ int CeedElemRestrictionCreate(Ceed ceed, CeedInt num_elem, CeedInt elem_size,
 
   ierr = CeedCalloc(1, rstr); CeedChk(ierr);
   (*rstr)->ceed = ceed;
-  ierr = CeedIncrementRefCounter(ceed); CeedChk(ierr);
+  ierr = CeedReference(ceed); CeedChk(ierr);
   (*rstr)->ref_count = 1;
   (*rstr)->num_elem = num_elem;
   (*rstr)->elem_size = elem_size;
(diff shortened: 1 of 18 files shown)


jeremylt:
Note: I can't reproduce the CI issue locally yet. I'm investigating - it might be a cargo_tarpaulin issue?

interface/ceed-elemrestriction.c
```diff
@@ -574,6 +588,33 @@ int CeedElemRestrictionCreateBlockedStrided(Ceed ceed, CeedInt num_elem,
...
+  @ref User
+**/
+int CeedElemRestrictionCopyReference(CeedElemRestriction rstr,
+                                     CeedElemRestriction *rstr_copy) {
+  int ierr;
+
+  ierr = CeedElemRestrictionIncrementRefCounter(rstr); CeedChk(ierr);
+  if (*rstr_copy) {
+    ierr = CeedElemRestrictionDestroy(rstr_copy); CeedChk(ierr);
+  }
```
jedbrown, line 613: We shouldn't need the guard because destroy checks for NULL.
```c
  if (!*rstr || --(*rstr)->ref_count > 0) return CEED_ERROR_SUCCESS;
```

jeremylt: Duh, I should have seen that. Thank you

jeremylt:
This crate isn't building correctly in CI: https://github.com/KyleMayes/clang-sys
but I'm not sure why. It doesn't have anything to do with us though

jeremylt:
I don't know how I fixed it, but I fixed CI. Ready to merge.

jedbrown (approved):
All looks fine. Would using "reference" as a verb, as in `CeedReference(Ceed ceed)`, with `CeedReferenceCopy(Ceed src, Ceed *dest)` or `CeedReferenceShare(Ceed src, Ceed *dest)` be better for autocomplete?
PETSc does not have a "copy" function like this, but it could be used most places where we currently spell out the operations.

jeremylt:
I updated the function names. A fix is in for rust-lang/rust on what I think the Rust issue is, but I don't know when it will merge. We should be ok to just merge whenever you're ready @jedbrown

YohannDudouit (approved):
@jeremylt Was there a specific reason for you to think this could be useful in MFEM?
Otherwise this looks good to me.

jeremylt:
I don't have a current specific use case in mind - it just seems like something that might be useful in the future

jeremylt:
Rust is happy, but GitHub is confused. This is good but no green button for some reason

PR #740, opened 2021-04-12
Update device ID selection for HIP/CUDA/MAGMA backends

### LINKED ISSUE #709: GPU resource names and device IDs
Recently, in fdf826bb480f6f0c657574035d43bde2679baed8, the code used by the CUDA and HIP backends to determine the device ID was changed slightly, to avoid resetting to device 0 if another library had already set it to a different value.  However, it seems this has had the (unintended?) effect of removing the "pick the best HIP/CUDA backend" resource option of `/gpu/cuda` or `/gpu/hip` (which would pick the cuda-gen or hip-gen backends currently).  This affects MFEM, as it uses only `/gpu/hip` for its default `ceed-hip` backend (done originally rather than hard-code `/gpu/hip/ref`, the first-available HIP backend).  

The reason `/gpu/cuda` and `/gpu/hip` no longer work is because  in `Ceed[Cuda/Hip]Init`, `deviceID` is set to -1, then used for `*GetDeviceProperties`, which will throw an error for the invalid ID.  An easy fix for this is to set `deviceID = currentDeviceID` as an `else` condition if we don't end up calling `hipSetDevice` to change the `deviceID`.  However, after looking some more at this, I realized I wasn't sure what the expected behavior was really intended to be. 

From looking at the code in `CeedHipInit`, I assumed that the way a user would specify a device ID would be to add it at the end, e.g., `/gpu/hip/shared/1`.  However, because the `nrc` is set to 8 (or 9 for cuda) to allow matching for the "pick the best case" scenario, it seems that the line
```
int deviceID = (slash && rlen > nrc + 1) ? atoi(&resource[nrc + 1]) : -1;
```

would result in `atoi(s)` in this case, which returns 0 -- the number at the end of the string is ignored.   (And this is also why `/gpu/hip/ref`, `/gpu/hip/shared`, and `/gpu/hip/gen` do not encounter the same error as `/gpu/hip` - because they all have the case where `slash && rlen > nrc + 1` is true, but thanks to the short `nrc` value, the `atoi` call is on a non-numeric character, which returns 0.) 

Also, I realized we don't have this sort of check/expectation that the user can set the device ID from the resource string in the MAGMA backend at all -- we only get the current device.  Thus, I want to make sure I understand the way this capability is supposed to work.

nbeams:
This PR fixes the device ID selection for the CUDA/HIP backends while retaining the option to use `/gpu/cuda` or `/gpu/hip` as shortcuts for the best backends (currently chooses `gen`).   It also adds the ability to set the device ID for the MAGMA backends. 

The device ID string was chosen to match that of the OCCA backends, i.e., `:device_id=#`.  An example full resource string would be `/gpu/hip/shared:device_id=1`.  

Other minor changes include:
 - Change some headers to fix compiler warnings
 - Remove the backend data structs for the `shared` and `gen` backends, and have those backends use the base `Ceed_Hip` or `Ceed_Cuda` structs.  (Previously, the `shared` and `gen` structs only contained a `Ceed_Hip` or `Ceed_Cuda` pointer called `base`, but nothing was ever accessed through `base` and the CUDA/HIP `Init` functions only ever assumed they were given the base struct directly.)

Resolves #709

### DIFF
--- a/backends/cuda-shared/ceed-cuda-shared.c
+++ b/backends/cuda-shared/ceed-cuda-shared.c
@@ -16,7 +16,6 @@
 
 #include <ceed/ceed.h>
 #include <ceed/backend.h>
-#include <stdbool.h>
 #include <string.h>
 #include "ceed-cuda-shared.h"
 #include "../cuda/ceed-cuda.h"
@@ -34,14 +33,14 @@ static int CeedInit_Cuda_shared(const char *resource, Ceed ceed) {
   // LCOV_EXCL_STOP
   ierr = CeedSetDeterministic(ceed, true); CeedChk(ierr);
 
-  Ceed ceedref;
-  CeedInit("/gpu/cuda/ref", &ceedref);
-  ierr = CeedSetDelegate(ceed, ceedref); CeedChk(ierr);
-
-  Ceed_Cuda_shared *data;
+  Ceed_Cuda *data;
   ierr = CeedCalloc(1, &data); CeedChk(ierr);
   ierr = CeedSetData(ceed, data); CeedChk(ierr);
   ierr = CeedCudaInit(ceed, resource, nrc); CeedChk(ierr);
+
+  Ceed ceedref;
+  CeedInit("/gpu/cuda/ref", &ceedref);
+  ierr = CeedSetDelegate(ceed, ceedref); CeedChk(ierr);
 
   ierr = CeedSetBackendFunction(ceed, "Ceed", ceed, "BasisCreateTensorH1",
                                 CeedBasisCreateTensorH1_Cuda_shared);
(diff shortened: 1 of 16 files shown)


backends/cuda-shared/ceed-cuda-shared.c
```diff
@@ -16,7 +16,6 @@
 
 #include <ceed/ceed.h>
 #include <ceed/backend.h>
-#include <stdbool.h>
 #include <string.h>
 #include "ceed-cuda-shared.h"
 #include "../cuda/ceed-cuda.h"
```
jeremylt, line 21: One of these days we'll have to decide how we want this reaching over into other backend's headers to work.

jeremylt (approved):

nbeams:
@jeremylt Do you think the Rust test failure is the same as what you were talking about on #739?

jeremylt:
Yes, there is a bug in the nightly builds of Rust, but the nightly builds are required for our current way of doing code coverage in Rust.

PR #741, opened 2021-04-14
Fix python test docs

jedbrown:
Just a minor inconsistency and add make targets because my fingers can't remember the process without tab completion.


jeremylt (approved):
Makes sense to me. Do we want `test-py` as a top level target?

jedbrown:
It would be nice, but I usually contain my testing in a virtualenv and I'm not sure I should push those conventions on others. What do you think?

jeremylt:
Ah, that is a fair point. I think our testing with a virtualenv may differ? I'd be curious to see you workflow. I clone libCEED while inside the virtualenv then build and execute tests in that clone.

Other thought: We might want to use these in our GitHub action to tidy up a bit.

Other other thought: should these tests actuall go in the Python folder? The other language specific tests live in their language impl folders.

tests/python/Makefile
```diff
@@ -17,4 +17,10 @@
 clean:
 	rm -rf build __pycache__ .pytest_cache *.so
 
-.PHONY: cln clean
+setup:
+	$(PYTHON) setup-qfunctions.py build
+
+test:
```
jeremylt, line 23:
```
TEST_OPTS ?= --ceed /cpu/self/ref/serial -vv
test:
	$(PYTHON) -m pytest test-*.py $(TEST_OPTS)
```

what about this, so in our GitHub action we can

```
      run: |
        make info
        make -j2
        export CEED_DIR=$PWD
        pip install .
        cd tests/python
        make setup && make test
        cd ../..
```

jeremylt: I added a couple of commits to this effect. Hopefully they seem reasonable

python/tests/Makefile
```diff
@@ -17,4 +17,11 @@
 clean:
 	rm -rf build __pycache__ .pytest_cache *.so
 
-.PHONY: cln clean
+setup:
+	$(PYTHON) setup-qfunctions.py build
+
+TEST_OPTS ?= --ceed /cpu/self/ref/serial
+test: setup
```
jedbrown, line 24: I'm fine with this because it's more foolproof. If we knew the name of the generated file, we could write a proper dependency for it instead of using a phony target.

.github/workflows/python-test-with-style.yml
```diff
@@ -30,9 +30,8 @@ jobs:
         make -j2
         export CEED_DIR=$PWD
         pip install .
-        cd tests/python
-        python setup-qfunctions.py build
-        pytest test-*.py --ceed /cpu/self/ref/serial -vv
+        cd python/tests
```
jedbrown, line 33: Rename seems fine to me, and is clear if we ever split Python bindings into a new repository. Can you please just confirm that the tests don't end up in a wheel?

jedbrown (commented):
I can't approve my own PR, but I approve of merging whenever you're done.

.github/workflows/python-test-with-style.yml
jeremylt: I checked and they don't show up, at least on my machine

jeremylt (approved):

PR #744, opened 2021-04-14
PointwiseMult and AXPY convenience functions

jeremylt:
These would be nice to have for my BDDC example.

It shouldn't be hard to add CUDA/HIP kernels down the road.

### DIFF
--- a/interface/ceed-vector.c
+++ b/interface/ceed-vector.c
@@ -561,6 +561,15 @@ int CeedVectorAXPY(CeedVector y, CeedScalar alpha, CeedVector x) {
                      "Cannot use same vector for x and y in CeedVectorAXPY");
   // LCOV_EXCL_STOP
 
+  Ceed ceed_parent_x, ceed_parent_y;
+  ierr = CeedGetParent(x->ceed, &ceed_parent_x); CeedChk(ierr);
+  ierr = CeedGetParent(y->ceed, &ceed_parent_y); CeedChk(ierr);
+  if (ceed_parent_x != ceed_parent_y)
+    // LCOV_EXCL_START
+    return CeedError(y->ceed, CEED_ERROR_INCOMPATIBLE,
+                     "Vectors x and y must be created by the same Ceed context");
+  // LCOV_EXCL_STOP
+
   // Backend implementation
   if (y->AXPY)
     return y->AXPY(y, alpha, x);
@@ -594,7 +603,7 @@ int CeedVectorPointwiseMult(CeedVector w, CeedVector x, CeedVector y) {
   int ierr;
   CeedScalar *w_array;
   CeedScalar const *x_array, *y_array;
-  CeedInt n_x, n_y, n_w;
+  CeedInt n_w, n_x, n_y;
 
   ierr = CeedVectorGetLength(w, &n_w); CeedChk(ierr);
   ierr = CeedVectorGetLength(x, &n_x); CeedChk(ierr);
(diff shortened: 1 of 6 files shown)


interface/ceed-vector.c
```diff
@@ -493,6 +493,103 @@ int CeedVectorNorm(CeedVector vec, CeedNormType norm_type, CeedScalar *norm) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Compute y = alpha x + y
+
+  @param y[in/out]  target vector for sum
```
jedbrown, line 499:
```suggestion
  @param y[in,out]  target vector for sum
```
This is what we use elsewhere and I've seen documented.

interface/ceed-vector.c
```diff
@@ -493,6 +493,103 @@ int CeedVectorNorm(CeedVector vec, CeedNormType norm_type, CeedScalar *norm) {
...
+  if (y->AXPY)
+    return y->AXPY(y, alpha, x);
+
+  ierr = CeedVectorGetLength(y, &n_y); CeedChk(ierr);
+  ierr = CeedVectorGetLength(x, &n_x); CeedChk(ierr);
+  if (n_x != n_y)
+    // LCOV_EXCL_START
+    return CeedError(y->ceed, CEED_ERROR_UNSUPPORTED,
+                     "Cannot add vector of different lengths");
+  // LCOV_EXCL_STOP
```
jedbrown, line 522: Seems like we should do this input validation, as well as checking that `x->ceed == y->ceed`, before calling `y->AXPY`.

interface/ceed-vector.c
```diff
@@ -493,6 +493,103 @@ int CeedVectorNorm(CeedVector vec, CeedNormType norm_type, CeedScalar *norm) {
...
+    return w->PointwiseMult(w, x, y);
+
+  ierr = CeedVectorGetLength(w, &n_w); CeedChk(ierr);
+  ierr = CeedVectorGetLength(x, &n_x); CeedChk(ierr);
+  ierr = CeedVectorGetLength(y, &n_y); CeedChk(ierr);
+  if (n_w != n_x || n_w != n_y)
+    // LCOV_EXCL_START
+    return CeedError(w->ceed, CEED_ERROR_UNSUPPORTED,
+                     "Cannot multiply vectors of different lengths");
+  // LCOV_EXCL_STOP
```
jedbrown, line 565: Same comment regarding input validation.

jedbrown (approved):

jeremylt:
@jedbrown, it looks like this will require your manual intervention too

jedbrown:
Did we change a configuration such that these Required checks have vanished? Shall I update branch permissions accordingly?
image

jeremylt:
I changed the complier for the Rust tests from GCC to Clang during my early debugging of the CI problem, so that might be the issue

jedbrown:
Thoughts about `x->ceed == y->ceed`? What if one is HIP and the other CUDA? `CEED_MEM_DEVICE` would mean different things and crash, right? Or must each backend be responsible for checking consistency when required (e.g., when using direct access to device memory)?

jeremylt:
I don't think we can be to aggressive in checking that the two ceeds match. The default impl should run with any pair of vectors from any backend. 

I'm not sure how much we'd see it, but in theory OCCA and CUDA backends using the same device should be able to use the same device pointers (when that OCCA fix happens). I'm not sure if that's a realistic use case though.

Would two ceeds on the same host pointing to different devices communicating via GPUDirect or something akin be a practical use case?

jedbrown:
I think transferring/viewing a vector to a different `ceed` may be a desirable feature (sometimes it can be no-copy), but making it happen implicitly inside an operation feels like a lot of implicitness.

jeremylt:
That's a fair point. We can add a check, but we have to be a bit careful. Perhaps:

```C
Ceed ceed_parent_x, ceed_parent_y;
ierr = CeedGetParent(x->ceed, &ceed_parent_x); CeedChk(ierr);
ierr = CeedGetParent(y->ceed, &ceed_parent_y); CeedChk(ierr);
if (ceed_parent_x != ceed_parent_y)
    // LCOV_EXCL_START
    return CeedError(ceed->y, CEED_ERROR_INCOMPATIBLE, "Vectors x and y must be created by the same Ceed context");
  // LCOV_EXCL_STOP
```

jedbrown:
That would be fine with me. It's easier to relax the check once there are use cases than to have people depending on some artifact of the implementation, such as naively mapping to host arrays.

jeremylt:
Added in both places. CodeCov is reporting a false change in the backend coverage :shrug: but we should be good to go now

.github/workflows/rust-test-with-style.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        compiler: [clang]
+        compiler: [gcc-10]
```
jedbrown, line 41: Do you still want this change?

jeremylt: I'm fine either way; I don't think it makes any difference. I removed it in this commit because it was originally a debugging commit that slipped into main.

Issue #645, opened 2020-10-18
Symmetric Schur decomposition fails for diagonal matrix


pazner:
The following simple example fails with error `Symmetric QR failed to converge`:
```c
#include <ceed.h>

int main() {
  Ceed ceed;
  CeedInit("/cpu/self/ref", &ceed);
  double A[9] = {
    1, 0, 0,
    0, 1, 0,
    0, 0, 1};
  CeedScalar lambda[3] = {0.0};
  CeedSymmetricSchurDecomposition(ceed, A, lambda, 3);
  return 0;
}
```
In particular, this causes `CeedOperatorCreateFDMElementInverse` to fail when the quadrature points and nodal points are the same, so the mass matrix is diagonal.

In general, `t540-operator.c` seems pretty fragile — changing `P` to values other than 4 does not work for me.

jeremylt:
That is a pretty experimental feature.

 Can you provide some more context? A constant coefficient mass matrix with collocated quadrature is not the intended use of this function since that is trivial to create by hand.

pazner:
> Can you provide some more context? A constant coefficient mass matrix with collocated quadrature is not the intended use of this function since that is trivial to invert by hand.

Of course, but it doesn't make sense for this feature to work for the nontrivial cases, and then to break for the trivial case.

> That said, it is strange that the symmetric QR did not converge.
> 
> Whe you say that p=4 only works, are you changing both p and q?

Yes, I haven't been able to find a `Q` that works, given, for example, `P=5`. This could be because of some other issue in the test that is distinct from my original issue, but it seemed worth mentioning.

jeremylt:
Though, I would still need to factor the collocated mass matrix if the user has any collocated problem. I don't see an obvious reason why it's not working. I might have mangled the algorithm from the matrix methods book I have.

PR #749, opened 2021-04-19
fdm - minor fixes

jeremylt:


jeremylt:
I have additional fixes to add tomorrow but I will put them in a follow-up. This is ready to merge.

jeremylt:
So this is definitely working better, but `t540` still only works for P from 4 to 6. I'm really surprised that 2 and 3 don't work.

jeremylt:
Something is screwy, I found a way to make it even worse
```
-- CEED Benchmark Problem 3 -- libCEED + PETSc + BDDC --
  PETSc:
    PETSc Vec Type                     : seq
  libCEED:
  Mesh:
    Number of 1D Quadrature Points (q) : 4
    Global Nodes                       : 125
    Owned Nodes                        : 125
    DoF per node                       : 1
  BDDC:
    Injection                          : scaled
    Global Interface Nodes             : 8
    Owned Interface Nodes              : 8
  KSP:
    Total KSP Iterations               : 61
    Final rnorm                        : 7.065059e-10
  BDDC:
    PC Type                            : shell
  Performance:
    Pointwise Error (max)              : 4.185030e-02
    CG Solve Time                      : 0.0171467 (0.0171467) sec
```

jeremylt:
I have chased one of the problems into the simultaneous diagonalization.

Given `M` and `K`, strangely `X^T M X = I` but `X^T K B != Lambda`

jeremylt:
ahhh, something screwy is going on with the singular value
```
Error in diagonalization of K [2, 2]: 6.666667 != 15.000000
Error in diagonalization of K [2, 3]: -7.453560 != 0.000000
Error in diagonalization of K [3, 2]: -7.453560 != 0.000000
Error in diagonalization of K [3, 3]: 8.333333 != 0.000000
```

jeremylt:
"Matrix Computations" seems to think that reordering the intermediate symmetric QR to put eigenvalues in ascending order may help stability. But I don't know that will fix this issue

tests/t541-operator.c
```diff
@@ -0,0 +1,341 @@
...
+    CeedVectorSetValue(V, 0.0);
+    CeedVectorGetArray(V, CEED_MEM_HOST, &v);
+    v[0]     = v_Pi[0];
+    v[P-1]   = v_Pi[1];
+    v[P*P-P] = v_Pi[2];
+    v[P*P-1] = v_Pi[3];
+    CeedVectorRestoreArray(V, &v);
+
+    // -- Apply full FDM inverse again
+    CeedOperatorApply(op_inv, V, U, CEED_REQUEST_IMMEDIATE);
```
jeremylt, line 304: Ok, this is a single element test for scalar diffusion that replicates the mapping from the FDM element inv to the A_rr inv. It's currently failing, so either there is a problem with the math, a problem with the FDM code, or a problem with the test. That narrows it down :)

backends/ref/ceed-ref-operator.c
```diff
@@ -1089,39 +1089,39 @@ int CeedOperatorCreateFDMElementInverse_Ref(CeedOperator op,
...
-  ierr = CeedMatrixMultiply(ceed, (const CeedScalar *)work,
-                            (const CeedScalar *)interp_1d, mass, P_1d, P_1d, Q_1d);
-  CeedChkBackend(ierr);
+  for (CeedInt i=0; i<P_1d; i++)
+    for (CeedInt j=0; j<P_1d; j++) {
+      CeedScalar sum = 0.0;
+      for (CeedInt k=0; k<Q_1d; k++)
+        sum += interp_1d[k*P_1d+i]*q_weight_1d[k]*interp_1d[k*P_1d+j];
+      mass[i+j*P_1d] = sum;
+    }
```
jeremylt, line 1109: This helped a bit, but there is still a small difference that's resulting in error being ~1000x larger

jeremylt:
Now the accuracy seems to depend upon the compiler. That's bad

jeremylt:
Ok, so optimization levels and multiplication associativity change the accuracy. Fun.

jeremylt:
This branch contains bugfixes that should be included for v0.8.1

include/ceed/backend.h
```diff
@@ -32,6 +32,39 @@
...
+  This macro provides the ablitiy to disable optimization flags for functions that
+  are sensitive to floting point optimizations.
+**/
+#ifndef CeedPragmaOptimizeOff
+#  if defined(__clang__)
+#    define CeedPragmaOptimizeOff _Pragma("clang optimize off")
+#  elif defined(__GNUC__)
+#    define CeedPragmaOptimizeOff _Pragma("GCC push_options") _Pragma("GCC optimize 0")
+#  elif defined(__INTEL_COMPILER)
+#    define CeedPragmaOptimizeOff _Pragma('optimize("", off)')
```
jedbrown, line 46: Can single quote possibly be correct here?

interface/ceed-basis.c
```diff
@@ -1303,82 +1307,89 @@ int CeedSymmetricSchurDecomposition(Ceed ceed, CeedScalar *mat,
...
-      v[j] = matT[i+n*(j+1)];
-      matT[i+n*(j+1)] = 0;
+    if (tau[i] > CEED_EPSILON) {
+      v[i] = 1;
+      for (CeedInt j=i+1; j<n-1; j++) {
+        v[j] = mat_T[i+n*(j+1)];
+        mat_T[i+n*(j+1)] = 0;
+      }
+      CeedHouseholderReflect(&mat[(i+1)+n*(i+1)], &v[i], tau[i],
+                             n-(i+1), n-(i+1), n, 1);
```
jedbrown, line 1359: What is this conditional intended to do? If `v[]` contains large entries, then `tau` would be very small, and yet skipping the reflection doesn't seem right. Note that `tau v^T v = 2` so if v has size 1e10, then tau has size 2e-20.

interface/ceed-basis.c
```diff
@@ -1303,82 +1307,89 @@ int CeedSymmetricSchurDecomposition(Ceed ceed, CeedScalar *mat,
...
     //   norm = sqrt(v[i]*v[i] + sigma) / v[i];
     //   tau = 2 / (norm*norm)
-    if (sigma > 10*CEED_EPSILON)
+    if (sigma > CEED_EPSILON)
       tau[i] = 2 * v[i]*v[i] / (v[i]*v[i] + sigma);
     else
       tau[i] = 0;
 
+    // Don't apply Householder reflector if row already diagonal
+    if (sigma < CEED_EPSILON) continue;
```
jedbrown, line 1327: Why not? It's harmless and consistent.

jeremylt: It was adding small (~e-15/16) noise that was making the Python/C differences even larger

jeremylt: Woops, no this is for fixing #645

interface/ceed-basis.c
jeremylt: This is the one that was added as an effort to reduce noise. I pushed a change.

include/ceed/backend.h
jeremylt: No, it can't. The intel job compiled, so perhaps that job does not believe that `__INTEL_COMPILER` is defined?

jeremylt: I think its ok now

jeremylt:
I've rebased the commits to clean up the history

interface/ceed-basis.c
```diff
@@ -1249,20 +1249,23 @@ int CeedQRFactorization(Ceed ceed, CeedScalar *mat, CeedScalar *tau,
...
+    CeedScalar R_ii = -copysign(norm, v[i]);
+    v[i] -= R_ii;
+    mat[i+n*i] = R_ii;
     // norm of v[i:m] after modification above and scaling below
     //   norm = sqrt(v[i]*v[i] + sigma) / v[i];
     //   tau = 2 / (norm*norm)
     tau[i] = 2 * v[i]*v[i] / (v[i]*v[i] + sigma);
 
+    // Don't apply Householder reflector if row already diagonal
+    if (sigma < CEED_EPSILON) continue;
```
jedbrown, line 1261: I think you mean that the column is already diagonal, but I'm still not convinced this is good or desirable. Perhaps it's harmless for our unblocked algorithm, but it's unusual and breaks the mathematical purity of using computed factors, including the semantics of blocking. (Sure, we may not plan to implement blocked/CompactWY methods.) My concern is not so much that it's wrong as written for the matrices we're working with, but that it's no better and thus spurious complexity that someone down the road will grapple with and after painstaking testing, rip out.

From the definition of `sigma` up above, I think this condition can trigger if you have a well-behaved non-triangular matrix that has been multiplied by 1e-15. I.e., to work with general matrices, any such tolerance must be suitably scaled relative to the norm of the column (acknowledging that `sigma` has been squared). And I still don't think this logic should be needed.

jeremylt: If I just remove it, 7 tests fail with
```
/home/jeremy/Dev/libCEED/interface/ceed-basis.c:1422 in CeedSymmetricSchurDecomposition(): Symmetric QR failed to converge
```

jedbrown: How does it factor `[2e-10   -1e-10; -1e-10  2e-10]`?

jeremylt: It makes a disaster
```
7.071068e-01	5.000000e-01	5.000000e-01	0.000000e+00	
-7.071068e-01	5.000000e-01	5.000000e-01	0.000000e+00	
0.000000e+00	-5.000000e-01	5.000000e-01	-7.071068e-01	
0.000000e+00	-5.000000e-01	5.000000e-01	7.071068e-01
```

jeremylt: Currently, what we have in main is broken, as it takes in diagonal matrices and un-diagonalizes them.

jedbrown: I'm lost. Shouldn't the result be 2x2?

jeremylt: I'm trying to re-acquire my context from April. I don't remember the full series of changes and observations that lead me to this change.

I don't think (might be wrong though) this sigma check I added matters for the QR factorization, but it seems to matter for using the same algorithm from the QR factorization for the tridiagonalization inside of the symmetric Schur decomposition. But I'm not seeing why that  should make a difference; as far as I can tell, if it works ok for the QR then it should work ok as part of the symmetric Schur. So I figure that I must be missing something about this version of QR. Unfortunately, your modifications make it different enough from my copy of Golub and Van Loan that whatever is going wrong is not obvious to me.

But the problem in the tridiagonalization only shows up with a 4x4 matrix or larger, so I ran the 4x4 version of the test case you proposed. The test case did emit a correct decomposition that when multiplied together did in fact recreate the original test case matrix.

jedbrown: The example I gave is catastrophic if you run the old code (with the condition to skip the reflection). First I made the test print accurately so we can see.
<details>

```diff
diff --git i/tests/t301-basis.c w/tests/t301-basis.c
index b8de71ee..3d96152b 100644
--- i/tests/t301-basis.c
+++ w/tests/t301-basis.c
@@ -5,19 +5,17 @@
 
 int main(int argc, char **argv) {
   Ceed ceed;
-  CeedScalar qr[12] = {1, -1, 4, 1, 4, -2, 1, 4, 2, 1, -1, 0};
-  CeedScalar tau[3];
+  CeedScalar qr[4] = {2e-10, -1e-10, -1e-10, 2e-10};
+  CeedScalar tau[2];
 
   CeedInit(argv[1], &ceed);
 
-  CeedQRFactorization(ceed, qr, tau, 4, 3);
-  for (int i=0; i<12; i++) {
-    if (qr[i] <= 1E-14 && qr[i] >= -1E-14) qr[i] = 0;
-    fprintf(stdout, "%12.8f\n", qr[i]);
+  CeedQRFactorization(ceed, qr, tau, 2, 2);
+  for (int i=0; i<4; i++) {
+    fprintf(stdout, "%12.8e\n", qr[i]);
   }
-  for (int i=0; i<3; i++) {
-    if (tau[i] <= 1E-14 && tau[i] >= -1E-14) tau[i] = 0;
-    fprintf(stdout, "%12.8f\n", tau[i]);
+  for (int i=0; i<2; i++) {
+    fprintf(stdout, "%12.8e\n", tau[i]);
   }
   CeedDestroy(&ceed);
   return 0;
```
</details>

then (with the conditional early continue)
```console
$ build/t301-basis /cpu/self
-2.23606798e-10
-1.00000000e-10
-1.00000000e-10
-2.00000000e-10
1.89442719e+00
2.00000000e+00
```
and without
```console
$ build/t301-basis /cpu/self
-2.23606798e-10
1.78885438e-10
-2.36067977e-01
-1.34164079e-10
1.89442719e+00
2.00000000e+00
```
These are totally different and the first one doesn't represent an orthogonal matrix because `I - tau v v^T` is not a reflector when `tau=1.894427189` and `v=[1, -1e-10]`.

I don't understand why this fails for the diagonalization, but a correct QR should not early continue like this.

jedbrown: BTW, LAPACK skips the _last_ reflection (via `tau=0`) for square matrices because there's nothing below the diagonal.
```
julia> using LinearAlgebra
julia> A = [2 -1; -1 2] * 1e-10; tau = [0.,0]; LAPACK.geqrf!(A, tau)
([-2.23606797749979e-10 1.7888543819998312e-10; -0.23606797749978967 1.341640786499874e-10], [1.8944271909999157, 0.0])
```

interface/ceed-basis.c
```diff
@@ -1314,21 +1310,13 @@ int CeedSymmetricSchurDecomposition(Ceed ceed, CeedScalar *mat,
     }
     CeedScalar norm = sqrt(v[i]*v[i] + sigma); // norm of v[i:n-1]
     CeedScalar R_ii = -copysign(norm, v[i]);
-    v[i] -= R_ii;
```
jedbrown: I presume deleting this line was unintentional. Fixed in 3ceb3616063706ae53d944dabb70706b67b16aa8 and all tests pass locally.

jeremylt: Hmm, yes. That was an unintentional deletion.

jedbrown (approved):
Looks good to me now. Assuming you're ready, will you `git rebase -i` to clean up or would you prefer auto-squash merge?

jeremylt:
I was thinking of cleaning up the history again since there are a few different threads of work in the PR

jedbrown:
Sounds great, thanks.

jeremylt:
Should be good to merge once CI does its thing

backends/ref/ceed-ref-operator.c
```diff
@@ -1026,6 +1026,38 @@ static int CeedOperatorLinearAssembleAddPointBlockDiagonal_Ref(CeedOperator op,
...
+    for (CeedInt j=0; j<P_1d; j++) {
+      CeedScalar volatile sum = 0.0;
+      for (CeedInt k=0; k<Q_1d; k++)
+        sum += interp_1d[k*P_1d+i]*q_weight_1d[k]*interp_1d[k*P_1d+j];
+      mass[i+j*P_1d] = sum;
+    }
+  // -- Laplacian
+  for (CeedInt i=0; i<P_1d; i++)
+    for (CeedInt j=0; j<P_1d; j++) {
+      CeedScalar volatile sum = 0.0;
```
jedbrown, line 1049: `volatile` is almost never the answer, but is frequently part of magical lore. I'd prefer that it be removed.
```suggestion
      CeedScalar sum = 0.0;
```

backends/ref/ceed-ref-operator.c
```diff
@@ -1026,6 +1026,38 @@ static int CeedOperatorLinearAssembleAddPointBlockDiagonal_Ref(CeedOperator op,
...
+CeedPragmaOptimizeOff
+static int CeedBuildMassLaplace(const CeedScalar *interp_1d,
+                                const CeedScalar *grad_1d,
+                                const CeedScalar *q_weight_1d, CeedInt P_1d,
+                                CeedInt Q_1d, CeedInt dim,
+                                CeedScalar *mass, CeedScalar *laplace) {
+
+  for (CeedInt i=0; i<P_1d; i++)
+    for (CeedInt j=0; j<P_1d; j++) {
+      CeedScalar volatile sum = 0.0;
```
jedbrown, line 1041:
```suggestion
      CeedScalar sum = 0.0;
```

backends/ref/ceed-ref-operator.c
```diff
@@ -1171,27 +1194,33 @@ int CeedOperatorCreateFDMElementInverse_Ref(CeedOperator op,
...
+    for (CeedInt n=0; n<elem_size; n++) {
+      if (interp)
+        fdm_diagonal[c*elem_size + n] = 1.0;
+      if (grad)
+        for (CeedInt d=0; d<dim; d++) {
+          CeedInt i = (n / CeedIntPow(P_1d, d)) % P_1d;
+          fdm_diagonal[c*elem_size + n] += lambda[i];
+        }
+      if (fabs(fdm_diagonal[c*elem_size + n]) < 1000*CEED_EPSILON)
+        fdm_diagonal[c*elem_size + n] = 1000*CEED_EPSILON;
```
jedbrown, line 1209: This is internal and for a specific purpose, but I don't love thresholding based on absolute values instead of relative values. Changing the entries based on choice of units or mesh resolution shouldn't cause qualitatively different local behavior. Thresholding could be based on max value or some other relative quantity.

jedbrown (approved):

backends/ref/ceed-ref-operator.c
```diff
@@ -1168,13 +1168,13 @@ int CeedOperatorCreateFDMElementInverse_Ref(CeedOperator op,
   ierr = CeedVectorGetArrayRead(q_weight, CEED_MEM_HOST, &q_weight_array);
   CeedChkBackend(ierr);
   ierr = CeedCalloc(num_elem, &elem_avg); CeedChkBackend(ierr);
+  const CeedScalar qf_value_bound = max_norm*100*CEED_EPSILON;
```
jeremylt, line 1171: What about this relative value as the cutoff in both places @jedbrown? In practice, due to the perturbation, I don't know that we actually get fdm diagonal values that are below that threshold now.

jedbrown: I think this relative bound would be fine, perhaps with size-dependent defaults. For example, SciPy uses `max(m,n) * eps`: https://github.com/scipy/scipy/blob/v1.7.0/scipy/linalg/basic.py#L1330-L1337

jeremylt: In our case that would be something like `sqrt(elem_size)*CEED_EPSILON`

jedbrown: Looks good to me.

PR #750, opened 2021-04-19
qf - allow identity qfunctions to have in and out EVAL_NONE

jeremylt:
I found a use case for this


jeremylt:
An issue may be showing up for qfunctions with both modes none. Will investigate

jeremylt:
I added operator side changes to accommodate this.

jeremylt:
Note: I did not add anything for handling the code generation backends - I just error as unsupported in this case. Eventually I'd like to add proper support of identity QFs in the code generation backends, and we can bundle this update in there as well.

jeremylt:
somewhere around PR #716 Nek tests stopped running on Noether, but they still work fine locally for me

jedbrown (approved):

PR #751, opened 2021-04-20
Allow number of quad points per element to be set for Op

jeremylt:
I found a use case for all collocated bases - restricting into the broken space for BDDC. In order to make that work though, we need to be able to set the number of "quadrature points" per element. See also #750

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -228,7 +228,8 @@ static int CeedOperatorCheckReady(CeedOperator op) {
     if (op->num_qpts == 0)
       // LCOV_EXCL_START
       return CeedError(ceed, CEED_ERROR_INCOMPLETE,
-                       "At least one non-collocated basis required");
+                       "At least one non-collocated basis is required "
+                       "or the number of quadrature points must be set");
     // LCOV_EXCL_STOP
   }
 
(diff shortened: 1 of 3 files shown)


interface/ceed-operator.c
```diff
@@ -2164,6 +2165,35 @@ int CeedOperatorCreateFDMElementInverse(CeedOperator op, CeedOperator *fdm_inv,
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Set the number of quadrature points associated with a CeedOperator.
+           This should be used when creating a CeedOperator where every
+           field has a collocated basis.
```
jedbrown, line 2171: Is it better to do this or to have a sort of phantom `CeedBasis` that only specifies the number of quadrature points?

In any case, we should be clear that this concept does not apply to composite operators.

jedbrown (approved):

interface/ceed-operator.c
jeremylt: One way or another this is a sort of strange case in our current model
```
  CeedQFunctionCreateIdentity(ceed, num_comp_u, CEED_EVAL_NONE, CEED_EVAL_NONE,
                              &qf_identity);
  CeedOperatorCreate(ceed, qf_identity, CEED_QFUNCTION_NONE, CEED_QFUNCTION_NONE,
                     &op_inject_r);
  CeedOperatorSetField(op_inject_r, "input", data_fine->elem_restr_u,
                       CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
  CeedOperatorSetField(op_inject_r, "output", elem_restr_r, CEED_BASIS_COLLOCATED,
                       CEED_VECTOR_ACTIVE);
  CeedOperatorSetNumQuadraturePoints(op_inject_r, elem_size);
```
To me, it looks more descriptive to say that the basis actions are both none and collocated to emphasize that this is a restriction-only operation. But that approach involves this extra call that isn't exactly intuitive in its naming (but is consistent)

jeremylt:
CI settings are strange again - GitHub won't let me merge

PR #752, opened 2021-04-21
python - add pointwisemult and axpy

jeremylt:


jeremylt:
@DiffeoInvariant GitHub isn't letting me add you to review, but here's the PR.

DiffeoInvariant:
@jeremylt I can't review because I don't have write access to the libCEED repo, but this looks good to me

jeremylt:
CUDA and HIP impls added

backends/cuda/ceed-cuda-vector.c
```diff
@@ -441,7 +441,6 @@ static int CeedVectorReciprocal_Cuda(CeedVector vec) {
   case CEED_CUDA_BOTH_SYNC:
     ierr = CeedDeviceReciprocal_Cuda(data->d_array, length); CeedChkBackend(ierr);
     data->memState = CEED_CUDA_DEVICE_SYNC;
-    ierr = CeedVectorSyncArray(vec, CEED_MEM_HOST); CeedChkBackend(ierr);
```
jeremylt, line 443: I added this when I first added the device versions of the vector reciprocal function. (It is actually buggy as it is in main; needs line 443, which I added in previous commit)

I'm dropping this sync in this commit - we generally have a philosophy of not forcing a sync unless we need it, so I see no reason to maintain `CEED_*_BOTH_SYNC` in this case. If the data needs to be on the host, it can be pulled down there when needed.

Thoughts @YohannDudouit @nbeams?

DiffeoInvariant (approved):
This looks good to me! I've never written CUDA code myself so I probably shouldn't sign off on that, but the Python and other small changes look good to me

backends/cuda/ceed-cuda-vector.c
nbeams: Makes sense to me.

backends/hip/kernels/hip-vector.hip.cpp
```diff
@@ -67,3 +67,59 @@ extern "C" int CeedDeviceReciprocal_Hip(CeedScalar* d_array, CeedInt length) {
...
+    return;
+  y[idx] += alpha * x[idx];
+}
+
+//------------------------------------------------------------------------------
+// Compute y = alpha x + y on device
+//------------------------------------------------------------------------------
+extern "C" int CeedDeviceAXPY_Hip(CeedScalar *y_array, CeedScalar alpha,
+    CeedScalar *x_array, CeedInt length) {
+  const int bsize = 512;
```
nbeams, line 89: This line reminds me that I probably shouldn't have left all the vector blocksizes at 512 for HIP after we discovered the issue with the compiler assuming 256 threads/block.  And now we don't match either the 4.0 assumption (256) or 4.1 assumption (1024), so perhaps this is something I should look into more...though with such simple kernels, it's unlikely to cause failures (we're not exactly taxing the register or shared memory usage here), I guess there's still a chance performance could change.  I did check the detailed compiler output from the `--save-temps` flag (something Scott told me about) and the result with/without adding `__launch_bounds(512)` is identical.

So I guess I'd be fine with leaving it for now, but we could also switch to 256 or add `__launch_bounds__` to all the kernels in this file for completeness.  (Since 4.1 assumes 1024, there will no longer be the potential for issues with errors like were seen previously.)   Or we could leave this for a future PR since we might want to add launch bounds other places after the change in 4.1; I've currently been investigating this quite a bit with the hipMAGMA backend because the results with/without launch bounds even with 4.0 are not what I expected (haven't been able to try 4.1 yet).

nbeams (approved):
Looks good to me.  

One thing I happened to notice that isn't part of this PR (but this PR may be a good place to fix it) is the typo "poinwise" in the description comments at the top of `t122-vector.c`.

backends/hip/kernels/hip-vector.hip.cpp
jedbrown: Should I upgrade Noether to 4.1? I left it at 4.0 because Scott said there were known issues and I didn't have enough context to know whether those were likely to impact us.

nbeams: I'm not sure -- I was under the impression that the issues would maybe affect performance more than cause failures -- were you thinking the same?  Maybe we can ask Scott for more clarification.  I don't know what all the users of Noether need, either... I mostly just test/debug (and am very grateful for the access on days like today, when the machine I use here at ICL has been busy all day with others' tests) but obviously I don't know what everyone else uses Noether for.

jedbrown: If it's pure performance, then not a big deal. We sometimes test performance, but don't put any particular confidence on it.

jeremylt:
I fixed the typo. I'll merge once CI finishes.

YohannDudouit (approved):

PR #753, opened 2021-04-22
CI - unset PETSC_DIR

jeremylt:


jeremylt:
Minor change to CI - at some point CI started picking up on the env variable `PETSC_DIR`. I unset so we keep the isolated core library tests.

I also added a few debugging outputs because I thought I saw something strange in the logs. It was just funny output before, but this new output is useful, so I left it.

jeremylt:
The issue with the PETSC_DIR variable becoming set in the CI environment disappeared again, but I think it's still probably a good idea to merge this so that we make sure that we aren't running those tests twice.

jedbrown (approved):

PR #754, opened 2021-04-23
Fluids - Refactor

LeilaGhaffari:
Tasks:
- [x] Use `PetscFunctionList` for different problems
- [x] Add a regression test for `EULER_VORTEX` 
- [x] Refactor the miniapp following our PETSc examples
- [x] Fix C style
- [x] Enable use of GPU vectors
- [x] Change the default BCs for `ADVECTION(2D)` from `slip` to `wall` and modify the corresponding regression tests accordingly.
- [x] Dedicate a separate section in `print problem summary` for each problem
- [x] Change some of the problem specific command line arguments (e.g. we can make `problem_advection_wind_translation` shorter)
- [x] More cleanup ...
- [x] Turn on some HIP/CUDA tests


LeilaGhaffari:
I am done with the plan I had envisioned for the refactoring. I don't understand what is going on with the GPU vectors. @jeremylt and @jedbrown  , could any of you give a look? Thanks!

valeriabarra:
Thank you @LeilaGhaffari for all of this work! I think this refactoring is going to be very helpful!

I'm trying to build and test your branch for the code review, but I am seeing a bunch of unused variable and function compilation warnings. For example,
```
~/CEED/libCEED/examples/fluids$ make -B PETSC_ARCH=$PETSC_ARCH PETSC_DIR=$PETSC_DIR
          CC build/navierstokes.o
In file included from /home/valeria/CEED/libCEED/examples/fluids/navierstokes.h:18,
                 from /home/valeria/CEED/libCEED/examples/fluids/navierstokes.c:55:
/home/valeria/CEED/libCEED/examples/fluids/qfunctions/eulervortex.h: In function ‘Exact_Euler’:
/home/valeria/CEED/libCEED/examples/fluids/qfunctions/eulervortex.h:92:44: warning: unused variable ‘z’ [-Wunused-variable]
   92 |   const CeedScalar x     = X[0], y = X[1], z = X[2]; // Coordinates
```
and so on for others. 

Which compiler do you usually test this with? 
I'm using `gcc 9.3.0` . Perhaps adding compilation flags such as  `-Wunused-variable` and `-Wunused-function` would help identifying these types of issues? Should I have done anything different than the usual? (i.e., `make -B`). 
Thanks

LeilaGhaffari:
Thanks @valeriabarra for pointing that out. I just fixed them. Some headers were included in more places than they need to be. There were some other warnings that had been hidden in ~2K lines of warnings that are fixed now!

examples/fluids/qfunctions/densitycurrent.h
```diff
@@ -32,47 +32,6 @@
 #define M_PI    3.14159265358979323846
 #endif
 
-#ifndef setup_context_struct
```
jeremylt, line 35: Any structs used in the QFunction must be defined in the QFunction source file or the Qfunctions cannot run on the GPU

LeilaGhaffari: Yep! I just fixed it. Thanks but what is the reason?

jeremylt: The entire QFunction source file is copied into the CUDA or HIP kernel that is generated at runtime, so those files need to contain every typedef the kernel will need to compile.

jedbrown:
Does anyone else get this sort of error?
```
$ build/fluids-navierstokes -test -degree 3 -dm_plex_box_faces 1,1,2 -units_kilogram 1e-9 -lx 125 -ly 125 -lz 250 -center 62.5,62.5,187.5 -rc 100. -thetaC -35. -ts_dt 1e-3 -compare_final_state_atol 1E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-dc-explicit.bin
[0]PETSC ERROR: Unexpected data in file
[0]PETSC ERROR: Vector in file different size (416) than input vector (224)
```

LeilaGhaffari:
> Does anyone else get this sort of error?
I don't get this error. I  also tried it on Rezgar's machine (after checking out the branch afresh) and it passed there as well. I don't see why this particular test should fail since I never changed this file in this branch nor any configurations related to this test! It's confusing! IDK!

examples/fluids/src/setuplibceed.c
```diff
@@ -549,7 +549,7 @@ PetscErrorCode SetupContextForProblems(Ceed ceed, CeedData ceed_data,
     CeedQFunctionContextCreate(ceed, &ceed_data->dc_context);
     CeedQFunctionContextSetData(ceed_data->dc_context, CEED_MEM_HOST,
                                 CEED_USE_POINTER,
-                                sizeof phys->dc_ctx, phys->dc_ctx);
+                                sizeof(*phys->dc_ctx), phys->dc_ctx);
```
jeremylt, line 552: @jedbrown Good guess yesterday

jedbrown: Was this enough to get it working on GPUs? @LeilaGhaffari What does your simplest GPU failure look like?

jeremylt:
Density current won't run for me
```
./navierstokes -ceed /gpu/hip
/projects/jeremy/libCEED/backends/hip/ceed-hip-compile.cpp:95 in CeedCompileHip(): HIPRTC_ERROR_COMPILATION
/tmp/comgr-736eb4/input/CompileSource:208:5: error: no matching function for call to 'Exact_DC'
    Exact_DC(3, 0., x, 5, q, ctx);
    ^~~~~~~~
/tmp/comgr-736eb4/input/CompileSource:143:12: note: candidate function not viable: call to __host__ function from __device__ function
static int Exact_DC(CeedInt dim, CeedScalar time, const CeedScalar X[],
           ^
1 error generated when compiling for gfx906.
Error: Failed to compile opencl source (from CL or HIP source to LLVM IR).

Aborted
```

Turns out that the hip code generation ignores helper functions defined in the same source file as the QFunctions, unlike CUDA.

jeremylt:
But the confusing thing is that CI on Noether seems to be indicating that the fluids test suite is running fine
```
$ make -k -j$NPROC_GPU BACKENDS="$BACKENDS_GPU" junit search="petsc fluids solids"
make: 'lib' with optional backends: /gpu/hip/occa /gpu/hip/ref /gpu/hip/shared /gpu/hip/gen /gpu/hip/magma /gpu/hip/magma/det
        TEST petsc-area
        TEST petsc-bps
        TEST petsc-bpsraw
        TEST petsc-bpssphere
        TEST petsc-multigrid
        TEST fluids-navierstokes
        TEST solids-elasticity
```

jedbrown:
What toolchain version are you using? Do we need a macro to add `__device__` for those helpers (all of which I think we intend for the compiler to inline as part of its vectorization)?

jeremylt:
Ah, that is because these tests are completely skipped for HIP. We need to see if we can get some of these tests to run on the GPU in CI.

jeremylt:
> What toolchain version are you using? Do we need a macro to add `__device__` for those helpers (all of which I think we intend for the compiler to inline as part of its vectorization)?

Yeah, it looks like we need to add a macro for this. The code seems to run ok when I manually add `__device__`.

This macro may want to expand to `__device__ __forceinline__`

jeremylt:
It needs more work, but I added a first version of a helper function macro

jeremylt:
Ok, the macro should work now.

jeremylt:
@LeilaGhaffari, which test cases should we try to reproduce the errors you were seeing? I think updating the QFunction context data size and fixing the helper function macros may have fixed the problems.

LeilaGhaffari:
Thanks, @jeremylt.
I can't run the tests right now. I will be able to be more responsive in evening. But you can try the regression tests in `navierstokes.c`, especially those for `advection` that would crash immediately! `density_current` and `euler_vortex` were returning some norms.

jeremylt:
The first advection test case runs clean for me
```
./navierstokes -ceed /gpu/hip -degree 3 -dm_plex_box_faces 1,1,2 -units_kilogram 1e-9 -lx 125 -ly 125 -lz 250 -center 62.5,62.5,187.5 -rc 100. -thetaC -35. -ksp_atol 1e-4 -ksp_rtol 1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3 -implicit -ts_type alpha  -compare_final_state_atol 5E-4 -compare_final_state_filename tests-output/fluids-navierstokes-dc-implicit-stab-none.bin -test
```

jeremylt:
Ok, I think this is almost ready. Thanks for all the hard work @LeilaGhaffari!

The last open ToDo in my eyes is turning on some tests for the GPU backends. Right now every fluids test is excluded for GPU backends in `tests/junit.py` and `tests/tap.sh`.

LeilaGhaffari:
Thanks, @jeremylt for your help.

Yes, triggering the tests has been one of the ToDo's in my head. I can try this evening. Also, I would prefer to create the coordinate vector once. I can fix that too.

backends/cuda/ceed-cuda-qfunction-load.cpp
```diff
@@ -77,6 +77,7 @@ extern "C" int CeedCudaBuildQFunction(CeedQFunction qf) {
 
   // Defintions
   code << "\n#define CEED_QFUNCTION(name) inline __device__ int name\n";
+  code << "#define CEED_QFUNCTION_HELPER(name) inline __device__ int name\n";
```
jedbrown, line 80: Shall we just make it
```c
#define CEED_QFUNCTION_HELPER inline __device__
```
so "helpers" can return something other than `int`?

examples/solids/qfunctions/FSCurrent-NH2.h
```diff
@@ -65,7 +66,8 @@ static inline CeedScalar log1p_series_shifted(CeedScalar x) {
   sum += y / 5;
   y *= y2;
   sum += y / 7;
-  return 2 * sum;
+  *log1p_shifted = 2 * sum;
```
jedbrown, line 69: That's the issue here, but I don't think there's a reason the macro needs to capture the name. We needed it for `CEED_QFUNCTION` because we had to create the location string in addition to the function.

backends/cuda/ceed-cuda-qfunction-load.cpp
jeremylt: It is my understanding that HIP sometimes ignores `__inline__` unless we force it. I was 50-50 on making the helper return an error code or making backends force inlining. I suppose it is easier to specify behavior on backend devs.

examples/solids/qfunctions/FSCurrent-NH2.h
jeremylt: Good point with the name.

jeremylt:
Hmm, this doubles how long CI takes on Noether.

jedbrown:
Where is most of that time spent? JIT? We don't do a lot of iterations, right?

jeremylt:
I believe so; JIT seemed to take a while when I was running the tests manually. Only 8 processes running a large number of test cases with slow JIT = 12ish minutes to run the fluids test suite on HIP

examples/fluids/src/setuplibceed.c
```diff
@@ -510,7 +512,8 @@ PetscErrorCode SetupLibceed(Ceed ceed, CeedData ceed_data, DM dm, User user,
   // CEED Operator Apply
   // *****************************************************************************
   // -- Apply CEED Operator for the geometric data
-  CeedOperatorApply(ceed_data->op_setup_vol, ceed_data->x_coord, ceed_data->q_data,
+  CeedOperatorApply(ceed_data->op_setup_vol, ceed_data->x_coord,
+                    ceed_data->q_data,
```
jeremylt, line 516: Every once in a while, `make style` results in awkward breaks like this one.

LeilaGhaffari: Ouch! Thank you for the catch!

LeilaGhaffari:
Is there a reason I don't see `cuda` backends in the CI list of tests?

LeilaGhaffari:
> We don't do a lot of iterations, right?

Right! There are only 10 time steps!

jeremylt:
> Is there a reason I don't see `cuda` backends in the CI list of tests?

Correct. We do not currently have CI for CUDA, only HIP.

LeilaGhaffari:
Ah, okay! Were you able to run the tests with cuda on Noether. I can try it this evening or tomorrow morning but `Advection(2d)` tests would still crash on my local machine with `/gpu/cuda/gen`.

jeremylt:
Noether has no CUDA. I ran CUDA on my local machine and HIP on Nother and everything seems ok.

LeilaGhaffari:
Hmm! Weird! I am glad to hear it though! Thanks!

jedbrown:
I think it's okay and we can live with it being slow for now. We could consider making it possible to solve multiple problem instances within the same execution (similar to the batches of runs we can do with `petsc-bps -degree 3,4,5,7,9`), but hip/gen will still need to recompile for each operator so maybe the gains are modest. We could have caching based on a hash of the sources, but I think this is premature optimization for _us_ to work on.

jedbrown:
And correct, we don't have CUDA CI, but we have a CUDA GPU in Qin Lv's machine and I plan to add CUDA CI there. It only has a few cores and her students use it for ML/data mining so we'll want to throttle resources. Maybe just limit ourselves to using our one GPU and run only the CUDA tests rather than everything.

examples/solids/qfunctions/FSCurrent-NH1.h
```diff
@@ -74,7 +74,8 @@ static inline CeedScalar log1p_series_shifted(CeedScalar x) {
 // -----------------------------------------------------------------------------
 #ifndef DETJM1
 #define DETJM1
-static inline CeedScalar computeJM1(const CeedScalar grad_u[3][3]) {
+CEED_QFUNCTION_HELPER static CeedScalar computeJM1(const CeedScalar
```
jedbrown, line 77: one ordering

examples/solids/qfunctions/FSCurrent-NH1.h
```diff
@@ -88,9 +89,10 @@ static inline CeedScalar computeJM1(const CeedScalar grad_u[3][3]) {
 // -----------------------------------------------------------------------------
 // Common computations between Ftau and dFtau
 // -----------------------------------------------------------------------------
-static inline int commonFtau(const CeedScalar lambda, const CeedScalar mu,
-                             const CeedScalar grad_u[3][3], CeedScalar F_inv[3][3],
-                             CeedScalar tau_work[6], CeedScalar *llnj) {
+static int CEED_QFUNCTION_HELPER commonFtau(const CeedScalar lambda,
```
jedbrown, line 92: another ordering

examples/solids/qfunctions/FSCurrent-NH2.h
```diff
@@ -44,7 +44,7 @@ struct Physics_private {
 // -----------------------------------------------------------------------------
 #ifndef LOG1P_SERIES_SHIFTED
 #define LOG1P_SERIES_SHIFTED
-static inline CeedScalar log1p_series_shifted(CeedScalar x) {
+static CEED_QFUNCTION_HELPER CeedScalar log1p_series_shifted(CeedScalar x) {
```
jedbrown, line 47: a third ordering

I think we can put the `static` inside the definition of `CEED_QFUNCTION_HELPER` and always write `CEED_QFUNCTION_HELPER return_type function_name(...)`.

include/ceed/ceed.h
```diff
@@ -67,6 +67,16 @@
...
 #endif
 
+/**
+  @ingroup CeedQFunction
+  This macro populates the correct function annotations for User QFunction
+    helper function source for code generation backends or populates default
+    values for CPU backends.
+**/
+#ifndef CEED_QFUNCTION_HELPER
+#define CEED_QFUNCTION_HELPER inline
```
jedbrown, line 77:
```suggestion
#define CEED_QFUNCTION_HELPER static inline
```

jedbrown (approved):
Looks good to me after resolving conflicts and the noted inconsistencies. Let's add CUDA tests in a new PR after this and #770 merges. Great work. I assume squash-merge is planned because there are so many commits, but if you want to split out separate commits for prerequisite refactoring (outside of `examples/fluids/`) that would be nicer for posterity.

jeremylt:
Good point, I cleaned up the macro.

jeremylt:
@LeilaGhaffari  can you resolve the merge conflicts? (from #769)

LeilaGhaffari:
> Looks good to me after resolving conflicts and the noted inconsistencies. Let's add CUDA tests in a new PR after this and #770 merges. Great work. I assume squash-merge is planned because there are so many commits, but if you want to split out separate commits for prerequisite refactoring (outside of `examples/fluids/`) that would be nicer for posterity.

Thanks! Yes, I was hoping to squash-merge since I went back and forth between multiple designs and never dropped/squashed commits. I am not sure if I can resolve merge conflicts if I split out separate commits but I can give it a try this evening. Sorry for mess!

LeilaGhaffari:
> #769



> @LeilaGhaffari can you resolve the merge conflicts? (from #769)

Yeah! Let me see.

jedbrown:
Cool. I'm not opposed to a single squash-merge. (It touches a lot, but only in a simple way outside examples/fluids.) You can just merge `main` into this branch and resolve the minor conflicts in solids, then we'll squash-merge.

LeilaGhaffari:
> Cool. I'm not opposed to a single squash-merge. (It touches a lot, but only in a simple way outside examples/fluids.) You can just merge `main` into this branch and resolve the minor conflicts in solids, then we'll squash-merge.

Done!

PR #735, opened 2021-04-09
Style Unification

jeremylt:
This includes style updates to all C code except

* PETSc examples (#726)
* Fluids example (@LeilaGhaffari  feel free to push to this branch, or make a separate one)
* CUDA, OCCA, HIP, and MAGMA backends (@YohannDudouit, @dmed256, @nbeams backend maintainers get to pick the style that is most ideomatic for their backends, feel free to add changes to this PR if you want to change the style in your backends as part of this effort)

I will probably do a check for funny line breaks later, but feel free to comment on any of the renaming.

This PR includes **zero** functionality change - style changes only!

PR #757, opened 2021-04-26
vec - add CeedVectorScale

jeremylt:
Another convenience vector method


python/tests/test-1-vector.py
```diff
@@ -289,6 +289,25 @@ def test_122(ceed_resource, capsys):
...
+    n = 10
+    x = ceed.Vector(n)
+
+    a = np.arange(10, 10 + n, dtype="float64")
+    x.set_array(a, cmode=libceed.COPY_VALUES)
+
+    x.scale(-0.5)
+    with x.array() as b:
+        for i in range(len(b)):
+            assert abs(b[i] + (10 + i) / 2) < 1e-14
```
jedbrown, line 309: This is likely sufficient.
```suggestion
        assert np.allclose(-.5 * a, b)
```

jedbrown (approved):
Rust binding?

jeremylt:
I updated the Python tests, added some Rust bindings, and fixed a Rust CI issue. We're currently having intermittent readthedocs issues, so I merging is currently blocked on this PR.

valeriabarra:
Nice. Should this also be added to `CeedVector.jl` ?

jeremylt:
The Julia bindings won't actually build locally for me. Let's see if CI thinks I did it right

valeriabarra:
> The Julia bindings won't actually build locally for me. Let's see if CI thinks I did it right

I had some problems with `LibCEED.jl` locally, too. Then, I feel less alone

jeremylt:
Because of the way the Julia bindings are tested, I looks like I can't just add development features. I'll let @pazner decide if/when/how to add these three convenience functions [CeedVectorScale, CeedVectorAXPY, CeedVectorPointwiseMult].

FFI in Julia seems to be far more painful than FFI in Python or Rust.

pazner:
> The Julia bindings won't actually build locally for me. Let's see if CI thinks I did it right

> I had some problems with LibCEED.jl locally, too. Then, I feel less alone

I could try to help with these issues if you have some more details to share.

pazner:
I added these new functions to LibCEED.jl.

I also added the distinction between "development tests" and "release tests" for the Julia interface. Release tests are those tests that use only the library features from the most recent release. That way, these tests are run both against the most recent binary build of libCEED_jll and against the current build in the CI.

On the other hand, development tests are tests for features which have been introduced *since* the latest release, and are not available in the binary build. These are run in the CI only with the current build of the library.

After each new release, the development tests can become release tests. This is documented in the release procedures in #725.

julia/LibCEED.jl/docs/src/CeedVector.md
```diff
@@ -15,4 +15,7 @@ witharray_read
 setarray!
 syncarray!
 takearray!
+scale!
+LinearAlgebra.axpy!(a::CeedScalar, x::CeedVector, y::CeedVector)
```
jeremylt, line 19: Do we want to make this `LinearAlgebra.axpy!` or `LibCEED.axpy!`?

pazner: `LinearAlgebra` (which is in the standard library) defines `axpy!`, so it makes sense for LibCEED to extend that definition for this case (with dispatch based on the types). The other alternative would cause a function name conflict.

This line just adds the function to the documentation.

This is similar to what is done already for `norm`, `det`, etc.

jeremylt: Will our axpy still work without them importing LinearAlgebra though?

julia/LibCEED.jl/test/rundevtests.jl
```diff
@@ -0,0 +1,33 @@
...
+        v3 = rand(n)
+
+        cv1 = CeedVector(c, v1)
+        cv2 = CeedVector(c, v2)
+        cv3 = CeedVector(c, v3)
+
+        alpha = rand()
+
+        scale!(cv1, alpha)
+        v1 .*= alpha
```
jeremylt, line 22: I think Julia will let us overload .*= if we want

julia/LibCEED.jl/docs/src/CeedVector.md
pazner: > Will our axpy still work without them importing LinearAlgebra though?

Yes, it should still work (it's exported, so it can be accessed as either `axpy!` or `LibCEED.axpy!`).

julia/LibCEED.jl/test/rundevtests.jl
pazner: It's actually not so easy to properly overload the broadcast operators, unfortunately.

It is possible to do, but there are some design decisions to be made, and it will likely have some limitations. I think keeping functions like `scale!` and `axpy!` explicit is a better choice, because it will also prevent users from writing accidentally slow code.

valeriabarra:
> > The Julia bindings won't actually build locally for me. Let's see if CI thinks I did it right
> 
> > I had some problems with LibCEED.jl locally, too. Then, I feel less alone
> 
> I could try to help with these issues if you have some more details to share.

Hi @pazner , yes, I haven't had the time to give it another try since my first attempt a while ago. I will try again and will report on that, if I still encounter issues. Thanks!

jedbrown:
Thanks @pazner. Do we need the generated file committed in the repository?

pazner:
> Thanks @pazner. Do we need the generated file committed in the repository?

I think that is the most straightforward solution. The reason the diff is so big in this PR is because of #735.

jedbrown:
Thanks. Is the issue that `Clang.jl` is a big dependency? Should we have CI re-generate the file and perform sanity checks? Can the "minor manual modifications" be automated, or made unimportant by implementing some conventions in `ceed.h`?

pazner:
One issue is that it would add a dependency on `Clang.jl`. I don't believe Julia's package manager has the concept of a "build-only" dependency, i.e. needed only for the build stage. Packages are discouraged from modifying their source directories, and generating the files would also require access to the C headers, which are not included with the Julia package. It would also somewhat increase the package install time.

Some of these issues could be worked around, but it's not clear that the benefit would be worth the (significant) added complexity. While it may not be the most "elegant" solution, I believe including the generated files in the repo is the most common approach for packages that wrap a C API.

I think the process could be automated (or at least made more automatic than it currently is). We could try to do some checks in the CI. Of course, these files (at least the parts of them that are used by the Julia interface) are tested with the unit tests, so we do have some level of confidence that they didn't get messed up somehow.

PR #758, opened 2021-04-28
Small Code Coverage Updates

jeremylt:


jedbrown (approved):
Seems fine. A test case or comment on rationale for the OCCA change would be helpful for posterity.

jeremylt:
Something is funny with the coverage for the 'fails successfully' tests right now. The modified t404 now checks for that  bug case, but at some point I'd like to add another the 'fails successfully' test to check that the context data was taken once I figure out why they aren't showing on coverage.

Issue #756, opened 2021-04-26
Review of the JOSS Paper


thelfer:
Dear libCEED developpers,

As part of the review of the JOSS paper (https://github.com/openjournals/joss-reviews/issues/2945), please find attached some remarks and questions.

The paper is well written and I was very pleased to review it and play with the library.  Installation is straighforward and the various tests I tried worked out of the box.

However, most of the remarks are related to the fact that the paper shall be made more self-explanatory, notably by picking some information already present in the introduction of the user material.

Do not hesitate to discuss the various questions/remarks and discard irrelevant points.

Regards,
Thomas Helfer

[review.pdf](https://github.com/CEED/libCEED/files/6376430/review.pdf)
[review.md](https://github.com/CEED/libCEED/files/6376431/review.md)

thelfer:
Note: for the question concerning boundary conditions, two boundary conditions are of general interest:

- Dirichlet boundary conditions
- Neuman boundary conditions

jeremylt:
Thank you for your review!

I added some additional content to the paper. We'll need to edit flow flow, content, size, etc, but I have initially addressed many of the comments.

---

1.1 - I added a comment on the two use cases as I see them - low level integration and high level use via a library such as MFEM (PETSc integration is still WIP so I de-emphasized it in the intro)
```
Users and library developers can integrate `libCEED` at a low level into existing applications in place of existing matrix-vector products without significant refactoring of their own discretization infrastructure.
Alternatively, users can utilize integrated `libCEED` support in MFEM [@MFEMlibrary; @mfem-paper].
```
We may wish to say more/edit

1.2 - I explained that users can supply their own bases
```
The discrete basis $B$ is the purely algebraic expression of a finite element basis (shape functions) and quadrature; it often possesses structure that is exploited to speed up its action.
Some constructors are provided for arbitrary order $H^1$ Lagrange bases with a tensor-product representation due to the computational efficiency of computing solution values and derivatives at quadrature points via tensor contractions.
However, the user can define a `CeedBasis` for arbitrary geometry elements, to include tetrahedra, by providing quadrature weights and the matrices used to compute solution values and derivatives at quadrature points from the DoFs on the element.
```

1.3 - Clarified libCEED's 'out of the box' MFEM integration
```
If MFEM is built with `libCEED` support, existing MFEM users can pass `-d ceed-cuda:/gpu/cuda/gen` to use a `libCEED` fastest CUDA backend, and similarly for other backends.
[...]

PR #762, opened 2021-04-28
Docs: Tiny rewording in Intro

valeriabarra:
Since this Introduction was mentioned lately in issue #756 , I was re-reading it and realized this sentence could have been improved. 
The use of the word `they` was a bit confusing in my opinion. 

 @jeremylt I just changed this sentence expanding its line, as recommended in #735 , although the original file had the older style (the max 80 char per line), but I wouldn't change the entire file style now. Do you agree?


jeremylt:
Re file style, agreed, it will be done en masse in a later PR

jedbrown (approved):

PR #763, opened 2021-05-03
Update Rust Github Action

jeremylt:
Bugfix allows us to use latest cargo tarpaulin again https://github.com/xd009642/tarpaulin/issues/747

A v0.19 should be coming soon, and we'll get that when it drops with this change


jedbrown (approved):
Is there a scenario where it may be important for us to "bisect" a GitHub Action, or would we always want to use the latest version here (modulo temporary pins due to bugs)?

jeremylt:
My thinking is that with 'support' functionality like this we're probably ok to use the lastest/fastest

PR #764, opened 2021-05-03
junit - set CEED_ERROR_HANDLER to exit

jeremylt:
Update error handler for junit to fix code coverage

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -27,7 +27,7 @@ noether-rocm:
     - echo "-------------- OCCA ----------------" && make -C $OCCA_DIR info
 # libCEED
     - make configure HIP_DIR=/opt/rocm OPT='-O -march=native -ffp-contract=fast'
-    - BACKENDS_CPU=$(make info-backends | grep -o '/cpu[^ ]*') && BACKENDS_GPU=$(make info-backends | grep -o '/gpu[^ ]*')
+    - BACKENDS_CPU=$(make info-backends-all | grep -o '/cpu[^ ]*') && BACKENDS_GPU=$(make info-backends | grep -o '/gpu[^ ]*')
     - echo "-------------- libCEED -------------" && make info
     - echo "-------------- BACKENDS_CPU---------" && echo $BACKENDS_CPU
     - echo "-------------- BACKENDS_GPU---------" && echo $BACKENDS_GPU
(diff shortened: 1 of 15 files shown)


jeremylt:
Bonus commit to fix us accidentally running tests 2x on Noether

jeremylt:
@jedbrown, the Nek5000 tests aren't running in CI, but they do run when I use the exact command in our bash script from `.gitlab-ci.yml` locally. I suspect there is some issue with the cached version of Nek5000. Can you delete the `Nek5000-19.0` folder in the CI cache? I can't see anything in the `projects/gitlab-runner` folder, so I assume some permissions magic is at work.

Makefile
```diff
@@ -591,7 +592,7 @@ prv : ;@$(MAKE) $(MFLAGS) V=$(V) prove
 prove-all :
 	+$(MAKE) prove realsearch=%
 
-junit-t% : BACKENDS += $(TEST_BACKENDS)
+junit-t% : BACKENDS = $(BACKENDS) $(TEST_BACKENDS)
```
jeremylt, line 595: I have no clue why this line would result in junit running each backend twice but this change does not.

jeremylt:
There is some parallel build problem with the petsc examples (which is only intermittent, as these things tend to do)

.gitlab-ci.yml
```diff
@@ -37,7 +37,8 @@ noether-rocm:
     - echo '[{"subject":"/","metrics":[{"name":"Transfer Size (KB)","value":"19.5","desiredSize":"smaller"},{"name":"Speed Index","value":0,"desiredSize":"smaller"},{"name":"Total Score","value":92,"desiredSize":"larger"},{"name":"Requests","value":4,"desiredSize":"smaller"}]}]' > performance.json
 #    Note: PETSC_DIR is set by default in GitLab runner env, unsetting to isolate core tests
     - export PETSC_DIR= PETSC_ARCH=
-    - make -k -j$NPROC_CPU BACKENDS="$BACKENDS_CPU" junit realsearch=%
+    - make -k -j$NPROC_CPU BACKENDS="$BACKENDS_CPU /cpu/self/tmpl /cpu/self/tmpl/sub" junit search=t
+    - make -k -j$NPROC_CPU BACKENDS="$BACKENDS_CPU" junit search=ex
```
jedbrown, line 41: What's the rationale for this manual split? I'm concerned that tests may get missed if any are introduced with a different naming convention.

jeremylt: This is because the ex can't be run with the tmpl backends but the t should be.

I've come back around to the idea that we should perhaps just remove those backends. The delegation mechanism is already tested by the CPU family of backends.

jeremylt: I'll merge this now to fix the CI and then save the question of removing the `/cpu/tmpl` backends (or leaving them as uncomplied examples? We have an example gallery QFunction, but I'm thinking we should perhaps remove it)

jedbrown (approved):
That's fine by me.

PR #767, opened 2021-05-20
Update GPU arch argument for hiprtc

nbeams:
Use `gcnArchName` instead of `gcnArch` for the `--gpu-architecture` argument of `hiprtc`. 
This change is required for ROCm 4.2, but is compatible with >= 3.6 (the README was updated from 3.5 to 3.6).

Thanks to @moes1 for figuring this out!


jeremylt (approved):
Looks reasonable and passes CI

jedbrown (approved):
Great, thanks for tracking this down.

PR #768, opened 2021-05-21
Remove Template Gallery and Backend Code

### LINKED ISSUE #260: Libceed_test and Ceed_Init
Travis defaults to `--as-needed` for gcc, which makes our current approach of making a separate `libceed_test.so` for `/cpu/self/tmpl` and `/cpu/self/tmpl/sub` somewhat delicate. We should create a `Ceed_Init_Test` to help with linking `libceed_test.so` more simply and proving more comprehensive test suite error messages.

jeremylt:
This PR does two things

- Remove QFunction gallery template - this is untested, uncompiled code, so I think this is an obvious win

- Remove template backends - I think the ref backends serve the same purpose, and removing these backends simplifies our makefile and testing

This could be used to fix the issue IDed in #764

closes #260


jeremylt:
Updated based on merge of #764

jedbrown (approved):

PR #771, opened 2021-05-25
Minor fixes

nbeams:
Two minor things I ran into today that I decided to combine into a little PR.

One is a typo introduced in #701.
The other is a potential string length issue introduced by #767 (my fault, sorry).

### DIFF
--- a/backends/hip/ceed-hip-compile.cpp
+++ b/backends/hip/ceed-hip-compile.cpp
@@ -45,7 +45,6 @@ int CeedCompileHip(Ceed ceed, const char *source, hipModule_t *module,
 
   // Macro definitions
   // Get kernel specific options, such as kernel constants
-  const int optslen = 32;
   const int optssize = 2;
   const char *opts[optssize];
   if (numopts > 0) {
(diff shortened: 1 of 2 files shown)


backends/hip/ceed-hip-compile.cpp
```diff
@@ -72,8 +71,9 @@ int CeedCompileHip(Ceed ceed, const char *source, hipModule_t *module,
   Ceed_Hip *ceed_data;
   ierr = CeedGetData(ceed, (void **)&ceed_data); CeedChkBackend(ierr);
   CeedChk_Hip(ceed, hipGetDeviceProperties(&prop, ceed_data->deviceId));
-  char buff[optslen];
   std::string archArg = "--gpu-architecture="  + std::string(prop.gcnArchName);
+  const int optslen = archArg.length() + 1;
+  char buff[optslen];
```
jedbrown, line 76: Hmm, variable length arrays are not allowed in C++ so this may be a portability problem. Meanwhile, gcc and clang allow them without warnings by default (including in `-std=c++11` dialect), but warn with `-Wvla` or `-Wpedantic`.

jedbrown (approved):

backends/hip/ceed-hip-compile.cpp
nbeams: Ack, good point.  I'd rather not have to overestimate the required size since I don't really know what the limit is (or could be in the future).  

What if I just do `opts[1] = archArg.c_str();` ?
`-Wpedantic` seems okay with that.

jedbrown: Great so long as `archArg` is guaranteed to live at least as long as `opts`. (Rust lifetime envy.)

nbeams: When you say "at least as long as", does the order of destruction matter?  Otherwise they should both go out of scope at the end of the function.

jedbrown: That's fine; order doesn't matter here because there's no destructor called in the `const char *opts`. It's just something whoever hacks on this next would need to keep in mind, but that's normal enough in C/C++.

nbeams:
Hmm, guess I should change the commit message now, it doesn't really make sense anymore...

nbeams:
Can I go ahead and merge this?

PR #772, opened 2021-05-28
doc - use docs.rs over self hosted documentation

jeremylt:


jedbrown (approved):
Cool. Could you also update the `documentation` in Cargo.toml?

PR #774, opened 2021-06-01
Check Release Notes

jeremylt:
It seems that GitHub actions forgot to implement a proper 'warn on failure' type mode
https://github.com/actions/toolkit/issues/399

This test should generate a hard failure if we forget to update the release notes.

### DIFF
--- a/.github/workflows/release-notes.yml
+++ b/.github/workflows/release-notes.yml
@@ -1,0 +1,20 @@
+name: Release Notes
+
+on:
+  push:
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - name: Environment setup
+      uses: actions/checkout@v2
+    - name: Check release notes
+      run: |
+        git fetch origin main
+        git diff origin/main --exit-code include/ceed/* && ! git diff origin/main --exit-code doc/sphinx/source/releasenotes.rst
--- a/doc/sphinx/source/releasenotes.rst
+++ b/doc/sphinx/source/releasenotes.rst
@@ -18,12 +18,15 @@ New features
 ^^^^^^^^^^^^
 
 * Add :c:func:`CeedVectorAXPY` and :c:func:`CeedVectorPointwiseMult` as a convenience for stand-alone testing and internal use.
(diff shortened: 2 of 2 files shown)


.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+    - name: Environment setup
+      uses: actions/checkout@v2
+    - name: Check release notes
+      env:
+        CC: ${{ matrix.compiler }}
+        FC: gfortran-9
+      run: |
+        git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin/main
+        git checkout main && git pull && git checkout -
```
jedbrown, line 24:
```suggestion
        git fetch origin main
```

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,25 @@
...
+        git diff main --exit-code include/ceed/* && ! git diff main --exit-code doc/sphinx/source/releasenotes.rst
```
jedbrown, line 25:
```suggestion
        git diff origin/main --exit-code include/ceed/* && ! git diff origin/main --exit-code doc/sphinx/source/releasenotes.rst
```

.github/workflows/release-notes.yml
jedbrown: Oh, or use this syntax in the fancier invocation above.

jeremylt: Thanks, this isn't a way I've used Git very much

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - name: Environment setup
+      uses: actions/checkout@v2
+    - name: Check release notes
+      env:
+        CC: ${{ matrix.compiler }}
+        FC: gfortran-9
```
jedbrown, line 21: I don't know if it saves time, but we don't need a Fortran compiler (or any kind of compiler) for this check. Just Git.

jedbrown (approved):

.github/workflows/release-notes.yml
jeremylt: Good point - fixed

PR #779, opened 2021-06-07
Jeremy/periodic bps

jeremylt:


examples/petsc/bps.c
```diff
@@ -163,6 +164,11 @@ static PetscErrorCode RunWithDM(RunParams rp, DM dm,
   ierr = MatShellSetOperation(mat_O, MATOP_GET_DIAGONAL,
                               (void(*)(void))MatGetDiag); CHKERRQ(ierr);
   ierr = MatShellSetVecType(mat_O, vec_type); CHKERRQ(ierr);
+  if (rp->bp_choice > 1) {
+    ierr = MatNullSpaceCreate(rp->comm, PETSC_TRUE, 0, NULL, &null_space);
+    CHKERRQ(ierr);
+    ierr = MatSetNullSpace(mat_O, null_space); CHKERRQ(ierr);
+  }
```
jeremylt, line 171: @jedbrown, I thought running a BP on a periodic mesh would be as simple as using `-dm_plex_box_bd periodic,periodic,periodic` and setting the null space to the constant vector, but I am getting `DIVERGED_INDEFINITE_MAT` (and `DIVERGED_INDEFINITE_PC` for bp1/2). Am I missing something obvious here?

jeremylt:
Side note, using `-dm_plex_box_bd periodic,periodic,periodic` sets the *cells* as periodic rather than the faces, which is annoying when paired with the fact that for `bps.c` we removed the ability to directly set the number of cells in each direction.

examples/petsc/bps.c
jedbrown: There used to be an issue in with tensor indices for periodic meshes, but @knepley has been hacking on this recently and it might have gone away. (I got part way through a somewhat tedious workaround in a branch that I think should be obsolete now.)

knepley: It should be easy to check. Can you pull it down to Q1 and view the matrix?

knepley:
@jeremylt I do not understand the distinction between periodic cells and faces. Maybe I can make the interface easier if I understand.

jeremylt:
Really, I should have said *element* instead of *cell*. But I think I was misunderstanding the information I was seeing.

examples/petsc/bps.c
jeremylt: When I use Q1 and grab a a row out of the center of the assembled matrix with Dirichlet BCs, I get
```
row 32: (16, 2.18331e-16)  (17, -0.0333333)  (20, -0.0333333)  (21, -0.0166667)  (32, 0.533333)  (33, -2.00809e-17)  (36, 2.08309e-17)  (37, -0.0333333)  (48, 2.18696e-16)  (49, -0.0333333)  (52, -0.0333333)  (53, -0.0166667) 
```

But when I grab a row out of the matrix with periodic BCs, I get
```
row 0: (0, 0.0555556)  (1, -0.0555556)  (2, 0.0555556)  (3, -0.0555556)  (4, 0.0416667)  (5, -0.0555556)  (6, 0.0555556)  (7, -0.0555556)  (8, 0.0555556)  (9, -0.0555556)  (10, 0.0416667)  (11, -0.0555556)  (12, 0.0416667)  (13, -0.0277778)  (14, 0.0416667)  (15, -0.0555556)  (16, 0.0416667)  (17, -0.0555556)  (18, 0.0555556)  (19, -0.0555556)  (20, 0.0555556)  (21, -0.0555556)  (22, 0.0416667)  (23, -0.0555556)  (24, 0.0555556)  (25, -0.0555556)  (26, 0.0555556) 
```

So something looks strange to me

knepley: Periodic has 27 nonzeros, which seems right for Q1 3D periodic hexes.

jeremylt: Sure, but I'm used to the Laplacian  only having only one positive entry in a row and everything else negative

knepley: If I had to guess, I would say you are using the coordinate field which jumps across Riemann sheets. When I calculate the geometry in Plex, I use the "localized" coordinates.

jeremylt: Hmm, that could be... We use whatever comes out of DMGetCoordinatesLocal

knepley: Both sets of coordinates are in there. The coordinates associated with vertices cannot express periodicity, so we also have a SG coordinate field stored on the cells.

jeremylt: So is there a different set of coordinates I should be retrieving than what is in DMGetCoordinatesLocal?

jedbrown: I'm surprised this doesn't generate an error in restriction setup due to mismatched closure sizes. This has been one of the longest standing disagreements between Matt and me. The "localized" coordinates are like a little DG space associated with the cell, but I don't think they're currently put into tensor ordering and I don't think it's easy to make that general. It also leads to singular operators if you have an ALE-type solve (unless you add Lagrange multipliers to enforce continuity elsewhere; much more complicated).

I have an old WIP to use the anchor system for periodicity, but it's now more than a year old. I have it on my plate to revive and finish in the next couple weeks, but Matt twisted my arm into being on a committee. :rofl:  https://gitlab.com/petsc/petsc/-/merge_requests/2178

jeremylt: Ah, if I add 
```
ierr = DMLocalizeCoordinates(dm); CHKERRQ(ierr); /* needed for periodic */
```
then I get an error about sizes when creating the coordinate restriction.

jedbrown:
I rebased my very (!) old branch here. https://gitlab.com/petsc/petsc/-/merge_requests/2178

It makes `DMPlexSetClosurePermutationTensor` work with periodicity. Maybe we should merge clean and merge that up-front, but the coordinates are still a mess because you have normal coordinates and "localized" coordinates in the elements that wrap.

jedbrown:
I want to fix that by making the local space always distinguish and using the sf mapping local to global spaces to tie it back together. Matt says that won't work, but I think I'm going to need to try it and see what breaks.

The solution fields should be managed correctly in the PETSc MR so we could test and merge it, then hack coordinates to make them good enough for us in fluids (@jrwrigh would really like this).

jedbrown:
Periodic boundaries work in https://github.com/CEED/libCEED/tree/jrwrigh/fluid_example, which is otherwise rough. I had to remove the `DMProjectCoordinates` and I hope @knepley can advise on what to do with that failing test case in PETSc because we do need an actual coordinate space at some point once we venture beyond the box with periodicity. https://gitlab.com/petsc/petsc/-/merge_requests/5107

jedbrown:
Click through to the MR with failing test that I created and tagged you in. https://gitlab.com/petsc/petsc/-/merge_requests/5107

The question is how to get an FE for the localized space. I still think it's a big mess and we should excise "localized", but if we can't do that yet, we gotta decide how to make the test work.

jeremylt:
I'm closing this since we sorted this out in fluids

PR #780, opened 2021-06-07
git - add *.a to petsc gitignore

jeremylt:


jeremylt:
Ugh. This error is back
```
libgcov profiling error:/var/lib/gitlab-runner/builds/N8GCsqus/0/libceed/libCEED/build/gallery/ceed-gallery-weak.gcda:overwriting an existing profile data with a different timestamp
...
```

jedbrown (approved):

jedbrown:
It seems like we have a race, but I don't know what it is. Maybe there's a way to make it softer.

jeremylt:
It is infrequent, which makes it harder to pin down.

PR #785, opened 2021-06-17
test - sync data to device arrays

jeremylt:
I think this should force the GPU backends to test the device versions of these functions.


jeremylt:
CI failure is our intermittent GitLab problem again

tests/t122-vector.c
```diff
@@ -23,6 +23,12 @@ int main(int argc, char **argv) {
   CeedVectorSetArray(y, CEED_MEM_HOST, CEED_COPY_VALUES, a);
 
   // Test multiplying two vectors into third
+  {
+    // Sync memtype to device for GPU backends
+    CeedMemType type = CEED_MEM_HOST;
+    CeedGetPreferredMemType(ceed, &type);
+    CeedVectorSyncArray(w, type);
```
nbeams, line 30: I don't think these are actually needed -- this test already ran on the GPU before.  Since `w` (and `y`) isn't set to anything yet, it isn't set to `CEED_*_HOST_SYNC`, the output vec's `memState` is the one used to determine whether to call the host or device function. 
Not that it hurts anything to have it here, though.

tests/t122-vector.c
```diff
@@ -53,6 +59,12 @@ int main(int argc, char **argv) {
   CeedVectorRestoreArrayRead(w, &b);
 
   // Test multiplying vector by itself and putting product into self
+  {
+    // Sync memtype to device for GPU backends
+    CeedMemType type = CEED_MEM_HOST;
+    CeedGetPreferredMemType(ceed, &type);
+    CeedVectorSyncArray(y, type);
```
nbeams, line 66: Same as above

nbeams (approved):
LGTM.  I tested locally and confirmed these tests run on the GPU now for cuda-ref.

I think the sync should be added to `t101-vector` as well, after the Read/Restore but before the first call of `CeedVectorSetValue`.  Otherwise, it only runs on the GPU for the second call to `CeedVectorSetValue`, instead of both.  (Unless this is what you want it to do.)

tests/t122-vector.c
jeremylt: Good point. I'm thinking of leaving it there it make it clear that this occurs in the backend's preferred memory space.

PR #787, opened 2021-06-28
Use more `__restrict__` in `cuda-gen` backend.

YohannDudouit:
This PR generalizes the usage of `__restrict__` to all inputs that can benefit from it.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -64,7 +64,7 @@ typedef struct {
 // Load matrices for basis actions
 //------------------------------------------------------------------------------
 template <int P, int Q>
-inline __device__ void loadMatrix(BackendData& data, const CeedScalar* d_B, CeedScalar* B) {
+inline __device__ void loadMatrix(BackendData &data, const CeedScalar *__restrict__ d_B, CeedScalar *B) {
   for (CeedInt i = data.tid; i < P*Q; i += blockDim.x*blockDim.y*blockDim.z)
     B[i] = d_B[i];
 }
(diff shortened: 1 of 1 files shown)


jedbrown:
Any performance consequences? Do we need to get this into qfunctions as well?

YohannDudouit:
I think it only makes sense where we read from device memory, which is not the case for QFunctions where the data has already been loaded into registers. I haven't done any benchmarking at the moment.

jedbrown (approved):
Looks good. Would you mind making a note about `restrict` semantics in developer notes (doc/sphinx/source/libCEEDdev.rst) and do some minor performance comparison to see if it makes any difference. If it does, we should mention it in release notes.

jeremylt:
This PR just needs the documentation notes, right?

jedbrown:
Yes, I'd meant to circle back and test it myself. It'd be nice to know if there's any impact, but the implication on aliasing should be communicated. It might help to use `restrict` in QFunctions also for CPU. We've been seeing good code without thus far and thus I've omitted it, but I could imagine it affecting code size and/or speed when there are more inputs/outputs.

jeremylt:
I did do some benchmarking on Lassen: https://github.com/CEED/libCEED/tree/jeremy/cuda-gen/restrict

YohannDudouit:
@jed @jeremylt I will try to get to this soon.

YohannDudouit:
> I did do some benchmarking on Lassen: https://github.com/CEED/libCEED/tree/jeremy/cuda-gen/restrict

Are you sure this is the right link?

jeremylt:
Should be. It has your commit 359613d7e4c8b7f55ddd77690d954abe28f7bc88 and my data b9c872237ce4d53b369bd28d265e3afd8d8590b2

jedbrown (approved):
This has no performance impact that I can tell, but I think it's still fair to add.

I've added a note to the docs and will merge if you think it's adequate.

PR #788, opened 2021-07-07
Allow CeedScalar to be single precision 

### LINKED ISSUE #784: Compiling with CeedScalar defined as float
As a first step toward #778, I tried building and testing libCEED with `CeedScalar` set to `float` instead of `double`.  I ran into some minor issues.  I was thinking we may want to address some or all of them in a PR prior to adding any mixed-precision functionality.  *(Of course, we would probably want to keep any potential future changes for mixed precision in mind when deciding what to do about these issues, or whether we want to address them at this time at all.)*

### Issue 1: Compiler warnings for three `CeedVector` functions
  
The affected functions are `CeedVectorSetValue`, `CeedVectorScale`, `CeedVectorAXPY`, which are currently only implemented in the cuda/hip-ref backends.

The warnings come from `CeedSetBackendFunction`, e.g.

```
backends/cuda/ceed-cuda-vector.c:659:33: warning: passing argument 5 of 'CeedSetBackendFunction'
       from incompatible pointer type [-Wincompatible-pointer-types]
                                 CeedVectorSetValue_Cuda); CeedChkBackend(ierr);
                                 ^~~~~~~~~~~~~~~~~~~~~~~
./include/ceed/backend.h:93:69: note: expected 'int (*)()' but argument is of type
       'int (*)(struct CeedVector_private *, CeedScalar)' {aka 'int (*)(struct CeedVector_private *, float)'}
                                 const char *func_name, int (*f)());
                                                        ~~~~~~^~~~
```

The reason only these three functions are affected is that they are the only ones that take a parameter of type `CeedScalar` (rather than `CeedScalar *`).  From the C standard, section 6.5.2.2, a function declaration with unspecified parameters will result in "default argument promotion," which includes converting `float` to `double`.  And from section 6.7.5.3, that means the function declaration of `int (*f)()` is incompatible with a prototype containing a `CeedScalar` parameter when `CeedScalar` is `float`, but it *is* compatible when `CeedScalar = double`.

Now, from what I can tell, this incompatibility doesn't actually cause incorrect execution of the three functions, since the functions are still called through function pointers with the correct parameter types for the backend implementation, from the function pointers in the `CeedVector` object.  Perhaps this is obvious to some libCEED developers/expert-level C coders, but I struggled a bit to make sure I understood the distinction between what libCEED does and things that *would* produce errors in this case. To convince myself, I confirmed this behavior from a minimal test code.  Consider three functions with prototypes `int func_dbl(double x)`, `int func_flt(float x)`, and `int func_pflt(float *x)` and the following `main()`:

```
  int ierr;
  float a = 0.5;
  double b = 0.5;

  int (*bar)() = &func_dbl;
  // OK: all floating-point arguments considered as double if unspecified
  ierr = bar(b);

  // Create function pointer with correct parameter type (like backend fcn pointers in libCEED objects);
  //  will set/use later
  int (*fbar)(float);
  // Pointer with unspecified parameters: compiler warning due to float promotion causing incompatible types
  int (*fbar_u)() = &func_flt; 

  // Not OK: undefined behavior to call function through this pointer
  ierr = fbar_u(a);

  // Mimimal recreation of behavior of CeedSetBackendFunction and libCEED objects:
  // Create pointer to fcn pointer with no parameters; set equal to address of fbar, with cast.
  int (**pfbar_v)(void) = (int (**)(void))(&fbar);
  // Set value of pfbar_v to the pointer with unspecified paramters (which points to func_flt);
  // this also sets fbar due to previous line
  *pfbar_v = fbar_u;
  // Call function through pointer with correct parameter specification; OK
  ierr = fbar(a);

  // Version where the parameter is a pointer to a float instead of a float.
  // This one does not give a warning because there is no incompatible type promotion
  // when using a pointer.
  int (*ifbar)() = &func_pflt;
  // OK
  ierr = ifbar(&a);
```

I tried this with GCC 8.3.0 and clang 11.0.1. Both gave the "incompatible pointer type" warning seen in compiling the cuda-ref backend for the line creating `fbar_u`, and both produced the same runtime behavior. So there's no problem with the functions' behavior in single precision, despite the warnings.

Still, we most likely wouldn't want these warnings to be produced for every backend that implements these three functions (though it's currently only cuda-ref and hip-ref).

A simple solution, though maybe not "nice" since it would ruin the symmetry of setting up all the backend functions of a `CeedVector` object, would be to perform a cast on these three functions during `CeedSetBackendFunction`, e.g.

```
 ierr = CeedSetBackendFunction(ceed, "Vector", vec, "SetValue",
                                (int (*)())(CeedVectorSetValue_Cuda)); CeedChkBackend(ierr);
```

This removes the compiler warning and is valid per C function pointer conversion rules. I think this is preferable to having these functions take `CeedScalar *` instead, given how the variables are used in the functions.  However, I would like to get input on this from other libCEED developers.

### Issue 2: Places where we need to know the type of `CeedScalar` 

There are several places we would need to know exactly what type `CeedScalar` is, e.g.:

 - In the runtime compilation for CUDA and HIP backends, namely the `Ceed[Cuda/Hip]Compile` function,
   where the definition of `CeedScalar` is added directly to the code string to be compiled,
   to avoid having to include any headers

 - In functions that call library routines/instructions for a specific precision: MAGMA non-tensor
   routines (GEMMs), AVX basis routines, the vector norm functions for CUDA/HIP backends...

 - Perhaps in the tests (see issue 3).

The question here is whether we want the user to have to define anything else while compiling, or *only* change `CeedScalar`. One option I tried to get the CUDA/HIP RTC working was to create variables of type `CeedScalar`, `float`, and `double`, then compare their sizes to determine which definition to add to the code string. Perhaps something like that could be done in a general helper function somewhere, so that any routine can check?  Or is there a more elegant way to do this in C?

Of course moving toward mixed precision support in the future, we may just want to make it so that the compilation functions take a parameter/parameter(s) indicating the main/computational precision of the code to be compiled (since the same compilation functions are used for all kernels, even if not all kernels will have the option for differing computation and argument precisions, the compilation functions must be able to support it).

### Issue 3: Required changes to libCEED tests

Many libCEED tests fail when `CeedScalar` is float.  In the case of t104, it just needs a cast to `CeedScalar` in the comparison, e.g. `if (a[3] != (CeedScalar)(-3.14))`, otherwise the comparison is done in double and fails.  Many other tests seem to fail due to the tolerances being set too low for what we are guaranteed in single precision, though of course not all of them do.

However, these tolerances are not all the same and seem tailored to the individual tests, somewhat complicating the situation if trying to set different tolerances based on `CeedScalar` (or future mixed-precision routines).  This might also require a way to check what `CeedScalar` is or another definition from the header (mentioned in issue 2).

nbeams:
This PR would expand the use of `CeedScalar` to be double or single precision.  See also discussion in #784.

Summary of main changes thus far:

 - Introduction of two different headers, `ceed-f32.h` and `ceed-f64.h`.  To change between the types, you change which file is included in `ceed.h`.  These files contain different values for `CEED_EPSILON` and `CEED_ALIGN`.
 - Addition of `CeedScalarType` enum and `CEED_SCALAR_TYPE` definition.  In libCEED C/C++ code, `CEED_SCALAR_TYPE` can be used to check for type-dependent code (e.g. whether to use the AVX basis functions, several places in the MAGMA backend, CUDA/HIP vector norms, etc.). 
 - Updating Python bindings: Most notably, places that used to have `dtype="float64"` now have `dtype=scalar_types[lib.CEED_SCALAR_TYPE]`, where `scalar_types` is a Python dictionary with type strings.
 - Moved `CeedQFunctionContextGetContextSize` to be a public/user function -- this was actually done because of the Python bindings.  It was used in `get_data` (not sure why it wasn't causing errors?  It did when I was running tests directly in ipython) but not available through `libceed.lib`, and not having the correct size (in bytes) was causing failures in one of the Python QFunction tests when using single precision.  (This was fixed by 9c4a53cccd47c3c111b9e680b929acab1fc1124b)

Outstanding issues/todos:

- [x] Test Python bindings with CUDA -- made some adjustments based on changes required for fix QFunctionContext-related failures mentioned above, but haven't tested yet
- [x] Possibly add a macro to avoid compiling MAGMA DGEMM non-tensor basis code when CeedScalar is float (currently gives compiler warnings, but code should never be executed), or add full SGEMM support.
- [x] Fix/decide what to do about the following tests: t312, t314, t315, t316, t321, t322, t323 in the basis tests, and t501, t502, t503, t505, t506, t507, t511, t521, t524, t530, t550, t551, t552, t553 in the operator tests.  These tests have a tolerance such that adjusting them directly for float as was done for other tests in 2c97b8407388333d0a3c7f6a9259adfa268cde52 would result in a very large value for the tolerance (probably more than we want). 
- [x] Fix/decide what to do about the tests that compare directly to output files: t300, t301, t302, t304, t305, t320; and t402.  (Also affects the Python versions of these tests)
- [x] Update other examples as needed to work with float (the examples in `/ceed` have been updated already)
- [x] Fix Fortran tests for float
- [x] Update Rust bindings
- [x] Update Julia bindings

Closes #784.


include/ceed/ceed.h
```diff
@@ -596,6 +606,8 @@ CEED_EXTERN int CeedQFunctionContextGetData(CeedQFunctionContext ctx,
     CeedMemType mem_type, void *data);
 CEED_EXTERN int CeedQFunctionContextRestoreData(CeedQFunctionContext ctx,
     void *data);
+CEED_EXTERN int CeedQFunctionContextGetContextSize(CeedQFunctionContext ctx,
+    size_t *ctx_size);
```
pazner, line 610: This declaration is also found in `backend.h`, which confuses the Julia bindings

pazner:
I started working on the modifying the Julia interface to allow single precision. The work is in [this branch](https://github.com/CEED/libCEED/tree/will/julia-ceedscalar-generalize).

For the most part I think it should be relatively straightforward.

It would be useful if there was a runtime function that returned e.g. 32 or 64 depending on the value of `CEED_SCALAR_TYPE`. That would allow the Julia interface to detect whether or not it was misconfigured. Since the user can supply their own version of the library, we cannot know a priori whether CeedScalar is 32 or 64 bits.

include/ceed/ceed.h
nbeams: Whoops, sorry about that -- I definitely forgot to delete from `backend.h` after adding to `ceed.h`.  
Are you planning to eventually bring over the commits from your branch to this one (meaning I don't need to go ahead and make this change here as well)?

nbeams:
Thanks @pazner!

I'm not sure I understand exactly how the Julia interface works.  You mean the user could supply a compiled version of libCEED from a directory other than the one they're using the Julia package from?  

There was some discussion on #784 about leaving things open to potentially more "exotic" types in the future, which may not be entirely defined by size.  Would this be possible with the Julia interface?  I think the idea was that it would be better to know the actual type rather than just its size.

include/ceed/ceed.h
pazner: I would suggest making the change here. Eventually I'll merge the changes from the Julia branch, but I am experimenting with some different options, so it might not happen immediately.

pazner:
> I'm not sure I understand exactly how the Julia interface works. You mean the user could supply a compiled version of libCEED from a directory other than the one they're using the Julia package from?

Yeah, the idea would be that the Julia package comes pre-installed with a "basic version" of libCEED that is precompiled. This basic version does not include CUDA or HIP support, is compiled with the "generic" compiler flags, would use double precision for CeedScalar, etc.

If the user wants to use more advanced features (GPU support, architecture-specific compiler flags, single precision, etc.), they would have to compile their own version of libCEED from source, and tell LibCEED.jl to use that library instead of its preinstalled basic version.

> There was some discussion on #784 about leaving things open to potentially more "exotic" types in the future, which may not be entirely defined by size. Would this be possible with the Julia interface? I think the idea was that it would be better to know the actual type rather than just its size.

That's a good point. Depending on the type, that should be fine. Any interface would be fine as far as I'm concerned, it could even just return a `CeedScalarType` enum value.

I just think it would be good to expose this information as a runtime function. If the user provides their own version of libCEED that uses e.g. float instead of double, then the Julia interface needs to be aware of this.

include/ceed/ceed.h
nbeams: Sounds good, I fixed it in 38c8540f60fa93923871da0dbdba0fe187fb12bf

nbeams:
What does everyone think about a library-level function like `CeedGetVersion`?  Maybe `CeedGetScalarType`?  (Since in this case, it would be for the whole library, and not a specific `Ceed` object.)   Is there no other way for Julia to find the information (e.g. by parsing ceed.h to see which header is included)?

pazner:
Julia could parse the header, but the problem is that the header might not be available — the user only needs to provide a binary file.

nbeams:
> the problem is that the header might not be available

Ah, yes, that's right.  Though, theoretically, could the user be required to supply both the compiled library file and the location of the headers for the binary?  Assuming this is the case where said user has compiled the library locally in order to have non-default options, the headers should be available.  Would that be totally antithetical to how Julia works?

If add the function, what about something like
```
int CeedGetScalarType(CeedScalarType *scalar_type) {  
  *scalar_type = CEED_SCALAR_TYPE;
  return CEED_ERROR_SUCCESS;
  }
```
(trying to match `CeedGetVersion` and `CeedGetPreferredMemType`).  Would this work for Julia?  You said returning an enum type would be okay, right?

pazner:
> Ah, yes, that's right. Though, theoretically, could the user be required to supply both the compiled library file and the location of the headers for the binary? Assuming this is the case where said user has compiled the library locally in order to have non-default options, the headers should be available. Would that be totally antithetical to how Julia works?

Yeah, that's a possibility, but I really think having the type available at runtime would be simpler. No headers would be required, and there is no possibility of a mismatch (e.g. the header and binary get out of sync somehow).

> If add the function, what about something like
> 
> ```
> int CeedGetScalarType(CeedScalarType *scalar_type) {  
>   *scalar_type = CEED_SCALAR_TYPE;
>   return CEED_ERROR_SUCCESS;
>   }
> ```
> 
> (trying to match `CeedGetVersion` and `CeedGetPreferredMemType`). Would this work for Julia? You said returning an enum type would be okay, right?

Yep, that would work perfectly. The function could possibly also just return `CEED_SCALAR_TYPE` rather than assign it to the output parameter, since there is no possibility of failure. On the other hand, always returning an error code would be more consistent with the rest of the libCEED API. Anything along those lines would work great for me.

nbeams:
The changes in ccbb23166e099ba1862a40d7766c12ca961c9405 should take care of all the remaining C tests, other than the ones that compare to things in the output directory.  These were the tests that used 1.e-10 for the tolerance.  

After looking at the output for these tests in double and single, I was able to come up with bounds that are tighter than what was previously used in double, while (I hope) still being acceptable in single: 
 - A few basis tests (312, 321, 322, 323) use `100*CEED_EPSILON`
 - Operator tests use `1000*CEED_EPSILON`.  (These should be okay at `500*CEED_EPSILON` as well, based on my tests, if we want a stricter check.) 
 - Only three tests seemed to require something "fancier": 314, 315, and 316.  For these, I just added custom functions to get a tolerance based on dimension and `CEED_SCALAR_TYPE`, so that we can use the necessary large tolerance for the `dim=3`/`float` combination, while using something tighter in dimensions 1 and 2, and a different tolerance for double precision.

pazner:
@nbeams, I put your `CeedGetScalarType` in my branch and added the Julia integration. The Julia package will check to make sure that its scalar type and the library's scalar type match — if not, it will reconfigure itself and notify the user.

```
julia> using LibCEED
[ Info: Precompiling LibCEED [2cd74e05-b976-4426-91fa-5f1011f8952b]
┌ Warning: libCEED is compiled with Float32 but LibCEED.jl is using Float64.
│ 
│ Configuring LibCEED.jl to use Float32.
│ 
│ The Julia session must be restarted for changes to take effect.
```

All Julia tests (except those comparing against text output files) are passing with single precision.

nbeams:
@pazner Awesome!  Do you want to go ahead and merge your branch into this one?  Or are you still considering some more options?  

It seems the "compare to output" tests will have to be handled slightly differently for the different languages (Python does its own check, for example, vs the C tests just running a diff in `tap.sh`, I think?) -- do you have any ideas for what would work for Julia?

pazner:
> @pazner Awesome! Do you want to go ahead and merge your branch into this one? Or are you still considering some more options?

Just merged it. The Julia CI will fail, I think we can ignore it for now (I have some workarounds in mind).

> It seems the "compare to output" tests will have to be handled slightly differently for the different languages (Python does its own check, for example, vs the C tests just running a diff in `tap.sh`, I think?) -- do you have any ideas for what would work for Julia?

Julia also does its own check, but it's just a simple textual comparison with a baseline output text file that's stored in the repo. I imagine that whatever solution we come up with for C and Python, we should be able to implement the same idea for Julia.

nbeams:
> I imagine that whatever solution we come up with for C and Python, we should be able to implement the same idea for Julia.

I was thinking Python will actually be easier than C.  In Python, the check happens inside the test itself, where we have imported the libceed module and have access to the scalar type, and so we can use that to decide which output file to check against, for example.  But since the C tests do the `diff` in `tap.sh`, rather than in the tests themselves, we don't have access to the scalar type.  Maybe we could have the tests output something about `CEED_SCALAR_TYPE` on a line by itself first, then in the test scripts, parse that line to pick which output file to check against?   And we could add that line to the output files, so they will still match after selecting the correct one?
(And I guess there are two options for running the C tests, and whatever we do has to work for both `tap.sh` and `junit.py`.)

pazner:
I see. Yeah, perhaps the best option for C is to somehow choose different files to compare from within `tap` (I don't have much experience with `tap` so I'm not clear on how the details would work).

The option that you proposed for Python would also work for Julia — I'll try to prototype it.

jeremylt:
TAP is just the output format. The logic is all in a bash script.

Ideally, I would like to have all tests use a tolerance except for on test per libCEED object that tests the display function for that object.

nbeams:
@jeremylt So you would like to change some of the t3* tests that currently require output checking?  (And maybe others, but only the basis tests have multiple failures in single precision.) 

Another question I forgot to put on the todo list by itself, but which ties to these t3* tests, is whether we want to change how many digits are output by `CeedBasisView` in the case of single precision -- right now it outputs 8 -- should we change to, say, 6?  Then we could be sure the output would exactly match that of the comparison file for float.

jeremylt:
Yeah, I've been wanting to refactor those tests to use less fixed output files but haven't made the time yet. I think ultimately there only something along the lines of 8 C tests that need to actually check against output for the view functionality?

pazner:
All the Julia tests are passing now, including the ones that compare "view" output — baseline files are included for both single and double precision, and the tests will compare against the correct file depending on the type of CeedScalar.

nbeams:
By my simple `grep` investigation, I count 11 tests where the `*View` call seems to be a main "point" of the test:

- `CeedView` - t003
- `CeedVectorView` - t107
- `CeedElemRestrictionView` - t210, 211, 212 (tests standard, strided, blocked strided)
- `CeedBasisView` t300, 320 (tensor, non-tensor)
- `CeedQFunctionView` - t402, t413 (header, gallery)
- `CeedQFunctionContextView` - t402 again
- `CeedOperatorView` - t504, 523 (regular, composite)

Of these tests, only `t300`, `t320`, and `t402` fail in single precision.

jeremylt:
Ah, better count than mine. I believe you are correct. We pass a format into our vector view functions. I feel like it may be appropriate for the basis as well.

nbeams:
> We pass a format into our vector view functions. I feel like it may be appropriate for the basis as well.

Ah, good idea.  What if we:
 - Add format to the `CeedBasisView` (it will get passed to the `CeedScalarView` that `CeedBasisView` calls) function
 - In the basis tests that compare output, change the format to only print, say, 6 digits after the decimal.  Update the output files.  Both double and single can use the same files.  (I'm assuming 6 digits will be enough that single will be the same in all output, but I haven't tried it yet)
 - In the remaining test that fails, t402, change the test itself (in C and Python versions) so that if `CEED_SCALAR_TYPE` is float, it will make the context twice as long (just repeat the array).  The context doesn't actually get used for anything in this test/QFunction, so it doesn't matter, and then the size will be the same (5 * 8 = 40 or 10 * 4 = 40) for both outputs. 
 
 Then, we actually shouldn't need different sets of output files to compare to...

nbeams:
I guess there is a question of what to do with the Python bindings -- going off what `CeedVector` does (using `CeedVectorView` in the `__str__` method), we can't easily pass a format there or to `CeedBasisView`.  We could either make the default have fewer digits than it uses now, or have separate output files to check for the Python basis tests only (and the Julia tests as well, if Will thinks that would still work better for Julia even after the `CeedBasisView` change).  As mentioned previously, checking for separate output comparisons in the Python tests is easier than for C, anyway.

nbeams:
Hmm, I see the Fortran interface also doesn't take a format (`ceedvectorview` just uses the default of "%12.8f" for its call to `CeedVectorView`)...so with the Fortran tests, we'd be back to the same question of how to pick the output to compare against (unlike Python), unless we're okay with the Fortran interface always using fewer digits by default (are we?). I guess it'd be confusing to have the Fortran take a "C-style format" string as an argument -- Fortran users would expect to be able to pass a Fortran-style specification -- I'm assuming that's why the format was left out of the Fortran interface for the vector view?

jeremylt:
That's exactly why we did that.

nbeams:
So, thoughts on what to do about Fortran?  
a) change default format of `CeedBasisView` to have fewer digits in Fortran, use same output files
b) figure out a way to change which output file to compare to for the C/Fortran tests

And if b), do we want to scrap the planned changes above and just have separate output files for C as well?
(Note that I also haven't yet figured out a good way to know which data type to use in the Fortran tests...)

(The `t402` change could go ahead either way)

nbeams:
As an update, in case this changes anyone's opinions: it seems we would need to go to only 5 digits after the decimal for the basis tests in order for the double and single precision output to match.

interface/ceed.c
```diff
@@ -1182,4 +1182,10 @@ int CeedGetVersion(int *major, int *minor, int *patch, bool *release) {
   return 0;
 }
 
+int CeedGetScalarType(CeedScalarType *scalar_type) {
+  *scalar_type = CEED_SCALAR_TYPE;
+  return 0;
+}
```
jedbrown, line 1188: Do we want to commit to supporting this global interface, or might we do object-specific `CeedVectorGetScalarType(CeedVector, CeedScalarType *)`? I'm also fine if this is a stop-gap so we can get this PR to a mergeable state, and then we'll revisit. In any case, I'd hope we can avoid making heavy dependencies on scalar type being a compile-time constant.

nbeams: This was only added for the Julia interface, and that's the only place it's used at the moment.  My understanding of Will's comments was that it was necessary in order for Julia to know which definitions to use for things related to the scalar type (most importantly, `CeedScalar`), if the user provides a compiled binary of the library to use with the Julia interface.  An alternative was that the user would have to provide a location for the headers as well, but Will was not in favor of that.  We could maybe change its documentation (and even name?) to make it clear that it's specifically for the Julia interface, and not really meant to be used elsewhere?

jedbrown: @pazner What if precision was per-object instead of per-library? (Perhaps every object would return the same thing now, but someday they could be different.) Would that serve your needs?

pazner: It's a bit hard for me to say, because I don't completely understand what you are proposing. It could be more challenging if the type is not known globally at compile time, but rather per object at runtime, because in order for Julia to compile efficient code, it has to be "type stable", meaning that the types are inferable at compile time. It's possible that this would work with what you are describing, but I would need more information before being able to say for sure.

But I am also a bit confused: there are plenty of places in the C API that use `CeedScalar`. This type has to be known at compile time… or am I missing something? Would the proposal be to never pass `CeedScalar` by value and only use pointers instead (i.e. `CeedScalar` would be kind of like an opaque type)?

jedbrown: This PR was meant to be a stepping stone toward a library in which scalar type could be chosen per-object similarly to how [BLIS supports fine-grained run-time selection of precision](https://github.com/flame/blis/blob/master/docs/MixedDatatypes.md). So for example, you could take a `CeedVector` in single precision and promote to double in restriction, or you could go the other way. User-provided QFunctions would have specified precision (or perhaps you'd set multiple variants if you have them available), but everything else would be run-time selectable. I think we have a strategy where users who don't want mixed precision will be able to basically ignore it by including a suitable header or using C11 `_Generic`, while those who want dynamically-selected mixed precision will also be able to do so.

pazner: I see, that makes sense to me.

Since this PR is just a stepping stone, I think we can consider something like `CeedGetScalarType` to be a provisional solution to get the Julia interface working with this PR. It doesn't need to be included in the public API (it doesn't even have to be declared in the header file).

When it is clearer what the final C API for mixed precision will look like, we can try to come up with a design for the Julia interface that will work.

nbeams:
Regarding the external examples:

 - My understanding is that we don't expect the Nek5000 examples to work with float, and can't make them work
 
 - MFEM also doesn't do single precision at the moment, so we don't expect these tests to work in single, anyway
   - What's the current feeling about the MFEM examples?  Now that MFEM has libCEED support directly, such that all you need to do is build MFEM with libCEED and select a `ceed` device configuration, do we still need the MFEM examples? 
 
 - Petsc examples would require that Petsc is also built with `PetscScalar = float`

jeremylt:
I see a big part of the purpose of the examples as a way to show others how to integrate libCEED into their own code, so I think we still need all of the examples as they are.

nbeams:
Then for the purposes of this PR, I think I can consider the examples as "done"?

jedbrown:
The existing MFEM examples are useful demonstrations as Jeremy says, but also allow testing within the repository. That gets complicated if testing libCEED requires building `libceed.so` and then building an MFEM to use libCEED.

We can create a PETSC_ARCH on Noether built for single precision and use that in CI. I would have no objections if only a subset of current tests are done with single precision. It can be a different job in `.gitlab-ci.yml`.

tests/t402-qfunction.c
```diff
@@ -24,9 +24,16 @@ int main(int argc, char **argv) {
...
   CeedQFunctionContextCreate(ceed, &ctx);
-  CeedScalar ctxData[5] = {1, 2, 3, 4, 5};
-  CeedQFunctionContextSetData(ctx, CEED_MEM_HOST, CEED_COPY_VALUES,
-                              sizeof(ctxData), &ctxData);
+  if (CEED_SCALAR_TYPE == CEED_SCALAR_FP64) {
+    CeedScalar ctxData[5] = {1, 2, 3, 4, 5};
+    CeedQFunctionContextSetData(ctx, CEED_MEM_HOST, CEED_COPY_VALUES,
+                                sizeof(ctxData), &ctxData);
+  }
+  else { // Make context twice as long so the size is the same in output
```
jeremylt, line 32: `} else {`

python/tests/test-3-basis.py
```diff
@@ -158,6 +165,7 @@ def test_305(ceed_resource, capsys):
         print("%12.8f" % lam[i])
 
     stdout, stderr, ref_stdout = check.output(capsys)
+# TODO: fix this for float or double
```
jeremylt, line 168: Refactored this test on main to avoid this output checking

python/tests/test-3-basis.py
```diff
@@ -122,6 +126,7 @@ def test_304(ceed_resource, capsys):
...
```
jeremylt, line 129: refactored

python/tests/test-4-qfunction.py
```diff
@@ -282,7 +302,8 @@ def test_412(ceed_resource):
 
     q = 8
 
-    u_array = np.zeros(q * size, dtype="float64")
+    u_array = np.zeros(q * size,
+                       dtype=libceed.scalar_types[libceed.lib.CEED_SCALAR_TYPE])
```
jeremylt, line 306: This feels a little clunky. How about adding a `ceed.scalar_type()` helper function to do this? Eventually I would see us doing this per object.

python/tests/test-5-operator.py
```diff
@@ -23,7 +23,7 @@
 import check
 import buildmats as bm
 
-TOL = np.finfo(float).eps * 256
+TOL = libceed.lib.CEED_EPSILON * 256
```
jeremylt, line 26: I think we should add this to `constants.py` so users have `libceed.EPSILON`

tests/t506-operator.c
```diff
@@ -138,10 +138,10 @@ int main(int argc, char **argv) {
...
     sum_2 += hv[2*i+1];
   }
-  if (fabs(sum_1-1.)>1e-10) printf("Computed Area: %f != True Area: 1.0\n",
-                                     sum_1);
-  if (fabs(sum_2-2.)>1e-10) printf("Computed Area: %f != True Area: 2.0\n",
-                                     sum_2);
+  if (fabs(sum_1-1.)>1000.*CEED_EPSILON)
+    printf("Computed Area: %f != True Area: 1.0\n", sum_1);
+  if (fabs(sum_2-2.)>1000.*CEED_EPSILON)
+    printf("Computed Area: %f != True Area: 2.0\n", sum_2);
```
jeremylt, line 144: If error print statements spill onto their own line(s), they should have LCOV exclusion comments

include/ceed/ceed-f32.h
```diff
@@ -0,0 +1,35 @@
...
+/// Set base scalar type to FP32.  (See CeedScalarType enum in ceed/ceed.h
+/// for all options.)
+#define CEED_SCALAR_TYPE CEED_SCALAR_FP32
+typedef float CeedScalar;
+
+/// Data alignment of 32 bits
+#define CEED_ALIGN 32
+
+/// Machine epsilon
+#define CEED_EPSILON 6e-08
```
jeremylt, line 33: Why 6e-08? What am I missing? Should the f64 value be different or should this one?

nbeams: I was going with the LAPACK definition of machine epsilon for single, based on using 1E-16 for double https://www.netlib.org/lapack/lug/node74.html (instead of 2E-16)  -- just trying to be consistent.  I don't have a strong opinion on which definition to use.

python/tests/test-4-qfunction.py
nbeams: I thought we were wanting to avoid this, as with the `CeedGetScalarType` discussion above, if it would later be changed to per-object anyway?  (The difference being that the Julia interface required something like this to work in the interim.)

jeremylt: We could go directly to a per device version if you prefer. I'm just objecting to repeating `libceed.scalar_types[libceed.lib.CEED_SCALAR_TYPE]` over and over and over all over the Python test suite

nbeams:
Looks like I need to do a rebase or merge and update after the latest changes in `main`. 

I also realized that the way I adjusted the Python interface in `build_ceed_cffi.py` to check for which header was included will fail if someone comments out another `#include ceed/ceed-f[num].h` line instead of replacing it, even though that would be fine for C.  How big of a problem do you think this is?

include/ceed/ceed-f32.h
jedbrown: This definition produces the same value as what I use in class: the smallest `eps` such that `1 + eps > 1`.

YohannDudouit:
I need to rewrite the MFEM examples using the libCEED integration, their current form has been inherited from times where libCEED was not integrated into MFEM.

jedbrown:
@YohannDudouit I think we need think carefully about the circular dependency that would induce. The purpose of these may not be to show best practices for using libCEED from MFEM, but to continually test changes in libCEED. If that test coverage is captured in other tests, then we could even delete them and just clearly document that MFEM users should look at examples X,Y,Z in MFEM. A stable documentation page on mfem.org would be great.

tests/t402-qfunction.c
nbeams: This was fixed in b628dc116a24e09db24d9f0fa78c6f57e5018f9d

include/ceed/fortran.h
```diff
@@ -30,6 +30,16 @@
...
 
+!-----------------------------------------------------------------------
+! CeedScalarType
+!-----------------------------------------------------------------------
+
+      integer ceed_scalar_fp32
+      parameter(ceed_scalar_fp32   = 0)
+
+      integer ceed_scalar_fp64
+      parameter(ceed_scalar_fp64 = 1)
```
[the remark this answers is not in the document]
nbeams: Should be fixed in 6f64236971afd3229e775a13b70ff24c359b08d0

tests/t506-operator.c
nbeams: LCOV exclusions added here and in several other operator tests in 1de5c84dffc2b50077960f7a325fd878a56568fb

python/tests/test-5-operator.py
nbeams: This was added in 6cd68c5894ee60cd26737e0b42d38c37386d6369

python/tests/test-4-qfunction.py
nbeams: Please check out ebfc85f15683e73b56e841e98b4bfbe5c12304c5 and see if this is what you had in mind.

python/tests/test-3-basis.py
nbeams: These changes have been brought over from main, so I'm marking this as resolved

nbeams:
After @jeremylt's refactoring of some of the tests that check output, 1a87bfa92f7f41a9dbebe9158774bd97032c287a should take care of the only remaining two (t300 and t320) that still failed in single precision, provided this solution is acceptable.  tap.sh/junit should check for either output file and pass if one of the two matches.

The other change in 1a87bfa92f7f41a9dbebe9158774bd97032c287a fixes the last C test that failed in single: 541.  After a lot of combing through output differences and some discussion with @jeremylt, we decided it seemed this test should be skipped in single precision, because it's too poorly conditioned. 

In f753e60a736bc00ea5b13b30ee2397ca0ea5bfc7 the Python test setup gets the ability to add an optional file suffix parameter (other than '.out') to select a different output file for comparison, which is used for `test_300`, the remaining "check against output" test for Python.

jeremylt:
I think Fortran and Rust are the biggest pieces open here? Rust should be straightforward, but I'm not sure about Fortran. I don't think we need to mess with adding single precision to the NEK5000 example.

jedbrown:
I have no objections to just not supporting Fortran single precision unless/until we have user interest. I suspect that because Nek5000 -> NekRS has made an end-run around Fortran90, any future Fortran users are likely to favor Fortran2008. I think it's wonderful that we have the nice test suite, but I think we should minimize effort on Fortran until users articulate their preferred environment.

nbeams:
Yes, I believe the two biggest open questions are Rust and Fortran.  I have no experience with Rust, so I have no thoughts there (I don't even know how best to set up a testing environment for the Rust tests). 

I have spent a bit of time trying to come up with something elegant for the Fortran tests, but haven't had any success thus far.  (I can think of clunky options that would work, but would involve lots of code duplication.)

I'm not sure if the Julia tests need to be updated after the refactoring of some of the "output" tests.  I see the Julia CI is currently failing -- I don't recall if it was before I merged in the changes from main -- did I inadvertently screw something up while doing that?

The other (smaller?) question is the issue I raised previously:

> I also realized that the way I adjusted the Python interface in build_ceed_cffi.py to check for which header was included will fail if someone comments out another #include ceed/ceed-f[num].h line instead of replacing it, even though that would be fine for C. How big of a problem do you think this is?

Just wanted to make sure everyone saw this question since a lot has been going on in this PR.

nbeams:
> I have no objections to just not supporting Fortran single precision unless/until we have user interest.

Do you have a preference for how the testing framework handles this?  For example, what @jeremylt and I currently discussed for t541 is to have the test output an error statement if `CEED_SCALAR_TYPE` is single precision, and then `tap.sh` and `junit.py` can catch that particular error statement to pass the test.  But would we want to do that with *every* Fortran test?  Probably not?

jeremylt:
We could just make a ctests target and error when trying to build against the Fortran interface in single precision.

Re:
>I also realized that the way I adjusted the Python interface in build_ceed_cffi.py to check for which header was included will fail if someone comments out another #include ceed/ceed-f[num].h line instead of replacing it, even though that would be fine for C. How big of a problem do you think this is?

I'm ok with this. The monolithic single vs double setting is a stepping stone to more fine-grained decisions, so this decision won't stay with us for too long.

Re: Rust
I should be able to make that change fairly quickly. If you're interested, I can give you a tour of the Rust interface - long term I'd like to figure out a clean way to use Rust functions in our GPU backends as is done with our Julia interface, but I haven't had the time to spend on this yet and Rust's GPU support isn't great at the moment.

pazner:
> I'm not sure if the Julia tests need to be updated after the refactoring of some of the "output" tests. I see the Julia CI is currently failing -- I don't recall if it was before I merged in the changes from main -- did I inadvertently screw something up while doing that?

The Julia CI is currently failing because when the package is loaded, it checks that it is configured correctly (i.e. if the libCEED library is set to f64 mode, then the Julia interface will use Float64, if the library is set to f32 mode, then the Julia interface will use Float32). This requires the function `CeedGetScalarType`, and so loading the package fails with the last released version of libCEED (it will only work with a library built from this branch).

One option is to change how the CI is run somehow to avoid this failure (or possibly just ignore it for now…)

Another option is for the Julia package to check if `CeedGetScalarType` is defined in the shared library, and if not then assume double precision. This would be a temporary fix that we could remove after the next release of libCEED (we'd have to make a note of this so we remember).

rust/libceed/src/lib.rs
```diff
@@ -54,12 +54,18 @@ pub mod operator;
...
+// -----------------------------------------------------------------------------
+pub type Scalar = bind_ceed::CeedScalar;
+
 // -----------------------------------------------------------------------------
 // Constants for library
 // -----------------------------------------------------------------------------
 const MAX_BUFFER_LENGTH: u64 = 4096;
 pub const MAX_QFUNCTION_FIELDS: usize = 16;
 pub const CEED_STRIDES_BACKEND: [i32; 3] = [0; 3];
+pub const EPSILON: crate::Scalar = bind_ceed::CEED_EPSILON as crate::Scalar;
```
jeremylt, line 68: This was pretty painless - I just propagated this typedef from libCEED all of the way through to our Rust examples

rust/libceed/src/lib.rs
```diff
@@ -54,12 +54,18 @@ pub mod operator;
 pub mod qfunction;
 pub mod vector;
 
+// -----------------------------------------------------------------------------
+// Typedef for scalar
+// -----------------------------------------------------------------------------
+pub type Scalar = bind_ceed::CeedScalar;
```
jeremylt, line 60: Woops, meant to highlight this line

nbeams:
> We could just make a ctests target and error when trying to build against the Fortran interface in single precision.
I'm not sure I understand what you mean by this.  In the Makefile?  

Thanks for taking care of the Rust changes!

> If you're interested, I can give you a tour of the Rust interface 

That would be useful, I think.

jeremylt:
Right, so currently in the Makefile we have just the C tests
https://github.com/CEED/libCEED/blob/b5daa9482a167c8fe65ae14d3b389f8754935dd0/Makefile#L195
but we don't have an easy way to only target these tests with `make prove` and `make junit`.

So as far as testing goes, I think we should

1) Add single precision Rust, Julia, and Python GitHub Actions

2) Add an easy way to target only the C tests in the Makefile

3) Add another GitLab CI job (Noether only or Noether and LV?)

Then we would also need to

4) Modify the Fortran to error if the library was built in single precision

jeremylt:
I think the LIBXSMM backend will need the same changes that were added to MAGMA to offer either smm or dmm kernels

nbeams:
> Modify the Fortran to error if the library was built in single precision

You mean the tests themselves?  The Fortran interface should work fine in single precision provided the user has `real*4` type data instead of `real*8` (I tried this with a few of the Fortran tests and confirmed they worked), so we don't want to give the impression that Fortran users can't use single precision. 

> I think the LIBXSMM backend will need the same changes that were added to MAGMA to offer either smm or dmm kernels

LIBXSMM is a good point that I'd forgotten about because I never test with it on any of my libCEED setups.  I haven't actually added any SGEMM functionality to MAGMA yet -- this PR is already so massive that I was leaning toward doing that in a follow-up PR.  What I did currently for both MAGMA non-tensor basis create and the AVX backend's `TensorContractCreate` is just put a check on `CEED_SCALAR_TYPE` before calling the "connection" for the function in the backend (`CeedSetBackendFunction`); so in single precision, these backends should still work, but using their delegate backends instead.  However, the way the MAGMA backend is currently set up, that does mean a lot of compiler warnings when `CeedScalar` is single, even though that code shouldn't execute.  

LIBXMM could temporarily be handled the same way (only set up `TensorContractCreate` if using double precision), or we could go ahead and add single precision versions of it, or we could go ahead and add single precision versions of everything (including MAGMA).  The optimal parameters for the MAGMA non-tensor backend will no doubt be different for SGEMM vs DGEMM, but that could be changed in a future PR if we want something fast to add now.

jeremylt:
> You mean the tests themselves? The Fortran interface should work fine in single precision provided the user has real*4 type data instead of real*8 (I tried this with a few of the Fortran tests and confirmed they worked), so we don't want to give the impression that Fortran users can't use single precision.

Ah, I see. Yeah, I feel like there should be some loud warning and explanation as to why our Fortran test suite utterly fails for single precision.

>  What I did currently for both MAGMA non-tensor basis create and the AVX backend's TensorContractCreate is just put a check on CEED_SCALAR_TYPE before calling the "connection" for the function in the backend (CeedSetBackendFunction); so in single precision, these backends should still work, but using their delegate backends instead. 

Hmm, I'm not a fan of this. This gives a very mistaken impression to users that they are successfully using AVX, LIBXSMM, etc when they are in fact not. I'd prefer to never actually register backends that do not have single precision support yet so users cannot mistakenly call them.

nbeams:
>  Yeah, I feel like there should be some loud warning and explanation as to why our Fortran test suite utterly fails for single precision.

Do you mean adding something to the documentation, then just letting Fortran tests fail in single (since any required CI testing won't test them)?  Alternatively, if we make separate targets for C and Fortran tests, should we add some single precision Fortran tests which never run in automatic testing, but could be manually tested if desired?  Or is that too messy?

> Hmm, I'm not a fan of this. This gives a very mistaken impression to users that they are successfully using AVX, LIBXSMM, etc when they are in fact not. I'd prefer to never actually register backends that do not have single precision support yet so users cannot mistakenly call them.

I agree it's not ideal for AVX or LIBXSMM in particular.  In the MAMGA case, I didn't want to disable the backend because it's only the non-tensor basis that's affected -- the tensor kernels work in single precision now.  In that way I figured it was kind of like cuda-shared or cuda-gen with non-tensor basis -- they just use cuda-ref instead of giving an error.

But, it sounds like maybe you prefer adding single precision support to each of these backends in this PR, or at least for MAGMA and LIBXSMM?  (I don't know enough about AVX to say whether supporting both double and single would be simple or difficult.)

nbeams:
Oh, another thing I forgot about the Fortran tests is that they would work if the user had `real*4` *AND* we updated the tolerances as we have done for all the C tests.  Except `CEED_EPSILON` isn't exposed to the Fortran tests...

nbeams:
For the XSMM backend, I tried adding 
```
#define LIBXSMM_GEMM libxsmm_dgemm
#define LIBXSMM_MMFUNCTION libxsmm_dmmfunction
#define LIBXSMM_MMDISPATCH libxsmm_dmmdispatch
```
to `ceed-f64.h` (and the corresponding definitions to `ceed-f32.h`), then replacing the function calls with the macros in 
the backend itself, in order to avoid having `if (CEED_SCALAR_TYPE == ` checks for each call.  Does this seem like a good idea, or would the `if` checks be preferred?

The preference for XSMM could affect what we pick for MAGMA.

I also realized my "solution" for checking both output files (double and single) in `tap.sh` may not have been such a solution.  When testing on Noether, I get this error from `make prove`: 
```
 Parse errors: Tests out of sequence.  Found (3) but expected (2)
                Tests out of sequence.  Found (4) but expected (3)
                Tests out of sequence.  Found (6) but expected (4)
                Tests out of sequence.  Found (7) but expected (5)
                ...(and many more lines like this)...
                Tests out of sequence.  Found (39) but expected (26)
                Bad plan.  You planned 39 tests but ran 26.
```
Strangely, I didn't see this on the machine I was testing CUDA on -- I guess different versions of `prove`? 
This was easily fixed by changing `tap.sh` to always run both tests (if both output files exist), and keep track of whether one of them passed with a variable.

I haven't pushed any of the above-discussed changes yet, but obviously I can/will.

However, after doing that, I realized we had another problem -- the last few digits of output for t300 and t320 appear to be sensitive enough to change with the compiler in single precision.  Using gcc on Noether, the tests pass (with the output files I created with a different version of gcc on an ICL machine), but clang fails...

We could go back to the idea of having a format as an optional parameter to the basis view, and use fewer digits in single precision, though we'd still have Python issues...unless we want to just skip these two tests in Python in single precision, or unless we want the Python interface to be able to take an optional format parameter as well, and just document that it should be a C format string? Or add something in to convert the Python parameter (however it's defined) into a C format string? Or some other option...?

jeremylt:
> For the XSMM backend, I tried adding
```
#define LIBXSMM_GEMM libxsmm_dgemm
#define LIBXSMM_MMFUNCTION libxsmm_dmmfunction
#define LIBXSMM_MMDISPATCH libxsmm_dmmdispatch
```
>to ceed-f64.h (and the corresponding definitions to ceed-f32.h), then replacing the function calls with the macros in
the backend itself, in order to avoid having if (CEED_SCALAR_TYPE == checks for each call. Does this seem like a good idea, or would the if checks be preferred?

I think I would prefer that logic living in the xsmm backend. Eventually we are going to want both single and double precision. I'd be ok with replicating `ceed-xsmm-tensor.c` to create a `ceed-xsmm-tensor-single.c` (and the same with AVX). I can put that together in a commit later this morning.

> However, after doing that, I realized we had another problem -- the last few digits of output for t300 and t320 appear to be sensitive enough to change with the compiler in single precision. Using gcc on Noether, the tests pass (with the output files I created with a different version of gcc on an ICL machine), but clang fails...

Oh, that makes sense. Honestly, I'd be ok with skipping the basis output tests in single precision. That's probably simpler overall and it is not like single precision basis view is a critical test that we need.

Thoughts on either of these @jedbrown?

jeremylt:
And for AVX: 5308b899823ee47e5e54b675c0066e581026e9ec

jeremylt:
Small updates to my proposed single and double variants for LIBXSMM and AVX backends:
9ca92f29db7ba5f06cede47ed64551a9ca3a0310
9398dcb0d5165dc2bb197d86204cad0dd9f11ef3

include/ceed/ceed-f64.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef _ceed_f64_h
+#define _ceed_f64_h
+
+/// Set base scalar type to FP64.  (See CeedScalarType enum in ceed/ceed.h
+/// for all options.)
+#define CEED_SCALAR_TYPE CEED_SCALAR_FP64
+typedef double CeedScalar;
+
+/// Data alignment of 64 bits
+#define CEED_ALIGN 64
```
jedbrown, line 29: Same here

README.rst
```diff
@@ -0,0 +1,518 @@
+libCEED: Efficient Extensible Discretization
+============================================
+
+.. image:: https://github.com/CEED/libCEED/workflows/C/Fortran/badge.svg
+   :alt: GitHub Actions
+   :target: https://github.com/CEED/libCEED/actions
```
jedbrown, line 6: It's README.md now, so this can be removed.

jedbrown:
Do you have passing f32 tests?
```
not ok 38 t121-vector /gpu/cuda/gen stdout
# + Error in alpha x + y, computed: 10.000000 actual: 5.000000
# + Error in alpha x + y, computed: 11.000000 actual: 5.500000
# + Error in alpha x + y, computed: 12.000000 actual: 6.000000
# + Error in alpha x + y, computed: 13.000000 actual: 6.500000
# + Error in alpha x + y, computed: 14.000000 actual: 7.000000
# + Error in alpha x + y, computed: 15.000000 actual: 7.500000
# + Error in alpha x + y, computed: 16.000000 actual: 8.000000
# + Error in alpha x + y, computed: 17.000000 actual: 8.500000
# + Error in alpha x + y, computed: 18.000000 actual: 9.000000
# + Error in alpha x + y, computed: 19.000000 actual: 9.500000
```

nbeams:
Yes, all f32 tests pass for me locally (until it gets to the Fortran tests).  I've tried both GCC 8.3 and clang 9.0 with CUDA 11.0.  What combination are you using?

README.rst
nbeams: Whoops, not sure how this made it in after the merge from main.  Removed now in  593b5cac93e94ddcc5cbe975940a49cb9548585c

include/ceed/ceed-f32.h
```diff
@@ -0,0 +1,35 @@
...
+#ifndef _ceed_f32_h
+#define _ceed_f32_h
+
+/// Set base scalar type to FP32.  (See CeedScalarType enum in ceed/ceed.h
+/// for all options.)
+#define CEED_SCALAR_TYPE CEED_SCALAR_FP32
+typedef float CeedScalar;
+
+/// Data alignment of 32 bits
+#define CEED_ALIGN 32
```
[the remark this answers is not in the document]
nbeams: Thanks for the explanation.  This should be fixed in 3a4531abfacee51245ead60c6fa3476017935548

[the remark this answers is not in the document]
nbeams: (I just moved it back to where it was for now, and left as 64)

jedbrown:
I think the lv-cuda test will pass on `main` after #803 was merged. I get the full test suite passing with `make test FC=`. We should have some CI test, even if it's a limit subset of `f32` functionality.

nbeams:
All of the C tests and ex1/ex2 should pass with f32, if we can come up with a good way to change ceed.h to include the f32 header and recompile the library first without interfering with other CI jobs on the machine.

I've never tried editing anything for the CI setups before.  Is there a way to test what it would do without actually pushing to the branch to try it?

jeremylt:
I added a draft of the code to add to the GitLab CI YAML above. I think running the core tests in f32 with FC= as a separate CI job on Noether should be fine.

nbeams:
But what about the separate compilation and change to `ceed.h`?  Can we ensure this job will not run in parallel with any of the other CI jobs on Noether?

jedbrown:
Each CI job has its own working directory and I think we only do one CI job at a time (could increase number of slots, but this could be an issue for GPUs).

nbeams:
Is the working directory just for building, or does it temporarily copy the source (since we'd need to edit the header in the source)?  Though I guess it wouldn't matter as long as the jobs run in serial.

jedbrown:
We'd just use `sed` to edit the header. gitlab-runner is responsible for cleaning the repo before the next job runs.

jedbrown:
Let's just drop the `float` label for now. I can add the runner slot, but we'll want to avoid overloading the GPU. I probably need to configure 1-node SLURM or make a different solution to moderate over-subscription. (CPUs have the lovely property that oversubscription just makes them slow, but GPUs crash.)

nbeams:
Since it's "stuck," does it have to be manually stopped or something?  I'm not seeing GitLab showing up after my force push.

nbeams:
Looks like the float job passed!  (Assuming it truly ran in single precision, that is.  The sed commands show up on the output, so it should have worked.) 

I think I'm going to official remove the WIP tag from this PR --  in my opinion, we should be ready for review with the intent of squash-merging.

pazner:
Are there plans to also run the tests for the various language bindings in single precision mode? Or just the C test suite?

jeremylt:
It would be nice to have single precision versions of Python, Rust, and Julia CI. These should be easy to add as separate jobs in those GitHub actions. I could do that Tuesday, possibly in a separate PR if we're eager to merge this one.

jedbrown:
As this is experimental and evolving, I'm fine merging it without every-language CI.

jedbrown (approved):

pazner (approved):

jedbrown:
@nbeams Would you like to selectively squash commits (e.g., you have some WIP commit messages) or squash-merge the whole thing?

nbeams:
With so many commits, I was worried that getting a nice rebase might be quite the headache, but I'll give it a go and see if I can clean it up a bit.

jedbrown:
Cool. One (somewhat extreme) approach is to squash the entire history into one, then amend (`git citool` and some others like Emacs Magit can help) to split out some parts that deserve to be separate. I'm also okay with the whole thing being one commit if you think that makes sense.

nbeams:
Well, I spent almost an hour trying to come up with a nice history through rebase, and I'm not sure it's that much better (and I once again appear to have messed up the Julia conflict resolution process during the rebase, because the final result has some differences from the branch as it is here, so I definitely won't be pushing that version). 

Personally, I'm okay with squash-merging the whole thing and replacing the default squash-merge commit message with a descriptive message explaining the major changes (addition of the new headers, addition of new enum, changes to the tests to use epsilon, and various changes to the other language bindings to get CeedScalar information).  There are a lot of files changed, but most everything fits within those categories, I think?  However, I can try again if there is a strong preference for keeping separate commits.  I don't think I understand your suggestion @jedbrown, how to pick out individual commits after a squash -- if you want me to try this, do you have a link to an explanation of this somewhere?

There were some minor bugfixes included in this branch (notably changing the location of `CeedQFunctionContextGetContextSize`).  @jeremylt mentioned that I might want to do a PR with just those changes  that points to `release` rather than `main` -- are we starting this new workflow yet?  The problem is that changes that are strictly bugfixes are not nicely separated from other changes in the commits, since I wasn't yet thinking in terms of this framework when "dividing up" commits.  So we'd still ultimately have some differences of git history between `release` and this branch/`main` even without a squash-merge over here (can't just cherry-pick commits to go to `release`).

jedbrown:
I'll just go ahead and merge. We can screenshare demo on a smaller PR sometime. Basically, you run `git citool`, which brings up a GUI and click "Amend", which allows you to deselect files/hunks/lines from the last commit, commit, then add new commits.

jedbrown:
Congrats @nbeams on this big lift. :rocket: And thanks to @pazner @YohannDudouit @jeremylt for the discussion and contributions.

nbeams:
Yes, a big thank you to everyone who contributed to this PR!  A team effort, for sure.  🎉

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -28,15 +28,15 @@ static const char *atomicAdd = QUOTE(
 //------------------------------------------------------------------------------
 // Atomic add, for older CUDA
 //------------------------------------------------------------------------------
-__device__ double atomicAdd(double *address, double val) {
+__device__ CeedScalar atomicAdd(CeedScalar *address, CeedScalar val) {
```
YohannDudouit, line 31: I think this one should stay `double` or simply be removed with `float`,  `atomicAdd` for `float` is always defined.

jedbrown: It's in CUDA-8 and later? We don't test ancient versions and I think it doesn't hurt us to just not support versions less than 8.

PR #789, opened 2021-07-08
[WIP] Julia interface updates (Preferences.jl)

pazner:
Some updates to the Julia interface.

#### Preferences system
The main change is using [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) to allow the user to provide their own libCEED shared library binary.

This is more consistent with the rest of the Julia ecosystem, and means we don't have to roll our own configuration. It will also work with [depot-wide overrides](https://docs.binarybuilder.org/stable/jll/#Overriding-the-artifacts-in-JLL-packages).

It also enables some more features, like using different libCEED libraries in different projects/environments simultaneously.

The same preference system can be used in a straightforward way to enable single precision (cf. #788).

#### Other changes

* Change `CC_VENDOR` in the makefile so that it works better with cross-compilation (e.g. to build libCEED_jll). This will allow for the simplification of the build script in libCEED_jll.
* Bumps the libCEED_jll version to 0.9 (latest libCEED release)
* Moves the development tests to release tests (since new version was released)
* Makes the FFI code closer to the auto-generated code, working towards automating it completely (will probably wait for the next release of Clang.jl since there will be breaking interface changes in the generator code)


pazner:
@jedbrown @jeremylt does one of you want to take a look at this PR?

jeremylt:
There is a conflict in the top level README because we changed over our documentation to MyST instead of RST, but it should be easy to resolve.

The Julia changes look good to me. For my own education, can you explain how the change to `CC` works? I feel like I'm missing something obvious.

pazner:
> There is a conflict in the top level README because we changed over our documentation to MyST instead of RST, but it should be easy to resolve.

Done.

> The Julia changes look good to me. For my own education, can you explain how the change to `CC_VENDOR` works? I feel like I'm missing something obvious.

This was @jedbrown's suggestion, it's already merged in main (commit [f0d101d](https://github.com/CEED/libCEED/commit/f0d101d5edb6e1600bc546c173bf76a954ebcea6))

jeremylt (approved):

julia/LibCEED.jl/src/LibCEED.jl
```diff
@@ -145,11 +149,104 @@ include("Request.jl")
...
+documentation for more information. If a depot-wide Overrides.toml file is present (e.g. in
+~/.julia/artifacts, a different libCEED library may still be used).
+"""
+function use_prebuilt_libceed!()
+    set_preferences!(C.libCEED_jll, "libceed_path" => nothing; force=true)
+    @info("""
+          Using the prebuilt libCEED binary.
+          Restart the Julia session for changes to take effect.
+          """)
+end
```
jedbrown, line 242: Do we need both modes? It sounds like they're similar and it may be confusing which should be used.

jedbrown (approved):
Thanks!

julia/LibCEED.jl/src/LibCEED.jl
pazner: Probably not… they are similar, and in most cases will give the same result.

The difference is that `use_prebuilt_libceed!` will "force" the library to use the prebuilt library, whereas `unset_libceed_path!` will tell LibCEED.jl that the user has no preference. Usually, this will also mean using the prebuilt library, but there are situations where this default could be changed somewhere else (for example, if there is a shared installation of libCEED and LibCEED.jl, then the default library path could be changed for everyone using this shared installation).

jedbrown: Would it be idiomatic to consolidate something like this?
```julia
set_libceed_path!("path/to/libceed.so")
set_libceed_path!(:prebuilt) # Always use pre-built, regardless of environment
set_libceed_path!(:default)  # Try environment, then fall back to prebuilt
```

pazner: To me it seems a little strange to call `set_libceed_path!` with an argument that isn't a path.

`use_prebuilt_libceed!` is more expressive in my opinion. We can probably just get rid of `unset_libceed_path!`, the user can always unset the preferences manually (either by calling `set_preferences!` or just editing the preferences TOML file).

Do you think that would make sense?

jedbrown: Yeah, that's fine. My thinking was that when you give a user one lever, they `set` it wherever they want and the implication is that the history doesn't matter. But when you give them several levers, it can be hard to tell whether they interact, whether there can be any order/history dependence between `use_prebuilt` and `set`. Similar for a key in the TOML. If you set one key, then it always has a unique and well-specified state after parsing. If there are several keys, then you need some logic to determine whether the state is invalid and whether to report the error or resolve it according to some rule.

pazner: Yet another option is to implement the interface you proposed, and keep `use_prebuilt_libceed!` as an alias. The docstring would be something like:
```
    use_prebuilt_libceed!()

Indicates that the prebuilt version of the libCEED library (bundled with libCEED_jll) should be
used. Equivalent to `set_libceed_path!(:prebuilt)`.
```

jedbrown: It's more than one way to do it, but entirely clear about effects/side-effects and tab-completes better than `set_libceed_path!(:prebuilt)`.

pazner: Agreed, done.

jedbrown (approved):
Looks good. Shall we squash-merge, merge with all these commits (some look like noise), or consolidate to more than one commit and merge that?

pazner:
> Looks good. Shall we squash-merge, merge with all these commits (some look like noise), or consolidate to more than one commit and merge that?

I consolidated some of the commits, I think each commit should be meaningful now for the most part.

PR #793, opened 2021-07-21
ex - remove recast of fixed length array as vla

jeremylt:
GCC 11 complains (rightly so) about this re-cast of a fixed length array as a VLA in the function argument list.

### DIFF
--- a/examples/ceed/ex1-volume.c
+++ b/examples/ceed/ex1-volume.c
@@ -56,12 +56,12 @@
 #include "ex1-volume.h"
 
 // Auxiliary functions.
-int GetCartesianMeshSize(int dim, int degree, int prob_size, int num_xyz[3]);
-int BuildCartesianRestriction(Ceed ceed, int dim, int num_xyz[3], int degree,
+int GetCartesianMeshSize(int dim, int degree, int prob_size, int num_xyz[dim]);
+int BuildCartesianRestriction(Ceed ceed, int dim, int num_xyz[dim], int degree,
                               int num_comp, CeedInt *size, CeedInt num_qpts,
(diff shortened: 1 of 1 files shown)


examples/ceed/ex1-volume.c
```diff
@@ -303,7 +303,7 @@ int main(int argc, const char *argv[]) {
   return 0;
 }
 
-int GetCartesianMeshSize(int dim, int degree, int prob_size, int num_xyz[dim]) {
+int GetCartesianMeshSize(int dim, int degree, int prob_size, int num_xyz[3]) {
```
jedbrown, line 306:
```suggestion
int GetCartesianMeshSize(int dim, int degree, int prob_size, int num_xyz[]) {
```
Isn't it shorter when `dim < 3`?

jeremylt: Ah, the real issue is line 59

jeremylt: fixed

jedbrown (approved):
I think this is okay because it's limited to an example, but we should avoid explicit use of VLA in public headers because some user compilation units/parsers won't support it.

jeremylt:
There are a couple of small places internally where we use VLA over malloc and I've thought about switching over to malloc/free for a user that might prefer a compiler that doesn't support VLA anymore.

jedbrown:
I'm not concerned about internal use (hasn't been a portability issue yet), but we can use `-std=c99` for those compilation units. The public header is included in the user environment, which might use `-std=c89`, other toolchain, or not be a compiler at all (binding generation tool).

PR #795, opened 2021-07-23
Refactor Preconditioning Ingredients

jeremylt:
This PR refactors the preconditioning ingredients to ~~remove the fragile fallback mechanism.~~ raise appropriate parts to interface level.

ToDo:
- [x] Raise default impl to interface level
- [x] Documentation improvements
- [x] Reduce redundency
- [x] Interface default allow alternate qvec orderings


jeremylt:
As I am taking pieces apart, the fallback mechanism might still be the best way to handle this on the GPU side, but I still think moving this up to an interface level default is the right call. This PR is also pulling the preconditioning related content into a separate file to help break up that massive operator interface level file.

jeremylt:
Mostly done except for some tidying, so commentary on this general approach is welcome.

interface/ceed-preconditioning.c
```diff
@@ -0,0 +1,1993 @@
...
+  @param[in] op      CeedOperator to assemble nonzero pattern
+  @param[in] offset  Offset for number of entries
+  @param[out] rows   Row number for each entry
+  @param[out] cols   Column number for each entry
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Developer
+**/
+int CeedSingleOperatorAssembleSymbolic(CeedOperator op, CeedInt offset,
```
jedbrown, line 462: `static` if they can't be called by a user. (I don't think these should be callable by a user.)

jedbrown (commented):
I'm generally not opposed to fallbacks living at interface level, but they should all have backend-specific overrides.

We have `CeedOperatorLinear*` and this could be `CeedOperatorSingle.*`. This naming is useful when grepping/tab-completing.

It's not 100% clear to me why this is more maintainable than the previous fallback mechanism.

interface/ceed-preconditioning.c
jeremylt: Good call. I'll do a pass for that while I'm seeing if I can reactor any of the full assembly to work reduce overlap with the diagonal assembly.

jeremylt:
Ultimately I ended up keeping the fallback mechanism largely the same for the GPU backends. The biggest change is moving diagonal assembly and FDM inverses up to the interface level so every backend has at least one impl. I'm not sure if I will be able to cleanly do so with QFunctions as well. 

Partly of the motivation is to provide a basic impl of these functions for any new backend, and part of the motivation is to make it easier for someone who is not as familiar with the fallback mechanism to make additional contributions to our preconditioning ingredients.

jeremylt:
Should be ready now. It is a fair amount of code movement, but but a smaller amount of logic change. Now the `t54*` series tests pass on CUDA and HIP rather than being skipped.

doc/sphinx/source/releasenotes.rst
```diff
@@ -9,6 +9,10 @@ for each release of libCEED.
 Current ``main``_branch
 -----------------------
 
+Maintainability
+^^^^^^^^^^^^^^^
+* Refactored preconditioner support internally to facilitate future development
```
jedbrown, line 14:
```suggestion
* Refactored preconditioner support internally to facilitate future development and improve GPU completeness/test coverage.
```
I think this is accurate?

jedbrown (approved):
Looks good to me.

PR #796, opened 2021-07-28
doc - add citing yaml

jeremylt:


jedbrown:
Does this work for you?
image

jeremylt:
The button does not work for me either. Unfortunately I don't know if its a bug with it being on a non-main branch or a bug with the yml or a bug with my browser or what.

jeremylt:
Ah-ha. It works better now.

CITATION.cff
```diff
@@ -0,0 +1,178 @@
...
+    given-names: Thilina
+    orcid: 0000-0001-6102-6560
+  - family-names: Thompson
+    given-names: Jeremy L
+    orcid: 0000-0003-2980-0899
+  - family-names: Tomov
+    given-names: Stanimire
+    orcid: 0000-0002-5937-7959
+date-released: 2021-07-07
+
```
jedbrown, line 49:
```suggestion
doi: 10.5281/zenodo.5080234

```
This should work according to this example https://github.com/citation-file-format/citation-file-format

jeremylt: I got that one. The "view citation file" button still doesn't work for me, and I'm not sure exactly why, but the generation of the APA and Bibtex is working correctly.

jedbrown: Yeah, I was just hoping to add the DOI because the current approach seems really poorly thought out (no URL, no publisher, no DOI, just the software name).

jeremylt: Yea. I swept that change up into my other changes to the base file.

jeremylt: Looking at the generation citations, it seems that the DOI is the most important of the three - it is the only one present in the APA format

CITATION.cff
```diff
@@ -0,0 +1,186 @@
+cff-version: 1.1.0
+title: "libCEED: the CEED Library"
```
jedbrown, line 2: Do we want this to be our tagline? I know it's in the README, but it seems circular and uninformative. Just throwing out ideas:

1. libCEED: CEED library for Efficient Extensible Discretization
2. libCEED: Community library for Efficient Extensible Discretization
3. libCEED: Code for Efficient Extensible Discretization  *(the original)*
4. libCEED: Efficient Extensible Discretization

jedbrown: Is the a `CITATION.cff` convention for escaping? The bibtex has
```
title = {libCEED: the CEED Library},
```
which will not capitalize correctly.

jeremylt: I think it will put in whatever we have it the quotes literally.

I agree that this tagline isn't great. We could replace both at the same time. I like 3 or 4 personally but perhaps we should poll others?

jeremylt: Update: escaping in the string has some mildly silly results:
```
@misc{Thompson_{LFAT}oolkit.jl_2021,
author = {Thompson, Jeremy L},
doi = {10.5281/zenodo.5064056},
month = {7},
title = {{LFAT}oolkit.jl},
url = {https://jeremylt.github.io/LFAToolkit.jl/stable/},
year = {2021}
}
```

jedbrown: https://github.com/citation-file-format/citation-file-format/issues/258

jedbrown:
This would help when they implement it https://twitter.com/stdruskat/status/1420289082211672064

CITATION.cff
jedbrown: I'll resolve this. You can follow the above issue for proper case handling, which looks like it may take a while.

README.rst
```diff
@@ -33,12 +33,13 @@ libCEED: the CEED Library
...
+Summary and Purpose
 --------------------------------------------
 
-This repository contains an initial low-level API library for the efficient
-high-order discretization methods developed by the ECP co-design
-`Center for Efficient Exascale Discretizations (CEED) <http://ceed.exascaleproject.org>`_.
+libCEED provides fast algebra for element-based discretizations, designed for
+performance portability, run-time flexibility, and clean embedding in higher
+level libraries and applications. It offers a C99 interface as well as bindings
+for Fortran, Python, Julia, and Rust.
```
jedbrown, line 42: I feel like this portrayal of ECP CEED may discourage community contributions so would rather leave it out of the opening sentences.

jedbrown (approved):

Issue #778, opened 2021-06-03
Mixed precision functionality


nbeams:
This issue is for discussions related to adding mixed precision capabilities to libCEED, including:

 - Which user-facing functions should have mixed precision capability?
 - How should the libCEED interface support this (what will the user need to do)?
 - What combinations should be available (e.g. mix-and-match input and output precisions, ability to decouple input/output precision from precision of computations, ...)
 - Potential for backend specializations, and how this will fit with the interface changes
 
An idea from @jedbrown was to handle precision conversions through the `CeedVector` objects, which could be modified to store multiple pointers for different precisions, along with a tag to indicate which precision that vector's data currently uses.

jedbrown:
I think qfunction variation is the hardest part. Do we need to support different arguments to the qfunction having different precisions or can they be homogeneous (all arguments in matching precision)? Let's defer this part for later.

As a first step, I would add functions like `CeedVectorGetArray_f32(CeedVector, CeedMemType, float **)` and have the implementation convert as needed. We can use C11 `_Generic` so C11 users don't need to specify these types and C++ overloading to do the same for C++ users.

We can have a macro `CEED_SCALAR_F32` that users can define before including `ceed.h` so that C99 callers get suitable `CeedScalar` and appropriate specialization. Perhaps better is to provide `#include <ceed/ceed_f32.h>` that sets the macro and includes `ceed/ceed.h`). This would be for callers who want to choose precision at compile time.

We then add a `CeedElemRestrictionSetScalarType(CeedElemRestriction, CeedScalarType evec_type)` that converts if necessary (when provided with an L-vector). This will require visiting some backends for the cross-type packing, but I don't think will be a lot of code.

I kinda think `CeedBasis` public interfaces (setting matrices and doing linear algebra) could just stay double precision all the time, and find it more compelling to possibly add `__float128` than to support direct specification in single precision (they can be converted as needed behind the scenes).

We can have `CeedVectorGetArrayUntyped(CeedVector, CeedMemType, CeedScalarType, void *)` to make life easier for dynamic callers who prefer to cast somewhere else.

nbeams:
> I think qfunction variation is the hardest part. Do we need to support different arguments to the qfunction having different precisions or can they be homogeneous (all arguments in matching precision)? Let's defer this part for later.

Just to make sure I'm clear: are you referring to the user-provided per-point QFunction, the libCEED QFunction interface, or both?

[...]

PR #797, opened 2021-08-02
Refactor tests to reduce use of output

jeremylt:
This PR adjusts the C tests so that only the tests of the `Ceed*View` functions make output.


jeremylt:
Woops, forgot the Python tests

jeremylt:
Ok, all set now

tests/t202-elemrestriction-f.f90
```diff
@@ -2,66 +2,98 @@
...
-
+      call ceedvectorgetarrayread(x,ceed_mem_host,xx,xoffset,err)
+      do i=1,ne+1
+        diff=xx(xoffset+i)
+        if (i > 1 .and. i < ne+1) then
+          diff=diff-2*(10+i-1)
+        else
+          diff=diff-(10+i-1)
+        endif
+        if (abs(diff) > 1.0D-15) then
```
jedbrown, line 83: Should we refrain from hard-coding new constants, or fix all these in a batch with #778 #788 ?

tests/t208-elemrestriction-f.f90
```diff
@@ -2,62 +2,94 @@
...
       character arg*32
-
+  
       call getarg(1,arg)
       call ceedinit(trim(arg)//char(0),ceed,err)
-
+  
       call ceedvectorcreate(ceed,ne+1,x,err)
-
+
```
jedbrown, line 29: Are these adding trailing whitespace?

jedbrown (approved):

tests/t202-elemrestriction-f.f90
jeremylt: I was thinking this could be fixed over there

tests/t208-elemrestriction-f.f90
jeremylt: Oh, that was my IDE adding bad extra spaces. I'll fix that

jeremylt: Fixed

Issue #729, opened 2021-04-06
Style Proposal


jeremylt:
Edit: Updating with a ToDo list

ToDo:
- [x] Update C style (#735)
- [x] Update fluids example
- [x] Update PETSc example suite (#726)
- [x] Update documentation (#703)

---

As @valeriabarra has pointed out a few times, our mix of styles is a bit of a mess. I can do a pass to unify and write a style guide, but I'd like to have some buy-in on the format.

Here is what I propose

* Documentation (currently very inconsistent)
  - One sentence per line
* C99/C++ (currently very inconsistent)
  - Google style variable/function names (we use modified Google style in `.astyle`)  (https://google.github.io/styleguide/cppguide.html)
  - Function names: `PascalCase` (current)
  - Type names: `PascalCase` (current)
  - Constant names: `UPPER_SNAKE_CASE` (current, different than Google)
  - Variable names: `snake_case` (currently we use a mix)
  - Struct variables: `snake_case`  (currently we use a mix)
* Python (currently consistent, change suggested)
  - PEP 8
  - Function/method names: `snake_case` (current)
  - Variable names: `snake_case` (current)
  - Whitespace: tab (proposed change, better accessibility)
* Rust (currently consistent)
  - Default Rust style (https://rust-lang.github.io/api-guidelines/naming.html)
  - Function/method names: `snake_case` (current)
  - Variable names: `snake_case` (current)
* Julia (currently consistent)
  - Default Julia style (https://docs.julialang.org/en/v1/manual/style-guide/)
  - Function/method/variable names: `nocase` or `snake_case` (current)
* Fortran (currently consistent)
  - defer modification until F77 support decision

I'm sure there are other things that can/should be added. This was just my first pass.

valeriabarra:
Thanks Jeremy for opening the issue and the dicsussion.
I personally prefer anything but `nocase` and the current Fortran style (LOL), for acccessibility.

jedbrown:
Underscores in Fortran are fraught. Some compilers mangle `foo` to `foo_` but `foo_bar` to `foo_bar__`, for example. The simplest convention is to avoid underscores in Fortran names, which are case-insensitive. Alternatively, we could use ISO C bindings, which fixes all of this and would allow an idiomatic modern Fortran interface.

I don't follow the desire to switch Python indentation to tabs. An advantage of spaces is you can always copy/paste and display using any tool while preserving indentation width. I'm also not wild about different conventions in different languages.

[...]

PR #799, opened 2021-08-05
Solids - some cleanup

rezgarshakeri:
I have cleaned-up Qfunctions. They are consistent with the documentation.

### DIFF
--- a/examples/solids/qfunctions/FSCurrent-NH1.h
+++ b/examples/solids/qfunctions/FSCurrent-NH1.h
@@ -89,34 +89,34 @@ CEED_QFUNCTION_HELPER CeedScalar computeJM1(const CeedScalar grad_u[3][3]) {
 // Common computations between Ftau and dFtau
 // -----------------------------------------------------------------------------
 CEED_QFUNCTION_HELPER int commonFtau(const CeedScalar lambda,
-                                     const CeedScalar mu, const CeedScalar grad_u[3][3],
+                                     const CeedScalar mu, const CeedScalar Grad_u[3][3],
                                      CeedScalar F_inv[3][3], CeedScalar tau_work[6],
                                      CeedScalar *llnj) {
 
 
-  // Compute The Deformation Gradient : F = I3 + grad_u
+  // Compute The Deformation Gradient : F = I3 + Grad_u
   // *INDENT-OFF*
-  CeedScalar F[3][3] =  {{grad_u[0][0] + 1,
-                          grad_u[0][1],
-                          grad_u[0][2]},
-                         {grad_u[1][0],
-                          grad_u[1][1] + 1,
-                          grad_u[1][2]},
-                         {grad_u[2][0],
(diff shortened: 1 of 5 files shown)


jedbrown (approved):
LGTM. We may want to write a little matrix math utility library to do things like compute `F^T F` and put the result in Voigt notation, compute the inverse of a Voigt matrix as a Voigt matrix, etc. I know our headers don't currently include other headers (a place to put such inline utility functions instead of copy-pasta into each file), but I think we could make that work and it would greatly clean up this sort of code.

examples/solids/qfunctions/FSCurrent-NH1.h
```diff
@@ -85,66 +85,79 @@ CEED_QFUNCTION_HELPER CeedScalar computeJM1(const CeedScalar grad_u[3][3]) {
...
+
+  return 0;
+};
+#endif
+
 // -----------------------------------------------------------------------------
 // Common computations between Ftau and dFtau
 // -----------------------------------------------------------------------------
 CEED_QFUNCTION_HELPER int commonFtau(const CeedScalar lambda,
-                                     const CeedScalar mu, const CeedScalar grad_u[3][3],
```
jeremylt, line 92: The underscore for `F_inv` was intentionally added. I'm not sure why `Grad_` should be capitalized here - it isn't in other QFunctions

jeremylt: I haven't thoroughly made every variable in all of the different solids QFunctions match our style: https://github.com/CEED/libCEED/issues/729

rezgarshakeri:
Capital “Grad” means derivative respect to initial coordinate while “grad” is in current configuration.
I don’t know maybe we need to show the difference between “initial” and “current “ by something else here, but what I used is normal notation in continuum mechanics

examples/solids/qfunctions/FSCurrent-NH1.h
jeremylt: That makes sense if it is typical. I'm not an expert in continuum mechanics notation so I'm not sure what the common convention would be in that case.

We still are missing a bunch of underscores between words though. `Finv` vs `F_inv` and `tempGradu` vs `temp_Grad_u` though

jedbrown: I wouldn't stress about this because I think we'll be refactoring this anyway to decouple material from structural concerns. (The exact way in which we refactor depends on empirical tests of what optimizations we can rely on compilers to do.)

jedbrown:
Should we merge this? Does it interact with #801 (@jeremylt)?

jeremylt:
I don't see any reason this should cause problems for the refactor - I'm not planning on reaching into the qfunctions

jedbrown:
Cool. I'll merge if you approve.

jeremylt (approved):

PR #800, opened 2021-08-06
Fluids - Small refactor

LeilaGhaffari:
This sets up QFunction contexts in each compilation unit.


jedbrown (approved):
LGTM. Some parts, like `PetscSectionSetComponentName` calls in `setupdm.c`, will need to be abstracted for the port to primitive variables. It might make sense to knock that out soon, either by adding a callback there or by having each "problem" provide its list of names.

LeilaGhaffari:
> LGTM. Some parts, like `PetscSectionSetComponentName` calls in `setupdm.c`, will need to be abstracted for the port to primitive variables. It might make sense to knock that out soon, either by adding a callback there or by having each "problem" provide its list of names.

Agreed! I should be able to fix that when we add the primitive variables. 
Is there a reason cuda is failing? I don't think it is related to this PR.

jedbrown:
Yeah, we've been getting this failure due to cuda drivers on lv.
https://gitlab.com/libceed/libCEED/-/pipelines/349263618/test_report
image

I restarted and re-ran the jobs. One unrelated failed on CUDA, perhaps because a tolerance is slightly too strict.

jeremylt:
Do we want to go ahead and merge this and the solids refactor while we work on the CUDA CI fix?

LeilaGhaffari:
I just rebased it and I don't think it needs to wait for the CUDA CI fix.

PR #801, opened 2021-08-11
Refactoring Solids to Remove Switches

jeremylt:
ToDo:
- [x] Create struct with problem data from `problem_options` in `src/setup-libceed.c`
- [x] Need clean way to handle array of field names in problem data struct (array of strings in struct)
- [x] Refactor `SetupLinearFineLevel()` and `SetupLibceedLevel()` to accept pointer to struct of problem data instead of using switches, will use loop over field name array to handle different number of storage
- [x] Swap to kabob case file names as refactoring
- [x] Create separate functions to process MR/NH physics data from command line and return phys struct
- [ ] and more I'm not thinking of....


examples/solids/Makefile
```diff
@@ -79,11 +79,11 @@ $(PETSc.pc):
...
 # E = 2 * mu + (1 + nu)
-output/NH-strain.csv: args = -problem FSInitial-NH1 -E 2.8 -nu 0.4
-output/MR-strain.csv: args = -problem FSInitial-MR1 -mu_1 1 -mu_2 0.0 -nu .4
-output/MR-strain1.csv: args = -problem FSInitial-MR1 -mu_1 .5 -mu_2 0.5 -nu .4
+tests-output/NH-strain.csv: args = -problem FSInitial-NH1 -E 2.8 -nu 0.4
+tests-output/MR-strain.csv: args = -problem FSInitial-MR1 -mu_1 1 -mu_2 0.0 -nu .4
+tests-output/MR-strain1.csv: args = -problem FSInitial-MR1 -mu_1 .5 -mu_2 0.5 -nu .4
 
-output/%.csv: elasticity
+tests-output/%.csv: elasticity
```
jedbrown, line 86: These files are referenced from the Sphinx documentation.

jedbrown (commented):
I take it this is just starting a branch toward this intended refactoring.

examples/solids/Makefile
jeremylt: Good catch, thank you

jeremylt:
I'm starting to pull this apart, but getting the pieces to line up cleanly will take some work.

The basic idea is that there will be a common function without switches for setting up the geometry, residual, jacobian, energy, and diagnostic operators. Each problem compilation unit will call this common setup with a struct holding the QFunction pointers and all of the information for the stored fields.

Depending upon how we want to structure it, we can compose these operators into composite operators.

examples/solids/include/petscmacros.h
```diff
@@ -0,0 +1,12 @@
...
+
+#if PETSC_VERSION_LT(3,14,0)
+#  define DMAddBoundary(a,b,c,d,e,f,g,h,i,j,k,l,m,n) DMAddBoundary(a,b,c,e,h,i,j,k,f,g,m)
+#elif PETSC_VERSION_LT(3,16,0)
+#  define DMAddBoundary(a,b,c,d,e,f,g,h,i,j,k,l,m,n) DMAddBoundary(a,b,c,e,h,i,j,k,l,f,g,m)
+#else
+#  define DMAddBoundary(a,b,c,d,e,f,g,h,i,j,k,l,m,n) DMAddBoundary(a,b,c,d,f,g,h,i,j,k,l,m,n)
+#endif
+
+#endif
```
jedbrown, line 12: Can you ensure your editor ends files with a newline?

examples/solids/problems/smallStrainNeoHookean.c
```diff
@@ -0,0 +1,24 @@
+#include <ceed.h>
+#include "../qfunctions/common.h"
+#include "../qfunctions/smallStrainNeoHookean.h"
+#include "../qfunctions/manufacturedTrue.h"
```
jedbrown, line 4: camelCase naming seems unusual for us in file names.

jedbrown (commented):
I think we want something like `PetscFunctionListAdd(&materials, "neo-Hookean", MaterialSetUp_NH)` [^1] where the callback might have a prefix, set defaults, (this or a *SetFromOptions) read run-time options, and register hooks for any custom diagnostic information. These compilation units can either provide (declarative mode) the auxilliary fields or could create them.

Whatever step works with a real mesh should have a set and collection of values. For example, when we have many beads, we'd have a `DMLabel` (e.g., `"Cell Sets"`) and a list of values (e.g., one per glass bead). This code could be centralized based on a specification from the material or we could have interfaces that the material calls to set itself up on the set of the mesh.

[^1]: We could combine these as `FSCurrent-NH2` and `FSInitial-MR1`, versus having sub-options or a global choice of current vs initial. I suspect when we move to a stand-alone application, we're most likely to formulate in current configuration because it's more convenient for boundary conditions and contact (and for viscoelastics using MPM), but having both is useful in the libCEED demo because their different coordinate mapping offers some inspiration for prospective users.

jeremylt:
The primary shape of this is in place, but it segfaults, so debugging is needed.

rezgarshakeri:
Thanks @jeremylt , It is really nice.

rezgarshakeri (approved):

examples/solids/elasticity.c
```diff
@@ -92,27 +90,11 @@ int main(int argc, char **argv) {
...
   num_levels = app_ctx->num_levels;
   fine_level = num_levels - 1;
 
-  if (app_ctx->problem_choice != ELAS_FSInitial_MR1) {
-    // -- Set Poison's ratio, Young's Modulus
-    ierr = PetscMalloc1(1, &units); CHKERRQ(ierr);
-    ierr = PetscMalloc1(1, &phys); CHKERRQ(ierr);
-    ierr = ProcessPhysics_General(comm, app_ctx, phys, phys_MR, units);
-    CHKERRQ(ierr);
-    if (fabs(app_ctx->nu_smoother) > 1E-14) {
```
jeremylt, line 104: Ahha, this is the one dangling piece. We didn't really do a lot with this separate smoother physics object. Is this something that we want? Is there a useful analog of this for Mooney-Rivlin?

jedbrown: We expressed Mooney-Rivlin with a Poisson ratio nu. (We were initially using bulk modulus, but Poisson ratio is easier for me to reason about.)

jeremylt: Ah, cool. I went ahead and added another function list for updating smoother physics. This PR should be good to merge now.

Issue #802, opened 2021-08-14
CUDA bug for BP1 degree 5 q_extra 1


jedbrown:
Test passes (`CONVERGED_RTOL`) when using `/gpu/cuda/shared`, but not with `/gpu/cuda/ref` (`DIVERGED_ITS`).

It works for `-degree` 4, 6, and 7, and it works with `-q_extra` 0 and 2. This is using CUDA-11.4.

<details>

### Working
```console
$ build/petsc-bpsraw -ceed /gpu/cuda/shared -problem bp1 -degree 5 -q_extra 1 -ksp_max_it_clip 15,15

-- CEED Benchmark Problem 1 -- libCEED + PETSc --
  PETSc:
    PETSc Vec Type                     : seqcuda
  libCEED:
    libCEED Backend                    : /gpu/cuda/shared
    libCEED Backend MemType            : device
  Mesh:
    Number of 1D Basis Nodes (P)       : 6
    Number of 1D Quadrature Points (Q) : 7
    Global nodes                       : 1331
    Process Decomposition              : 1 1 1
    Local Elements                     : 8 = 2 2 2
    Owned nodes                        : 1331 = 11 11 11
    DoF per node                       : 1
  KSP:
    KSP Type                           : cg
    KSP Convergence                    : CONVERGED_RTOL
    Total KSP Iterations               : 9
    Final rnorm                        : 8.185631e-11
  Performance:
    Pointwise Error (max)              : 6.064044e-04
    CG Solve Time                      : 0.00187707 (0.00187707) sec
    DoFs/Sec in CG                     : 6.38176 (6.38176) million
```

### Broken
```console
$ build/petsc-bpsraw -ceed /gpu/cuda/ref -problem bp1 -degree 5 -q_extra 1 -ksp_max_it_clip 15,15

-- CEED Benchmark Problem 1 -- libCEED + PETSc --
  PETSc:
    PETSc Vec Type                     : seqcuda
  libCEED:
    libCEED Backend                    : /gpu/cuda/ref
    libCEED Backend MemType            : device
  Mesh:
    Number of 1D Basis Nodes (P)       : 6
    Number of 1D Quadrature Points (Q) : 7
    Global nodes                       : 1331
    Process Decomposition              : 1 1 1
    Local Elements                     : 8 = 2 2 2
    Owned nodes                        : 1331 = 11 11 11
    DoF per node                       : 1
  KSP:
    KSP Type                           : cg
[...]

PR #803, opened 2021-08-17
testing updates: junit classname, bpsraw tolerances, CUDA on lv

jedbrown:


examples/fluids/src/setupdm.c
```diff
@@ -24,12 +24,14 @@ PetscErrorCode CreateDistributedDM(MPI_Comm comm, ProblemData *problem,
                                    SetupContext setup_ctx, DM *dm) {
   DM               dist_mesh = NULL;
   PetscPartitioner part;
-  PetscInt         dim = problem->dim;
+  PetscInt         dim = problem->dim, faces[3] = {3, 3, 3};
```
LeilaGhaffari, line 27: Does this mean three faces per each dimension? I am not sure if this understanding is correct since I think the default was two faces per dimension?!

Makefile
```diff
@@ -535,7 +535,7 @@ $(OBJDIR)/petsc-% : examples/petsc/%.c examples/petsc/libutils.a.PHONY $(libceed
 	  PETSC_DIR="$(abspath $(PETSC_DIR))" OPT="$(OPT)" $*
 	cp examples/petsc/$* $@
 
-$(OBJDIR)/fluids-% : examples/fluids/%.c $(libceed) $(ceed.pc) | $$(@D)/.DIR
+$(OBJDIR)/fluids-% : examples/fluids/%.c examples/fluids/src/*.c examples/fluids/*.h examples/fluids/problems/*.c examples/fluids/qfunctions/*.h $(libceed) $(ceed.pc) | $$(@D)/.DIR
```
LeilaGhaffari, line 538: Oops! Sorry I missed this. What was the problem it was creating that I never noticed?

jedbrown: I like to build and run from the top level. So I edit a source file in the library *or* in the example, and it should always build cleanly with minimal work. Without these listed prerequisites, editing source files other than `navierstokes.c` would not call the sub-directory make.

LeilaGhaffari: I see! Thanks for clarifying!

examples/fluids/src/setupdm.c
jedbrown: Hmm, I'm not seeing where that would be specified. When Matt refactored DMPlex creation, `DMPlexCreateBoxMesh` no longer calls `DMSetFromOptions` and thus doesn't process options like `-dm_plex_box_faces`. Instead, you can now do everything (read in or create any kind of mesh, distribute, interpolate, refine, extrude) from three lines
```c
DMCreate(comm, &dm);
DMSetType(dm, DMPLEX);
DMSetFromOptions(dm);
```
This will simplify our code, but I figured we'd wait for 3.16 to be released before shifting over to this pattern.

jedbrown: I'd just copied the 3,3,3 default from solids, but it was probably 1,1,1 in fluids (not that that's likely to be very helpful).

knepley: The `faces` array is asking how many faces you want on each side of the box. So `[2, 3, 5]` would give 30 hex cells in the box.

LeilaGhaffari: Oh, then I was confusing two separate things! I thought that was determining number of box faces which was 6 for 3D. I completely forgot for a moment that the mesh is a _box_ and the number of box faces is fixed! :laughing:

jedbrown: One reason to have 3,3,3 as a default is that even with full Dirichlet boundary conditions and basis degree 1, _something_ can happen.

LeilaGhaffari: Yes, `3,3,3` makes more sense @jedbrown. 

Thanks, @jedbrown and @knepley for the clarification!

jedbrown:
This makes testing reliable and fixes #802 (circumvents what may be a cuda-11.2+ toolchain or driver issue). The lv (cuda) tests are slow because they're now serialized, but that's better than routinely failing. I think this is ready for merging.

PR #804, opened 2021-09-03
make CUDA block sizes fit according to number of used registers

jedbrown:


backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -41,7 +41,7 @@ static int CeedQFunctionApply_Cuda(CeedQFunction qf, CeedInt Q,
   CeedInt numinputfields, numoutputfields;
   ierr = CeedQFunctionGetNumArgs(qf, &numinputfields, &numoutputfields);
   CeedChkBackend(ierr);
-  const int blocksize = ceed_Cuda->optblocksize;
+  const int blocksize = ceed_Cuda->optblocksize / 2;
```
jedbrown, line 44: Dirty hack to make this work
```console
$ nsys profile --trace=cuda build/solids-elasticity -E 10 -nu .3 -degree 2 -mesh ~/meshes/holes.msh -problem FSInitial-NH1 -num_steps 3 -bc_clamp 1 -bc_traction 2 -bc_traction_2 0,1,0 -view_soln -snes_monitor -ceed /gpu/cuda/shared
```

The default is to use the max number of threads per block (1024), but this requires that register count stays quite low. Real qfunctions can get bigger and we need to choose correspondingly smaller block sizes.

backends/cuda/ceed-cuda.c
```diff
@@ -71,6 +71,8 @@ int CeedCompileCuda(Ceed ceed, const char *source, CUmodule *module,
   char buff[optslen];
   snprintf(buff, optslen,"-arch=compute_%d%d", prop.major, prop.minor);
   opts[numopts + 3] = buff;
+  opts[numopts + 4] = "--maxrregcount=63";
```
jedbrown, line 74: This doesn't seem to actually be used. Analyzing the output says the kernel used 126 registers. I haven't checked yet if it's coincidence that this is twice the requested number.

backends/cuda/ceed-cuda.c
```diff
@@ -71,6 +71,8 @@ int CeedCompileCuda(Ceed ceed, const char *source, CUmodule *module,
...
+  opts[numopts + 5] = "--optimization-info=inline";
```
jedbrown, line 75: This produces confusingly contradictory output
```
remark: default_program:673:29: _Z20log1p_series_shiftedd will not be inlined into CeedKernel_Cuda_ref_ElasFSInitialNH1Diagnostic

remark: default_program:673:29: _Z20log1p_series_shiftedd inlined into CeedKernel_Cuda_ref_ElasFSInitialNH1Diagnostic with cost=-19925 (threshold=1625)
```

backends/cuda/ceed-cuda.c
```diff
@@ -110,8 +121,17 @@ int CeedGetKernelCuda(Ceed ceed, CUmodule module, const char *name,
 //------------------------------------------------------------------------------
 int CeedRunKernelCuda(Ceed ceed, CUfunction kernel, const int gridSize,
                       const int blockSize, void **args) {
-  CeedChk_Cu(ceed, cuLaunchKernel(kernel, gridSize, 1, 1, blockSize, 1,
-                                  1, 0, NULL, args, NULL));
+  CUresult result = cuLaunchKernel(kernel, gridSize, 1, 1, blockSize, 1,
+                                   1, 0, NULL, args, NULL);
+  if (result == CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES) {
+    int max_threads_per_block, shared_size_bytes, num_regs;
+    cuFuncGetAttribute(&max_threads_per_block, CU_FUNC_ATTRIBUTE_MAX_THREADS_PER_BLOCK, kernel);
```
jedbrown, line 128: Seems we should just set the number of threads per block according to this, rather than the max system reported.

jedbrown (commented):
I can pick this up when I have time, either in the afternoon or weekend/next week. It's important to be able to handle complicated qfunctions. I assume something similar may be needed for hiprtc, though this example has been running in noether CI so it may be fine.

backends/cuda/ceed-cuda.c
nbeams: Interestingly, I commented out the inline option due to the error I mentioned below, and I'm seeing 104 registers/thread. This is also with an older CUDA, so perhaps some changes to nvrtc... The QFunction kernel is the only one above 63, though I tried the same test without the maxrregcount flag, and the QFunction kernel was still the only one above 63.  So it does indeed look like it's not being respected.

I wonder if somehow nvrtc gets confused about usage of the device functions inside the QFunction kernel?  The QFunction inputs and outputs at a quadrature point are read into register arrays through device functions prior to calling the user function. Does nvrtc count this use together with whatever is required within the main user-defined device function?

nbeams: Also, is there a specific reason you chose 63?

backends/cuda/ceed-cuda.c
jedbrown: This was debugging and diagnostics, not a proposal to merge. I just wanted to share what I'd tried. We don't want to dump diagnostics in general, though perhaps when the user requests (say, via an environment variable).

backends/cuda/ceed-cuda.c
jedbrown: It was arbitrary, just seeing if it would be used.

YohannDudouit: It's used as a recommendation.

YohannDudouit: I wouldn't recommend using this option, it sometimes result in worst performance.

backends/cuda/ceed-cuda.c
YohannDudouit: Quite often the number of threads is specific to the number of quads, but this would be a good default, for instance to a function where we don't provide the blocksize.

backends/cuda/ceed-cuda.c
jedbrown: I was thinking we should cap it at *some* size to ensure that blocks could be scheduled at some block size. It doesn't matter here (with shared) because the block sizes can be as small as we choose, but cuda-gen needs specific block sizes (which might need to be bigger than 32).

Also, I see that the cuda-shared `grad` kernel (for BP3, degree 3, Q=5 quadrature points per dimension) might have a block size of `5,5,1` while the corresponding cuda-gen has a block size of `5,5,4`. Probably both would be better off as `5,5,5` (size 125 is close to 128) unless I'm missing something. Anyway, my thought was to be able to find a small viable block size (`5,5,5`) and insist on using few enough registers for it. But this option doesn't seem to be processed as of cuda-11.4 so there's no point pursuing. (We should report it though.)

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -32,6 +32,81 @@ static int CeedOperatorDestroy_Cuda_gen(CeedOperator op) {
...
+// are effectively committed to that block. Now suppose
+// cuOccupancyMaxPotentialBlockSize returned 352. We can schedule 2 blocks of
+// size 98 (196 useful threads using 256 hardware threads), but not a third
+// block (which would need a total of 384 hardware threads).
+//
+// If instead, we had packed 3 elements, we'd have 2*49=147 useful threads
+// occupying 160 slots, and could schedule two blocks. Alterntively, we could
+// pack a single block of 7 elements (2*49=343 useful threads) into the 354
+// slots. The latter has the least "waste", but __syncthreads()
+// over-synchronizes and it might not pay off relative to smaller blocks.
```
jedbrown, line 82: @YohannDudouit @nbeams Does this description of the performance tradeoffs for cuda-gen sound right to you? I did some profiling and viewing locally it it's how I would explain it.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -32,6 +32,81 @@ static int CeedOperatorDestroy_Cuda_gen(CeedOperator op) {
...
+    // We want to minimize waste, but smaller kernels have lower latency and
+    // less __syncthreads() overhead so when a larger block size has the same
+    // waste as a smaller one, go ahead and prefer the smaller block.
+    if (i_waste < waste || (i_waste == waste && threads_per_elem * i <= 128)) {
+      elems_per_block = i;
+      waste = i_waste;
+    }
+  }
+  block[2] = elems_per_block;
+  *grid = (nelem + elems_per_block - 1) / elems_per_block;
```
jedbrown, line 102: This is kind of heuristic and we might want to be more aggressive (beyond breaking ties) in creating smaller blocks so that `__syncthreads()` is less costly.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -32,6 +32,81 @@ static int CeedOperatorDestroy_Cuda_gen(CeedOperator op) {
...
+// large thread blocks. Suppose I have elements with 7x7x7 quadrature points.
+// This will loop over the last dimension, so we have 7*7=49 threads per
+// element. Suppose we have two elements = 2*49=98 useful threads. CUDA
+// schedules in units of full warps (32 threads), so 128 CUDA hardware threads
+// are effectively committed to that block. Now suppose
+// cuOccupancyMaxPotentialBlockSize returned 352. We can schedule 2 blocks of
+// size 98 (196 useful threads using 256 hardware threads), but not a third
+// block (which would need a total of 384 hardware threads).
+//
+// If instead, we had packed 3 elements, we'd have 2*49=147 useful threads
```
nbeams, line 78: Typo: 3*49

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -32,6 +32,81 @@ static int CeedOperatorDestroy_Cuda_gen(CeedOperator op) {
...
+// occupying 160 slots, and could schedule two blocks. Alterntively, we could
```
nbeams, line 79: Typo: Alternatively

backends/cuda-gen/ceed-cuda-gen-operator.c
nbeams: Makes sense to me.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -122,28 +199,19 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector invec,
...
-    CeedInt grid = nelem/elemsPerBlock + ( (nelem/elemsPerBlock*elemsPerBlock<nelem)
-                                           ? 1 : 0 );
-    CeedInt sharedMem = elemsPerBlock*thread1d*thread1d*sizeof(CeedScalar);
-    ierr = CeedRunKernelDimSharedCuda(ceed, data->op, grid, thread1d, thread1d,
-                                      elemsPerBlock, sharedMem, opargs);
-  }
+  int max_threads_per_block, min_grid_size;
+  CeedChk_Cu(ceed, cuFuncSetCacheConfig(data->op, CU_FUNC_CACHE_PREFER_L1));
+  CeedChk_Cu(ceed, cuFuncSetSharedMemConfig(data->op,
+             CU_SHARED_MEM_CONFIG_EIGHT_BYTE_BANK_SIZE));
```
nbeams, line 205: Have you seen a noticeable difference in performance when using this option?  What if we are using single precision?

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -122,28 +199,19 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector invec,
...
-  } else if (dim==3) {
-    const CeedInt elemsPerBlock = thread1d<6? 4 : (thread1d<8? 2 : 1);
-    CeedInt grid = nelem/elemsPerBlock + ( (nelem/elemsPerBlock*elemsPerBlock<nelem)
-                                           ? 1 : 0 );
-    CeedInt sharedMem = elemsPerBlock*thread1d*thread1d*sizeof(CeedScalar);
-    ierr = CeedRunKernelDimSharedCuda(ceed, data->op, grid, thread1d, thread1d,
-                                      elemsPerBlock, sharedMem, opargs);
-  }
+  int max_threads_per_block, min_grid_size;
+  CeedChk_Cu(ceed, cuFuncSetCacheConfig(data->op, CU_FUNC_CACHE_PREFER_L1));
```
nbeams, line 203: If we want to use this, should be consider setting it at higher level than just for this kernel?  Per the CUDA documentation, "[Launching a kernel with a different preference than the most recent preference setting may insert a device-side synchronization point.](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXEC.html#group__CUDA__EXEC_1g40f8c11e81def95dc0072a375f965681)"  I don't have any experience indicating how likely this is to happen, though.

nbeams (approved):
LGTM.  I like the `CeedRunKernelAutoblockCuda` function.

backends/cuda-gen/ceed-cuda-gen-operator.c
jedbrown: Thanks, fixed.

backends/cuda-gen/ceed-cuda-gen-operator.c
jedbrown: I honestly don't know and I'm having trouble measuring a consistent difference so I'll remove it. This just seemed like it would be analytically beneficial given the relative size of the needed shared memory, but defaults are better than magic that may or may not make a difference.

backends/cuda-gen/ceed-cuda-gen-operator.c
YohannDudouit: It's great, I think it describes things well.

Issue #805, opened 2021-09-05
GPU: persist kernels that will be reused in nonlinear and transient problems.


jedbrown:
Notably, grid transfer operators and diagonal assembly. Running (for example) the solid mechanics example with a breakpoint in `CeedCompileCuda` and collecting stack traces is informative. For example, this occurs on every level and every Newton step.

```
#0  CeedCompileCuda (ceed=0x555555ca17c0, source=source@entry=0x7ffff5597fa0 "extern \"C\" __global__ void noTrStrided(const CeedInt nelem, const CeedScalar *__restrict__ u,
 CeedScalar *__restrict__ v) { for (CeedInt node = blockIdx.x * blockDim.x + threadIdx.x; node < nelem*REST"..., module=0x55559c14bd50, numopts=numopts@entry=8) at /home/jed
/src/libceed/backends/cuda/ceed-cuda.c:93
#1  0x00007ffff54e5eb8 in CeedElemRestrictionCreate_Cuda (mtype=CEED_MEM_HOST, cmode=<optimized out>, indices=0x0, r=0x55559c562aa0) at /home/jed/src/libceed/backends/cuda/c
eed-cuda-restriction.c:440
#2  0x00007ffff54be790 in CeedElemRestrictionCreateStrided (ceed=0x555555ca17c0, num_elem=2880, elem_size=27, num_comp=num_comp@entry=81, l_size=6298560, strides=strides@ent
ry=0x7fffffffb75c, rstr=0x7fffffffb848) at /home/jed/src/libceed/interface/ceed-elemrestriction.c:435
#3  0x00007ffff54e2ec1 in CeedOperatorLinearAssembleQFunction_Cuda (op=<optimized out>, assembled=0x7fffffffb840, rstr=0x7fffffffb848, request=<optimized out>) at /home/jed/
src/libceed/backends/cuda/ceed-cuda-operator.c:615
#4  0x00007ffff54c5118 in CeedOperatorLinearAssembleQFunction (op=op@entry=0x55559c5283f0, assembled=assembled@entry=0x7fffffffb840, rstr=rstr@entry=0x7fffffffb848, request=
request@entry=0x7ffff5755608 <ceed_request_immediate>) at /home/jed/src/libceed/interface/ceed-preconditioning.c:1060
[...]

Issue #792, opened 2021-07-15
Stable release branch?


jedbrown:
Some of you may have noticed that there is a `release` branch in the repository now. I'm curious if we should start distinguishing main development from ABI-compatible bug fixes, perhaps as a warm-up for practicing more strict SemVer when we reach 1.0. It's another thing to pay attention to when fixing bugs and creating/reviewing PRs, but not bad once you get the hang of it. I think we should eventually do this, but I'm not sure it's worth doing yet. A related discussion is what should be present to release 1.0. So a poll, and solicitation of comments:

:+1: Keep `release` branch, potentially tag minor releases like `v0.9.1`. Update contributing guidelines to recommend starting bug-fix branches from `release` and making pull requests to `release`. We'll merge forward (`release` into `main`) routinely as part of maintainer workflow. New feature branches should still start from `main` and the next feature release will be `v0.10.0` (presumably) tagged on `main`.

:-1: Let's wait (for any number of reasons). It's added complexity and not enough value to justify. Let's keep `main` moving for now and perhaps revisit after future feature releases.

ArashMehraban:
I periodically read the source code to see if any significant changes have occurred that might be problematic in case older versions of PETSc is used for example (3.13, 3.14). Of course, I mostly check out the solids code for testing. In addition, I have found numerous cases that the provisions we have put in that code (solid) do not resolve the issues. That is on top of the fact that I feel like the solids code, and perhaps fluids code should be restructured to have a more flexible design for the add-ons I have in mind. To provide a few examples from the solids code:
1. The elasticity.c (or main in general) code is too cluttered up and many things from there must be moved somewhere else. For example, the preconditioning is almost in the middle of that code in is entirety. 
2. ProcessUserOptions computes things that are related to the natural BC. That function has no business computing things!
3. I’d like to have more flexibility with BC and that needs to become its own module, i,e. its dependencies will matter
4. I want an api in ‘main’ to control PETSc solvers, etc.
5. I want multi-materials

So, I personally, prefer to be able to work with a stable ‘release’ and get something working and call that example the one that works with version x.y.z.. So thumbs up 👍.

jedbrown:
[...]

PR #806, opened 2021-09-08
Promote Field Getters to Public API

jeremylt:
I also want to add this to the Rust interface to make my mini-app development in Rust easier.

### DIFF
--- a/doc/sphinx/source/releasenotes.md
+++ b/doc/sphinx/source/releasenotes.md
@@ -13,6 +13,10 @@ for each release of libCEED.
 - QFunction and Operator field objects, `CeedQFunctionField` and `CeedOperatorField`, and associated getters, {c:func}`CeedQFunctionGetFields`; {c:func}`CeedQFunctionFieldGetName`; {c:func}`CeedQFunctionFieldGetSize`; {c:func}`CeedQFunctionFieldGetEvalMode`; {c:func}`CeedOperatorGetFields`; {c:func}`CeedOperatorFieldGetElemRestriction`; {c:func}`CeedOperatorFieldGetBasis`; and {c:func}`CeedOperatorFieldGetVector`, promoted to the public API.
 - Clarify and document conditions where `CeedQFunction` and `CeedOperator` become immutable and no further fields or suboperators can be added.
 
+### New features
+
+- `CeedScalar` can now be set as `float` or `double` at compile time.
+
 ### Maintainability
 
 - Refactored preconditioner support internally to facilitate future development and improve GPU completeness/test coverage.
--- a/doc/sphinx/source/releasenotes.md
+++ b/doc/sphinx/source/releasenotes.md
@@ -7,6 +7,10 @@ for each release of libCEED.
 
 ## Current `main` branch
 
+### Interface changes
+
+- Update {c:func} `CeedQFunctionGetFields` and {c:func} `CeedOperatorGetFields` to include number of fields.
+
 ### Maintainability
 
 - Refactored preconditioner support internally to facilitate future development and improve GPU completeness/test coverage.
--- a/doc/sphinx/source/releasenotes.md
+++ b/doc/sphinx/source/releasenotes.md
@@ -10,7 +10,7 @@ for each release of libCEED.
 ### Interface changes
 
 - Update {c:func} `CeedQFunctionGetFields` and {c:func} `CeedOperatorGetFields` to include number of fields.
-- QFunction and Operator field objects, `CeedQFunctionField` and `CeedOperatorField`, and associated getters, {c:func}`CeedQFunctionGetFields`; {c:func}`CeedQFunctionFieldGetName`; {c:func}`CeedQFunctionFieldGetSize`; {c:func}`CeedQFunctionFieldGetEvalMode`; {c:func}`CeedOperatorGetFields`; {c:func}`CeedOperatorFieldGetElemRestriction`; {c:func}`CeedOperatorFieldGetBasis`; and {c:func}`CeedOperatorFieldGetVector`, promoted to the public API.
+- Promote to the public API: QFunction and Operator field objects, `CeedQFunctionField` and `CeedOperatorField`, and associated getters, {c:func}`CeedQFunctionGetFields`; {c:func}`CeedQFunctionFieldGetName`; {c:func}`CeedQFunctionFieldGetSize`; {c:func}`CeedQFunctionFieldGetEvalMode`; {c:func}`CeedOperatorGetFields`; {c:func}`CeedOperatorFieldGetElemRestriction`; {c:func}`CeedOperatorFieldGetBasis`; and {c:func}`CeedOperatorFieldGetVector`.
 - Clarify and document conditions where `CeedQFunction` and `CeedOperator` become immutable and no further fields or suboperators can be added.
 
 ### New features
--- a/doc/sphinx/source/releasenotes.md
+++ b/doc/sphinx/source/releasenotes.md
@@ -11,6 +11,7 @@ for each release of libCEED.
 
 - Update {c:func} `CeedQFunctionGetFields` and {c:func} `CeedOperatorGetFields` to include number of fields.
 - QFunction and Operator field objects, `CeedQFunctionField` and `CeedOperatorField`, and associated getters, {c:func}`CeedQFunctionGetFields`; {c:func}`CeedQFunctionFieldGetName`; {c:func}`CeedQFunctionFieldGetSize`; {c:func}`CeedQFunctionFieldGetEvalMode`; {c:func}`CeedOperatorGetFields`; {c:func}`CeedOperatorFieldGetElemRestriction`; {c:func}`CeedOperatorFieldGetBasis`; and {c:func}`CeedOperatorFieldGetVector`, promoted to the public API.
+- Clarify and document conditions where `CeedQFunction` and `CeedOperator` become immutable and no further fields or suboperators can be added.
 
 ### Maintainability
 
--- a/doc/sphinx/source/releasenotes.md
+++ b/doc/sphinx/source/releasenotes.md
@@ -10,6 +10,7 @@ for each release of libCEED.
 ### Interface changes
 
 - Update {c:func} `CeedQFunctionGetFields` and {c:func} `CeedOperatorGetFields` to include number of fields.
+- QFunction and Operator field objects, `CeedQFunctionField` and `CeedOperatorField`, and associated getters, {c:func}`CeedQFunctionGetFields`; {c:func}`CeedQFunctionFieldGetName`; {c:func}`CeedQFunctionFieldGetSize`; {c:func}`CeedQFunctionFieldGetEvalMode`; {c:func}`CeedOperatorGetFields`; {c:func}`CeedOperatorFieldGetElemRestriction`; {c:func}`CeedOperatorFieldGetBasis`; and {c:func}`CeedOperatorFieldGetVector`, promoted to the public API.
 
 ### Maintainability
 
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -146,11 +146,11 @@ int CeedOperatorCheckReady(CeedOperator op) {
   Ceed ceed;
   ierr = CeedOperatorGetCeed(op, &ceed); CeedChk(ierr);
 
-  if (op->interface_setup)
+  if (op->is_interface_setup)
     return CEED_ERROR_SUCCESS;
 
   CeedQFunction qf = op->qf;
-  if (op->composite) {
+  if (op->is_composite) {
     if (!op->num_suboperators)
       // LCOV_EXCL_START
       return CeedError(ceed, CEED_ERROR_INCOMPLETE, "No sub_operators set");
@@ -178,18 +178,18 @@ int CeedOperatorCheckReady(CeedOperator op) {
   }
 
   // Flag as immutable and ready
-  op->interface_setup = true;
+  op->is_interface_setup = true;
   if (op->qf && op->qf != CEED_QFUNCTION_NONE)
     // LCOV_EXCL_START
-    op->qf->operators_set++;
+    op->qf->is_immutable = true;
   // LCOV_EXCL_STOP
   if (op->dqf && op->dqf != CEED_QFUNCTION_NONE)
     // LCOV_EXCL_START
-    op->dqf->operators_set++;
+    op->dqf->is_immutable = true;
   // LCOV_EXCL_STOP
   if (op->dqfT && op->dqfT != CEED_QFUNCTION_NONE)
     // LCOV_EXCL_START
-    op->dqfT->operators_set++;
+    op->dqfT->is_immutable = true;
   // LCOV_EXCL_STOP
   return CEED_ERROR_SUCCESS;
 }
@@ -363,7 +363,7 @@ int CeedOperatorGetCeed(CeedOperator op, Ceed *ceed) {
 **/
 
 int CeedOperatorGetNumElements(CeedOperator op, CeedInt *num_elem) {
-  if (op->composite)
+  if (op->is_composite)
     // LCOV_EXCL_START
     return CeedError(op->ceed, CEED_ERROR_MINOR,
                      "Not defined for composite operator");
@@ -385,7 +385,7 @@ int CeedOperatorGetNumElements(CeedOperator op, CeedInt *num_elem) {
 **/
 
 int CeedOperatorGetNumQuadraturePoints(CeedOperator op, CeedInt *num_qpts) {
-  if (op->composite)
+  if (op->is_composite)
     // LCOV_EXCL_START
     return CeedError(op->ceed, CEED_ERROR_MINOR,
                      "Not defined for composite operator");
@@ -407,7 +407,7 @@ int CeedOperatorGetNumQuadraturePoints(CeedOperator op, CeedInt *num_qpts) {
 **/
 
 int CeedOperatorGetNumArgs(CeedOperator op, CeedInt *num_args) {
-  if (op->composite)
+  if (op->is_composite)
     // LCOV_EXCL_START
     return CeedError(op->ceed, CEED_ERROR_MINOR,
                      "Not defined for composite operators");
@@ -429,7 +429,7 @@ int CeedOperatorGetNumArgs(CeedOperator op, CeedInt *num_args) {
 **/
 
 int CeedOperatorIsSetupDone(CeedOperator op, bool *is_setup_done) {
-  *is_setup_done = op->backend_setup;
+  *is_setup_done = op->is_backend_setup;
   return CEED_ERROR_SUCCESS;
 }
 
@@ -445,7 +445,7 @@ int CeedOperatorIsSetupDone(CeedOperator op, bool *is_setup_done) {
 **/
 
 int CeedOperatorGetQFunction(CeedOperator op, CeedQFunction *qf) {
-  if (op->composite)
+  if (op->is_composite)
     // LCOV_EXCL_START
     return CeedError(op->ceed, CEED_ERROR_MINOR,
                      "Not defined for composite operator");
@@ -467,7 +467,7 @@ int CeedOperatorGetQFunction(CeedOperator op, CeedQFunction *qf) {
 **/
 
 int CeedOperatorIsComposite(CeedOperator op, bool *is_composite) {
-  *is_composite = op->composite;
+  *is_composite = op->is_composite;
   return CEED_ERROR_SUCCESS;
 }
 
@@ -483,7 +483,7 @@ int CeedOperatorIsComposite(CeedOperator op, bool *is_composite) {
 **/
 
 int CeedOperatorGetNumSub(CeedOperator op, CeedInt *num_suboperators) {
-  if (!op->composite)
+  if (!op->is_composite)
     // LCOV_EXCL_START
     return CeedError(op->ceed, CEED_ERROR_MINOR, "Not a composite operator");
   // LCOV_EXCL_STOP
@@ -504,7 +504,7 @@ int CeedOperatorGetNumSub(CeedOperator op, CeedInt *num_suboperators) {
 **/
 
 int CeedOperatorGetSubList(CeedOperator op, CeedOperator **sub_operators) {
-  if (!op->composite)
+  if (!op->is_composite)
     // LCOV_EXCL_START
     return CeedError(op->ceed, CEED_ERROR_MINOR, "Not a composite operator");
   // LCOV_EXCL_STOP
@@ -570,7 +570,7 @@ int CeedOperatorReference(CeedOperator op) {
 **/
 
 int CeedOperatorSetSetupDone(CeedOperator op) {
-  op->backend_setup = true;
+  op->is_backend_setup = true;
   return CEED_ERROR_SUCCESS;
 }
 
@@ -670,7 +670,7 @@ int CeedCompositeOperatorCreate(Ceed ceed, CeedOperator *op) {
   ierr = CeedCalloc(1, op); CeedChk(ierr);
   (*op)->ceed = ceed;
   ierr = CeedReference(ceed); CeedChk(ierr);
-  (*op)->composite = true;
+  (*op)->is_composite = true;
   ierr = CeedCalloc(16, &(*op)->sub_operators); CeedChk(ierr);
 
   if (ceed->CompositeOperatorCreate) {
@@ -731,10 +731,15 @@ int CeedOperatorReferenceCopy(CeedOperator op, CeedOperator *op_copy) {
 int CeedOperatorSetField(CeedOperator op, const char *field_name,
                          CeedElemRestriction r, CeedBasis b, CeedVector v) {
   int ierr;
-  if (op->composite)
+  if (op->is_composite)
     // LCOV_EXCL_START
     return CeedError(op->ceed, CEED_ERROR_INCOMPATIBLE,
                      "Cannot add field to composite operator.");
+  // LCOV_EXCL_STOP
+  if (op->is_immutable)
+    // LCOV_EXCL_START
+    return CeedError(op->ceed, CEED_ERROR_MAJOR,
+                     "Operator cannot be changed after set as immutable");
   // LCOV_EXCL_STOP
   if (!r)
     // LCOV_EXCL_START
@@ -833,6 +838,9 @@ found:
 /**
   @brief Get the CeedOperatorFields of a CeedOperator
 
+  Note: Calling this function asserts that setup is complete
+          and sets the CeedOperator as immutable.
+
   @param op                  CeedOperator
   @param[out] input_fields   Variable to store input_fields
   @param[out] output_fields  Variable to store output_fields
@@ -845,11 +853,14 @@ int CeedOperatorGetFields(CeedOperator op, CeedInt *num_input_fields,
                           CeedOperatorField **input_fields,
                           CeedInt *num_output_fields,
                           CeedOperatorField **output_fields) {
-  if (op->composite)
+  int ierr;
+
+  if (op->is_composite)
     // LCOV_EXCL_START
     return CeedError(op->ceed, CEED_ERROR_MINOR,
                      "Not defined for composite operator");
   // LCOV_EXCL_STOP
+  ierr = CeedOperatorCheckReady(op); CeedChk(ierr);
 
   if (num_input_fields) *num_input_fields = op->qf->num_input_fields;
   if (input_fields) *input_fields = op->input_fields;
@@ -933,7 +944,7 @@ int CeedCompositeOperatorAddSub(CeedOperator composite_op,
                                 CeedOperator sub_op) {
   int ierr;
 
-  if (!composite_op->composite)
+  if (!composite_op->is_composite)
     // LCOV_EXCL_START
     return CeedError(composite_op->ceed, CEED_ERROR_MINOR,
                      "CeedOperator is not a composite operator");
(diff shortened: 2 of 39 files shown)


doc/sphinx/source/releasenotes.md
```diff
@@ -7,6 +7,14 @@ for each release of libCEED.
 
 ## Current `main` branch
 
+### Interface changes
+
+- QFunction and Operator field objects, `CeedQFunctionField` and `CeedOperatorField`, and associated getters, {c:func}`CeedQFunctionGetFields`; {c:func}`CeedQFunctionFieldGetName`; {c:func}`CeedQFunctionFieldGetSize`; {c:func}`CeedQFunctionFieldGetEvalMode`; {c:func}`CeedOperatorGetFields`; {c:func}`CeedOperatorFieldGetElemRestriction`; {c:func}`CeedOperatorFieldGetBasis`; and {c:func}`CeedOperatorFieldGetVector`, promoted to the public API.
+
+### New features
+
+- `CeedScalar` can now be set as `float` or `double` at compile time.
```
jeremylt, line 16: Not sure what more to say. That was a big PR, but this seems to cover the key point.

rust/libceed/src/operator.rs
```diff
@@ -20,11 +20,218 @@
...
+            let mut ptr_copy = std::ptr::null_mut();
+            unsafe {
+                bind_ceed::CeedVectorReferenceCopy(ptr, &mut ptr_copy);
+            }
+            Some(Vector::from_raw(self.ceed, ptr_copy).unwrap())
+        } else {
+            None
+        }
+    }
+}
```
jeremylt, line 228: It works, but I'm not really a fan of this because it does not distinguish between `CEED_VECTOR_NONE` and `CEED_VECTOR_ACTIVE`. Currently, we have the options enum `VectorOpt`, but that's technically more of a `VectorRefOpt`. I think we will want to have both an option enum for vectors and vector refs as well as a way to interop between the two.

jeremylt: Ok, I have updated this with a better approach. If we like it, I can squash the two relevant commits together.

doc/sphinx/source/releasenotes.md
jedbrown: I think this is fine at this time -- we have more ambitious plans longer term.

interface/ceed-operator.c
```diff
@@ -904,6 +830,95 @@ int CeedOperatorSetField(CeedOperator op, const char *field_name,
...
+
+  @param op                  CeedOperator
+  @param[out] input_fields   Variable to store input_fields
+  @param[out] output_fields  Variable to store output_fields
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+
```
jedbrown, line 844:
```suggestion
```

interface/ceed-operator.c
```diff
@@ -904,6 +830,95 @@ int CeedOperatorSetField(CeedOperator op, const char *field_name,
...
+int CeedOperatorGetFields(CeedOperator op, CeedOperatorField **input_fields,
+                          CeedOperatorField **output_fields) {
```
jedbrown, line 846: Should we return the number of fields?

interface/ceed-qfunction.c
```diff
@@ -681,6 +617,70 @@ int CeedQFunctionAddOutput(CeedQFunction qf, const char *field_name,
...
+/**
+  @brief Get the CeedQFunctionFields of a CeedQFunction
+
+  @param qf                  CeedQFunction
+  @param[out] input_fields   Variable to store input_fields
+  @param[out] output_fields  Variable to store output_fields
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
```
jedbrown, line 629: Perhaps Advanced or something suggesting that these are useful for higher level abstractions/libraries, but you probably don't need it when writing straight-line application code.

interface/ceed-qfunction.c
```diff
@@ -681,6 +617,70 @@ int CeedQFunctionAddOutput(CeedQFunction qf, const char *field_name,
...
+**/
+int CeedQFunctionGetFields(CeedQFunction qf, CeedQFunctionField **input_fields,
+                           CeedQFunctionField **output_fields) {
```
jedbrown, line 632: Number of fields?

rust/libceed/src/qfunction.rs
```diff
@@ -191,6 +278,71 @@ impl<'a> QFunctionCore<'a> {
...
+        let mut inputs = Vec::new();
+        for i in 0..num_inputs as usize {
+            inputs.push(QFunctionField {
+                ptr: inputs_slice[i] as bind_ceed::CeedQFunctionField,
+            })
+        }
+        Ok(inputs)
+    }
+
+    pub fn outputs(&self) -> crate::Result<Vec<crate::QFunctionField>> {
```
jedbrown, line 314: I'm a little uncomfortable with this returning *owned* `QFunctionField` (modifications of which would be reflected as mutation on the C side). Wouldn't it be more natural to an immutable slice?

jedbrown (commented):
Thanks for taking the lead on this.

interface/ceed-qfunction.c
jeremylt: We made those two separate functions originally for... reasons? I agree that extra outputs make sense here.

rust/libceed/src/operator.rs
```diff
@@ -20,18 +20,292 @@
...
+    /// ```
+    pub fn name(&self) -> String {
+        let mut name_ptr: *mut std::os::raw::c_char = std::ptr::null_mut();
+        unsafe {
+            bind_ceed::CeedOperatorFieldGetName(self.ptr, &mut name_ptr);
+        }
+        let c_str: &CStr = unsafe { CStr::from_ptr(name_ptr) };
+        let str_slice: &str = c_str.to_str().unwrap().clone();
+        str_slice.to_owned()
+    }
```
jedbrown, line 87: Should we return a `&str`?

jeremylt: I'm not sure why we would? `String` seems to be more common in Rust code than `&str`

jeremylt:
The Julia CI issue seems to be related the Release/Dev test split.

rust/libceed/src/operator.rs
jedbrown: Standard practice [C-CALLER-CONTROL](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control) (somewhat different context) says one should not clone unnecessarily inside library interfaces. I think the caller would typically match on the return, use it as a key, or interpolate it into a new string (for end-user output or to create a new related string). They can always call `to_owned()` if they want, but I think it makes more sense for us to return `&str`.

rust/libceed/src/qfunction.rs
jedbrown: I think if these interfaces can be turned into the following, we'll be good with this PR. Unless you have a rationale for for these being owned `Vec`.

```suggestion
    pub fn outputs(&self) -> crate::Result<&[crate::QFunctionField]> {
```

jeremylt: I'm currently digging through this. The tricky thing is that the slice has to be owned *somewhere*. I'm looking at the interior mutability pattern in Rust, but it is tricky to make a clean return of a slice instead of a ref. It might be easier to return a `QFunctionFieldView` that only has immutable deref.

jeremylt: Current working RefCell version
```
    pub fn inputs(&self) -> crate::Result<&[crate::QFunctionField]> {
        // Get array of raw C pointers for inputs
        let mut num_inputs = 0;
        let mut inputs_ptr = std::ptr::null_mut();
        let ierr = unsafe {
            bind_ceed::CeedQFunctionGetFields(
                self.ptr,
                &mut num_inputs,
                &mut inputs_ptr,
                std::ptr::null_mut() as *mut bind_ceed::CeedInt,
                std::ptr::null_mut() as *mut *mut bind_ceed::CeedQFunctionField,
            )
        };
        self.ceed.check_error(ierr)?;
        // Convert raw C pointers to fixed length array
        let inputs_slice = unsafe { std::slice::from_raw_parts(inputs_ptr, num_inputs as usize) };
        let mut inputs = Vec::new();
        for i in 0..num_inputs as usize {
            inputs.push(QFunctionField {
                ptr: inputs_slice[i] as bind_ceed::CeedQFunctionField,
            })
        }
        self.inputs.replace(inputs);
        Ok(unsafe { self.inputs.try_borrow_unguarded().unwrap() })
    }
```

jedbrown: Since `QFunctionField` just wraps a single pointer, it seems you could just transmute it. Then you wouldn't need an owned `Vec` or to even traverse the array (trusting that the pointers are non-NULL) because everything is contained in the slice, and the backing memory for the slice lives as long as the `QFunctionCore<'a>`.

jeremylt: I think the reasoning about this gets easier if I restrict `*GetFields()` to only be valid if the Operator is fully set-up.

jedbrown: Good point, and I think that's a reasonable constraint. (`QFunctionGetFields` requires a complete `CeedQFunction` and similar for `CeedOperator`.)

jeremylt: It is easy to check if a `CeedOperator` is fully specified - we check if it has as many fields as the `CeedQFunction`

Currently, we don't have a way to check that all `CeedQFunction` fields have been added. Should our QFunction constructor specify the number of inputs and outputs in the C interface?

jedbrown: Could we make it so that doing anything non-creationary with a QFunction checks that it's complete (to the extent that's a thing) and makes it immutable? This would avoiding making the API more burdensome, but still prevent weird conditions.

jeremylt: Currently we make the QF immutable once the OP is successfully checked for 'completeness' prior to Apply, Assemble*, etc. We don't really have any way of knowing that a QFunction is complete on its own.

jedbrown: But if all you have is a QFunction, we'd still like to be able to ask it questions. Can we mark it immutable in the same way as would happen when an Operator checks it? (Bottom line is that we shouldn't need an Operator to put the QFunction in that state.) It could be `CeedQFunctionVerify()` that the user may optionally call and that we call for certain queries and when Operator uses it.

jeremylt: But how do we know all fields have been provided unless the user tells us how many fields should be provided? Unless we just assume that if the user calls GetFields or Apply then they are no longer allowed to make any modifications.

jedbrown: Right, we just document the semantic and the user will get an error if they try to add more fields later.

jeremylt: Ok, I've rearranged this PR a bit. I still need to play more with returning a slice, though what I wrote above should be perfectly fine now that the fields cannot be changed after we call *GetFields

There is still some Julia issue I need to figure out, but that's for tomorrow.

jeremylt: Nope, the code above won't work. Rust will let you take the immutable borrow in the return value and put it in a mutable variable and then change properties of the immutable borrow willy-nilly.

jeremylt: Ah, new plan

rust/libceed/src/operator.rs
```diff
@@ -20,18 +20,254 @@
...
+    ///     "Incorrect field ElemRestriction"
+    /// );
+    /// ```
+    pub fn elem_restriction(&self) -> ElemRestrictionOpt {
+        if self.elem_restriction.ptr == unsafe { bind_ceed::CEED_ELEMRESTRICTION_NONE } {
+            ElemRestrictionOpt::None
+        } else {
+            ElemRestrictionOpt::Some(&self.elem_restriction)
+        }
+    }
```
jeremylt, line 145: Since the purpose of this is to facilitate construction of one operator/qfunction from another, I reworked my approach. Now access is only provided to a *Opt, which contains an immutable ref, so users can only use these fields to construct fields for additional qfunction/operators. They have no direct mutable access to the underlying libCEED object. I think it is now fine to hand the user a vector that they can own because they cannot change the underlying data.

jeremylt: We can now, for instance
```rust
// ...
let inputs = op_1.inputs()?;

let op_2 = ceed
     .operator(&qf, QFunctionOpt::None, QFunctionOpt::None)?
     .field(inputs[1].name(), inputs[1].elem_restriction(), inputs[1].basis(), inputs[1].vector())?;
     // ...
```

interface/ceed-operator.c
```diff
@@ -904,6 +835,101 @@ int CeedOperatorSetField(CeedOperator op, const char *field_name,
...
+  @param[out] output_fields  Variable to store output_fields
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+int CeedOperatorGetFields(CeedOperator op, CeedInt *num_input_fields,
+                          CeedOperatorField **input_fields,
+                          CeedInt *num_output_fields,
+                          CeedOperatorField **output_fields) {
```
jeremylt, line 855: @pazner, relevant to both the Julia interface and the MFEM integration, this is an API change. The old function signature was
```C
int CeedOperatorGetFields(CeedOperator op,
                          CeedOperatorField **input_fields,
                          CeedOperatorField **output_fields)
```
A similar change was made for `CeedQFunctionGetFields()`.

jeremylt: You could get around it fine in MFEM with
```C
#if CEED_VERSION_GE(0, 9, 1)
#  define CeedOperatorGetFields(a,b,c) CeedOperatorGetFields(a,NULL,b,NULL,c)
#else
#  define CeedOperatorGetFields(a,b,c) CeedOperatorGetFields(a,b,c)
#endif

Comment (pazner) [reply interface/ceed-operator.c:855]: Does it make sense to bump the library version since this is a breaking change?
```c++
#define CEED_VERSION_MAJOR 0
#define CEED_VERSION_MINOR 10
#define CEED_VERSION_PATCH 0
#define CEED_VERSION_RELEASE false
```

pazner: > You could get around it fine in MFEM with

I could do something similar in Julia too, but maybe it's better to just adopt the new interface entirely (and require a more up-to-date version of libCEED)?

jeremylt: I think we're planning on a release this month, but we're a little busy here at CU the next couple of weeks though, so it would be later in the month.

pazner: Does it make sense to bump the version already in PRs? i.e. now that 0.9.0 is released, this PR is targeting 0.10.0?

pazner: In particular, the Julia CI is currently failing because it's attempting to use the new API with the most recent release of libCEED.

I could change the `minimum_libceed_version` in LibCEED.jl, which would then recognize that the version is incompatible and not run the tests.

jeremylt:
Fun new OCCA backend bug! The OCCA backend attempts to fully build and validate an Operator as part of its process for destroying an Operator, resulting in t508 failing. I'm investigating how best to fix this. There is no reason whatsoever to validate the operator fields as a prerequsite to destroying an Operator.

jeremylt:
Ok, the OCCA backend is pretty opaque to me and my C++ is not too great. I'll probably need help to fix this the right way. We could also just skip that test for OCCA and file another issue against the backend to be fixed sometime in the future.

interface/ceed-operator.c
jedbrown: `CEED_VERSION_GE` compares as +infinity when `CEED_VERSION_RELEASE` is `false`. The next release will be 0.10. I think we can have the same logic is the Julia code; cf. #792

pazner: That sound reasonable. To make sure I understand what you are suggesting, I can set `minimum_libceed_version` to 0.10.0, but `libceed_version >= minimum_libceed_version` will evaluate to `true` whenever `CEED_VERSION_RELEASE` is false (even if the version number is 0.9.0)?

rust/libceed/src/operator.rs
jeremylt: Note: I plan to squash this into the previous commit if we like this approach.

interface/ceed-operator.c
jeremylt: Thanks for those commits. With them, it looks like we're down to just the OCCA bug as a CI failure

julia/LibCEED.jl/src/LibCEED.jl
```diff
@@ -149,15 +149,14 @@ include("Request.jl")
 include("Operator.jl")
 include("Misc.jl")
 
-const minimum_libceed_version = v"0.8.0"
+const minimum_libceed_version = v"0.10.0"
```
pazner, line 152: @jeremylt, do you prefer if this is 0.10.0 or 0.9.1?

jeremylt: It looks like our next version number will be 0.10.0, so this should be good

jeremylt:
While we're here, any objections to raising these to the public API level?

```C
CEED_EXTERN int CeedOperatorGetNumElements(CeedOperator op, CeedInt *num_elem);
CEED_EXTERN int CeedOperatorGetNumQuadraturePoints(CeedOperator op,
    CeedInt *num_qpts);
```

maybe also this one?
```C
CEED_EXTERN int CeedOperatorGetQFunction(CeedOperator op, CeedQFunction *qf);
```

jedbrown:
> While we're here, any objections to raising these to the public API level?
> 
> ```c
> CEED_EXTERN int CeedOperatorGetNumElements(CeedOperator op, CeedInt *num_elem);
> CEED_EXTERN int CeedOperatorGetNumQuadraturePoints(CeedOperator op,
>     CeedInt *num_qpts);
> ```
> 
> maybe also this one?
> 
> ```c
> CEED_EXTERN int CeedOperatorGetQFunction(CeedOperator op, CeedQFunction *qf);
> ```

Just need to clarify what happens if you call those on a composite operator. Either it consults each of its sub-operators and aggregates the result or it returns error (or something out of band, like `-1`).

jeremylt:
Currently we return a minor error code for these from composite operators.

It might be helpful to expose IsComposite as well

rust/libceed/src/operator.rs
```diff
@@ -20,18 +20,226 @@
...
+// -----------------------------------------------------------------------------
+// CeedOperator Field context wrapper
+// -----------------------------------------------------------------------------
+#[derive(Debug)]
+pub struct OperatorField<'a> {
+    ceed: &'a crate::Ceed,
+    pub(crate) ptr: bind_ceed::CeedOperatorField,
+    elem_restriction: crate::ElemRestriction<'a>,
+    basis: crate::Basis<'a>,
+    vector: crate::Vector<'a>,
```
jedbrown, line 32: Is it necessary that all these fields get mirrored into the Rust object rather than referenced on-demand through the API?

Compare `QFunctionField`, which only contains the pointer and thus can readily be used in slices a la f3e3b2465013c0e26ea7151c50f20d31a15b3659

rust/libceed/src/qfunction.rs
```diff
@@ -24,10 +24,102 @@ use crate::prelude::*;
 pub type QFunctionInputs<'a> = [&'a [crate::Scalar]; MAX_QFUNCTION_FIELDS];
 pub type QFunctionOutputs<'a> = [&'a mut [crate::Scalar]; MAX_QFUNCTION_FIELDS];
 
+// -----------------------------------------------------------------------------
+// CeedQFunction Field context wrapper
+// -----------------------------------------------------------------------------
+#[derive(Debug)]
+pub struct QFunctionField {
+    ptr: bind_ceed::CeedQFunctionField,
```
jedbrown, line 32: Does this need a `PhantomData<'a>`?

rust/libceed/src/operator.rs
jeremylt: When I tried something similar with the operator fields, I was able to declare the returned slice as mut and start changing the underlying objects on the C side, which is why switched to only returning *Opts from the field, but an *Opt is only a ref and the actual Rest object needs to be owned somewhere

jedbrown: Do you have a strawman commit that does that? It sounds like something unsound that needs a fix elsewhere.

jeremylt: I can recreate on Monday. Really, I don't think that making the slice mut or not makes any difference. If I return a Rust object from the rest equivalent of OperatorFieldGet*, then the user will have the ability to capture the object in a mutable variable and change as they want. If I return an immutable ref or an opt holding the immutable ref, then the underlying rust object has to live somewhere.

Really, this comes down to if we want to allow users to get the vectors out of the fields of an object and change the contents of those vectors, as only the vectors present a way to change the underlying data after creation.

Side note: Are we 100% certain that transmute will be portable? The docs seem to imply otherwise?

jedbrown: As for the mutability leak you've found, you can't have a non-mut accessor return a mutable reference (or an owned object that behaves like a mutable reference, which is the case if it points into mutable C objects). Currently, `VectorOpt::Some` holds an immutable reference so the vector inside can't be mutated.

```rust
let mut inputs = op_mass.inputs()?;
if let libceed::prelude::VectorOpt::Some(v) = inputs[1].vector() {
    v.set_value(3.);
}
```
fails with
```
error[E0596]: cannot borrow `*v` as mutable, as it is behind a `&` reference
  --> src/vector.rs:96:5
   |
41 | if let libceed::prelude::VectorOpt::Some(v) = inputs[1].vector() {
   |                                          - help: consider changing this to be a mutable reference: `&mut Vector<'_>`
42 |     v.set_value(3.);
   |     ^ `v` is a `&` reference, so the data it refers to cannot be borrowed as mutable

error: aborting due to previous error
```

We may need a way for a user to mutate vectors in this context, but I think we want to localize it and use `RefCell` or otherwise checked exclusion. I don't think `op.inputs()` should return something that can mutate vectors.

Here's a revision without `transmute`, following ["This is the proper no-copy, unsafe way of "transmuting"](https://doc.rust-lang.org/std/mem/fn.transmute.html): 70141ce359b58a9ebdcec77190151f921783050e

jeremylt: Right, but that's with returning a VectorOpt. That's why I switched to a returning a VectorOpt. But we need to keep the Vector that the VectorOpt is referencing somewhere. Which is why I created the more complex OperatorField.

jeremylt: I just realized that currently we bypass the borrow checker and silently mutate the vectors for our passive fields. We should probably fix that.

jedbrown: We have to keep it somewhere *if* it's more than just the C pointer. I'm not sure I understand the motivation for the mirroring of related objects from C into Rust. Like why does `OperatorCore<'a>` contain a `&'a Ceed` instead of just a `PhantomData` to hold the lifetime?

Mutability in passive outputs is a common pattern that I don't know an idiomatic solution to, other than runtime checking similar to the read/write locking checking in vectors. With smart references, this could look like reference counting in which there can only ever be one mutable container, but an arbitrary number of immutable containers that are capable of holding it until needed. For example, a setup Operator could create an output that is fed into other Operators (solvers and diagnostics). The setup Operator might be kept around for multiple iterations, say while mesh geometry is optimized, then dropped with its output still used (immutably). Way out of scope for this PR, but the exclusivity of the mutable container could be checked at compile time. (This might exist in crate form.)

jeremylt: I don't understand how we would prevent mutation without hiding the ownership of the underlying C object from the user. If a field returns a C pointer to a vector or other object, the user can make that return value mut unless we borrow it from somewhere, unless I'm missing something.u

We carry a Rust Ceed in all objects to simplify error handling. Otherwise we would need to retrieve the C Ceed to check for errors after each bind_ceed call.

jedbrown: They can make what `mut`? We either return an object that derefs to `Vector` (could just be wrapping the C pointer) or a reference to `Vector`. It's true that if we return an (owned) `Vector` then the user can do anything with it, but we shouldn't do that.

I'm not sure the error-checking is that different from having a method to access it. Saving a few instructions in an error handling path doesn't seem important. A possible benefit would be possibly better resilience to memory corruption, in which calling into C again could give an erroneous `Ceed`. That seems like an edge case.

As for lifetimes, you can use `PhantomData` to match lifetimes without holding a direct reference. But I don't even think that matters because the C `Ceed` is reference counted so it doesn't matter if the Rust `Ceed` is dropped while objects live on.

jeremylt: If we return a reference, I don't understand where ownership lives. I'll just play around on Monday, because I'm more confused at this point and need to just tinker with code until I see what you're seeing.

jedbrown: Ownership would live on the C side with reference counting. We can return a reference or slice or struct that implements `Deref` (all of which have a lifetime) without incrementing the reference count of the C object because Rust checks that these ref-like things cannot outlive the parent object (which holds a pointer to the C object with in-built reference counting). I think if you can point to a strawman of what you think is problematic, I'll be able to help.

rust/libceed/src/operator.rs
```diff
@@ -529,10 +746,205 @@ impl<'a> Operator<'a> {
...
+            )
+        };
+        self.op_core.check_error(ierr)?;
+        // Convert raw C pointers to fixed length slice
+        let inputs_slice = unsafe {
+            std::slice::from_raw_parts(
+                inputs_ptr as *const crate::OperatorField,
+                num_inputs as usize,
+            )
+        };
```
jeremylt, line 808: I got it to work, but I think this is the part that is throwing me. How do we know that rustc and gccrs will continue to implement our  libCEED Rust structs in such a way that this will always work?

jedbrown: It has nothing to do with C layout, just whether a newtype holding a single pointer has the same memory representation as the pointer it holds. I believe that is guaranteed. We could check that each pointer is non-null, but if pointers being non-null is already a contract enforced on the C side, then it would be redundant.

rust/libceed/src/lib.rs
```diff
@@ -132,11 +134,24 @@ pub enum EvalMode {
...
+}
 
 // -----------------------------------------------------------------------------
 // Ceed error
 // -----------------------------------------------------------------------------
-type Result<T> = std::result::Result<T, CeedError>;
+pub type Result<T> = std::result::Result<T, CeedError>;
 
 #[derive(Debug)]
 pub struct CeedError {
```
jeremylt, line 157: We should probably just call this an `Error`, as it is already `libceed::[Ceed]Error`

jedbrown: Agreed.

jeremylt:
Ok, this all should be fine to merge now - I think I'll talk a bit about our modern interfaces in my PSAAP slides

doc/sphinx/source/releasenotes.md
```diff
@@ -7,6 +7,16 @@ for each release of libCEED.
 
 ## Current `main` branch
 
+### Interface changes
+
+- Update {c:func} `CeedQFunctionGetFields` and {c:func} `CeedOperatorGetFields` to include number of fields.
+- QFunction and Operator field objects, `CeedQFunctionField` and `CeedOperatorField`, and associated getters, {c:func}`CeedQFunctionGetFields`; {c:func}`CeedQFunctionFieldGetName`; {c:func}`CeedQFunctionFieldGetSize`; {c:func}`CeedQFunctionFieldGetEvalMode`; {c:func}`CeedOperatorGetFields`; {c:func}`CeedOperatorFieldGetElemRestriction`; {c:func}`CeedOperatorFieldGetBasis`; and {c:func}`CeedOperatorFieldGetVector`, promoted to the public API.
```
jedbrown, line 13: Tiny nit, but better to use consistent tense. (Others in this block use imperative, which is consistent with most strict commit message conventions.)
```suggestion
- Promote to the public API: QFunction and Operator field objects, `CeedQFunctionField` and `CeedOperatorField`, and associated getters, {c:func}`CeedQFunctionGetFields`; {c:func}`CeedQFunctionFieldGetName`; {c:func}`CeedQFunctionFieldGetSize`; {c:func}`CeedQFunctionFieldGetEvalMode`; {c:func}`CeedOperatorGetFields`; {c:func}`CeedOperatorFieldGetElemRestriction`; {c:func}`CeedOperatorFieldGetBasis`; and {c:func}`CeedOperatorFieldGetVector`.
```

rust/libceed/src/operator.rs
```diff
@@ -20,18 +20,275 @@
...
+        if ptr == unsafe { bind_ceed::CEED_ELEMRESTRICTION_NONE } {
+            ElemRestrictionOpt::None
+        } else {
+            let slice = unsafe {
+                std::slice::from_raw_parts(
+                    &ptr as *const bind_ceed::CeedElemRestriction as *const crate::ElemRestriction,
+                    1 as usize,
+                )
+            };
+            ElemRestrictionOpt::Some(&slice[0])
```
jedbrown, line 144: I found this safe and less contorted way to do this. https://docs.rs/ref-cast/1.0.6/ref_cast/

jedbrown (approved):
LGTM any time. We could merge now and use `ref_cast` later. I'd love to make progress on #807 or #805 this week.

PR #809, opened 2021-09-11
[#808][OCCA] Avoid verifying objects in destructors

### LINKED ISSUE #808: OCCA Backend Builds Operator to Destroy
See PR #806 

The OCCA backend currently checks the operator fields as part of destroying the operator. This doesn't really make any sense and it results in a CI failure in PR #806 

Offending code:
```C++
    int Operator::ceedDestroy(CeedOperator op) {
      delete Operator::from(op); // Is there a way to only delete this if its already been created?
      return CEED_ERROR_SUCCESS;
    }
```

It should be perfectly reasonable for a developer to create and then immediately destroy an operator as part of some incremental development process.

dmed256:
## Description

Destructors can be called on partial objects. Avoid building the full object during the destructor

Should close #808

### DIFF
--- a/backends/occa/ceed-occa-qfunction.cpp
+++ b/backends/occa/ceed-occa-qfunction.cpp
@@ -42,7 +42,7 @@ namespace ceed {
       ierr = CeedQFunctionGetData(qf, &qFunction);
       CeedOccaFromChk(ierr);
 
-      return qfunction;
+      return qFunction;
     }
 
     QFunction* QFunction::from(CeedQFunction qf) {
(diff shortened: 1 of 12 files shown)


backends/occa/ceed-occa-qfunction.cpp
```diff
@@ -30,17 +30,28 @@ namespace ceed {
...
       }
 
       int ierr;
-      QFunction *qFunction;
+      QFunction *qFunction = NULL;
 
       ierr = CeedQFunctionGetData(qf, &qFunction);
       CeedOccaFromChk(ierr);
 
+      return qfunction;
```
jeremylt, line 45:
```suggestion
      return qFunction;
```

jeremylt (approved):
Works in CI with the suggested change: https://gitlab.com/libceed/libCEED/-/jobs/1585697355

PR #810, opened 2021-09-13
Adjust cuda-gen launch to limit z dimension of thread block 

nbeams:
I was not thorough enough in my testing of #804.  In the case of low-order elements (e.g. linear elements in 3D), you can sometimes end up with such a small `threads_per_elem` that the new `BlockGridCalculate` function picks a size greater than the allowable thread block size for z.  With linear elements and applying a mass operator, it was trying to use 71, when the limit was 64.

This PR just adds the z-dimension limit as a parameter to `BlockGridCalculate` and checks the final `block[2]` value against it.


jedbrown (approved):
Thank you!

PR #812, opened 2021-09-15
Promote Operator check to public API; Strengthen composite Operator check

jeremylt:
Promote `CeedOperatorCheckReady` to public API


rust/libceed/src/operator.rs
```diff
@@ -428,6 +428,12 @@ impl<'a> OperatorCore<'a> {
     }
 
     // Common implementations
+    pub fn check(&self) -> crate::Result<i32> {
+        let mut is_ready = false;
+        let ierr = unsafe { bind_ceed::CeedOperatorIsReady(self.ptr, &mut is_ready) };
+        self.check_error(ierr)
```
jedbrown, line 434: Doesn't actually use the result `is_ready`

rust/libceed/src/operator.rs
```diff
@@ -750,6 +756,50 @@ impl<'a> Operator<'a> {
...
+    /// // Check
+    /// assert!(op_build.check().is_ok(), "Operator not correctly setup");
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[allow(unused_mut)]
+    pub fn check(mut self) -> crate::Result<Self> {
+        self.op_core.check()?;
+        Ok(self)
+    }
```
jedbrown, line 801: Why `mut`?

rust/libceed/src/operator.rs
```diff
@@ -1993,6 +2043,62 @@ impl<'a> CompositeOperator<'a> {
...
+    ///     .sub_operator(&op_build_mass)?
+    ///     .sub_operator(&op_build_diff)?;
+    ///
+    /// // Check
+    /// assert!(op_build.check().is_ok(), "Operator not correctly setup");
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[allow(unused_mut)]
+    pub fn check(mut self) -> crate::Result<Self> {
```
jedbrown, line 2097: And here; tests pass without it.

interface/ceed-operator.c
```diff
@@ -907,6 +910,29 @@ int CeedCompositeOperatorAddSub(CeedOperator composite_op,
...
+  @ref Advanced
+**/
+int CeedOperatorIsReady(CeedOperator op, bool *is_ready) {
+  int ierr;
+
+  *is_ready = false;
+  ierr = CeedOperatorCheckReady(op); CeedChk(ierr);
+  *is_ready = true;
+
+  return CEED_ERROR_SUCCESS;
```
jedbrown, line 933: These are surprising semantics. One would expect to return with `is_ready=false` and no error code. `CeedOperatorCheckReady` provides more information than a bool (in the form of an error code). If we want that to be available in a non-error path, then perhaps the innermost function should be this non-error check. I'm skeptical that it's worth doing that -- a caller can just handle the error.

jedbrown (commented):
May be useful, but semantics are weird now.

jeremylt:
CI failure due to usual intermittent runner problem

jedbrown (approved):

PR #815, opened 2021-09-17
Jeremy/occa fix array methods

jeremylt:
I updated this for changes to main. The `KernelBuilder` is still broken though
https://github.com/libocca/occa/pull/493#issuecomment-867814731


jeremylt:
I think I sorta have the builder stuff fixed, but there is something fishy going on. Also, need to fix `::occa::linalg`

jeremylt:
Ok, it looks like this OCCA PR (https://github.com/libocca/occa/pull/442) removed the empty initalizer for `kernerBuilder`s, which is the remaining broken piece of this PR.
https://github.com/libocca/occa/commit/9443ce0d8c78d9b99435e55aa0046bb82ba7a0ad#diff-2922e658903db536f88398883ba8d03260716330634f947493981228db63f11cL11

jeremylt:
It seems like there have been enough changes to OCCA that the kernels in this backend just don't build on OCCA main anymore.

PR #818, opened 2021-09-27
Examples - doc

LeilaGhaffari:
Included "Running the mini-app" section to the Navier-Stokes mini-app.


jeremylt:
Small fix we can probably wrap into this PR:
image
It would make much more sense for the 'CEED Benchmark Problems' to be mentioned before the 'PETSc demos and BPs'

examples/fluids/index.md
```diff
@@ -297,8 +267,9 @@ $$
 
 where $P_0$ is the atmospheric pressure.
 For this problem, we have used no-slip and non-penetration boundary conditions for $\bm{u}$, and no-flux for mass and energy densities.
-This problem can be run with:
 
-```
-./navierstokes -problem density_current
+## Running the mini-app
```
jeremylt, line 271: Can we move this to the top like we have it for the solids example? At least to me it makes more sense there.

examples/solids/qfunctions/finite-strain-neo-hookean-initial-2.h
```diff
@@ -355,7 +355,7 @@ CEED_QFUNCTION(ElasFSInitialNH2dF)(void *ctx, CeedInt Q,
                                       {q_data[7][i],
                                        q_data[8][i],
                                        q_data[9][i]}
-                                      };                              
+                                      };
```
jeremylt, line 358: We should modify the top level Makefile so that `make style` reaches deep enough into these example files to catch this sort of thing for us.

examples/fluids/README.md
```diff
@@ -7,390 +7,499 @@ The state variables are mass density, momentum density, and energy density.
...
-| `-viz_refine`                   | Use regular refinement for visualization                                            | `0`                     |
-| `-degree`                       | Polynomial degree of tensor product basis (must be >= 1)                            | `1`                     |
-| `-q_extra`                      | Number of extra quadrature points                                                   | `2`                     |
-| `-output_freq`                  | Frequency of output, in number of steps                                             | `10`                    |
-| `-continue`                     | Continue from previous solution                                                     | `0`                     |
-| `-output_dir`                   | Output directory                                                                    | `.`                     |
+% inclusion-fluids-marker
 
-### Advection
+The Navier-Stokes mini-app is controlled via command-line options, the following of which are common among all problem cases.
```
jeremylt, line 24:
```suggestion
The Navier-Stokes mini-app is controlled via command-line options.
The following options are common among all problem cases:
```

"problem cases" feels a little awkward to me?

examples/fluids/README.md
```diff
@@ -7,390 +7,499 @@ The state variables are mass density, momentum density, and energy density.
...
+  - View comprehensive information about run-time options
+  -
+:::
 
-Energy Density:
-    0.0 flux
+In addition to the command line options just shown, one may set options specific to each problem case following:
 
-##### Translation
+### Runtime options for Advection 2D (-problem advection2d)
```
jeremylt, line 100: perhaps
```
For the 2D advection problem, the following additional command-line options are available:
```

and so on for the other problems?

examples/fluids/README.md
```diff
@@ -7,390 +7,499 @@ The state variables are mass density, momentum density, and energy density.
...
+  -
+
+* - `-euler_test`
+  - Euler test option (`t1`-`t4` and `none`)
+  - `none`
+  -
+:::
+
+This problem represents the exact solution to five test cases, which are selected by `-euler_test`, where
+- `none`: Isentropic vortex
```
jeremylt, line 338: perhaps we should make this `t0`? or `test_i` for all of the options?

examples/fluids/README.md
LeilaGhaffari: Haha, right! `problems` would be enough!

examples/fluids/index.md
jeremylt: In my head, the 'running the mini-app' section provides a quick reference that you would frequently use and the individual problem sections are a deep dive into the details that you look at less often.

Ultimately, I have a mild preference for 'running the mini-app' before the theory, but I'm fine with either order so long as the two mini-apps match.

examples/fluids/index.md
```diff
@@ -6,6 +6,13 @@ This example is located in the subdirectory {file}`examples/fluids`.
 It solves the time-dependent Navier-Stokes equations of compressible gas dynamics in a static Eulerian three-dimensional frame using unstructured high-order finite/spectral element spatial discretizations and explicit or implicit high-order time-stepping (available in PETSc).
 Moreover, the Navier-Stokes example has been developed using PETSc, so that the pointwise physics (defined at quadrature points) is separated from the parallelization and meshing concerns.
 
+## Running the mini-app
+
+```{include} README.md
+:start-after: inclusion-fluids-marker
+```
```
LeilaGhaffari, line 13: OK, this makes it compatible with the solids example.

examples/fluids/README.md
LeilaGhaffari: Done!

jeremylt (approved):
Overall this LGTM. Not sure if @jedbrown has any comments.

LeilaGhaffari:
Thanks, @jeremylt for the review!

jedbrown (approved):
Thanks for putting this together. My only comment is that `-euler_test isentropic_vortex` or `-analytic_test euler_isentropic_vortex` would be more informative than a pure numbering system and that the simplifications (which were mostly made for debugging purposes) probably don't need to be user-facing or run as part of the test suite. (After all, they don't exercise libCEED differently, but it takes quite a some time to run each of these in GPU tests/CI.)

LeilaGhaffari:
Very good points. I just dropped those tests from the documentation and the test suite. Do you think I should remove the corresponding CL options as well? I thought those could help developers at some point, maybe!

examples/fluids/README.md
```diff
@@ -326,21 +326,9 @@ For the Isentropic Vortex problem, the following additional command-line options
...
-- `test_1`: Constant zero velocity, density constant, total energy constant
-- `test_2`: Constant nonzero velocity, density constant, total energy constant
-- `test_3`: Constant zero velocity, pressure constant, density and internal energy non-constant
-- `test_4`: Constant nonzero velocity, pressure constant, density and internal energy non-constant
-
-and it can be run with:
+This problem can be run with:
 
 ```
 ./navierstokes -problem euler_vortex -mean_velocity .5,-.8,0. -euler_test test_0
```
jedbrown, line 334: I think this would become `-euler_test isentropic_vortex`, though that's kind of repetitive with `-problem euler_vortex`. If it's the default, then this option wouldn't need to be documented here. It can still be found via `-help`.

jedbrown:
I think keeping the option is okay because it makes debugging the example easier, though we might take it out if we have an independent fluids/solids app that we can use for mechanics/solvers research.

examples/fluids/README.md
LeilaGhaffari: Oops! Sorry, I missed that one! Just fixed it, thanks!

jedbrown (approved):
LGTM after squashing what makes sense to squash (some updates/fixes to previous commits in this branch).

LeilaGhaffari:
This is ready to merge if the commits look good now.

PR #820, opened 2021-09-30
Add OCCA backend warning

jeremylt:
See #816, the OCCA backend is not receiving regular maintenance, is less feature complete than the other backends, and is less performant. This change alerts users that this backend is experimental.

### DIFF
--- a/include/ceed-backend.h
+++ b/include/ceed-backend.h
@@ -1,3 +1,0 @@
-#include <ceed/backend.h>
-
-#warning <ceed-backend.h> is deprecated and will be removed in libCEED v0.9; use <ceed/backend.h>
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -29,8 +29,8 @@ noether-rocm:
     - echo "-------------- OCCA ----------------" && make -C $OCCA_DIR info
 # libCEED
     - make configure HIP_DIR=/opt/rocm OPT='-O -march=native -ffp-contract=fast'
-#    Note: OCCA CPU backends currently disabled in CI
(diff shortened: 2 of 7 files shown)


include/ceed-backend.h
```diff
@@ -1,3 +1,3 @@
 #include <ceed/backend.h>
 
-#warning <ceed-backend.h> is deprecated and will be removed in libCEED v0.9; use <ceed/backend.h>
+#warning "<ceed-backend.h> is deprecated and will be removed in libCEED v0.9; use <ceed/backend.h>"
```
jedbrown, line 3: We already release v0.9....

jedbrown (approved):
Should we disable occa/openmp tests that are failing on noether?

include/ceed-backend.h
jeremylt: Yeah. That was an out of cycle release, so perhaps we delete these files now in anticipation of v0.10?

jeremylt:
The openmp mode was somewhat flakey iirc, so I don't object to disabling those tests in CI. I'll add that in the morning.

include/ceed-backend.h
jedbrown: Harmless, but I'm fine either way.

jeremylt:
Strange. I disabled the OCCA openmp backend for testing and the problem moved over to OCCA hip.

jeremylt:
I adjusted the CI. I'm not sure what's going on, but now we just verify that the OCCA backend can build.

jeremylt:
lv is stuck, but otherwise I think we're good to merge now

PR #822, opened 2021-10-13
Add support for additional QF source files

jeremylt:
**Feedback needed!**

This PR adds support for multiple QF source files. This will help reduce duplication, as common helper functions can be provided as `CEED_QFUNCTION_HELPER`s that get included and inlined as needed.

All backends except for OCCA pass the new test I wrote.

The actual change is pretty straightforward - The user provides a semicolon separated list of source file paths. I search the provided source filepath string for semicolons and add the contents of each file to the string buffer for the JiT.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,7 +7,7 @@ trigger:
 - main
 
 pool:
-  vmImage: 'ubuntu-latest'
+  vmImage: 'Ubuntu-20.04'
 
 container: jedbrown/petsc-root:latest
 


jeremylt:
Previously these parts of the CUDA and HIP backends were already pretty copy-paste, but I went ahead and refactored. 

I'm chasing down minor cross-language changes for the refactor.

YohannDudouit:
@jeremylt I'm very confused as why you're constantly using `force-push`, I seldom use it, and it is usually encouraged to never use it on `public` branches.

YohannDudouit:
When your branch is under review it is similar to having multiple developers. When you force push, GitHub and I are lost, and I can't track what you actually changed since my last review... I mostly have to review everything from scratch, that's unnecessarily tedious. If I also checked locally the branch for testing for instance, I also have to do all sorts of manipulations with git to get back to a normal state. If you want more arguments, a simple google search will give you a ton.

jeremylt:
I put all of my changes since your last review into a single commit and didn't modify the two commits from before your review. It's all in the last commit.

Anyways, we've talked offline quite a bit about refactoring the copy-paste parts of the GPU  backends. Do we actually want to make an issue and PR to finally address this? Like I said, I feel like that should be intentionally with a premeditated design rather than squished into other PRs with no big picture design.

PR #826, opened 2021-10-20
Separate JiT Tools

jeremylt:
Edit: This is a new approach to adding mulit-QFunction source support, as suggested from my last attempt.

What about this approach? If we like this, then we can use this to also help reduce repeated kernel source and various `#define`s across the CUDA/HIP backend families.


nbeams:
The naming problem, as I see it, is that arguably the most crucial aspect of the actual JIT compilation -- the CUDA/HIP compile functions -- won't be in this header...  But we probably also don't want to make it too specific in case there are future common (exact same for CUDA or HIP backends) functions related to JIT that we want to add here...

What if we a) move compilation-related functions for CUDA into a new file (the functions corresponding to the ones in `ceed-hip-compile`), 
b) name this new file `ceed-cuda-jit` and rename the HIP file `ceed-hip-jit`
c) go with `ceed-jit-tools` for this new file as you suggested (or something else indicating it's for "extra" common JIT stuff)? 

Just throwing ideas out...I realize that would be a lot more code movement and renaming.

nbeams (approved):
Thanks for all the work on the organization @jeremylt.  I think this is a lot cleaner, and I like that it brings the CUDA and HIP backends more in line with each other. 

Do you know what caused the GitLab CUDA failure?  I tried CUDA tests on a machine locally, and everything looked OK.

jeremylt:
CI on that machine is mildly flakey, so it is probably just a suprrious issue. I'm re-running.

jeremylt:
Yep, CI on that machine is clean on the re-run.

backends/avx/ceed-avx-blocked.c
```diff
@@ -56,6 +57,8 @@ static int CeedInit_Avx(const char *resource, Ceed ceed) {
 // Backend Register
 //------------------------------------------------------------------------------
 CEED_INTERN int CeedRegister_Avx_Blocked(void) {
-  return CeedRegister("/cpu/self/avx/blocked", CeedInit_Avx, 30);
+  const char prefix[] = "/cpu/self/avx/blocked";
+  if (getenv("CEED_DEBUG")) fprintf(stderr, "Backend Register: %s\n", prefix);
+  return CeedRegister(prefix, CeedInit_Avx, 30);
```
YohannDudouit, line 62: For a PR that claims to factorize code, it looks like the diffs in the 10 first files is just duplicated code, I think this calls for a function.

YohannDudouit (changes requested):
I think all this duplicated code could be replaced by one function that convey the intent.

backends/cuda/ceed-cuda-jit.c
```diff
@@ -0,0 +1,173 @@
...
+//------------------------------------------------------------------------------
+int CeedRunKernelDimSharedCuda(Ceed ceed, CUfunction kernel, const int gridSize,
+                               const int blockSizeX, const int blockSizeY,
+                               const int blockSizeZ, const int sharedMemSize,
+                               void **args) {
+  CeedChk_Cu(ceed, cuLaunchKernel(kernel, gridSize, 1, 1,
+                                  blockSizeX, blockSizeY, blockSizeZ,
+                                  sharedMemSize, NULL, args, NULL));
+  return CEED_ERROR_SUCCESS;
+}
```
YohannDudouit, line 173: missing newline.

backends/cuda/ceed-cuda-jit.h
```diff
@@ -0,0 +1,49 @@
...
+CEED_INTERN int CeedRunKernelDimCuda(Ceed ceed, CUfunction kernel,
+                                     const int gridSize,
+                                     const int blockSizeX, const int blockSizeY,
+                                     const int blockSizeZ, void **args);
+
+CEED_INTERN int CeedRunKernelDimSharedCuda(Ceed ceed, CUfunction kernel,
+    const int gridSize, const int blockSizeX, const int blockSizeY,
+    const int blockSizeZ, const int sharedMemSize, void **args);
+
+#endif // _ceed_cuda_jit_h
```
YohannDudouit, line 49: newline

backends/cuda/ceed-cuda-jit.h
```diff
@@ -0,0 +1,49 @@
...
+
+CEED_INTERN int CeedGetKernelCuda(Ceed ceed, CUmodule module, const char *name,
+                                  CUfunction *kernel);
+
+CEED_INTERN int CeedRunKernelCuda(Ceed ceed, CUfunction kernel,
+                                  const int gridSize,
+                                  const int blockSize, void **args);
+
+CEED_INTERN int CeedRunKernelAutoblockCuda(Ceed ceed, CUfunction kernel,
+    size_t size, void **args);
```
YohannDudouit, line 38: style.

backends/cuda/ceed-cuda-jit.h
```diff
@@ -0,0 +1,49 @@
...
+
+CEED_INTERN int CeedRunKernelDimCuda(Ceed ceed, CUfunction kernel,
+                                     const int gridSize,
+                                     const int blockSizeX, const int blockSizeY,
+                                     const int blockSizeZ, void **args);
+
+CEED_INTERN int CeedRunKernelDimSharedCuda(Ceed ceed, CUfunction kernel,
+    const int gridSize, const int blockSizeX, const int blockSizeY,
+    const int blockSizeZ, const int sharedMemSize, void **args);
```
YohannDudouit, line 47: style.

backends/cuda/ceed-cuda-qfunction-load.cpp
```diff
@@ -66,9 +46,17 @@ extern "C" int CeedCudaBuildQFunction(CeedQFunction qf) {
   ierr = CeedQFunctionGetFields(qf, &numinputfields, &qfinputfields, &numoutputfields, &qfoutputfields);
   CeedChkBackend(ierr);
 
+  // Additional kernels
+  char source_path[CEED_MAX_PATH_LEN] = __FILE__;
+  CeedInt end = strrchr(source_path, '/') - source_path;
+  strncpy(&source_path[end], "/kernels/cuda-qfunction.h", 26);
+  char *qFunctionKernels;
+  ierr = CeedLoadSourceToBuffer(ceed, source_path, &qFunctionKernels);
```
YohannDudouit, line 54: It seems that this pattern appears in multiple places, this should be factorized into a function I think.

backends/cuda/ceed-cuda-qfunction.c
```diff
@@ -183,17 +123,14 @@ int CeedQFunctionCreate_Cuda(CeedQFunction qf) {
...
   // Empty source path indicates user must supply Q-Function
-  if (source[0] != '\0') {
-    const char *funname = strrchr(source, ':') + 1;
-    data->qFunctionName = (char *)funname;
-    const int filenamelen = funname - source;
-    char filename[filenamelen];
-    memcpy(filename, source, filenamelen - 1);
-    filename[filenamelen - 1] = '\0';
-    ierr = CeedCudaLoadQFunction(qf, filename); CeedChkBackend(ierr);
+  if (source_path[0] != '\0') {
```
YohannDudouit, line 129: I would prefer self-documented code, I would suggest making a function with a name that convey what is in the doc and that does that.

backends/cuda/kernels/cuda-qfunction.h
```diff
@@ -0,0 +1,44 @@
...
+// Write at quadrature points
+//------------------------------------------------------------------------------
+template <int SIZE>
+inline __device__ void writeQuads(const CeedInt quad, const CeedInt nquads, const CeedScalar* r_v, CeedScalar* d_v) {
+  for(CeedInt comp = 0; comp < SIZE; ++comp) {
+    d_v[quad + nquads * comp] = r_v[comp];
+  }
+}
+
+#endif // _cuda_qfunction_kernels
```
YohannDudouit, line 44: newline.

backends/hip-gen/ceed-hip-gen-qfunction.c
```diff
@@ -124,15 +61,19 @@ int CeedQFunctionCreate_Hip_gen(CeedQFunction qf) {
...
-  const char *funname = strrchr(source, ':') + 1;
-  data->qFunctionName = (char *)funname;
-  const int filenamelen = funname - source;
-  char filename[filenamelen];
-  memcpy(filename, source, filenamelen - 1);
-  filename[filenamelen - 1] = '\0';
-  ierr = loadHipFunction(qf, filename); CeedChkBackend(ierr);
+  char *source_path;
+  ierr = CeedQFunctionGetSourcePath(qf, &source_path); CeedChkBackend(ierr);
+  if (source_path[0] != '\0') {
```
YohannDudouit, line 66: You forgot the documentation, as a result the code becomes obscure -> function.

backends/hip/ceed-hip-qfunction.c
```diff
@@ -175,15 +114,20 @@ int CeedQFunctionCreate_Hip(CeedQFunction qf) {
...
+    ierr = CeedQFunctionGetKernelName(qf, &data->qFunctionName);
+    CeedChkBackend(ierr);
+    ierr = CeedLoadSourceToBuffer(ceed, source_path, &data->qFunctionSource);
+    CeedChkBackend(ierr);
+  } else {
+    data->module = NULL;
+    data->qFunctionName = "";
+    data->qFunctionSource = "";
+    data->qFunction = NULL;
+  }
```
YohannDudouit, line 130: Actually this whole pattern seems to repeat, so the whole pattern should probably be a function.

YohannDudouit (changes requested):
I would unify `hip/ kernels/*` and `cuda/kernels/*` in to `gpu_common_kernels/*` (or something similar), currently the code is duplicated. In the future it could be different, Hip/Cuda could have different code, but I think that the design would be better if we abstracted the fact that we target Hip or Cuda. This would allow to test new implementations in a more modular way.

jeremylt:
I really don't think this is the right PR to combine CUDA and HIP kernel code together. I'm still (again) trying to keep the work to a series of smaller PRs rather than making a super PR that takes forever for anyone to review.

backends/cuda/ceed-cuda-jit.h
jeremylt: How so? `make style` doesn't object to this line.

backends/cuda/ceed-cuda-jit.h
jeremylt: same question

backends/hip-gen/ceed-hip-gen-qfunction.c
jeremylt: Which part did I forget the documentation of? `CeedQFunctionGetSourcePath()` is documented. `CeedQFunctionGetKernelName()` is documented. `CeedLoadSourceToBuffer()` is documented.

nbeams:
Also most of the kernels are duplicated, but not completely (there are some differences in cuda-shared and hip-shared, for example.  Changing cuda-shared to do what hip-shared does with the basis data memory might work well, but we definitely don't want to go back the other way, since it was a major performance problem for HIP).

backends/avx/ceed-avx-blocked.c
jeremylt: I added `CeedDebugEnv()`

YohannDudouit:
@jeremylt You'll have to explain how reviewing two times the (almost) same code is faster than reviewing it once.

@nbeams This is actually a good argument that goes in favor of what I'm proposing, if there was the modularity I'm talking about we could test your Hip improvements with Cuda.

backends/cuda/ceed-cuda-jit.h
YohannDudouit: Just look at the style of the other functions in this file.............

jeremylt:
This PR should *only* be about adding support for QFunction header files that include other QFunction header files. We shouldn't turn this into multi-QFunction header file support + HIP/CUDA kernel reunification + HIP/CUDA gen backend cleanup + ...

backends/cuda/ceed-cuda-jit.h
jeremylt: The indentation changes based upon the length of the function name. The style tool automatically handles that change for us.

backends/hip-gen/ceed-hip-gen-qfunction.c
YohannDudouit:
```c++
if (source_path[0] != '\0') {
```
usually you have this comment above:
```c++
// Empty source path indicates user must supply Q-Function
```

YohannDudouit:
@jeremylt This was really not obvious from the title and the description of the PR, I inferred the intent of your PR from the code...

backends/cuda-shared/kernel-strings/cuda-shared-basis.h
```diff
@@ -0,0 +1,752 @@
...
+//
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+#ifndef cuda_shared_basis_kernels
+#define cuda_shared_basis_kernels
```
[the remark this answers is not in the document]
YohannDudouit: I proposed that a long time ago, we discussed that with @nbeams to also have some sort of `cuda-gen/magma`, I think the idea is good.

jeremylt:
Sorry, the context was a little more obvious 3 weeks ago when this PR was fresh on the heels of those 2 rejected PRs so I didn't include as good of a description as I should have. I updated the description.

backends/hip-gen/ceed-hip-gen-qfunction.c
jeremylt: That comment isn't applicable in the HIP/CUDA gen backends though. We must error if no QFunction source is provided. I suspect that in the HIP/CUDA ref backends we don't actually need the `else` clause though.

nbeams:
> This is actually a good argument that goes in favor of what I'm proposing, if there was the modularity I'm talking about we could test your Hip improvements with Cuda.

Yeah, but there's a chance that what's an improvement for HIP isn't an improvement for CUDA, so we may need to maintain separate files, anyway.  Plus, in the case of cuda-shared, there are also changes in the *cu file compared to its HIP counterpart, not just the kernel strings in the *h files.

We can obviously cut down on code duplication, but I'm not sure we can (or should) get _everything_ into a common directory, and I definitely think those changes and discussion would warrant a separate PR.

backends/hip-gen/ceed-hip-gen-qfunction.c
YohannDudouit: That's even more reasons to centralize this logic into a function.

jeremylt: I don't understand. We do two different things in four different places so we should make a function that does these two different things in four different places?

Instead of just saying 'make it into a function' can you please provide more details? Right now the logic is just 'if there is a QFunction source path, then call these two functions' and for two code gen backends there is the extra clause 'if there is not, then error'.

YohannDudouit:
>Yeah, but there's a chance that what's an improvement for HIP isn't an improvement for CUDA, so we may need to maintain separate files, anyway. Plus, in the case of cuda-shared, there are also changes in the *cu file compared to its HIP counterpart, not just the kernel strings in the *h files.

@nbeams I think you didn't get my point, we can have different implementations of the same algorithms that live under `common_gpu_kernels`, then which one we decide on using is chosen when loading the source, and it can be different for hip and cuda. The purpose being to try different implementations for different scenarios. We could imagine loading different implementations not only based on hip/cuda, but also on the number of quadrature points and degrees of freedom. We already know that there is no best performance approach for all cases, that different implements work better for different cases, this design could also handle that.

>We can obviously cut down on code duplication, but I'm not sure we can (or should) get everything into a common directory, and I definitely think those changes and discussion would warrant a separate PR.

I don't see any argument supporting your point of view. I actually think the opposite, I think generic code is superior to specialized code for maintainability, readability, extensibility, modularity, performance, and safety.

jeremylt:
I created a separate issue for the discussion of a massive overhaul of all of the GPU backends. It really shouldn't be part of this PR. This PR should just be enabling multi-QFunction source support in a way that allows for future improvements.

nbeams:
> I don't see any argument supporting your point of view that we should not get everything into a common directory.

My argument is that if we end up with any kernels that are meant for *only* CUDA or HIP or whatever (as the case with the current cuda-shared), so those kernels really shouldn't be used with other hardware, then it's misleading to put them in a directory called `common-gpu` (or similar), because they're _not_ common.  

But mostly, I was supporting @jeremylt's opinion that this should happen in a separate PR.

backends/hip-gen/ceed-hip-gen-qfunction.c
YohannDudouit: > I don't understand. We do two different things in four different places so we should make a function that does these two different things in four different places?

Yes, the idea is to have more self-documented code, currently when I read your code I see __how__ you do things, but I don't know __what__ you do, I have to deduce the "what" from the "how". Functions allow you to separate level of abstractions so that when I read code I stay at the same level of abstraction, understand "what"it does without me having to know implementation details, and you don't mix low level code (ex: `source_path[0] != '\0'`) with high level code. The function names should suggest their high level intent (the "what"). You can either make one function that does it all if the details don't matter, if you think the distinction between the two different processing matters, then you can write two functions that convey the difference in the "what".

> Instead of just saying 'make it into a function' can you please provide more details? Right now the logic is just 'if there is a QFunction source path, then call these two functions' and for two code gen backends there is the extra clause 'if there is not, then error'.

I think the important part here is to find "what" it does, your "logic" only describes the "how".

jeremylt: This is C, so to some degree we are are stuck using some low level approaches. Is there a different way to check if the source file path string is empty that you would prefer?

In this chunk

```c
// Read source
char *source_path;
ierr = CeedQFunctionGetSourcePath(qf, &source_path); CeedChkBackend(ierr);
// Empty source path indicates user must supply Q-Function
if (source_path[0] != '\0') {
  ierr = CeedQFunctionGetKernelName(qf, &data->qFunctionName); CeedChkBackend(ierr);
  ierr = CeedLoadSourceToBuffer(ceed, source_path, &data->qFunctionSource); CeedChkBackend(ierr);
}
```

we are loading the kernel name and source file contents into the correct buffer for the implementation dependent object. We currently have 4 different objects being populated with this information, CUDA ref QFunction backend data, CUDA gen QFunction backend data, HIP ref QFunction data, and HIP gen QFunction backend data. I don't see an obvious clean way to make one function that takes in these four different function definitions.

I could wrap `CeedLoadSourceToBuffer()` in `CeedQFunctionLoadSourceToBuffer()` if you think it would be clearer. We could then check if that string is empty and error if the source buffer is empty for the code generation backends. That would get us down to 2 function calls in 4 places and 2 if statements in 2 places.

Side note: This is code you wrote that has been modified since that time and then copied into the HIP backend. When making changes to the GPU backends, I try to be somewhat conservative in my changes since I am not the primary maintainer of the GPU backends.

jeremylt: Ok, how about this pattern:
```c
ierr = CeedQFunctionGetKernelName(qf, &data->qFunctionName); CeedChkBackend(ierr);
ierr = CeedQFunctionLoadSourceToBuffer(qf, &data->qFunctionSource); CeedChkBackend(ierr);
```

YohannDudouit: I agree that the code I wrote for the cuda-gen backend years ago is difficult to read, difficult to maintain, and difficult to extend, and overall not a nice looking code. I have to say that it was a pain for me to write this in C, I had to mimic C++ generic template programming using C-strings, that was not fun. Just a reminder that this code was written to get performance on GPU which wasn't the case before for libCEED, the ambition was not to rewrite all the little things that I find badly written. I think though that once code has been reviewed and accepted we should not consider that the author is the only one responsible for maintaining it, the purpose of reviewing is also to make sure that the code is good enough so that pretty much __anyone__ in the project can read, maintain, and extend it. I don't think it's completely fair to complain on the quality of the code I wrote which the team reviewed (reviewers have their share of responsibility), and it should absolutely not be an excuse to write bad code now.

This PR is 6500 lines diff, 106 files modified, and most of this diff is duplicated code, as a result it's relatively a pain to review, and it feels like a programming version of the 7 errors game.

This being said, I think that having this kind of code is an anti-pattern to me:
```c
if (source_path[0] != '\0') {
   ierr = CeedLoadSourceToBuffer(ceed, source_path, &data->qFunctionSource); CeedChkBackend(ierr);
} else {
   return CeedError(ceed, CEED_ERROR_UNSUPPORTED,
                  "/gpu/hip/gen backend requires QFunction source code file");
}
```
This means that a user/developer __has to__ be aware of this special semantic of the `source_path` in order to use `CeedQFunctionLoadSourceToBuffer` properly, and there is no reason for that. This ` if` could be inside `CeedQFunctionLoadSourceToBuffer`, and the code that access the ` source_path` could also be inside `CeedQFunctionLoadSourceToBuffer`. If you want to customize the error message you can add a parameter. Then for the case that is different, you can either have a different behavior when no message is given, and document the semantic of it, or you create a function with a different name built on top of the previous function that has the expected different behavior.
By doing this you indeed obtain something looking like your previous answer, the benefit is that I directly understand __what__ is happening, and I don't have to even know __how__ it's done. Another benefit is that all the logic about the __how__ is in one place, reducing significantly the risk of errors when refactoring, and also improving readability, maintainability, and safety. In many aspects this PR is refactoring code, and for instance I cannot know as is if you didn't omit to modify this pattern somewhere else in the code.

jeremylt: I think you are now commented about a pattern I have already removed. Did you miss the update that I put in? Do you like the update I put in? Are you requesting changes to the update I put in?

I agree with your comments about the sprawling nature of this PR. This is exactly why I think we should be making a larger number of smaller PRs. I would love to see a full refactor of the GPU backends, but that should be done in a series of smaller PRs.

To be honest, I don't know enough about the design of the GPU backends at this time to feel comfortable considering myself a primary maintainer of those backends. I'm in the awkward position of pushing forward library features into code that others have designed. It really isn't reasonable to omit GPU implementations of these features so I'm tinkering with backend code I don't really feel like I have ownership over. It is a *very* frustrating situation that I have been trying to remedy by learning about and documenting the GPU backend code as I have time, but less hostile code review would be helpful.

YohannDudouit: I'm sorry that me telling you how I would improve this work makes you so emotional, this is not my intention, and I apologize for that. I already have spent a lot of time on this PR, and I really have much more important priorities right now, I don't have the time to micro-manage this work. Please ask me to review again your work when you feel confident that it's in a good state.

If you want me to explain anything to you about code that I wrote I'm always available for that. You can also open issues for documentation requests.

jedbrown: It's a lot of lines of code touched, but it is half the size.
```
106 files changed, 1812 insertions(+), 3815 deletions(-)
```
I'd hope we can lean on tests so that the reviewing need is mostly for whether this is boxing us into a position where future features (including performance optimization) will require excessive churn. If this is an improvement, but the concept can be pushed further, I think it's reasonable to approve and leave the next increment for another PR, which would hopefully be quite a bit smaller. We really appreciate your time reviewing -- I'll note that Jeremy also has other deliverables (for PSAAP, which funds him now). I think we should have a long-term governance and sustainability discussion sometime in the coming months.

On the specific issue of strings, I'm not a fan of the empty string being special. A pointer that can be NULL is like `Option<String>`, okay, and sometimes we need a `String`. But empty string (path) is in my book like any path that doesn't exist -- we can try to open it and give a reasonable error if it fails.

jeremylt: I was actually explicitly `NULL`ing the pointer to the string buffer in `CeedQFunctionLoadSourceToBuffer()` but didn't use that in my error checking. I swapped that in instead of using `strlen` since it is a bit clearer (though the effect is identical for a `NULL` pointer)
```C
// Read QFunction source
ierr = CeedQFunctionGetKernelName(qf, &data->qFunctionName); CeedChkBackend(ierr);
ierr = CeedQFunctionLoadSourceToBuffer(qf, &data->qFunctionSource); CeedChkBackend(ierr);
if (!data->qFunctionSource)
  return CeedError(ceed, CEED_ERROR_UNSUPPORTED,
                   "/gpu/cuda/gen backend requires QFunction source code file");
```
Those 2/3 statements seem really clear to me now.

PR #827, opened 2021-10-21
Levenshtein distance - Backend closest match

laylagi:
This uses Levenshtein distance to find the closest match to the user defined backend. 
I am seeing two problems though and I would appreciate your input:

1. Some tests, including Julia and Rust, are failing and since I am not familiar with these interfaces, I don't see why this change should affect them!
2. This algorithm works fine with something like `/cup/self/ref/serial` (finds `/cpu/self/ref/serial`) but with `/cpu`, it finds `/gpu/hip/ref` which is not what I was hoping to see! Any thoughts/recommendations?

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -723,6 +723,7 @@ int CeedInit(const char *resource, Ceed *ceed) {
   }
   // Using Levenshtein distance to find closest match
   if (match_len <= 1 || match_len != stem_length) {
+    // LCOV_EXCL_START
     size_t lev_dis = UINT_MAX;
     size_t lev_idx = UINT_MAX, lev_priority = CEED_MAX_BACKEND_PRIORITY;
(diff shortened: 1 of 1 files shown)


jeremylt:
I think we might need a modified Levenshtein algorithm where `lev(requested, backend) = 0` when all characters of `requested` match `backend` no matter how long `backend` is.

jedbrown:
I think that's right: we need to check exact prefix matches before evaluating distance. I think I'd prefer only evaluating distance when we fail to match, so it would never affect correct execution and it's facing can't be relied on, but it would produce a more useful error message.

jedbrown:
I think that's right: we need to check exact prefix matches before evaluating distance. I think I'd prefer only evaluating distance when we fail to match, so it would never affect correct execution and it's facing can't be relied on, but it would produce a more useful error message.

PR #833, opened 2021-10-29
makefile - respond to either V or VERBOSE

jeremylt:


jedbrown (approved):
This is the CMake convention and Bill tried it first (CMake muscle memory) so we may as well support it. Thanks for the clean impl.

PR #834, opened 2021-10-29
Enzyme in libCEED

LeilaGhaffari:
We have been able to reproduce `examples/solids/finite-strain-neo-hookean-initial-1.h` with [Enzyme](https://github.com/wsmoses/Enzyme) where Enzyme computes `dSdE` (\partial S / \partial E) in `deltaS = dSdE:deltaE` which was previously derived analytically.

CC: @KennethEJansen @wsmoses


LeilaGhaffari (commented):
@wsmoses , could you take a look at these changes and see if this makes sense?

wsmoses:
The changes seem fine, can you rebuild that pull request on Enzyme again, there were a few changes added which should ensure this works.

LeilaGhaffari:
> The changes seem fine, can you rebuild that pull request on Enzyme again, there were a few changes added which should ensure this works.

Indeed it did work! Thanks!

examples/solids/problems/cl-problems.h
```diff
@@ -3,12 +3,13 @@
...
 typedef enum {
-  ELAS_LINEAR = 0, ELAS_SS_NH = 1, ELAS_FSInitial_NH1 = 2, ELAS_FSInitial_NH2 = 3,
-  ELAS_FSCurrent_NH1 = 4, ELAS_FSCurrent_NH2 = 5, ELAS_FSInitial_MR1 = 6
+  ELAS_LINEAR = 0, ELAS_SS_NH = 1, ELAS_FSInitial_NH1 = 2, ELAS_FSInitial_NH1_AD = 3, ELAS_FSInitial_NH2 = 4,
+  ELAS_FSCurrent_NH1 = 5, ELAS_FSCurrent_NH2 = 6, ELAS_FSInitial_MR1 = 7
 } problemType;
 static const char *const problemTypes[] = {"Linear",
                                            "SS-NH",
                                            "FSInitial-NH1",
+                                           "FSInitial-NH1-AD",
```
jeremylt, line 12: I'd just call this `FSInitial-NH-AD`, personally since we're bypassing the mathematical formulation of the Jacobian.

examples/solids/Makefile
```diff
@@ -34,6 +34,16 @@ LDFLAGS = $(call pkgconf, --libs-only-L --libs-only-other $(PETSc.pc) $(ceed.pc)
...
 
+# Enzyme-AD
+ENZYME_LIB ?=
+ENZYMEFLAGS = -Xclang -load -Xclang $(ENZYME_LIB) -O2 -fno-unroll-loops -mllvm -enzyme-print
+
+ifneq ($(ENZYME_LIB),)
+  CFLAGS += $(ENZYMEFLAGS)
+else
+  $(error Must export ENZYME_LIB=/path/to/ClangEnzyme-XX.so)
+endif
```
jeremylt, line 45: Before merging, we should make this an optional dependency and make it a run-time error if the user tries to use a problem that uses Enzyme without having Enzyme installed.

examples/solids/problems/finite-strain-neo-hookean-initial-1-ad.c
```diff
@@ -0,0 +1,60 @@
...
+static CeedInt field_sizes[] = {9, 6, 1};
+
+ProblemData finite_strain_neo_Hookean_initial_1_ad = {
+  .setup_geo = SetupGeo,
+  .setup_geo_loc = SetupGeo_loc,
+  .geo_data_size = 10,
+  .quadrature_mode = CEED_GAUSS,
+  .residual = ElasFSInitialNH1F_AD,
+  .residual_loc = ElasFSInitialNH1F_AD_loc,
+  .number_fields_stored = sizeof(field_sizes) / sizeof(*field_sizes),
```
jeremylt, line 19: Probably worth propagating this change to all problems.

examples/solids/qfunctions/finite-strain-neo-hookean-initial-1-ad.h
```diff
@@ -0,0 +1,710 @@
+// Copyright (c) 2017, Lawrence Livermore National Security, LLC. Produced at
```
jeremylt, line 1: We can cut out the completely repeated functions (energy, diagnostic) and use both this header and `finite-strain-neo-hookean-initial-1.h` in `finite-strain-neo-hookean-initial-1-ad.c`

examples/solids/Makefile
LeilaGhaffari: Right, I should make that optional but since `ENZYME_LIB` is needed to compile the code, I think we should export it before the compilation.

examples/solids/problems/finite-strain-neo-hookean-initial-1-ad.c
LeilaGhaffari: Will do!

examples/solids/problems/cl-problems.h
LeilaGhaffari: Oh, I see the difference between `FSInitial-NH1` and `FSInitial-NH2` now. Fair point!

examples/solids/qfunctions/finite-strain-neo-hookean-initial-1-ad.h
LeilaGhaffari: How about [this approach](https://github.com/CEED/libCEED/pull/834/commits/764d88583e092121b5e5bf778d417cebf47926ec)?

examples/fluids/problems/problem-weak.c
```diff
@@ -0,0 +1,20 @@
...
+#include <stdbool.h>
+#include <stdlib.h>
+#include <ceed/backend.h>
+#include "../navierstokes.h"
+
+// This function provides a debug target for weak symbols
+// LCOV_EXCL_START
+static PetscErrorCode ProblemRegister_Weak(const char *name) {
+  // TODO: What should be the error message?
+  if (getenv("CEED_DEBUG")) fprintf(stderr, "Weak %s\n", name);
```
LeilaGhaffari, line 11: @jedbrown , I am not sure if this If-statement works fine since this function doesn't get called at all. Do I need to change the Makefile as well or is there something wrong in the implementation?

jedbrown: The weak symbol will get discarded by the linker if there is a strong (normal) symbol of the same name. To test this being called, you can build without Enzyme (or otherwise skip linking the strong symbol) or just create an extra weak symbol for which there is no corresponding strong symbol. This (when you run with `CEED_DEBUG=1` in your environment) is just logging that this symbol is weak and thus will error if selected at run-time.

LeilaGhaffari: I tried to build without Enzyme in the last commit which I hadn't pushed. It should be able to discard `EULER_VORTEX` when not using clang or ENZYME_LIB not provided, right?
Note I started with the fluids example since I am more familiar with that code.

jeremylt: To get this to work right requires changes in the makefile and the `problems` directory.

The existing register function for all current (and new) problems should remain as is.

You also need that new file for the weak registration of any problem when the required libs are not present. It looks like you have that. It should be pretty similar to the backend and gallery weak registration functions.

In the makefile, you will want to separately add the source file for the new Enzyme 'problem' to `src.c`
https://github.com/CEED/libCEED/blob/5b1362b9c136ba2a203fe6f05a8cefc3f59aa1b3/examples/fluids/Makefile#L39
sort of like what we do when we detect that AVX is supported by the compilation environment
https://github.com/CEED/libCEED/blob/5b1362b9c136ba2a203fe6f05a8cefc3f59aa1b3/Makefile#L333

You will then need this weak_last rule to help make sure that the weak registration of problems occurs last
https://github.com/CEED/libCEED/blob/5b1362b9c136ba2a203fe6f05a8cefc3f59aa1b3/Makefile#L479

LeilaGhaffari: Thanks, @jeremylt . These were very helpful.

jeremylt: awesome!

jeremylt: For your point 2) in the commit message, I made some changes in #826 along those lines. Each register function reports in debug mode so you can see everything that is registered. To see the effect, build the branch and run `CEED_DEBUG=1 ./build/t000-ceed /cpu/self`
You get
```
---------- Registering Backends ----------

Backend Register: /cpu/self/avx/blocked
Backend Register: /cpu/self/avx/serial
Backend Register: /gpu/cuda/ref
Backend Register: /gpu/cuda/gen
Backend Register: /gpu/cuda/shared
Backend Register: /gpu/hip/ref
Backend Register: /gpu/hip/gen
Backend Register: /gpu/hip/shared
** Weak Register: /gpu/cuda/magma
** Weak Register: /gpu/hip/magma
** Weak Register: /gpu/cuda/magma/det
** Weak Register: /gpu/hip/magma/det
Backend Register: /cpu/self/memcheck/blocked
Backend Register: /cpu/self/memcheck/serial
** Weak Register: /cpu/self/occa
** Weak Register: /cpu/openmp/occa
** Weak Register: /gpu/hip/occa
** Weak Register: /gpu/cuda/occa
Backend Register: /cpu/self/opt/blocked
Backend Register: /cpu/self/opt/serial
Backend Register: /cpu/self/ref/serial
Backend Register: /cpu/self/ref/blocked
Backend Register: /cpu/self/xsmm/blocked
Backend Register: /cpu/self/xsmm/serial
```

examples/solids/qfunctions/finite-strain-neo-hookean-initial-ad.h
```diff
@@ -509,6 +509,26 @@ CEED_QFUNCTION(ElasFSInitialNHdF_AD)(void *ctx, CeedInt Q,
...
+                                       CeedScalar *const *out) {
+  // Inputs
+  void *(*tape) = (void *(*))in[0];
+  // No outputs
+
+  // Quadrature Point Loop
+  CeedPragmaSIMD
+  for (CeedInt i=0; i<Q; i++) {
+    // Free allocated memory for tape
+    free(tape[i]);
```
LeilaGhaffari, line 526: This seems to be an invalid pointer. 
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff3feb859 in __GI_abort () at abort.c:79
#2  0x00007ffff40563ee in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff4180285 "%s\n") at ../sysdeps/posix/libc_fatal.c:155
#3  0x00007ffff405e47c in malloc_printerr (str=str@entry=0x7ffff417e4ae "free(): invalid pointer") at malloc.c:5347
#4  0x00007ffff405fcac in _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:4173
#5  0x000000000042e319 in ElasFSInitialNHFree_AD (ctx=<optimized out>, Q=216, in=<optimized out>, out=<optimized out>)
    at problems/../qfunctions/finite-strain-neo-hookean-initial-ad.h:526
#6  0x00007ffff529839f in CeedQFunctionApply_Ref (qf=0xfc3900, Q=216, U=0x24f89d0, V=0x24f8a60) at /home/leila/git/libCEED/backends/ref/ceed-ref-qfunction.c:54
```
I am either missing something or this is not the right approach to free the allocated memory for tape. @jedbrown @jeremylt , your insight would be appreciated!

examples/solids/src/setup-libceed.c
```diff
@@ -734,6 +733,35 @@ PetscErrorCode SetupLibceedFineLevel(DM dm, DM dm_energy, DM dm_diagnostic,
...
+  // Enzyme-AD
+  // ---------------------------------------------------------------------------
+  // Create the QFunction and Operator that frees memory created for Enzyme-AD
+  // ---------------------------------------------------------------------------
+  if (problem_data.tape) {
+    CeedBasis basis_tape;
+    CeedOperator op_tape;
+    CeedQFunction qf_tape;
+    CeedInt idx = problem_data.number_fields_stored - 1, num_comp_tape = 1;
+    CeedBasisCreateTensorH1Lagrange(ceed, dim, num_comp_tape, P, Q,
```
jeremylt, line 746: The tape is at the quadrature points, so the basis should be `CEED_BASIS_COLLOCATED` and an eval mode of `CEED_EVAL_NONE`

jeremylt:
```C

    CeedOperatorSetField(op_residual, problem_data.field_names[i],
                         data[fine_level]->elem_restr_stored_fields_i[i],
                         CEED_BASIS_COLLOCATED,
                         data[fine_level]->stored_fields[i]);
```
From `setup-libceed.c`

LeilaGhaffari: Thanks! It does free it now.

examples/solids/elasticity.c
```diff
@@ -914,6 +914,18 @@ int main(int argc, char **argv) {
     ierr = PetscFree(diagnostic_ctx); CHKERRQ(ierr);
   }
 
+  if(ceed_data[fine_level]->qf_tape) {
```
jeremylt, line 917:
```suggestion
  if (ceed_data[fine_level]->qf_tape) {
```

jeremylt:
This is looking pretty good overall. This is marked draft. What's left before it's ready for review? I'd like to migrate it over to Ratel when this branch is ready.

examples/fluids/Makefile
```diff
@@ -36,7 +38,16 @@ OBJDIR := build
 SRCDIR := src
 PROBLEMDIR := problems
 
-src.c := navierstokes.c $(sort $(wildcard $(PROBLEMDIR)/*.c)) $(sort $(wildcard $(SRCDIR)/*.c))
+src.c := navierstokes.c $(filter-out $(PROBLEMDIR)/%-ad.c, $(sort $(wildcard $(PROBLEMDIR)/*.c))) $(sort $(wildcard $(SRCDIR)/*.c))
+
+# Enzyme-AD
+ENZYME_LIB ?=
+ENZYMEFLAGS = -Xclang -load -Xclang $(ENZYME_LIB) -O2 -fno-unroll-loops -mllvm -enzyme-print
```
wsmoses, line 45: You can likely remove the `-fno-unroll-loops -mllvm -enzyme-print`

jedbrown: Good; yes, best to be minimally intrusive.

@wsmoses do you think it's feasible to use Enzyme within hiprtc? I think this would require building with the LLVM behind `hipcc`, or perhaps just an LLVM from the same LLVM major version (and expect it to be ABI compatible). We're starting to run on OLCF Crusher (same arch as Frontier) and it would be wonderful if we could use Enzyme there. (We can move to an issue in the Enzyme repo if that's a better place to discuss.)

wsmoses: We've used Enzyme on AMD kernels via Julia, and I don't see a reason that shouldn't work for hipcc (since indeed we should be compatible across LLVM 7-current). Frankly the only limiting factor is a lack of compatible ROCm-compatible GPU to try it out on. I can probably ask @jdoerfert to lend me one on ALCF.

jedbrown: If you email me your ssh public key, I can add you to our development box (Radeon VII -- a little old, but still useful for testing). We can probably also get you access to OLCF via ECP -- that takes longer because they mail you a token and you have to mail back a notarized form. If you have JLSE access (https://www.jlse.anl.gov/hardware-under-development/), that would also work.

LeilaGhaffari: Thanks, @wsmoses ! Just removed those flags.

examples/solids/elasticity.c
LeilaGhaffari: Oops! Just fixed it. Thanks!

examples/solids/qfunctions/finite-strain-neo-hookean-initial-ad.h
```diff
@@ -509,6 +524,26 @@ CEED_QFUNCTION(ElasFSInitialNHdF_AD)(void *ctx, CeedInt Q,
...
+
+  // Quadrature Point Loop
+  CeedPragmaSIMD
+  for (CeedInt i=0; i<Q; i++) {
+    // Free allocated memory for tape
+    free_tape(tape[i]);
+  } // End of Quadrature Point Loop
+
+  return 0;
+}
```
LeilaGhaffari, line 545: @jedbrown 
Here is where the main issue lies. We should instead allocate and free `tape` outside the QFunctions.

LeilaGhaffari:
I am trying to wrap this up but I am not able to compile it. Nothing has really changed since the last time that I was able to compile and run the code. @wsmoses , do you think I need to change anything in the implementation?

```
/libCEED/examples/solids $ make V=1
/home/leila/git/petsc/arch-clang/bin/mpicc -L/home/leila/git/petsc/arch-clang/lib -L/home/leila/git/libCEED/lib  -Wl,-rpath,/home/leila/git/petsc/arch-clang/lib  -Wl,-rpath,/home/leila/git/libCEED/lib   build/elasticity.o build/problems/finite-strain-mooney-rivlin-initial-1.o build/problems/finite-strain-neo-hookean-current-1.o build/problems/finite-strain-neo-hookean-current-2.o build/problems/finite-strain-neo-hookean-initial-1.o build/problems/finite-strain-neo-hookean-initial-2.o build/problems/linear.o build/problems/mooney-rivlin.o build/problems/neo-hookean.o build/problems/problem-register.o build/problems/problem-weak.o build/problems/small-strain-neo-hookean.o build/src/boundary.o build/src/cl-options.o build/src/matops.o build/src/misc.o build/src/setup-dm.o build/src/setup-libceed.o build/problems/finite-strain-neo-hookean-initial-ad.o  -lpetsc -lceed -lm -o elasticity
/usr/bin/ld: build/problems/finite-strain-neo-hookean-initial-ad.o: in function `getEnzymeSize':
/home/leila/git/libCEED/examples/solids/problems/../qfunctions/finite-strain-neo-hookean-initial-ad.h:179: undefined reference to `__enzyme_augmentsize'
/usr/bin/ld: build/problems/finite-strain-neo-hookean-initial-ad.o: in function `grad_S_fwd':
/home/leila/git/libCEED/examples/solids/problems/../qfunctions/finite-strain-neo-hookean-initial-ad.h:185: undefined reference to `__enzyme_augmentfwd'
/usr/bin/ld: build/problems/finite-strain-neo-hookean-initial-ad.o: in function `grad_S_rev':
/home/leila/git/libCEED/examples/solids/problems/../qfunctions/finite-strain-neo-hookean-initial-ad.h:193: undefined reference to `__enzyme_reverse'
/usr/bin/ld: /home/leila/git/libCEED/examples/solids/problems/../qfunctions/finite-strain-neo-hookean-initial-ad.h:197: undefined reference to `__enzyme_reverse'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:56: elasticity] Error 1
```

examples/solids/qfunctions/finite-strain-neo-hookean-initial-ad.h
```diff
@@ -0,0 +1,494 @@
...
+    for (CeedInt m = 0; m < 6; m++) {
+      E2work[m] = tempgradu[indj[m]][indk[m]] + tempgradu[indk[m]][indj[m]];
+      for (CeedInt n = 0; n < 3; n++)
+        E2work[m] += tempgradu[n][indj[m]]*tempgradu[n][indk[m]];
+    }
+
+    // Compute Swork with Enzyme-AD and get tape
+    CeedScalar Swork_[6], packed_tape[6];
+    S_fwd(Swork_, E2work, lambda, mu, packed_tape);
+    for (int j=0; j<6; j++) tape[j][i] = packed_tape[j];
```
jedbrown, line 314: Hi, @wsmoses. We normally expect the outer loop `for (CeedInt i=0; i<Q; i++)` up above to be vectorized, and thus we'd like the data here to have `i` as its inner index. Clang doesn't do as good a job as gcc vectorizing this variant, but clang and gcc both vectorize the current configuration variants (which we like better anyway, but this one is "harder" in what we ask Enzyme to do). The 6 here comes from `__enzyme_augmentsize` returning `48 = 6 * sizeof(CeedScalar)`, which makes practical sense. If the compiler does its job, this packing will vanish after inlining and the contents of `S_fwd` will used packed instructions. Is there a better way to do this?

I can imagine we add explicit blocking (say batches of 8 elements) and manually write the inner loop, but I'd love to keep that complexity separate.

wsmoses: From a pure performance standpoint, I don't think any extra massaging within the derivative to place the tape values at a location would be any different from the packing you have right now (at least in theory). Of course this doesn't mean that we can't make some nice syntactic sugar, though I'm not sure what that would be.

Regarding the size 8 to make it better for vectorization, can we not just round the result of `__enzyme_augmentsize` to the nearest `8*sizeof(CeedScalar)`?

jedbrown: What I meant by 8 was blocking over the quadrature point index `i`. Returning `6 * sizeof(CeedScalar)` is totally fine (better because it's less data), but this pack/unpack needs the result of `__enzyme_augmentsize` as a compile-time constant for the loop and data rearrangement to be eliminated. What I meant by unrolling is to do

```c
for (int i=0; i<Q; i+=8) {
   S_fwd(...);
}
```
and `S_fwd` would have its own fixed length-8 loop. I think this clutters the code and at the end of the day doesn't make optimization better. But I see we can assign `__enzyme_augmentsize` to a `static const` and that the optimizer replaces it with a numeric literal so we can have our own packing without magic numbers. So I think we're set, thanks.

jedbrown: Oh, or it's sometimes a compile-time constant. Hmm.

jedbrown: To clarify, I tried this on a simple function and it worked, but doesn't work here. We could have a run-time or compile-time size compatibility check and just assign `ElasFSInitialNH_AD_tape_bytes = 6 * sizeof(CeedScalar)` here.
```
/home/jed/src/libCEED/examples/solids/problems/../qfunctions/finite-strain-neo-hookean-initial-ad.h:179:53: error: initializer element is not a compile-time constant
static const size_t ElasFSInitialNH_AD_tape_bytes = __enzyme_augmentsize(computeS,
                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jed/src/libCEED/examples/solids/problems/finite-strain-neo-hookean-initial-ad.c:29:16: error: initializer element is not a compile-time constant
  .tape_size = ElasFSInitialNH_AD_tape_bytes,
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

wsmoses: The result of the augmentsize becomes a compile-time constant when the derivatives are generated (which in its current location in the compiler optimization pipeline, is shortly before vectorization). Because it depends on what the definition of the differentiated function is, it isn't available as a constant before then (and thus cannot be constexpr since c++ templates wouldn't be able to tell its value).

That said if you pass a constant as the tape size, Enzyme will check if the memory allocated is at least what it requires and throw an error if not.

jedbrown: Okay. I think I'd rather have a run-time error/warning saying to update this variable name to this size than let it fail inside Enzyme. @LeilaGhaffari Could you update your branch to make a constant for the tape size, then get the value produced by `__enzyme_augmentsize` and:
1. if the const size matches, run
2. if the const size is bigger, print a warning that it can be reduced to the run-time size
3. if the const size is smaller, error saying that it must be increased to the run-time size

I think this will be easy enough to maintain and also avoid surprises as we're working on constitutive models. (The checked number is useful information to know.)

jeremylt:
@LeilaGhaffari, this sounds pretty close to ready. I'd like to add this work to Ratel. It think it will take a bit of effort to adapt, but it shouldn't be too bad and it would be interesting to run this at a larger scale

LeilaGhaffari:
@jeremylt , this works fine on CPU but the example runs for ever on HIP. I haven't figured out the problem yet.

jedbrown:
Would you like to finish this branch or close it because the relevant code is in Ratel?

LeilaGhaffari:
I think we decided to close this and migrate the remaining pieces to Ratel. I'll go ahead and close it but would like to keep the branch until it is completely moved to Ratel.

PR #836, opened 2021-11-05
Preferred CITATION isn't an array.

arfon:
:wave: @jedbrown – this should now work properly.


jedbrown (approved):
Aha, thanks! This dialog has the JOSS citation now. When you click `View citation file` while looking at a branch, it still takes you to the version on `main`.

PR #838, opened 2021-11-08
Fix iwyu target

jeremylt:
Ugh, I hate make. Does this work better? It works locally for me.


jeremylt:
This should be good to merge.

jedbrown (approved):
I upgraded to clang-13 and now can't trivially test, but I think it would be ideal that if no environment variables are set and `include-what-you-use` is in the user's path, it would just be used without needing to set environment variables. But not a big deal.

PR #840, opened 2021-11-10
Macro for Debug without Ceed Context

jeremylt:
This is a small change that cleans up our debugging macros and provides a macro for debugging without a ceed context by directly checking for the environment variable.


backends/avx/ceed-avx-blocked.c
```diff
@@ -56,6 +56,8 @@ static int CeedInit_Avx(const char *resource, Ceed ceed) {
 // Backend Register
 //------------------------------------------------------------------------------
 CEED_INTERN int CeedRegister_Avx_Blocked(void) {
-  return CeedRegister("/cpu/self/avx/blocked", CeedInit_Avx, 30);
+  const char prefix[] = "/cpu/self/avx/blocked";
+  CeedDebugEnv("Backend Register: %s", prefix);
```
YohannDudouit, line 60: Why is this code outside `CeedRegister`?

YohannDudouit (dismissed):

backends/avx/ceed-avx-blocked.c
jeremylt: Because `CeedRegister` is not where the weak vs normal registration distinction is made.

YohannDudouit: Are you saying that weak vs normal registration is also handled outside the `CeedRegister` function? Then maybe `CeedRegister` and `CeedWeakRegister`?

jeremylt: `CeedInit` calls `CeedRegisterAll`
`CeedRegisterAll` calls `CeedRegister_*` for each backend
Any backend that is not compiled has a weak link instead to `CeedRegister_Weak`
`CeedRegisterAll` doesn't "know" if it is calling the actual `CeedRegister_*` or `CeedRegister_Weak`.

include/ceed/backend.h
```diff
@@ -66,14 +66,18 @@
 
 /// CEED_DEBUG_COLOR default value, forward CeedDebug* declarations & macros
 #ifndef CEED_DEBUG_COLOR
-#define CEED_DEBUG_COLOR 0
+#define CEED_DEBUG_COLOR 255
```
jedbrown, line 69: Hard-coding this color is not my idea of a good time.
image

backends/avx/ceed-avx-blocked.c
jedbrown: I gotta agree that needing 3 lines to register a backend *and* provide debugging output is not good. We can modify `CeedRegister` to do the debug printing if requested and call `CeedRegisterImpl` for the actual work, then have `CeedRegister_Weak` do the printing and directly call `CeedRegisterImpl`. That way you get debug printing without extra work for the caller.

include/ceed/backend.h
jeremylt: Fixed. I meant to use that as a flag to use the env default since its not a super common color.

jeremylt:
Should be good to merge.

backends/ceed-backend-weak.c
```diff
@@ -13,13 +12,13 @@ static int CeedInit_Weak(const char *resource, Ceed ceed) {
...
 static int CeedRegister_Weak(const char *name, int num_prefixes, ...) {
-  if (getenv("CEED_DEBUG")) fprintf(stderr, "Weak %s\n", name);
-
   va_list prefixes;
   va_start(prefixes, num_prefixes);
   int ierr;
   for (int i=0; i<num_prefixes; i++) {
-    ierr = CeedRegister(va_arg(prefixes, const char*), CeedInit_Weak, CEED_MAX_BACKEND_PRIORITY);
+    const char *prefix = va_arg(prefixes, const char*);
+    CeedDebugEnv("** Weak Register **");
```
jedbrown, line 20: This is okay, though we can't grep the results easily. If you skipped the `\n` in the Weak annotation, it would appear on one line.

jedbrown (approved):

Issue #839, opened 2021-11-08
CUDA/HIP Backend Refactor


jeremylt:
As a result of how they were designed, there is a bunch of code duplication in the CUDA backends, and as a result of this CUDA duplication, the HIP backends inherited this same code duplication.

We should make a PR or series of PRs that is actually designed to refactor and reduce code duplication across these backends.

* What code could/should be combined?

* Where and how do we need to allow for differences between the backends and platforms?

*  Where do we want to test to prevent regression from aggressive or incorrect amalgamation between CUDA and HIP?

* What needs to be done to allow the code generation backends (`gpu/cuda/gen` and `gpu/hip/gen`) to share kernels from the other backends?

@tcew, @jedbrown, any anyone else who I'm missing but is interested, please feel free to jump into this issue or the discussion with thoughts I'm overlooking.

YohannDudouit:
I would unify `hip/ kernels/*` and `cuda/kernels/*` into `gpu_common_kernels/*` (or something similar), currently the code is duplicated. In the future it could be different, Hip/Cuda could have different code, but I think that the design would be better if we abstracted the fact that we target Hip or Cuda architectures. This would allow to test new implementations in a more modular way. We can have different implementations of the same algorithms that live under `common_gpu_kernels`, then which one we decide on using is chosen when loading the source. The implementations could be different for hip and cuda. The purpose being to try different implementations for different scenarios. We could imagine loading different implementations not only based on hip/cuda, but also on the number of quadrature points and degrees of freedom. We already know that there is no best performance approach for all cases, that different implementations work better for different cases, this design could also handle that.
Also, there is implementations that result in less register pressure. In certain applications, register pressure can become an issue when the QFunction gets big, this design would also allow to change the parallelization strategy to accommodate this kind of issues.
This design would potentially allow to fuse `magma` backend kernels too.

On the topic of implementations that would be specific to hip or cuda, I am not aware of such things.
Generalizing code to target either Hip or Cuda is relatively trivial, the architecture specific keywords can easily be abstracted behind macros (`CEED_DEVICE`, `CEED_HOST`, `CEED_HOST_DEVICE`, `CEED_MEM_SHARED`, etc...).

jeremylt:
I think a smaller first step could be refactoring the code generation backends to share the kernels that other backends use. Currently there are some minor differences, but I don't know why those differences were added.

YohannDudouit:
[...]

PR #841, opened 2021-11-10
QFunction multi source support

jeremylt:
This PR adds support for multiple QFunction source files.

See the new `t406-qfunction` for usage.

I think further downstream consequences of the new `CeedLoadSourceToBuffer()` should be put in a downstream PR.

This the bulk of what was in #826, but with some of the possible downstream changes removed. The minor changes to the debugging macros went into #840.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-qfunction.c
+++ b/backends/cuda-gen/ceed-cuda-gen-qfunction.c
@@ -50,72 +50,6 @@ static int CeedQFunctionDestroy_Cuda_gen(CeedQFunction qf) {
 }
 
 //------------------------------------------------------------------------------
-// Load QFunction
-//------------------------------------------------------------------------------
-static int loadCudaFunction(CeedQFunction qf, char *c_src_file) {
-  int ierr;
-  Ceed ceed;
-  CeedQFunctionGetCeed(qf, &ceed);
-  CeedQFunction_Cuda_gen *data;
-  ierr = CeedQFunctionGetData(qf, &data); CeedChkBackend(ierr);
-
-  // Find source file
-  char *cuda_file;
-  ierr = CeedCalloc(CUDA_MAX_PATH, &cuda_file); CeedChkBackend(ierr);
-  memcpy(cuda_file, c_src_file, strlen(c_src_file));
-  const char *last_dot = strrchr(cuda_file, '.');
-  if (!last_dot)
-    return CeedError(ceed, CEED_ERROR_BACKEND, "Cannot find file's extension!");
-  const size_t cuda_path_len = last_dot - cuda_file;
-  strncpy(&cuda_file[cuda_path_len], ".h", 3);
-
-  // Open source file
-  FILE *fp;
-  long lSize;
-  char *buffer;
-  fp = fopen (cuda_file, "rb");
-  if (!fp)
-    // LCOV_EXCL_START
-    return CeedError(ceed, CEED_ERROR_BACKEND,
-                     "Couldn't open the Cuda file for the QFunction.");
-  // LCOV_EXCL_STOP
-
-  // Compute size of source file
-  fseek(fp, 0L, SEEK_END);
-  lSize = ftell(fp);
-  rewind(fp);
-
-  // Allocate memory for entire content
-  ierr = CeedCalloc(lSize+1, &buffer); CeedChkBackend(ierr);
-
-  // Copy the file into the buffer
-  if (1 != fread(buffer, lSize, 1, fp)) {
-    // LCOV_EXCL_START
-    fclose(fp);
-    ierr = CeedFree(&buffer); CeedChkBackend(ierr);
-    return CeedError(ceed, CEED_ERROR_BACKEND,
-                     "Couldn't read the Cuda file for the QFunction.");
-    // LCOV_EXCL_STOP
-  }
-
-  // Append typedef and save source string
-  // FIXME: the magic number 16 should be defined somewhere...
-  char *fields_string =
-    "typedef struct { const CeedScalar* inputs[16]; CeedScalar* outputs[16]; } Fields_Cuda_gen;";
-  ierr = CeedMalloc(1 + strlen(fields_string) + strlen(buffer),
-                    &data->qFunctionSource); CeedChkBackend(ierr);
-  memcpy(data->qFunctionSource, fields_string, strlen(fields_string));
-  memcpy(data->qFunctionSource + strlen(fields_string), buffer,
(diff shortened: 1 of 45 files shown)


backends/cuda-gen/ceed-cuda-gen-qfunction.c
```diff
@@ -126,15 +60,16 @@ int CeedQFunctionCreate_Cuda_gen(CeedQFunction qf) {
...
+  // Read QFunction source
+  ierr = CeedQFunctionGetKernelName(qf, &data->qFunctionName);
+  CeedChkBackend(ierr);
+  ierr = CeedQFunctionLoadSourceToBuffer(qf, &data->qFunctionSource);
+  CeedChkBackend(ierr);
+  if (!data->qFunctionSource)
+    // LCOV_EXCL_START
+    return CeedError(ceed, CEED_ERROR_UNSUPPORTED,
+                     "/gpu/cuda/gen backend requires QFunction source code file");
+  // LCOV_EXCL_STOP
```
YohannDudouit, line 72: Why is this code outside `CeedQFunctionLoadSourceToBuffer`?

YohannDudouit: As I said in the other PR, I would prefer details of this kind to be inside a function.

YohannDudouit (dismissed):

backends/cuda-gen/ceed-cuda-gen-qfunction.c
jeremylt: As I said in the other PR, it is not a problem if no QFunction source file exists for `/gpu/cuda/ref`, `/gpu/cuda/shared`, `/gpu/hip/ref`, and `/gpu/hip/shared`. Only the `/gpu/cuda/gen` and `/gpu/hip/gen` backends cannot accept a user provided `CUfunction`/`hipFunction_t` in place of a source file.

This check is in the only two places where it needs to be.

jeremylt:
This one change should be ready to merge. Future changes related to the design of GPU backends can be captured in the refactor discussed in #839.

doc/sphinx/source/releasenotes.md
```diff
@@ -16,10 +16,14 @@ for each release of libCEED.
 - Promote {c:func} `CeedOperatorCheckReady`to the public API to facilitate interactive interfaces.
 - Warning added when compiling OCCA backend to alert users that this backend is experimental.
 - `ceed-backend.h`, `ceed-hash.h`, and `ceed-khash.h` removed. Users should use `ceed/backend.h`, `ceed/hash.h`, and `ceed/khash.h`.
+- Added {c:func} `CeedQFunctionGetKernelName`; refactored {c:func} `CeedQFunctionGetSourcePath` to exclude function kernel name.
 
 ### New features
 
 - `CeedScalar` can now be set as `float` or `double` at compile time.
+- Added JiT utilities in `ceed/jit-tools.h` to simplify code generation backends.
```
nbeams, line 24: I'm not sure if this first line of the new features in the release notes is completely accurate now (the purpose of the change was not really to simplify code generation backends...), so maybe we should wait until one of the later PRs to mention that?

nbeams (approved):
I'm fine with this as a step to go ahead and add the desired new feature (multi-file QFunction support) while limiting the scope of the PR.  At first I wasn't sure how I felt about elevating the `LoadSourceToBuffer` to the interface level since it's only used for GPU backends, but we already do that in, e.g., `CeedVectorSyncArray`, which is only really needed for GPU backends, as well, so I think it's fine.

doc/sphinx/source/releasenotes.md
jeremylt: Sure, I can edit the message. It sort of simplifies since it collects common code from 4 places to one, but that wasn't the primary purpose.

tests/t406-qfunction.c
```diff
@@ -0,0 +1,75 @@
+/// @file
+/// Test QFunction helper macro
+/// \test Test QFunction helper macro
+#include <ceed.h>
+#include <math.h>
+#include <stdlib.h>
+#include <string.h>
+#include "t406-qfunction-helper.h"
```
jedbrown, line 8:
```suggestion
```
Not used directly so no need to include.

doc/sphinx/source/releasenotes.md
```diff
@@ -16,10 +16,14 @@ for each release of libCEED.
 - Promote {c:func} `CeedOperatorCheckReady`to the public API to facilitate interactive interfaces.
 - Warning added when compiling OCCA backend to alert users that this backend is experimental.
 - `ceed-backend.h`, `ceed-hash.h`, and `ceed-khash.h` removed. Users should use `ceed/backend.h`, `ceed/hash.h`, and `ceed/khash.h`.
+- Added {c:func} `CeedQFunctionGetKernelName`; refactored {c:func} `CeedQFunctionGetSourcePath` to exclude function kernel name.
 
 ### New features
 
 - `CeedScalar` can now be set as `float` or `double` at compile time.
+- Added JiT utilities in `ceed/jit-tools.h` to reduce duplicated code in GPU backends.
+- Added support for JiT of QFunctions with `#include` statements for additional local files.
```
jedbrown, line 25:
```suggestion
- Added support for JiT of QFunctions with `#include "relative/path/local-file.h"` statements for additional local files. Note that files included with `""` are searched relative to the current file first, then by compiler paths (as with `<>` includes). To use this feature, one should adhere to relative paths only, not compiler flags like `-I`, which the JiT will not be aware of.
```

include/ceed/jit-tools.h
```diff
@@ -0,0 +1,26 @@
...
+/// the planning and preparation of a capable exascale ecosystem, including
+/// software, applications, hardware, advanced system engineering and early
+/// testbed platforms, in support of the nation's exascale computing imperative.
+
+#ifndef _ceed_jit_h
+#define _ceed_jit_h
+
+#include <ceed/ceed.h>
+
+#define CEED_MAX_PATH_LEN 1024
```
jedbrown, line 22: This is never used? It'd be best if we can avoid needing to hard-code such a thing. There is `PATH_MAX` and `pathconf` that can be used, but see this BUGS section.
https://www.man7.org/linux/man-pages/man3/realpath.3.html#BUGS

jedbrown (approved):

include/ceed/jit-tools.h
jeremylt: Good catch - holdover from old approach

jeremylt: fixed

PR #842, opened 2021-11-11
SVE Backend

### LINKED ISSUE #631: Arm SVE intrinsics
Arm SVE intrinsics are vector-length agnostic, so represent a nontrivial difference in strategy from Intel intrinsics. An example shows how one might program with these intrinsics:

https://developer.arm.com/documentation/100891/0612/coding-considerations/using-sve-intrinsics-directly-in-your-c-code

Scatters and gathers are supported.

The new Neoverse V1 and N2 will support SVE, so we can expect to see it appearing in more server environments (i.e., not just A64FX).
https://fuse.wikichip.org/news/4564/arm-updates-its-neoverse-roadmap-new-bfloat16-sve-support/

jeremylt:
Closes #631


jeremylt:
This now compiles and passes the t3 tests on Ookami.

ToDo:
- [ ] Performance comparisons
- [ ] Makefile flag fix (not sure how the AVX one even works)
- [ ] Improved vectorization instructions?
- [ ] Do we want manual unrolling for ref? The unrolling is pretty straightforward

backends/opt/ceed-opt-tensor.c
```diff
@@ -0,0 +1,142 @@
...
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+#include <ceed/ceed.h>
+#include <ceed/backend.h>
+#include "ceed-opt.h"
+
+//------------------------------------------------------------------------------
+// Tensor Contract Blocked
+//------------------------------------------------------------------------------
+int CeedTensorContractApply_Blocked_Opt(CeedTensorContract contract, CeedInt A,
```
jeremylt, line 24: Interestingly, this seems to run worse with `mpiexec ./build/petsc-bps -ceed /cpu/self/opt/blocked`, with a peak near `DoFs/Sec in CG: 12.2369 (12.2369) million` vs `DoFs/Sec in CG: 16.7052 (16.7052) million` for the `ref` tensor

The SVE performance in previous commits seems similar to this unrolling

backends/opt/ceed-opt-tensor.c
```diff
@@ -0,0 +1,142 @@
...
+          v[(a*J+j)*C+(c+cc)] += tq * u[(a*B+b)*C+(c+cc)];
+      }
+
+  return CEED_ERROR_SUCCESS;
+}
+
+//------------------------------------------------------------------------------
+// Tensor Contract Serial
+//------------------------------------------------------------------------------
+int CeedTensorContractApply_Serial_Opt(CeedTensorContract contract, CeedInt A,
```
jeremylt, line 53: While here this does better than the ref tensor contract `DoFs/Sec in CG: 5.69806 (5.69806) million` vs `DoFs/Sec in CG: 1.30158 (1.30158) million`

backends/opt/ceed-opt-tensor.c
jeremylt: Specializing on `t_mode` doesn't seem to have much of an effect

LeilaGhaffari:
I remember from Ookami's talk that performance with GCC was the worst among all compilers. I did a brief experiment before losing access to Ookami.
`sve/blocked` with `armclang`: `DoFs/Sec in CG: 1.43289 (1.43289) million`
`opt/blocked` with `gcc`:  `DoFs/Sec in CG: 0.729637 (0.729637) million`

I am not sure why these numbers are so small compared to what we had Friday but I think the poor performance with sve might partly have something to do with compiler, maybe?! I have to apply for an account to do more experiments though.

jedbrown:
Is this ready for review? Should we include it in v0.11?

jeremylt:
The two big todos are fixing the makefile magic and seeing if this actually does anything different than OPT in terms of performance.

jedbrown:
I noticed that libxsmm contains aarch64/SVE code and it's announced as supported for the next release.

jedbrown:
Do we have a place where we can measure performance? There is a machine at Sandia that you can access if you put in a Sarape request and AWS c7g also has SVE. JLSE also has a system that I could try requesting.

sebastiangrimberg:
Hi @jedbrown @jeremylt, if there is any way I can help here let me know. Happy to take the time to test on AWS's c6g/c7g for you if there's a specific case you're interested in, also against whatever libxsmm has for Arm optimizations as well.

jeremylt:
I haven't had time to close the loop on this one. The big two todos are fixing the makefile detection of SVE support and running some basic performance comparisons to the /CPU/self/opt backends. I'm not sure the best way to do the first but we have a script to run our PETSc examples for the second.

sebastiangrimberg:
Sounds good, no rush. I'd be very happy to run the PETSc benchmarks on whatever instance types you want whenever it is ready, if that would be helpful to you.

jeremylt:
Side note - finally got the compile flags right to test in GitHub Actions. Yay. But the test is failing. Boo.

jrwrigh:
This was well before my time, so I'm not 100% on the motivation and scope of the PR. But that logic makes sense to me.

jedbrown:
I think XSMM did not target SVE at the time when this work was started. If XSMM covers the scope of what we might use SVE intrinsics for, I don't have a problem sticking with that and just closing this.

PR #844, opened 2021-11-12
Opt - specalize tensor product for C=1 case

jeremylt:
Initial tests on Ookami seem to indicate that this is helpful for the `opt/serial` backend.

### DIFF
--- a/backends/opt/ceed-opt-tensor.c
+++ b/backends/opt/ceed-opt-tensor.c
@@ -1,0 +1,89 @@
+// Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory. LLNL-CODE-734707.
+// All Rights reserved. See files LICENSE and NOTICE for details.
+//
+// This file is part of CEED, a collection of benchmarks, miniapps, software
+// libraries and APIs for efficient high-order finite element and spectral
+// element discretizations for exascale applications. For more information and
+// source code availability see http://github.com/ceed.
+//
+// The CEED research is supported by the Exascale Computing Project 17-SC-20-SC,
+// a collaborative effort of two U.S. Department of Energy organizations (Office
+// of Science and the National Nuclear Security Administration) responsible for
+// the planning and preparation of a capable exascale ecosystem, including
+// software, applications, hardware, advanced system engineering and early
+// testbed platforms, in support of the nation's exascale computing imperative.
+
+#include <ceed/ceed.h>
+#include <ceed/backend.h>
+#include "ceed-opt.h"
+
+//------------------------------------------------------------------------------
+// Tensor Contract Core loop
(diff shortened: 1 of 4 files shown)


jeremylt:
I feel like this is a small and easy enough change that we can just go ahead and merge it before we finish tinkering with SVE.

backends/opt/ceed-opt-tensor.c
```diff
@@ -0,0 +1,113 @@
...
+    t_stride_0 = 1; t_stride_1 = J;
+  }
+
+  for (CeedInt a=0; a<A; a++)
+    for (CeedInt b=0; b<B; b++)
+      for (CeedInt j=0; j<J; j++)
+        v[a*J+j] += t[j*t_stride_0 + b*t_stride_1] * u[a*B+b];
+
+  return CEED_ERROR_SUCCESS;
+}
```
jedbrown, line 66: Why is this written separately instead of as a specialization of the above with static `C=1`? That should allow the compiler to eliminate the loop.

Wouldn't we want to specialize this implementation for some static sizes like we do with the AVX backend? We could do more setup-time specialization if we create different tensor contraction objects for each different size. The joy of `blocked` is that we don't need that so long as `C` is a multiple of the vector length.

jedbrown (approved):

backends/opt/ceed-opt-tensor.c
jeremylt: Good point about writing it separately. I updated the commit.

At least in my tests on Ookami, it seems like manually unrolling multiples of 8 for  didn't help much.
https://github.com/CEED/libCEED/pull/842/commits/c9881c54a31ae6782ce35137d099b39f6a1a9c51

jedbrown: We should examine the assembly. You could do this on Compiler Explorer -- [example](https://gcc.godbolt.org/z/EvGE3Yx5c). Note that the default `-msve-vector-bits=scalable` gives somewhat different code (and should probably be preferred), but it's worth being able to interpret the inner loop both ways.

Also, A64FX will stay pretty bespoke. I would use `-march=armv9 -mtune=neoverse-v1` or `neoverse-n2` since those are more representative of what might be widely available in the next year or so. The available optimizations should be similar.

jeremylt: This current change on Noether results in ~2.5 higher throughput for `/cpu/self/opt/serial`. We still get better performance with `/cpu/self/avx/serial`, but I feel like this is a good change to merge.
```
make prove -j search=pet OPT="-O3 -march=native -funsafe-math-optimizations -ffp-contract=fast"
./build/petsc-bps -ceed /cpu/self/opt/serial
```
I saw ~13 million DoFs/Sec on main vs ~30 million DoFs/Sec on this PR for that limited run of sizes.

I'll go ahead and merge this small change and we can continue with future improvements.

PR #846, opened 2021-11-18
doc - fix minor spacing bug in release notes

jeremylt:
Anyone can review - this fixes a small bug in the release notes that prevents correct rendering.

Before:
image

After:
image


LeilaGhaffari (approved):
Nice catch!

PR #849, opened 2021-11-29
petsc - use PetscFECreateLagrange

jeremylt:
I just discovered `PetscFECreateLagrange` - I don't remember this function being in PETSc when I first set up the multigrid example.


jedbrown (approved):
Yeah, it first appeared in PETSc 3.13 and we wanted compatibility when it first came out, but it's been plenty long enough to use it now.
https://gitlab.com/petsc/petsc/-/commit/e703855ddb8d48a97309a309e1282b0ee453d443

PR #850, opened 2021-11-29
Slight modifications for hiprtc usage in ROCm 4.5

nbeams:
Some minor changes are required to use hiprtc with ROCm 4.5.  I added a check so that ROCm < 4.5 will still work.

See [ROCm release notes](https://rocmdocs.amd.com/en/latest/Current_Release_Notes/Current-Release-Notes.html), particularly:

>  Applications can no longer explicitly include HIP runtime header files. Minor code changes are required to remove the HIP runtime header files.

Edit: if @moes1 has a chance to look at this as well, that would be great.

### DIFF
--- a/backends/hip/ceed-hip-compile.cpp
+++ b/backends/hip/ceed-hip-compile.cpp
@@ -39,9 +39,21 @@ int CeedCompileHip(Ceed ceed, const char *source, hipModule_t *module,
   hipFree(0); // Make sure a Context exists for hiprtc 
   hiprtcProgram prog;
 
-  // Add hip runtime include to string for generation
   std::ostringstream code;
-  code << "\n#include <hip/hip_runtime.h>\n";
+  // Add hip runtime include statement for generation if runtime < 40400000
+  // (implies ROCm < 4.5)
+  int runtime_version;
+  CeedChk_Hip(ceed, hipRuntimeGetVersion(&runtime_version));
+  if (runtime_version < 40400000) {
+    code << "\n#include <hip/hip_runtime.h>\n";
+  }
+  // With ROCm 4.5, need to include these definitions specifically for hiprtc
+  // (but cannot include the runtime header)
+  else {
+    code << "#include <stddef.h>\n";
+    code << "#define __forceinline__ inline __attribute__((always_inline))\n";
+    code << "#define HIP_DYNAMIC_SHARED(type, var) extern __shared__ type var[];\n";
+  }
 
   // Macro definitions
   // Get kernel specific options, such as kernel constants


backends/hip/ceed-hip-compile.cpp
```diff
@@ -39,9 +39,18 @@ int CeedCompileHip(Ceed ceed, const char *source, hipModule_t *module,
...
   std::ostringstream code;
-  code << "\n#include <hip/hip_runtime.h>\n";
+  // Add hip runtime include to string for generation if runtime < 40400000
+  // (implies ROCm < 4.5)
+  int runtime_version;
+  CeedChk_Hip(ceed, hipRuntimeGetVersion(&runtime_version));
+  if (runtime_version < 40400000)
+    code << "\n#include <hip/hip_runtime.h>\n";
+  else { // With ROCm 4.5, need to include these specifically for hiprtc
+    code << "#define NULL 0\n";
```
nbeams, line 50: Not sure if this is the preferred choice for defining NULL...

backends/hip/ceed-hip-compile.cpp
```diff
@@ -39,9 +39,18 @@ int CeedCompileHip(Ceed ceed, const char *source, hipModule_t *module,
...
   hiprtcProgram prog;
 
-  // Add hip runtime include to string for generation
   std::ostringstream code;
-  code << "\n#include <hip/hip_runtime.h>\n";
+  // Add hip runtime include to string for generation if runtime < 40400000
+  // (implies ROCm < 4.5)
+  int runtime_version;
+  CeedChk_Hip(ceed, hipRuntimeGetVersion(&runtime_version));
+  if (runtime_version < 40400000)
```
jeremylt, line 47: Minor style preference - in an if -else, when at least one of the two body statements requires {}s, I generally prefer to guard both bodies with {}s. It is just easier for me to read, personally.

backends/hip/ceed-hip-compile.cpp
jeremylt: Do you know which kernels use `NULL`? I think it would be nice to remove `NULL` where able.

nbeams: From the basic test suite, at least, it's just two lines in the operator diagonal assembly:

https://github.com/CEED/libCEED/blob/239c65026b8097e4437d6328197c2f08879797b1/backends/hip/ceed-hip-operator.c#L781
and
https://github.com/CEED/libCEED/blob/239c65026b8097e4437d6328197c2f08879797b1/backends/hip/ceed-hip-operator.c#L788

Do you have a preference for what to do with those?  We could just set them to 0 instead of NULL, or maybe leave them uninitialized (this works for the tests but not sure if there are any potential problem cases from doing that).

jeremylt: It would be fine to leave them as uninitialized since it is immediately set by the next function call. I set those two pointers to `NULL` to make `clang-tidy` happy  for the CPU version of this function.

Though, would users expect to be able to use `NULL` in QFunctions?

nbeams: > Though, would users expect to be able to use NULL in QFunctions?

That's a good point.  They might be a bit surprised to find that it doesn't work with HIP but does with CUDA...

jedbrown: I'm in favor of defining it so we don't make users jump through the awkward hoops. Maybe we can report it to hiprtc to fix in a future release.

backends/hip/ceed-hip-compile.cpp
nbeams: I added the {} (and tweaked the comments) in 682583ce514e65ab3543cf99b41ac809995991cd

backends/hip/ceed-hip-compile.cpp
moes1: @nbeams I have communicated several examples of issues caused by this change to hiprtc internally. 

For now it might be better to include these files that are typically included in hip_runtime.h instead of defining NULL directly yourself (I believe it's defined in stdlib.h):

`#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>
`

jedbrown: Yes, good point. By include-what-you-use, these should be included by the *.h files anyway, since we now support that.

jeremylt: Though, we strip out installed headers from user QFunction source, so we would want to include it here in `ceed-hip-compile.cpp`.

nbeams: The problem with including `stdlib` directly is that I get errors like this:
` error: typedef redefinition with different types ('__int64_t' (aka 'long') vs 'long long')`

Have you not run into that with any tests?

jedbrown:
Is this ready to merge? I'd need to log into Spock/Crusher to test with ROCm 4.5 and I'm happy to take your word that it works there.

backends/hip/ceed-hip-compile.cpp
jedbrown: Can you `#include <stddef.h>`?

jedbrown (approved):

nbeams:
Spock didn't have 4.5 yet when I opened this PR.  I can test with the latest ROCm 4.5 module tomorrow.  I will also see about including stddef.

backends/hip/ceed-hip-compile.cpp
nbeams: Good call, including `stddef` works without causing any of the errors from `stdlib`.  I have changed that line in  4637df7

nbeams:
I have tested this version of the branch with ROCm 4.5.0 and 4.5.2 (and also with 4.0 on Noether).  It should be good to go.

jedbrown:
Thanks!

Issue #755, opened 2021-04-23
Fluids - SU/SUPG broken in DC


LeilaGhaffari:
While refactoring the fluids example, I noticed that `SU/SUPG` stabilization methods are broken in `density_current` and they don't converge.

PR #851, opened 2021-12-06
Shock Tube + YZB Discontinuity Capturing

tt-aiken:
## Overview 
New example within /examples/fluids that's based on the [Sod Shock Tube](https://en.wikipedia.org/wiki/Sod_shock_tube) problem. This problem includes YZB discontinuity capturing and SU stabilization for the explicit formulation of the Euler equations. These features were not included for the implicit solver, which produces highly oscillatory solutions as a result. 

The problem can be run with the following command:
`./navierstokes -problem shocktube -yzb -stab su -bc_slip_z 3,4 -bc_slip_y 1,2 -bc_wall 5,6 -dm_plex_dim 3 -dm_plex_box_lower 0,0,0 -dm_plex_box_upper 1000,100,100 -dm_plex_box_faces 200,1,1 -units_second 0.1 `

## YZB detail
The YZB formulation was implemented following the first formulation presented in [this paper](https://www.tafsm.org/PUB_PRE/jALL/j130-CF-SHOC.pdf), which was preferred over the second formulation because of its simplicity. Implementation required a slight modification from the originally published form, since the real space gradient of the shape function is not available at the QFunction level. This gradient, which is used in the calculation of an element lengthscale, is replaced in the current formulation with a function of the real->reference space Jacobian. The exact formulation was proposed by @KennethEJansen and is shown below:

### Original length scale:
Screen Shot 2021-12-06 at 11 07 43 AM

### Modified length scale:
Screen Shot 2022-01-30 at 12 19 19 PM

## Other implementation details:
* C<sub>YZB</sub> is set to 0.1, based on testing
* c<sub>v</sub> for the SU scheme is set to 50.0, again based on testing since the formulation is nondimensional. Larger values produce more dissipation by increasing the magnitude of the energy equation stabilization term.
* Reference velocity for YZB is the acoustic speed
* Reference density for YZB is the local density 
* Convective flux Jacobians for the SU formulation came from [this page](https://www.theoretical-physics.net/dev/fluid-dynamics/euler.html#:~:text=then%20we%20can%20calculate%20the%20Jacobians). The Jacobian from `densitycurrent.h` was not used since that SU implementation is [broken](https://github.com/CEED/libCEED/issues/755).
* Aside from the SU and YZB implementations, the RHS QFunction is unchanged from that used in `eulervortex.h`. The form used in that example is still the default, with the current upgrades requiring the additional command line arguments `-yzb` and  `-stab su`
* `-units_second` sometimes requires tuning based on mesh density. Perhaps a different time integrator could be more robust and willing to adapt more aggressively? Or, this is just a sign that implicit implementation should be prioritized.


examples/fluids/navierstokes.h
```diff
@@ -216,6 +216,11 @@ struct SetupContext_ {
   CeedScalar dc_axis[3];
   CeedScalar wind[3];
   CeedScalar time;
+  CeedScalar midpt;
```
LeilaGhaffari, line 219: Nice work, @tt-aiken ! 
What is `midpt`? Is it `midpoint`?

examples/fluids/problems/shocktube.c
```diff
@@ -0,0 +1,246 @@
...
+  PetscFunctionBeginUser;
+
+  // Default boundary conditions
+  //    slip BC on y and z faces
+  bc->num_slip[0] = 0;
+  bc->num_slip[1] = bc->num_slip[2] = 4;
+  bc->slips[1][0] = bc->slips[2][0] = 1;
+  bc->slips[1][1] = bc->slips[2][1] = 2;
+  bc->slips[1][2] = bc->slips[2][2] = 3;
+  bc->slips[1][3] = bc->slips[2][3] = 4;
```
LeilaGhaffari, line 192: @tt-aiken , I don't understand how the BCs are specified. If you want to set slip BCs on y and z directions (based on your comment), the following should do the job.

```
  bc->num_slip[1] = bc->num_slip[2] = 2;
  bc->slips[1][0] = 3;
  bc->slips[1][1] = 4;
  bc->slips[2][0] = 1;
  bc->slips[2][1] = 2;
```

Is there a reason you have considered 4 faces for each direction?

examples/fluids/navierstokes.h
tt-aiken: Thanks! 
Yes, this is midpt. Probably better to just use the full word for clarity.

examples/fluids/problems/shocktube.c
tt-aiken: I see that setting slip BC for y-velocity on the y-face is redundant, and same with z-velocities on z-faces - two faces should do the trick. However, from my understanding of the face numbering convention, this would look like:

```
bc->num_slip[1] = bc->num_slip[2] = 2;
bc->slips[1][0] = 1;
bc->slips[1][1] = 2;
bc->slips[2][0] = 3;
bc->slips[2][1] = 4;
```

These PETSc BC assignment routines are a bit opaque to me though. Am I missing something here?

LeilaGhaffari: Right, it took me a while to understand the numbering convention. Hmm, faces `1` and `2` are in the `z` direction, `3` and `4` in the `y` direction, and `5` and `6` in the `x` direction. So, the one that I wrote above should be correct. Just like what we have in [densitycurrent](https://github.com/CEED/libCEED/blob/239c65026b8097e4437d6328197c2f08879797b1/examples/fluids/problems/densitycurrent.c#L283).

examples/fluids/navierstokes.h
LeilaGhaffari: Yeah, `midpoint`, `mid_point`, or even `mid_pt` are easier to read, IMO.

examples/fluids/problems/shocktube.c
knepley: The numbering is from here: https://gitlab.com/petsc/petsc/-/blob/main/src/dm/impls/plex/plexinterpolate.c#L177 . The z faces come first so that the closure of an interpolated mesh produces the same vertex order as the uninterpolated mesh.

tt-aiken: Thank you both, I'll change this now.

examples/fluids/qfunctions/shocktube.h
```diff
@@ -0,0 +1,510 @@
...
+    switch (context->stabilization) {
+    case 0:        // Galerkin
+      break;
+    case 1:        // SU
+      for (int j=0; j<3; j++)
+        for (int k=0; k<5; k++)
+          for (int l=0; l<5; l++) {
+            stab[k][j] = jacob_F_conv_T[j][k][l] * Tau[l] * strong_conv[l];
+            // printf("strong_conv[%d], Tau[%d] = %1.5e, %1.5e \n",l,l,strong_conv[l],Tau[l]);
+            // printf("stab[%d][%d] = %1.5e \n",k,j,stab[k][j]);
```
jeremylt, line 406: Don't forget to tidy up dead code!

examples/fluids/qfunctions/shocktube.h
```diff
@@ -0,0 +1,510 @@
...
+        dv[j][4][i] -= wdetJ * nu_shoc * dEdx[j];
+    }
+
+    // Stabilization
+
+    // Need the Jacobian for the advective fluxes for stabilization
+    //    indexed as: jacob_F_conv[direction][flux component][solution component]
+    CeedScalar jacob_F_conv[3][5][5] = {{{0.}}};
+    CeedScalar velsq = u[0]*u[0] + u[1]*u[1] + u[2]*u[2];
+    CeedScalar rhosq = rho*rho;
```
jeremylt, line 336: libCEED's convention is to use full words in general, which underscores between them, so something like `vel_sq` or `velocity_sq`

examples/fluids/qfunctions/shocktube.h
```diff
@@ -0,0 +1,510 @@
...
+
+    // -- YZB stabilization
+    if (context->yzb) {
+      CeedScalar drho_norm = 0.0;   // magnitude of the density gradient
+      CeedScalar j_vec[3] = {0.0};  // unit vector aligned with the density gradient
+      CeedScalar jdgn[3] = {0.0};   // j * grad(N)
+      CeedScalar h_shoc = 0.0;      // element lengthscale
+      CeedScalar u_cha = 0.0;       // characteristic velocity, set to acoustic speed
+      CeedScalar tau_shoc = 0.0;    // timescale
+      CeedScalar nu_shoc = 0.0;     // artificial diffusion
```
jeremylt, line 295: why `shoc` instead of `shock` and why not `j_gradn`? The idea is that people less familiar with fluids should still be able to use this example to help them understand how to write complex physics in QFunctions.

jeremylt:
I'm not really familiar with this model, but I included a few comments where I think names can be clearer. I see our examples as tutorials on using our codebase, so we've been moving towards clearer variable names in our code.

examples/fluids/qfunctions/shocktube.h
```diff
@@ -300,14 +301,14 @@ CEED_QFUNCTION(EulerShockTube)(void *ctx, CeedInt Q,
       // Approximate dot(j_vec,grad(N)) using the metric tensor
       for (int j=0; j<3; j++)
         j_gradn[j] = j_vec[0] * dXdx[0][j]
-                   + j_vec[1] * dXdx[1][j]
-                   + j_vec[2] * dXdx[2][j];
+                     + j_vec[1] * dXdx[1][j]
+                     + j_vec[2] * dXdx[2][j];
```
LeilaGhaffari, line 305: This is more consistent with other terms and I think it's easier to read.
```
      for (int j=0; j<3; j++)
        j_gradn[j] = j_vec[0] * dXdx[0][j] +
                     j_vec[1] * dXdx[1][j] +
                     j_vec[2] * dXdx[2][j];
```

examples/fluids/qfunctions/shocktube.h
```diff
@@ -287,7 +287,8 @@ CEED_QFUNCTION(EulerShockTube)(void *ctx, CeedInt Q,
       CeedScalar j_vec[3] = {0.0};        // unit vector aligned with the density gradient
       CeedScalar j_gradn[3] = {0.0};      // j * grad(N)
       CeedScalar h_shock = 0.0;           // element lengthscale
-      CeedScalar acoustic_vel = 0.0;      // characteristic velocity, set to acoustic speed
+      CeedScalar acoustic_vel =
+        0.0;      // characteristic velocity, set to acoustic speed
```
LeilaGhaffari, line 291: `make style` sometimes does weird things and we probably need to fix some of them manually.
how about
```
CeedScalar acoustic_vel = 0.0;      
// characteristic velocity, set to acoustic speed
```

examples/fluids/qfunctions/shocktube.h
```diff
@@ -0,0 +1,510 @@
...
+#endif
+
+// *****************************************************************************
+// This function sets the initial conditions
+//
+//   Temperature:
+//     T   = P / (rho * R)
+//   Density:
+//     rho = 1.0        if x <= midpt
+//         =            if x >  midpt
```
LeilaGhaffari: 0.125?

LeilaGhaffari:
Also, I think it would be helpful to add a regression test like what we have for other problems [here](https://github.com/CEED/libCEED/blob/4a63c7da274c5bcaedf5323dad2ae658837a33b1/examples/fluids/navierstokes.c#L34).

examples/fluids/qfunctions/shocktube.h
tt-aiken: This is what happened for the above comment on `j_gradn[j]`. I'll go through and clean these up.

tt-aiken:
Thanks for the further comments @LeilaGhaffari. I'll be offline for most of the next three weeks but will definitely be making these changes once I'm back so this PR can get merged.

LeilaGhaffari (changes requested):

LeilaGhaffari:
> Thanks for the further comments @LeilaGhaffari. I'll be offline for most of the next three weeks but will definitely be making these changes once I'm back so this PR can get merged.

No problem. Sorry I have been slow in the review. I was trying to compare your impl of stabilization with the one in `density_current`. It seems that the only difference is in the flux Jacobian. In fact, your implementation helped me find a couple of bugs in `density_current`. I believe three terms in shocktube need to change as well. I checked them with the source you provided us with and @KennethEJansen 's lecture notes. I might be wrong though. Could you please double-check those terms when you are back? Thanks!

examples/fluids/qfunctions/shocktube.h
```diff
@@ -0,0 +1,505 @@
...
+    const CeedScalar uiujgij = uX[0]*uX[0] + uX[1]*uX[1] + uX[2]*uX[2];
+    const CeedScalar Cc      = 1.;
+    const CeedScalar Ce      = 1.;
+    const CeedScalar f1      = rho * sqrt(uiujgij);
+    const CeedScalar TauC    = (Cc * f1) /
+                               (8. * (dXdxdXdxT[0][0] + dXdxdXdxT[1][1] + dXdxdXdxT[2][2]));
+    const CeedScalar TauM    = 1. / (f1>1. ? f1 : 1.);
+    const CeedScalar TauE    = TauM / (Ce * cv);
+
+    const CeedScalar Tau[5]  = {TauC, TauM, TauM, TauM, TauE};
```
jedbrown, line 392: Could you share a citation for this stabilization formulation? Is it intended for conservative variables? The units are different from what I would expect.

jedbrown:
Welcome back @tt-aiken. We worked out some things in the just-merged #865, including bypassing the tau definition here (which has inconsistent units when used to stabilize conservative variables). You may want to rebase and read the updated docs https://libceed.readthedocs.io/en/latest/examples/fluids/#equation-eq-tau-conservative (in the "Stabilization scale \tau" dropdown) 

I'll be curious to see how this SUPG variant works with your shock capturing scheme. We also have a thermal bubble experiment (see `test_5`) that would be interesting to try with shock capturing turned on. One thing I'd ask is that you add a section to `examples/fluids/index.md` to describe the formulation. You can put paper citations in `doc/sphinx/source/references.bib`.

tt-aiken:
Thanks for all the comments @jedbrown and @LeilaGhaffari! Sorry for the delay on my end, I'm finally back to normal life and am excited to help push this over the line.

It sounds like I have three things to do:
- Update SU stabilization based on the formulation for tau
- Document the shock capturing operator implementation
- Fix those three terms in the flux Jacobian

Am I missing anything?

jedbrown:
I think that's it, thanks.

examples/fluids/qfunctions/shocktube.h
tt-aiken: `make style` consistently yells at me about this. I don't know what I should be doing to make it accept the formatting. Do you have any ideas?

jeremylt: You can add style exclusion tags.

LCOV_EXCL_START 
And
LCOV_EXCL_STOP

jeremylt: You can also put the + on the preceding line.

jedbrown: I don't recall astyle being able to handle the `+` at the beginning of the line, but I think it's easier to read that way. I'd just ignore style here, or write it as a loop. Maybe we can use clang-format some day.

tt-aiken: I'll move the + sign, I missed that detail in Leila's original suggestion

tt-aiken:
I've gone through and done almost everything discussed in this thread so far. There are two points that I need a bit of help to finish:

- Documentation. Is there a way I can see readthedocs.io with my updates? It would be great to see what I've added in the actual compiled documentation.
- How can I add test output to a .bin file in /fluids/examples/test-output? I wrote the regression test line in `navierstokes.c` but there is currently no result file to compare with.

jedbrown:
You can write the result `-vec_view that-file.bin` and then run with the comparison flag.

jedbrown:
The rendered documentation can be viewed in the checks, or you can `make doc` locally. You'll need `pip install -r doc/sphinx/requirements.txt` before running `make doc` so you'll have the Python toolchain.
image

LeilaGhaffari:
>     * How can I add test output to a .bin file in /fluids/examples/test-output? I wrote the regression test line in `navierstokes.c` but there is currently no result file to compare with.

Thanks, @tt-aiken for the hard work. 
I usually create the *.bin files as following:
1. Run the test with all run time flags that I would normally use + `-ts_max_steps 10`. If you look at other tests, you can see that we avoid large mesh sizes for regression tests.
2. Rename the generated solution file, `ns-solution.bin`, and move it to `tests-output/`.
3. Add the test arguments to [navierstokes.c](https://github.com/CEED/libCEED/blob/3c6176c9ba1a3cbcebb4b8e7be9c58288fd240e9/examples/fluids/navierstokes.c#L44). You would need to add `//TESTARGS(name="some-name") -ceed {ceed_resource} -test -compare_final_state_atol some-tol -compare_final_state_filename examples/fluids/tests-output/your-new-test.bin` to the arguments and drop `-ts_max_steps 10`.

This should do the job. Let me know if I can help with that.

jedbrown:
@tt-aiken Just checking in if you need anything from us to wrap this up. It looks close to me.

tt-aiken:
Thanks for the heads up here. I'm working to get this totally wrapped up today, sorry for the delay.

tt-aiken:
Two questions: 
1) The pipeline failure happens because some `.junit` and `.json` files are not found. Is this related to anything I have done?
2) The docs build is failing during `__init__.py`, and I cannot figure out what's happening for the life of me. I've removed all of my changes to the docs and it still crashes. What could I be missing here?

jeremylt:
We have changed the CI environment. I recommend rebaseing off main to get those changes into this branch.

tt-aiken:
One more hurdle: Where can I find PETSc 3.17? I can't find the tarball anywhere, and the most recent branch among `release` and `main` is 3.16.4.

jedbrown:
We've been using `main` recently because some new features make our lives a lot easier. 3.17 will be released next month. Are you okay to use `main`?

tt-aiken:
All good with the `main` build of PETSc. Hopefully in the final stretches here! 

Were the BCs generalized in the fluids examples? I see that my example is the only one that still uses `DMAddBoundary`. The BCs for the other problems seem to be handled in `setupdm.c`. My problem is currently crashing and I think this might be to blame.

jedbrown:
Yeah, they were moved while @LeilaGhaffari was refactoring. I think her or I can help after our SIAM talks tomorrow, or maybe you can see what's needed to fix your current crashing.

LeilaGhaffari:
Sure, I can fix it today.

LeilaGhaffari:
@tt-aiken , I think you forgot to commit your test output file. Could you commit it so I could check whether my changes help to pass your test and then push?

tt-aiken:
@LeilaGhaffariVSCode inserted a bunch of weird characters during rebasing and I had to hard reset to a previous commit - losing the output in the process. I can't make a new test output because it doesn't run now that I have rebased.

tt-aiken:
I'm going to look over some of the new additions in the other example files and will try to figure out the crashes myself this morning. I see there are a few new things to do with the mesh object that I haven't included in the `shocktube` files yet.

LeilaGhaffari:
I've already fixed the issue [here](https://github.com/CEED/libCEED/commit/30eecbc59a7614d43cc2f2a7df5cfe7935fa6810). You could use this commit or just push your test file so that I could test my changes with it.

LeilaGhaffari:
Sorry I missed your first comment. Let me see if I can create a new test.

jedbrown:
If the file was ever committed, you can use `git reflog` to find a commit that included the file and `git checkout THATHASH -- file/name.log` to recover it.

LeilaGhaffari:
All tests are passing locally. I rebased the branch with main as well. The regression tests in navierstokes.c weren't updated with the main branch and that caused all tests to fail.

tt-aiken:
@LeilaGhaffari, thank you! The problem runs just fine now. Fixing a few small things that have come up with the new formulation and then everything should be good to go.

LeilaGhaffari:
No problem, @tt-aiken . Though it would be great for sanity check if you could find your primary test file and compare/replace it with the one that I just created.

tt-aiken:
Everything checks out except the `noether-rocm` test, which seems to have failed on other PRs and may not be related to my changes here.

LeilaGhaffari:
I don't know what had happened previously but the commits were doubled (some code were repeated). I got rid of those repetitions and rebased onto main. You can checkout this version with `git reset --hard origin/tim/yzb-shocktube` to avoid further conflicts. Tests are passing locally but I am not sure what is going on with the CI.

examples/fluids/index.md
```diff
@@ -351,6 +351,40 @@ $$
...
 There is no perturbation in the entropy $S=P/\rho^\gamma$ ($\delta S=0)$.
 
+(problem-shock-tube)=
+
+## Shock Tube
+
+This test problem is based on Sod's Shock Tube (from{cite}`sodshocktubewiki`), a canonical test case for discontinuity capturing in one dimension. For this problem, the three-dimensional Euler equations are formulated exactly as in the Isentropic Vortex problem. The default initial conditions are $P=1$, $\rho=1$ for the driver section and $P=0.1$, $\rho=0.125$ for the driven section. The initial velocity is zero in both sections. Slip boundary conditions are applied to the side walls and wall boundary conditions are applied at the end walls.
+
+SU upwinding and discontinuity capturing have been implemented into the explicit timestepping operator for this problem. Discontinuity capturing is accomplished using a modified version of the YZB operator described in {cite}`tezduyar2007yzb`. This discontinuity capturing scheme involves the introduction of a dissipation term of the form
```
jedbrown, line 360:
```suggestion
SU upwinding and discontinuity capturing have been implemented into the explicit timestepping operator for this problem. Discontinuity capturing is accomplished using a modified version of the $YZ\beta$ operator described in {cite}`tezduyar2007yzb`. This discontinuity capturing scheme involves the introduction of a dissipation term of the form
```

examples/fluids/index.md
```diff
@@ -351,6 +351,40 @@ $$
...
+
+$$
+\nu_{SHOCK} = \tau_{SHOCK} u_{cha}^2
+$$
+where,
+$$
+\tau_{SHOCK} = \frac{h_{SHOCK}}{2u_{cha}} \left( \frac{ \,|\, \nabla \rho \,|\, h_{SHOCK}}{\rho_{ref}} \right)^{\beta}
+$$
+
+$\beta$ is a tuning parameter set between 1 (smoother shocks) and 2 (sharper shocks. The parameter $h_{SHOCK}$ is a length scale that is proportional to the element length in the direction of the density gradient unit vector. This density gradient unit vector is defined as $\hat{\bm j} = \frac{\nabla \rho}{|\nabla \rho|}. The original formulation of Tezduyar and Senga relies on the shape function gradient to define the element length scale, but this gradient is not available to qFunctions in libCEED. To avoid this problem, $h_{SHOCK}$ is defined in the current implementation as
```
jedbrown, line 376: Covariant and contravariant length measures are discussed at the top of the "stabilization scale \tau" drop-down here:
https://libceed.readthedocs.io/en/latest/examples/fluids/#equation-eq-peclet

John Evans says he uses the covariant measure and it's more common among former Hughes students. That's what we use in SU/SUPG.

examples/fluids/navierstokes.h
```diff
@@ -249,6 +254,62 @@ struct EulerContext_ {
...
+struct ShockTubeContext_ {
+  CeedScalar cv;
+  CeedScalar Cyzb;
+  CeedScalar Byzb;
+  bool implicit;
+  bool yzb;
+  int stabilization;
+};
+#endif
+
```
jedbrown, line 312: These duplicates must not be used
```suggestion
```

examples/fluids/qfunctions/shocktube.h
```diff
@@ -0,0 +1,527 @@
...
+// Model from:
+//   On the Order of Accuracy and Numerical Performance of Two Classes of
+//   Finite Volume WENO Schemes, Zhang, Zhang, and Shu (2011).
+
+#ifndef shocktube_h
+#define shocktube_h
+
+#ifndef __CUDACC__
+#  include <math.h>
+#endif
```
jedbrown, line 30: It's safe to just include these days.
```suggestion
#include <math.h>
```

examples/fluids/qfunctions/shocktube.h
```diff
@@ -0,0 +1,527 @@
...
+    for (int j=0; j<3; j++)
+      dv[j][4][i]  -= wdetJ * (E + P) * (u[0]*dXdx[j][0] + u[1]*dXdx[j][1] +
+                                         u[2]*dXdx[j][2]);
+  } // End Quadrature Point Loop
+
+  // Return
+  return 0;
+}
+// *****************************************************************************
+
```
jedbrown, line 526: Do we need to duplicate this unstabilized Euler qfunction?

examples/fluids/problems/shocktube.c
```diff
@@ -0,0 +1,198 @@
...
+  problem->setup_vol               = Setup;
+  problem->setup_vol_loc           = Setup_loc;
+  problem->setup_sur               = SetupBoundary;
+  problem->setup_sur_loc           = SetupBoundary_loc;
+  problem->ics                     = ICsShockTube;
+  problem->ics_loc                 = ICsShockTube_loc;
+  problem->apply_vol_rhs           = EulerShockTube;
+  problem->apply_vol_rhs_loc       = EulerShockTube_loc;
+  problem->apply_vol_ifunction     = IFunction_EulerShockTube;
+  problem->apply_vol_ifunction_loc = IFunction_EulerShockTube_loc;
```
jedbrown, line 53:
```suggestion
  problem->apply_vol_ifunction     = NULL;
  problem->apply_vol_ifunction_loc = NULL;
```

jedbrown (commented):
@tt-aiken Thanks for all your work on this, and thanks to @LeilaGhaffari for shepherding. I see some fragments that can be deleted. I think it would be better to use the covariant length measure for element size, unless I'm missing something important (I might be). Or at least leave a TODO and corresponding mention in the docs.

examples/fluids/navierstokes.h
LeilaGhaffari: Thanks, @jedbrown . I thought I removed all the duplicates. Missed this one!

examples/fluids/navierstokes.h
```diff
@@ -211,6 +211,11 @@ struct SetupContext_ {
   CeedScalar dc_axis[3];
   CeedScalar wind[3];
   CeedScalar time;
+  CeedScalar mid_point;
+  CeedScalar P_high;
+  CeedScalar rho_high;
+  CeedScalar P_low;
+  CeedScalar rho_low;
```
LeilaGhaffari, line 218: Ah, HIP is failing because these are not added to SetupContext in `qfunctions/advection.c`, `qfunctions/advection2d.c`, and `qfunctions/densitycurrent.c`. You would also need to add this Context to `qfunctions/shocktube.h` to make HIP happy.

examples/fluids/qfunctions/shocktube.h
tt-aiken: Should I just remove this QFunction? Any addition of YZB into the implicit timestepping qfunction can occur in a generalized IFunction that will exist outside of `shocktube.h` and `eulervortex.h`?

examples/fluids/navierstokes.h
tt-aiken: Is there any action for me here? I can add these to the relevant files if needed.

examples/fluids/index.md
tt-aiken: This is measured in the direction of velocity, not the density gradient. Are you suggesting applying the velocity-based measurement or are you suggesting I employ a similar metric based on the density gradient?

jedbrown: Yeah, same scale measured in the direction of the density gradient. We should probably make this a helper function (give the Jacobian and a physical-space vector, return covariant length in that direction). Then you could call the same function for SUPG stabilization with velocity and here with the density gradient.

jedbrown: Maybe also check if it cancels, since you have the density gradient on the reference element already.

tt-aiken: If I understand correctly, this would replace the p vector that I have defined in original post for this PR. For the current formulation, this vector is exactly the element length in the x-direction. Would switching to the covariant length scale be more accurate for arbitrary grids? I don't want to use the covariant length scale for h_SHOCK directly because I had to do some tuning of C_YZB to get good behavior for this testcase.

jedbrown: If you're on a structured grid, the lengths should be exactly the same. It's when you have non-orthogonal grids that the covariant and contravariant lengths differ.

jedbrown: We should make a visual explainer on the website, not specific to the fluids example, because this is a common thing other libCEED users will be thinking about, perhaps for the first time.

examples/fluids/navierstokes.h
LeilaGhaffari: Yeah, the GPU tests won't pass if we don't add them. Do you mind making this change?

examples/fluids/qfunctions/shocktube.h
tt-aiken: I went ahead and removed this.

examples/fluids/navierstokes.h
tt-aiken: Done!

LeilaGhaffari: Thanks! Could you add SetupContext to shocktube.h as well?

examples/fluids/index.md
tt-aiken: Helper function is in the code and working as expected. I would appreciate a double check on my indices for `dot(dXdx,vec)` inside the new helper function.

LeilaGhaffari (approved):

jeremylt:
It looks like this just needs merge conflicts handled and a last look over from @jedbrown before merging. @LeilaGhaffari can you handle the merge conflicts?

LeilaGhaffari:
Yes, I can clean up the git history and resolve the merge conflicts after @jedbrown 's final review.

jedbrown:
This looks good in general. I merged with 'main' to pick up the light refactoring from James. There are two things I'd like to explain.

1. Instability
```console
$ build/fluids-navierstokes -problem shocktube -degree 1 -dm_plex_box_faces 100,1,1 -units_meter 1e-2 units_second 1e-2 -dm_plex_box_lower 0,0,0 -dm_plex_box_upper 1000,10,10 -dm_plex_dim 3 -bc_wall 5,6 -bc_slip_y 3,4 -bc_Slip_z 1,2 -yzb -stab su -compare_final_state_atol 1E-11 -ts_adapt_monitor
[...]
      TSAdapt basic rk 0:5f step 378 accepted t=4.71639    + 1.125e-02 dt=2.975e-02  wlte=0.00457  wltea=   -1 wlter=   -1
      TSAdapt basic rk 0:5f step 379 accepted t=4.72765    + 2.975e-02 dt=2.735e-02  wlte=0.899  wltea=   -1 wlter=   -1
[0]PETSC ERROR: Floating point exception
[0]PETSC ERROR: Infinite or not-a-number generated in norma
[0]PETSC ERROR: See https://petsc.org/release/faq/ for trouble shooting.
[0]PETSC ERROR: Petsc Development GIT revision: v3.16.5-1288-g4f9ab2b475  GIT Date: 2022-03-18 17:15:23 -0600
[0]PETSC ERROR: build/fluids-navierstokes on a ompi named kichatna by jed Sun Mar 20 19:41:19 2022
[0]PETSC ERROR: Configure options --download-chaco --download-ctetgen --download-exodusii --download-glvis --download-ml --download-mumps --download-netcdf --download-pnetcdf --download-ptscotch --download-scalapack --download-superlu --download-triangle --with-cgns --with-cuda --with-hip=0 --with-debugging=0 --with-fortran-bindings=0 --with-hdf5 --with-med --with-metis --with-mpi-dir=/home/jed/usr/ccache/ompi --with-opengl --with-openmp --with-parmetis --with-suitesparse --with-zlib COPTFLAGS="-O2 -march=native -ffp-contract=fast -g" PETSC_ARCH=ompi --download-hypre
[0]PETSC ERROR: #1 TSErrorWeightedNorm2() at /home/jed/petsc/src/ts/interface/ts.c:5203
[0]PETSC ERROR: #2 TSErrorWeightedNorm() at /home/jed/petsc/src/ts/interface/ts.c:5389
[0]PETSC ERROR: #3 TSAdaptChoose_Basic() at /home/jed/petsc/src/ts/adapt/impls/basic/adaptbasic.c:30
[0]PETSC ERROR: #4 TSAdaptChoose() at /home/jed/petsc/src/ts/adapt/interface/tsadapt.c:919
[0]PETSC ERROR: #5 TSStep_RK() at /home/jed/petsc/src/ts/impls/explicit/rk/rk.c:835
[0]PETSC ERROR: #6 TSStep() at /home/jed/petsc/src/ts/interface/ts.c:3615
[0]PETSC ERROR: #7 TSSolve() at /home/jed/petsc/src/ts/interface/ts.c:4014
[0]PETSC ERROR: #8 TSSolve_NS() at /home/jed/src/libCEED/examples/fluids/src/setupts.c:342
[0]PETSC ERROR: #9 main() at /home/jed/src/libCEED/examples/fluids/navierstokes.c:280
```

Exploring in the debugger with the argument `-fp_trap` added to the application arguments, i.e.,
```console
$ gdb -ex r --args build/fluids-navierstokes -fp_trap -problem shocktube -degree 1 -dm_plex_box_faces 100,1,1 -units_meter 1e-2 units_second 1e-2 -dm_plex_box_lower 0,0,0 -dm_plex_box_upper 1000,10,10 -dm_plex_dim 3 -bc_wall 5,6 -bc_slip_y 3,4 -bc_Slip_z 1,2 -yzb -stab su -compare_final_state_atol 1E-11 -ts_adapt_monitor
```
(compile with `OPT='-O0 -g'`), I see both internal energy and density negative. It seems like we should be able to deal with this.

2. Boundary conditions: Density is fixed at 2 on the left endpoint and total energy density is fixed at 5, and similarly at the right boundary. Would this be better as an insulated slip boundary condition, specifying only that the x component of momentum is zero? Replacing `-bc_wall 5,6` with `-bc_slip_x 5,6` allows my simulation to run to completion without obvious artifacts. So this fixes number 1 if you think it's appropriate.

jeremylt:
Note - this branch needs the updates in main for the latest development version of PETSc: #940

jedbrown:
@AdelekeBankole will push a fix for this in a minute. Here's an animation of density and internal energy.

```console
$ build/fluids-navierstokes -problem shocktube -degree 1 -dm_plex_box_faces 200,1,1 -units_meter 1e-2 units_second 1e-2 -dm_plex_box_lower 0,0,0 -dm_plex_box_upper 1000,5,5 -dm_plex_dim 3 -bc_slip_x 5,6 -bc_slip_y 3,4 -bc_slip_z 1,2 -yzb -stab su -ts_adapt_monitor -ts_max_time 20
```

https://user-images.githubusercontent.com/3303/163734249-e836d883-0e3e-4c9d-813f-e08a96c355b0.mp4

jedbrown:
This finally looks okay to me. We just need to fix the test and resolve the merge conflict with main.
```console
$ ./navierstokes -problem shocktube -degree 1 -dm_plex_box_faces 200,1,1 -units_meter 1e-2 units_second 1e-2 -dm_plex_box_lower 0,0,0 -dm_plex_box_upper 1000,5,5 -dm_plex_dim 3 -bc_slip_x 5,6 -bc_slip_y 3,4 -bc_slip_z 1,2 -yzb -stab su -ts_adapt_monitor -ts_max_time 20
```

https://user-images.githubusercontent.com/3303/166250103-4a0c8250-988a-4d0e-ad88-03562834d9f3.mp4

jedbrown:
I merged from main and resolved conflicts. I think this will finally be ready to squash-merge when the tests pass. Thanks for everyone's work on this.

jedbrown:
Thanks @tt-aiken. :rocket: Enjoy your summer.

PR #852, opened 2021-12-07
rust - add slice borrowing

jeremylt:
In this PR, I port `CeedVectorSetArray`/`CeedVectorTakeArray` for borrowed slices.

### DIFF
--- a/rust/libceed/src/vector.rs
+++ b/rust/libceed/src/vector.rs
@@ -131,6 +131,29 @@ impl<'a> VectorOpt<'a> {
 }
 
 // -----------------------------------------------------------------------------
+// CeedVector borrowed slice wrapper
+// -----------------------------------------------------------------------------
+pub struct VectorSliceBorrow<'a> {
+    pub(crate) vector: crate::Vector<'a>,
+    pub(crate) _slice: &'a mut [crate::Scalar],
+}
+
+// -----------------------------------------------------------------------------
+// Destructor
+// -----------------------------------------------------------------------------
+impl<'a> Drop for VectorSliceBorrow<'a> {
+    fn drop(&mut self) {
+        unsafe {
+            bind_ceed::CeedVectorTakeArray(
+                self.vector.ptr,
+                crate::MemType::Host as bind_ceed::CeedMemType,
+                std::ptr::null_mut(),
+            )
+        };
+    }
+}
+
+// -----------------------------------------------------------------------------
 // CeedVector context wrapper
 // -----------------------------------------------------------------------------
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


jeremylt:
Note: this is a blocker for the `meles-rs` PETSc + libCEED Rust examples, currently we can't borrow PETSc vector arrays to wrap in libCEED vectors in Rust

rust/libceed/src/vector.rs
```diff
@@ -375,6 +398,76 @@ impl<'a> Vector<'a> {
...
+    /// // This line will error, as the Vector has dropped the reference
+    /// //   to the slice and no new array was allocated for the Vector
+    /// // vec.view()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn set_borrowed_slice<'b>(
+        &mut self,
+        slice: &'b mut [crate::Scalar],
+    ) -> crate::Result<VectorSliceBorrow<'b>> {
```
jedbrown, line 445: Sorry to lose track of this. My concern is with naming. Seems to me a slice is always borrowed and this terminology (but not type signature) is ambiguous as to whether the slice is provided to the vector or the vector is setting the slice. But what about `wrap_slice` or perhaps `wrap_slice_mut`, which "Uses pre-sized `Vector` to wrap a provided mutable slice". Looks useful and clean otherwise.

jeremylt: Good point. I switched over to `wrapper` based language.

rust/libceed/src/vector.rs
```diff
@@ -375,6 +398,80 @@ impl<'a> Vector<'a> {
...
+    /// // 'array' remains changed
+    /// array.iter().for_each(|v| {
+    ///     assert_eq!(*v, 5.0 as Scalar, "Array not mutated correctly");
+    /// });
+    ///
+    /// // While changes to `vec` no longer affect `array`
+    /// vec.set_value(6.0)?;
+    /// array.iter().for_each(|v| {
+    ///     assert_eq!(*v, 5.0 as Scalar, "Array mutated without permission");
+    /// });
```
jedbrown, line 442: Do you prefer the above style? You seem to be using it consistently, though this alternative has slightly less syntactic noise.
```suggestion
    /// for v in array.iter() {
    ///     assert_eq!(*v, 5.0 as Scalar, "Array mutated without permission");
    /// }
```
From [the docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each)
> This is equivalent to using a for loop on the iterator, although break and continue are not possible from a closure. It’s generally more idiomatic to use a for loop, but for_each may be more legible when processing items at the end of longer iterator chains. In some cases for_each may also be faster than a loop, because it will use internal iteration on adapters like Chain.

jedbrown (approved):
LGTM, thanks.

rust/libceed/src/vector.rs
```diff
@@ -375,6 +398,80 @@ impl<'a> Vector<'a> {
...
+                host,
+                copy_mode,
+                slice.as_ptr() as *mut crate::Scalar,
+            )
+        };
+        self.check_error(ierr)?;
+
+        let mut ptr_copy = std::ptr::null_mut();
+        let ierr = unsafe { bind_ceed::CeedVectorReferenceCopy(self.ptr, &mut ptr_copy) };
+        self.check_error(ierr)?;
```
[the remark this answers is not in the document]
jeremylt: The reference is because I'm making and then dropping a Vector, but I could avoid that by just keeping a raw C pointer instead of a C pointer converted to a Rust Vector

rust/libceed/src/vector.rs
jeremylt: No particular reason, I'm just not used to mixing and matching an iterator and older style loops

rust/libceed/src/vector.rs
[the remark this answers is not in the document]
jedbrown: No strong preference from me.

rust/libceed/src/vector.rs
jeremylt: I switched to `for ... in` when the `.for_each` isn't at the 'end of longer iterator chains'.

PR #853, opened 2021-12-07
vec/qf - initial valid/borrowed/owned split for data

jeremylt:
As we talked about on the telecon. I still want to do a pass to simplify the logic, but this runs and passes the tests.


jeremylt:
I think we should make sure this design lines ups with @nbeams's path ahead for CeedVector precision, but I think the owned vs borrowed changes in here are important to get into main as they represent a current [minor] flaw in our interface.

backends/cuda/ceed-cuda-qfunctioncontext.c
```diff
@@ -41,8 +41,20 @@ static inline int CeedQFunctionContextSyncH2D_Cuda(
   CeedQFunctionContext_Cuda *impl;
   ierr = CeedQFunctionContextGetBackendData(ctx, &impl); CeedChkBackend(ierr);
 
+
+  if (impl->d_data_borrowed) {
+    impl->d_data = impl->d_data_borrowed;
+  } else if (impl->d_data_owned) {
+    impl->d_data = impl->d_data_owned;
+  } else {
```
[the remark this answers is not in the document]
jeremylt: We could find ourselves in a situation where the user
1) sets the host pointer
2) sets the device pointer
3) gets the host pointer (making device pointer stale)
4) takes host pointer

I would expect that we would still use the device pointer that the user told us to use in that situation even though the data at this location is not up to date.

backends/cuda/ceed-cuda-qfunctioncontext.c
```diff
@@ -160,95 +199,97 @@ static int CeedQFunctionContextTakeData_Cuda(const CeedQFunctionContext ctx,
...
-    impl->memState = CEED_CUDA_DEVICE_SYNC;
     *(void **)data = impl->d_data;
     break;
   }
+
+  // Mark only pointer for requested memory as valid
+  ierr = CeedQFunctionContextSetAllStale_Cuda(ctx); CeedChkBackend(ierr);
+  switch (mtype) {
+  case CEED_MEM_HOST:
+    impl->h_data = *(void **)data;
```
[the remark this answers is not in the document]
jeremylt: Right, we assume that write access is always granted for context data

backends/cuda/ceed-cuda-vector.c
```diff
@@ -268,78 +329,44 @@ static int CeedVectorGetArrayRead_Cuda(const CeedVector vec,
...
-                                   CeedScalar **array) {
+                                   const CeedMemType mtype, CeedScalar **array) {
   int ierr;
-  Ceed ceed;
-  ierr = CeedVectorGetCeed(vec, &ceed); CeedChkBackend(ierr);
   CeedVector_Cuda *data;
   ierr = CeedVectorGetData(vec, &data); CeedChkBackend(ierr);
 
-  // Sync array to requested memtype and update pointer
+  ierr = CeedVectorGetArrayRead_Cuda(vec, mtype, (const CeedScalar **)array);
```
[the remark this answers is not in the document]
jeremylt: Given the complexity of the logic, I strongly dislike duplicating the code for this case

backends/cuda/ceed-cuda-vector.c
```diff
@@ -164,49 +234,45 @@ static int CeedVectorSetArray_Cuda(const CeedVector vec,
...
-    if (impl->memState == CEED_CUDA_DEVICE_SYNC) {
+    if (!impl->h_array_borrowed)
+      // LCOV_EXCL_START
+      return CeedError(ceed, CEED_ERROR_BACKEND,
+                       "No host array set with CeedVectorSetArray and CEED_USE_POINTER");
+    // LCOV_EXCL_STOP
+
+    if (!impl->h_array && impl->d_array) {
       ierr = CeedVectorSyncD2H_Cuda(vec); CeedChkBackend(ierr);
     }
```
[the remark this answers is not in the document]
jeremylt: `!impl->h_array` means that the host data is stale. `impl->d_array` means the device data is not stale. The two together means that the device has more up to date information than the host and we should update the host before returning the host pointer to the user.

backends/cuda/ceed-cuda-qfunctioncontext.c
```diff
@@ -160,95 +199,97 @@ static int CeedQFunctionContextTakeData_Cuda(const CeedQFunctionContext ctx,
...
-    if (impl->h_data == NULL) {
-      ierr = CeedMalloc(bytes(ctx), &impl->h_data_allocated);
-      CeedChkBackend(ierr);
-      impl->h_data = impl->h_data_allocated;
-    }
-    if (impl->memState == CEED_CUDA_DEVICE_SYNC) {
+    if (!impl->h_data_borrowed)
+      // LCOV_EXCL_START
+      return CeedError(ceed, CEED_ERROR_BACKEND,
+                       "No host context data set with CeedQFunctionContextSetData and CEED_USE_POINTER");
```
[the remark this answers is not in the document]
jeremylt: What about these updated error messages I just pushed?

jeremylt:
> I assume we also want to use `void *` instead of `CeedScalar *` in the `CeedQFunctionContext_[Cuda/Hip]` backend data structs? And in the parameter for the data variable in `CeedQFunctionContextTakeData_Cuda`? (The type in the HIP version got changed.)

Thanks for catching that! Updated.

backends/cuda/ceed-cuda-qfunctioncontext.c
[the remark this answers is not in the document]
jedbrown: > I would expect that we would still use the device pointer that the user told us to use in that situation even though the data at this location is not up to date.

I would not allow `CeedVectorGetArrayRead` (or other operations that would read values) in that circumstance. The memory is allocated, but there is no current valid state.

One problem with promoting that stale array to being active (besides losing strictness) is that you'd get a different result if you took a diagnostic norm between step 3 and 4. (Computing the norm on the GPU will copy the newly-valid state there as well.) But if you're debugging some corrupt data, you might check the norm there and expect that doing so doesn't produce observable side-effects.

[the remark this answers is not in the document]
jeremylt: I agree. I need to double check the logic, but I think that should be the current behavior (or my intention). The promotion to valid should only occur if we call something like set value.

jeremylt:
I rearranged and clarified the logic. I think this should be easier to understand/remember and modify in future work.

jeremylt:
With this latest pass, I pulled the data validity checks up into the interface, simplifying the logic in the backends. I think this should be pretty close to what we need.

backends/cuda/ceed-cuda-vector.c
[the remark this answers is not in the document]
nbeams: But is code duplication really worse than lying to the compiler about `const`ness?  I guess we have different opinions about which is the lesser of two evils :)

backends/cuda/ceed-cuda-qfunctioncontext.c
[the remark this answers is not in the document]
nbeams: Do I need to go through and look at all the latest changes to understand this part of the logic?  I see you said you have moved some checks up to the interface level.  I just did a fresh pull of the latest version of the branch.

backends/cuda/ceed-cuda-vector.c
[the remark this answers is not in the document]
jeremylt: We could just remove the const from the signature of this backend internal function.

backends/cuda/ceed-cuda-qfunctioncontext.c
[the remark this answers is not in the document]
jeremylt: Yeah, parts of the logic changed, so I'm hoping that the new version is clearer.

interface/ceed-vector.c
```diff
@@ -326,6 +352,15 @@ int CeedVectorTakeArray(CeedVector vec, CeedMemType mem_type,
                      "has read access");
   // LCOV_EXCL_STOP
 
+  bool has_valid_array = true;
+  ierr = CeedVectorHasValidArray(vec, &has_valid_array); CeedChk(ierr);
+  if (!has_valid_array)
+    // LCOV_EXCL_START
+    return CeedError(vec->ceed, CEED_ERROR_BACKEND,
+                     "CeedVector has no valid data to take, "
+                     "must set data with CeedVectorSetValue or CeedVectorSetArray");
```
nbeams, line 361: Is this a change in the new rule for `TakeArray`?  The error messages in `CeedVectorTakeArray_Cuda` still say that you must call `CeedVectorSetArray` with `CEED_USE_POINTER` in order to call `TakeArray`, while here it says `CeedVectorSetValue` would be fine, as well (though `SetValue` could potentially allocate its own memory).  Maybe this one is supposed to have a different error message than the other functions using `HasValidArray`?

backends/cuda/ceed-cuda-qfunctioncontext.c
```diff
@@ -57,198 +74,278 @@ static inline int CeedQFunctionContextSyncD2H_Cuda(
...
+  *has_valid_data = !!impl->h_data || !!impl->d_data;
+
+  return CEED_ERROR_SUCCESS;
+}
+
+//------------------------------------------------------------------------------
+// Check if data of given type needs sync
+//------------------------------------------------------------------------------
+static inline int CeedQFunctionContextNeedSync_Cuda(
+  const CeedQFunctionContext ctx, CeedMemType mtype, bool *need_sync) {
```
nbeams, line 144: I really like this way of checking whether we need to sync (and in the vector stuff, too).  I think it's easier to follow than what we had before.

backends/cuda/ceed-cuda-vector.c
```diff
@@ -54,46 +71,149 @@ static inline int CeedVectorSyncD2H_Cuda(const CeedVector vec) {
...
+// Check if CeedVector has any valid pointer
+//------------------------------------------------------------------------------
+static inline int CeedVectorHasValidArray_Cuda(const CeedVector vec,
+    bool *has_valid_array) {
+  int ierr;
+  CeedVector_Cuda *impl;
+  ierr = CeedVectorGetData(vec, &impl); CeedChkBackend(ierr);
+
+  *has_valid_array = !!impl->h_array || !!impl->d_array;
+
```
nbeams, line 135: Is there a reason for `!!` rather than just checking for `impl->h_array`?

backends/cuda/ceed-cuda-vector.c
```diff
@@ -257,89 +390,69 @@ static int CeedVectorSetValue_Cuda(CeedVector vec, CeedScalar val) {
...
+  CeedVector_Cuda *impl;
+  ierr = CeedVectorGetData(vec, &impl); CeedChkBackend(ierr);
 
-  // Sync array to requested memtype and update pointer
+  bool has_valid_array = true, is_array_of_type = true;
+  ierr = CeedVectorHasValidArray(vec, &has_valid_array); CeedChkBackend(ierr);
+  ierr = CeedVectorIsArrayOfType_Cuda(vec, mtype, &is_array_of_type);
+  CeedChkBackend(ierr);
+  if (!has_valid_array && is_array_of_type) {
+    // LCOV_EXCL_START
```
nbeams, line 405: I don't understand why we want `is_array_of_type` here.  It seems that `is_array_of_type` will be true if either of the `borrowed` or `owned` arrays exists for the requested memtype, correct?  Why do we want to check specifically that the vector does *not* have a valid array but *does* have some invalid/stale data for the memtype before throwing an error?  Isn't no valid data alone enough to throw an error for `GetArrayRead`?

backends/cuda/ceed-cuda-vector.c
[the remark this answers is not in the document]
nbeams: But then it wouldn't match the prototype in the interface for `CeedVector`...

interface/ceed-vector.c
jeremylt: Two things need to happen

1) the user must set the pointer with SetArray

2) the data must be valid

HasValidArray only tells us if the data is valid, condition 2. Calling SetValue on a user provided array can fix the invalid data issue. Not sure if there is a better way to put that in the two error messages.

backends/cuda/ceed-cuda-vector.c
jeremylt: Just being explicit that I want a boolean value

backends/cuda/ceed-cuda-vector.c
jeremylt: We use the pattern of calling GetArray before setting any allocations all over the place.

```
CeedVectorCreate(ceed, 10, &vec);
CeedVectorGetArray(vec, &array);
for (i=0; i<10; i++) {
  // code initializing values
}
```

I want to support this while preventing access to stale data.

backends/cuda/ceed-cuda-vector.c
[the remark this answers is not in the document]
jeremylt: Ok. Then I can take all of the contents out of this function, put it in a third function, and then call this third function by GetArray and GetArrayRead. Would this be better?

We do this, but in the reverse direction on the CPU side, but that will need this same change once we switch to mixed precision.

interface/ceed-vector.c
jeremylt: Perhaps a interface level check HasBorrowedArrayOfType

backends/cuda/ceed-cuda-vector.c
nbeams: Let me see if I understand now: the goal is basically to make sure that the vector either a) has valid data, or b) doesn't have *any* data (for the memtype) at all?  Otherwise, if it has data but the data is stale, it is an error?  

My understanding from our discussion of the future multiprecision vectors was that once a vector is given access to a memory location through `SetArray` + `USE_POINTER`, it should continue to assume it can use that memory location until told otherwise (through `TakeArray` or a new `SetArray` call).  For `GetArrayRead`, it makes sense that we wouldn't want access to stale data, since we presumably care about the current values in the vector (otherwise, why use `GetArrayRead`?).  But for `GetArray`, couldn't we want access to the stale data *location* in order to write new values (and then make the data valid again)?  

Actually, this leads me to another question: if we presume we can use the borrowed location until told otherwise, and we also end up allocating memory later, which one (owned or borrowed) do we prefer later, if neither is currently valid?  ...Could such a scenario even ever happen?

interface/ceed-vector.c
nbeams: Right, I meant that for this function (and this function only) the rule is not just that the data has to be valid, but that it had to have been set with `SetArray` + `CEED_USE_POINTER` combo, so using `SetValue` for validity purposes not allowed.  Or more specifically, I was noting that the error messages for the interface level function and the Cuda backend version don't match (and the backend version is the one that follows what I thought was the desired rule).

So then the `HasBorrowedArrayOfType` would specifically check for borrowed data?

backends/cuda/ceed-cuda-vector.c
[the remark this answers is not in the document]
nbeams: Personally, I would prefer that, but obviously my opinion isn't the only one that matters here.

backends/cuda/ceed-cuda-vector.c
jeremylt: > the goal is basically to make sure that the vector either a) has valid data, or b) doesn't have any data (for the memtype) at all? Otherwise, if it has data but the data is stale, it is an error?

Exactly.

>  But for `GetArray`, couldn't we want access to the stale data location in order to write new values (and then make the data valid again)?

For `GetArray` I am treating it as read/write access. I could see a user wanting to read some values and write some values, which would lead to an array with a mix of valid and invalid data.

We would only fall into this use case if the user Sets the array and then Takes the array back without ensuring synchronization. I would expect a user to call `SyncArray' before taking the pointer back if they want to make sure the data is valid in a particular memory location that they are not taking or to call `SetValue` to reset the validity of the data at the location.

> if we presume we can use the borrowed location until told otherwise, and we also end up allocating memory later, which one (owned or borrowed) do we prefer later, if neither is currently valid? ...Could such a scenario even ever happen?

I think we should try to prevent this from ever happening. Setting a borrowed array should free a corresponding allocation and vise versa.

interface/ceed-vector.c
jeremylt: >  but that it had to have been set with `SetArray` + `CEED_USE_POINTER` combo, so using `SetValue` for validity purposes not allowed

I think it would be ok for a user to call `SetValue` after `SetArray` to restore validity of data at a location, which is what I was trying to say. Let me see if I can clean up this error message.

backends/cuda/ceed-cuda-vector.c
nbeams: >  I would expect a user to call `SyncArray` before taking the pointer back if they want to make sure the data is valid in a particular memory location that they are not taking or to call `SetValue` 

Wouldn't this be a departure from how it currently works, though?  Didn't we do the checking for them and sync if necessary when requesting memory access on host/device, no explicit `Sync` required?

> I could see a user wanting to read some values and write some values, which would lead to an array with a mix of valid and invalid data.

But in this setup, `GetArray` and `GetArrayRead` work the same way, so `GetArray` would also error if you were trying to get access to stale data -- but it shouldn't, right?  `GetArray` should let you access stale data for write purposes?

Actually, this is something I was already wondering about in terms of multiprecision -- should we have a special case for write-only access, where we know we don't care about syncing the current values even if they are stale?  And a version that will do the sync (for host/device now, for precision later) if required?

backends/cuda/ceed-cuda-vector.c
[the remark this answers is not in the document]
jeremylt: Updated

backends/opt/ceed-opt-operator.c
```diff
@@ -248,6 +248,10 @@ static inline int CeedOperatorSetupInputs_Opt(CeedInt num_input_fields,
           CeedChkBackend(ierr);
           impl->input_states[i] = state;
         }
+        // Get evec
+        ierr = CeedVectorGetArrayRead(impl->e_vecs_full[i], CEED_MEM_HOST,
```
jeremylt, line 252: This PR caught a benign misuse where we were requesting Read access to a vector with no data. This had no practical impact since we never used this array pointer, but it is a nice bug to have fixed.

backends/cuda/ceed-cuda-vector.c
jeremylt: I have no objection to adding a `GetArrayInvalid` where the user requests access to a dirty array, but it isn't something that we currently need internally. With the exception of the one small bug I found in the `/cpu/self/opt` backends, internally and in the examples we have only been calling `GetArray` if we have valid data or no arrays yet allocated.

jedbrown: PETSc has `VecGetArrayWrite()`, which makes it clear at the call site that old values are to be considered invalid. In debug mode, these can be initialized with NaN to prevent accidental use.

jeremylt: That name makes sense to me. I added this function to the interface.

We could adopt the convention that `GetArray` and `GetArrayRead` require valid data while `GetArrayWrite` is the only function that allows access to uninitialized or invalid data. That would simplify the error checking, but would require a couple of small changes in the backends. Any votes yea on nay?

nbeams: > We could adopt the convention that GetArray and GetArrayRead require valid data while GetArrayWrite is the only function that allows access to uninitialized or invalid data. 

I like it, yea vote from me 👍

nbeams: I still do think this may cause minor changes somewhere with usage.  Previously, if you called `GetArray` or `GetArrayRead` with device memtype, and the host data was set, the vector would sync before giving you the device pointer.  Now, it would cause an error unless you explicitly call `Sync` first, is that correct?

jeremylt: Not quite. I think that automatic synchronization should be expected, unless there exists no valid data to synchronize. If there is no valid data, then in this branch read or read/write access will fail.

This should be pretty uncommon. The  only use case I can think of would be if the user calls `TakeArray` on the only valid memory location and then wants to get read access to the array.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -88,10 +88,45 @@ and
 are purely implicit -- one just indexes the same array using the
 appropriate convention.
 
-## `restrict` semantics
+## `restrict` Semantics
 
 QFunction arguments can be assumed to have `restrict` semantics. That is, each input and output array must reside in distinct memory without overlap.
 
+## CeedVector Array Access Semantics
```
jeremylt, line 95: I tried to distill the changes and discussion into the documentation here. These are the rules I have coded vector access to follow in this branch.

nbeams: I think this is very clear, thank you.

jeremylt: Thanks for your help and review on this PR - I think we have some nice updates

jeremylt:
Everything passes the test suite. This PR got bigger than I wanted, but I think we should be good for review + squash + merge, modulo anything else that we want to fixup.

nbeams:
A quick test tells me we've broken something in the MFEM integration for MFEM's algebraic CEED solver -- from MFEM's ex1 with `-d ceed-cuda -a -pa` options.  That part seems to be fixed by swapping a `GetArray` for `GetArrayWrite`, but then I'm getting a segfault from `CeedOperatorGetActiveElemRestriction`, so most likely not related to this PR... (I cherry-picked the commits from https://github.com/mfem/mfem/pull/2569 since I know the MFEM integration was already broken.)

At any rate, this may be a breaking change for some people using libCEED, depending on how it's integrated (where they should use `GetArrayWrite` now instead of `GetArray`).

jeremylt:
Since this is a breaking change, it might be a fair point to cut a new release. We haven't done a fall release yet.

python/tests/test-1-vector.py
```diff
@@ -321,6 +322,7 @@ def test_199(ceed_resource):
     ceed = libceed.Ceed(ceed_resource)
 
     vec = ceed.Vector(12)
+    vec.set_value(0.0)
     with vec.array(4, 3) as x:
```
nbeams, line 326: Should we allow write/invalid access for Python users (wrapping `GetArrayWrite`) so initialization is not always required (like using `numpy.empy`)?

nbeams (approved):
Still have a few minor questions, but I'm going ahead and marking as "approve" for this version of the interface.  Thanks for all the work @jeremylt

nbeams:
Though I guess, do we need to investigate the GitLab failure on Noether?

python/tests/test-1-vector.py
jeremylt: I'm not sure. I definitely do not want to do it for the Rust interface. Python is more 'wild west' so I suppose I could go either way.

backends/cuda/ceed-cuda-operator.c
```diff
@@ -150,6 +150,9 @@ static int CeedOperatorSetupFields_Cuda(CeedQFunction qf, CeedOperator op,
         ierr = CeedElemRestrictionCreateVector(Erestrict, NULL,
                                                &evecs[i + starte]);
         CeedChkBackend(ierr);
+        // Allocate array
+        ierr = CeedVectorSetArray(evecs[i + starte], CEED_MEM_DEVICE, CEED_COPY_VALUES,
+                                  NULL); CeedChkBackend(ierr);
```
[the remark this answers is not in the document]
jeremylt: `CeedElemRestriction` needs to use `GetArray` instead of `GetArrayWrite` because the convention is that we sum into the target for transpose mode, so we need to use initialized arrays for restrictions.

nbeams:
Oh wait, I have another question about `GetArrayWrite`, prompted by the issue with the MFEM integration.  It seems I may have run into a situation where calling `GetArrayRead` after `GetArrayWrite` throws the "no valid data to read" error, though `GetArrayWrite` should leave the vector in a valid state.  Hmmm... will need to investigate further to make sure the issue is on the MFEM side rather than libCEED.

jeremylt:
ICC/IFort failure is Intel's flakey installer, unrelated to this PR

jeremylt:
> It seems I may have run into a situation where calling `GetArrayRead` after `GetArrayWrite` throws the "no valid data to read" error, though `GetArrayWrite` should leave the vector in a valid state. Hmmm... will need to investigate further to make sure the issue is on the MFEM side rather than libCEED.

I think it's on the MFEM side of the integration? `t124` tests this specific use case and all backends pass.

backends/blocked/ceed-blocked-operator.c
```diff
@@ -650,6 +651,7 @@ static inline int CeedOperatorLinearAssembleQFunctionCore_Blocked(
   if (!l_vec) {
     ierr = CeedVectorCreate(ceed, num_blks*blk_size*Q*num_active_in*num_active_out,
                             &l_vec); CeedChkBackend(ierr);
+    ierr = CeedVectorSetValue(l_vec, 0.0); CeedChkBackend(ierr);
     impl->qf_l_vec = l_vec;
```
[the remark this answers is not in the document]
jeremylt: I'm trying to be sparing with the use of `GetArrayWrite`. I'm trying to restrict its use to places where we absolutely know no synchronization will ever be wanted.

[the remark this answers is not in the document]
jeremylt: I think the usage should all be consistent now.

nbeams:
> > It seems I may have run into a situation where calling `GetArrayRead` after `GetArrayWrite` throws the "no valid data to read" error, though `GetArrayWrite` should leave the vector in a valid state. Hmmm... will need to investigate further to make sure the issue is on the MFEM side rather than libCEED.
> 
> I think it's on the MFEM side of the integration? `t124` tests this specific use case and all backends pass.

Yes, after looking into it some more, I do think it's just that we broke the MFEM integration (but it was already broken).

backends/cuda/ceed-cuda-operator.c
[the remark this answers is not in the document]
nbeams: Ah, that makes sense -- but each backend already takes care of that inside those functions/kernels, no?  Since they're only for the transpose mode, and we don't want to unnecessarily initialize. Also, `CEED_COPY_VALUES` and `NULL` just allocates memory instead of initializing, so we're still not getting an initialized array here.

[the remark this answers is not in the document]
jeremylt: Woops, wrong word. We can't use `GetArrayWrite` inside of the restriction because of the summing convention, but this allocation is for the inputs, which use the NoTranspose mode, so we overwrite all values. 

Transpode mode : need initialized values
NoTransposed mode : need allocated memory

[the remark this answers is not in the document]
jeremylt: Backends could choose `GetArray` for Transpose mode and `GetArrayWrite` for NoTranspose mode though...

python/tests/test-1-vector.py
nbeams: Yeah, I think we should provide write access to Python, since we "accidentally" allowed it before (`GetArray` with uninitialized data was okay), and indeed, that is what these tests were doing.  I don't think we should force Python users to pay an unnecessary initialization cost when other standard Python computing libraries (like `numpy`) allow use of uninitialized data.  Of course, I don't know how often Python users will be manipulating `CeedVector`s directly, outside of other libCEED components, but I'd still vote for adding it to the Python API.

backends/cuda/ceed-cuda-operator.c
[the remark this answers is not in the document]
nbeams: True.  That makes a good case for using `SetArray` to allocate (but not initialize) here.  👍

python/tests/test-1-vector.py
jeremylt: I added `vector.array_write()`

PR #854, opened 2021-12-07
doc - minor doxygen fixes

jeremylt:
This PR fixes some missing arguments in function documentation.


jedbrown (approved):
Is there a way to test that the documented arguments match the actual named arguments? Like make this a Doxygen error without too many false positives from the copious other Doxygen output?

jeremylt:
I haven't yet found a way to convert the Doxygen warnings into errors for missing arguments, but I agree that it would be helpful

PR #855, opened 2021-12-08
Remove magic number 16

jeremylt:
This removes the use of the magic number 16 as the maximum number of input or output fields for a QFunction/Operator and instead declares the value as a constant.

I then use this value to [finally] remove the `e_data` array from my pre-vector-access-control hack in the CPU operator data.

### DIFF
--- a/backends/blocked/ceed-blocked.h
+++ b/backends/blocked/ceed-blocked.h
@@ -29,17 +29,16 @@ typedef struct {
 typedef struct {
   bool is_identity_qf, is_identity_restr_op;
   CeedElemRestriction *blk_restr; /* Blocked versions of restrictions */
-  CeedVector *e_vecs;      /* All E-vectors, inputs followed by outputs */
-  uint64_t *input_state;   /* State counter of inputs */
-  CeedVector *e_vecs_in;   /* Input E-vectors needed to apply operator */
-  CeedVector *e_vecs_out;  /* Output E-vectors needed to apply operator */
-  CeedVector *q_vecs_in;   /* Input Q-vectors needed to apply operator */
-  CeedVector *q_vecs_out;  /* Output Q-vectors needed to apply operator */
-  CeedInt    num_e_vecs_in;
-  CeedInt    num_e_vecs_out;
-  CeedInt    qf_num_active_in, qf_num_active_out;
+  CeedVector *e_vecs_full; /* Full E-vectors, inputs followed by outputs */
+  uint64_t *input_states;  /* State counter of inputs */
+  CeedVector *e_vecs_in;   /* Element block input E-vectors  */
+  CeedVector *e_vecs_out;  /* Element block output E-vectors */
+  CeedVector *q_vecs_in;   /* Element block input Q-vectors  */
+  CeedVector *q_vecs_out;  /* Element block output Q-vectors */
+  CeedInt    num_inputs, num_outputs;
+  CeedInt    num_active_in, num_active_out;
   CeedVector *qf_active_in;
-  CeedVector qf_lvec;
+  CeedVector qf_l_vec;
   CeedElemRestriction qf_blk_rstr;
 } CeedOperator_Blocked;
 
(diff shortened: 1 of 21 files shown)


jeremylt:
I think this is relevant to the mixed precision discussion. I didn't tackle it, but a similar change can be made in the CUDA and HIP operators.

jeremylt:
The CUDA failures seem to be unrelated to the changes in this branch.

backends/ref/ceed-ref.h
```diff
@@ -53,9 +53,7 @@ typedef struct {
 
 typedef struct {
   bool is_identity_qf, is_identity_restr_op;
-  CeedVector
-  *e_vecs;   /* E-vectors needed to apply operator (input followed by outputs) */
-  CeedScalar **e_data;
+  CeedVector *e_vecs;      /* All E-vectors, inputs followed by outputs */
   uint64_t *input_state;   /* State counter of inputs */
   CeedVector *e_vecs_in;   /* Input E-vectors needed to apply operator */
   CeedVector *e_vecs_out;  /* Output E-vectors needed to apply operator */
```
nbeams, line 59: Wait, why do we need `e_vecs` AND `e_vecs_in`/`e_vecs_out`?

nbeams (commented):
The CUDA/HIP backends also use the `edata` setup in their backend Operator data structs.  Were you planning to modify that, too, or did you want me to do it?

Though, while I'm not opposed to these changes or removing the cryptic 16, I'm not sure I see the immediate connection to changing the use of `e_data` in this way and making it easier to change to a flexible-precision version in the future.  We'll still need to change its type in the prototype of the backend's internal functions (can't use `CeedScalar **`) unless we want to have a different version of OperatorApply for every possible E-vector precision.  Is it to simplify memory management/allocation issues if the `e_data` isn't part of the backend data struct?

nbeams:
> The CUDA failures seem to be unrelated to the changes in this branch.

I can confirm that when I tested this branch locally (CUDA), all tests passed.

jeremylt:
> The CUDA/HIP backends also use the edata setup in their backend Operator data structs. Were you planning to modify that, too, or did you want me to do it?

Either of us doing it was fine. I just wanted to make sure this change was fine before going through with the work.

> Is it to simplify memory management/allocation issues if the `e_data` isn't part of the backend data struct?

This is exactly what I was thinking of. It is just a holdover from older code, so it's not really needed.

backends/ref/ceed-ref.h
jeremylt: Good point - more legacy consequences. Let me see if it removes nicely.

jeremylt: Actually... both do need to be there. One is the *full* evec and the other is the partial evec for the current block of elements.

backends/blocked/ceed-blocked.h
```diff
@@ -29,17 +29,16 @@ typedef struct {
 typedef struct {
   bool is_identity_qf, is_identity_restr_op;
   CeedElemRestriction *blk_restr; /* Blocked versions of restrictions */
-  CeedVector *e_vecs;      /* All E-vectors, inputs followed by outputs */
+  CeedVector *e_vecs_full; /* Full E-vectors, inputs followed by outputs */
```
jeremylt, line 32: I update this name and a few others to something clearer

backends/ref/ceed-ref.h
nbeams: Ah, yes, thank you.  The new names and comments are much more clear.  (I'm so used to thinking about the GPU backends that I forgot this distinction existed in the CPU backends.)

jeremylt: Cool. I'll start on the CUDA/HIP version of these changes tomorrow.

jeremylt:
I mirrored the change in CUDA/HIP. This should be good to merge.

jedbrown (approved):

PR #856, opened 2021-12-09
junit - append instead of overwrite

jeremylt:
Oops. We were only reporting HIP CI on Noether.

### DIFF
--- a/tests/junit.py
+++ b/tests/junit.py
@@ -164,11 +164,19 @@ if __name__ == '__main__':
         backends = os.environ['BACKENDS'].split()
 
         result = run(args.test, backends)
-        output = (os.path.join('build', args.test + '.junit')
+
+        junit_batch = ''
+        try:
+            junit_batch = '-' + os.environ['JUNIT_BATCH']
+        except:
+            pass
+        output = (os.path.join('build', args.test + junit_batch + '.junit')
                   if args.output is None
                   else args.output)
+
         with open(output, 'w') as fd:
             TestSuite.to_file(fd, [result])
+
         for t in result.test_cases:
             failures = len([c for c in result.test_cases if c.is_failure()])
             errors = len([c for c in result.test_cases if c.is_error()])
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -42,25 +42,25 @@ noether-rocm:
     - echo '[{"subject":"/","metrics":[{"name":"Transfer Size (KB)","value":"19.5","desiredSize":"smaller"},{"name":"Speed Index","value":0,"desiredSize":"smaller"},{"name":"Total Score","value":92,"desiredSize":"larger"},{"name":"Requests","value":4,"desiredSize":"smaller"}]}]' > performance.json
 #    Note: PETSC_DIR is set by default in GitLab runner env, unsetting to isolate core tests
     - export PETSC_DIR= PETSC_ARCH=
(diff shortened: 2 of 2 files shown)


jeremylt:
hmm, that isn't quite right either

tests/junit.py
```diff
@@ -4,6 +4,7 @@
 import sys
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), 'junit-xml')))
 from junit_xml import TestCase, TestSuite
+from junitparser import JUnitXml
```
jeremylt, line 7: @jedbrown, the `junit_xml` module does not support merging a preexisting `xml` with current results. This is one module that I found that does. What do you think about adding?

jeremylt: It would need to be installed on Noether, LV (easy), and the Azure image. I have verified that this works locally.

tests/junit.py
```diff
@@ -167,8 +168,22 @@ def run(test, backends):
         output = (os.path.join('build', args.test + '.junit')
                   if args.output is None
                   else args.output)
-        with open(output, 'w') as fd:
-            TestSuite.to_file(fd, [result])
+
+        if os.path.exists(output):
```
jedbrown, line 172: I don't like this race condition. Can we not write different files?

jeremylt: I'm not sure under what use case we'd see a race condition? A single invocation of `make junit -j` will only ever write once to a `.junit` file with the results for all backends. But currently if we call `make junit -j` a second time (once for all CPU backend, then again for all GPU backends), then the results from the first invocation are overwritten. This change just appends the new data to the old results.

Do we want to support two independent processes simultaneously calling `make junit` in the same folder at the same time? Wouldn't we always use `make junit -j` in that case?

jeremylt: We could certainly instead just append a random value into the filename if that is preferred, it just seemed like a lot of extra clutter.

jedbrown: I'd rather just make a unique file name for now. Perhaps pass an env variable to put into the file name. I'd prefer that repeated invocation be idempotent and not leave stale results. Like if you run, fix a bug, and re-run, you should see a clean result. I know that workflow isn't relevant to CI, but minimizing surprise is good.

jedbrown: For example, `make junit -j JUNIT_BATCH=cuda` would generate `result.cuda.junit`.

tests/junit.py
```diff
@@ -164,11 +164,19 @@ def run(test, backends):
         backends = os.environ['BACKENDS'].split()
 
         result = run(args.test, backends)
-        output = (os.path.join('build', args.test + '.junit')
+
+        junit_batch = ''
+        try:
+            junit_batch = '-' + os.environ['JUNIT_BATCH'] + '-'
```
jedbrown, line 170:
```suggestion
            junit_batch = '-' + os.environ['JUNIT_BATCH']
```

jedbrown (approved):
LGTM, but needs rebase/squash with first commit.

tests/junit.py
```diff
@@ -167,7 +167,7 @@ def run(test, backends):
 
         junit_batch = ''
         try:
-            junit_batch = '-' + os.environ['JUNIT_BATCH'] + '-'
+            junit_batch = '-' + os.environ['JUNIT_BATCH']
```
jeremylt, line 170: ha, same thought

PR #858, opened 2021-12-15
Strip redundant/outdated license info duplication

jeremylt:


jeremylt:
This should be good to squash+merge now

jedbrown:
@tzanio I'd like your input on this. (Anyone else, feel free to comment as well.) The issue is that these are redundant with project-level docs and are increasingly incorrect as people not funded by CEED contribute and/or code is refactored. It's untenable to accurately identify the copyright holders on a per-file basis. If we are to merge this, we should consider the following:
1. Use `SPDX-License-Identifier: BSD-2-Clause` in each file. [More on SPDX](https://spdx.dev/ids/). The Linux kernel does this. The precise copyright status requires tracing who modified the files, but it's better than incorrect information.
2. Have a project-level document storing authors and time periods in which copyright on their work may be assigned to another entity (employer or sponsor).
3. Update `NOTICE` to limit its scope to portions of the software (which need not be identified precisely).
4. Developer documentation for new process.

tzanio:
@jedbrown, just wanted to let you know that I saw this and I'm thinking about it. These notices were a requirement for us to be able to release (all) CEED software at LLNL. I am checking what can we do at this point.

jedbrown:
Thanks. Does LLNL have a policy on contribution to projects in which LLNL does not have a controlling stake in the project governance? One would think that a project-level notice would be sufficient and all they could reasonably ask of such projects. Anecdotally, PETSc has received contributions from LLNL people with no special notices. Meanwhile, the standard spack notice (in every file) is more compact.
```python
# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
# Spack Project Developers. See the top-level COPYRIGHT file for details.
#
# SPDX-License-Identifier: (Apache-2.0 OR MIT)
```
This date range (2013-2021) is present in all files I've checked, even those that are brand new in 2021 and authored solely by non-LLNL people.

jeremylt:
Any updates from above? Simplifying/condensing this information would be a real nice-to-have.

jedbrown:
I would like to do the. I understand LLNL wants the dates (even though they're always wrong) and to name LLNS as done in the Spack notice. I don't know if that's a "wants" as in there would be consequences for @tzanio or myself if we merge a change as below (with clear explanation in COPYRIGHT).
```c
// Copyright libCEED Developers. See the top-level COPYRIGHT file for details.
//
// SPDX-License-Identifier: BSD-2-Clause
```

jedbrown:
Or perhaps this (shed the always-wrong date and use replace "and" with "and/or").
```c
// Copyright Lawrence Livermore National Security, LLC and/or other libCEED authors.
// See the top-level COPYRIGHT file for details.
//
// SPDX-License-Identifier: BSD-2-Clause

**Comment by jeremylt** [comment]
I'd like to knock this out soon as part of our prep for the v0.10 release. @tzanio, is there any issue on the Livermore side with us trimming down the boilerplate file heading to what Jed proposed above and having all files reference the full version stored in a single file at the root of the repo? It seems to meet the intent of the requirement while making it easier for us to maintain up-to-date information.

**Comment by jeremylt** [comment]
@tzanio any updated word on what DoE will let us do here? I'm hoping to simplify these before the next CEED release

**Comment by jedbrown** [comment]
It's not about DOE, but rather LLNS (which manages our subcontract). I think this should be acceptable. I know it's not quite what they're asking for, but it's as close as we can get while being uniform and factual. https://github.com/CEED/libCEED/pull/858#issuecomment-1022721401

**Comment by tzanio** [comment]
Sorry, this slipped through the cracks.

The suggestion from LLNS is:

```
# Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and other CEED contributors
# All Rights Reserved. See the top-level COPYRIGHT and NOTICE files for details.
#
# SPDX-License-Identifier: (BSD-2-Clause)
#
# This file is part of CEED:  http://github.com/ceed.
```

**Comment by tzanio** [comment]
More details with examples here: https://software.llnl.gov/about/licenses/

**Comment by jedbrown** [comment]
Thanks. We're single-licensed (though could consider doing Rust-style dual with Apache) so can just refer to `LICENSE`, per that guidance.

Do you think it would be a serious issue to keep the text intact, but drop the date (which either becomes stale or causes dumb churn every year)?

**Comment by tzanio** [comment]
> Thanks. We're single-licensed (though could consider doing Rust-style dual with Apache) so can just refer to LICENSE, per that guidance.

Yes, but the `NOTICE` should remain, here is [another example](https://github.com/mfem/mfem/blob/master/NOTICE) from mfem.

> Do you think it would be a serious issue to keep the text intact, but drop the date (which either becomes stale or causes dumb churn every year)?

I don't know, but I will recommend keeping the date range. Even if it is stale, it is not a big deal, and updating it is pretty easy.

**Comment by jeremylt** [comment]
Updated per above

File: backends/avx/ceed-avx-blocked.c [hunk 1]
```diff
@@ -1,18 +1,9 @@
-// Copyright (c) 2017-2018, Lawrence Livermore National Security, LLC.
-// Produced at the Lawrence Livermore National Laboratory. LLNL-CODE-734707.
-// All Rights reserved. See files LICENSE and NOTICE for details.
+// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and other CEED contributors.
+// All Rights Reserved. See the top-level COPYRIGHT and NOTICE files for details.
```
jedbrown, line 2:
```suggestion
// All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
```
The copyright date range in `LICENSE` should be updated.

backends/avx/ceed-avx-blocked.c
```diff
@@ -1,18 +1,9 @@
...
 //
-// This file is part of CEED, a collection of benchmarks, miniapps, software
-// libraries and APIs for efficient high-order finite element and spectral
-// element discretizations for exascale applications. For more information and
-// source code availability see http://github.com/ceed.
+// SPDX-License-Identifier: (BSD-2-Clause)
```
jedbrown, line 4:
```suggestion
// SPDX-License-Identifier: BSD-2-Clause
```

My understanding is that parens are only when you have multiple licenses. An official example usage: https://github.com/spdx/spdx-examples/blob/5b8b8b4d533d89959f1a59bae94f2fbc3e0ff14f/example3/content/src/lib.c

jedbrown (approved):

PR #860, opened 2021-12-16
ctx - define interface for updating ctx fields

jeremylt:
If we like this interface, I can add functionality pretty quickly/easily in this PR, but I'd like to get some feedback on the API first.

The idea is that we can write code like this:

```C
CeedOperatorGetQFunction(op, &qf);
CeedQFunctionGetContext(qf, &ctx);
CeedQFunctionContextSetField(ctx, "time", t /* given to us by some solver */);
```

@karenlstengel, this would greatly simplify what you need to do for the TS IFunction.

### DIFF
--- a/include/ceed/ceed.h
+++ b/include/ceed/ceed.h
@@ -286,7 +286,6 @@ typedef enum {
   /// Major error, operation unsupported by current backend
   CEED_ERROR_UNSUPPORTED = -3,
 } CeedErrorType;
-
 CEED_EXTERN const char *const *CeedErrorTypes;
 
 /// Specify memory type
@@ -300,7 +299,6 @@ typedef enum {
   /// Memory resides on a device (corresponding to \ref Ceed resource)
   CEED_MEM_DEVICE,
 } CeedMemType;
-
 CEED_EXTERN const char *const CeedMemTypes[];
 
 CEED_EXTERN int CeedGetPreferredMemType(Ceed ceed, CeedMemType *type);
@@ -321,6 +319,7 @@ typedef enum {
   /// memory that can be freed using free(3).
   CEED_OWN_POINTER,
 } CeedCopyMode;
+CEED_EXTERN const char *const CeedCopyModes[];
 
 /// Denotes type of vector norm to be computed
 /// @ingroup CeedVector
@@ -332,8 +331,6 @@ typedef enum {
   /// L_Infinity norm: max_i |x_i|
   CEED_NORM_MAX,
 } CeedNormType;
-
-CEED_EXTERN const char *const CeedCopyModes[];
 
 CEED_EXTERN int CeedVectorCreate(Ceed ceed, CeedInt len, CeedVector *vec);
 CEED_EXTERN int CeedVectorReferenceCopy(CeedVector vec, CeedVector *vec_copy);
@@ -402,7 +399,6 @@ typedef enum {
   /// Apply the transpose
   CEED_TRANSPOSE
 } CeedTransposeMode;
-
 CEED_EXTERN const char *const CeedTransposeModes[];
 
 /// Argument for CeedElemRestrictionCreateStrided that L-vector is in
@@ -478,7 +474,6 @@ typedef enum {
   /// Using no input, evaluate quadrature weights on the reference element
   CEED_EVAL_WEIGHT = 16,
 } CeedEvalMode;
-
 CEED_EXTERN const char *const CeedEvalModes[];
 
 /// Type of quadrature; also used for location of nodes
@@ -489,7 +484,6 @@ typedef enum {
   /// Gauss-Legendre-Lobatto quadrature
   CEED_GAUSS_LOBATTO = 1,
 } CeedQuadMode;
-
 CEED_EXTERN const char *const CeedQuadModes[];
 
 /// Type of basis shape to create non-tensor H1 element basis
@@ -513,7 +507,6 @@ typedef enum {
   /// Hexehedron - 3D shape
   CEED_HEX = 3 << 16 | 6,
 } CeedElemTopology;
-
 CEED_EXTERN const char *const CeedElemTopologies[];
 
 CEED_EXTERN int CeedBasisCreateTensorH1Lagrange(Ceed ceed, CeedInt dim,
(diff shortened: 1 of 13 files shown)


include/ceed/ceed.h
```diff
@@ -622,6 +619,27 @@ CEED_EXTERN int CeedQFunctionFieldGetSize(CeedQFunctionField qf_field,
...
+
+ @param[in,out] ctx     CeedQFunctionContext data to update
+ @param[in] field_name  Name of field to update
+ @param[in] data        New value to set for CeedQFunctionContext field
+
+ @return An error code: 0 - success, otherwise - failure
+
+ @ingroup CeedQFunction
+**/
+typedef int (*CeedQFunctionContextUser)(void *ctx, const char *field_name,
```
nbeams, line 633: Would a better name for this be something like `CeedQFunctionContextSetter`?  How important is it it have `User` in the name?  I found this kind of confusing at first, like we were making the `QFunctionContext` itself a function, but in reality, the function specifically must be used for setting fields in a context.

jedbrown (commented):
Ugh, I see my comments from yesterday are still "pending". Submit again.

include/ceed/ceed.h
```diff
@@ -634,6 +652,14 @@ CEED_EXTERN int CeedQFunctionContextGetData(CeedQFunctionContext ctx,
...
 CEED_EXTERN int CeedQFunctionContextRestoreData(CeedQFunctionContext ctx,
     void *data);
+CEED_EXTERN int CeedQFunctionContextSetFieldFunction(CeedQFunctionContext ctx,
+    CeedQFunctionContextUser f);
+CEED_EXTERN int CeedQFunctionContextSetFieldDescriptions(CeedQFunctionContext ctx,
+    CeedQFunctionContextFieldDescription *field_descriptions, CeedInt num_fields);
+CEED_EXTERN int CeedQFunctionContextGetFieldDescriptions(CeedQFunctionContext ctx,
+    CeedQFunctionContextFieldDescription **field_descriptions, CeedInt *num_fields);
+CEED_EXTERN int CeedQFunctionContextSetField(CeedQFunctionContext ctx,
+    const char *field_name, void *data);
```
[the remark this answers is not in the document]
jeremylt: `offsetof` is new to me, but I think I like this better. I'll do another pass on Monday.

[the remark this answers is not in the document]
jeremylt: Haven't started on tests or using this interface, but what about the latest push?

interface/ceed-qfunctioncontext.c
```diff
@@ -19,10 +19,106 @@
...
+    return CeedError(ctx->ceed, CEED_ERROR_UNSUPPORTED,
+                     "QFunctionContext field with name \"%s\" already registered",
+                     field_name);
+  // LCOV_EXCL_STOP
+
+  // Allocate space for field data
+  if (ctx->num_fields == 0) {
+    ierr = CeedCalloc(1, &ctx->field_descriptions); CeedChk(ierr);
+  } else {
+    ierr = CeedRealloc(ctx->num_fields + 1, &ctx->field_descriptions);
```
jedbrown, line 88: Likely no big deal here, but this becomes quadratic complexity if there are many fields so usually one wants to separate the amount of space allocated (which might double each time it's exceeded) from the amount used.

include/ceed/ceed.h
```diff
@@ -624,6 +621,15 @@ CEED_EXTERN int CeedQFunctionFieldGetSize(CeedQFunctionField qf_field,
 CEED_EXTERN int CeedQFunctionFieldGetEvalMode(CeedQFunctionField qf_field,
     CeedEvalMode *eval_mode);
 
+/// Handle for object describing CeedQFunctionContext fields
+/// @ingroup CeedQFunction
+typedef struct {
+  const char *name;
+  const char *type;
```
jedbrown, line 628: Better as an enum. Can have an array of names indexed by the enum values for pretty printing.

interface/ceed-qfunctioncontext.c
```diff
@@ -19,10 +19,106 @@
...
+  }
+
+  // Copy field data
+  {
+    size_t len = strlen(field_name);
+    char *tmp;
+    ierr = CeedCalloc(len + 1, &tmp); CeedChk(ierr);
+    memcpy(tmp, field_name, len+1);
+    ctx->field_descriptions[ctx->num_fields].name = tmp;
+  }
```
jedbrown, line 99: May want a `CeedStrallocpy` like `PetscStrallocpy` for this purpose.

interface/ceed-qfunctioncontext.c
```diff
@@ -399,6 +495,147 @@ int CeedQFunctionContextRestoreData(CeedQFunctionContext ctx, void *data) {
...
+    // LCOV_EXCL_START
+    return CeedError(ctx->ceed, CEED_ERROR_UNSUPPORTED,
+                     "QFunctionContext field with name \"%s\" registered as %s, "
+                     "not registered as double",
+                     field_name, ctx->field_descriptions[field_index].type);
+  // LCOV_EXCL_STOP
+
+  char *data;
+  ierr = CeedQFunctionContextGetData(ctx, CEED_MEM_HOST, &data); CeedChk(ierr);
+  *(double *)&data[ctx->field_descriptions[field_index].offset] = value;
```
jedbrown, line 591: It's possible to make this generic and use `memcpy` in place of assignment.

interface/ceed-qfunctioncontext.c
```diff
@@ -399,6 +495,147 @@ int CeedQFunctionContextRestoreData(CeedQFunctionContext ctx, void *data) {
...
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedQFunctionContextRegisterDouble(CeedQFunctionContext ctx,
+                                       const char *field_name, size_t field_offset,
+                                       const char *field_description) {
+  return CeedQFunctionContextRegisterGeneric(ctx, field_name, field_offset,
+         field_description, "double", sizeof(double));
+}
```
jedbrown, line 515: Do we want this to always be single values or should it be an array (perhaps usually of length 1)? I honestly don't know. What's here will often be sufficient and maybe the ergonomics are worth it.

jedbrown (approved):

interface/ceed-qfunctioncontext.c
jeremylt: I'll comment more in the morning - I was wondering (honestly not sure)... If we add an array of a fixed length to the struct and then add fields
```
typedef struct {
  double forcing[3],
  double time,
} myContext;
```

Could we just add "forcing_x", "forcing_y", and "forcing_z" as separate setable fields? That feels like it should be legal.

include/ceed/ceed.h
jeremylt: Agreed, I was just being quick/lazy to get the idea running

interface/ceed-qfunctioncontext.c
jedbrown: We could, but I'm concerned about a higher level library/application writing loops that construct many strings. As an extreme, but not hypothetical example, the context could hold weights of a neural network used by CRIKit to implement the constitutive relation. It'd be nice to provide the hundreds of values in an array rather than lots of string queries.

jeremylt: Wouldn't that scenario be better served by using GetContextData though?

jedbrown: Perhaps, but that requires knowing the layout of the entire struct, not just one part. For example, suppose one is using a standard model for thermal diffusivity (some function of temperature, thus needing some coefficients), but an NN representation of stress-strain. Access by name seems to make it easier to compose and to set interpretable parameters in an input file while also supporting programmatic access to some less interpretable parameters that you probably won't list out in an input file.

jeremylt: We perhaps leave the single variants for now but design the internals to support arrays when we get to a problem where that is helpful?

jedbrown: Sounds good to me.

PR #863, opened 2021-12-21
GPU - separate common code into separate folder

jeremylt:
I still want to do a bit more, but this PR moves the compilation logic for CUDA backends into separate files, as was done with HIP.


jeremylt:
I will do another pass to check things over in the morning, but this should be the basic shape of the result for this PR. This will facilitate future PRs where we reduce repetition and decrease code complexity in the GPU backends.

jedbrown (approved):
The value of this rearrangement isn't all clear to me, but it's minor and I have no objections.

PR #865, opened 2021-12-28
Fluids - Fix bug in Stabilization (SU/SUPG)

LeilaGhaffari:
There were some bugs in the convective flux Jacobian which is fixed now. I have also added SU/SUPG to the traveling vortex problem. However, there should be some remaining issues with the impl since both traveling vortex and density current get completely distorted with `supg`&`degree=1` which should be numerically consistent.
I'd appreciate another pair of eyes to see what I am missing in the stabilization terms. Thanks!

# Traveling Vortex with no stabilization (degree=1):
euler_none_d1_x75
Ran with:
```
./navierstokes -problem euler_vortex -ts_adapt_monitor -lx 1000 -ly 1000 -lz 50 -dm_plex_box_faces 20,20,1 -ts_adapt_safety 0.8 -implicit -snes_atol 1e-3 -ksp_atol 1e-4 -ksp_rtol 1e-3 -ts_type alpha -ts_dt 2e-3 -ksp_type bcgs -pc_type lu  -log_view  -snes_monitor -ksp_monitor -snes_converged_reason -output_freq 1 -degree 1 -ts_max_time 5. -stab none 
```

# Traveling vortex with SUPG (degree=1):
euler_supg_d1_x75
Ran with:
```
./navierstokes -problem euler_vortex -ts_adapt_monitor -lx 1000 -ly 1000 -lz 50 -dm_plex_box_faces 20,20,1 -ts_adapt_safety 0.8 -implicit -snes_atol 1e-3 -ksp_atol 1e-4 -ksp_rtol 1e-3 -ts_type alpha -ts_dt 2e-3 -ksp_type bcgs -pc_type lu  -log_view  -snes_monitor -ksp_monitor -snes_converged_reason -output_freq 1 -degree 1 -ts_max_time 5. -stab supg 
```

# Density Current with no stabilization (degree=1):
dc_none_d1_x75
Ran with:
```
./navierstokes -problem density_current -lx 2000 -ly 125 -lz 1000 -rc 400. -dm_plex_box_faces 16,1,8 -bc_wall 1,2,5,6 -bc_slip_y 3,4  -mu 75. -thetaC -25. -ts_adapt_monitor -ts_adapt_safety 0.8 -snes_atol 1e-3 -ksp_atol 1e-4 -ksp_rtol 1e-3 -ts_type alpha -ts_dt 2e-3 -ksp_type bcgs -pc_type lu -snes_monitor -ksp_monitor -snes_converged_reason -log_view -ts_max_time 2. -output_freq 1 -implicit -degree 1 -stab none
```

# Density Current with SUPG (degree=1):
dc_supg_d1_x75
Ran with:
```
./navierstokes -problem density_current -lx 2000 -ly 125 -lz 1000 -rc 400. -dm_plex_box_faces 16,1,8 -bc_wall 1,2,5,6 -bc_slip_y 3,4  -mu 75. -thetaC -25. -ts_adapt_monitor -ts_adapt_safety 0.8 -snes_atol 1e-3 -ksp_atol 1e-4 -ksp_rtol 1e-3 -ts_type alpha -ts_dt 2e-3 -ksp_type bcgs -pc_type lu -snes_monitor -ksp_monitor -snes_converged_reason -log_view -ts_max_time 2. -output_freq 1 -implicit -degree 1 -stab supg
```


jedbrown:
@LeilaGhaffari Would you mind writing the intended formulation (equations) into the documentation (fluids/index.md), with citations to relevant literature? My typical reference is https://doi.org/10.1007/s10915-008-9233-5 (it has consistent notation and equations are hyperlinked so I get hover preview with evince). Anyway, it'd be helpful to describe our formulation in notation consistent with the rest of our documentation. It doesn't need to be *derived*, but we should explain any differences with the literature. This will help clarify the math and make it easier to "check" the code.

LeilaGhaffari:
@jedbrown , sure, I can update the documentation but I am not sure to what extent. We have referenced the paper you mentioned above to add the following
Screenshot from 2021-12-29 11-55-34

I just noticed that I had made a mistake in **P** which in fact includes the Jacobian of `convective` flux but I haven't specified that (F=F_conv + F_diff). Beside correcting that term, what improvements are you wanting me see?

jedbrown:
I've done some further writing on the stabilization. The docs here (and Hughes et al 2010 eq 23) describe `\tau` as a (diagonal) 3x3 matrix, but Whiting et al 2003, Hughes eq 39 and your code have it as a (diagonal) 5x5 matrix. The 2003 paper was for primitive variables and Hughes eq 39 is for entropy variables. Could you review/edit the docs I wrote and explain tau for NS? Note that I'm suggesting variational notation for expressions involving the flux Jacobian, similar to how we do in solids to avoid 4th order tensors -- the corresponding code is generally faster and I find the notation more comfortable.

image

examples/fluids/index.md
```diff
@@ -185,21 +190,75 @@ Our formulation follows {cite}`hughesetal2010`, which offers a comprehensive rev
...
+In both {eq}`eq-weak-vector-ns-su` and {eq}`eq-weak-vector-ns-supg`, $\mathcal P$ is called the *perturbation to the test-function space*, since it modifies the original Galerkin method into *SUPG* or *SU* schemes.
 It is defined as
 
 $$
-\bm{P}(\bm v) \equiv \left(\bm{\tau} \cdot \frac{\partial \bm{F} \, (\bm{q}_N)}{\partial \bm{q}_N} \right)^T \, \nabla \bm v\,,
+\mathcal P(\bm v) \equiv \left(\bm{\tau} \cdot \frac{\partial \bm{F}_{\text{adv}} (\bm{q}_N)}{\partial \bm{q}_N} \right)^T \, \nabla \bm v\,,
+$$
+
+where parameter $\bm{\tau} \in \mathbb R^{3\times 3}$ is an intrinsic time scale matrix.
+This expression contains the flux Jacobian, which we express in variational notation by differentiating the advective flux $\bm F_{\text{diff}}$ of {eq}`eq-ns-flux`
```
valeriabarra, line 208: I think you have a typo here: advective flux `$\bm F_{\text{adv}}$`

jedbrown: Fixed, thank you.

examples/fluids/qfunctions/eulervortex.h
```diff
@@ -386,14 +547,92 @@ CEED_QFUNCTION(IFunction_Euler)(void *ctx, CeedInt Q,
...
+      uX[j] = dXdx[j][0]*u[0] + dXdx[j][1]*u[1] + dXdx[j][2]*u[2];
+    const CeedScalar uiujgij = uX[0]*uX[0] + uX[1]*uX[1] + uX[2]*uX[2];
+    const CeedScalar Cc      = 1.;
+    const CeedScalar Ce      = 1.;
+    const CeedScalar f1      = rho * sqrt(uiujgij);
+    const CeedScalar TauC    = (Cc * f1) /
+                               (8. * (dXdxdXdxT[0][0] + dXdxdXdxT[1][1] + dXdxdXdxT[2][2]));
+    const CeedScalar TauM    = 1. / (f1>1. ? f1 : 1.);
+    const CeedScalar TauE    = TauM / (Ce * cv);
+    const CeedScalar Tau[5]  = {TauC, TauM, TauM, TauM, TauE};
```
jedbrown, line 604: I want to think of all the tau terms as having units of time, but these apparently don't. I recall discussing this formulation (or something quite like it), but I think we didn't write down how we got these expressions.

jedbrown:
Thanks @LeilaGhaffari. This plumbing looks good. We discussed having an initial condition with both the smooth vortex and a temperature bubble discontinuity so we can see both at once. Weak (small `c_tau`) stabilization might be okay for low-Mach turbulence, but it wouldn't surprise me if 0.5 is needed.

@KennethEJansen Do you have any specific test that you like to use to illustrate the tradeoff between oscillations and accuracy? I'm imagining a plot with axes being a norm of error over the vortex part and the max/min values in the vicinity of the thermal bubble, with a line traced by varying tau at fixed resolution. I'd like to know if this formulation is good enough or if we need a different tau (or primitive variables) to have good enough performance on your lower Mach turbulence experiments.

I recall John having a standard recommendation for c_tau as a function of the polynomial degree, but I'm having trouble finding the expression or reference. It was slightly more than just `c_tau = 0.5 / degree`.

jedbrown (approved):
- [ ] Can we automate choosing `c_tau` for arbitrary degree?
- [x] Test order of convergence for the vortex with at least linear and quadratic elements
- [ ] Test that a thermal bubble is adequately stabilized

examples/fluids/qfunctions/eulervortex.h
LeilaGhaffari: I am writing this for posterity. Yes, those terms were taken from Whiting et al 2003 which are for primitive variables. Thanks for the fix!

examples/fluids/index.md
```diff
@@ -185,21 +190,75 @@ Our formulation follows {cite}`hughesetal2010`, which offers a comprehensive rev
...
+(E + P)\diff\bm U/\rho + (\diff E + \diff P)\bm U/\rho - (E + P) \bm U/\rho^2 \diff\rho
+\end{pmatrix},
+\end{aligned}
+$$
+
+where $\diff P$ is defined by differentiating {eq}`eq-state`.
+In this notation, we may equivalently write the stabilization term as
+
+$$
+\mathcal P(\bm v)^T \bm r = \nabla \bm v \bm\tau \diff\bm F_{\text{adv}}(\bm r),
```
LeilaGhaffari, line 225: Does the parenthesis in `\diff\bm F_{\text{adv}}(\bm r)` indicate dF_adv as a function of residual or is it dF_adv multiplied by the residual?

examples/fluids/index.md
```diff
@@ -258,12 +258,12 @@ $$ (eq-test-perturbation-advdiff)
...
 See {cite}`hughesetal2010` equations 15-17 and 34-36 for further discussion of this formulation.
 
-For the Navier-Stokes and Euler equations in conservative variables, we follow {cite}`hughesetal2010` in defining a $3\times 3$ diagonal stabilization
-according to spatial criterion 2 (equation 27), whereas in the case of primitive variables, we follow {cite}`whiting2003hierarchical` in defining a $5\times 5$ diagonal stabilization consisting of
+For the Navier-Stokes and Euler equations in primitive variables, {cite}`whiting2003hierarchical` defines a $5\times 5$ diagonal stabilization consisting of
 1. continuity stabilization $\tau_c$
 2. momentum stabilization $\tau_m$
 3. energy stabilization $\tau_E$
 
+while we follow {cite}`hughesetal2010` in defining a $3\times 3$ diagonal stabilization according to spatial criterion 2 (equation 27), since our equations are in conservative form.
```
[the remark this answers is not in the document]
jedbrown: Yeah, can you state the equation from Hughes2010 and how the eigenvalues of the flux Jacobian are the wave speeds `u-a, u, u+a`, because knowing this requires a different chunk of knowledge than what we're clearly referencing here.

For derivation, one can see, for example, lambda in Toro page 88-90.
image
image

examples/fluids/index.md
jedbrown: My intent in the previous display equation was to define `dF_adv(dq; q)` as a linear function of its first argument with the current state defining the linearization (after the semicolon). This tracks with the "efficient" code I would anticipate writing and that Enzyme can generate, but I'm certainly open to other notation.

examples/fluids/index.md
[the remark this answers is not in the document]
LeilaGhaffari: Sure, will do!

examples/fluids/index.md
```diff
@@ -185,21 +190,75 @@ Our formulation follows {cite}`hughesetal2010`, which offers a comprehensive rev
...
-\bm{P}(\bm v) \equiv \left(\bm{\tau} \cdot \frac{\partial \bm{F} \, (\bm{q}_N)}{\partial \bm{q}_N} \right)^T \, \nabla \bm v\,,
+\mathcal P(\bm v) \equiv \left(\bm{\tau} \cdot \frac{\partial \bm{F}_{\text{adv}} (\bm{q}_N)}{\partial \bm{q}_N} \right)^T \, \nabla \bm v\,,
+$$
+
+where parameter $\bm{\tau} \in \mathbb R^{3\times 3}$ is an intrinsic time scale matrix.
+This expression contains the flux Jacobian, which we express in variational notation by differentiating the advective flux $\bm F_{\text{adv}}$ of {eq}`eq-ns-flux`
+
+$$
+\begin{aligned}
+\diff\bm F_{\text{adv}}(\diff\bm q; \bm q) &= \frac{\partial \bm F_{\text{adv}}}{\partial \bm q} \diff\bm q \\
```
jedbrown, line 212: @LeilaGhaffari The idea here is that `\frac{\partial \bm F_{\text{adv}}}{\partial \bm q}` is the (5,3,5) tensor that we don't actually form because we really only need its action. IMO, this makes notation and implementation for differentiation more mechanical, though it doesn't create the full tensor to be fed to an eigensolver or otherwise inspected. The function `dF(dq; q)` can be used to build the matrix by running it on `dq` chosen as columns of the identity.

examples/fluids/index.md
```diff
@@ -263,7 +263,26 @@ For the Navier-Stokes and Euler equations in primitive variables, {cite}`whiting
 2. momentum stabilization $\tau_m$
 3. energy stabilization $\tau_E$
 
-while we follow {cite}`hughesetal2010` in defining a $3\times 3$ diagonal stabilization according to spatial criterion 2 (equation 27), since our equations are in conservative form.
+However, since our equations are in conservative form, we follow {cite}`hughesetal2010` in defining a $3\times 3$ diagonal stabilization according to spatial criterion 2 (equation 27) as follows.
+
+$$
+\bm \tau = \frac{ 2 c_{\tau} \xi(\mathrm{Pe})}{\rho(\diff \bm F) \lVert \nabla_{\bm x} \bm X \rVert}
```
LeilaGhaffari, line 269: @jedbrown , is this the correct vector notation for tau?

jedbrown: I pushed an update here. `\diff\bm F` is an "increment" or "variation" in `\bm F`, which has the same shape as `\bm F` (a 5x3 matrix). We want the 5x5 matrices for each directional split. I pushed a revision of this section.

examples/fluids/index.md
```diff
@@ -185,21 +190,106 @@ Our formulation follows {cite}`hughesetal2010`, which offers a comprehensive rev
...
+A velocity vector $\bm u$ can be pulled back to the reference element as $\bm u_{\bm X} = \nabla_{\bm x}\bm X \cdot \bm u$, with units of reference length (non-dimensional) per second.
+To build intuition, consider a boundary layer element of dimension $(1, \epsilon)$, for which $\nabla_{\bm x} \bm X = \bigl(\begin{smallmatrix} 2 & \\ & 2/\epsilon \end{smallmatrix}\bigr)$.
+So a small normal component of velocity will be amplified (by a factor of the aspect ratio $1/\epsilon$) in this transformation.
+The ratio $\lVert \bm u \rVert / \lVert \bm u_{\bm X} \rVert$ is a covariant measure of (half) the element length in the direction of the velocity.
+A contravariant measure of element length in the direction of a unit vector $\hat{\bm u}$ is given by $\lVert \bigl(\nabla_{\bm X} \bm x\bigr)^T \hat{\bm n} \rVert$.
+While $\nabla_{\bm X} \bm x$ is readily computable, its inverse $\nabla_{\bm x} \bm X$ is needed directly in finite element methods and thus more convenient for our use.
+If we consider a parallelogram, the covariant measure is larger than the contravariant measure for vectors pointing between acute corners and the opposite holds for vectors between oblique corners.
```
jedbrown, line 238: Rendered:
image
The figures here are informative for a stretched element.
https://rinterested.github.io/statistics/covariant_contravariant_metric_tensor.html
The dotted lines show a parallelogram and the blue show the contravariant measure of length.
image
Given the same stretched element (and looking diagonally across it), the covariant lengths are here
image

I haven't seen these compared in the literature. Hughes2010 mentions both metrics in different places, but not their relation. They're the same for rectangular elements (diagonal stretch + rotation). If we write a paper about stretched grids, we may want to compare these.

examples/fluids/index.md
LeilaGhaffari: Thanks for the fix! For some reason, I thought we wanted to stick with vector notation.

jedbrown: We can't talk about the eigenvalues or spectral radius of a vector -- that needs to be a matrix. But we don't have to form the matrix because we know its eigenvalues analytically. Our code will work in incremental form (only dealing with state vectors and state fluxes), but we'll use this analytic property of the flux Jacobian.

LeilaGhaffari:
> * [ ]  Can we automate choosing `c_tau` for arbitrary degree?
> 
>     * [x]  Test order of convergence for the vortex with at least linear and quadratic elements
> 
>     * [x]  Test that a thermal bubble is adequately stabilized

I tried to clean up the commits. I am happy with `squash-merge` as well. The only remaining task is a automating c_tau for arbitrary degree but I haven't figured that out yet. If there are no other comments, I think this is ready to merge.

jedbrown:
I'll merge this now. Thanks!

PR #867, opened 2022-01-11
rust - looser lifetime restrictions

jeremylt:
I was getting some strange behavior where the lifetimes for the `ceed` and any objects it created were strangely interlinked. This removes those entanglements.

### DIFF
--- a/rust/libceed/src/basis.rs
+++ b/rust/libceed/src/basis.rs
@@ -144,7 +144,7 @@ impl<'a> fmt::Display for Basis<'a> {
 impl<'a> Basis<'a> {
     // Constructors
     pub fn create_tensor_H1(
-        ceed: &'a crate::Ceed,
+        ceed: &crate::Ceed,
         dim: usize,
         ncomp: usize,
         P1d: usize,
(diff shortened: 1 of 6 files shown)


jedbrown:
Deleting characters is good. Was the previous code making the objects inherit the lifetime of the `Ceed` rather than a distinct lifetime that is no longer than the `Ceed`? Do you have an example handy of what broke previously? (This could go in a test that would both be educational regarding the design and avoid making such a mistake in the future.)

jedbrown (approved):

jeremylt:
> Was the previous code making the objects inherit the lifetime of the `Ceed` rather than a distinct lifetime that is no longer than the `Ceed`

I think it was requiring that a `Ceed` live as long as the objects, but since we're keeping the `Ceed` alive via refcounts, this was adding an artificial restriction.

I'm fighting with inner mutability over in https://github.com/jeremylt/meles-rs and this artificial restriction was making my job harder. I can try to boil down to a smaller test at some point.

Side note - `meles-rs` is getting closer to working now. I have DM and libCEED setup working. Now I just need to get the matshell sorted out and make a KSP.

PR #868, opened 2022-01-13
Fluids - Add STG inflow

jrwrigh:
Add capability to use Shur et. al 2014 synthetic turbulence generation (STG) inflow boundary condition to the fluids examples.

TODO:
- [x] Rebase onto main
- [x] Fix up documentation
    - [x] \kappa_e vs \kappa^e
    - [x] Fix up modify mesh documentation
- [x] Fix `!!weakT` in blasius


jedbrown:
Looks reasonable. Please write in complete sentences and cite the Shur paper (add it to `references.bib`). It isn't totally clear from the summary here how \bar u and R are specified. I take it they are analytic functions of the distance y from the wall? Anything that can logically be a tunable/run-time parameter should use `PetscOptionsRealArray` or similar to obtain. (You can set a default, but that allows specifying it at the command line or in a yaml input file.)

For others, this is the direct link to the rendered docs. It'll be updated each time James pushes.
https://libceed--868.org.readthedocs.build/en/868/examples/fluids/#flat-plate-boundary-layer

KennethEJansen:
> Looks reasonable. Please write in complete sentences and cite the Shur paper (add it to `references.bib`). It isn't totally clear from the summary here how \bar u and R are specified. I take it they are analytic functions of the distance y from the wall? Anything that can logically be a tunable/run-time parameter should use `PetscOptionsRealArray` or similar to obtain. (You can set a default, but that allows specifying it at the command line or in a yaml input file.)
> 
> For others, this is the direct link to the rendered docs. It'll be updated each time James pushes. https://libceed--868.org.readthedocs.build/en/868/examples/fluids/#flat-plate-boundary-layer

\bar u is the mean velocity profile (the components) that we read from file.  R is the Reynolds stress tensor (read from the same file).  At least for the flat plate and other span wise homogenous flows these would be read in a  as a table along with a y coordinate or a distance to the wall and this would be interpolated to any point you want to add the fluctation to get an unsteady inflow with a prescribed mean.    Note in this pass James is leaving out T but  there would be a bar T to read in as well for the same y/wall distance points.

jedbrown:
Okay, so what's called "Init Func Data" here would also be interpolating those profiles from the ~200 point table to the quadrature points. It's all your choice on format. Somewhat less filesystem metadata for a single rank to read these and broadcast. (You could use MPI-IO, perhaps via HDF5, but it's not much data. Not a big deal because it's one file read once.)

KennethEJansen:
@jrwrigh made a good start but I think it is probably good to, at least in the short term, keep a clear separation of the two input data sources.  We may decide to merge them later but that should come after it is clear what they are. 

We started off discussing the random number file/table.  This is the data numbered one to $$n$$, the number of wave numbers.  

Then there is a second file/table that has profile data.  The md document needs to be enhanced to document what comes from there AND how that file is generated.  I know Jed prefers to express as much stuff analytically in the code as possible which I mostly agree with.  Our past practice should be improved to reflect that to the greatest extent possible.  For example, while we do not have analytic expressions for \bar u  or the Reynolds stresses (they come from other solutions as noted above), PHASTA currently reads in 14 columns of data in that profile table which is generated by another MATLAB helper file.  At least some of that data is derived data but James should review and potentially call in help from Riccardo and or I if necessary to be more clear on how things like $$l_t$$ are computed.  My guess is that while we could in the short term continue to make that  file that contains the 14 columns of profile data (we call it STGInflow.dat) that might not be the best long term plan.  Note some of the columns don't come into play until we try to do IDDES which has 1 or 2 more columns depending upon whether you are solving one or two transport equations for turbulence model variables.  I suppose that accounts for 12 of the 14 columns already: dw(1), velocity (3), stress(6), turbVars(2).  So, it is probably just $$l_t$$ and one other that are "extra".

jrwrigh:
I've made a few additions/corrections to the documentation. Namely, specing out the `STGInflow.dat` and `STGRand.dat` files that would (presumably) be used here. I've made that note in the data flowchart as well, and added notes to arrows to hopefully clarify the actions being done in the initialization step.

jrwrigh:
I've uploaded an "outline" of the general data architecture I'm thinking of using. I'd appreciate any feedback on whether this is what y'all had in mind, ideas, or even corrections to my (probably subpar) C code.  I've commented a decent bit of the code to hopefully fill in the gaps, but let me know if anything doesn't make sense or is confusing.

jrwrigh:
One outlying part of the implementation is with the element size. For right now, I think just the bounding box of the element is fine, especially since we're dealing with meshes that are well-aligned with the global coordinate system.

I'm not sure what the best way of getting that data from PETSc is. I can dig into the DMPlex functions to see if there's anything useful there, but any direction would be helpful as I'm in very unfamiliar territory with PETSc. 


Also, once I have element sizes, I'm not sure how to get them into a relevant format for the qfunction. Are q functions are aware of which quadrature point belongs to what element? If so, has that been implemented in libCeed anywhere? If not, any ideas on how to get a element-wise quantity to quadrature points?

knepley:
My understanding of the CEED workflow was that there is a preliminary geometry kernel that computes everything you want, including the cell size, and stick it onto quadrature points. There are functions in Plex that do this, but they are not setup to be called from a QFunction. Maybe we should refactor to enable this.

examples/fluids/qfunctions/stg_shur14.h
```diff
@@ -0,0 +1,133 @@
...
+int SetupSTG_Rand(STGShur14Context stg_ctx){
+
+  //TODO will probably want to have the paths to the STGRand.dat and
+  // STGInflow.dat as inputs for this function.
+
+  stg_ctx->alpha = 1.01; // Get from CLI, yaml/toml, etc.
+
+  //TODO Read STGRand.dat
+  stg_ctx->nmodes = 2;
+  stg_ctx->sigma = malloc(sizeof(CeedScalar)*stg_ctx->nmodes*3);
```
jeremylt, line 59: I'm not understanding why you would be mallocing inside of a QFunction? I don't think that will be GPU friendly.

jeremylt: This QFunction code will get called multiple times for the same application of the FE operator in most backends, so I really don't think this is the approach you want to take. This will leak like a sieve.

jeremylt:
> For right now, I think just the bounding box of the element is fine

Do you want points *outside* and "inside" of each element?

> Are q functions are aware of which quadrature point belongs to what element

No. I'm not sure exactly why you want this? The ElemRestriction and Operator orders the arrays fed into the QFunction so that indexing by quadrature point `q` gives you the data for that quadrature point. The assumption is that the QFunction is independent at each quadrature point.

Have you seen this diagram?
https://libceed.readthedocs.io/en/latest/libCEEDapi/#finite-element-operator-decomposition

Big picture, inputs and outputs are stored in Vectors. Data for a given element is extracted with an ElemRestriction. A Basis provides interpolation/gradients at quadrature points as needed. The QFunction acts independently on each quadrature point. Then a Basis applies the transpose of the test functions and an ElemRestriction sums the result into an output Vector.

If you're going to be storing lots of data per element, we usually store that data in a Vector and create a Strided ElemRestriction to index into the Vector.

examples/fluids/qfunctions/stg_shur14.h
jrwrigh: This isn't a QFunction. It's a "normal" function that will be called only once in the program. Though I can see now that it being in the `qfunctions` directory might lead to the confusion. Would `src` would be a better place to move this (and `TearDownSTG`)?

jrwrigh:
> My understanding of the CEED workflow was that there is a preliminary geometry kernel that computes everything you want, including the cell size, and stick it onto quadrature points. There are functions in Plex that do this, 

That's precisely what I'm after! Cell size "stuck" onto quadrature points.

> but they are not setup to be called from a QFunction. Maybe we should refactor to enable this.

I don't think a refactor is necessary. I think it makes sense to have them be outside the QFunction and is what I was really wanting/needing for this particular application

examples/fluids/qfunctions/stg_shur14.h
jeremylt: Ah, that makes more sense. It would be best to move it. It's best to put only QFunctions in these QFunction headers. It makes JiT easier. 

This still will be a problem on the GPU though. The context data can't be copied to the GPU with a struct that only has pointers and not the data.

jedbrown: Yeah, better to put it in `src` -- that's where setup stuff goes. This header will end up being read by the JIT so it should be as slim as possible.

jeremylt:
> Cell size "stuck" onto quadrature points

What do you mean by this? The change of coordinates and the Jacobian? That's in the other QFunctions as 'qdata'.

examples/fluids/qfunctions/stg_shur14.h
jedbrown: That said, we'd rather this be an array inside the `struct` (the last one can have dynamic size, allocated as `malloc(sizeof(struct S) + n*sizeof(double))`) instead of a pointer to other data. The reason being that when a context is copied to the GPU, we'd rather it be plain data rather than needing deep copies.

examples/fluids/qfunctions/stg_shur14.h
```diff
@@ -0,0 +1,133 @@
...
+  CeedScalar *sigma, *d, *phi; // !< Random number set, [nmodes,3], [nmodes,3], [nmodes]
+  CeedScalar *kappa;           // !< Wavemode frequencies, [nmodes]
+  CeedScalar alpha;            // !< Geometric growth rate of kappa
+
+  CeedInt nprof;       // !< Number of profile points in STGInflow.dat
+  CeedScalar *prof_dw; // !< Distance to wall for Inflow Profie, [nprof]
+  CeedScalar *ubar;    // !< Mean velocity, [nprof, 3]
+  CeedScalar *cij;     // !< Cholesky decomposition [nprof, 6]
+  CeedScalar *eps;     // !< Turbulent Disspation [nprof, 6]
+  CeedScalar *lt;      // !< Tubulent Length Scale [nprof, 6]
```
jedbrown, line 46: What you're doing here will work fine on a CPU, but it will need some changes for a GPU. Not a big deal to do then, but if you want to do it now, one approach is
```suggestion
  CeedInt nmodes;              // !< Number of wavemodes
  CeedScalar alpha;            // !< Geometric growth rate of kappa
  CeedInt nprof;       // !< Number of profile points in STGInflow.dat
  struct {
    CeedInt sigma, d, phi;
    CeedInt prof_dw, ubar, cij, eps, lt;
  } offsets;
  CeedScalar data[];
```

and then access as
```c
const CeedScalar *sigma = &ctx->data[ctx->offsets.sigma];
```
This is called a [flexible array member](https://en.wikipedia.org/wiki/Flexible_array_member). Note that it's only strictly a C feature, but in practice, g++ and clang++ accept it at `-Wall -Wextra`, and only warn at `-Wpedantic`.

jedbrown:
For local element size, would one of these definitions be okay?
image

jrwrigh:
> What do you mean by this? The change of coordinates and the Jacobian? That's in the other QFunctions as 'qdata'.

I don't need the change of coordinates and the Jacobian, but the bounding box of the element itself. I feel like there's probably a way to calculate the bounding box from the change of coordinates and/or the Jacobian, but I'm not entirely sure. 

To explicitly define what I mean by bounding box, here's a Python code-equivalent:

```python
x = np.rand.random((8,3)) # coordinates of nodes of the element, can be more/less than 8
bounding_box = np.zeros((3))
for i in range(3):
	bounding_box[i] = np.abs( x[:,i].max() - x[:,i].min() )
```

jrwrigh:
> For local element size, would one of these definitions be okay?

It probably would be, but since we're going to be comparing directly to PHASTA, I figured we should(?) try to stick as closely to what PHASTA currently does, which is element bounding box.

jedbrown:
Is the bounding box still what you want as the FE degree increases? In this test problem where we're comparing, are the elements affine?

jedbrown:
Also, does it bother you that this definition of bounding box isn't equivariant under rotation, even for an orthogonal, affine element?

The covariant and contravariant definitions proposed above are equivalent for orthogonal elements. They'll agree with the bounding box definition if the elements are also grid aligned and affine (as for lowest order element geometry).

jrwrigh:
So for this first push on STG/Flat Plate, my understanding is that we're effectively trying to make this a side-by-side comparison to PHASTA. So while there maybe better mathematical implementations of the methods to get STG running, the end goal is to match PHASTA. If I'm wrong in that understanding, please let me know, as I share many of the concerns that y'all have outlined here. @KennethEJansen I think your insight may be useful here.

> Is the bounding box still what you want as the FE degree increases? 

To keep aligned with PHASTA, yes. We had [discussed in Zulip](https://phypid.zulipchat.com/#narrow/stream/232375-ceed-phasta/topic/mesh/near/266465917) about possible adding a coefficient handle the fact that as polynomial order increases, the effective cutoff lengthscale decreases. As a rough estimate, we could make it be 1/p (though we'd expect the rate of decrease to be higher than 1/p).

> In this test problem where we're comparing, are the elements affine?

I am assuming affine elements here. Again, more as a comparison to PHASTA rather than something to take into the future.

> Also, does it bother you that this definition of bounding box isn't equivariant under rotation, even for an orthogonal, affine element?

Yes, it does bother me. But that is the current implementation in PHASTA.

> The covariant and contravariant definitions proposed above are equivalent for orthogonal elements. They'll agree with the bounding box definition if the elements are also grid aligned and affine (as for lowest order element geometry).

Sweet! I think there was also discussion about possibly also doing unstructured meshes for the Flat Plate problem as well. It looks like this would hold for affine tetrahedral/pyramidal elements as well?

examples/fluids/qfunctions/stg_shur14.h
jrwrigh: Ok, cool. Just so I understand correctly, if we assume that the data for `sigma`, `d`, and `phi` follow each other (which are sizes `[nmodes, 3]`, `[nmodes,3]`, `[nmodes]`, repsectively), then the offsets would be:

```c
ctx->offsets.sigma = 0
ctx->offsets.d = ctx->nmodes * 3
ctx->offsets.phi = ctx->offsets.d + ctx->nmodes*3
ctx->offsets.prof_dw = ctx->offsets.phi + ctx->nmodes // Or whatever data comes after phi
```
?

jedbrown: Yes to both. Sorry it's a bit clumsy -- we could give it some interface sugar, but let's not get distracted by that just now.

jedbrown:
We won't match PHASTA exactly because PHASTA uses primitive variables. (Unless we add a primitive variable solver in libCEED, which is also possible.)

If the elements are affine and grid-aligned, the definitions will be the same so I think we can stick with what we're currently using (the covariant one).

As for tet elements, it's measuring relative to the reference element. But the reference tet is usually chosen to not be invariant under rotations. This might be a reason why we should choose a reference tet that is symmetric (so its topological symmetry group matches its geometric symmetry group). This can be done when defining the basis, but it's something we should think about when we start running these h-dependent models on tets (and prisms).

jedbrown:
As a simple mental check, if you take a 2D mesh of unit-length squares and slash it into triangles, what lengths would you like to report looking along the original axes, looking diagonally across the slashes, and looking diagonally along the slashes? Now what if the triangles are not orthogonal? (We can investigate this to a separate issue or a Zulip conversation.)

knepley:
@jedbrown I do not think you gain much by moving to an equilateral tet reference cell. You still cannot tile space with that, so your real cells will always be as distorted as the original reference.

jedbrown:
@knepley Right, but we're measuring length of the physical element by how much it's distorted from the reference element. In the example with triangles from slashed squares, I think you don't want to report length 1 in every direction.

knepley:
@jedbrown I guess it depends how you want it interpreted in your method. If you really want the distortion from equilateral. You can just compose it with the known Jacobian of your reference cell, rather than changing the reference.

jedbrown:
> If you really want the distortion from equilateral. You can just compose it with the known Jacobian of your reference cell, rather than changing the reference.

Yes, I would preprocess the PetscFE (which becomes a `CeedBasis`) to apply this transformation so that the local element Jacobian can be used as a rotationally equivariant measure of element size.

examples/fluids/qfunctions/stg_shur14.h
jrwrigh: No worries. I just wasn't familiar with that construct. I believe I've resolved this though with the most recent commits.

examples/fluids/qfunctions/stg_shur14.h
```diff
@@ -0,0 +1,63 @@
...
+/// STGShur14_CalcQF is run over quadrature points. Before the program exits,
+/// TearDownSTG is run to free the memory of the allocated arrays.
+
+#ifndef stg_shur14_h
+#define stg_shur14_h
+
+#include <math.h>
+#include <ceed.h>
+#include <stdlib.h>
+#include "../src/setupstg_shur14.h"
```
jedbrown, line 31: This (the struct definition) can be inlined here and `src/setupstg_shur14.h` can be deleted.

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,134 @@
...
+  CeedPragmaSIMD
+  for(int i=0; i<nprofs; i++){
+    Cij[0][i] = sqrt(Rij[0][i]);
+    Cij[3][i] = Rij[3][i] / Cij[0][i];
+    Cij[1][i] = sqrt(Rij[1][i] - pow(Cij[3][i], 2) );
+    Cij[4][i] = Rij[4][i] / Cij[0][i];
+    Cij[5][i] = (Rij[5][i] - Cij[3][i]*Cij[4][i]) / Cij[1][i];
+    Cij[2][i] = sqrt(Rij[2][i] - pow(Cij[4][i], 2) - pow(Cij[5][i], 2));
+  }
+return 1;
```
jedbrown, line 62: Most common `C` convention is to return `0` on success, or just return `void` if there is no error case.

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,134 @@
...
+}
+
+
+int SetupSTG_Rand(STGShur14Context stg_ctx){
+
+  //TODO will probably want to have the paths to the STGRand.dat and
+  // STGInflow.dat as inputs for this function.
```
jedbrown, line 69: That or you can use `PetscOptionsString` and the like so it can be specified via command-line or yaml file.

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,134 @@
...
+
+  stg_ctx->offsets.sigma = 0;
+  stg_ctx->offsets.d       = nmodes*3;
+  stg_ctx->offsets.phi     = stg_ctx->offsets.d       + nmodes*3;
+  stg_ctx->offsets.kappa   = stg_ctx->offsets.phi     + nmodes;
+  stg_ctx->offsets.prof_dw = stg_ctx->offsets.kappa   + nmodes;
+  stg_ctx->offsets.ubar    = stg_ctx->offsets.prof_dw + nprofs;
+  stg_ctx->offsets.cij     = stg_ctx->offsets.ubar    + nprofs*3;
+  stg_ctx->offsets.eps     = stg_ctx->offsets.ubar    + nprofs*6;
+  stg_ctx->offsets.lt      = stg_ctx->offsets.eps     + nprofs;
```
jedbrown, line 91: I think the size calculation is too small (misses one `nmodes`), but here's a different way to initialize that might be less error-prone.
```suggestion
  {
    STGShur14Context s;
    s.nmodes = nmodes;
    s.nprofs = nprofs;
    s.offsets.sigma = 0;
    s.offsets.d       = nmodes*3;
    s.offsets.phi     = stg_ctx->offsets.d       + nmodes*3;
    s.offsets.kappa   = stg_ctx->offsets.phi     + nmodes;
    s.offsets.prof_dw = stg_ctx->offsets.kappa   + nmodes;
    s.offsets.ubar    = stg_ctx->offsets.prof_dw + nprofs;
    s.offsets.cij     = stg_ctx->offsets.ubar    + nprofs*3;
    s.offsets.eps     = stg_ctx->offsets.ubar    + nprofs*6;
    s.offsets.lt      = stg_ctx->offsets.eps     + nprofs;
    int total_num_scalars = stg_ctx->offsets.lt + nprofs;
    stg_ctx = malloc(sizeof(*stg_ctx) + total_num_scalars*sizeof(stg_ctx->data[0]));
    *stg_ctx = s;
  }
```

examples/fluids/src/setupstg_shur14.h
```diff
@@ -0,0 +1,51 @@
...
+  CeedScalar alpha; // !< Geometric growth rate of kappa
+
+  struct {
+  CeedInt sigma, d, phi; // !< Random number set, [nmodes,3], [nmodes,3], [nmodes]
+  CeedInt kappa;   // !< Wavemode frequencies in increasing order, [nmodes]
+  CeedInt prof_dw; // !< Distance to wall for Inflow Profie, [nprof]
+  CeedInt ubar;    // !< Mean velocity, [nprof, 3]
+  CeedInt cij;     // !< Cholesky decomposition [nprof, 6]
+  CeedInt eps;     // !< Turbulent Disspation [nprof, 6]
+  CeedInt lt;      // !< Tubulent Length Scale [nprof, 6]
```
jedbrown, line 43: Indent (`make style`). These could be `size_t`, though not necessary.

jedbrown (commented):
Looks good so far.

examples/fluids/qfunctions/stg_shur14.h
jrwrigh: Ok. I was going more on a DRY principle. For my own understanding, is there a downside to defining the struct in a single file and then `#include`ing it where needed?

examples/fluids/src/setupstg_shur14.c
jrwrigh: Gotcha. I had that initially based on experience in shell programming, but saw that C interprets `0` as false and everything else as true.

examples/fluids/src/setupstg_shur14.c
jrwrigh: Yep, you're right. Thanks for the suggestion!

examples/fluids/src/setupstg_shur14.h
jrwrigh: Is there any advantage/disadvantage to using `size_t` vs `int`/`CeedInt`?

examples/fluids/src/setupstg_shur14.c
jeremylt: Yeah, so the convention is 
```C
error_code = MyFunctionCall(with, some, args);
if (error_code) {
  // then do error handling
}
```

examples/fluids/qfunctions/stg_shur14.h
jeremylt: Personally, I'll put the struct in a separate header if its going to be used in multiple QFunctions and keep in with the QFunction header if it is only used in that one QFunction. Less opening files to see the whole picture that way.

examples/fluids/src/setupstg_shur14.h
jeremylt: `size_t` is strictly positive, so I use it for array/allocation sizes.

jedbrown: Yeah, `size_t` is always positive and can't overflow when measuring sizes. In contrast, `int` can overflow. However, using `size_t` as a loop index is problematic because `for (size_t i=n-1; i >= 0; i--)` is an infinite loop. Also, `size_t` doesn't have an obvious flag value to indicate "not set" or "ignore this value". So PETSc uses signed integers for indices, even when negative values don't have a valid meaning. Rust (for example) allows you to make "niche" types that can be safer, though they're also louder.

examples/fluids/qfunctions/stg_shur14.h
jedbrown: This header is currently only included in one place. It may be problematic to find such headers during JIT compilation so I'd say this indirection is not necessary -- just put the struct here.

jrwrigh: That makes sense to me. Didn't know if it had any JIT/backend implications I wasn't aware of. In this case the struct will only be used in this QFunction, so I'll move it to `qfunctions/stg_shur14.h` and then `#include` it for `src/setupstg_shur14.c`.

> ... and `src/setupstg_shur14.h` can be deleted.

> This header is currently only included in one place.

`SetupSTGContext()` (and `TearDownSTG()` if that ends up getting used) is going to be called externally, which is why I initially made `src/setupstg_shur14.h`. I now see that the other setup function instead put their function signatures into `navierstokes.h`. Should I follow that pattern?

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,136 @@
...
+  CeedPragmaSIMD
+  for(int i=0; i<nprofs; i++) {
+    Cij[0][i] = sqrt(Rij[0][i]);
+    Cij[3][i] = Rij[3][i] / Cij[0][i];
+    Cij[1][i] = sqrt(Rij[1][i] - pow(Cij[3][i], 2) );
+    Cij[4][i] = Rij[4][i] / Cij[0][i];
+    Cij[5][i] = (Rij[5][i] - Cij[3][i]*Cij[4][i]) / Cij[1][i];
+    Cij[2][i] = sqrt(Rij[2][i] - pow(Cij[4][i], 2) - pow(Cij[5][i], 2));
+  }
+}
```
jrwrigh, line 63: Not to fall into the premature optimization trap, but I have a general question about n-dimensional array layouts. Above, I have the index being looped over as the last index so `Cij[5][0]` is non-contiguous to `Cij[4][0]` despite being looped together. This is against my general intuition, which is to keep data that is used in same loop iteration contiguous.  Looking at the assembly of the above function with both array orientations (`[6][nprofs]` and `[nprofs][6]`), having the looped index be the first array index (`[nprofs][6]`) results in less total operations and move vectorized operations. 

However, I wrote the function with `[6][nprofs]` orientation as this is what QFunctions with n-dimensional array data do (just replace `nprofs` with `CEED_Q_VLA`) https://github.com/CEED/libCEED/blob/9477ba7d4a95edebf665b0c36c17bb403727d847/examples/fluids/qfunctions/eulervortex.h#L314-L319.

Is there reason that the QFunctions use (to my eye) a less easily vectorizable data structure? Is my view point too simplified? Is it a limitation of VLA?

jedbrown: For our state variables, we want the quadrature point last so every operation is vectorized (uses SIMD instructions) over that dimension. Since all quadrature points are independent, that's good. For memory locality, you want data that you use together to be close. That's why (on CPU), the QFunction gets called in batches of elements that fit in (usually L1 or L2) cache. All the constant size loops (usually 3 or 5) inside the QFunction get unrolled, and the loop over quadrature points also gets unrolled (usually by 4 or 8) to use these vector instructions.

Some operations will prevent the compiler from vectorizing. We hit that with `log1p` in solids, and wrote our own log1p that's accurate enough in the relevant range. Variable length inner loops are another that will often make the compiler give up on vectorizing. We can profile and optimize this if it's an issue. The main way to do that would be to sum the series at a batch of points (e.g., 8), ensuring vectorization over that inner loop. With that, your `Cij` would only need to be accessed as scalars anyway, thus the ordering won't really matter, and interlacing the components (`[nprofs][6]`) will give somewhat better L1 cache reuse.

jedbrown: If you want to do it that way, I would make a struct with `Cij` and the other terms indexed up to `nprofs`, then the other. It'll be fewer offsets (maybe just two) and thus slightly fewer integer registers needed for indexing. (Very much premature optimization, but the code might be more clear if written this way.) I would just focus on writing clear code right now -- I have a couple theories on how best to optimize it, but that'll be easier if we can test our hypothesis with working code, and also, it might not need that much attention.

jrwrigh: Ok, that makes sense. I'll probably just leave the structures as is for right now, though I agree it'd probably make the code a little more clear.

examples/fluids/src/setupstg_shur14.h
jrwrigh: Changed to size_t in 0b45cf5

jrwrigh:
I've now added functions to read in the `STGInflow.dat` (containing the boundary layer profile data) and `STGRand.dat` (the random numbers required for each wavemode). The file reading mechanism is fairly fragile; it requires there to be exactly one space between numbers. I'd imagine there's probably a better way of doing this; please let me know y'all's thoughts.

I've also uploaded a mock `STGInflow.dat`, `STGRand.dat`, and code for running `SetupSTGContext` using the `-stg` flag for testing. I'm guessing the more permanent location would be within a new boundary layer "problem". Is there a plan set for creating that structure right now? Is that something I should be looking at picking up?

Lastly, is there any unit testing setup for the examples? I see evidence of regression testing (though I honestly don't know what reads/uses the `// TESTARGS` comments in `navierstokes.c`), but nothing on unit testing. I find it useful to make sure I don't accidentally break code as I continue to move forward, but if there's no desire for that in the examples, no worries.

jrwrigh:
I've also remembered that I still need to implement the MPI communication of the file data as well. Which will happen in two steps, communicate `nmodes` and `nprofs` so each rank can allocate memory, and then communicate `stg_ctx->data`.

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,333 @@
...
+    CeedScalar *eps = &stg_ctx->data[stg_ctx->offsets.eps];
+    CeedScalar *lt = &stg_ctx->data[stg_ctx->offsets.lt];
+    CeedScalar (*ubar)[stg_ctx->nprofs] = (CeedScalar (*)[stg_ctx->nprofs])
+                                          &stg_ctx->data[stg_ctx->offsets.ubar];
+
+    for (int i=0; i<stg_ctx->nprofs; i++) {
+      if ( fgets(line, char_array_len, fp) != NULL) {
+        ierr = PetscStrToArray(line, ' ', &ndims, &array); CHKERRQ(ierr);
+        if(ndims < dims[1]) {
+          printf("Line %d of %s does not contain enough columns (%d instead of %d)", i,
```
jedbrown, line 161: use `SETERRQ` here.

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,333 @@
...
+  FILE *fp;
+  const int char_array_len = 512;
+  char line[char_array_len];
+  MPI_Comm comm = PETSC_COMM_WORLD;
+  char **array;
+
+  PetscFunctionBeginUser;
+  ierr = PetscFOpen(comm, path, "r", &fp); CHKERRQ(ierr);
+
+  if ( fgets(line, char_array_len, fp) != NULL) {
```
jedbrown, line 91: You can use `PetscSynchronizedFGets` for the reads, which allows all ranks to pass through the same code (and then you don't have to broadcast later). It does the broadcast internally, which is fine for a file of this size (relatively small number of lines). If not, then best not to use collective `PetscFOpen` above.

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,333 @@
...
+ */
+static inline PetscErrorCode ReadSTGInflow(const char path[PETSC_MAX_PATH_LEN],
+    STGShur14Context stg_ctx) {
+
+  PetscErrorCode ierr;
+  int ndims, dims[2];
+  FILE *fp;
+  const int char_array_len = 512;
+  char line[char_array_len];
+  MPI_Comm comm = PETSC_COMM_WORLD;
```
jedbrown, line 127: Best to pass the communicator in as a function argument rather than guess inside.

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,333 @@
...
+    s->offsets.cij     = s->offsets.ubar    + nprofs*3;
+    s->offsets.eps     = s->offsets.cij     + nprofs*6;
+    s->offsets.lt      = s->offsets.eps     + nprofs;
+    int total_num_scalars = s->offsets.lt + nprofs;
+    stg_ctx = malloc(sizeof(*stg_ctx) + total_num_scalars*sizeof(stg_ctx->data[0]));
+    *stg_ctx = *s;
+  }
+
+  stg_ctx->alpha = 1.01;
+  ierr = PetscOptionsGetReal(NULL, NULL, "-stg_alpha", &stg_ctx->alpha, NULL);
```
jedbrown, line 300: `PetscOptionsBegin` and `PetscOptionsReal` would allow these options to be displayed in `-help` output.

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,333 @@
...
+
+#ifndef max
+#define max(a,b)            (((a) > (b)) ? (a) : (b))
+#endif
+
+#ifndef min
+#define min(a,b)            (((a) < (b)) ? (a) : (b))
+#endif
+
+#endif  /* NOMINMAX */
```
jedbrown, line 41: Looks like these aren't used? There are `PetscMin` with the same definition.

jedbrown:
We don't have a unit testing framework for examples, just the regression tests. You can of course use asserts anywhere within the examples, and the setup here won't be performance sensitive so feel free to put in any consistency checks you like. The `//TESTARGS` is read by `tap.sh` (for `make test` and `make prove`) and `junit.py` (for `make junit`).

examples/fluids/src/setupstg_shur14.c
jrwrigh: `max` is used in the calculation of kappa, but `min` isn't used, correct. Didn't realize Petsc already had those definitions (is there anything that PETSc can't do?? :joy:).

examples/fluids/src/setupstg_shur14.c
jrwrigh: Ah, didn't realize that. I was just going off the PETSc user manual, which only mentions the `PetscOptionsGet*` functions. Does it also pass the default value through to `-help`?

examples/fluids/src/setupstg_shur14.c
jrwrigh: Ah. The documentation for `PetscFOpen` says only the first rank actually opens the file, while `PetscSynchronizedFGets` says all the processors get it. I had interpreted that as  `PetscSynchronizedFGets` has all the ranks read the same file at once, but synchronizes the line reading. Hindsight, it makes sense that a HPC library wouldn't include a function that does something like that.

examples/fluids/src/setupstg_shur14.c
jedbrown: Yeah, `-help` will show the default at the time that code is reached. We should encourage people to use these in the users manual.

examples/fluids/src/setupstg_shur14.c
jedbrown: Right, nothing slows a supercomputer to a crawl like the filesystem metadata impact of every rank opening files (the same or different files). The bcasts are in the tens of microseconds range and thus no big deal to read once that way until you have many thousands of lines. It's somewhat faster to read larger blocks and broadcast, or create the whole data structure and broadcast, but sometimes that code gets out of sync with a changing file format and it's generally not worth the human time for smallish files.

jrwrigh:
Assuming the broken test is actually fine, all the preprocessing (that I can see) is done now. The next step would be implementing the QFunction itself. However, to do that I need a problem to apply it to (or more specifically an operator to add it too). Should I go ahead and mock up a framework for the boundary layer problem? Or is someone already working on that?

jeremylt:
Note - this branch needs the updates in main for the latest development version of PETSc: #940

jrwrigh:
Should be good to go now!

doc/sphinx/source/references.bib
```diff
@@ -190,3 +189,15 @@ @phdthesis{whitingStabilizedFEM1999
...
+
+@article{shurSTG2014,
+  title = {Synthetic {{Turbulence Generators}} for {{RANS-LES Interfaces}} in {{Zonal Simulations}} of {{Aerodynamic}} and {{Aeroacoustic Problems}}},
+  author = {Shur, Michael L. and Spalart, Philippe R. and Strelets, Michael K. and Travin, Andrey K.},
+  year = {2014},
+  journal = {Flow, Turbulence and Combustion},
+  volume = {93},
+  number = {1},
+  pages = {63--92},
+  doi = {10/f58pzd},
```
jedbrown, line 201: I don't understand how doi.org/10/f58pzd resolves, but the publisher's site says
```suggestion
  doi = {10.1007/s10494-014-9534-8},
```

jedbrown (approved):
Great, just super minor stuff.

examples/fluids/qfunctions/setupgeo.h
```diff
@@ -134,6 +141,11 @@ CEED_QFUNCTION(Setup)(void *ctx, CeedInt Q,
 //                [ J2 ]
 //                [ J3 ]
 //
+// Stored: dxdX_{i,j}
```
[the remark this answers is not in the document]
jrwrigh: Gotcha. I'll leave that as a future TODO for now.

examples/fluids/blasius.yaml
```diff
@@ -48,3 +48,7 @@ wall_comps: 1,2,3
 bc_inflow: 6
 bc_outflow: 5,4
 g: 0,0,0
+
+stg:
+  inflow_path: "./STGInflow_physical.dat"
```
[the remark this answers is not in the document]
jrwrigh: Ah, I included an example `STGRand.dat`, but not an `STGInflow.dat`. I'll add the example (read, not physically relevant) version of the `STGInflow.dat`. I'll have another PR to get the physically realistic ones in there, along with some mesh generation improvements.

doc/sphinx/source/references.bib
jrwrigh: It's a [short DOI](https://shortdoi.org/), which is an official way of getting shorter DOI links. Nicer for actual paper publications (less noise in citation, easier to re-type if copy/paste isn't an option), but I think either works perfectly fine here.

jedbrown:
Cool. @jeremylt Can you give a high level review?

@jrwrigh Is it possible to have a minimal test (comparing with reference output) to confirm that nothing major is broken? Also, would you like to squash selectively or squash-merge the whole branch?

examples/fluids/problems/blasius.c
```diff
@@ -76,11 +78,14 @@ PetscErrorCode modifyMesh(DM dm, PetscInt dim, PetscReal growth, PetscInt N,
 PetscErrorCode NS_BLASIUS(ProblemData *problem, DM dm, void *ctx) {
 
   PetscInt ierr;
-  User              user = *(User *)ctx;
-  MPI_Comm          comm = PETSC_COMM_WORLD;
-  BlasiusContext    blasius_ctx;
+  User           user     = *(User *)ctx;
+  MPI_Comm       comm     = PETSC_COMM_WORLD;
+  PetscBool      stg_bool = PETSC_FALSE;
```
jeremylt, line 83: Our convention for booleans is `verb_noun`, so probably in this case `use_stg`?

examples/fluids/problems/blasius.c
```diff
@@ -130,8 +133,18 @@ PetscErrorCode NS_BLASIUS(ProblemData *problem, DM dm, void *ctx) {
   ierr = PetscOptionsScalar("-top_angle",
                             "Geometric top_angle rate of boundary layer mesh",
                             NULL, top_angle, &top_angle, NULL); CHKERRQ(ierr);
+  ierr = PetscOptionsBool("-stg", "Use STG inflow boundary condition",
+                          NULL, stg_bool, &stg_bool, NULL); CHKERRQ(ierr);
   PetscOptionsEnd();
 
+  if (!stg_bool) {
```
jeremylt, line 140: With the name change above, I'd write this as 
```
if (use_stg) {
  // use the STGShur14 problem
} else {
  // use Blasis
}
```

and so on through the file

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,331 @@
...
+  PetscErrorCode ierr;
+  int ndims;
+  char line[char_array_len];
+  char **array;
+
+  PetscFunctionBeginUser;
+  ierr = PetscFOpen(comm, path, "r", fp); CHKERRQ(ierr);
+  ierr = PetscSynchronizedFGets(comm, *fp, char_array_len, line); CHKERRQ(ierr);
+  ierr = PetscStrToArray(line, ' ', &ndims, &array); CHKERRQ(ierr);
+  if(ndims != 2) SETERRQ(comm, -1,
```
jeremylt, line 82: The spacing here  looks funny. (need space for the `if (` and the stuff on the following line doesn't match up) I'd make sure 'make style' is touching these files

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,331 @@
...
+                           "Found %d dimensions instead of 2 on the first line of %s",
+                           ndims, path);
+
+  for (int i=0; i<ndims; i++)  dims[i] = atoi(array[i]);
```
jeremylt, line 86: We tend to use `PetscInt` or `CeedInt`, not `int`

examples/fluids/src/setupstg_shur14.c
```diff
@@ -0,0 +1,331 @@
...
+    rij[0][i]  = (CeedScalar) atof(array[4]);
+    rij[1][i]  = (CeedScalar) atof(array[5]);
+    rij[2][i]  = (CeedScalar) atof(array[6]);
+    rij[3][i]  = (CeedScalar) atof(array[7]);
+    rij[4][i]  = (CeedScalar) atof(array[8]);
+    rij[5][i]  = (CeedScalar) atof(array[9]);
+    lt[i]      = (CeedScalar) atof(array[12]);
+    eps[i]     = (CeedScalar) atof(array[13]);
+
+    if(prof_dw[i] < 0) SETERRQ(comm, -1,
```
jeremylt, line 169: Same thing  here with funky formatting - If you haven't run `make style`, I'd do so. If you have and these files weren't fixed, then I can investigate.

examples/fluids/src/setupstg_shur14.h
```diff
@@ -0,0 +1,14 @@
...
+//
+// This file is part of CEED:  http://github.com/ceed
+
+#include <ceed.h>
+#include <petsc.h>
+#include "../qfunctions/stg_shur14_type.h"
+
+extern PetscErrorCode CreateSTGContext(MPI_Comm comm, DM dm,
+                                       STGShur14Context *pstg_ctx, NewtonianIdealGasContext newt_ctx,
+                                       bool implicit, CeedScalar theta0);
```
jeremylt, line 14: Same thing here, boolean parameter should follow `verb_noun`, so `is_implicit` or `use_implicit`

doc/sphinx/source/references.bib
```diff
@@ -143,7 +143,6 @@ @book{bower2010applied
   publisher={CRC press},
   isbn={978-1-4398-0247-2}
 }
-
```
jeremylt, line 146: line deleted why? There is a newline between the other entries above

jeremylt (commented):
I have a few comments. I'll look over the new documentation next

jrwrigh:
> Is it possible to have a minimal test (comparing with reference output) to confirm that nothing major is broken? 
I can definitely write one in. 

> Also, would you like to squash selectively or squash-merge the whole branch?
I was just thinking to squash-merge the whole branch.

examples/fluids/problems/blasius.c
jrwrigh: Yeah, that makes sense. Should I also change `mean_only` to `use_mean_only` or `use_only_mean`?

Also, maybe change the flag in `blasius.c` to `-stg_use` (so that all the stg-related flags can use the:

```yaml
stg:
  use: true
  stg_inflow_path: './path/STGInflow.dat'
  ...[etc]
```

examples/fluids/src/setupstg_shur14.c
jrwrigh: Yeah, these files aren't being hit by `make style` for some reason. Running `astyle --options=.astylerc` on them explicitly results in changes, but `make style` doesn't.

examples/fluids/problems/blasius.c
```diff
@@ -167,8 +186,21 @@ PetscErrorCode NS_BLASIUS(ProblemData *problem, DM dm, void *ctx) {
...
+    CeedQFunctionContextReferenceCopy(blasius_context,
+                                      &problem->apply_inflow.qfunction_context);
+  } else {
+    CeedQFunctionContextCreate(user->ceed, &stg_shur14_context);
+    CeedQFunctionContextSetData(stg_shur14_context, CEED_MEM_HOST,
+                                CEED_USE_POINTER,
+                                sizeof(*stg_shur14_ctx), stg_shur14_ctx);
+    CeedQFunctionContextSetDataDestroy(stg_shur14_context, CEED_MEM_HOST,
+                                       FreeContextPetsc);
```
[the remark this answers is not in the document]
jrwrigh: Paging @jedbrown.

examples/fluids/src/setupstg_shur14.c
jrwrigh: Only thing I see different about these files is the underscore in the filename? The glob in the `Makefile` looks like it should handle that fine, but I'm not familiar with what special rules `make` may have for globbing.

jeremylt: The recursing deep enough into folders has been flaky, so I suspect that's the culprit

examples/fluids/problems/blasius.c
jeremylt: Yeah, `use_mean_only` would be consistent

for the command line flag I'm fine with the current

examples/fluids/problems/blasius.c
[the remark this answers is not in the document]
jedbrown: Yeah, I'm to blame for that name. The contexts were allocated with PetscMalloc so we have debug/tracing features, so we need to call `PetscFree` and return a libceed error code. Contexts generally need to be plain data (not contain pointers) to be valid on the GPU (without currently-unsupported processing) so this one function is fine for all current and planned contexts. It's used in all the other problems currently.

examples/fluids/problems/blasius.c
jedbrown: Should STG be a sub-type of Blasius or its own problem type (that might call Blasius setup and then do extra stuff)? If it's a subtype, then maybe `-blasius_inflow_type stg`. Or maybe the problem type should become `plate` with `-plate_inflow_type [blasius | stg]`?

examples/fluids/src/setupstg_shur14.c
jedbrown: Should we consider merging in-progress stuff and switching to clang-format here, as we did in Ratel? (Should check in with other libceed stakeholders before taking that step.)

jeremylt: Personally, I like the output better and find clang-format easier to use/configure so :+1: from me

examples/fluids/problems/blasius.c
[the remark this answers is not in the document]
jrwrigh: So is this good as is then?

examples/fluids/src/setupstg_shur14.c
jrwrigh: Should I manually run with astyle real quick, or leave it as is and wait for clang-format to come in? Granted, if the culprit is make globbing, I don't know if switching to clang-format necessarily fixes it.

jeremylt: Swapping to clang-format will take time. I'd go ahead and format this.

Recursing into folders doesn't seem work right with astyle, but it seemed to work fine with clang-format. Not really a make issue.

jrwrigh: Upon further investigation, make is correctly passing the file to astyle (verified via `make style-c -n`), _and_ is successfully running on the file. Astyle just 
A) doesn't care about the missing space after the `if` and
B) enforces shifting the `SETERRQ` arguments out of alignment (ie. making them aligned manually is undone by astyle).

The former can be enforced with `--pad-header`. Not sure about the latter though.

jedbrown:
I see some of these casts that need `const`. I don't understand the failing Julia test; it might not be related.
image

jrwrigh:
Interesting. I don't think those are related, as that expression has been in the code and passed the Julia test fine previously.

To the error itself, should I be specifying `const` in the rvalue of the cast? ie. `const CeedScalar (*prof_ubar)[nprofs] = (const CeedScalar (*)[nprofs]) &stg_ctx->data[stg_ctx->offsets.ubar];`

jedbrown:
Yes

jedbrown (approved):
Tests are passing and I think this can be squash-merged unless you want to consolidate into a few atomic commits.

PR #869, opened 2022-01-13
Update GetMultiplicity function (E^T * E * 1) to work for H(div) space

rezgarshakeri:


jedbrown (approved):
Could use the vector `mult` so you don't need to create another `l_vec`. Also, I think the first apply will overwrite `e_vec` so it doesn't need to be zeroed.

rezgarshakeri:
> 

Right I deleted `l_vec`, but we need to set it to zero (I tried without setting it) before applying restriction otherwise the output values is not correct.

interface/ceed-elemrestriction.c
```diff
@@ -866,15 +866,17 @@ int CeedElemRestrictionGetMultiplicity(CeedElemRestriction rstr,
   int ierr;
   CeedVector e_vec;
 
-  // Create and set e_vec
+  // Create and set l_vec, e_vec
   ierr = CeedElemRestrictionCreateVector(rstr, NULL, &e_vec); CeedChk(ierr);
-  ierr = CeedVectorSetValue(e_vec, 1.0); CeedChk(ierr);
-  ierr = CeedVectorSetValue(mult, 0.0); CeedChk(ierr);
+  ierr = CeedVectorSetValue(mult, 1.0); CeedChk(ierr);
+  ierr = CeedVectorSetValue(e_vec, 0.0); CeedChk(ierr);
```
jedbrown, line 872: Restrictions overwrite, transpose restrictions add. So this one isn't needed.
```suggestion
```

interface/ceed-elemrestriction.c
```diff
@@ -866,15 +866,17 @@ int CeedElemRestrictionGetMultiplicity(CeedElemRestriction rstr,
...
 
   // Apply to get multiplicity
+  ierr = CeedElemRestrictionApply(rstr, CEED_NOTRANSPOSE, mult, e_vec,
+                                  CEED_REQUEST_IMMEDIATE); CeedChk(ierr);
+  ierr = CeedVectorSetValue(mult, 0.0); CeedChk(ierr);
```
jedbrown, line 877: And you're correct that this one is needed.

interface/ceed-elemrestriction.c
```diff
@@ -866,15 +866,17 @@ int CeedElemRestrictionGetMultiplicity(CeedElemRestriction rstr,
   int ierr;
   CeedVector e_vec;
 
-  // Create and set e_vec
+  // Create and set l_vec, e_vec
```
jedbrown, line 869:
```suggestion
  // Create e_vec to hold intermediate computation in E^T (E 1)
```

jedbrown (approved):

interface/ceed-elemrestriction.c
```diff
@@ -866,15 +866,16 @@ int CeedElemRestrictionGetMultiplicity(CeedElemRestriction rstr,
   int ierr;
   CeedVector e_vec;
 
-  // Create and set e_vec
+  // Create e_vec to hold intermediate computation in E^T (E 1)
   ierr = CeedElemRestrictionCreateVector(rstr, NULL, &e_vec); CeedChk(ierr);
-  ierr = CeedVectorSetValue(e_vec, 1.0); CeedChk(ierr);
-  ierr = CeedVectorSetValue(mult, 0.0); CeedChk(ierr);
+  ierr = CeedVectorSetValue(mult, 1.0); CeedChk(ierr);
```
jeremylt, line 871: Minor quibble - the comment above now doesn't really apply to both lines here. You could adjust the comment or just move line 871 down to the next block of code?

jeremylt (approved):

interface/ceed-elemrestriction.c
rezgarshakeri: Thanks, How about now?

interface/ceed-elemrestriction.c
```diff
@@ -868,11 +868,12 @@ int CeedElemRestrictionGetMultiplicity(CeedElemRestriction rstr,
...
   ierr = CeedElemRestrictionCreateVector(rstr, NULL, &e_vec); CeedChk(ierr);
-  ierr = CeedVectorSetValue(mult, 1.0); CeedChk(ierr);
 
-  // Apply to get multiplicity
+  // Compute e_vec = E * 1
+  ierr = CeedVectorSetValue(mult, 1.0); CeedChk(ierr);
   ierr = CeedElemRestrictionApply(rstr, CEED_NOTRANSPOSE, mult, e_vec,
                                   CEED_REQUEST_IMMEDIATE); CeedChk(ierr);
+  // Compute multiplicity, mult = E^T * e_vec = E^T (E 1)
   ierr = CeedVectorSetValue(mult, 0.0); CeedChk(ierr);
```
rezgarshakeri, line 877: @jeremylt

jeremylt: LGTM, thanks!

Issue #695, opened 2021-01-21
Rustdoc math support


jedbrown:
This purports to do the trick.
https://docs.rs/katex-doc/0.1.0/katex_doc/

jedbrown:
Closed via #870

PR #870, opened 2022-01-14
Draft: rust doc: example documenting a method with math

jedbrown:
Cc: issue #695

@jeremylt This renders nicely. I haven't experimented further and I don't know off-hand how to make the dependency only apply to documentation, but not normal builds.
image

### DIFF
--- a/rust/libceed/Cargo.toml
+++ b/rust/libceed/Cargo.toml
@@ -18,7 +18,10 @@ categories = ["science"]
 
 [dependencies]
 libceed-sys = { version = "0.9", path = "../libceed-sys" }
-katexit = "0.1.1"
+katexit = { version = "0.1.1", optional = true }
 
 [dev-dependencies]
 version-sync = "0.9.2"
+
+[package.metadata.docs.rs]
+features = ["katexit"]
--- a/rust/libceed/Cargo.toml
+++ b/rust/libceed/Cargo.toml
@@ -18,6 +18,7 @@ categories = ["science"]
 
 [dependencies]
 libceed-sys = { version = "0.9", path = "../libceed-sys" }
+katexit = "0.1.1"
 
 [dev-dependencies]
 version-sync = "0.9.2"
--- a/rust/libceed/README.md
+++ b/rust/libceed/README.md
@@ -3,14 +3,9 @@
 [![GitHub Actions](https://github.com/CEED/libCEED/actions/workflows/rust-test-with-style.yml/badge.svg)](https://github.com/CEED/libCEED/actions/workflows/rust-test-with-style.yml)
 [![Documentation](https://docs.rs/libceed/badge.svg)](https://docs.rs/libceed)
 
-This crate provides an interface to [libCEED](https://libceed.readthedocs.io),
-which is a performance-portable library for extensible element-based
-discretization for partial differential equations and related computational
-problems. The formulation is algebraic and intended to be lightweight and easy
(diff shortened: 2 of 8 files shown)


jeremylt:
https://docs.rs/rustdoc-katex-demo/0.1.5/rustdoc_katex_demo/
https://docs.rs/katex-doc/latest/katex_doc/

These provide options to inject a small script for rendering into the docs so we don't have to drag around the dependency. 

There is also the possibility of turning katex into an optional feature and turning on that feature on by default for tests (easy) and docs (hard - cargo is touchy about passing arguments sanely to rustdoc for this sort of thing)

jeremylt:
The commit I just pushed shows using katexit as an optional feature, though to run locally you need to do `cargo doc --features=katexit`. Not sure how to enable that feature by default only for `cargo doc`.

jeremylt:
Note: Read the Docs failure unrelated and already fixed on main

jeremylt:
Note, we seem to be stuck with this warning
```
835 | / #[doc = " Handle for the user provided CeedQFunction callback function"]
836 | | #[doc = ""]
837 | | #[doc = "@param[in,out] ctx  User-defined context set using CeedQFunctionSetContext() or NULL"]
838 | | #[doc = "@param[in] Q        Number of quadrature points at which to evaluate"]
...   |
851 | | #[doc = ""]
852 | | #[doc = "@ingroup CeedQFunction"]
    | |_________________________________^
    |
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
    = note: the link appears in this line:
            
            @param[in,out] ctx  User-defined context set using CeedQFunctionSetContext() or NULL
                   ^^^^^^
    = note: no item named `in,out` in scope
    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
```
which is a know issue related to bindgen handling doxygen

jeremylt:
If you like my workaround for having to carry the katexit dependency on all builds, I think we can merge this. Not sure where else you might be thinking of putting fancy equations.

jedbrown:
Thanks. That was just meant as a sample to test that it works. My thought was to visit the key objects and insert the basic math, but mostly, I just wanted the infrastructure there so that it would be low marginal effort to properly document new features.

jeremylt (approved):
Seems good as an initial

jedbrown:
Could you document (maybe in `rust/libceed/README.md`) how to build the docs locally? I just ran this and the math did not render .
```console
$ cargo rustdoc --features katexit --package libceed
 Documenting libceed v0.9.0 (/home/jed/src/libCEED/rust/libceed)
    Finished dev [unoptimized + debuginfo] target(s) in 0.57s
$ firefox target/doc/libceed/struct.Ceed.html
```

jeremylt:
I've only ran with `features=katexit`. I'll test some and see. If skipping the = makes a difference, that might be a cargo bug?

jeremylt:
Ah, it was a matter of me having moved something I shouldn't have. I added a line in the README about building the docs.

jedbrown:
Thanks. Renders correctly now. `cargo doc` does not work on `+nightly` for me (breaks documenting dependencies), but `cargo +stable doc --features katexit` works fine. This also works fine: `cargo +nightly rustdoc --features katexit --package libceed`.

jeremylt:
odd, `cargo +nightly doc`, `cargo +nightly doc --features=katexit`, and `cargo +nightly doc --features katexit` all work ok for me. I wonder if we have different nightly versions?

PR #871, opened 2022-01-18
op - add ctx setting interface for operators

jeremylt:
This PR pushes the context field setting interface up to the operator level so we don't have to retrieve the QFunction and then the QFunctionContext from the Operator to update named fields in the context (like if we are updating `time` in a TS)


jeremylt:
Read the Docs failure seems to be unrelated to this PR.

interface/ceed-operator.c
```diff
@@ -264,6 +264,64 @@ int CeedOperatorGetActiveElemRestriction(CeedOperator op,
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Set QFunctionContext field
```
jedbrown, line 268: Something more specific about how this applies to all sub-operators.
```suggestion
  @brief Set QFunctionContext field for every QFunction containing `field_name` within the provided CeedOperator
```

interface/ceed-operator.c
```diff
@@ -1027,6 +1085,41 @@ int CeedOperatorGetNumQuadraturePoints(CeedOperator op, CeedInt *num_qpts) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Set QFunctionContext field holding a double precision value
```
jedbrown, line 1089: better comments in this file too

jedbrown (approved):

tests/t525-operator.c
```diff
@@ -0,0 +1,103 @@
...
+
+  CeedOperatorCreate(ceed, qf_sub_2, CEED_QFUNCTION_NONE, CEED_QFUNCTION_NONE,
+                     &op_sub_2);
+
+  // Composite operator
+  CeedCompositeOperatorCreate(ceed, &op_composite);
+  CeedCompositeOperatorAddSub(op_composite, op_sub_1);
+  CeedCompositeOperatorAddSub(op_composite, op_sub_2);
+
+  // Check setting field in context of single sub-operator for composite operator
```
jeremylt, line 76: In addition to the documentation, I also expanded this test to be a bit clearer about what happens with composite operators.

jeremylt:
I can't reproduce the Read the Docs error locally, so it seems to be specific to how their env is setup.

jeremylt:
Read the Docs seems to have broken in the jump from Sphinx 3.5 to 4.4

jeremylt:
`sphinxcontrib-bibtex` seems to be the offending party

jeremylt:
Pinning `docutils` seems to help for now. A fix is in the works for `sphinxcontrib-bibtex`:
https://github.com/mcmtroffaes/sphinxcontrib-bibtex/commit/d026c156580c4909a934b4f8daa9985c9fc070f0

PR #872, opened 2022-01-18
Rust - impl cloning Ceed

jeremylt:
This PR lets Rust users clone the Ceed object.

### DIFF
--- a/rust/libceed/src/lib.rs
+++ b/rust/libceed/src/lib.rs
@@ -219,6 +219,27 @@ impl Drop for Ceed {
 }
 
 // -----------------------------------------------------------------------------
+// Cloning
+// -----------------------------------------------------------------------------
+impl Clone for Ceed {
+    /// Perform a shallow clone of a Ceed context
+    ///
+    /// ```
+    /// let ceed = libceed::Ceed::init("/cpu/self/ref/serial");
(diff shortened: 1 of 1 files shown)


jedbrown (approved):
Best to document that this is a shallow clone, similar to cloning a `RefCell`.

jeremylt:
I think it's different than cloning a RefCell - it looks like cloning the RefCell also clones the interior value so the behavior of cloning a RefCell can vary widely depending upon how the interior object impls clone.
<https://doc.rust-lang.org/src/core/cell.rs.html#1147-1165>

rust/libceed/src/lib.rs
```diff
@@ -218,6 +218,27 @@ impl Drop for Ceed {
     }
 }
 
+// -----------------------------------------------------------------------------
+// Cloning
+// -----------------------------------------------------------------------------
+impl Clone for Ceed {
+    /// Performe a shallow clone of a Ceed context
```
jedbrown, line 225:
```suggestion
    /// Perform a shallow clone of a Ceed context
```

jedbrown (approved):
Ah, yes. I should have said `Rc`. https://doc.rust-lang.org/src/alloc/rc.rs.html#1481-1503

PR #873, opened 2022-01-18
actions - bump version of run-on-arch-action

jeremylt:


jeremylt:
Should be good to squash+merge. I was hoping it would be slightly faster, but no. Newer has to be better though, right?

jedbrown (approved):

PR #874, opened 2022-01-18
vec/ctx - make backend restore function optional

jeremylt:
Really, we should just let the backend opt out of specialized restoring logic.


backends/cuda-ref/ceed-cuda-ref-qfunctioncontext.c
```diff
@@ -351,14 +351,6 @@ static int CeedQFunctionContextGetData_Cuda(const CeedQFunctionContext ctx,
   return CEED_ERROR_SUCCESS;
 }
 
-//------------------------------------------------------------------------------
```
jeremylt, line 354: Note - I deleted these for CUDA and HIP but intentionally left them in for `/cpu/self/ref` since that's our explainer backend for new backends

jeremylt:
CI failure is updating LV to PETSc main.

jedbrown (approved):

PR #878, opened 2022-01-21
H(div) basis constructor

rezgarshakeri:
I have created H(div) basis constructor and added unit tests.
You can see the Julia notebook version of this constructor [here](https://github.com/rezgarshakeri/H-div-Tests).


jedbrown:
Are the tests supposed to print all these values (thus fail), versus checking properties and only printing when there's an error?
```console
$ make test search=t330 BACKENDS=/cpu/self/ref/serial
make: 'lib' with optional backends:
1..3
# TESTARGS: {ceed_resource}
ok 1 t330-basis /cpu/self/ref/serial
not ok 2 t330-basis /cpu/self/ref/serial stdout
# + CeedBasis (H(div) discretization): dim=2 P=8 Q=9
[...]
```
Alternatively, you can compare with reference output.

interface/ceed-basis.c
```diff
@@ -391,6 +391,7 @@ int CeedMatrixMultiply(Ceed ceed, const CeedScalar *mat_A,
   @param num_comp    Number of field components (1 for scalar fields)
   @param P_1d        Number of nodes in one dimension
   @param Q_1d        Number of quadrature points in one dimension
+  @param basis_space 1 for H^1 discretization (2 for H(div), 3 for H(curl))
```
jedbrown, line 394:
```suggestion
```
Not actually a parameter. (`make doxygen` warns, though you have to troll through the logs to find it).

interface/ceed-basis.c
```diff
@@ -557,6 +559,7 @@ int CeedBasisCreateTensorH1Lagrange(Ceed ceed, CeedInt dim, CeedInt num_comp,
   @param num_comp    Number of field components (1 for scalar fields)
   @param num_nodes   Total number of nodes
   @param num_qpts    Total number of quadrature points
+  @param basis_space 1 for H^1 discretization (2 for H(div), 3 for H(curl))
```
jedbrown, line 562:
```suggestion
```

interface/ceed-basis.c
```diff
@@ -621,6 +625,75 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
...
 }
 
+/**
+  @brief Create a non tensor-product basis for H(div) discretizations
+  @param ceed        A Ceed object where the CeedBasis will be created
+  @param topo        Topology of element, e.g. hypercube, simplex, ect
+  @param num_comp    Number of componenet, we have 1 vector componenet in H(div)
+  @param num_nodes   Total number of nodes (dofs per element)
+  @param num_qpts    Total number of quadrature points
+  @param basis_space 2 for H(div) discretization (1 for H^1, 3 for H(curl))
```
jedbrown, line 635:
```suggestion
```

rezgarshakeri:
> 

`t330-basis.c` is just displays the basis. Similar to t300-basis.c. 
By the way I checked the values of the libCEED output in julia code that I shared above (t330-basis.ipynb)

interface/ceed-basis.c
```diff
@@ -621,6 +625,75 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Create a non tensor-product basis for H(div) discretizations
```
jedbrown, line 629:
```suggestion
  @brief Create a non tensor-product basis for H(div) discretizations

```

interface/ceed-basis.c
```diff
@@ -621,6 +625,75 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
...
+  @param num_comp    Number of componenet, we have 1 vector componenet in H(div)
+  @param num_nodes   Total number of nodes (dofs per element)
+  @param num_qpts    Total number of quadrature points
+  @param basis_space 2 for H(div) discretization (1 for H^1, 3 for H(curl))
+  @param interp      Row-major (dim*num_qpts * num_nodes) matrix expressing the values of
+                       nodal basis functions at quadrature points
+  @param div        Row-major (num_qpts * num_nodes) matrix expressing
+                       divergence of nodal basis functions at quadrature points
+  @param q_ref       Array of length num_qpts holding the locations of quadrature
+                       points on the reference element [-1, 1]
```
jedbrown, line 641: The reference element will generally be something else.

interface/ceed-basis.c
```diff
@@ -621,6 +625,75 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Create a non tensor-product basis for H(div) discretizations
+  @param ceed        A Ceed object where the CeedBasis will be created
+  @param topo        Topology of element, e.g. hypercube, simplex, ect
```
jedbrown, line 631:
```suggestion
  @param topo        Topology of element (`CEED_QUAD`, `CEED_PRISM`, etc.), dimension of which is used in some array sizes below
```

interface/ceed-basis.c
```diff
@@ -621,6 +625,75 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
...
+  @param num_comp    Number of componenet, we have 1 vector componenet in H(div)
```
jedbrown, line 632:
```suggestion
  @param num_comp    Number of components (usually 1 for vectors in H(div) bases)
```
VSCode has a spell checker for code that helps to find typos

interface/ceed-basis.c
rezgarshakeri: Yeah, I just copied from H^1 discretizations (line 568), perhaps we need to change that.

jedbrown:
> 
> `t330-basis.c` is just for displaying the basis, similar to t300-basis.c. By the way I checked the values of the libCEED output in julia code that I shared above (t330-basis.ipynb)

Perhaps you created the file `tests/output/t330-basis.out` but forgot to `git add` it? That's necessary if you want to compare text output with a reference.

interface/ceed-basis.c
jedbrown: Yeah, it was probably copied from the tensor constructor.

rezgarshakeri:
> > `t330-basis.c` is just for displaying the basis, similar to t300-basis.c. By the way I checked the values of the libCEED output in julia code that I shared above (t330-basis.ipynb)
> 
> Perhaps you created the file `tests/output/t330-basis.out` but forgot to `git add` it? That's necessary if you want to compare text output with a reference.

Oh no, I copied libCEED output to julia and compare it inside julia. I have nothing to add to this test.

interface/ceed-basis.c
rezgarshakeri: I deleted [-1,1] on non-tensor basis constructor.

jedbrown:
So run the example to generate that file and commit the result so that `make test search=t330` can succeed. This is saying "I've certified this output by some external procedure" and tests need to match it to succeed.

rezgarshakeri:
> So run the example to generate that file and commit the result so that `make test search=t330` can succeed. This is saying "I've certified this output by some external procedure" and tests need to match it to succeed.
Ok, I'll create a file from julia results to libCEED.

jeremylt:
I don't think you'll be able to make the correct file in Julia, you have to make it with libCEED. The test suite checks that the output for this test exactly matches the file you provide, so you need to provide the exact output libCEED will produce.

rezgarshakeri:
> 

Thanks, I didn't know that.

jedbrown:
Yeah, literally just run the example (manually) and redirect the output to `tests/output/t330-basis.out`, then commit that file. The corresponding `make test` should pass.

interface/ceed-basis.c
```diff
@@ -657,30 +728,46 @@ int CeedBasisReferenceCopy(CeedBasis basis, CeedBasis *basis_copy) {
...
-                          basis->q_weight_1d, stream); CeedChk(ierr);
-    ierr = CeedScalarView("interp1d", "\t% 12.8f", basis->Q_1d, basis->P_1d,
-                          basis->interp_1d, stream); CeedChk(ierr);
-    ierr = CeedScalarView("grad1d", "\t% 12.8f", basis->Q_1d, basis->P_1d,
-                          basis->grad_1d, stream); CeedChk(ierr);
-  } else {
-    fprintf(stream, "CeedBasis: dim=%d P=%d Q=%d\n", basis->dim, basis->P,
-            basis->Q);
+  switch (basis->basis_space) {
+  case 1: // H^1 discretization
```
jeremylt, line 732: Instead of magic numbers, this should be an enum

interface/ceed-basis.c
```diff
@@ -621,6 +623,75 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
...
+  CeedInt Q = num_qpts, P = num_nodes, dim = 0;
+  ierr = CeedBasisGetTopologyDimension(topo, &dim); CeedChk(ierr);
+  if (!ceed->BasisCreateHdiv) {
+    Ceed delegate;
+    ierr = CeedGetObjectDelegate(ceed, &delegate, "Basis"); CeedChk(ierr);
+
+    if (!delegate)
+      // LCOV_EXCL_START
+      return CeedError(ceed, CEED_ERROR_UNSUPPORTED,
+                       "Backend does not support BasisCreateHdiv");
```
jedbrown, line 661:
```suggestion
                       "Backend does not implement BasisCreateHdiv");
```
This will get GPU tests passing (because the string "Backend does not implement" is used to identify SKIP tests).

interface/ceed-basis.c
```diff
@@ -1059,6 +1146,18 @@ int CeedBasisGetGrad1D(CeedBasis basis, const CeedScalar **grad_1d) {
...
 
+/**
+  @brief Get divergence matrix of a CeedBasisHdiv
+  @param basis        CeedBasisHdiv
+  @param[out] div  Variable to store divergence matrix
+  @return An error code: 0 - success, otherwise - failure
+  @ref Advanced
+**/
+int CeedBasisGetDiv(CeedBasis basis, const CeedScalar **div) {
+  *div = basis->div;
```
jeremylt, line 1157: Should there be a check for when this is unsupported?

tests/t330-basis.h
```diff
@@ -0,0 +1,86 @@
...
+//  b7|         |b3
+//    |         |
+//  b6|         |b2
+//    0---------1
+//     b0     b1
+// Bx[0-->7] = b0_x-->b7_x, By[0-->7] = b0_y-->b7_y
+// To see how the nodal basis is constructed visit:
+// https://github.com/rezgarshakeri/H-div-Tests
+int NodalHdivBasisQuad(CeedScalar *X, CeedScalar *Bx, CeedScalar *By) {
+  CeedScalar xhat = X[0];
```
jeremylt, line 30: we use `x_hat` over `xhat` typically

jedbrown: Yeah. And since H^1 is contained in H(div), should we automatically construct it from the gradient if requested (and number of components = dimension)? You tend to use H^1 and H(div) elements differently (Piola map) so this might cause more harm than good (say, in evaluating the off-diagonal block in mixed velocity-pressure).

rezgarshakeri: > we use `x_hat` over `xhat` typically

Thanks.

interface/ceed-basis.c
rezgarshakeri: Yeah, agreed. Can we add enum to the basis constructor? This is something like `tensor_basis` we have it already but it is not 0,1 only. We have L^2, and H(curl) as well. I set 0,1,2,3 for L^2, H^1, H(div), H(curl) (last one not implemented)

tests/t330-basis.h
rezgarshakeri: > 

So you mean add `div` to H^1 constructor for the case when we need the divergence in H^1?! Isn't the sum of the diagonal of gradient?!

interface/ceed-basis.c
rezgarshakeri: We can add that switch case `basis_space` (with those magic numbers or whatever we will do) to work only for H(div).

interface/ceed-basis.c
jeremylt: I was thinking something like this 
https://github.com/CEED/libCEED/blob/7ed177db354372d7a27bd4ae30dbb2e96a56ab87/include/ceed/ceed.h#L490

rezgarshakeri: Ah, nice. So with that I think we need to add a new parameter to the basis constructor, right? Parameter like `CeedFESpace`.

tests/t330-basis.h
jedbrown: I think I'm making the case that even though divergence appears in weak forms when using H^1 elements (as in `q * div(u) + p * div(v)` for Stokes or mixed elasticity), we should not create it by default because it can't be used on stretched elements (reference element divergence doesn't tell you physical element divergence for these elements). So let's refuse to emulate in this way that would likely only cause confusing bugs.

interface/ceed-basis.c
jeremylt: Which basis constructor? I thought our current basis constructors all specify the space in their names?

rezgarshakeri: Yeah, the space are specified in their name. In the link that you shared `CeedElemTopology` is an input for basis constructor. If we need something like that we should add it to the input of the constructor, right?or we don't need to.!?

rezgarshakeri: I think we don't need to added as input.Let me check.

interface/ceed-basis.c
```diff
@@ -729,12 +729,22 @@ int CeedBasisReferenceCopy(CeedBasis basis, CeedBasis *basis_copy) {
...
+  if (basis->tensor_basis) {
+    fprintf(stream, "CeedBasis (%s on a %s element): dim=%d P=%d Q=%d\n",
+            CeedFESpaces[FE_space], CeedElemTopologies[topo],
+            basis->dim, basis->P_1d, basis->Q_1d);
+  } else {
+    fprintf(stream, "CeedBasis (%s on a %s element): dim=%d P=%d Q=%d\n",
+            CeedFESpaces[FE_space], CeedElemTopologies[topo],
+            basis->dim, basis->P, basis->Q);
+  }
+  // Print quadrature data, interpolation/gradient/divergene/curl of the basis
```
rezgarshakeri, line 743: @jeremylt how about that?

jeremylt: You'll need to change some .out files, but it looks good to me

rezgarshakeri: I'll do. Thanks

include/ceed/ceed.h
```diff
@@ -494,22 +494,36 @@ CEED_EXTERN const char *const CeedQuadModes[];
...
   /// Pyramid - 3D shape
-  CEED_PYRAMID = 3 << 16 | 4,
+  CEED_FE_TOPO_PYRAMID = 3 << 16 | 4,
   /// Prism - 3D shape
-  CEED_PRISM = 3 << 16 | 5,
+  CEED_FE_TOPO_PRISM = 3 << 16 | 5,
   /// Hexehedron - 3D shape
-  CEED_HEX = 3 << 16 | 6,
+  CEED_FE_TOPO_HEX = 3 << 16 | 6,
 } CeedElemTopology;
```
jeremylt, line 510: Small detail, we probably should call this enum CeedFETopology

jedbrown, line 510: I'd rather not have `FE` here. libCEED is for *element-based* discretization, but not strictly *finite element* discretization. We don't have examples, but I think we'd like to support methods like spectral volume and spectral difference. We can say `CEED_ELEM_HEX` or `CEED_TOPOLOGY_HEX` or `CEED_ELEM_TOPOLOGY_HEX`.

include/ceed/ceed.h
```diff
@@ -494,22 +494,36 @@ CEED_EXTERN const char *const CeedQuadModes[];
...
+  /// L2 FE space
+  CEED_FE_SPACE_L2 = 0,
+  /// H1 FE space
+  CEED_FE_SPACE_H1 = 1,
+  /// H(div) FE space
+  CEED_FE_SPACE_HDIV = 2,
+  /// H(curl) FE space
+  CEED_FE_SPACE_HCURL = 3,
+} CeedFESpace;
+CEED_EXTERN const char *const CeedFESpaces[];
```
jedbrown, line 525: I'm also a little concerned about this amount of finite element terminology. Do we need to talk about these spaces or could we just say which subset of {values, gradients, divergence, curl} is supplied (all on a reference element)?

include/ceed/ceed.h
jeremylt: Fair point here. I think CEED_TOPOLOGY_HEX makes more sense and is about the same length as the current

include/ceed/ceed.h
jeremylt: Question - can the Hdiv interp and grad be written to run from the same code as the H1 interp and grad? If so, then we can simplify this PR a fair amount. 

In terms of the printing, we could change this just to print whichever of the interp, grad, and div matrices are avaliable

backends/ref/ceed-ref-basis.c
```diff
@@ -306,7 +306,112 @@ static int CeedBasisApply_Ref(CeedBasis basis, CeedInt num_elem,
...
+    switch (eval_mode) {
+    // Interpolate to/from quadrature points
+    case CEED_EVAL_INTERP: {
+      CeedInt P = num_nodes, Q = dim*num_qpts;
+      const CeedScalar *interp;
+      ierr = CeedBasisGetInterp(basis, &interp); CeedChkBackend(ierr);
+      if (t_mode == CEED_TRANSPOSE) {
+        P = dim*num_qpts; Q = num_nodes;
+      }
+      ierr = CeedTensorContractApply(contract, num_comp, P, num_elem, Q,
```
jeremylt, line 361: Does this support multiple components? Should it?

include/ceed/ceed.h
rezgarshakeri: Right. I'll update that.

include/ceed/ceed.h
rezgarshakeri: In my mixed PR, I used P0, L^2 basis which has only interpolation matrix of size (QxP=1) and it is constant at all quadrature points. I tried to create a constant interpolation using H^1, but got error (P must be >1). 
I am not sure that we can use H^1 interpolation for H(div) case. In H(div), interpolation is a matrix of size `(dim*QxP)`. The interpolation for x-component is different from y-component (or z-comp). But in H^1 we use the same interpolation matrix of size `(QxP)` for all other components.

backends/ref/ceed-ref-basis.c
rezgarshakeri: Because of the way we created the restriction from plex, we set num_comp=1. It is like we have 1 vector dof, and the components of this dofs goes to interpolation matrix (dim*QxP, as I mentioned above)

include/ceed/ceed.h
jedbrown: Hmm, they're both (as non-tensor) just matrix multiplies. The shape calculation is different as you say. Though I think components (living in E-vector) can be handled the same way as we do for H^1 (Kronecker with identity).

@rezgarshakeri I think you can call `CeedBasisCreateH1` with `P`=1. You can't call `CeedBasisCreateTensorH1Lagrange` because we're hard-coded to putting nodes at Lobatto points and you can't have just one Lobatto point. We could make a variant that puts dofs at Lobatto points, but then you can't make globally H^1 conforming spaces (without significant pain).

backends/ref/ceed-ref-basis.c
jeremylt: But is it possible mathematically to have a multiple components on a single Hdiv element? That feels like a natural possibility

rezgarshakeri: They really have multiple components, dof is a vector. At the end we use the H(div) basis to interpolate a vector variable (like velocity). But can we handle multiple components using tensor product as we do for H^1?! I don't know how.! 
For this element I implemented, we have P=8 values for x and P=8 values for y at quadrature points.

jeremylt: I understand that at the quadrature points there are dim components. That's not what I mean. I mean that it feels natural that we should be able to represent multiple Hdiv fields of the same order with the same basis. Doing one interpolation with three fields seems it would be more effecient than three separate interpolations, for instance.

include/ceed/ceed.h
rezgarshakeri: Right, I can give a constant interpolation array to the `CeedBasisCreateH1`.
Right now I had to create my `dm` using 1 component and 2 dof on each face to give me exactly linear BDM element. So the dmplex, is another part that we need to handle it.

backends/ref/ceed-ref-basis.c
rezgarshakeri: Hmm, I see what you mean. So like what we have in H^1, we have one interpolation but use it for all other components (to handle vector values), right?

jeremylt: Exactly.

 Right now the ref code uses ncamp in some places but not others, which is why I noticed it and asked.

include/ceed/ceed.h
jeremylt: Hmm, the shape being different is an issue. Would it help to let the basis set ncomp_dofs and ncomp_qpts separately? That could either make our  ref backend code smaller and more flexible or just more confusing.

backends/ref/ceed-ref-basis.c
jedbrown: Yeah, there are methods for elasticity in which a tensor (stress or strain, without symmetry) lives in an H(div) space. From https://doi.org/10.1007/s10915-010-9373-2
image

rezgarshakeri: To my knowledge at least for this element (BDM element) I implemented, I don't know how to do that.
Perhaps we need to check `FEnics` to see how they handle it. https://launchpadlibrarian.net/83776282/fenics-book-2011-10-27-final.pdf.

jedbrown: To make an H(div) space for matrices by treating each row as a standard H(div) space? I think your code would almost work, with `ncomp=dim`.

include/ceed/ceed.h
rezgarshakeri: What do you mean by basis set ncomp_dofs and ncomp_qpts separately?

jeremylt: I'm imagining for a 3D two component H1 element ncomp_nodes = ncomp_qpts = 3 but for a 3D two component Hdiv element ncomp_nodes = 2 and ncomp_qpts = 6

backends/ref/ceed-ref-basis.c
jedbrown: There are layers of abstraction here. LibCEED is only responsible for the algebraic operation. As far as I'm concerned, it doesn't need to know anything about H(div) or H^1, it just has a matrix of shape `(dim*Q, P)` that may be Kroneckered with an identity to transform a vector of length `c*P` to a vector of length `c*dim*Q`. We want the basis to *also* be able to transform the vector of length `c*P` to `c*Q`. These operations have names, but perhaps there's nothing special about the operation (except internal structure most relevant to tensor elements).

Let k be the number of Q-vec components in each E-vec dof, and c be the number of E-vec components.

| operation | basis data shape | Q-vec shape | E-vec shape | notes |
|---|---|---|---|---|
| interp | (k, Q, P) | (c, k, Q) | (c, P) |
| grad | (dim, k, Q, P) | (dim, c, k, Q) | (c, P) |
| div | (Q, P) | (c, Q) | (c, P) | requires k=dim |
| curl(dim=2) | (1, Q, P) | (1, c, Q) | (c, P) | requires k=dim |
| curl(dim=3) | (dim, Q, P) | (dim, c, Q) | (c, P) | requires k=dim |

We have Poisson (k=1, c=1), standard elasticity (k=1, c=3), mixed/Darcy (k=dim, c=1), mixed elasticity/weak symmetry (k=dim, c=3), Nedelec for Maxwell (k=3, c=1). The "requires" notes are for that operation to be compatible with the corresponding interpolation. But perhaps someone will conceive of another operation so maybe we can make it possible to define new operations via their shape semantics. (Not in this PR.)

jeremylt: I could be wrong, but it seems like from this chart we could have the basis separately track c and k to reduce repetition in the ref apply code and provide a simpler path forward to extending this to the GPU backends.

rezgarshakeri: > 

Yeah, that looks great. So we need to update `CeedTensorContractApply` function to take `k` as an input as well, right?

jeremylt: The tensor contract shouldn't care. I think it would be a property of the basis during BasisApply.

rezgarshakeri: Ah, right. So we add this to the basis and for example set Q = k*num_qpts, in BasisApply. Like what we have here (dim*num_qpts).

jeremylt: Right, that's roughly what I'm thinking. Though I would keep Q still as the number of qpts and just scale pre/post as needed by k.

tests/t330-basis.h
rezgarshakeri: Should we leave `CEED_EVAL_DIV` for H^1 in another PR? perhaps when we do incompressible elasticity?!

include/ceed/ceed.h
rezgarshakeri: I have created/used `CeedBasisCreateL2` in my mixed branch (not include it in this PR to make review easier). Do you want me delete that and use `CeedBasisCreateH1`?

tests/t330-basis.h
jedbrown: I think `CEED_EVAL_DIV` should probably error for H^1, at leas for now. So yes, not this MR.

jeremylt: Though, it seems reasonable to me to collapse down to a single backend BasisApply by introducing this c/k division below. I think that would give us a smaller about of code, but a wider range of cases we could handle. In the H1 case we can error because no div matrix is set without knowing specifically that we're doing a H1 Apply

backends/ref/ceed-ref-basis.c
rezgarshakeri: How about here, I think it is easy to add `k` to the basis constructor and set it to 1 for H^1 and `dim` for H(div), so I don't need to create `CeedBasisApplyHdiv_Ref`.

jeremylt: Agreed. I think we had the same idea/comment at the same time in two different threads

jeremylt: What would this k be called? `CeedBasisGetNumQuadratureComponents`?

tests/t330-basis.h
rezgarshakeri: Yeah, right now we have `CEED_EVAL_DIV not supported` for H^1. I can add switch case that div works for H(div) only.

backends/ref/ceed-ref-basis.c
rezgarshakeri: Isn't `Q-vec` component?

rezgarshakeri: Perhaps add `CeedBasisGetQVectorComponents`?

jeremylt: QVector is just short for QuadraturePointVector

rezgarshakeri: Thanks, so what we should call it? `NumQuadratureComponenets` is a little confusing to me.

jeremylt: Names are hard. I don't have a good  idea here. My thought of 

`CeedBasisGetNumQuadratureComponents` was from the discussion above
>Let k be the number of Q-vec components in each E-vec dof, and c be the number of E-vec components.

But I'm open to other ideas

rezgarshakeri: I added `Q_comp` and `CeedBasisGetNumQuadratureComponents` so we don't need `BasisApplyHdiv_Ref`.

interface/ceed-basis.c
```diff
@@ -1059,6 +1174,32 @@ int CeedBasisGetGrad1D(CeedBasis basis, const CeedScalar **grad_1d) {
...
+  @brief Get divergence matrix of a CeedBasisHdiv
+  @param basis        CeedBasisHdiv
+  @param[out] div  Variable to store divergence matrix
+  @return An error code: 0 - success, otherwise - failure
+  @ref Advanced
+**/
+int CeedBasisGetDiv(CeedBasis basis, const CeedScalar **div) {
+
+  CeedFESpace FE_space = basis->basis_space;
+  switch (FE_space) {
```
jeremylt, line 1187: I think we can simplify/future proof this function a bit:
```suggestion
int CeedBasisGetDiv(CeedBasis basis, const CeedScalar **div) {
  if (!basis->div)
     return CeedError(basis->ceed, CEED_ERROR_MINOR,
                     "CeedBasis does not have divergence matrix.");
 
   *div = basis->div;
   return CEED_ERROR_SUCCESS;
 }
```

interface/ceed-basis.c
```diff
@@ -1059,6 +1174,32 @@ int CeedBasisGetGrad1D(CeedBasis basis, const CeedScalar **grad_1d) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Get divergence matrix of a CeedBasisHdiv
+  @param basis        CeedBasisHdiv
+  @param[out] div  Variable to store divergence matrix
+  @return An error code: 0 - success, otherwise - failure
+  @ref Advanced
+**/
```
jeremylt, line 1183:
```suggestion
  @brief Get divergence matrix of a CeedBasisHdiv
  
  @param basis        CeedBasisHdiv
  @param[out] div  Variable to store divergence matrix
  
  @return An error code: 0 - success, otherwise - failure
  
  @ref Advanced
**/
```

interface/ceed-basis.c
```diff
@@ -657,30 +731,56 @@ int CeedBasisReferenceCopy(CeedBasis basis, CeedBasis *basis_copy) {
...
-            basis->Q);
+    fprintf(stream, "CeedBasis (%s on a %s element): dim=%d P=%d Q=%d\n",
+            CeedFESpaces[FE_space], CeedElemTopologies[topo],
+            basis->dim, basis->P, basis->Q);
+  }
+  // Print quadrature data, interpolation/gradient/divergene/curl of the basis
+  switch (FE_space) {
+  case CEED_FE_SPACE_L2: // L^2 discretization
+  case CEED_FE_SPACE_H1: // H^1 discretization
+    if (basis->tensor_basis) { // H^1 tensor basis
```
jeremylt, line 750: Similarly, we can simplify here, so that the difference is tensor vs non-tensor and then div is only displayed if available. We can use the new `Q_dim` to handle the interp shape difference.

interface/ceed-basis.c
rezgarshakeri: maybe just `CeedBasis` not `CeedBasisHdiv` ?!

jeremylt: good call

interface/ceed-basis.c
rezgarshakeri: Right, I forgot to update here, we can use `Q_comp`.

include/ceed-impl.h
```diff
@@ -197,18 +202,23 @@ struct CeedBasis_private {
   *q_weight_1d; /* array of length Q1d holding the quadrature weights on
                                the reference element */
   CeedScalar
-  *interp;    /* row-major matrix of shape [Q, P] expressing the values of
-                   nodal basis functions at quadrature points */
+  *interp;    /* row-major matrix of shape [Q_com*Q, P] expressing the values of
+                   nodal basis functions at quadrature points.*/
```
jeremylt, line 206:
```suggestion
  *interp;    /* row-major matrix of shape [Q_comp*Q, P] expressing the values of
                   nodal basis functions at quadrature points */
```

interface/ceed-basis.c
```diff
@@ -621,6 +625,76 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Create a non tensor-product basis for H(div) discretizations
+
```
jeremylt, line 630: same suggestion here wrt spacing inside of the comment block

interface/ceed-basis.c
```diff
@@ -1187,7 +1187,16 @@ int CeedBasisGetGrad1D(CeedBasis basis, const CeedScalar **grad_1d) {
 int CeedBasisGetDiv(CeedBasis basis, const CeedScalar **div) {
 
   CeedFESpace FE_space = basis->basis_space;
-  switch (FE_space) {
+int CeedBasisGetDiv(CeedBasis basis, const CeedScalar **div) {
```
jeremylt, line 1190: woops, the suggestion went in wrong so this commit will need tidying

YohannDudouit:
Is this interface only meant for non-tensor elements?

jeremylt:
I assume we'll eventually add tensor elements but they aren't in this PR, no

include/ceed/ceed.h
YohannDudouit: > Question - can the Hdiv interp and grad be written to run from the same code as the H1 interp and grad? If so, then we can simplify this PR a fair amount.
> 
> In terms of the printing, we could change this just to print whichever of the interp, grad, and div matrices are avaliable

For the GPU backends this will require some refactoring and redesign to take into account that the number of dofs per dimension is different for each component on tensor elements.

jeremylt: Right, I was thinking the GPU work would come in a follow up PR and right now we'd merge with the standard "not implemented" message.

YohannDudouit: Regarding the initialization of tensor basis for H(div) and H(curl) (which require two different "H1 1D basis" -- GL and GLL), it may be interesting to spend a little extra time on the design to make it generic enough to allow spectral finite element methods too.

jedbrown: Do you have a favorite reference on tensor decomposition for H(div) and H(curl) spaces? We have an immediate need for the non-tensor case, but I agree that libCEED will want high order spaces at some point. I assumed it would be a specialized constructor for tensor H(div) bases.

interface/ceed-basis.c
```diff
@@ -623,6 +627,79 @@ int CeedBasisCreateH1(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
...
+  (*basis)->Q_comp = dim;
+  (*basis)->basis_space = 2; // 2 for H(div) space
+  ierr = CeedMalloc(Q*dim,&(*basis)->q_ref_1d); CeedChk(ierr);
+  ierr = CeedMalloc(Q,&(*basis)->q_weight_1d); CeedChk(ierr);
+  if (q_ref) memcpy((*basis)->q_ref_1d, q_ref, Q*dim*sizeof(q_ref[0]));
+  if(q_weight) memcpy((*basis)->q_weight_1d, q_weight, Q*sizeof(q_weight[0]));
+  ierr = CeedMalloc(dim*Q*P, &(*basis)->interp); CeedChk(ierr);
+  ierr = CeedMalloc(Q*P, &(*basis)->div); CeedChk(ierr);
+  if(interp) memcpy((*basis)->interp, interp, dim*Q*P*sizeof(interp[0]));
+  if(div) memcpy((*basis)->div, div, Q*P*sizeof(div[0]));
```
jeremylt, line 697: minor style to reflect main
```suggestion
  ierr = CeedCalloc(1, basis); CeedChk(ierr);

  (*basis)->ceed = ceed;
  ierr = CeedReference(ceed); CeedChk(ierr);
  (*basis)->ref_count = 1;
  (*basis)->tensor_basis = 0;
  (*basis)->dim = dim;
  (*basis)->topo = topo;
  (*basis)->num_comp = num_comp;
  (*basis)->P = P;
  (*basis)->Q = Q;
  (*basis)->Q_comp = dim;
  (*basis)->basis_space = 2; // 2 for H(div) space
  ierr = CeedMalloc(Q*dim, &(*basis)->q_ref_1d); CeedChk(ierr);
  ierr = CeedMalloc(Q, &(*basis)->q_weight_1d); CeedChk(ierr);
  if (q_ref) memcpy((*basis)->q_ref_1d, q_ref, Q*dim*sizeof(q_ref[0]));
  if (q_weight) memcpy((*basis)->q_weight_1d, q_weight, Q*sizeof(q_weight[0]));
  ierr = CeedMalloc(dim*Q*P, &(*basis)->interp); CeedChk(ierr);
  ierr = CeedMalloc(Q*P, &(*basis)->div); CeedChk(ierr);
  if (interp) memcpy((*basis)->interp, interp, dim*Q*P*sizeof(interp[0]));
  if (div) memcpy((*basis)->div, div, Q*P*sizeof(div[0]));
```

jeremylt:
@rezgarshakeri you're good to squash+merge once CI passes. Thanks for all the work in this branch!

rezgarshakeri:
> @rezgarshakeri you're good to squash+merge once CI passes. Thanks for all the work in this branch!

Thanks @jeremylt , @jedbrown and @YohannDudouit for all your helps and comments.

include/ceed/ceed.h
YohannDudouit: [This book](https://link.springer.com/book/10.1007/978-94-017-7761-2) describes the tensor decomposition p.86

rezgarshakeri: Thanks for sharing.

rezgarshakeri: I need to digest that case first and extend it to BDM element case we have here.

jeremylt (approved):
Sorry about that, I thought I had already approved. Good for squash+merge.

rezgarshakeri:
> Sorry about that, I thought I had already approved. Good for squash+merge.

No worries, Thanks.

include/ceed/ceed.h
```diff
@@ -494,22 +494,32 @@ CEED_EXTERN const char *const CeedQuadModes[];
...
 
+/// Type of FE space;
+/// @ingroup CeedBasis
+typedef enum {
+  /// H1 FE space
+  CEED_FE_SPACE_H1 = 1,
+  /// H(div) FE space
+  CEED_FE_SPACE_HDIV = 2,
+} CeedFESpace;
+CEED_EXTERN const char *const CeedFESpaces[];
```
jedbrown, line 521: I just noticed that this isn't used in the public interface (now, and I'm not sure whether it should ever be).

jeremylt: I think internally its only used in `CeedBasisView` now.

rezgarshakeri: Yeah, it is used only in `CeedBasisView` so far.

PR #879, opened 2022-01-22
Doxygen cleanup

jedbrown:
Speeds things up a bit and reduces noisy output so it's easier to notice Doxygen warnings.


jeremylt (approved):
Much nicer locally

PR #881, opened 2022-01-25
Fluids - update mesh creating and remove problem-specific BCs

LeilaGhaffari:
All the code related to creating meshes and sizing the domain is removed. The problem-specific BCs are also dropped. This will enable the user to specify wall, slip, inflow, and/or outflow BCs according to their own  problem setup.

TODOs:

- [x] Rebase onto main
- [x] Update the doc


examples/fluids/src/setuplibceed.c
```diff
@@ -96,25 +95,21 @@ PetscErrorCode CreateOperatorForDomain(Ceed ceed, DM dm, SimpleBC bc,
...
-      CeedElemRestriction elem_restr_x_sur, elem_restr_q_sur,
-                          elem_restr_qd_i_sur;
+      CeedOperator        op_setup_sur, op_apply_inflow;
+      CeedElemRestriction elem_restr_x_sur, elem_restr_q_sur, elem_restr_qd_i_sur;
 
       // ---- CEED Restriction
-      ierr = GetRestrictionForDomain(ceed, dm, height, domain_label, i+1,
-                                     Q_sur, q_data_size_sur, &elem_restr_q_sur,
-                                     &elem_restr_x_sur, &elem_restr_qd_i_sur);
+      ierr = GetRestrictionForDomain(ceed, dm, height, domain_label, bc->inflows[i],
```
jeremylt, line 110: is this change giving the same value?

LeilaGhaffari: Yes! Just checked it again and it is giving me the correct value for Traveling Vortex.

examples/fluids/qfunctions/advection2d.h
```diff
@@ -448,6 +447,57 @@ CEED_QFUNCTION(Advection2d_Sur)(void *ctx, CeedInt Q,
...
+                                    };
+    // Normal velocity
+    const CeedScalar u_normal = norm[0]*u[0] + norm[1]*u[1];
+
+    // No Change in density or momentum
+    for (CeedInt j=0; j<4; j++)
+      v[j][i] = 0;
+
+    // Implementing inflow BC
+    if (u_normal < 0)
```
jeremylt, line 479: is this `if` needed here?

LeilaGhaffari: Good point. For advection and advection2d, I prefer to leave the if-statement since it's safer to define inflow and outflow by the u_normal. In advection2d, it is easy to figure out the inflow and outflow face IDs by looking at the velocity components but it becomes tedious/tricky in advection.

jeremylt: Is it ok to possibly end up in a state where v[0..3] are changed but not v[4]?

LeilaGhaffari: I had the same thought and tried to zero v[4] before the if-statement but it didn't affect the results. I can zero it again to be safe.

jeremylt: I don't understand what splitting this QFunction and repeating most of it gives us?

LeilaGhaffari: This setup works for advection and euler but when we have viscosity (Navier-Stokes) `u_normal` is not a proper criterion in defining the in/outflow faces anymore. In the presence of viscosity, the user should give us the inflow and outflow face IDs. By splitting this QFunction, I was mostly envisioning a generic design for all our fluid problems, especially the one that @jrwrigh is working on. For turbulence simulation, we would need to treat the inflow and outflow boundaries differently.

jeremylt: That makes sense. My concern here though is we have taken 1 QFunction with 2 possible outcomes and converted it into 2 QFunctions with 3 (or 4?) possible outcomes. Was this the intention?

jedbrown: I think it'd be best to either leave it in one function or to error in case an inflow isn't an inflow. It's likely that a user will want to make an inflow condition with different (perhaps time-dependent) profile, but probably always the same outflow condition.

jeremylt: @LeilaGhaffari, you can return a non-zero value from a QFunction and on the CPU side this will get `CeedChk`ed and stop the code.

LeilaGhaffari: Ok, I just rebased this branch afresh and combined the inflow and outflow BCs for Advection(2d). We can change this later. I am seeing the original error that I reported on Zulip. Advection and Advection2d are failing for similar errors. Other tests are passing.
Screenshot from 2022-01-29 16-57-09

LeilaGhaffari: > @LeilaGhaffari, you can return a non-zero value from a QFunction and on the CPU side this will get `CeedChk`ed and stop the code.

@jeremylt , sorry, I am not following. What is this for?

jedbrown: Does the above run correctly locally?

LeilaGhaffari: No, same issue on my machine.

jedbrown: Returning nonzero is the way to signal an error, as when a boundary designated as inflow receives an outflow velocity.

jedbrown: I pushed a fix for the missing `DMPlexLabelComplete`, which happens automatically when there is any essential boundary condition defined over `"Face Sets"`. But I don't think it should be required here. Specifically, `DMPlexGetLocalOffsets` calls
```c
  if (domain_label) {
    ierr = DMGetFirstLabelEntry_Internal(dm, dm, domain_label, 1, &label_value, dim, NULL, &ds);CHKERRQ(ierr);
  }
```
which effectively requires that a vertex is returned by `DMGetEnclosurePoint`:
```c
    ierr = DMLabelGetStratumIS(label, ids[i], &pointIS);CHKERRQ(ierr);
    if (!pointIS) continue; /* No points with that id on this process */
    ierr = ISGetIndices(pointIS, &points);CHKERRQ(ierr);
    ierr = DMGetEnclosurePoint(dm, odm, enc, points[0], &point);CHKERRQ(ierr);
    ierr = DMLabelGetValue(depthLabel, point, &pdepth);CHKERRQ(ierr);
    if (pdepth == cdepth) {
      ls = point;
      if (ds) {ierr = DMGetCellDS(dm, ls, ds);CHKERRQ(ierr);}
    }
```

@knepley What would be the correct logic here so that a height 1 label does not need to be propagated to vertices?

knepley: I am not understanding the question. `DMGetFirstLabelEntry_Internal` tries to return a point to you of the height that you ask for in the call. `DMGetEnclosurePoint` is just trying to handle asking for points in related submeshes.

jedbrown: I need this to handle both cases, where the label has been completed and when it hasn't. Is there a better way? It's kind of wasteful to do these hefty intersections for every boundary.
https://gitlab.com/petsc/petsc/-/merge_requests/4786

knepley: As long as you are always asking for a face, it should not matter whether the label has been completed.

LeilaGhaffari: Thanks for the fix @jedbrown . You have probably noticed that `euler_vortex` is now failing.

jedbrown: I think it's clean with my PETSc branch above. If @knepley approves, I'll document and get that branch merged quickly. We don't have to wait for it to merge this branch, but I'd like to know it's okay in principle before proceeding here.

jedbrown (approved):
I rebased on `main` so we can use new PETSc. I think we can merge once the pipeline passes.

jedbrown:
Erm, I missed that the `*.md` documentation needs updating (task up top). Good otherwise as far as I'm concerned.

jedbrown:
What do you make of these HIP-only test failures? https://gitlab.com/libceed/libCEED/-/pipelines/461302258/test_report

I think the lv test failures are just needing a PETSc update from @jeremylt

examples/fluids/README.md
```diff
@@ -107,16 +146,6 @@ For the 2D advection problem, the following additional command-line options are
...
-* - `-lx`
-  - Length scale in x direction
-  - `8000`
-  - `m`
-
-* - `-ly`
-  - Length scale in y direction
-  - `8000`
-  - `m`
-
```
jedbrown, line 119: Somewhere, can we be explicit that lengths like `-dm_plex_box_upper` are given in meters, and will be nondimensionalized according to `-units_meter`?

jedbrown (approved):
Thanks. Looks good to me. In the future, we'll want to allow slip boundaries that are not grid aligned. This is subtle on curved boundaries, but the normal would be obtained from quadrature points (and thus the user won't need to specify the normal).

examples/fluids/README.md
LeilaGhaffari: Good point. If the last commit looks good, I can squash the last two commits and merge.

jedbrown: Docs look good. Have you noticed the failing HIP tests?

LeilaGhaffari: Sorry I missed [your comment](https://github.com/CEED/libCEED/pull/881#issuecomment-1027434984). It seems that HIP is not happy about excluding the outflow QFunction from `eulervortex.h`. I am not sure about the reason though!

jedbrown: I think it's better not to duplicate the struct definition anyway so I think it's fine to merge. We can find the root cause if we have a good reason for splitting these headers in the future.

examples/fluids/qfunctions/euler-outflow.h
```diff
@@ -1,117 +0,0 @@
...
-
-/// @file
-/// Euler outflow BC
-
-#ifndef euler_outflow_h
-#define euler_outflow_h
-
-#include <math.h>
-
-#ifndef euler_context_struct
```
jeremylt: To make this work the way you want, you would need to put the struct in a third file and include it in both qfunction headers.

LeilaGhaffari: Thanks @jeremylt for the hint. I will go ahead and merge this since we don't need to split the headers right now. We can split them when we add other problems that use `euler_outflow`.

PR #882, opened 2022-01-26
CI - update to xsmm 1.17

jeremylt:
When it comes time for the next LIBXSMM release, it looks like there has been a breaking change to some signatures we use.


jedbrown:
Will it be hard to support both (so we'll just require 1.18 or whatever the next release is) or something we can add support for now?

jedbrown (approved):

jeremylt:
Looks like they just deleted the last three args for calling a `libxsmm_*mmfunction`, which we were already setting to `NULL` anyways.

PR #883, opened 2022-01-28
basis - handle null pointers to basis matrices

jeremylt:
Do we want to make some of these matrices mandatory and  error if the pointer is NULL?


rezgarshakeri (approved):
Thanks! LGTM.

jeremylt:
*sigh* it looks like HIP is still throwing a memory tantrum on Noether.

PR #807, opened 2021-09-09
PETSc examples: assemble matrix analytically instead of coloring

jedbrown:
This currently requires development PETSc to correctly handle negative indices through the LocalToGlobalMapping.

PR #885, opened 2022-01-31
PETSc - set minimum version for examples

jeremylt:
This sets the minimum version of PETSc to v3.17 and mentions https://gitlab.com/petsc/petsc/-/commit/3aa6300419c71e062c3b79f2357c12af68281a1b

This commit will be the minimum required once we merge #807, though we may need to push that out again after https://gitlab.com/petsc/petsc/-/merge_requests/4786

I am not wild about repeating this version number in 3 places, but it feels like the easiest way to communicate the requirement and the version number should change very infrequently.


jeremylt:
Updated to set https://gitlab.com/petsc/petsc/-/commit/0e95d8424937f5c91c24c9ee84bd617c5e0333ee as the minimum PETSc dev commit.

examples/petsc/include/petscversion.h
```diff
@@ -0,0 +1,8 @@
+#ifndef petsc_version
+#define petsc_version
```
jedbrown, line 2: This might be legacy, but this isn't our namespace. We can define `libceed_petsc_version_h` or whatever, but not `petsc_version`.

jedbrown (approved):
Looks good. I'm updating Noether and then we can rebase #881

PR #887, opened 2022-02-01
QFContext - create labels for registered fields

jeremylt:
In Ratel we have code that generates an operator that may or may not have a particular named context field. This PR stops `Ceed*ContextSet*()` from failing when the field is not found. Now the function only fails if there is a field/type mismatch. The caller can consult the boolean `is_set` and decide if not finding the field is a problem.

### DIFF
--- a/include/ceed-impl.h
+++ b/include/ceed-impl.h
@@ -274,7 +274,7 @@ struct CeedQFunctionContext_private {
   int (*Destroy)(CeedQFunctionContext);
   CeedInt num_fields;
   CeedInt max_fields;
-  CeedQFunctionContextFieldDescription *field_descriptions;
+  CeedContextFieldLabel *field_labels;
   uint64_t state;
   size_t ctx_size;
   void *data;
@@ -301,6 +301,16 @@ struct CeedFortranContext_private {
             CeedScalar *v15, int *err);
 };
 typedef struct CeedFortranContext_private *CeedFortranContext;
+
+struct CeedContextFieldLabel_private {
+  const char *name;
+  const char *description;
+  CeedContextFieldType type;
+  size_t size;
+  size_t offset;
+  CeedInt num_sub_labels;
+  CeedContextFieldLabel *sub_labels;
+};
 
 struct CeedOperatorField_private {
   CeedElemRestriction elem_restr; /* Restriction from L-vector */
@@ -355,6 +365,9 @@ struct CeedOperator_private {
   CeedOperator *sub_operators;
   CeedInt num_suboperators;
   void *data;
+  CeedInt num_context_labels;
+  CeedInt max_context_labels;
+  CeedContextFieldLabel *context_labels;
 };
 
 #endif
--- a/include/ceed-impl.h
+++ b/include/ceed-impl.h
@@ -307,6 +307,7 @@ struct CeedContextFieldLabel_private {
   const char *description;
   CeedContextFieldType type;
   size_t size;
+  size_t num_values;
   size_t offset;
   CeedInt num_sub_labels;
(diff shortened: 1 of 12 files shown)


jedbrown:
Is this what we want (implies string processing on each iteration) or should we have an interface to get a key and then to set that key? The key can just be the integer offset, but it's intended to be opaque to the user.

jeremylt:
How would the key work for a composite operator? An array of offsets, one for each sub-operator?

I'd be fine adding this new `QFunctionContextLabel` or whatnot to this PR.

jedbrown:
Yeah, in that case the handle could be a new opaque type where setting pokes all the values. If many values need to be set, we could provide them all in a batch. I think it would typically poke the pointers and initiate one or more async memcopies to the device.

jeremylt:
I know the tests aren't quite working, but what about this interface?

include/ceed-impl.h
```diff
@@ -302,6 +302,16 @@ struct CeedFortranContext_private {
...
 
+struct CeedContextFieldLabel_private {
+  const char *name;
+  const char *description;
+  CeedContextFieldType type;
+  size_t size;
+  size_t offset;
+  CeedInt num_sub_labels;
+  CeedContextFieldLabel *sub_labels;
+};
```
jedbrown, line 313: Do we need a tree here or could we make a flat array with the address of all the structs? Perhaps
```c
struct {
  char *host, *device;
  size_t size, offset;
} *locations;

for (i=0; i<n; i++) {
  memcpy(locations[i].host + locations[i].offset, &value_to_set, bytes_of_value);
}
```
followed by an async copy to device (if needed).

jeremylt: My plan was to let the GPU backends handle a copy to the GPU if it is needed and only update the CPU side in the interface.

I don't understand why an array of offsets is better than an array of related labels?

jedbrown: Probably premature optimization, but also trying to make this structure flat and simple so it doesn't get abused for more semantic operations. Your solution is fine -- we'll be able to refactor if it becomes necessary.

jedbrown: Yep, good point. We should do that.

tests/t525-operator.c
```diff
@@ -74,14 +82,20 @@ int main(int argc, char **argv) {
...
   if (ctx_data_2.time != 2.0)
     // LCOV_EXCL_START
     printf("Incorrect context data for time: %f != 2.0", ctx_data_2.time);
   // LCOV_EXCL_STOP
 
   // Check setting field in context of multiple sub-operators for composite operator
-  CeedOperatorContextSetDouble(op_composite, "other", 9000.);
+  CeedOperatorContextGetFieldLabel(op_composite, "other", &other_label);
+  // No issue requesting same label twice
+  CeedOperatorContextGetFieldLabel(op_composite, "other", &other_label);
```
jedbrown, line 96: Looks good. How should we manage lifetimes for this concept in Rust? It's like an immutable borrow (because you need to call `CeedOperatorContextSetDouble` to mutate), but it's also safe to mutate the Operator while borrowed. But this label (maybe we should call it a Key) can't outlive the Operator. The GhostCell concept is similar, but we don't need its performance (lock elision).

jedbrown (approved):

tests/t525-operator.c
jeremylt: We can only express this concept in the Rust interface with gallery QFunctions since the context holds the user QFunction closure for non-gallery functions. I'm not sure how this will shape up once start adding in GPU capability or Gallery QFs with labeled fields. I think we can say that a derived object cannot outlive it's parent.

PR #891, opened 2022-02-03
CI Update

jeremylt:
This
1) Drops Azure
2) Reorganizes the GitLab CI and adds a fast CPU only job


jeremylt:
Should be good, but @jedbrown you'll need to remove the Azure requirement for merges.

jedbrown (approved):
Looks good, thanks.

PR #892, opened 2022-02-07
op - fix small leak in composite ctx label

jeremylt:


pazner:
> The CI failure is unrelated - something is broken in the Julia style formatter. @pazner do you have any theories on what it could be? I ended up swapping to the default formatting on LFAToolkit because I couldn't figure this out when my copy of this file broke over in that repo.

We can pin the version of the Tokenizer.jl package (a transitive dependency), or upgrade to a more recent JuliaFormatter version, which might result in some code style changes… I can look into it.

jeremylt:
Either strategy seems reasonable to me, so whichever you prefer

tests/t525-operator.c
```diff
@@ -105,6 +105,13 @@ int main(int argc, char **argv) {
     printf("Incorrect context data for other: %f != 2.0", ctx_data_2.other);
   // LCOV_EXCL_STOP
 
+  // Check requesting label for field that doesn't exist returns NULL
```
jeremylt, line 108: One of these days we should set up a CI job to run our test suite with a simple CPU backend against Valgrind to check for leaks.

jedbrown (approved):

tests/t525-operator.c
jedbrown: Would `-faddress=sanitize` have caught this? It runs much faster than Valgrind and somewhat avoids needing to build a special MPICH to be valgrind-clean.

jeremylt: I'm not sure; I can go back in history and check. I was thinking of running Valgrind only on the core library tests to bypass all the MPICH, LIBXSMM, CUDA, etc noise.

PR #893, opened 2022-02-07
HIP: add atomics flag and more kernel launch bounds for performance improvements

nbeams:
This PR adds some changes to improve performance of the hip-gen and hipMAGMA backends:
 - Add launch bounds to hip-gen kernels, hip-ref QFunction kernels, and MAGMA backend kernels
 - Add flag for double precision atomics on MI250x: to the Makefile for code compiled when building the library (currently just the MAGMA restrictions), and to the flags passed to hiprtc for most HIP backend kernels.  I have confirmed that adding this flag to hiprtc changes the emitted instruction from `global_atomic_cmpswap_x2` to `global_atomic_add_f64` for gfx90a.  I think it is fine to add it all the time because it seems to be ignored if the `HIP_ARCH` does not support double precision atomics (e.g., when testing on Noether).
 
**Important question to resolve**
The atomics flag only works with ROCm >= 4.2.  Do we want to: 
  - update information such that ROCm 4.2 or newer is required, or 
  - add checks to only add the flag if ROCm is >= 4.2?  (Is there much demand out there to support ROCm 4.1 or older at this point?)

### DIFF
--- a/backends/hip-gen/ceed-hip-gen-operator-build.cpp
+++ b/backends/hip-gen/ceed-hip-gen-operator-build.cpp
@@ -26,6 +26,35 @@
 #include "../hip-shared/ceed-hip-shared.h"
 #include "../hip/ceed-hip-compile.h"
 
+
+//------------------------------------------------------------------------------
+// Calculate the block size used for launching the operator kernel 
+//------------------------------------------------------------------------------
+CEED_INTERN int BlockGridCalculate(const CeedInt dim, const CeedInt nelem,
+		                   const CeedInt P1d, const CeedInt Q1d,
+				   CeedInt *block_sizes) {
+  
+  const CeedInt thread1d = CeedIntMax(Q1d, P1d);
+  if (dim==1) {
+    CeedInt elemsPerBlock = 64*thread1d > 256? 256/thread1d : 64;
+    elemsPerBlock = elemsPerBlock>0?elemsPerBlock:1;
+    block_sizes[0] = thread1d;
+    block_sizes[1] = 1;
+    block_sizes[2] = elemsPerBlock;
+  } else if (dim==2) {
+    const CeedInt elemsPerBlock = thread1d<4? 16 : 2;
+    block_sizes[0] = thread1d;
+    block_sizes[1] = thread1d;
+    block_sizes[2] = elemsPerBlock;
+  } else if (dim==3) {
+    const CeedInt elemsPerBlock = thread1d<6? 4 : (thread1d<8? 2 : 1);
+    block_sizes[0] = thread1d;
+    block_sizes[1] = thread1d;
+    block_sizes[2] = elemsPerBlock;
+  }
+  return CEED_ERROR_SUCCESS;
+}
+
 static const char *deviceFunctions = QUOTE(
 
 //------------------------------------------------------------------------------
@@ -858,7 +887,8 @@ CEED_INTERN int CeedHipGenOperatorBuild(CeedOperator op) {
 
   // Setup
   code << "\n// -----------------------------------------------------------------------------\n";
-  code << "\nextern \"C\" __global__ void "<<oper<<"(CeedInt nelem, void* ctx, HipFieldsInt indices, HipFields fields, HipFields B, HipFields G, CeedScalar* W) {\n";
+  code << "\nextern \"C\" __launch_bounds__(BLOCK_SIZE)\n";
+  code << "__global__ void "<<oper<<"(CeedInt nelem, void* ctx, HipFieldsInt indices, HipFields fields, HipFields B, HipFields G, CeedScalar* W) {\n";
   for (CeedInt i = 0; i < numinputfields; i++) {
     ierr = CeedQFunctionFieldGetEvalMode(qfinputfields[i], &emode);
     CeedChkBackend(ierr);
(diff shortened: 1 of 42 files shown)


jeremylt:
I'm not aware of a strong demand for ROCM 4.1 or earlier. That said, does it take much to check for 4.1 and selectively add the flag?

nbeams:
I know I would need to update the specification of the required ROCm version in the README -- are there any other changes associated with that?

nbeams:
> That said, does it take much to check for 4.1 and selectively add the flag?

Not really -- I think it would mean
1) an additional check in `CeedCompileHip` like the one added to check for ROCm 4.5, and some minor rearrangement of the code that adds the flags (to make it so that the length of the array of flags is changed and the last flag only appended if ROCm >= 4.2)
2) an additional `hipconfig` command in the Makefile used to conditionally add the flag for `hipcc` -- we can use `hipconfig -v` to get the version, and then do some Make string checking stuff 

But it's a little bit of extra complexity that may be obsolete pretty quickly if no one is actually using ROCm 4.1 or older anymore, so I'm not sure there's much of an advantage to adding it.

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -26,6 +26,35 @@
...
+    block_sizes[0] = thread1d;
+    block_sizes[1] = thread1d;
+    block_sizes[2] = elemsPerBlock;
+  } else if (dim==3) {
+    const CeedInt elemsPerBlock = thread1d<6? 4 : (thread1d<8? 2 : 1);
+    block_sizes[0] = thread1d;
+    block_sizes[1] = thread1d;
+    block_sizes[2] = elemsPerBlock;
+  }
+  return CEED_ERROR_SUCCESS;
```
jedbrown, line 55: We still have a function with exactly this name (but `static`) in `ceed-cuda-gen-operator.c`, but that version does a search to minimize "waste". I think these should have different names even though there isn't a symbol collision.

Also, `CEED_INTERN` doesn't have to be repeated here so long as the header with that declaration is included first. (This isn't good or bad, but in most cases it's less surprising to have a single point of authority to determine symbol visibility.)

jedbrown (approved):
Thanks! This looks good in general. I'm having failures in this one test, but it seems that actually started before this so I'll need to track it down. (This is with mobile RTX 2050 Ti, cuda-11.6.)
```
not ok 36 t506-operator /gpu/cuda/shared stderr
# +/home/jed/src/libCEED/backends/cuda/ceed-cuda-compile.c:165 in CeedRunKernelDimSharedCuda(): CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES
```

jedbrown:
The issue with t506 and /gpu/cuda/shared is also present on a 2080 Super with cuda-11.6. @jeremylt Do you see this locally as well?

backends/hip-gen/ceed-hip-gen-operator-build.cpp
jedbrown: `BlockGridCalculate_HIP` would be fine. I think we need to make do with what we've got for now (can't wait for new HIP interfaces). We probably need to query the wavefront size (ROCm now supports Navi, but it's 32 while MI is 64).

Can we have a crude bound so we can ensure the solutions are safe even if not truly optimal?

nbeams: Do you mean a bound in terms of total threads?  The code as-is (which is what was already in hip-gen, I just pulled it out into a function that can be used by both the compilation and application functions for now) should never have more than 256 threads/block, unless thread1d*thread1d is greater than 256, in which case it will have max thread1d*thread1d threads -- and we'd have to go above p1d/q1d = 32 to have too many.

jedbrown: Right, so we can add extra shared memory requirement when calling the function to ensure the solution is safe. If it returns fewer than 256 threads, we could call again with a smaller shared memory placeholder. This may either converge to a fixed point or a limit cycle. In the latter case, we need to choose the safe value.

nbeams: Oh, I see, you meant safe in terms of shared memory usage.

Another problem, whether we have a variable-smem version of `hipOccupancyMaxPotentialBlockSize` or not, is that with hiprtc, using the launch bounds seems to be more important for performance than with nvrtc.  We need the compiled function to call `hipOccupancyMaxPotentialBlockSize` -- but the compiled function will need to include launch bounds -- then if we change the blocksize based on `hipOccupancyMaxPotentialBlockSize`, we would actually need to compile again (and build the code string again, to replace the launch bounds), which then might change the outcome of `hipOccupancyMaxPotentialBlockSize` if the compiler makes changes...meaning this iterative process might become costly in terms of overhead.

jeremylt:
> @jeremylt Do you see this locally as well?

Trying to install ROCm 5.0 borked my ability to use CUDA or HIP, so I can't investigate locally right now.

backends/hip-gen/ceed-hip-gen-operator-build.cpp
jedbrown: Yeah, that is indeed frustrating and relevant to https://github.com/ROCm-Developer-Tools/HIP/issues/924

Seems it would be useful to have the compiler's optimization pass choose launch bounds and we query instead of us probing via an entire compilation unit.

nbeams:
Getting back to this now...

I updated the minor changes discussed above.  For now, everything should be exactly the same as it was before, except that we can get performance improvements from having the launch bounds given to the compiler.  Is this worth merging while we work on/wait for better options?

backends/hip-gen/ceed-hip-gen-operator-build.cpp
jedbrown: Let's resolve this thread so we can merge and we can follow up with AMD liaison about the underlying issue. Thanks for your careful work on this.

PR #897, opened 2022-02-12
Vector: error-free path for get/take array when size=0

jedbrown:
Among other things, this is important so that parallel callers can
easily keep collective control flow even when some subdomains (materials
or boundary surfaces) are size 0 on some ranks.

### DIFF
--- a/interface/ceed-vector.c
+++ b/interface/ceed-vector.c
@@ -387,29 +387,30 @@ int CeedVectorTakeArray(CeedVector vec, CeedMemType mem_type,
                      "Cannot take CeedVector array, a process "
                      "has read access");
   // LCOV_EXCL_STOP
-
-  bool has_borrowed_array_of_type = true;
-  ierr = CeedVectorHasBorrowedArrayOfType(vec, mem_type,
-                                          &has_borrowed_array_of_type);
-  CeedChk(ierr);
-  if (!has_borrowed_array_of_type)
-    // LCOV_EXCL_START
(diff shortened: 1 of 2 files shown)


jeremylt (approved):
This looks good to me. I patched around this in Ratel, as you saw, but hadn't had a chance to make a proper fix yet.

This is the sort of thing that I would like to have a test for but would prefer not to test against every backend. There are a few other tests we have that I would also prefer to only run once.

I think it would be nice to have a way to run some tests with only the default CPU backend to verify our interface logic.

jedbrown:
I think test naming is the easiest way to have a subset of the tests that runs with only one backend, though it adds complexity to our bespoke test running scripts.

Alternative is to add it to an existing test because it takes microseconds at most to make a zero-sized vector and dance with it. This is easy, but tests become less granular.

Another alternative would be to test it from Rust, where `rust/libceed/tests/*.rs` are compiled into one executable and also very fast. (If we want code coverage stats to reflect that coverage, we need to learn how to do that.)

jeremylt:
I think line 32 used to/is supposed to set b to some other value.

jedbrown:
> I think line 32 used to/is supposed to set b to some other value.

It's in a different slot so it's okay. I just didn't notice the direction of the equality because all it needs to check is that `a` was NOT modified, but instead it's checking if it's within epsilon of the value set in a different array.

FWIW, floating point equality is just fine when you're just storing and examining values. It's when you compute it via possibly-reordered arithmetic that you need approximate comparison.

PR #898, opened 2022-02-14
Read-only access for QFContext

jeremylt:


interface/ceed-vector.c
```diff
@@ -858,6 +865,8 @@ int CeedVectorPointwiseMult(CeedVector w, CeedVector x, CeedVector y) {
     y_array = x_array;
   }
 
+  assert(w_array); assert(x_array); assert(y_array);
```
jeremylt, line 868: Added to make `make tidy` happy, unless you have a different preference?

jeremylt:
hmm, I'm seeing `CeedRunKernelDimSharedCuda(): CUDA_ERROR_INVALID_VALUE` on LV

interface/ceed-qfunction.c
```diff
@@ -328,6 +387,65 @@ int CeedQFunctionGetInnerContext(CeedQFunction qf, CeedQFunctionContext *ctx) {
...
+  CeedQFunctionContext ctx;
+
+  ierr = CeedQFunctionGetInnerContext(qf, &ctx); CeedChk(ierr);
+  if (ctx) {
+    ierr = CeedQFunctionIsContextWritable(qf, &is_writable); CeedChk(ierr);
+    if (is_writable) {
+      ierr = CeedQFunctionContextGetData(ctx, mem_type, data); CeedChk(ierr);
+    } else {
+      ierr = CeedQFunctionContextGetDataRead(ctx, mem_type, data); CeedChk(ierr);
+    }
```
jedbrown, line 415: It looks like the default will be `is_context_writable = false` so we'll act as though it's read-only even though the user may write to it. Should the default be writable and the user needs to mark it if not (for the performance benefit)? Or should we make it explicit when the context is first set (`CeedQFunctionSetContextWritable`)?

jedbrown (approved):
At a high level, what does it mean for a context to be writable? Is it just a promise that the qfunction itself won't write or is it a promise that `CeedQFunctionContextSetDouble` won't be used? It seems the docs are ambiguous about this and we need to clarify what writable/mutable means.

jeremylt:
> At a high level, what does it mean for a context to be writable? Is it just a promise that the qfunction itself won't write

This is the intent. `CeedQFunctionSetContextWritable()` is asserting if the user QFunction function writes into the context data or not.

jedbrown:
That sounds good, and we can (eventually) bind it safely on the Rust side.

jeremylt:
Huh, CI is stuck because LV is not avalible

jedbrown:
Just ignore lv for a while. It's having kernel problems.
```
Message from syslogd@lv at Feb 18 08:33:30 ...
 kernel:watchdog: BUG: soft lockup - CPU#14 stuck for 22s! [migration/14:97]

Message from syslogd@lv at Feb 18 08:33:58 ...
 kernel:watchdog: BUG: soft lockup - CPU#14 stuck for 22s! [migration/14:97]

Message from syslogd@lv at Feb 18 08:34:06 ...
 kernel:watchdog: BUG: soft lockup - CPU#8 stuck for 22s! [t410-qfunction-:126264]
```

PR #901, opened 2022-02-17
Fix context memory leak

jeremylt:


interface/ceed-qfunction.c
```diff
@@ -742,6 +742,11 @@ int CeedQFunctionFieldGetEvalMode(CeedQFunctionField qf_field,
 **/
 int CeedQFunctionSetContext(CeedQFunction qf, CeedQFunctionContext ctx) {
   int ierr;
+  if (qf->ctx) {
+    // LCOV_EXCL_START
+    ierr = CeedQFunctionContextDestroy(&qf->ctx); CeedChk(ierr);
+    // LCOV_EXCL_STOP
+  }
```
jedbrown, line 749: The check is done internally so it's safe to just call, like `free(NULL)` is also safe.
```suggestion
  ierr = CeedQFunctionContextDestroy(&qf->ctx); CeedChk(ierr);
```

jedbrown (approved):

PR #902, opened 2022-02-18
backends cuda-shared: fix launch bounds to avoid invalid z dimension

jedbrown:
The typical max z dimension size of a thread block is 64 and we were
computing larger values (like 85) in some cases.


jedbrown:
This fixes CI that we've been bumping into on other branches (including the magma pinned memory todo) so I'll just merge it.

PR #903, opened 2022-02-18
Extract generic Navier-Stokes code into 'newtonian.h'

jrwrigh:
- Extract out the generic NS solver from `densitycurrent.h` into `newtonian.h`
 - `densitycurrent.h` is then dependent on `newtonian.h` to get the "solver" qfunction.


jrwrigh:
Yep, was looking at that now.

jrwrigh:
All the Setup functions have been moved over to `newtonian.c`, extracting out the problem-specific settings. I've also created a "still" initial condition for generic Newtonian problems. Not sure if there's anything I need to do for "generic boundary conditions" though.

jedbrown (commented):
Looks pretty good. I see a bunch of `-Wunused-function` and `-Wunused-variable` warnings. I also get compilation errors with CUDA. Is there a qfunction 
```console
$ make BACKENDS=/gpu/cuda test search=fluids
make: 'lib' with optional backends: /gpu/cuda
1..33
# Test Name: dc_explicit
# TESTARGS: -ceed {ceed_resource} -test -degree 3 -dm_plex_box_faces 1,1,2 -dm_plex_box_lower 0,0,0 -dm_plex_box_upper 125,125,250 -dm_plex_dim 3 -bc_slip_x 5,6 -bc_slip_y 3,4 -bc_Slip_z 1,2
-units_kilogram 1e-9 -center 62.5,62.5,187.5 -rc 100. -thetaC -35. -ts_dt 1e-3 -compare_final_state_atol 1E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-d
c-explicit.bin
not ok 1 fluids-navierstokes /gpu/cuda
ok 2 fluids-navierstokes /gpu/cuda stdout
not ok 3 fluids-navierstokes /gpu/cuda stderr
# +/home/jed/src/libCEED/backends/cuda/ceed-cuda-compile.c:87 in CeedCompileCuda(): NVRTC_ERROR_COMPILATION
# +default_program(57): error: function call is not allowed in a constant expression
# +
# +default_program(65): error: identifier "CeedMemType" is undefined
# +
# +default_program(65): error: identifier "PetscMemType" is undefined
# +
# +default_program(66): error: identifier "PetscMemTypeDevice" is undefined
# +
# +default_program(66): error: identifier "CEED_MEM_DEVICE" is undefined
# +
# +default_program(66): error: identifier "CEED_MEM_HOST" is undefined
# +
# +default_program(150): error: identifier "PETSC_MAX_PATH_LEN" is undefined
# +
# +default_program(151): error: identifier "PetscInt" is undefined
[...]
```
I can debug this if it's not clear to you why it's happening.

jrwrigh:
I've addressed the `-Wunused-variable` issues, but I'm a bit confused on the `-Wunused-function`. All three of the functions listed are used and set in `problems/newtonian.c`, which `problems/densitycurrent.c` then uses to setup it's required Qfunctions. The only difference I see is that the current functions are being called more indirectly?

jedbrown:
I pushed a fix for the `-Wunused-function` (due to the unnecessary include) and the missing context stuff. I'd like to refactor this in a future PR, but no reason to block this one on the bit of repetition. Let's see the pipeline.

jrwrigh:
I've also been working on laminar flat plate boundary layer problem setup. I know you mentioned possibly throwing that in this PR, but I've been running into some difficulties with the boundary conditions. I think it makes sense to throw that in a different PR or shift it over to #868.

jedbrown (approved):
CI host is finally back up and this passes. @LeilaGhaffari Would you like to review or shall we merge now?

examples/fluids/navierstokes.h
```diff
@@ -262,19 +262,35 @@ struct AdvectionContext_ {
...
-  AdvectionContext     advection_ctx;
-  WindType             wind_type;
-  BubbleType           bubble_type;
-  BubbleContinuityType bubble_continuity_type;
-  EulerTestType        euler_test;
-  StabilizationType    stab;
-  PetscBool            implicit;
-  PetscBool            has_curr_time;
-  PetscBool            has_neumann;
+  NewtonianIdealGasContext newt_ig_ctx;
```
LeilaGhaffari, line 283: Can we change `newt_ig_ctx` to `newtonian_ig_ctx`? We have been trying to choose more descriptive variable names.

LeilaGhaffari (commented):
This looks great. Thanks, @jrwrigh .
I just have a question (@jedbrown ), I was envisioning something like `newtonian.h` for the solvers and some other header files for different ICs. e.g., `newtonian-densitycurrent.h` and `newtonian-idealgas.h`. That would be more readable  to me (even by looking at the files' names). We could always refactor more in follow-up PRs but this has been my plan for refactoring the fluids mini-app and I am wondering if this is not a good design?!

jedbrown:
I have no objection to that naming, but I also don't know anyone who wants to write a non-ideal/non-Newtonian solver in the near future so I'm not opposed to renaming when that time comes (and hopefully we've been able to remove the odd duplication).

examples/fluids/navierstokes.h
jedbrown: I just pushed this, thanks.

PR #906, opened 2022-02-23
make - fix env BACKENDS detection

jeremylt:
Small bug I made ages ago and just discovered

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -453,11 +453,8 @@ ifneq ($(wildcard $(MAGMA_DIR)/lib/libmagma.*),)
   BACKENDS_MAKE += $(MAGMA_BACKENDS)
 endif
 
-BACKENDS_ENV ?= BACKENDS
-ifneq ($(BACKENDS_ENV),)
-  BACKENDS = $(BACKENDS_MAKE)
-  export BACKENDS
-endif
+BACKENDS ?= $(BACKENDS_MAKE)
+export BACKENDS
 
 _pkg_ldflags = $(filter -L%,$(PKG_LIBS))
 _pkg_ldlibs = $(filter-out -L%,$(PKG_LIBS))


Makefile
```diff
@@ -453,8 +453,8 @@ ifneq ($(wildcard $(MAGMA_DIR)/lib/libmagma.*),)
   BACKENDS_MAKE += $(MAGMA_BACKENDS)
 endif
 
-BACKENDS_ENV ?= BACKENDS
-ifneq ($(BACKENDS_ENV),)
+BACKENDS_ENV = $(BACKENDS)
+ifeq ($(BACKENDS_ENV),)
   BACKENDS = $(BACKENDS_MAKE)
   export BACKENDS
 endif
```
jedbrown, line 460: What semantic are you looking for in which this isn't sufficient?
```suggestion
BACKENDS ?= $(BACKENDS_MAKE)
```

jedbrown (approved):

Makefile
jeremylt: It's really just an artifact of getting the old broken code to work. I think your change would be preferable.

jeremylt: Woops, we still need to export ttis variable for CI... We should meld the two

Makefile
```diff
@@ -453,11 +453,7 @@ ifneq ($(wildcard $(MAGMA_DIR)/lib/libmagma.*),)
   BACKENDS_MAKE += $(MAGMA_BACKENDS)
 endif
 
-BACKENDS_ENV ?= BACKENDS
-ifneq ($(BACKENDS_ENV),)
-  BACKENDS = $(BACKENDS_MAKE)
-  export BACKENDS
-endif
+BACKENDS ?= $(BACKENDS_MAKE)
```
jedbrown, line 456:
```suggestion
BACKENDS ?= $(BACKENDS_MAKE)
export BACKENDS
```

PR #907, opened 2022-02-23
Fix basis qref size mismatch

jeremylt:
Small bigfix. Didn't change correctness but caused Valgrind confusion


jedbrown (approved):
I think this is okay because reference element geometric dimension matches its topological dimension, even if the coordinate transform embeds it in a higher physical geometric dimension. Does that sound right?

jeremylt:
I believe that is correct

PR #908, opened 2022-02-25
Documentation: small changes to notation

nbeams:
This PR changes the documentation to use E instead of G for element restriction operators, as is used in the libCEED JOSS paper.  

It also adds a new schematic to the libCEED API section.


nbeams:
Any and all comments/requests for changes on the new schematic welcome (its location in the documentation and/or the schematic itself).

Once we are happy with all changes, we can squash-merge, or I can do a rebase and squash into a succinct set of commits, whatever is preferred.

doc/sphinx/source/libCEEDapi.md
```diff
@@ -307,16 +307,22 @@ specific devices (e.g. GPUs) or specific polynomial orders. A simple reference
 backend implementation is provided in the file
 [ceed-ref.c](https://github.com/CEED/libCEED/blob/main/backends/ref/ceed-ref.c).
 
+(fig-operator-schematic)=
+
+:::{figure} ../../img/libceed_schematic.png
+Flow of data through vector types inside libCEED Operators
```
valeriabarra, line 313: Hi @nbeams thank you so much for adding this and renaming all the G's with \mathcal{E}. Would you consider adding a bit of descriptive text of this image, perhaps above in section "Terminology and Notation" (L109), where we introduce our "Vector representation/storage categories"

jeremylt:
Probably not worth adding to the diagram, but it might be worth noting that we support passive outputs (and use heavily in our solids/Ratel)

doc/sphinx/source/libCEEDapi.md
nbeams: Hmm, do you think the schematic would fit better in a different location?  I'm not sure if the descriptive text should be so far away from the figure itself...?

nbeams:
> Probably not worth adding to the diagram, but it might be worth noting that we support passive outputs (and use heavily in our solids/Ratel)

I don't mind adding if you think it would be worthwhile.  Should I?

jeremylt:
> I don't mind adding if you think it would be worthwhile. Should I? Or just note in the text somewhere?

A note to that effect? I think it would be helpful. Maybe wherever the diagram is referenced? I'm not sure if we actually talk about passive input/outputs at all?

nbeams:
Can you give me a concrete example of a passive output in Ratel?  Though I know the code allows it, I'm having a hard time understanding what it means in practical terms -- if it's an output which is created/computed by the operator from its inputs, what makes it passive vs active?

jeremylt:
A passive output in Ratel would be the state related data in the residual evaluation to be used in the jacobian evaluation

nbeams:
Would it be worth it to change the schematic so labels match the operator in the example (u, v, rho)?
Or maybe two versions: a generic one included first (maybe also with a line for passive outputs?) and then one with specific labels corresponding to the example?  Maybe also with the names in the boxes corresponding to the object names in the code snippets?  Overkill?  Or useful?

jedbrown:
I find it hard to exaggerate the value of well-annotated figures. We'll use them in talks and they help people get started and then teach their colleagues. So if you have the appetite for it, by all means go wild. I used Inkscape to create the SVG version. This plugin makes it easy to write LaTeX math inside Inkscape. https://textext.github.io/textext/usage/gui.html

nbeams:
In 02076a18a7ce56edb5e6b37c4443befb188aec8a I updated the schematic and added an attempt at an "operator-specific" version that corresponds directly to the sample code used in the documentation. 

I was thinking I might make a specific version for the setup operator as well, but wanted to get feedback first to reduce the amount of "propagated editing" across all versions of the schematic...

jedbrown:
I like it these flow diagrams. I'm not sure what "basis data" is here. Is it just what's part of `B`? https://github.com/CEED/libCEED/blob/02076a18a7ce56edb5e6b37c4443befb188aec8a/doc/img/libceed_schematic.svg

Would it make sense to write one for a nonlinear solve, with geometric setup -> nonlinear residual -> Jacobian action?

I think it's not necessary to annotate line numbers -- function names are enough in my opinion.

jeremylt:
I agree. I think a simpler diagram showing our 3 stage workflow for non-linear problems generally would be most helpful.

nbeams:
>  I'm not sure what "basis data" is here. Is it just what's part of B? 

Yeah, it's just the interp/grad/weight data for the basis.  Since I originally made the schematic as a tool for talking about mixed-precision use in the GPU backends, I included it since the basis data arrays are additional inputs to the basis kernels, and could potentially be in lower or higher precision based on the options.  Do I gather you think we should remove it entirely from the documentation version of the schematic?

> Would it make sense to write one for a nonlinear solve, with geometric setup -> nonlinear residual -> Jacobian action

Do you have a particular example/set of operators in mind to use?  I think it might be useful to have the object names exactly correlate to the ones in a particular example, as with the mass operator.  Though probably no annotations necessary on these, if we've already had one example with that level of detail.

jedbrown:
I think the basis data can be elided. It's set in advance to define the basis, but not passed as fields.

The solids example (here and Ratel) are good examples of passive output from the residual flowing on as passive input to the Jacobian-apply. The bottom line (`FSCurrent-NH2`) is perhaps most interesting, with derivative of reference coordinate through to current configuration (a state that depends on the solution) and the Kirchoff stress `\tau` being passive outputs of residual and passive inputs to Jacobian.

jedbrown:
And yes with respect to `QFunctionAddOutput`.

nbeams:
So based on the discussion of how three operators work in succession in the solid mechanics example, I decided to try combining setup and mass schematics into one figure, making it clear that the active output of the setup operator becomes a passive input to the mass operator.  I also changed the annotations as suggested.  Here is a (low-quality, sorry) screenshot of the current SVG:
image

But when I try building the documentation locally and then looking at the result in a browser, some of the text annotations are rendering in the wrong place.  They look fine when I open the SVG in Inkscape.  Has anyone run into this before?  I tried doing some editing in the SVG file directly to remove translations in the location and just set the x and y, but the result looks the same after building the documentation. 

I also realized I probably need to be more careful about which fonts I use if doing SVG vs a PNG.  Is there a list of which fonts will work with the doc build system (not just what I can do locally on my computer)?

jedbrown:
That's really nice annotation. Can you open the raw SVG directly in your browser (as in `firefox the/figure.svg`)?

Inkscape has an export as "plain SVG" that might help. I don't know the difference between inkscape SVG and plain SVG.

nbeams:
Hmm, yes, when I open the SVG directly in Firefox (or Chrome), I get the same issue.  Occurs with plain SVG or Inkscape SVG.   Weird.  Well, I'll keep playing with it.

jedbrown:
Shot in the dark, but does grouping fix this issue?
https://alpha.inkscape.org/vectors/www.inkscapeforum.com/viewtopic9991.html?t%3D30964#p93298

Also, a trimmed version of this figure (dropping API annotations, but with the flow for Ratel residual and Jacobian) would be really useful in the paper.

valeriabarra:
> image
> 
I personally liked the original sketch as it was simpler than this more annotated one, which seems a bit too crowded. But I can see the value of these annotations. Perhaps do we want to add both images in different sections? One more introductory and one when we go over in details? I'd understand if people just want to keep one though. So no problem with me

nbeams:
@valeriabarra My plan was to have the simple/plain version first, then add this detailed one directly by the part of the documentation that discusses the t500 test and how to set up the libCEED operators, which I did in the commits I just pushed. I was worried about the annotated version seeming too crowded, especially after changing line numbers for function names.  Do you think it would be better to have the annotations only for one of the operators (setup or mass)?  Personally, when I was learning libCEED it took me a bit to understand that the basis eval mode was set with a "QFunction function" rather than an "Operator function," for example, so I think making this very obvious somewhere could be useful to new users.

@jedbrown Unfortunately, grouping didn't help the problem.  Eventually I gave up and removed the offending text boxes and replaced them -- for some reason this worked...until I changed the fonts and some of the basis text boxes had the same issue...also finally fixed through delete and replace... 

> Also, a trimmed version of this figure (dropping API annotations, but with the flow for Ratel residual and Jacobian) would be really useful in the paper.

Would you still want this level of detail with basis eval modes, QFunction name, E and Q vector "flow," etc., or more just a series of operator boxes with L-Vector inputs and outputs being connected to each other? (Basically not showing the "insides" of the operators at all?)

I changed all the fonts to be Arial or Courier New since the all-knowing Google told me these are among the most widely-supported and portable fonts.  What about the text rendered through the TexText extension?  Will that work in a portable way?  I've never used it before.  In the SVG file itself there is stuff like 

```
       ns3:version="1.8.1"
       ns3:texconverter="pdflatex"
       ns3:pdfconverter="inkscape"
       ns3:text="\\texttt{rho}$^\\text{L}$"
       ns3:preamble="/Users/natalie/Library/Application Support/org.inkscape.Inkscape/config/inkscape/extensions/textext/default_packages.tex"
``` 
which worries me...?

My main problem now is the appearance of two random backslashes in the SVG when rendered in the browser: 
image

Like with the spacing issues, these do not show up in Inkscape, but do show up in Firefox/Chrome...arrrrrrggggh!

I wonder if my issues are caused in part by the fact that I imported from a PDF to Inkscape?  (The original schematic was made in Keynote, then I exported that to PDF, then imported to Inkscape for SVG...)

valeriabarra:
> @valeriabarra My plan was to have the simple/plain version first, then add this detailed one directly by the part of the documentation that discusses the t500 test and how to set up the libCEED operators, which I did in the commits I just pushed.

Oh, yes, that sounds like a reasonable plan to me! 

>  Do you think it would be better to have the annotations only for one of the operators (setup or mass)? Personally, when I was learning libCEED it took me a bit to understand that the basis eval mode was set with a "QFunction function" rather than an "Operator function," for example, so I think making this very obvious somewhere could be useful to new users.

Oh, yes, I can definitely see that type of confusion and I personally think it would be great to annotate both operators here in this more detail-oriented version of the picture (since they show different types of basis evals), while keeping the basic one just plain for the introduction. Thanks a lot Natalie, this is great and greatly appreciated!!

jedbrown:
Doh, I'm so sorry you're up against this saga. I think textext embeds its glyphs so they should be perfectly portable. I'm afraid I don't know why this is rendering weird. I think I had this experience some years ago when I imported a PNG to annotate and wanted to view the result as an SVG (so my annotation text would be sharp even if the figure behind was raster).

For the paper, I'm thinking something simple and you're right that distinguishing the full (restriction, operator, qfunction) triplets is too much detail. Actually, horizontal is probably best, just showing geometry and state `u` coming into the residual op with two outputs, residual `f` down and linearization `\tau, \partial \hat X/\partial x, \lambda \log J` to the right. Then `\delta u` and that stuff :point_left: comes into the Jacobian op and `\delta f` comes out. I can draw it up if this is unclear or you're busy.

jedbrown:
Hooray. I think it's a common pattern for nonlinear problems and not one people reading the docs may have encountered before so figures are good. We'd use it here and in Ratel docs (where our ongoing solids work lives).

jedbrown:
I'd think the simplified schematic would also appear here and in the paper. So if you've got things worked out to do both, I'd say yes. If it's a lot more work, then no problem to triage it.

jedbrown (approved):
This looks good to me. Is it ready to merge?

PR #910, opened 2022-03-01
examples/solids: Fix leftover rst formatting

wence-:
Some minor `:code:` sigils left after the transition from rst to myst.


jedbrown (approved):
Thanks!

PR #900, opened 2022-02-16
PETSc BPs with tetrahedral elements

rezgarshakeri:

PR #912, opened 2022-03-04
examples/petsc: let DMSetFromOptions manage distribution

jedbrown:
Sphere meshes are automatically snapped to the unit sphere when refined.

I left explicit partitioning in solids (though Ratel uses the new
paradigm) because it has documented command-line arguments and maybe
someone wants to know how to avoid this distribution.

@rezgarshakeri This is somewhat related to #900. Partitioning and distribution is done automatically now in `DMSetFromOptions`, and our parallel tests are broken with latest PETSc as a result. Automatic distribution can be disabled with `DMPlexDistributeSetDefault()`.

### DIFF
--- a/examples/petsc/area.c
+++ b/examples/petsc/area.c
@@ -136,28 +136,15 @@ int main(int argc, char **argv) {
     // Create the mesh as a 0-refined sphere. This will create a cubic surface, not a box
     ierr = DMPlexCreateSphereMesh(PETSC_COMM_WORLD, topo_dim, simplex, 1., &dm);
     CHKERRQ(ierr);
+    if (problem_choice == CUBE) {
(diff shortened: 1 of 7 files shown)


rezgarshakeri:
> Sphere meshes are automatically snapped to the unit sphere when refined.
> 
> I left explicit partitioning in solids (though Ratel uses the new paradigm) because it has documented command-line arguments and maybe someone wants to know how to avoid this distribution.
> 
> @rezgarshakeri This is somewhat related to #900. Partitioning and distribution is done automatically now in `DMSetFromOptions`, and our parallel tests are broken with latest PETSc as a result. Automatic distribution can be disabled with `DMPlexDistributeSetDefault()`.

Ah, thanks. I didn't know that, I'll delete `dm_dist` part in my branch.

rezgarshakeri (approved):
LGTM, thanks.

PR #913, opened 2022-03-09
Create CeedSize as ptrdiff_t

jeremylt:


jedbrown:
Do any language bindings need to be updated? I guess they're all passing, which is great. I don't know if this path fully works with int32 PETSc, but we should at least test that we can assemble with COO arrays beyond 2B on Noether with an int64 PETSc. I guess Ratel is the place to test that because it can convert the index arrays.

jeremylt:
I don't believe that the assembly functions have made it to the other language bindings.

jeremylt:
The question becomes - which 'length like' quantities should use `CeedSize`?

ElemRestriction `elem_size`, `num_comp`, `comp_stride`, and `l_size` (probably `l_size` at least because its an array length)
Basis `P`, `Q`, `dim`, `num_comp`
QFunction `field_size` (used to create vectors of inputs for User QF)
QFunctionContext `size` (currently use `size_t`)

include/ceed/ceed.h
```diff
@@ -119,15 +119,20 @@
...
+#include <stdint.h>
+#include <stdio.h>
 
 /// Integer type, used for indexing
 /// @ingroup Ceed
 typedef int32_t CeedInt;
 
+/// Integer type, used array sizes
+/// @ingroup Ceed
+typedef ptrdiff_t CeedSize;
```
jedbrown, line 134: Should it be named like this because `size_t` is unsigned and this is signed? On the other hand, this is like Rust's `isize`. No strong preference, but want the decision to be deliberate.
```suggestion
typedef ptrdiff_t CeedCount;
```

interface/ceed-elemrestriction.c
```diff
@@ -991,7 +991,7 @@ int CeedElemRestrictionView(CeedElemRestriction rstr, FILE *stream) {
   else
     sprintf(stridesstr, "%d", rstr->comp_stride);
 
-  fprintf(stream, "%sCeedElemRestriction from (%d, %d) to %d elements with %d "
+  fprintf(stream, "%sCeedElemRestriction from (%ld, %d) to %d elements with %d "
```
jedbrown, line 994:
```suggestion
  fprintf(stream, "%sCeedElemRestriction from (%td, %d) to %d elements with %d "
```
From `man 3 printf`
>  `t`      A following integer conversion corresponds to a ptrdiff_t argument, or a following n conversion corresponds to a pointer to a ptrdiff_t argument.

jedbrown (approved):
This is a breaking change that needs release notes. Once merged, we should update https://github.com/mfem/mfem/pull/2569#issuecomment-1065223885 ASAP so it doesn't cause them pain.

include/ceed/ceed.h
jeremylt: I was thinking size like isize but I'm fine with either.

jedbrown: That'll smooth things out when we change the library core to Rust. :smile:

PR #916, opened 2022-03-16
release -  prepare for v0.10

jeremylt:


jeremylt:
I haven't started publishing crates or anything - this is just the version numbering changes so far

pazner:
Julia changes look good to me.

CI is failing because it depends on libCEED_jll v0.10, which will only exist after libCEED v0.10 is released.

Once that happens, I can handle updating libCEED_jll and registering the new version of LibCEED.jl.

jeremylt:
@pazner, I removed the Julia changes so this passes CI. I'm going to tag v0.10 once this merges. Can you include the needed Julia changes in a follow-up PR after your updates via Yggdrasil?

pazner:
Yes, I think that should work!

PR #917, opened 2022-03-16
Visibility and AVX flag detection

jedbrown:


jeremylt (approved):
LGTM, should be fine in CI with #918

PR #920, opened 2022-03-17
Explicit casting of vector sizes

### LINKED ISSUE #927: GPU inline helpers
The GPU backends have a few static inline helper functions that return values instead of error codes but use CeedChkBackend internally. CeedChk returns an error code if an error is found, violating the assumed return type for these functions

jeremylt:
I added some explicit casting around `CeedSize` for CeedVector lengths

Closes #927

### DIFF
--- a/backends/blocked/ceed-blocked-operator.c
+++ b/backends/blocked/ceed-blocked-operator.c
@@ -29,6 +29,7 @@ static int CeedOperatorSetupFields_Blocked(CeedQFunction qf,
     CeedVector *e_vecs_full, CeedVector *e_vecs, CeedVector *q_vecs,
     CeedInt start_e, CeedInt num_fields, CeedInt Q) {
   CeedInt ierr, num_comp, size, P;
+  CeedSize e_size, q_size;
   Ceed ceed;
   ierr = CeedOperatorGetCeed(op, &ceed); CeedChkBackend(ierr);
   CeedBasis basis;
@@ -94,8 +95,8 @@ static int CeedOperatorSetupFields_Blocked(CeedQFunction qf,
     switch(eval_mode) {
     case CEED_EVAL_NONE:
       ierr = CeedQFunctionFieldGetSize(qf_fields[i], &size); CeedChkBackend(ierr);
-      ierr = CeedVectorCreate(ceed, Q*size*blk_size, &q_vecs[i]);
-      CeedChkBackend(ierr);
+      q_size = (CeedSize)Q*size*blk_size;
+      ierr = CeedVectorCreate(ceed, q_size, &q_vecs[i]); CeedChkBackend(ierr);
       break;
     case CEED_EVAL_INTERP:
     case CEED_EVAL_GRAD:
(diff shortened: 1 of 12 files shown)


backends/blocked/ceed-blocked-operator.c
```diff
@@ -94,23 +95,24 @@ static int CeedOperatorSetupFields_Blocked(CeedQFunction qf,
     switch(eval_mode) {
     case CEED_EVAL_NONE:
       ierr = CeedQFunctionFieldGetSize(qf_fields[i], &size); CeedChkBackend(ierr);
-      ierr = CeedVectorCreate(ceed, Q*size*blk_size, &q_vecs[i]);
-      CeedChkBackend(ierr);
+      q_size = Q*size*blk_size;
```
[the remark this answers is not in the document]
jeremylt: Thanks, updated

backends/blocked/ceed-blocked-operator.c
```diff
@@ -94,23 +95,24 @@ static int CeedOperatorSetupFields_Blocked(CeedQFunction qf,
     switch(eval_mode) {
     case CEED_EVAL_NONE:
       ierr = CeedQFunctionFieldGetSize(qf_fields[i], &size); CeedChkBackend(ierr);
-      ierr = CeedVectorCreate(ceed, Q*size*blk_size, &q_vecs[i]);
-      CeedChkBackend(ierr);
+      q_size = (CeedSize)Q*size*blk_size;
+      ierr = CeedVectorCreate(ceed, q_size, &q_vecs[i]); CeedChkBackend(ierr);
```
jedbrown, line 99: Your choice here. It isn't necessary to make the temp variables, but they don't hurt.

jedbrown (approved):
This one is a real bug if there's ever an error.
```
/home/jed/src/libCEED/backends/cuda-ref/ceed-cuda-ref-qfunctioncontext.c: In function ‘bytes’:
./include/ceed/backend.h:103:105: warning: unsigned conversion from ‘int’ to ‘size_t’ {aka ‘long unsigned int’} changes value from ‘-2’ to ‘18446744073709551614’ [-Wsign-conversion]
  103 | #define CeedChkBackend(ierr) do { int ierr_ = ierr; if (ierr_) { if (ierr_ > CEED_ERROR_SUCCESS) return CEED_ERROR_BACKEND; else return ierr_; } } while (0)
      |                                                                                                         ^~~~~~~~~~~~~~~~~~
```

The relevant code
```c
static inline size_t bytes(const CeedQFunctionContext ctx) {
  int ierr;
  size_t ctxsize;
  ierr = CeedQFunctionContextGetContextSize(ctx, &ctxsize); CeedChkBackend(ierr);
  return ctxsize;
}
```
then
```c
    ierr = cudaMalloc((void **)&impl->d_data_owned, bytes(ctx));
```

backends/blocked/ceed-blocked-operator.c
jeremylt: I added it because it made the casing and the purpose of the value a bit clearer to me, personally.

PR #924, opened 2022-03-18
readme - drop azure badge

jeremylt:
We aren't testing on Azure pipelines anymore: https://github.com/CEED/libCEED/pull/891


jedbrown (approved):
This is just hung up on lv CI so may as well merge.

PR #925, opened 2022-03-18
Add some matrix assembly support to GPU backends

nbeams:
This PR adds a basic reference implementation of matrix assembly support for the GPU backends.  It works for the same types of operators as the current interface CPU implementation (i.e., same element restriction and basis objects for active input and output fields); expansion to more complicated operators will be a future feature for both CPU and GPU.  ~The main limitation of this GPU version compared to the current CPU implementation is that the total number of basis nodes in an element is limited to 32 (P3 tets, Q2 hexes).~  I added a fallback kernel with a 1D threadblock; it gets much worse performance than the 2D version when P is small enough to use the 2D version, but it extends the orders that can be built on the GPU.

### DIFF
--- a/backends/hip-ref/ceed-hip-ref-operator.c
+++ b/backends/hip-ref/ceed-hip-ref-operator.c
@@ -1249,10 +1249,10 @@ static int CeedOperatorLinearAssembleAddPointBlockDiagonal_Hip(CeedOperator op,
 }
 
 //------------------------------------------------------------------------------
-// Matrix assembly kernel
+// Matrix assembly kernel for low-order elements (2D thread block)
 //------------------------------------------------------------------------------
 // *INDENT-OFF*
-static const char *assemblykernels = QUOTE(
+static const char *assemblykernel = QUOTE(
 extern "C" __launch_bounds__(BLOCK_SIZE) 
            __global__ void linearAssemble(const CeedScalar *B_in, const CeedScalar *B_out,
                    const CeedScalar *__restrict__ qf_array,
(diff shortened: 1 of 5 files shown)


backends/hip-ref/ceed-hip-ref-operator.c
```diff
@@ -1239,6 +1248,353 @@ static int CeedOperatorLinearAssembleAddPointBlockDiagonal_Hip(CeedOperator op,
...
+  asmb->nnodes = esize;
+
+  // Compile kernels
+  int elemsPerBlock = 1;
+  asmb->elemsPerBlock = elemsPerBlock;
+  // TODO: 1D threadblock version for larger elements
+  if (esize * esize * elemsPerBlock > 1024)
+    // LCOV_EXCL_START
+    return CeedError(ceed, CEED_ERROR_UNSUPPORTED,
+                     "GPU assembly not currently available for elements of this size");
```
jedbrown, line 1428: Could you maybe print the sizes and the bound to make this message more informative?

jedbrown (approved):
This looks great to me assuming it clears CI and the latest update is good with Ratel. Hugely valuable feature.

backends/hip-ref/ceed-hip-ref-operator.c
jeremylt: Can we make the error message use the standard "not implemented" verbiage or (better yet) fall back to CPU assembly?

jeremylt:
One Ratel test fails - t102. In this test we compare the action of a sparse matrix representation with the action of a matrix-free implementation on Q3 hexes. I can drop it down to Q2 in that test, but it would be nice to have some fallback plan for Q3+ element assembly to facilitate comparisons between matrix-free and sparse matrix representations.

backends/hip-ref/ceed-hip-ref-operator.c
```diff
@@ -1239,6 +1248,353 @@ static int CeedOperatorLinearAssembleAddPointBlockDiagonal_Hip(CeedOperator op,
...
+  // LCOV_EXCL_STOP
```
jeremylt, line 1429: maybe something like this?

```suggestion
  if (esize * esize * elemsPerBlock > 1024) {
    // Do some cleanup here
    /* ToDo */

  ierr = CeedSetBackendFunction(ceed, "Operator", op,
                                "LinearAssemble", NULL); CeedChkBackend(ierr);
  // Set some flag to re-try interface level impl and return
  }

// ...

// Back in CeedOperatorLinearAssemble_Hip
// if falling back
ierr = CeedOperatorLinearAssemble(op, values); CeedChkBackend(ierr);
return CEED_ERROR_SUCCESS;
```

jeremylt: If we do this, we probably want to include a `CeedDebug256` statement indicating we're falling back.

nbeams:
Since there is already a test that uses it, I just went ahead and added a basic fallback kernel.  We can go back to the CPU fallback idea if you want.  Also, this will still fail when P > 1024.  Should I still add a CPU fallback (or just a more descriptive error message than the one I had originally?  I hope most people won't try to assemble a matrix for elements with P > 1024....)

jedbrown:
Nah, P=1024 is plenty big and you want to use fast (FFT-like) transforms well before that anyway.

nbeams:
All the Ratel tests pass for me now, except 

```
$ ./build/t103-static-elasticity -ceed /gpu/hip -yml tests/fs-initial-nh-elasticity.yml 
Error in Jacobian: Y_aij - Y_shell = 1.22356e-13

$./build/t103-static-elasticity -ceed /gpu/hip -yml tests/fs-initial-mr-elasticity.yml 
Error in Jacobian: Y_aij - Y_shell = 1.65463e-13
```

However, if I specify `gpu/hip/ref` instead of `gpu/hip`, the tests pass, so I'm not sure what's going on...?

nbeams:
Hmm, I get the same error on t103-static-elasticity with `-ceed /cpu/self/ref`.  So maybe nothing to do with the HIP backend specifically?

jeremylt:
Maybe slight rounding issues. We can relax the tolerances a bit

jeremylt (approved):

jeremylt:
Once this updates, we should update the Ratel CI

jeremylt:
And Ratel CI is updated and passing. Thanks @nbeams!

PR #934, opened 2022-03-29
Prepare for v0.10.1 patch

jeremylt:
The flop counting interface currently only lives in the C API, so I think we only need to tag the patch on the C API.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,6 +1,6 @@
 cff-version: 1.2.0
 title: "libCEED: Efficient Extensible Discretization"
-version: 0.10.0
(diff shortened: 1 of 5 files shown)


jeremylt:
In a shocking turn of events, CI on LV is flaky. This PR is fine to merge though.

jedbrown (approved):
Should we make a `release` branch so we don't have to flip-flop `CEED_VERSION_RELEASE` but instead can just increment on `release` and merge to `main` where it always stays `false`?

jeremylt:
That all sounds reasonable, so perhaps we start in the fall with that?

PR #938, opened 2022-04-08
minor - unify strncpy vs memcpy, silence errors

jeremylt:
This standardizes `strncpy` vs `memcpy` in the interface level and silences those two warnings.

### DIFF
--- a/interface/ceed-jit-tools.c
+++ b/interface/ceed-jit-tools.c
@@ -36,7 +36,7 @@ int CeedCheckFilePath(Ceed ceed, const char *source_file_path, bool *is_valid) {
 
     ierr = CeedCalloc(source_file_path_length, &source_file_path_only);
     CeedChk(ierr);
-    strncpy(source_file_path_only, source_file_path, source_file_path_length - 1);
+    memcpy(source_file_path_only, source_file_path, source_file_path_length - 1);
   } else {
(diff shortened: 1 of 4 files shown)


interface/ceed-jit-tools.c
```diff
@@ -36,7 +36,7 @@ int CeedCheckFilePath(Ceed ceed, const char *source_file_path, bool *is_valid) {
 
     ierr = CeedCalloc(source_file_path_length, &source_file_path_only);
     CeedChk(ierr);
-    strncpy(source_file_path_only, source_file_path, source_file_path_length - 1);
+    memcpy(source_file_path_only, source_file_path, source_file_path_length - 1);
```
jedbrown, line 39: I'd honestly feel safer about using `CeedMalloc` and copying including the null byte.
```suggestion
    memcpy(source_file_path_only, source_file_path, source_file_path_length);
```

jedbrown (approved):

interface/ceed-jit-tools.c
jeremylt: The -1 in this line actually is important because we're copying a segment that stops before the end of the string. I think we could use Malloc or Calloc interchangeably in all other cases.
```
// Sometimes we have path/to/file.h:function_name
// Create temporary file path without name, if needed
```

PR #940, opened 2022-04-13
petsc - update for PetscOptionsBegin/End

jeremylt:
I have only updated for PetscOptionsBegin/End no longer returning error codes. Do we also want to include swapping over to PetscCall().


knepley:
I will do that after we get this working

jedbrown (approved):

PR #945, opened 2022-04-20
Remove 'quoted' operator assembly kernels

jeremylt:
I moved these JiTed kernels into separate files like we do with other similar kernels.


jedbrown (approved):
Cool, looks more maintainable this way.

PR #946, opened 2022-04-21
Move CeedQFunctionContextGetFieldLabel to backend interface

jeremylt:
With the addition of `CeedOperatorContextGetFieldLabel()`, users shouldn't need to use `CeedQFunctionContextGetFieldLabel()`. Same with `CeedQFunctionContextSet*()`. I left `CeedQFunctionContextGetAllFieldLabels()` and `CeedContextFieldLabelGetDescription()` as those should still be useful to general users who want to make pretty output about a QFunctionContext.

### DIFF
--- a/interface/ceed-qfunctioncontext.c
+++ b/interface/ceed-qfunctioncontext.c
@@ -231,7 +231,7 @@ int CeedQFunctionContextSetBackendData(CeedQFunctionContext ctx, void *data) {
 
   @return An error code: 0 - success, otherwise - failure
 
-  @ref User
+  @ref Backend
 **/
 int CeedQFunctionContextGetFieldLabel(CeedQFunctionContext ctx,
                                       const char *field_name,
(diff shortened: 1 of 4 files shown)


interface/ceed-qfunctioncontext.c
```diff
@@ -221,6 +221,36 @@ int CeedQFunctionContextSetBackendData(CeedQFunctionContext ctx, void *data) {
...
+  @brief Get label for a registered QFunctionContext field, or `NULL` if no
+           field has been registered with this `field_name`
+
+  @param[in] ctx           CeedQFunctionContext
+  @param[in] field_name    Name of field to retrieve label
+  @param[out] field_label  Variable to field label
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
```
jrwrigh, line 234: Should these `@refs` be changed to `backend`?

jrwrigh: (including the other moved functions)

jrwrigh (commented):
Should `CeedQFunctionContextSetGeneric` also have it's `@ref` changed to `backend`?

jrwrigh (approved):

PR #949, opened 2022-04-25
Small doc update

jeremylt:
Update for `q_extra` and use the same in both examples


jedbrown (approved):
Also `git grep qextra` for a couple more stragglers.

PR #950, opened 2022-04-25
Use backend functions for SyncArray in CUDA and HIP

nbeams:
Previously, the `SyncArray` interface function never used the backend implementations, and the sync functions inside the cuda-ref and hip-ref backends were only called directly by other backend functions.

(As mentioned in #948)

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -843,6 +843,7 @@ int CeedInit(const char *resource, Ceed *ceed) {
     CEED_FTABLE_ENTRY(CeedVector, SetArray),
     CEED_FTABLE_ENTRY(CeedVector, TakeArray),
     CEED_FTABLE_ENTRY(CeedVector, SetValue),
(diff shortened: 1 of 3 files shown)


interface/ceed.c
```diff
@@ -843,6 +843,7 @@ int CeedInit(const char *resource, Ceed *ceed) {
     CEED_FTABLE_ENTRY(CeedVector, SetArray),
     CEED_FTABLE_ENTRY(CeedVector, TakeArray),
     CEED_FTABLE_ENTRY(CeedVector, SetValue),
+    CEED_FTABLE_ENTRY(CeedVector, SyncArray),
```
nbeams, line 846: `SyncArray` was already part of the `CeedVector_private` struct's list of backend functions, but it had never been added to the table here

jeremylt (approved):
Looks good to me, thanks for making these pieces match up

PR #951, opened 2022-04-27
Use CeedMallocArray for void data pointers in CUDA/HIP QFunctionContext

nbeams:
Avoids `CeedMalloc` (which uses `sizeof`) for `void *` data in QFunctionContext, as already done in the CPU ref backend.


jeremylt (approved):
LGTM. Should be good to merge once CI finishes

nbeams:
Hmm, it's still showing the GitLab as pending (at least as I'm looking at it now), but when I click on "Details," it shows all the GitLab tests have completed successfully?  Does it look that way to you, too?

jeremylt:
Yeah, that is odd

PR #953, opened 2022-04-30
Cleaner density inflow

KennethEJansen:
This pull request adds the option to choose between two variables, density or temperature, set weakly on the inflow.  So far, density  is far more stable than temperature which tends to produce oscillations for both SUPG and Galkerkin for p=1.  Without setting any flags, density will be chosen but with the flag -weakT it will be replaced by weak temperature.  Both options are subsonic inflow boundary conditions meaning one of the two will be applied with the inflow velocity leaving the last variable free to be determined by the solution  as is consistent with 4 incoming characteristics.  Note that no change was made to the application of weak traction at the inlet and this is still also applied at both the inflow and the outflow (where weak pressure is also applied).

### DIFF
--- a/examples/fluids/blasius.yaml
+++ b/examples/fluids/blasius.yaml
@@ -27,9 +27,9 @@ Ctau_t: 1
 #Ctau_v: 36,60,128 is what PHASTA has for p=1,2, 3
 ## linear Settings:
 Ctau_v: 36
-Ctau_C: 0.5
-Ctau_M: 0.5
-Ctau_E: 0.5
+Ctau_C: 0.25
+Ctau_M: 0.25
+Ctau_E: 0.125
 ## Quadratic Settings:
 #Ctau_v: 60
 #Ctau_C: 0.125
--- a/examples/fluids/blasius.yaml
+++ b/examples/fluids/blasius.yaml
@@ -4,7 +4,12 @@ implicit: true
 ts:
   adapt_type: 'none'
   type: 'beuler'
-  dt: 6.4e-5
+  dt: 0.2e-5
+  max_time: 1.0e-3
+  output_freq: 10
+
+#snes_max_it: 4 
+#snes_convergence_test: skip
 
 ## Linear Settings:
 degree: 1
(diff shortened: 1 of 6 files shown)


examples/fluids/blasius.yaml
```diff
@@ -4,7 +4,7 @@ implicit: true
 ts:
   adapt_type: 'none'
   type: 'beuler'
-  dt: 5.e-8
+  dt: 6.4e-5
```
jedbrown, line 7: Can we set a max time? I think this is about 10 velocity transit times through the domain, and the initial SNES residuals are really small by this time.
```suggestion
  dt: 6.4e-5
  max_time: 1e-3
```

examples/fluids/blasius.yaml
```diff
@@ -21,14 +21,15 @@ stab: 'supg'
 Ctau_t: 1
 #Ctau_v: 36,60,128 is what PHASTA has for p=1,2, 3
 Ctau_v: 36
-Ctau_C: 0.125
-Ctau_M: 1.0
-Ctau_E: 0.125
+Ctau_C: 0.5
+Ctau_M: 0.5
+Ctau_E: 0.5
```
jedbrown, line 26: Are the old ones appropriate for quadratic?
```suggestion
Ctau_C: 0.5
Ctau_M: 0.5
Ctau_E: 0.5
## Quadratic settings
#Ctau_v: 60
#Ctau_C: 0.125
#Ctau_M: 0.125
#Ctau_E: 0.125
```

jedbrown (approved):
Looks good. I need to reduce the time step a somewhat when switching to quadratic elements. A couple questions on defaults.

examples/fluids/blasius.yaml
KennethEJansen: Yes.  This is needed.

examples/fluids/blasius.yaml
KennethEJansen: I am running some tests on this as I am unsure.  What I am observing so far is better convergence for large time steps  with 0.5 than reduced values.  I need to dig into the docs to figure out how to force SNES to always take two iterations as it is converging well until SNES decides it needs only on iteration at which time one iteration and then on the next step it fails to converge.  Obviously this is just SNES settings but it needs to be worked out if we want to push large time steps.

jedbrown: I think that's happening once already very close to steady and the issue will go away with analytic Jacobian.

examples/fluids/navierstokes.h
```diff
@@ -299,6 +299,7 @@ struct BlasiusContext_ {
   CeedScalar Uinf;      // !< Velocity at boundary layer edge
   CeedScalar P0;        // !< Pressure at outflow
   CeedScalar theta0;    // !< Temperature at inflow
+  CeedInt weakT;        // !< flag to set Temperature at inflow
```
jrwrigh, line 302: Should these be `bool` instead of `int`?

```suggestion
  bool weakT;           // !< flag to set Temperature at inflow
```
(also includes changing to `PetscOptionsInt` -> `PetscOptionsBool`, and `if(weakT == 0)` to `if(weakT)`)

jedbrown: Done, thanks.

PR #954, opened 2022-05-05
op - include num_elem and num_qpts in view

jeremylt:
This adds a bit to the CeedOperatorView output. Sample output:

```
CeedOperator
  15 elements with 8 quadrature points each
  3 fields
  2 input fields:
    Input field 0:
      Name: "weight"
      No vector
    Input field 1:
      Name: "dx"
      Active vector
  1 output field:
    Output field 0:
      Name: "rho"
      Collocated basis
      Active vector
```

### DIFF
--- a/doc/sphinx/source/releasenotes.md
+++ b/doc/sphinx/source/releasenotes.md
@@ -8,6 +8,8 @@ On this page we provide a summary of the main API changes, new features and exam
 
 ### Interface changes
 
+- Added {c:func}`CeedOperatorSetName` for more readable {c:func}`CeedOperatorView` output.
+
 (v0-10-1)=
 
 ## v0.10.1 (Apr 11, 2022)
--- a/examples/rust/ex1-volume/src/main.rs
+++ b/examples/rust/ex1-volume/src/main.rs
@@ -178,6 +178,7 @@ fn example_1(options: opt::Opt) -> libceed::Result<()> {
     // Operator that build the quadrature data for the mass operator
     let op_build = ceed
         .operator(qf_build, QFunctionOpt::None, QFunctionOpt::None)?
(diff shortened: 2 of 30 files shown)


jedbrown:
What does output look like from a composite operator?

jeremylt:
```
Composite CeedOperator
  SubOperator [0]:
    6 elements with 4 quadrature points each
    3 Fields
    2 Input Fields:
      Input Field [0]:
        Name: "weight"
        No vector
      Input Field [1]:
        Name: "dx"
        Active vector
    1 Output Field:
      Output Field [0]:
        Name: "rho"
        Collocated basis
  SubOperator [1]:
    6 elements with 16 quadrature points each
    3 Fields
    2 Input Fields:
      Input Field [0]:
        Name: "weight"
        No vector
      Input Field [1]:
        Name: "dx"
        Active vector
    1 Output Field:
      Output Field [0]:
        Name: "rho"
        Collocated basis
```

jeremylt:
I'm not sure why the field numbers and sub-operator numbers are in `[]`. That was a strange choice by past me.

jeremylt:
I made the output a bit less odd looking (samples updated above)

jedbrown:
Cool. Should sub-operators print their name (maybe the `source` string) or would that be more confusing/noisy than it's worth?

jeremylt:
The suboperators often all share the same QFunction with different bases, so I think that grabbing the QF name wouldn't help. We could allow the user to set a name for human readability in the output though. I'll add that.

jedbrown:
Volumes and boundaries are usually different and that's informative. Fine to make it optional, named by the user.

jeremylt:
Oh yay, Fortran's crappy string handling bites again

PR #957, opened 2022-05-10
CI Debugging

jeremylt:


.gitlab-ci.yml
```diff
@@ -55,7 +55,7 @@ noether-cpu:
     - export COVERAGE=0
     - cd .. && export NEK5K_VERSION=Nek5000-19.0 && { [[ -d $NEK5K_VERSION ]] || { git clone --depth 1 --branch v19.0 https://github.com/Nek5000/Nek5000.git $NEK5K_VERSION && cd $NEK5K_VERSION/tools && ./maketools genbox genmap reatore2 && cd ../..; }; } && export NEK5K_DIR=$PWD/$NEK5K_VERSION && export PATH=$NEK5K_DIR/bin:$PATH MPI=0 && cd libCEED
     - echo "-------------- Nek5000 -------------" && git -C $NEK5K_DIR describe --tags
-    - make -k -j$NPROC_CPU BACKENDS="$BACKENDS_CPU" JUNIT_BATCH="cpu" junit search=nek
+    - make -k -j$NPROC_CPU BACKENDS="$BACKENDS_CPU" JUNIT_BATCH="cpu" junit search=nek NEK5K_DIR=$NEK5K_DIR
```
jeremylt, line 58: I have no idea why, but `make info` shows the `NEK5K_DIR` variable as unset unless I do this...

jedbrown: I assume you can't reproduce this either, outside of CI?

jeremylt: It works fine for me locally and on LV without this strange hack

jeremylt: Since the Fortran interface and Nek tests are low priority, I'm fine merging this fix to get the CI job running again and then investigating later if we have time. Running the Nek tests from the top level has always been oddly fragile

jedbrown (approved):

PR #958, opened 2022-05-10
examples/fluids cleaning

jedbrown:
This decouples the different problem setups, removing undesirable sharing and replication.


jedbrown:
@jrwrigh Can you give this a peek as to how it may interfere with your STG work? It has no result- or algorithm-changing effects, but I hope will make your development more contained.

jrwrigh:
I don't see any major issues merging the STG branch into this. I'll do a full comb through in a bit, but it looks good!

jedbrown:
I think it's largely mechanical so let's merge this and we can move on.

PR #961, opened 2022-05-11
Fix fluids for CUDA_ERROR_INVALID_PTX and enable lv-cuda

jedbrown:
Thanks @nbeams for help debugging.

### DIFF
--- a/examples/fluids/qfunctions/advection2d.h
+++ b/examples/fluids/qfunctions/advection2d.h
@@ -36,6 +36,8 @@ struct AdvectionContext_ {
   int stabilization; // See StabilizationType: 0=none, 1=SU, 2=SUPG
 };
 
+CEED_QFUNCTION_HELPER CeedScalar Square(CeedScalar x) { return x*x; }
+
 // *****************************************************************************
 // This QFunction sets the initial conditions and the boundary conditions
 //   for two test cases: ROTATION and TRANSLATION
@@ -124,15 +126,15 @@ CEED_QFUNCTION_HELPER int Exact_Advection2d(CeedInt dim, CeedScalar time,
     return 1;
   }
 
-  CeedScalar r = sqrt(pow(x - x0[0], 2) + pow(y - x0[1], 2));
+  CeedScalar r = sqrt(Square(x - x0[0]) + Square(y - x0[1]));
   CeedScalar E = 1 - r/rc;
 
   if (0) { // non-smooth initial conditions
     if (q[4] < E) q[4] = E;
-    r = sqrt(pow(x - x1[0], 2) + pow(y - x1[1], 2));
+    r = sqrt(Square(x - x1[0]) + Square(y - x1[1]));
     if (r <= rc) q[4] = 1;
   }
-  r = sqrt(pow(x - x2[0], 2) + pow(y - x2[1], 2));
+  r = sqrt(Square(x - x2[0]) + Square(y - x2[1]));
   E = (r <= rc) ? .5 + .5*cos(r*M_PI/rc) : 0;
   if (q[4] < E) q[4] = E;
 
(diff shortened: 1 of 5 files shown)


examples/fluids/qfunctions/advection2d.h
```diff
@@ -36,6 +36,8 @@ struct AdvectionContext_ {
   int stabilization; // See StabilizationType: 0=none, 1=SU, 2=SUPG
 };
 
+CEED_QFUNCTION_HELPER CeedScalar Square(CeedScalar x) { return x*x; }
```
jeremylt, line 39: Probably could put this in a common header shared between the QFunctions with an include via rpath

jedbrown: Yeah, I didn't want to create that coupling just yet, but I envision a common header with utilities for scalars and 3x3 matrices.

jeremylt:
There was some error in CI on LV that was before the part of the script touched by this PR, so I'm re-running.

jedbrown:
Was your PETSc built for sm_61? Maybe need to export `CUDA_VISIBLE_DEVICES` when building PETSc, or pass `--with-cuda-arch=75`.

> Kokkos::Cuda::initialize ERROR: running kernels compiled for compute capability 6.1 on device with compute capability 7.5 is not supported by CUDA!

jeremylt:
> Was your PETSc built for sm_61? Maybe need to export `CUDA_VISIBLE_DEVICES` when building PETSc, or pass `--with-cuda-arch=75`.
> 
> > Kokkos::Cuda::initialize ERROR: running kernels compiled for compute capability 6.1 on device with compute capability 7.5 is not supported by CUDA!

huh, I just used the same configure options that previously worked. I'll try adding more flags.

jeremylt:
Fully rebulit PETSc on LV. It works locally, so I'm letting CI try again

jeremylt:
ugh, I'm probably testing on a different device when I log in to try

```
Kokkos::Cuda::initialize ERROR: running kernels compiled for compute capability 6.1 on device with compute capability 7.5 is not supported by CUDA!
```

Still no idea why PETSc seems to be ignoring `--with-cuda-arch=75`. That's super convenient

jeremylt:
Ok, I seem to have rebuilt sucessfully. I'm letting CI try again, though 3/4 of the runs have some random error in a random *t*` series test, so its slow going.

jedbrown:
Maybe we should run a judicious small subset of the `t*` in CI, perhaps with lower `NPROC_GPU`. If fluids and solids pass on CUDA and full pipelines pass on ROCm, chances are the unit tests also pass on CUDA.

jedbrown:
I have this in my `~/.bash_profile`
```bash
export CUDA_VISIBLE_DEVICES=GPU-c4529365-8229-f689-b43d-ccd7f1677079
. /opt/rh/gcc-toolset-11/enable
export PETSC_DIR=$HOME/petsc PETSC_ARCH=ompi-cuda-arch75-g
export PATH=/usr/lib64/openmpi/bin:$PATH
```

jeremylt:
The strange thing is that sometimes `t000` fails and everything else passes, sometimes a random `t3*` or `t5*` fails. I haven't been able to pin down exactly why

---

Yeah, I set the `CUDA_VISIBLE_DEVICES` and that seemed to fix the PETSc build (I think)

jeremylt:
This is the intermittent error
```
libgcov profiling error:/home/gitlab-runner/builds/st7hbLD-/0/libceed/libCEED/build/t004-ceed-f.gcda:overwriting an existing profile data with a different timestamp
```
I really don't see why this should ever happen though. The directory is cleared out before we run junit

jeremylt:
This fix is now working, so I merged. I still need to dig more into why `lcov` isn't working on LV though

PR #962, opened 2022-05-13
examples/fluids: add analytic Jacobian

jedbrown:


jedbrown:
@jrwrigh @KennethEJansen Is there a problem with periodic+wall boundary conditions (but not inflow/outflow) that I can use to test the accuracy of the (frozen tau) SUPG part of this (currently elided). I've been running this and I'm reporting too-accurate Jacobians without tau/SUPG used in the Jacobian action at all.

```console
$ build/fluids-navierstokes -options_file examples/fluids/channel.yaml -snes_monitor -ksp_converged_reason -ts_dt 1e-4 -ts_adapt_monitor
```

@jeremylt This probably isn't worth your time to review just yet.

KennethEJansen:
I am not sure I follow.    The only cases we have without inflow/outflow are the channel and the isentropic vortex so adding walls that leaves just the channel.    

If your issue is that the IC is already completely satisfying the equations, it is perfectly valid to give it a poor IC (e.g.,   linear velocity  based on distance to the wall instead of the exact quadratic profile??).
If that is not appealing, you could put a vortex in he center of domain superposed on the unperturbed velocity.  That shape won't be preserved due to the vorticity in the channel profile but it would at least produce a non-trivial solution.

jedbrown:
Okay, lol. So Channel as currently specified had gravity in the opposite direction of the flow, but the domain was so big (and Reynolds number of 4645) that it would take a very long time for the flow to reverse. But if I increase mu by 100x, I get more interesting steps and the flow reversal as gravity eventually becomes significant. I'll shrink the problem domain to achieve this with more plausible viscosity.

jedbrown:
@nbeams @jeremylt I'm having an issue here with what appears to be incorrect numeric assembly on CUDA and HIP. The application of the Jacobian and diagonal appear to match (on GPU and CPU), but the assembled values `CeedOperatorLinearAssemble` are different (the CPU values are correct; all HIP and CUDA backends are incorrect). These matrices have 5x5 nonsymmetric blocks. Do you have any tips to debug?

jedbrown:
This is working now and ready for @jeremylt to have a look. It's exact for Galerkin (as near I can tell).

## Old approach (now use `-snes_fd_color`)
```console
$ build/fluids-navierstokes -options_file examples/fluids/blasius.yaml -ts_monitor -snes_converged_reason -ts_max_steps 10 -stab none -snes_monitor -pc_type lu -snes_fd_color
 TS dt 2e-06 time 0.
    0 SNES Function norm 2.566151599615e-05
    1 SNES Function norm 7.068374388419e-10
    2 SNES Function norm 1.060081842696e-14
  Nonlinear solve converged due to CONVERGED_FNORM_RELATIVE iterations 2
1 TS dt 2e-06 time 2e-06
    0 SNES Function norm 4.570362775498e-06
    1 SNES Function norm 3.678650480384e-11
    2 SNES Function norm 2.989231890881e-15
  Nonlinear solve converged due to CONVERGED_FNORM_RELATIVE iterations 2
2 TS dt 2e-06 time 4e-06
    0 SNES Function norm 3.920695386552e-06
    1 SNES Function norm 2.390930990566e-11
    2 SNES Function norm 5.145555849539e-15
  Nonlinear solve converged due to CONVERGED_FNORM_RELATIVE iterations 2
```
## New Jacobian
```console
0 TS dt 2e-06 time 0.
    0 SNES Function norm 2.566151599615e-05
    1 SNES Function norm 7.457042659032e-10
    2 SNES Function norm 4.921758938426e-14
  Nonlinear solve converged due to CONVERGED_FNORM_RELATIVE iterations 2
1 TS dt 2e-06 time 2e-06
    0 SNES Function norm 4.570362785126e-06
    1 SNES Function norm 1.385117778293e-10
    2 SNES Function norm 7.730836902993e-15
  Nonlinear solve converged due to CONVERGED_FNORM_RELATIVE iterations 2
2 TS dt 2e-06 time 4e-06
    0 SNES Function norm 3.920695392964e-06
    1 SNES Function norm 9.637990334989e-11
    2 SNES Function norm 4.932095702880e-15
```

jedbrown:
@jedbrown @nbeams The GPU assembly issue remains and I'm not sure the best way to reduce. I suppose a start would be to check whether the qfunction core is assembled correctly, then check the COO values (which I'm observing are incorrect). `MatShell` action and diagonal assembly are correct.

jeremylt:
It looks like the PETSc for libCEED CI doesn't have Kokkos currently

jeremylt:
If this problem has non-symmetric blocks and Ratel has only symmetric, I'd guess the issue is a mixup in the indexing in the GPU kernels. I can look tomorrow and see if its obvious or we need to dig harder.

jedbrown:
Yeah, relative to what's used in Ratel, it's 5x5 and non-symmetric.

nbeams:
I just now saw that I was tagged in this.  Would you mind pointing me to where the problematic operator is built/configured?

jedbrown:
It's the op_ijacobian. There's only one places that calls MatSetValuesCOO

nbeams:
Would the sample command you used above (`build/fluids-navierstokes -options_file examples/fluids/blasius.yaml -ts_monitor -snes_converged_reason -ts_max_steps 10 -stab none -snes_monitor -pc_type lu -snes_fd_color`) be good for testing this issue?

jedbrown:
Yeah, add `-ceed /gpu/cuda` to see it work correctly. Drop `-snes_fd_color` to assemble using libCEED (works with CPU, wrong with GPU).

nbeams:
When I use `-ksp_view_mat`, it appears the matrices I get from CUDA and HIP are not the same... 🤔  
Is that what you've seen, too?

jedbrown:
@nbeams No, it's just that a new feature landed upstream and I haven't written an analytic Jacobian for it (and it'll probably be superseded by a different form with a simpler Jacobian). The assembly issue can be observed fine in this example, which contains only a volume operator. This problem is periodic in the streamline direction, with walls top and bottom, and symmetry/slip in the z direction.
```console
$ build/fluids-navierstokes -options_file examples/fluids/channel.yaml -ts_max_steps 1 -ceed /gpu/cuda/ref -{ts,snes}_monitor
```

nbeams:
I tried making the problem smaller to make it easier to look at and compare the assembled matrix output values.  If I go to `examples/fluids/channel.yaml` and change `dm_plex_box_faces` to `2,1,1`, even the cpu-ref backend fails with the `DIVERGED_NONLINEAR_SOLVE` error (and the output matrix definitely looks wrong, values all ~1e-17 or smaller, even though the values array that I had libCEED print upon its assembly function completion looks more reasonable).

If I change to `1,1,1` -- which I thought was reducing the problem to just one element -- I get a libCEED error:

```
CeedElemRestrictionCreate_Ref(): Restriction offset 6 (14) out of range [0, 12]
```
which is coming from (partial backtrace):

```
#6  0x000000000042758d in CreateRestrictionFromPlex (ceed=ceed@entry=0xd9aa80, dm=<optimized out>, height=height@entry=0, domain_label=domain_label@entry=0x0, 
    value=value@entry=0, elem_restr=elem_restr@entry=0xd93a40) at /home/nbeams/CEED/libCEED/examples/fluids/src/setuplibceed.c:29
#7  0x000000000042791a in GetRestrictionForDomain (ceed=ceed@entry=0xd9aa80, dm=dm@entry=0xdd8af0, height=height@entry=0, domain_label=domain_label@entry=0x0, 
    value=value@entry=0, Q=Q@entry=4, q_data_size=10, elem_restr_q=0xd93a48, elem_restr_x=0xd93a40, elem_restr_qd_i=0xd93a50)
    at /home/nbeams/CEED/libCEED/examples/fluids/src/setuplibceed.c:62
#8  0x0000000000428878 in SetupLibceed (ceed=0xd9aa80, ceed_data=0xd939c0, dm=0xdd8af0, user=0xd932c0, app_ctx=0xd8e4a0, problem=0xd92b30, bc=0xd940d0)
    at /home/nbeams/CEED/libCEED/examples/fluids/src/setuplibceed.c:319
 ```
 
But if I use the cuda-ref backend, I don't see the element restriction error...? (though of course I still have the diverged error).  I don't see how that part of the setup code would be different between backends, so I'm confused.

I don't really understand enough about how the fluids examples work in general, or this example in particular, to know what I can and cannot do in terms of changing the input parameters... What other parameter changes do I need to make in order to cut the problem down as small as possible (1 or 2 elements to start, preferably)?

**Comment by jrwrigh** [comment]
@nbeams The smallest mesh size I was able to get to run successfully was `-dm_plex_box_faces 3,2,1`. 

I'd say the minimum for the problem is `1,2,1` (leaving 1 free node on the centerline of the channel), but I'm getting failure by `DIVERGED_FNORM_NAN`. For `2,2,1`, I get failure by `DIVERGED_LINEAR_SOLVE`. So unfortunately `3,2,1` is the smallest I can get it.

**Comment by nbeams** [comment]
I believe I found the bug; it was predictably silly (and my fault).  See https://github.com/CEED/libCEED/pull/976

When I include that fix in this branch, this command
`./build/fluids-navierstokes -options_file examples/fluids/channel.yaml -ts_max_steps 1 -ceed /gpu/cuda/ref -snes_monitor -snes_test_jacobian`
gives this output:

```
    0 SNES Function norm 1.051470360594e-02 
    ---------- Testing Jacobian -------------
    Run with -snes_test_jacobian_view and optionally -snes_test_jacobian <threshold> to show difference
      of hand-coded and finite difference Jacobian entries greater than <threshold>.
    Testing hand-coded Jacobian, if (for double precision runs) ||J - Jfd||_F/||J||_F is
      O(1.e-8), the hand-coded Jacobian is probably correct.
    ||J - Jfd||_F/||J||_F = 0.00018248, ||J - Jfd||_F = 0.00418989
    1 SNES Function norm 1.012604775618e-05 
    ---------- Testing Jacobian -------------
    ||J - Jfd||_F/||J||_F = 0.000182974, ||J - Jfd||_F = 0.00420118
    2 SNES Function norm 6.079311909419e-10 
    ---------- Testing Jacobian -------------
    ||J - Jfd||_F/||J||_F = 0.000182974, ||J - Jfd||_F = 0.00420118
    3 SNES Function norm 3.825921767798e-13 
```

which is the same as for `/cpu/self/ref`.

**Comment by jedbrown** [comment]
@jrwrigh This is now passing on GPUs. Could you give it a brief review so you understand what it's doing? Then we can change outflow handling as we discussed (to avoid referencing the Blasius profile) and add the viscous inflow flux.

File: examples/fluids/navierstokes.c [hunk 1]
```diff
@@ -35,8 +35,8 @@
...
-//TESTARGS(name="blasius_STG") -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG.bin
-//TESTARGS(name="blasius_STG_weakT") -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG_weakT.bin -weakT
+//TESTARGS(name="blasius_STG") -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG.bin -snes_fd_color
+//TESTARGS(name="blasius_STG_weakT") -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG_weakT.bin -weakT -snes_fd_color
```
jedbrown, line 39: @jrwrigh We'll be able to remove these `-snes_fd_color` once we differentiate the STG inflows. But let's use the in-plane gradients before doing that.

examples/fluids/qfunctions/newtonian.h
```diff
@@ -803,25 +736,164 @@ CEED_QFUNCTION(IFunction_Newtonian)(void *ctx, CeedInt Q,
...
+                                    q_data[5][i],
+                                    q_data[6][i]},
+                                   {q_data[7][i],
+                                    q_data[8][i],
+                                    q_data[9][i]}
+                                  };
+    // *INDENT-ON*
+
+    CeedScalar U[5], kmstress[6], Tau_d[3] __attribute((unused));
+    for (int j=0; j<5; j++) U[j] = jac_data[j][i];
```
jrwrigh, line 806: Should these be `CeedInt`?

examples/fluids/qfunctions/newtonian.h
```diff
@@ -803,25 +736,164 @@ CEED_QFUNCTION(IFunction_Newtonian)(void *ctx, CeedInt Q,
...
+                                    dXdx[j][1] * dFlux[k][1] +
+                                    dXdx[j][2] * dFlux[k][2]);
+      }
+    }
+
+    const CeedScalar dbody_force[5] = {0, ds.U.density *g[0], ds.U.density *g[1], ds.U.density *g[2], 0};
+    for (int j=0; j<5; j++)
+      v[j][i] = wdetJ * (context->ijacobian_time_shift * dU[j] - dbody_force[j]);
+
+    if (1) {
```
jrwrigh, line 856: Is this hardcoded `if` necessary? Maybe it should be this:
```suggestion
    if (context->stabilization != STAB_NONE) {
```

jrwrigh (approved):
I think I follow the general gist of it. Few minor suggestions, but looks good to me!

jrwrigh:
> Then we can change outflow handling as we discussed (to avoid referencing the Blasius profile) and add the viscous inflow flux.

Yeah, I started to work on that a bit yesterday.

examples/fluids/problems/channel.c
```diff
@@ -40,13 +40,16 @@ PetscErrorCode NS_CHANNEL(ProblemData *problem, DM dm,
...
   CeedScalar P0     = 1.e5; // Pa
+  PetscReal body_force_scale = 1.;
   PetscOptionsBegin(comm, NULL, "Options for CHANNEL problem", NULL);
   ierr = PetscOptionsScalar("-umax", "Centerline velocity of the Channel",
                             NULL, umax, &umax, NULL); CHKERRQ(ierr);
   ierr = PetscOptionsScalar("-theta0", "Wall temperature",
                             NULL, theta0, &theta0, NULL); CHKERRQ(ierr);
   ierr = PetscOptionsScalar("-P0", "Pressure at outflow",
                             NULL, P0, &P0, NULL); CHKERRQ(ierr);
+  ierr = PetscOptionsReal("-body_force_scale", "Multiplier for body force",
```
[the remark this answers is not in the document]
jedbrown: Done

examples/fluids/problems/newtonian.c
```diff
@@ -116,6 +172,9 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
   ierr = PetscOptionsBool("-implicit", "Use implicit (IFunction) formulation",
                           NULL, implicit=PETSC_FALSE, &implicit, NULL);
   CHKERRQ(ierr);
+  ierr = PetscOptionsBool("-newtonian_unit_tests", "Run Newtonian unit tests",
```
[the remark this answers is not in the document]
jedbrown: done

examples/fluids/problems/newtonian.c
```diff
@@ -12,13 +12,66 @@
...
+  StatePrimitive eY; // relative error
+  eY.pressure = (aY.pressure - bY.pressure) / sY.pressure;
+  PetscScalar u = sqrt(Square(sY.velocity[0]) + Square(sY.velocity[1]) + Square(
+                         sY.velocity[2]));
+  for (int j=0; j<3; j++) eY.velocity[j] = (aY.velocity[j] - bY.velocity[j]) / u;
+  eY.temperature = (aY.temperature - bY.temperature) / sY.temperature;
+  if (fabs(eY.pressure) > rtol_pressure) printf("%s: pressure error %g\n", name,
+        eY.pressure);
+  for (int j=0; j<3;
+       j++) if (fabs(eY.velocity[j]) > rtol_velocity)
```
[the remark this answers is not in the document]
jedbrown: done

examples/fluids/qfunctions/newtonian.h
jedbrown: It's inactive for Galerkin because the computed tau is zero. I kept the `if` block for development convenience and because I think that block of code can be refactored to be a bit cleaner and more efficient. I'd prefer that control flow continue to pass through this even for `-stab none` unless we want to optimize for Galerkin (in which case we'd want to create a separate qfunction that passes less data).

examples/fluids/problems/newtonian.c
```diff
@@ -12,13 +12,66 @@
...
+      printf("%s: velocity[%d] error %g\n", name, j, eY.velocity[j]);
+  if (fabs(eY.temperature) > rtol_temperature)
+    printf("%s: temperature error %g\n", name, eY.temperature);
+  PetscFunctionReturn(0);
+}
+
+static PetscErrorCode UnitTests_Newtonian(User user,
```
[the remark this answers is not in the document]
jedbrown: Our unit tests are already getting expensive so I'm not really inclined at this time. If we transfer this code to Ratel (or make it easier to subset), then it'd make sense to add it. From the perspective of the repository we're working in, `examples/fluids` exists to test libCEED, and thus runs everything with every backend. This code is not at all backend-dependent, but I didn't want to delete this scaffolding I wrote to check correctness of that helper function.

examples/fluids/qfunctions/newtonian.h
jedbrown: Jeremy has sometimes made changes to that effect. It needs to be `CeedInt` when the loop bounds are `CeedInt`, but here it's just the magic numbers 5 and 3 so anything is okay. I find `int` visually quietest in this context, which is good.

jrwrigh: Gotcha. Definitely agree on `int` being more visually quiet.

PR #963, opened 2022-05-17
Add test for non-symmetric assembly

jeremylt:
@jedbrown @nbeams 

I added t566, though it looks like this test shows the CPU and GPU failing in the same way?

This should help us track down the assembly issue for the fluids PR.

### DIFF
--- a/tests/t566-operator.c
+++ b/tests/t566-operator.c
@@ -27,10 +27,10 @@ int main(int argc, char **argv) {
   CeedInit(argv[1], &ceed);
 
   // DoF Coordinates
-  for (CeedInt i=0; i<n_x*2+1; i++)
-    for (CeedInt j=0; j<n_y*2+1; j++) {
-      x[i+j*(n_x*2+1)+0*num_dofs] = (CeedScalar) i / (2*n_x);
-      x[i+j*(n_x*2+1)+1*num_dofs] = (CeedScalar) j / (2*n_y);
+  for (CeedInt i=0; i<n_x*(P-1)+1; i++)
+    for (CeedInt j=0; j<n_y*(P-1)+1; j++) {
+      x[i+j*(n_x*2+1)+0*num_dofs] = (CeedScalar) i / (n_x * (P-1));
+      x[i+j*(n_x*2+1)+1*num_dofs] = (CeedScalar) j / (n_y * (P-1));
     }
   CeedVectorCreate(ceed, dim*num_dofs, &X);
   CeedVectorSetArray(X, CEED_MEM_HOST, CEED_USE_POINTER, x);
@@ -43,10 +43,10 @@ int main(int argc, char **argv) {
     CeedInt col, row, offset;
     col = i % n_x;
     row = i / n_x;
-    offset = col*(P-1) + row*(n_x*2+1)*(P-1);
+    offset = col*(P-1) + row*(n_x*(P-1)+1)*(P-1);
     for (CeedInt j=0; j<P; j++)
       for (CeedInt k=0; k<P; k++)
-        ind_x[P*(P*i+k)+j] = offset + k*(n_x*2+1) + j;
+        ind_x[P*(P*i+k)+j] = offset + k*P + j;
   }
 
   // Restrictions
@@ -55,7 +55,7 @@ int main(int argc, char **argv) {
   CeedElemRestrictionCreate(ceed, num_elem, P*P, num_comp, num_dofs,
                             num_comp*num_dofs,
                             CEED_MEM_HOST, CEED_USE_POINTER, ind_x, &elem_restr_u);
-  CeedInt strides_qd[3] = {1, Q*Q, Q*Q};
+  CeedInt strides_qd[3] = {1, Q*Q*num_elem, Q*Q};
   CeedElemRestrictionCreateStrided(ceed, num_elem, Q*Q, 1, num_qpts, strides_qd,
                                    &elem_restr_qd_i);
 
--- a/tests/t566-operator.c
+++ b/tests/t566-operator.c
@@ -17,7 +17,7 @@ int main(int argc, char **argv) {
   CeedInt P = 3, Q = 3, dim = 2, num_comp = 2;
   CeedInt n_x = 1, n_y = 1;
   CeedInt num_elem = n_x * n_y;
-  CeedInt num_dofs = (n_x*2+1)*(n_y*2+1), num_qpts = num_elem*Q*Q;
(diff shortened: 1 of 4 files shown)


tests/t566-operator.c
```diff
@@ -0,0 +1,184 @@
...
+  CeedElemRestriction elem_restr_x, elem_restr_u,
+                      elem_restr_qd_i;
+  CeedBasis basis_x, basis_u;
+  CeedQFunction qf_setup, qf_mass;
+  CeedOperator op_setup, op_mass;
+  CeedVector q_data, X, U, V;
+  CeedInt P = 3, Q = 3, dim = 2, num_comp = 2;
+  CeedInt n_x = 1, n_y = 1;
+  CeedInt num_elem = n_x * n_y;
+  CeedInt num_dofs = (n_x*2+1)*(n_y*2+1), num_qpts = num_elem*Q*Q;
```
jedbrown, line 20: I think this is intended. I'd like to be able to simplify as much as possible.
```suggestion
  CeedInt num_dofs = (n_x*(P-1)+1)*(n_y*(P-1)+1), num_qpts = num_elem*Q*Q;
```

tests/t566-operator.c
```diff
@@ -0,0 +1,184 @@
...
+        ind_x[P*(P*i+k)+j] = offset + k*(n_x*2+1) + j;
+  }
+
+  // Restrictions
+  CeedElemRestrictionCreate(ceed, num_elem, P*P, dim, num_dofs, dim*num_dofs,
+                            CEED_MEM_HOST, CEED_USE_POINTER, ind_x, &elem_restr_x);
+  CeedElemRestrictionCreate(ceed, num_elem, P*P, num_comp, num_dofs,
+                            num_comp*num_dofs,
+                            CEED_MEM_HOST, CEED_USE_POINTER, ind_x, &elem_restr_u);
+  CeedInt strides_qd[3] = {1, Q*Q, Q*Q};
```
jedbrown, line 58: So these strides are `[node, component, element]`, but given that we have two components, how can the stride between elements equal the stride between components? I guess it's benign when there's only one element, but we should do it right.

tests/t566-operator.c
```diff
@@ -0,0 +1,184 @@
...
+  CeedElemRestrictionCreateStrided(ceed, num_elem, Q*Q, 1, num_qpts, strides_qd,
+                                   &elem_restr_qd_i);
+
+  // Bases
+  CeedBasisCreateTensorH1Lagrange(ceed, dim, dim, P, Q, CEED_GAUSS, &basis_x);
+  CeedBasisCreateTensorH1Lagrange(ceed, dim, num_comp, P, Q, CEED_GAUSS,
```
jedbrown, line 64: It'll be easier to debug when changing these to `CEED_GAUSS_LOBATTO`, which produces many fewer points in my test.

tests/t566-operator.c
jedbrown: BTW, I think "dofs" used here to mean number of nodes (not `num_dofs = num_nodes * num_comp`) is confusing.

jedbrown (commented):
Do we have a test yet that nonsymmetric QFunction assembly is correct?

tests/t566-operator.c
jeremylt: Its only one component in the qdata, which is why it was written this way originally in a previous test file, I'd guess.

nbeams:
When I try the latest version locally, it passes for both cpu and hip-ref.  Is that what you see too?

jedbrown:
Yeah, it passes for me now, and I think the indexing in the test is now correct.

jedbrown:
I haven't tried to converge now on where between this test and the fluids problem we're getting incorrect results.

jeremylt:
I was thinking the next step would be t567, this test based upon Poisson instead of the Mass operator

nbeams:
Hmm, t567 appears to pass for cuda-ref (GTX 1060) but fail for hip-ref (Radeon VII).  Odd.  Though t567 passes for hip-ref if I change to one component instead of two...?

nbeams:
FWIW, the failure of the hip test seems to be with small differences in every value, ~1e-6.  Not crazy weird.  Same if I up `num_comp` to 3. 

The serial CPU backends seem okay, though blocked fails (and seems completely off, unlike the hip-ref small diffs).  This is true for `num_comp` 1 as well, where hip-ref passes.

tests/t567-operator.h
```diff
@@ -0,0 +1,73 @@
...
+                                        {q_data[2][i],
+                                         q_data[1][i]}
+                                       };
+    // *INDENT-ON*
+
+    // Apply Poisson operator
+    // j = direction of vg
+    for (CeedInt j=0; j<dim; j++)
+      vg[j][i] = (ug[0][i] * dXdxdXdxT[0][j] * scale[0][i] +
+                  ug[1][i] * dXdxdXdxT[1][j] * scale[1][i]);
```
nbeams, line 69: I think at least some of the current issues with this test are from this line -- `i` is the quadrature point loop, so I don't think it's what you intended for indexing the scale values, especially since Q is 3 and the size of scale is 2x2.

Switching `scale[*][i]` to `scale[*][j]` here fixes all the issues I was seeing locally (avx/blocked, hip-ref) but not sure if that's what you intended, either.  With the modified QFunction, I confirmed that the `values` array from the assembly was the same for cpu and hip-ref.

jeremylt: Good catch. I'm not if this is good or bad news for the fluids assembly bug.

nbeams: Hmm, yeah, though I don't know why, since they use the same (ref) kernels for assembly...

jeremylt: It seems to be the same error between hip-gen and cuda-gen.

jeremylt: If I'm understanding the output correctly, the `assembled_true_value` is the one that is wrong.

nbeams: Yeah, I've checked, and the output from the setup operator is the same, but then the outputs from the apply operator are wrong for cuda-gen, but not cuda-ref. ??  Been wasting a bunch of time trying to track down the reason, but no real good theories yet.

nbeams: Okay I think I found the issue.  The QFunction for the application in t567 doesn't include the multiple components anywhere, unlike t566-operator.h

jedbrown: Is there something like valgrind client requests? Or we could make libCEED device malloc initialize with NaN.

jeremylt:
Can someone approve this so we can merge to help debug the new assembly issue?

PR #964, opened 2022-05-17
examples/fluids: Create mesh from file distribution

jrwrigh:
For the flat plate boundary layer problems, allow for the wall-normal node distribution to be defined by a file of node locations.

Note this is currently feature complete, but awaits #868 to be merged. (though it could technically be merged before as well).

### DIFF
--- a/examples/fluids/README.md
+++ b/examples/fluids/README.md
@@ -656,22 +656,22 @@ addition to the Newtonian Ideal Gas options:
   - `1.01E5`
   - `Pa`
 
-* - `-refine_height`
-  - Height at which `-Ndelta` number of elements should refined into
+* - `-platemesh_refine_height`
+  - Height at which `-platemesh_Ndelta` number of elements should refined into
   - `5.9E-4`
   - `m`
 
-* - `-Ndelta`
-  - Number of elements to keep below `-refine_height`
+* - `-platemesh_Ndelta`
+  - Number of elements to keep below `-platemesh_refine_height`
   - `45`
   -
 
-* - `-growth`
+* - `-platemesh_growth`
   - Growth rate of the elements in the refinement region
   - `1.08`
   -
 
-* - `-top_angle`
+* - `-platemesh_top_angle`
   - Downward angle of the top face of the domain. This face serves as an outlet.
   - `5`
   - `degrees`
@@ -679,6 +679,11 @@ addition to the Newtonian Ideal Gas options:
 * - `-stg_use`
   - Whether to use stg for the inflow conditions
   - `false`
+  -
+
+* - `-platemesh_y_node_locs_path`
+  - Path to file with y node locations. If empty, will use mesh warping instead.
+  - `""`
   -
 :::
 
--- a/examples/fluids/index.md
+++ b/examples/fluids/index.md
@@ -618,3 +618,29 @@ numerous terms in the STG formulation.
 | $\{\kappa^n\}_{n=1}^N$ | k^n  | No           | Yes      |
 | $h_i$           | h_i    | Yes          | No   |
 | $d_w$           | d_w    | Yes          | No   |
+
+### Meshing
+
+The flat plate boundary layer example has custom meshing features to better
+resolve the flow. One of those is tilting the top of the domain, allowing for
+it to be a outflow boundary condition. The angle of this tilt is controled by
+`-platemesh_top_angle`
+
+The primary meshing feature is the ability to grade the mesh, providing better
+resolution near the wall. There are two methods to do this; algorithmically, or
+specifying the node locations via a file. Algorithmically, a base node
+distribution is defined at the inlet (assumed to be $\min(x)$) and then
+linearly stretched/squeezed to match the slanted top boundary condition. Nodes
+are placed such that `-platemesh_Ndelta` elements are within
+`-platemesh_refine_height` of the wall. They are placed such that the element
+height matches a geometric growth ratio defined by `-platemesh_growth`. The
+remaining elements are then distributed from `-platemesh_refine_height` to the
+top of the domain linearly in logarithmic space.
+
+Alternatively, a file may be specified containing the locations of each node.
+The file should be newline delimited, with the first line specifying the number
+of points and the rest being the locations of the nodes. The node locations
+used exactly at the inlet (assumed to be $\min(x)$) and linearly
+stretched/squeezed to match the slanted top boundary condition. The file is
+specified via `-platemesh_y_node_locs_path`. If this flag is given an empty
+string, then the algorithmic approach will be performed.
--- a/examples/fluids/problems/blasius.c
+++ b/examples/fluids/problems/blasius.c
@@ -21,8 +21,9 @@
  * The top surface is also angled downwards, so that it may be used as an
  * outflow. It's angle is controlled by `top_angle` (in units of degrees).
  */
-PetscErrorCode modifyMesh(DM dm, PetscInt dim, PetscReal growth, PetscInt N,
-                          PetscReal refine_height, PetscReal top_angle) {
+static PetscErrorCode ModifyMesh(DM dm, PetscInt dim, PetscReal growth,
+                                 PetscInt N, PetscReal refine_height,
+                                 PetscReal top_angle) {
 
   PetscInt ierr, narr, ncoords;
   PetscReal domain_min[3], domain_max[3], domain_size[3];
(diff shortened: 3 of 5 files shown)


examples/fluids/problems/blasius.c
```diff
@@ -12,6 +12,41 @@
...
+  char line[char_array_len];
+  char **array;
+  PetscReal *node_locs;
+  PetscFunctionBeginUser;
+
+  ierr = PetscFOpen(comm, path, "r", &fp); CHKERRQ(ierr);
+  ierr = PetscSynchronizedFGets(comm, fp, char_array_len, line); CHKERRQ(ierr);
+  ierr = PetscStrToArray(line, ' ', &ndims, &array); CHKERRQ(ierr);
+
+  for (PetscInt i=0; i<ndims; i++)  dims[i] = atoi(array[i]);
```
jedbrown, line 31: Maybe cleaner (and faster, but not important because these files are small) alternative is to use https://petsc.org/main/docs/manualpages/Viewer/PetscViewerASCIIRead/ and broadcast the result.

examples/fluids/problems/blasius.c
```diff
@@ -20,9 +55,13 @@
  *
  * The top surface is also angled downwards, so that it may be used as an
  * outflow. It's angle is controlled by `top_angle` (in units of degrees).
+ *
+ * If `node_locs` is not NULL, then the nodes will be placed at `node_locs`
+ * locations.
  */
-PetscErrorCode modifyMesh(DM dm, PetscInt dim, PetscReal growth, PetscInt N,
-                          PetscReal refine_height, PetscReal top_angle) {
+PetscErrorCode modifyMesh(MPI_Comm comm, DM dm, PetscInt dim, PetscReal growth,
```
jedbrown, line 62:
```suggestion
static PetscErrorCode modifyMesh(MPI_Comm comm, DM dm, PetscInt dim, PetscReal growth,
```

examples/fluids/index.md
```diff
@@ -618,3 +618,28 @@ numerous terms in the STG formulation.
...
+defined by `-growth`. The remaining elements are then distributed from
+`-refine_height` to the top of the domain linearly in logarithmic space.
+
+Alternatively, a file may be specified containing the locations of each node.
+The file should be newline delimited, with the first line specifying the number
+of points and the rest being the locations of the nodes. The node locations
+used exactly at the inlet (assumed to be $\min(x)$) and linearly
+stretched/squeezed to match the slanted top boundary condition. The file is
+specified via `-y_node_locs_path`. If this flag is given an empty string, then
+the algorithmic approach will be performed.
```
jedbrown, line 645: Does the length of this file need to be manually kept in sync with the `-dm_plex_box_faces`? If we want it to decouple (seem like a better user interface), then I'd think we should interpolate in this grading file so the mesh covers the domain implied by the grading.

examples/fluids/index.md
```diff
@@ -618,3 +618,28 @@ numerous terms in the STG formulation.
...
+resolve the flow. One of those is tilting the top of the domain, allowing for
+it to be a outflow boundary condition. The angle of this tilt is controled by
+`-top_angle`
+
+The primary meshing feature is the ability to grade the mesh, providing better
+resolution near the wall. There are two methods to do this; algorithmically, or
+specifying the node locations via a file. Algorithmically, a base node
+distribution is defined at the inlet (assumed to be $\min(x)$) and then linearly
+stretched/squeezed to match the slanted top boundary condition. Nodes are
+placed such that `-Ndelta` elements are within `-refine-height` of the wall.
```
jedbrown, line 634:
```suggestion
placed such that `-Ndelta` elements are within `-refine_height` of the wall.
```

jedbrown (commented):
Can we namespace the mesh spacing options so they at least sort together? This could be `-plate_bl_growth` or `-bl_y_growth` or something, but I think we sow confusion if these don't group together, especially with names like `-Ndelta` that have no obvious meaning on their own.

Looks good otherwise. Convention has been to write `FunctionName` rather than `functionName`, so maybe stick to that convention.

examples/fluids/problems/blasius.c
jrwrigh: I'm all for giving it a shot, but I'm not sure how to do it. Documentation (that I can find) doesn't describe the format that the viewer would expect, how to handle multiple columns, etc. Grepping through the examples doesn't show much either. Do you have an example I could reference?

examples/fluids/index.md
jrwrigh: Not entirely, no. If the file has more points than the mesh ( so `-dm_plex_box_faces + 1` ), then it will print out a warning, but just use the first `-dm_plex_box_faces + 1` points. If the file has less, then it will error out. 

Doing an interpolation would be interesting, but the primary intention of this is to exactly match an external mesh. If we do the decouple route, I think it should still default to erroring out/printing a warning if the file and `-dm_plex_box_faces` are not synced.

examples/fluids/problems/blasius.c
jedbrown: Maybe check `PetscViewerRead` calls in `plexgmsh.c`, which reads single and multi-column data (ASCII or Binary), a mix of integers and floating point.

examples/fluids/index.md
jedbrown: Yeah, I'm thinking of switching from linear to quadratic elements without needing to re-prep the inputs. But I'm okay with a clear error message for now.

jedbrown: Let me know when this is ready for review and merging.

jrwrigh: Yeah, it's ready for now if you're ok with the current file reading implementation. I can always switch to the Viewer implementation later, and I think strong STG is a higher priority at the moment.

jedbrown: Could you just address the naming of options for this mesh transform component (in the code, yaml, and docs)? Those will end up in input files and it'll be more disruptive to change them later, especially if they aren't grouped now.

jedbrown:
```suggestion
Alternatively, a file may be specified containing the locations of each node.
The file should be newline delimited, with the first line specifying the number
of points and the rest being the locations of the nodes. The node locations
used exactly at the inlet (assumed to be $\min(x)$) and linearly
stretched/squeezed to match the slanted top boundary condition. The file is
specified via `-platemesh_y_node_locs_path`. If this flag is given an empty string, then
the algorithmic approach will be performed.
```

examples/fluids/README.md
```diff
@@ -680,6 +680,11 @@ addition to the Newtonian Ideal Gas options:
   - Whether to use stg for the inflow conditions
   - `false`
   -
+
+* - `-y_node_locs_path`
```
jedbrown, line 684: And the names above
```suggestion
* - `-platemesh_y_node_locs_path`
```

examples/fluids/index.md
```diff
@@ -618,3 +618,28 @@ numerous terms in the STG formulation.
 | $\{\kappa^n\}_{n=1}^N$ | k^n  | No           | Yes      |
 | $h_i$           | h_i    | Yes          | No   |
 | $d_w$           | d_w    | Yes          | No   |
+
+### Meshing
+
+The flat plate boundary layer example has custom meshing features to better
+resolve the flow. One of those is tilting the top of the domain, allowing for
+it to be a outflow boundary condition. The angle of this tilt is controled by
+`-top_angle`
```
jedbrown, line 627:
```suggestion
`-platemesh_top_angle`
```

examples/fluids/index.md
```diff
@@ -618,3 +618,28 @@ numerous terms in the STG formulation.
...
+
+The primary meshing feature is the ability to grade the mesh, providing better
+resolution near the wall. There are two methods to do this; algorithmically, or
+specifying the node locations via a file. Algorithmically, a base node
+distribution is defined at the inlet (assumed to be $\min(x)$) and then linearly
+stretched/squeezed to match the slanted top boundary condition. Nodes are
+placed such that `-Ndelta` elements are within `-refine_height` of the wall.
```
jedbrown, line 634:
```suggestion
placed such that `-platemesh_Ndelta` elements are within `-platemesh_refine_height` of the wall.
```

examples/fluids/index.md
```diff
@@ -618,3 +618,28 @@ numerous terms in the STG formulation.
...
+They are placed such that the element height matches a geometric growth ratio
+defined by `-growth`. The remaining elements are then distributed from
```
jedbrown, line 636:
```suggestion
defined by `-platemesh_growth`. The remaining elements are then distributed from
```

examples/fluids/index.md
```diff
@@ -618,3 +618,28 @@ numerous terms in the STG formulation.
...
+`-refine_height` to the top of the domain linearly in logarithmic space.
```
jedbrown, line 637:
```suggestion
`-platemesh_refine_height` to the top of the domain linearly in logarithmic space.
```

jedbrown (commented):
You can use grep to check that all instances of these variables that should be have been updated.

examples/fluids/README.md
jrwrigh: Whoops, forgot about the documentation.

jrwrigh:
Fixed up the documentation and updated `blasius.yaml` as well. Let me know if you see anything else.

jedbrown (approved):
lv testing is being flakey. I'm ordering a CUDA GPU that will fit in Noether so we can hopefully stop bleeding time into that older GPU (whose weird shaped fan prevents it form fitting in Noether) having erratic false positives.

PR #967, opened 2022-05-24
Fluids: Strongly enforced STG

jrwrigh:
Implement the current STG boundary condition, but with strong enforcement rather than the current weak enforcement.

Also adds `problem->bc_from_ics` boolean to determine whether to set the boundary conditions to fixed values based on the initial condition. This needs to be `false` for strong STG, but breaks tests (and is not as performant).

TODO:
- [x] (Further) Validate inflow against PHASTA
- [x] Determine how to implement in code
	- Should I refactor the essential boundary condition creation to allow for easier expansion in the future (like strong implementation of wall temperature, etc.)?
- [x] Run laminar simulation with strong inflow boundary condition

Note: Currently branched off of `jrwrigh/mesh_from_file`, which is in #964 . Mesh used to validate against PHASTA.


jrwrigh:
The `adv2d_rotation_explicit_strong` test is failing on my laptop. I'll need to investigate it to figure out what's the issue there (failed error norm, so I'm effecting the solution somehow).

jedbrown (approved):

PR #969, opened 2022-05-25
ElemRestriction and Basis input checking

### LINKED ISSUE #966: ElemRestriction input verification
We should add improved checks for input parameters for ElemRestrictions - only `num_elem` should be able to be 0.

jeremylt:
Fixes #966


jedbrown (approved):
Looks reasonable. I think this would have caught the case I was debugging the other day.

PR #973, opened 2022-05-27
Use RTC for MAGMA tensor basis kernels and element restrictions

nbeams:
This PR shifts the MAGMA backend to use libCEED's RTC framework to compile tensor basis kernels and element restriction kernels.  It greatly reduces the time to compile the library (no template instantiations) and also lifts the restrictions on the tests/examples which previously couldn't use the MAGMA backend due to parameters not included in the instantiation ranges.  It should also help with mixed-precision usage in the future, assuming we can leverage RTC to our advantage across the GPU backends to avoid needing to compile kernels for many combinations of precisions.

Note that this should be ready to review, but *I'd prefer to not merge until* @abdelfattah83 *has had time to review/test.*


include/ceed/jit-tools.h
```diff
@@ -21,6 +21,7 @@
 
 CEED_EXTERN int CeedCheckFilePath(Ceed ceed, const char *source_file_path, bool *is_valid);
 CEED_EXTERN int CeedLoadSourceToBuffer(Ceed ceed, const char *source_file_path, char **buffer);
+CEED_EXTERN int CeedLoadSourceToInitializedBuffer(Ceed ceed, const char *source_file_path, char **buffer);
```
nbeams, line 24: The main change outside of the MAGMA backend is that I moved this function to the `jit-tools` header in order to easily load several kernel files before compiling.  Is this acceptable?  If not, should we move to a different header?  I'd rather not have to duplicate some common code in order to have everything in one file for each dimension's basis kernels, for example.

jedbrown:
This sounds great! We'll be running tets a lot in the coming months so this will be a great feature to have better supported. I'll be somewhat offline for a few days, but will test and review sometime next week.

jedbrown:
Sometime between #918 and now, we stopped testing MAGMA in noether-rocm.
```
MAGMA_DIR     = /projects/hipMAGMA [not found]
```
Can you or @jeremylt figure out what's up so that this actually gets tested?

jedbrown:
@abdelfattah83 :wave: When would you have a chance to review this?

nbeams:
> Sometime between #918 and now, we stopped testing MAGMA in noether-rocm.
> 
> ```
> MAGMA_DIR     = /projects/hipMAGMA [not found]
> ```
> 
> Can you or @jeremylt figure out what's up so that this actually gets tested?

I don't think I have the permissions to edit/build in /projects/hipMAGMA.  Not sure why it's not built currently.  I didn't think we switched to a new build somewhere else?

There will be an issue building with libCEED due to the error w/ defining `half` for HIP:

In `include/magma_types.h`, line 198,
`typedef __half            magmaHalf;` [in the HIP part of the header]
this will cause failures when compiling for C instead of C++, because `__half` in HIP is a struct only defined when `if defined(__cplusplus)` is true 

What I have tried locally that seemed to work was replacing `__half` with `hipblasHalf` (which is actually currently `uint16_t`, see https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/develop/library/include/hipblas.h).  This was Ahmad's idea; it may end up being an official change in MAGMA.  But it should definitely work for getting libCEED built while including the `magma_types` header (and thus being able to build with the MAGMA backend for HIP).

nbeams:
The build in `/projects/hipMAGMA` should be available and working now.

jeremylt:
I'm getting
```
In file included from /projects/hipMAGMA/include/magmablas.h:19,
                 from /projects/hipMAGMA/include/magma_v2.h:22,
                 from /var/lib/gitlab-runner/builds/N8GCsqus/0/libceed/libCEED/backends/magma/ceed-magma.h:14,
                 from /var/lib/gitlab-runner/builds/N8GCsqus/0/libceed/libCEED/backends/magma/ceed-magma-basis.c:10:
/projects/hipMAGMA/include/magmablas_h.h:76:1: error: expected identifier or ‘(’ before ‘}’ token
   76 | }
      | ^
```

nbeams:
Hmm, I saw that error too, but it should be fixed now.  I moved the brace up inside the `#endif` -- this was fixed in [this commit](https://bitbucket.org/icl/magma/commits/207035db5e868a2db017ed7a6325dd5540b7b477), just after the release, so the repository in `/projects/hipMAGMA` didn't have it.  But again, I fixed it back when I posted my earlier comment -- when was this run that failed started?

jeremylt:
This run was 25 min ago

nbeams:
> This run was 25 min ago

Hmmmm... ??

I am looking at the file right now, the `}` is on line 74 instead of 76.  The timestamp shows it was last edited at 13:42.  Since we don't install the library to another folder, I'm not sure what else there is to do to make sure the CI uses the latest version of the files?

When I build locally with `/projects/hipMAGMA` as the `MAGMA_DIR`, it's fine.

jeremylt:
Now we're getting a different error
```
Can't exec "/opt/rocm/bin/rocm_agent_enumerator": No such file or directory at /opt/rocm/bin//hipcc.pl line 609.
Use of uninitialized value $targetsStr in substitution (s///) at /opt/rocm/bin//hipcc.pl line 610.
Use of uninitialized value $targetsStr in split at /opt/rocm/bin//hipcc.pl line 616.
Can't exec "/opt/rocm/bin/rocm_agent_enumerator": No such file or directory at /opt/rocm/bin//hipcc.pl line 609.
Use of uninitialized value $targetsStr in substitution (s///) at /opt/rocm/bin//hipcc.pl line 610.
Use of uninitialized value $targetsStr in split at /opt/rocm/bin//hipcc.pl line 616.
```
@jedbrown, something going on with Noether?

jedbrown:
Huh, I did just upgrade rocminfo, which provides `rocm_agent_enumerator`. I can either revert or do regular upgrades and hopefully finishing the job will fix it.

jedbrown:
I think we're operational again. I did make this change (rather than recompiling with the new 5.1.3). Seems to work.
```console
# chrpath -r /opt/rocm-5.1.3/hip/lib:/opt/rocm-5.1.3/lib:/opt/rocm/lib:/opt/rocm-5.1.3/llvm/bin/../lib /projects/hipMAGMA/lib/libmagma.so
/projects/hipMAGMA/lib/libmagma.so: RUNPATH=/opt/rocm-5.1.1/hip/lib:/opt/rocm-5.1.1/lib:/opt/rocm/lib:/opt/rocm-5.1.1/llvm/bin/../lib
/projects/hipMAGMA/lib/libmagma.so: new RUNPATH: /opt/rocm-5.1.3/hip/lib:/opt/rocm-5.1.3/lib:/opt/rocm/lib:/opt/rocm-5.1.3/llvm/bin/../lib
```

backends/magma/ceed-magma-basis.c
```diff
@@ -8,6 +8,15 @@
 #include <ceed/ceed.h>
 #include <ceed/backend.h>
 #include "ceed-magma.h"
+#include <ceed/jit-tools.h>
+#include "ceed-magma.h"
+#ifdef HAVE_HIP
```
jeremylt, line 13: Where is this macro definition coming from? It looks a bit strange not having some `MAGMA_` or `CEED_` leading it.

nbeams: Oh yeah, back when I originally added the HIP option for the MAGMA backend, `HAVE_HIP` was the macro used in MAGMA itself, and it had to be defined for some header definitions or something like that (I don't remember the exact details).   But MAGMA itself has actually transitioned to `MAGMA_HAVE_HIP`, yet it doesn't cause errors with compilation anymore if they don't match...
Anyway, we should probably change to `MAGMA_HAVE_HIP` here, too.  I can make that change.

nbeams: Or actually, if the macro doesn't need to exactly match MAGMA anymore for header inclusion purposes, would it be preferred if we used a different one here?  like `CEED_MAGMA_USE_HIP` or something?

jeremylt: I like `CEED_BACKEND_SPECIFIC_MACRO` like `CEED_MAGMA_USE_HIP`

nbeams: Done in acc154bdaa78f7d5eb152bc74e55892088677db5

jeremylt:
Side note - I think this would be a good thing to add to `doc/sphinx/source/releasenotes.md` so that it gets highlighted in the next release

backends/magma/ceed-magma-basis.c
```diff
@@ -499,39 +608,127 @@ int CeedBasisCreateTensorH1_Magma(CeedInt dim, CeedInt P1d, CeedInt Q1d,
...
+  ierr = CeedLoadSourceToInitializedBuffer(ceed, weight_path,
+         &basis_kernel_source);
+  CeedChkBackend(ierr);
+  CeedDebug256(ceed, 2,
+               "----- Loading Basis Kernel Source Complete! -----\n");
+  // The RTC compilation code expects a Ceed with the common Ceed_Cuda or Ceed_Hip
+  // data
+  Ceed delegate;
+  ierr = CeedGetDelegate(ceed, &delegate); CeedChkBackend(ierr);
+  ierr = MAGMA_RTC_COMPILE(delegate, basis_kernel_source, &impl->module, 5,
```
nbeams, line 665: The compilation functions get `Ceed_Cuda` or `Ceed_Hip` backend data explicitly, for device properties (like ID), so we can't just pass a `Ceed` whose backend data is `Ceed_Magma`.  This seems to work, but am I breaking any rules here?  Do we need a different solution?

jeremylt: This is consistent with what we do with other backends now, though eventually we really should decide how we want backends to reach into each other's data and headers.

abdelfattah83 (approved):

backends/magma/ceed-magma.h
```diff
@@ -13,6 +13,34 @@
...
+#define MAGMA_RTC_RUN_KERNEL_DIM CeedRunKernelDimHip
+#define MAGMA_RTC_RUN_KERNEL_DIM_SH CeedRunKernelDimSharedHip
+#else
+#define MAGMA_RTC_MODULE CUmodule
+#define MAGMA_RTC_FUNCTION CUfunction
+#define MAGMA_RTC_COMPILE CeedCompileCuda
+#define MAGMA_RTC_GET_KERNEL CeedGetKernelCuda
+#define MAGMA_RTC_RUN_KERNEL CeedRunKernelCuda
+#define MAGMA_RTC_RUN_KERNEL_DIM CeedRunKernelDimCuda
+#define MAGMA_RTC_RUN_KERNEL_DIM_SH CeedRunKernelDimSharedCuda
```
jeremylt, line 41: Minor questions - 

Do we want to prepend `CEED_` here?

For the types, do we want to conditionally typedef `CeedMagmaModule` and `CeedMagmaFunction` as `hipModule_t/CUmodule` and `hipFunction_t/CUfunction`? Might make it easier to see that those are types

nbeams: Would you prefer switching the functions to CeedCamelCase as well?  Or just the types?
I guess I thought it would be good for them to be obviously macros, but we have things in libCEED like `CeedCalloc` which is a macro, so it would be consistent to use that style here.

jeremylt: For the types, I was thinking of something like
```C
#ifdef CEED_MAGMA_USE_HIP
typedef CeedMagmaModule hipModule_t;
typedef CeedMagmaFunction hipFunction_t;
#else
typedef CeedMagmaModule CUmodule;
typedef CeedMagmaFunction CUfunction;
#endif
```
so that IDEs know these are types.

I could go either way with the CamelCase for the function names. Whichever you feel will be clearer in 6 months when we've half forgotten the specific details of this PR.

nbeams: I decided to go with a style that matches the other backends better.  Hopefully I don't regret it in 6 months. :)

backends/magma/ceed-magma.h
```diff
@@ -13,6 +13,34 @@
...
+#define CeedRunKernelDimSharedMagma CeedRunKernelDimSharedHip
+#else
+#define CeedMagmaModule CUmodule
+#define CeedMagmaFunction CUfunction
+#define CeedCompileMagma CeedCompileCuda
+#define CeedGetKernelMagma CeedGetKernelCuda
+#define CeedRunKernelMagma CeedRunKernelCuda
+#define CeedRunKernelDimMagma CeedRunKernelDimCuda
+#define CeedRunKernelDimSharedMagma CeedRunKernelDimSharedCuda
+#endif
```
jeremylt, line 42:
```suggestion
#ifdef CEED_MAGMA_USE_HIP
#typedef CeedMagmaModule hipModule_t
#typedef CeedMagmaFunction hipFunction_t
#define CeedCompileMagma CeedCompileHip
#define CeedGetKernelMagma CeedGetKernelHip
#define CeedRunKernelMagma CeedRunKernelHip
#define CeedRunKernelDimMagma CeedRunKernelDimHip
#define CeedRunKernelDimSharedMagma CeedRunKernelDimSharedHip
#else
#typedef CeedMagmaModule CUmodule
#typedef CeedMagmaFunction CUfunction
#define CeedCompileMagma CeedCompileCuda
#define CeedGetKernelMagma CeedGetKernelCuda
#define CeedRunKernelMagma CeedRunKernelCuda
#define CeedRunKernelDimMagma CeedRunKernelDimCuda
#define CeedRunKernelDimSharedMagma CeedRunKernelDimSharedCuda
#endif
```

these two are types, right?

jeremylt (approved):
I like it! If CI is happy, then I think we can merge

backends/magma/ceed-magma.h
nbeams: Yeah, whoops, fixed now

nbeams:
I think the GitLab failures are not due to this branch specifically, is that correct?

jedbrown (approved):
Yeah, lv has issues and we have ordered a CUDA GPU that we intend to use in Noether.

nbeams:
Cool, should I go ahead and merge?

jedbrown:
Go for it.

PR #976, opened 2022-06-02
GPU assembly: fix possible basis loading offset error

nbeams:
Fix a bug that leads to incorrect basis matrix loading in the setup of GPU operator assembly if: 1) one of the input eval modes is `GRAD`, and 2) it is not the first eval mode.

Discovered through fluids examples (see https://github.com/CEED/libCEED/pull/962).


jeremylt (approved):
If CI likes it, I likes it

jedbrown (approved):
Thanks for tracking this down @nbeams! I'm so excited for this. And thanks @jeremylt for tests to reduce.

PR #981, opened 2022-06-06
pc - fix GPU single operator fallback

jeremylt:
Good catch @nbeams, this should fix it. It seems to be working locally for me when I use `CEED_DEBUG=1 ./build/t565-operator /gpu/cuda > out.txt` and look at the generated kernels.


nbeams (approved):
LGTM, thanks

Issue #975, opened 2022-06-02
Minor Memory Access Violation


jeremylt:
We have a minor memory access violation in certain cases for `CeedLoadSourceToInitalizedBuffer()`. Its is benign but worth fixing:

https://github.com/CEED/libCEED/blob/617436a9ced3e376309fe403f20ffa53029d2607/interface/ceed-jit-tools.c#L81

When a QFunction source file starts with a `#sometexthere` on the first line, the logic for checking if there is a `#include` can reach past the start of the file when checking if the line is `#include` (or some equivalent with spacing, like `#     include`).

We probably just need to modify this line to make sure we don't reach outside of the file for our memcpy:
https://github.com/CEED/libCEED/blob/617436a9ced3e376309fe403f20ffa53029d2607/interface/ceed-jit-tools.c#L125

PR #982, opened 2022-06-06
fixed include memcpy segv error

ajpedersen20:
Fix for issue #975


jeremylt:
This looks good. I'll be double checking with Valgrind tomorrow by making [this file ](https://github.com/CEED/libCEED/blob/main/tests/t406-qfunction-helper.h) reproduce the original error. I think it would be reasonable to move lines 8 and 9 in that file permanently (with a comment explaining what's up like on line 8 [here](https://github.com/CEED/libCEED/blob/main/tests/t406-qfunction.h)) so we can easily check for regression on this bug in the future.

jeremylt (approved):
Looks good to me. You can merge once CI finishes running. Thanks!

PR #984, opened 2022-06-07
Draft: add basis transformation procedure

AdelekeBankole:


interface/ceed-basis.c
```diff
@@ -1219,6 +1219,58 @@ int CeedBasisDestroy(CeedBasis *basis) {
...
+  @ref Utility
+**/
+int CeedTransformQuadrature(CeeddInt Q, CeedScalar *q_weight_1d) {
+  CeedScalar points, m_points, m_weights;
+  if (!mapping) {
+    if (!q_weight_1d)
+      points = Q;
+      else
+       points = Q;
+       weights = q_weight_1d;
```
jedbrown, line 1268: Note that the indentation here does not show the scope. Your editor should be able to re-indent to make the actual scope of the `if` and `else` clear.

interface/ceed-basis.c
```diff
@@ -1219,6 +1219,58 @@ int CeedBasisDestroy(CeedBasis *basis) {
...
+  @param[in] Q    Number of quadrature points
+  @param[in] rho  sum of semiminor and major axis
+  @param[out] g   conformal map
+  @param[out] g_prime   derivative of conformal map g
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Utility
+**/
+int CeedHaleTrefethenStripMap(CeedInt Q, CeedInt rho, CeedScalar *g, CeedScalar *g_prime) {
```
jedbrown, line 1234: If this is a public function, it'll need to be declared in a public header.

interface/ceed-basis.c
```diff
@@ -1270,6 +1322,8 @@ int CeedGaussQuadrature(CeedInt Q, CeedScalar *q_ref_1d,
     q_ref_1d[i] = -xi;
     q_ref_1d[Q-1-i]= xi;
   }
+  // Call transformed Gauss-Legendre quadrature
```
jedbrown, line 1325: You probably can't just call it here because you can't change the behavior of the existing interfaces, but you'll add a new interface that creates a Gauss quadrature and then transforms it.

interface/ceed-basis.c
```diff
@@ -1231,28 +1231,30 @@ int CeedBasisDestroy(CeedBasis *basis) {
 
   @ref Utility
 **/
-int CeedHaleTrefethenStripMap(CeedInt Q, CeedInt rho, CeedScalar *g, CeedScalar *g_prime) {
-  CeedScalar tau, d, C, PI2, PI = 4.0*atan(1.0);
+int CeedHaleTrefethenStripMap(CeedInt Q, CeedScalar rho, CeedScalar *g, CeedScalar *g_prime) {
```
jedbrown, line 1234: This still isn't a function $g(s)$ as we've discussed. You'll need to take `const CeedScalar *s` as an argument, then write `g[i]` as a function of `s[i]`.
image

interface/ceed-basis.c
```diff
@@ -1231,28 +1231,30 @@ int CeedBasisDestroy(CeedBasis *basis) {
...
+    g_prime[i] = -tau * C / sqrt(1.0 - sin(u[i]) * sin(u[i])) * g_prime_u[u[i]];
   }
   return CEED_ERROR_SUCCESS;
 }
 
 /**
   @brief Transform quadrature by applying a smooth mapping = (g, g_prime)
 
-  @param Q  Number of quadarture points
+  @param Q  Number of quadrature points
```
jedbrown, line 1257: In general, every argument has to be documented. I don't know what you intend this function to do; maybe it can be deleted or maybe it needs to take a `CeedBasis` as input and create a `CeedBasis` as output.

tests/t333-basis.c
```diff
@@ -0,0 +1,18 @@
...
+#include <ceed.h>
+#include <math.h>
+
+int main(int argc, char **argv) {
+    Ceed ceed;
+    CeedInt Q = 16;
+    CeedScalar rho;
+    CeedScalar *g, *g_prime;
+
+    CeedHaleTrefethenStripMap();
```
jedbrown, line 13: We talked about checking by spot-checking accuracy relative to figure 7.1a. How would you test that? Take 20-point Gauss quadrature and integrate the function, then transform the quadrature points and weights, and integrate that way in $[-1, 1]$. The second should be more accurate.

image

jedbrown (commented):
You have reference code in Julia so the job here is mainly to be able to create the same results. So start with a particular example in Julia, like the points and weights for 5 Gauss quadrature points using the $\rho=1.4$ transform.

jeremylt:
What's the plan for this PR?

jedbrown:
I think that's going to be a choice for @AdelekeBankole and perhaps another student will take interest in driving this further. I think it causes no harm to stay open for a while.

AdelekeBankole:
Yes that's correct, I plan to move move this PR forward.

zatkins-dev:
It's been a few more years. Is there intention to move forward with this PR? If not, I think it would be best to close.

PR #995, opened 2022-06-17
Fluids - Minor Refactor

LeilaGhaffari:
This PR aims at removing `computeFluxJacobian_NS(CeedScalar dF[3][5][5], ...)` and instead, using variational notation for computing the Flux Jacobian.


jedbrown:
I haven't forgotten about this (which I think is a valuable refactor). Can you just ping when you're ready to revisit this, I assume after #1011 merges and this is rebased (or merge from main into this branch)?

LeilaGhaffari:
> I haven't forgotten about this (which I think is a valuable refactor). Can you just ping when you're ready to revisit this, I assume after #1011 merges and this is rebased (or merge from main into this branch)?

Sure, I will update/fix the remaining parts (and rebase it onto main) after #1011 merges.

LeilaGhaffari:
Alright, this is ready for review.

examples/fluids/qfunctions/channel.h
```diff
@@ -113,10 +112,14 @@ CEED_QFUNCTION(Channel_Inflow)(void *ctx, CeedInt Q,
...
-  const CeedScalar cp     = context->newtonian_ctx.cp;
-  const CeedScalar gamma  = cp/cv;
+  const bool implicit          = context->implicit;
+  NewtonianIdealGasContext gas = &context->newtonian_ctx;
+  const CeedScalar cv          = gas->cv;
+  const CeedScalar cp          = gas->cp;
+  const CeedScalar gamma       = cp / cv;
+  gas->g[0] = 0.;
+  gas->g[1] = 0.;
+  gas->g[2] = 0.;
```
jrwrigh, line 122: Wouldn't this also set `g` to zero for the other functions too? Making the volumetric forcing term not work for periodic BCs?

examples/fluids/qfunctions/channel.h
```diff
@@ -83,22 +83,21 @@ CEED_QFUNCTION(ICsChannel)(void *ctx, CeedInt Q,
...
-      q0[0][i] = s.U.density;
-      for (CeedInt j=0; j<3; j++)
-        q0[j+1][i] = s.U.momentum[j];
-      q0[4][i] = s.U.E_total;
-    }
+    CeedScalar q[5] = {0};
+    if (context->newtonian_ctx.is_primitive)
+      UnpackState_Y(s.Y, q);
+    else
+      UnpackState_U(s.U, q);
```
jrwrigh, line 90: I like the `UnpackState*` functions! I think we could move this to a function pointer method similar to what I did with the `newtonian.h` boundary integral QFs, where we set a generic `SetStateFromQi` function at the start of the routine and just use that throughout, `qi` representing whatever state variables the solver is using at the time.

examples/fluids/qfunctions/channel.h
jrwrigh: If the attempt is to remove the potential energy, I think you could just set `x[0] = 0`.

examples/fluids/qfunctions/newtonian.h
```diff
@@ -373,65 +212,31 @@ CEED_QFUNCTION(RHSFunction_Newtonian)(void *ctx, CeedInt Q,
...
 
-    for (CeedInt j=0; j<3; j++) {
-      for (CeedInt k=0; k<5; k++) {
+    for (CeedInt j=0; j<3; j++)
+      for (CeedInt k=0; k<5; k++)
         Grad_v[j][k][i] = wdetJ * (dXdx[j][0] * Flux[k][0] +
                                    dXdx[j][1] * Flux[k][1] +
                                    dXdx[j][2] * Flux[k][2]);
-      }
-    }
```
jrwrigh, line 389: Any strong preference for curly braces vs not? I kinda prefer without when possible. There are a few other places in the code where we don't have curly braces on double for-loops, which is mainly why I ask.

examples/fluids/qfunctions/stabilization.h
```diff
@@ -0,0 +1,167 @@
...
+// *****************************************************************************
+// Helper function for computing Tau elements (stabilization constant)
+//   Model from:
+//     PHASTA
+//
+//   Tau[i] = itau=0 which is diagonal-Shakib (3 values still but not spatial)
+//
+// Where NOT UPDATED YET
+// *****************************************************************************
+CEED_QFUNCTION_HELPER void Tau_diagPrim(NewtonianIdealGasContext gas, State s,
```
jrwrigh, line 57: Can probably just be `Tau_diag` since it's using a `State` input now.

examples/fluids/qfunctions/newtonian.h
```diff
@@ -710,69 +461,57 @@ CEED_QFUNCTION(IJacobian_Newtonian)(void *ctx, CeedInt Q,
...
                                     dXdx[j][1] * dFlux[k][1] +
                                     dXdx[j][2] * dFlux[k][2]);
-      }
-    }
 
     const CeedScalar dbody_force[5] = {0, ds.U.density *g[0], ds.U.density *g[1], ds.U.density *g[2], 0};
     for (int j=0; j<5; j++)
       v[j][i] = wdetJ * (context->ijacobian_time_shift * dU[j] - dbody_force[j]);
 
-    if (1) {
```
jrwrigh, line 732: I know I had asked @jedbrown about removing this before (see [here](https://github.com/CEED/libCEED/pull/962#discussion_r889586560)). Not sure if the same reasoning applies now though. Jed mentions refactoring this block of code, so maybe this is good to go now.

examples/fluids/qfunctions/stabilization.h
```diff
@@ -0,0 +1,167 @@
...
+//
+// This file is part of CEED:  http://github.com/ceed
+
+/// @file
+/// Helper functions for computing stabilization terms of a newtonian simulation
+
+
+#ifndef stabilization_h
+#define stabilization_h
+
```
jrwrigh, line 14:
```suggestion
#include "newtonian_state.h"
#include <ceed.h>
```

Stop some unnecessary warnings from my LSP.

examples/fluids/qfunctions/stabilization.h
```diff
@@ -0,0 +1,167 @@
...
+  // but in that case energy tau is scaled by the product of Ctau_E * Ctau_M
+  // OR we could absorb cv into Ctau_E but this puts more burden on user to
+  // know how to change constants with a change of fluid or units.  Same for
+  // Ctau_v * mu * mu IF AND ONLY IF we don't add viscosity law =f(T)
+}
+
+// *****************************************************************************
+// Helper function for computing flux Jacobian of Primitive variables
+// *****************************************************************************
+CEED_QFUNCTION_HELPER void computeFluxJacobian_NSp(CeedScalar dF[3][5][5],
```
jrwrigh, line 131: Do we need to keep this?

examples/fluids/qfunctions/stabilization.h
```diff
@@ -0,0 +1,167 @@
...
+    const CeedScalar rho, const CeedScalar u[3], const CeedScalar E,
+    const CeedScalar Rd, const CeedScalar cv) {
+  CeedScalar u_sq = u[0]*u[0] + u[1]*u[1] + u[2]*u[2]; // Velocity square
```
jrwrigh, line 134:
```suggestion
  CeedScalar u_sq = Dot3(u,u); // Velocity square
```

examples/fluids/qfunctions/densitycurrent.h
```diff
@@ -152,17 +152,15 @@ CEED_QFUNCTION(ICsDC)(void *ctx, CeedInt Q,
...
-      q0[0][i] = s.U.density;
-      for (CeedInt j=0; j<3; j++)
-        q0[j+1][i] = s.U.momentum[j];
-      q0[4][i] = s.U.E_total;
-    }
+    CeedScalar q[5] = {0};
+    if (context->newtonian_ctx.is_primitive)
+      UnpackState_Y(s.Y, q);
+    else
+      UnpackState_U(s.U, q);
```
jrwrigh, line 159: Could do similar function pointer stuff.

jrwrigh (approved):
Few suggestions. Otherwise, looks good to me, and a few extra comments:

1) There are `__attribute ((unused))` by the initializations of `Tau_d` (https://github.com/CEED/libCEED/blob/779062404f6c183e987cbc971fc4cdeea0db0363/examples/fluids/qfunctions/newtonian.h#L431 and https://github.com/CEED/libCEED/blob/779062404f6c183e987cbc971fc4cdeea0db0363/examples/fluids/qfunctions/newtonian.h#L1063). I don't think these are necessary anymore?

2) Might be worth switching to Function pointers for `State` related functions for `IFunction_Newtonian`. Then we can move to only having one IFunction instead of 2. The only difference between them now is different treatments of `q_dot`. I don't think that extra bit of efficiency is worth keeping them separate. But that can also be a different PR, as I'm not sure if how I currently implemented the function pointers is good or not (or really if it can be improved).

examples/fluids/qfunctions/channel.h
jedbrown: I'd want to do a performance test before using function pointers in perf-critical code (like the main volume integrals). I'm less concerned elsewhere. It's possible to transform so the key function pointers are passed as arguments to a function that is meant to be inlined (topic of another PR if we want to go that route).

examples/fluids/qfunctions/newtonian.h
jedbrown: Not a big deal to me. Rust requires them always.

jeremylt: In Ratel I adopted the PETSc convention - {} required for all multi line statements.

So
good:
```C
for (i = 0; i < 42; i++) total += array[i];
```
good:
```C
for (i = 0; i < 42; i++) {
  total += array[i];
}
```
bad:
```C
for (i = 0; i < 42; i++)
  total += array[i];
```

I mostly did that because I have screwed it up a couple of times

examples/fluids/qfunctions/newtonian.h
```diff
@@ -710,69 +461,57 @@ CEED_QFUNCTION(IJacobian_Newtonian)(void *ctx, CeedInt Q,
...
+    // -- Stabilization method: none, SU, or SUPG
+    // Strong convective flux
+    CeedScalar dstrong_conv[5] = {0};
+    FluxInviscidStrong(context, s, grad_ds, dstrong_conv);
+
+    // Strong residual
+    CeedScalar dstrong_res[5];
+    for (int j=0; j<5; j++)
+      dstrong_res[j] = context->ijacobian_time_shift * dU[j] +
+                       dstrong_conv[j] - dbody_force[j];
```
jedbrown, line 485: I guess one way to write this would be to have NONE/SU/SUPG just change what terms are active on this line, and then you could remove the switch down below. The current structure is fine too.

jedbrown (approved):

examples/fluids/qfunctions/stabilization.h
jedbrown: It doesn't get called. We might want such a thing when implementing an eigenvalue/matrix-sqrt stabilization operator, though we can resurrect it from the history if we don't have a better way to compute the matrix at that time.

examples/fluids/qfunctions/stabilization.h
jedbrown: I think the point is that physically, it's meant to take a (conservative) residual and produce fluctuations Y' (primitive). The `Prim` isn't about how the input state is represented, but that the output (projected subgrid fluctuations) Y' is primitive.

jedbrown: The `NOT UPDATED YET` can presumably be removed.

examples/fluids/qfunctions/newtonian.h
jedbrown: :shrug: Not important to me any more, as our stabilization code has been exercised more.

examples/fluids/qfunctions/stabilization.h
```diff
@@ -0,0 +1,167 @@
...
+  const CeedScalar dx_i[3] = {0};
+  StateConservative dF[3];
+  dYFromTau(Y, Tau_d, dY);
+  State ds = StateFromY_fwd(gas, s, dY, x, dx_i);
+  FluxInviscid_fwd(gas, s, ds, dF);
+  for (CeedInt j=0; j<3; j++) {
+    CeedScalar dF_j[5];
+    UnpackState_U(dF[j], dF_j);
+    for (CeedInt k=0; k<5; k++)
+      stab[k][j] += dF_j[k];
```
jrwrigh, line 44: Should we move the zeroing of `stab[][]` into the function itself. Seems like a simple thing that might prevent a headache in the future.

examples/fluids/qfunctions/densitycurrent.h
jrwrigh: See https://github.com/CEED/libCEED/pull/995#discussion_r927745566

examples/fluids/qfunctions/channel.h
LeilaGhaffari: I just confirmed that it doesn't affect other QFunctions but the reason I chose to set g to zero is that I thought it was more intuitive and could help with the readability of the code. I was also concerned about  nonzero g[2] and g[2] but I am seeing now that they are set to zero. Anyways, just changed it in [this commit](https://github.com/CEED/libCEED/pull/995/commits/f712d7f79c535eaa1b03e0b87d72a30fbb5c7840) with more comments.

examples/fluids/qfunctions/newtonian.h
LeilaGhaffari: I was trying to make the style consistent throughout that file but if you are seeing more of those curly braces, then I have missed them.

jeremylt: haha, I have been trying to push code I write more towards the style *with* braces instead of *without* braces.

examples/fluids/qfunctions/stabilization.h
LeilaGhaffari: Good call!

examples/fluids/qfunctions/newtonian.h
LeilaGhaffari: I agree with you, @jeremylt , that the last style is more error-prone and I am happy to stick to the first two if we want to add it to our style convention.

jedbrown: I think clang-format can do this automatically if we migrate there at some point.

examples/fluids/qfunctions/newtonian.h
LeilaGhaffari: What do you think of [the last commit](https://github.com/CEED/libCEED/pull/995/commits/b70ae9d3e37855b20b547f7fe4ba07cd52b70e4b)?

examples/fluids/qfunctions/stabilization.h
```diff
@@ -52,6 +53,25 @@ CEED_QFUNCTION_HELPER void Stabilization(NewtonianIdealGasContext gas, State s,
...
+    break;
+  case STAB_SU:
+    FluxInviscidStrong(gas, s, ds, Y);
+    break;
+  case STAB_SUPG:
+    FluxInviscidStrong(gas, s, ds, Y);
+    for (CeedInt j=0; j<5; j++) Y[j] += U_dot[j] - body_force[j];
+    break;
+  }
+  StabilizationMatrix(gas, s, Tau_d, Y, x, stab);
```
jedbrown, line 72: To avoid confusion, the `Y` here has units of `[conservative]/[time]` so it should have a different name.

examples/fluids/qfunctions/newtonian.h
jedbrown: Great, modulo one comment I left there.

examples/fluids/qfunctions/stabilization.h
LeilaGhaffari: Oh, right. I meant to change that. `U_dot`?

LeilaGhaffari: Nope, we have that already!

examples/fluids/qfunctions/channel.h
jrwrigh: > I just confirmed that it doesn't affect other QFunctions

Interesting. Does libCEED make a copy of the struct if you don't set `CeedQFunctionSetContextWritable`? @jedbrown @jeremylt

examples/fluids/qfunctions/stabilization.h
jedbrown: `R` for residual is common in the literature.

examples/fluids/qfunctions/channel.h
jeremylt: It shouldn't. If you don't set the context read-only, then changes should persist. Could be that this ctx has a copy of the struct in creation

jedbrown: I'd thought this was copied out somewhere in the qfunctions, but perhaps not. I don't immediately know why this wouldn't affect other qfunctions.

examples/fluids/qfunctions/newtonian.h
LeilaGhaffari: Cool, just fixed renamed it.

jedbrown (approved):
Looks good to me now.

LeilaGhaffari:
Thanks! I am happy to `squash-merge` if there are no objections.

jedbrown:
No objections from me as soon as the last test passes.

LeilaGhaffari:
Sorry, I had messed up the sign for stabilization terms in `RHSFunction_Newtonian()` (caused by copy/pasting) but I think it is ready to merge after the tests pass.

PR #997, opened 2022-06-17
fluids - add preloading

jeremylt:
https://petsc.org/main/docs/manualpages/Profiling/PetscPreLoadBegin/

> command line option -preload true or -preload false

I think this is the command line arg

### DIFF
--- a/examples/fluids/src/setupts.c
+++ b/examples/fluids/src/setupts.c
@@ -481,19 +481,40 @@ PetscErrorCode TSSolve_NS(DM dm, User user, AppCtx app_ctx, Physics phys,
   }
 
   // Solve
-  double start, cpu_time_used;
-  start = MPI_Wtime();
-  ierr = PetscBarrier((PetscObject) *ts); CHKERRQ(ierr);
-  ierr = TSSolve(*ts, *Q); CHKERRQ(ierr);
-  cpu_time_used = MPI_Wtime() - start;
-  ierr = TSGetSolveTime(*ts, &final_time); CHKERRQ(ierr);
-  *f_time = final_time;
-  ierr = MPI_Allreduce(MPI_IN_PLACE, &cpu_time_used, 1, MPI_DOUBLE, MPI_MIN,
-                       comm); CHKERRQ(ierr);
+  PetscScalar start_time;
+  ierr = TSGetTime(*ts, &start_time); CHKERRQ(ierr);
+
+  PetscPreLoadBegin(PETSC_FALSE, "Fluids Solve");
+  PetscCall(TSSetTime(*ts, start_time));
+  PetscCall(TSSetStepNumber(*ts, 0));
+  if (PetscPreLoadingOn) {
+    // LCOV_EXCL_START
+    SNES      snes;
+    Vec       Q_preload;
+    PetscReal rtol;
+    PetscCall(VecDuplicate(*Q, &Q_preload));
+    PetscCall(VecCopy(*Q, Q_preload));
+    PetscCall(TSGetSNES(*ts, &snes));
+    PetscCall(SNESGetTolerances(snes, NULL, &rtol, NULL, NULL, NULL));
+    PetscCall(SNESSetTolerances(snes, PETSC_DEFAULT, .99, PETSC_DEFAULT,
+                                PETSC_DEFAULT, PETSC_DEFAULT));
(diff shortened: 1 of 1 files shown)


examples/fluids/src/setupts.c
```diff
@@ -479,12 +479,39 @@ PetscErrorCode TSSolve_NS(DM dm, User user, AppCtx app_ctx, Physics phys,
...
-  ierr = PetscBarrier((PetscObject) *ts); CHKERRQ(ierr);
-  ierr = TSSolve(*ts, *Q); CHKERRQ(ierr);
-  cpu_time_used = MPI_Wtime() - start;
-  ierr = TSGetSolveTime(*ts, &final_time); CHKERRQ(ierr);
+  PetscScalar start_time;
+  ierr = TSGetTime(*ts, &start_time); CHKERRQ(ierr);
+
+  PetscPreLoadBegin(PETSC_FALSE, "Fluids Solve");
+  PetscCall(TSSetTime(*ts, start_time));
+  PetscCall(TSSetStepNumber(*ts, 0));
```
jeremylt, line 487: Personally, I'm fine with starting to move to this `PetscCall` syntax since we need to do a pass for it through these examples after this proposal submision

jedbrown: No complaint from me.

examples/fluids/src/setupts.c
```diff
@@ -481,19 +481,40 @@ PetscErrorCode TSSolve_NS(DM dm, User user, AppCtx app_ctx, Physics phys,
...
+                                PETSC_DEFAULT, PETSC_DEFAULT));
+    PetscCall(VecDestroy(&Q_preload));
+    // LCOV_EXCL_STOP
+  } else {
+    ierr = PetscBarrier((PetscObject) *ts); CHKERRQ(ierr);
+    ierr = TSSolve(*ts, *Q); CHKERRQ(ierr);
+  }
+  PetscPreLoadEnd();
+
+  PetscCall(TSGetSolveTime(*ts, &final_time));
```
jrwrigh, line 513:
```suggestion
  PetscCall(TSGetSolveTime(*ts, &final_time));
  *f_time = final_time;
```
To keep the output from `PostProcess_NS` the same.

examples/fluids/src/setupts.c
```diff
@@ -481,19 +481,40 @@ PetscErrorCode TSSolve_NS(DM dm, User user, AppCtx app_ctx, Physics phys,
...
   if (!app_ctx->test_mode) {
     ierr = PetscPrintf(PETSC_COMM_WORLD,
                        "Time taken for solution (sec): %g\n",
-                       (double)cpu_time_used); CHKERRQ(ierr);
+                       final_time); CHKERRQ(ierr);
```
jrwrigh, line 517: The output from `TSGetSolveTime` appears to be the solution time, rather than the wall-clock time. So this print statement doesn't post the same information as before.

jrwrigh (changes requested):

examples/fluids/src/setupts.c
jrwrigh: Maybe need to use `PetscLogEventGetPerfInfo` instead, but for the entire Fluid Solve stage rather than a specific event?

jrwrigh: Maybe:

```
PetscStageLog stageLog;
PetscStageInfo stageInfo;
PetscCall(PetscLogGetStageLog(&stageLog));
stageInfo = stageLog->stageInfo;
wall_time = stageInfo[1].perfInfo.time;
```
Assuming that the fluid solve stage is the first stage. Probably a more robust way of selecting the stage though.

jeremylt: I think this might only turn on when the flag `-log_view` is used... which we really should be using for all timing runs

jedbrown: It's not rapidly changing and if anything changed, it would be a compiler error that we can easily notice and fix. I don't like that they're public structs, and I think it would be good to make them private, but I also don't envision anyone taking that initiative because it would involve writing/documenting a bunch of accessors.

jrwrigh (approved):
This actually works fine on my end without `-log_view`, so I think this is that fragile.

Looks like you accidentally added `fluids-navierstokes.kkxk`, otherwise LGTM.

PR #994, opened 2022-06-17
Fluids block diagonal

jedbrown:

PR #998, opened 2022-06-19
CeedOperatorAssemblePointBlockDiagonal: fix restriction sizes for subdomain integral

jedbrown:
Formerly would give errors such as

interface/ceed-elemrestriction.c:862 in CeedElemRestrictionApply(): Output vector size 125050 not compatible with element restriction (124050, 6000)

(noted in https://github.com/CEED/libCEED/pull/994#discussion_r899808040)


jeremylt (approved):
Oof, yeah, this is obviously better

PR #1001, opened 2022-06-21
gpu - fix setting device id

### LINKED ISSUE #999: Cannot set device ID through resource name
Setting GPU device ID when specifying a backend is broken.  Trying to add it gives an error, e.g.:

```
in CeedInit_Cuda(): Cuda backend cannot use resource: /gpu/cuda/ref:device_id=0
```

I think it stems from #863, specifically commit f87d896ccdb3ae7eb58212d71cf20ac055d48f9d ?

Was this unintended behavior?  Or do we want to add a different way of setting/configuring device ID?  

If removing this method of setting device ID was intentional, we should update the README (which currently explains how to set the device ID for CUDA/HIP/MAGMA backends).

jeremylt:
Fixes #999

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen.c
+++ b/backends/cuda-gen/ceed-cuda-gen.c
@@ -16,11 +16,16 @@
 static int CeedInit_Cuda_gen(const char *resource, Ceed ceed) {
   int ierr;
 
-  if (strcmp(resource, "/gpu/cuda") && strcmp(resource, "/gpu/cuda/gen"))
+  char *resource_root;
+  ierr = CeedCudaGetResourceRoot(ceed, resource, &resource_root);
+  CeedChkBackend(ierr);
+  if (strcmp(resource_root, "/gpu/cuda")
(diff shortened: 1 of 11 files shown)


jeremylt:
Note: This targets `release` and will need to be merged back into main

jeremylt:
CI is failing because we don't have a compatible version of PETSc on Noether, but everything works fine for me locally with this commit on main.

jedbrown (approved):
Seems fine at a glance, thanks. It may ultimately make life easier to containerize our CI, shipping Dockerfiles for each environment (and having them rebuild as needed).

nbeams (approved):
LGTM, and I tested locally on a multi-GPU machine (for CUDA) to confirm that the correct device was being used when I changed it. 

My only question is whether we want stricter matching for the MAGMA backends as well?

jeremylt:
Since this is a bugfix to `release`, I think we can decide to enforce stricter resource matching for MAGMA in a follow-up

PR #1002, opened 2022-06-21
Add CeedInt_FMT

jeremylt:
As discussed in #991

### DIFF
--- a/backends/opt/ceed-opt-operator.c
+++ b/backends/opt/ceed-opt-operator.c
@@ -820,7 +820,7 @@ int CeedOperatorCreate_Opt(CeedOperator op) {
   if (blk_size != 1 && blk_size != 8)
     // LCOV_EXCL_START
     return CeedError(ceed, CEED_ERROR_BACKEND,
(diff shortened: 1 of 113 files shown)


jrwrigh (approved):
I think there are some `%d` references in `backends/ceed-magma-basis.c` that need to be changed over. Otherwise LGTM

backends/opt/ceed-opt-operator.c
```diff
@@ -820,7 +820,7 @@ int CeedOperatorCreate_Opt(CeedOperator op) {
   if (blk_size != 1 && blk_size != 8)
     // LCOV_EXCL_START
     return CeedError(ceed, CEED_ERROR_BACKEND,
-                     "Opt backend cannot use blocksize: %d", blk_size);
+                     "Opt backend cannot use blocksize: " CeedInt_FMT, blk_size);
```
rezgarshakeri, line 823: I think `%` is missing here. I assume you replace `d` with `" CeedInt_FMT"`?

jeremylt: Good catch -  fixed

PR #1004, opened 2022-06-21
Doc fixes

jeremylt:
I'm still trying to figure out why Doxygen is duplicating the definition of `CeedEvalMode`

### DIFF
--- a/doc/sphinx/source/api/Ceed.rst
+++ b/doc/sphinx/source/api/Ceed.rst
@@ -31,3 +31,6 @@ Typedefs and Enumerations
 
 .. doxygenenum:: CeedMemType
    :project: libCEED
+
+.. doxygenenum:: CeedErrorType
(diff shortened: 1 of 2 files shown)


jeremylt:
Fixed:
image

@jrwrigh, can you confirm? I think the `docs/readthedocs.org:libceed` CI job will let you look at a preview of the changes

jrwrigh:
Yep, that looks better.

Doxyfile
```diff
@@ -964,6 +964,7 @@ RECURSIVE              = YES
 EXCLUDE                = benchmarks \
                          include/ceed/khash.h \
                          include/ceedf.h \
+                         include/ceed/jit-source \
```
jeremylt, line 967: A single header that gets included in the JiT would have prevented this as well.

jrwrigh (approved):

PR #1005, opened 2022-06-22
basis - add CeedBasisCreateProlongation

jeremylt:
This is a refactor that exposes the prolongation basis setup to users. This will let me take the coordinates basis and the solution basis and interpolate from a lower order coordinates basis to a higher order solution basis. I think that this is the last piece of the puzzle for Dirichlet BCs in libCEED.

### DIFF
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -830,6 +830,172 @@ int CeedBasisCreateHdiv(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
 }
 
 /**
+  @brief Create a CeedBasis for projection from the nodes of `basis_from`
+           to the nodes of `basis_to`. Only `CEED_EVAL_INTERP` will be
+           valid for the new basis, `basis_project`. This projection is
+           given by `interp_project = interp_to^+ * interp_from`, where
+           the pesudoinverse `interp_to^+` is given by QR factorization.
+         Note: `basis_from` and `basis_to` must have compatible quadrature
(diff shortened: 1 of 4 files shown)


interface/ceed-basis.c
```diff
@@ -829,6 +829,161 @@ int CeedBasisCreateHdiv(Ceed ceed, CeedElemTopology topo, CeedInt num_comp,
...
+}
+
+/**
+  @brief Create the interpolation matrix for prolongation from the nodes of `basis_c`
+           to the nodes of `basis_f`.
+
+  @param[in] basis_fine      Fine CeedBasis to prolong to
+  @param[in] basis_coarse    Coarse CeedBasis to prolong from
+  @param[out] interp_c_to_f  Address of the variable where the newly created
+                               prolongation matrix will be stored.
```
jedbrown, line 904: Let's just document (in the man page, in addition to the check you have below) that these must have matching quadratures.

And perhaps write this creates a $B_p$ such that $B_f B_p = B_c$. This is possible any time it's actually a prolongation, but there may be times when one wants to reverse the roles, asking for $B_c B_p \approx B_f$. There are stabilization methods that use this sort of operator so we could call this something like `CeedBasisCreateProjectionMatrix` so that it makes sense for both uses.

jedbrown (approved):

interface/ceed-basis.c
jeremylt: Good call on all counts.

PR #1009, opened 2022-06-23
Fluids - Use libCEED to compute Dirichlet boundary conditions

jrwrigh:
Use libCEED functions to strongly enforce (dynamic) boundary conditions. Similar to this [Ratel MR](https://gitlab.com/micromorph/ratel/-/merge_requests/150/), though will have to contend with ~~non-homogenous static Dirichlet boundary conditions.~~ For now we'll ignore non-homogenous static Dirichlet BCs. That can be easily fixed later via adding a restriction for them and an identity operator.

Todos:
- [ ] ~~Create restriction for the static Dirichlet nodes~~
- Either:
	- [ ] ~~Use Identity QF to skip the non-static stuff~~
- or:
	- [ ] ~~Create QFunction to populate static CeedVector with static Dirichlet data~~
	- [ ] ~~Create Qfunction to apply static Dirichlet data~~
- [x] Create QFunction for setting up Dirichlet BC data
- [x] Create Qfunction to apply dynamic Dirichlet data
	- Possibly just reuse the existing STG QF
- [ ] ~~Create Dirichlet Operator that combines the two together~~
- [x] Function: DMInsertBoundaryValues replacement
- [x] Function: Setup qfunctions, operators, etc.


KennethEJansen:
I can imagine future needs for coexistence with non-homogenous static BCs (e.g.Strong T on a wall) but what is the current need? we have periodicity on zmax and zmin, weak P on xmax and slanted top, zero heat flux and no slip on the wall and strong/dynamic STG on density and momentum on the inflow.

jedbrown:
I think there's no harm in considering all inhomogeneous strong BCs to be transient.

examples/fluids/src/setuplibceed.c
```diff
@@ -313,6 +315,159 @@ PetscErrorCode CreateOperatorForDomain(Ceed ceed, DM dm, SimpleBC bc,
...
+  PetscMemType q_mem_type;
+  PetscFunctionBeginUser;
+
+  PetscCall(DMGetApplicationContext(dm, &user));
+  // Setup libCEED vector
+  PetscCall(VecGetArrayAndMemType(Q_loc, &q, &q_mem_type));
+  CeedVectorSetArray(user->q_ceed, MemTypeP2C(q_mem_type), CEED_USE_POINTER, q);
+
+  // Apply libCEED operator
+  CeedOperatorApply(user->op_dirichlet, CEED_VECTOR_NONE, user->q_ceed,
```
jeremylt, line 333: Side note - this really should use ApplyAdd

We found a bug with viz by overwriting the target vector.
See one potential fix here:
https://gitlab.com/micromorph/ratel/-/blob/main/src/ratel-petsc-ops.c#L279

jeremylt: Right. In Ratel we are manually applying the mask before adding in to fix the correctness of visualization.

jrwrigh: In the Ratel example, what is stored in `ratel->X[0]`? Not sure how best to build the mask in this context.

jeremylt: X[0] is the global solution vector. To create the mask, I set the local vector to 0, set the global vector to 1, and then did a global to local.

examples/fluids/src/dirichlet.c
```diff
@@ -0,0 +1,194 @@
...
+PetscErrorCode SetupStrongBC_Ceed(Ceed ceed, CeedData ceed_data, DM dm,
+                                  User user, AppCtx app_ctx, ProblemData *problem,
+                                  SimpleBC bc, CeedInt Q_sur, CeedInt q_data_size_sur) {
+  PetscFunctionBeginUser;
+
+  {
+    Vec boundary_mask, global_vec;
+
+    PetscCall(DMGetNamedLocalVector(dm, "boundary mask", &boundary_mask));
+    PetscCall(DMCreateGlobalVector(dm, &global_vec));
```
jrwrigh, line 172: I've currently set for a global vector to be created and destroyed to get the boundary mask value. I could alternatively use the `Q` vector created in the `main` function, but that'd involve needing to pass it through several functions or adding it to a struct. This is currently run before `Q` is created, so there aren't two copies of a global vector. 

Wasn't sure if having two "utility" global vectors would actually take up space; Not entirely sure how global vs local vectors work on the backend (ie. whether PETSc knows that the only value in `global_vec` is just 1 and doesn't need to allocate a full array for that).

jedbrown: PETSc doesn't optimize that case, but you're doing this once so it's totally reasonable. You can use `DMGetGlobalVector` to have it be a work vector that doesn't get deallocated when returned (so the the next caller can have it). I think either is fine here.

examples/fluids/qfunctions/stg_shur14.h
```diff
@@ -373,4 +373,63 @@ CEED_QFUNCTION(STGShur14_Inflow_Jacobian)(void *ctx, CeedInt Q,
...
+
+    InterpolateProfile(coords[1][i], ubar, cij, &eps, &lt, stg_ctx);
+    if (!mean_only) {
+      CalcSpectrum(coords[1][i], eps, lt, h, mu/rho, qn, stg_ctx);
+      STGShur14_Calc(x, time, ubar, cij, qn, u, stg_ctx);
+    } else {
+      for (CeedInt j=0; j<3; j++) u[j] = ubar[j];
+    }
+
+    bcval[0][i] = scale[i] * rho;
```
jeremylt, line 427: aren't there 5 fields? You should set all output values, or you may get junk values in various GPU backends

examples/fluids/src/dirichlet.c
```diff
@@ -0,0 +1,194 @@
...
+                              &qf_setup);
+  CeedQFunctionAddInput(qf_setup, "x", num_comp_x, CEED_EVAL_INTERP);
+  CeedQFunctionAddInput(qf_setup, "multiplicity", num_comp_q, CEED_EVAL_NONE);
+  CeedQFunctionAddOutput(qf_setup, "x stored", num_comp_x, CEED_EVAL_NONE);
+  CeedQFunctionAddOutput(qf_setup, "scale", 1, CEED_EVAL_NONE);
+
+  // Compute contribution on each boundary face
+  for (CeedInt i=0; i < bc->num_inflow; i++) {
+    // -- Restrictions
+    PetscCall(GetRestrictionForDomain(ceed, dm, height, domain_label,
```
jeremylt, line 44: Here, this Restriction, I think, has 5 components, so that 5th component is dumping junk into the local vector for some of the GPU backends.

examples/fluids/qfunctions/stg_shur14.h
jrwrigh: Yep. Result of lazy copy/paste from the `bcFunc`. Though I guess it didn't really matter to the actual calculation since `DMPlexGlobalToLocal` ends up just overwriting the entry anyways. It'll still effect `VecView` though.

examples/fluids/src/dirichlet.c
jrwrigh: Yep, that should be fixed now, but it's still failing with `/gpu/hip/gen` on noether.

jrwrigh: It's still failing, even with the latest commit (which forces the unset output to be zeroed out).

examples/fluids/qfunctions/dirichlet_boundary.h
```diff
@@ -0,0 +1,32 @@
...
+  const CeedScalar (*multiplicity)       = (const CeedScalar(*))in[1];
+
+  // Outputs
+  CeedScalar (*coords_stored)[CEED_Q_VLA] = (CeedScalar(*)[CEED_Q_VLA])out[0];
+  CeedScalar (*scale_stored)              = (CeedScalar(*))out[1];
+
+  CeedPragmaSIMD
+  for(CeedInt i=0; i<Q; i++) {
+    for (int j=0; j<3; j++) coords_stored[j][i] = coords[j][i];
+    scale_stored[i] = 1.0 / multiplicity[i];
```
jeremylt, line 27: Minor - I don't think it has anything to do with correctness, but `multiplicity` has 5 components here

jeremylt: Actually, I'm not sure if it would cause an issue with the code generation or not

examples/fluids/qfunctions/stg_shur14.h
```diff
@@ -373,4 +373,64 @@ CEED_QFUNCTION(STGShur14_Inflow_Jacobian)(void *ctx, CeedInt Q,
...
+    const CeedScalar x[] = { coords[0][i], coords[1][i], coords[2][i] };
+    const CeedScalar dXdx[2][3] = {
+      {q_data_sur[4][i], q_data_sur[5][i], q_data_sur[6][i]},
+      {q_data_sur[7][i], q_data_sur[8][i], q_data_sur[9][i]}
+    };
+
+    CeedScalar h[3];
+    for (CeedInt j=0; j<3; j++)
+      h[j] = 2/sqrt(Square(dXdx[0][j]) + Square(dXdx[1][j]));
+    h[0] = dx;
```
jeremylt, line 417: minor - why set `h[0]` twice?

examples/fluids/qfunctions/dirichlet_boundary.h
jrwrigh: Hmmm. Didn't realize that. 

Is there a circumstance in which the five entries of multiplicity would be different? I think that would require that certain components not be shared at nodes, but I can't think of a circumstance where that's true other than (maybe?) mixed FEM.

jeremylt: all 5 fields should be the same in this case

examples/fluids/qfunctions/stg_shur14.h
jrwrigh: No particular reason. Pretty sure I wrote the loop before I had `dx` get passed in.

examples/fluids/problems/stg_shur14.c
```diff
@@ -486,3 +488,24 @@ PetscErrorCode SetupStrongSTG(DM dm, SimpleBC bc, ProblemData *problem) {
...
+  PetscFunctionBeginUser;
+  CeedQFunctionCreateInterior(ceed, 1, STGShur14_Inflow_StrongQF,
+                              STGShur14_Inflow_StrongQF_loc, &qf_strongbc);
+  CeedQFunctionAddInput(qf_strongbc, "surface qdata", q_data_size_sur,
+                        CEED_EVAL_NONE);
+  CeedQFunctionAddInput(qf_strongbc,  "x",     num_comp_x, CEED_EVAL_NONE);
+  CeedQFunctionAddInput(qf_strongbc,  "scale", 1,          CEED_EVAL_NONE);
+  CeedQFunctionAddOutput(qf_strongbc, "q",     num_comp_q, CEED_EVAL_NONE);
+
+  CeedQFunctionContextReferenceCopy(problem->ics.qfunction_context, &stg_context);
```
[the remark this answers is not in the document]
jrwrigh: I've run the code through valgrind, but am not sure how to interpret the output with regard to seeing if a specific context is destroyed. Running with `--leak-check=full --show-leak-kinds=all --track-origins=yes`, and the only references to contexts are for the creation of the STG context, but not this specific line.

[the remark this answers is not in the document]
jeremylt: If it's still showing as allocated at the end, that means the destructor wasn't called enough times. I suspect it needs another destory here. Alternatively you could directly set the context for the QF from the problem struct instead of copying the reference an additional time.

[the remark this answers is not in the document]
jrwrigh: > If it's still showing as allocated at the end, that means the destructor wasn't called enough times.

I'm not sure how to determine that based on the valgrind output. 

> I suspect it needs another destory here.

Yeah, I think that's right. I didn't realize that the context set to the `problem->apply_<bc>.qfunction_context`s was destroyed after being set, so I misunderstood the lifecycle of the QFContexts.

examples/fluids/src/dirichlet.c
jeremylt: I looked over the code a few times and don't see any obvious reason why this would be the case. It might be useful to check the l-vec after boundary value insertion on a small problem with `/gpu/hip/shared` and `/gpu/hip/gen`

It is probably too much information, but you can see the generated code by setting the environment variable `CEED_DEBUG=1` or `DEBUG=1`

jedbrown:
@jrwrigh How is this branch going? What do we need before merging it?

examples/fluids/src/dirichlet.c
jedbrown: Oh, is this `/gpu/hip/gen` issue still unresolved?

jrwrigh:
The only thing left is the `/gpu/hip/gen` failure (discussed a bit on zulip, but basically it fails with a large error norm for the strong STG test case, but only for `/gpu/hip/gen`). I was going to look at the L vector like Jeremy had suggested and also try and run the test at Summit/Perlmutter to see if the issue exists on CUDA.

jedbrown:
I can confirm it's also present with `/gpu/cuda/gen`.

examples/fluids/src/dirichlet.c
```diff
@@ -0,0 +1,194 @@
...
+
+    CeedOperatorSetField(op_dirichlet_sub, "surface qdata", elem_restr_qd_sur,
+                         CEED_BASIS_COLLOCATED, q_data_sur);
+    CeedOperatorSetField(op_dirichlet_sub, "x", elem_restr_x_stored,
+                         CEED_BASIS_COLLOCATED, x_stored);
+    CeedOperatorSetField(op_dirichlet_sub, "scale", elem_restr_scale,
+                         CEED_BASIS_COLLOCATED, scale_stored);
+    CeedOperatorSetField(op_dirichlet_sub, "q", elem_restr_q_sur,
+                         CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
+    CeedOperatorSetNumQuadraturePoints(op_dirichlet_sub, elem_size);
```
jeremylt, line 106: The number of quadratre points seems correct hore

jedbrown: I think this level must all be correct because the cuda/shared (and CPU) backends have correct results and no extra quadrature points.

jeremylt:
We could dump the JiT source with DEBUG=1 and verify the number of quadratre points set there?

nbeams:
In the all-collocated-basis case, this check never happens:
```
useCollograd = useCollograd && basis_data->d_collo_grad_1d;
```
so `useCollograd` is `true`.  I haven't dug enough to figure out which code difference(s) in trying to do the "collograd" version may be involved, but if I set it to false by checking whether all input/output basis eval modes are collocated:
```
--- a/backends/hip-gen/ceed-hip-gen-operator-build.cpp
+++ b/backends/hip-gen/ceed-hip-gen-operator-build.cpp
@@ -814,10 +814,12 @@ extern "C" int CeedHipGenOperatorBuild(CeedOperator op) {
 
   // Find dim and Q1d
   bool useCollograd = true;
+  bool allCollocated = true;
   data->maxP1d = 0;
   for (CeedInt i = 0; i < numinputfields; i++) {
     ierr = CeedOperatorFieldGetBasis(opinputfields[i], &basis); CeedChkBackend(ierr);
+      allCollocated = false;
       ierr = CeedBasisGetData(basis, &basis_data); CeedChkBackend(ierr);
       ierr = CeedQFunctionFieldGetEvalMode(qfinputfields[i], &emode);
       CeedChkBackend(ierr);
@@ -846,6 +848,7 @@ extern "C" int CeedHipGenOperatorBuild(CeedOperator op) {
     ierr = CeedOperatorFieldGetBasis(opoutputfields[i], &basis); CeedChkBackend(ierr);
 
+      allCollocated = false;
       ierr = CeedBasisGetData(basis, &basis_data); CeedChkBackend(ierr);
       ierr = CeedQFunctionFieldGetEvalMode(qfoutputfields[i], &emode);
       CeedChkBackend(ierr);
@@ -868,6 +871,8 @@ extern "C" int CeedHipGenOperatorBuild(CeedOperator op) {
   }
   data->dim = dim;
   data->Q1d = Q1d;
+  if (useCollograd && allCollocated)
+    useCollograd = false;
```
then the output looks like it matches `shared`:
```
$./build/fluids-navierstokes -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-10 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG_strongBC.bin -stg_strong true -ksp_converged_reason -{ts,snes}_monitor -ts_max_steps 3 -snes_rtol .99 -ceed /gpu/hip/shared
0 TS dt 2e-06 time 0.
    0 SNES Function norm 1.116472692852e-01 
    Linear solve converged due to CONVERGED_RTOL iterations 7
    1 SNES Function norm 2.141271691113e-04 
1 TS dt 2e-06 time 2e-06
    0 SNES Function norm 3.370340368714e-02 
    Linear solve converged due to CONVERGED_RTOL iterations 7
    1 SNES Function norm 2.421228116759e-04 
2 TS dt 2e-06 time 4e-06
    0 SNES Function norm 1.847902438292e-02 
    Linear solve converged due to CONVERGED_RTOL iterations 7
    1 SNES Function norm 2.514154676241e-04 
3 TS dt 2e-06 time 6e-06
Test failed with error norm 0.00730937

$ ./build/fluids-navierstokes -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-10 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG_strongBC.bin -stg_strong true -ksp_converged_reason -{ts,snes}_monitor -ts_max_steps 3 -snes_rtol .99 -ceed /gpu/hip/gen
0 TS dt 2e-06 time 0.
    0 SNES Function norm 1.116472692852e-01 
    Linear solve converged due to CONVERGED_RTOL iterations 7
    1 SNES Function norm 2.141271691104e-04 
1 TS dt 2e-06 time 2e-06
    0 SNES Function norm 3.370340368713e-02 
    Linear solve converged due to CONVERGED_RTOL iterations 7
    1 SNES Function norm 2.421228116743e-04 
2 TS dt 2e-06 time 4e-06
    0 SNES Function norm 1.847902438295e-02 
    Linear solve converged due to CONVERGED_RTOL iterations 7
    1 SNES Function norm 2.514154676271e-04 
3 TS dt 2e-06 time 6e-06
Test failed with error norm 0.00730937
```

jeremylt:
The collocated gradient shouldn't come into play though?

nbeams:
I agree it *shouldn't* matter, yet I can't argue with the results that it does seem to...

jeremylt:
This same error doesn't present in Ratel, so why is the fluids bc using the codepath for a collocated gradient?

examples/fluids/src/dirichlet.c
```diff
@@ -0,0 +1,194 @@
...
+    CeedOperatorSetField(op_setup, "x stored", elem_restr_x_stored,
+                         CEED_BASIS_COLLOCATED, x_stored);
+    CeedOperatorSetField(op_setup, "scale", elem_restr_scale, CEED_BASIS_COLLOCATED,
+                         scale_stored);
+
+    // -- Compute geometric factors
+    CeedOperatorApply(op_setup, ceed_data->x_coord, CEED_VECTOR_NONE,
+                      CEED_REQUEST_IMMEDIATE);
+
+    // -- Compute QData for the surface
```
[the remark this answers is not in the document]
jrwrigh: I'm not sure I follow. The only input or output that's at nodes is `ceed_data->x_coord`. Everything else is set as `CEED_BASIS_COLLOCATED`. And for the dirichlet routines we've set the quadrature points to be at the nodes.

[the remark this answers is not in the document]
jeremylt: You are computing qdata at the quadratre points and using it at the nodes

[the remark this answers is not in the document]
jeremylt: `basis_x_sur` goes from the coordinates to the quadratre space, not to the nodal space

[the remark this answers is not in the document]
jrwrigh: Ohhhh. I thought you were talking at the `op_setup` code, not the `op_setup_sur` code. Yeah, poor copy/paste. 

Should be `basis_x_to_q_sur` for `"dx"` and `"weight"` right? Granted we don't use anything from the quadrature weight in this case, so that one probably doesn't matter in this case.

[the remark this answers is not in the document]
jeremylt: The problem is that gradients are not defined for the projection basis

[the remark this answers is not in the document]
jrwrigh: Hmmm. Has it just not been implemented, or is that a permanent issue?

In the meantime, I can replace it with a "normal" basis that goes from points to nodes. Like:

```c
  CeedBasisCreateTensorH1Lagrange(ceed, dim, num_comp_x, 2, P,
                                  CEED_GAUSS_LOBATTO, &ceed_data->basis_xc_sur);
```

jeremylt:
Looking at the gen code, I think the fix might be to put in that patch for now and move up the priority for refactoring that code

examples/fluids/src/dirichlet.c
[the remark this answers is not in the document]
jeremylt: I am not sure of an obvious way to define the gradient of the projection with what we have in each basis?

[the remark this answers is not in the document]
jedbrown: We have $B_p = B_f^\dagger B_c$ and I think $D_p = B_F^\dagger D_c$ is the gradient, in the sense that $B_f D_p \approx D_c$ (exactly when c is a subset of the space f).

jedbrown:
Thanks @nbeams! I pushed this version (applied to both cuda/gen and hip/gen) and it seems to be passing tests for me. I think there is probably still a bug lurking here (or I really don't understand it), but this is enough for us to move forward with the fluids work.

jeremylt:
I think the issue is somehow related to the ordering being different for all fields in the collocated gradient codepath. I think it's a bit too messy to be worth finding before the gen refactor.

jedbrown (approved):
lv-cuda and noether-rocm pass so I think this is good to merge, and unblock #1016

examples/fluids/src/dirichlet.c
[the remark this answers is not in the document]
jedbrown: I'll resolve this thread because it isn't really part of this PR, but feel free to make it into an issue.

PR #1011, opened 2022-06-24
Fluids - Newtonian with Primitive variables

LeilaGhaffari:
This adds primitive variables to the `newtonian` solver, and `channel` and `density_current` problems.


jedbrown (approved):
Does the channel inflow and outflow actually get called anywhere? The `channel.yaml` example is periodic in x, so it won't get used there. It could be interesting to run non-periodic channel flow, but I think we don't have any tests of that so no idea if the code we have (for conservative OR primitive) is correct.

examples/fluids/qfunctions/channel.h
```diff
@@ -64,6 +64,34 @@ CEED_QFUNCTION_HELPER CeedInt Exact_Channel(CeedInt dim, CeedScalar time,
...
+  const CeedScalar Ec    = (umax*umax) / (cp*theta0);
+  const CeedScalar theta = theta0*(1 + (Pr*Ec/3)
+                                   * (1 - Square(Square((y-center)/H))));
+  q[0] = P0;
+  q[1] = umax*(1 - Square((y-center)/H));
+  q[2] = 0.;
+  q[3] = 0.;
+  q[4] = theta;
+
+  return 0;
```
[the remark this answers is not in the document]
LeilaGhaffari: Yes, that makes sense. I'll make that change.
Regarding the in/outflow, what do you think of dropping the primitive versions for now if the conservative versions are not tested yet?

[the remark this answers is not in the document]
jedbrown: I think that's fine. I'd like to make all steady inflow (and outflow to the extent relevant) qfunctions just use data anyway, which could come from "imprinting" the initial condition or from sampling a prescribed function. But we should only need one inflow qfunction for conservative and one for primitive. My inclination would be to prescribe state always in terms of primitive variables (since that's a better conditioned way to represent state) and I think the inflow/outflow qfunctions can test the upstream state to determine whether it's sub- or sonic. That would be a future PR.

LeilaGhaffari:
Can we `squash-merge` if the last commit looks good, @jedbrown ?

examples/fluids/qfunctions/densitycurrent.h
```diff
@@ -148,4 +148,81 @@ CEED_QFUNCTION(ICsDC)(void *ctx, CeedInt Q,
...
+
+  // -- Exner pressure, hydrostatic balance
+  const CeedScalar Pi = 1. + g*g*(exp(-N*N*z/g) - 1.) / (cp*theta0*N*N);
+
+  // Initial Conditions
+  q[0] = Pi;
+  q[1] = 0.0;
+  q[2] = 0.0;
+  q[3] = 0.0;
+  q[4] = theta;
```
jedbrown, line 195: Here also, could we have a helper that returns `State` and then `ICsDC_Cons` and `ICsDC_Prim` would just unpack the variables it needs?

examples/fluids/src/setupdm.c
```diff
@@ -97,11 +97,12 @@ PetscErrorCode SetUpDM(DM dm, ProblemData *problem, PetscInt degree,
...
-    // Empty name for conserved field (because there is only one field)
-    PetscSection section;
-    ierr = DMGetLocalSection(dm, &section); CHKERRQ(ierr);
-    ierr = PetscSectionSetFieldName(section, 0, ""); CHKERRQ(ierr);
+
+  // Empty name for conserved field (because there is only one field)
+  PetscSection section;
+  ierr = DMGetLocalSection(dm, &section); CHKERRQ(ierr);
+  ierr = PetscSectionSetFieldName(section, 0, ""); CHKERRQ(ierr);
+  if (!phys->primitive) {
```
jedbrown, line 105: Stylistically, I think it's more clear to say `if (cond) {} else {}` rather than `if (!cond) {} else {}` when there isn't a strong reason otherwise.

jedbrown (commented):
Looking pretty good. Just a few minor bits.

examples/fluids/problems/newtonian.c
```diff
@@ -140,14 +143,25 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
...
+  ierr = PetscOptionsBool("-primitive", "Use primitive variables",
+                          NULL, prim_var=PETSC_FALSE, &prim_var, NULL); CHKERRQ(ierr);
+  if (prim_var) {
+    problem->ics.qfunction                     = ICsNewtonianIG_Prim;
+    problem->ics.qfunction_loc                 = ICsNewtonianIG_Prim_loc;
+    problem->apply_vol_ifunction.qfunction     = IFunction_Newtonian_Prim;
+    problem->apply_vol_ifunction.qfunction_loc = IFunction_Newtonian_Prim_loc;
+    problem->apply_vol_ijacobian.qfunction     = IJacobian_Newtonian_Prim;
+    problem->apply_vol_ijacobian.qfunction_loc = IJacobian_Newtonian_Prim_loc;
+  }
```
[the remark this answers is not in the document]
LeilaGhaffari: Oh, right! When I was coding this up, my main concern was getting some results for my talk as quickly as possible and meant to revisit this. It's fixed now, I think!

LeilaGhaffari:
Thanks, @jedbrown . I tried to address them all. Could you give a look again? I can refactor the traveling vortex and add the primitive-variables version after this and [#995](https://github.com/CEED/libCEED/pull/995) is merged.

examples/fluids/qfunctions/channel.h
```diff
@@ -29,46 +30,70 @@ struct ChannelContext_ {
...
+  Y[4] = theta;
+
+  return StateFromY(gas, Y, X);
+}
 
-  const CeedScalar p = P0;
+// *****************************************************************************
+// These QFunctions sets the initial condition
+// *****************************************************************************
+CEED_QFUNCTION(ICsChannel_Prim)(void *ctx, CeedInt Q,
```
jrwrigh, line 70: Since `ExactChannel` is outputing a `State`, I think it makes sense to add primitive to the newtonian context and have a single ICs Channel, and the output of it is either `s.Y.*` or `s.U.*`.

examples/fluids/qfunctions/channel.h
```diff
@@ -29,46 +30,70 @@ struct ChannelContext_ {
...
+    State s = Exact_Channel(3, 0., x, 5, ctx);
+    CeedScalar q[5] = {0.};
+    q[0] = s.Y.pressure;
+    for (CeedInt j=0; j<3; j++)
+      q[j+1] = s.Y.velocity[j];
+    q[4] = s.Y.temperature;
 
+    for (CeedInt j=0; j<5; j++)
+      q0[j][i] = q[j];
+  } // End of Quadrature Point Loop
```
jrwrigh, line 91: Probably just condense into a single loop?
```suggestion
    q0[0][i] = s.Y.pressure;
    for (CeedInt j=0; j<3; j++)
      q0[j+1][i] = s.Y.velocity[j];
    q0[4][i] = s.Y.temperature;
  } // End of Quadrature Point Loop
```

examples/fluids/qfunctions/channel.h
```diff
@@ -79,11 +104,16 @@ CEED_QFUNCTION(ICsChannel)(void *ctx, CeedInt Q,
...
     CeedScalar q[5] = {0.};
-    Exact_Channel(3, 0., x, 5, q, ctx);
+    q[0] = s.U.density;
+    for (CeedInt j=0; j<3; j++)
+      q[j+1] = s.U.momentum[j];
+    q[4] = s.U.E_total;
 
     for (CeedInt j=0; j<5; j++)
       q0[j][i] = q[j];
+
```
jrwrigh, line 116: Can similarly contract into a single for loop.

examples/fluids/qfunctions/densitycurrent.h
```diff
@@ -108,25 +121,23 @@ CEED_QFUNCTION_HELPER int Exact_DC(CeedInt dim, CeedScalar time,
...
+  return StateFromY(gas, Y, X);
 }
 
 // *****************************************************************************
-// This QFunction sets the initial conditions for density current
+// These QFunctions set the initial conditions for density current
 // *****************************************************************************
-CEED_QFUNCTION(ICsDC)(void *ctx, CeedInt Q,
-                      const CeedScalar *const *in, CeedScalar *const *out) {
+CEED_QFUNCTION(ICsDC_Prim)(void *ctx, CeedInt Q,
```
jrwrigh, line 139: Could also use a `DensityCurrentContext->newtonian_ctx.primitive` to make this a single ICs.

jrwrigh (approved):
Looks good to me. Just some minor suggestions.

LeilaGhaffari:
Done! Thanks, @jrwrigh .

examples/fluids/qfunctions/densitycurrent.h
```diff
@@ -185,17 +159,19 @@ CEED_QFUNCTION(ICsDC_Cons)(void *ctx, CeedInt Q,
...
+      for (CeedInt j=0; j<3; j++)
+        q0[j+1][i] = s.Y.velocity[j];
+      q0[4][i] = s.Y.temperature;
+    } else {
+      q0[0][i] = s.U.density;
+      for (CeedInt j=0; j<3; j++)
+        q0[j+1][i] = s.U.momentum[j];
+      const CeedScalar Pi = 1. + g*g*(exp(-N*N*x[2]/g) - 1.) / (cp*theta0*N*N);
+      // Multiply internal energy by Exner pressure (Pi)
+      q0[4][i] = s.U.density * (cv*s.Y.temperature*Pi + g*x[2]);
```
jedbrown, line 173: I believe the `theta` defined earlier is the "potential temperature".  The absolute temperature is the product with the Exner function $T = \pi\theta$. But `s.Y.temperature` is meant to be absolute temperature. Ideally we can just use `s.U.E_total` on this line.

jedbrown (approved):
Looks great with one consistency fix for density current.

examples/fluids/qfunctions/densitycurrent.h
LeilaGhaffari: Alright, just fixed it. Needed to update pressure as well. It does pass locally now but I don't know why I don't see the failures on the CI!

jedbrown: Hmm, I manually started a pipeline here. https://gitlab.com/libceed/libCEED/-/pipelines/590481550

examples/fluids/qfunctions/densitycurrent.h
```diff
@@ -108,18 +121,16 @@ CEED_QFUNCTION_HELPER int Exact_DC(CeedInt dim, CeedScalar time,
...
   // Initial Conditions
-  q[0] = rho;
-  q[1] = 0.0;
-  q[2] = 0.0;
-  q[3] = 0.0;
-  q[4] = rho * (cv*theta*Pi + g*z);
-
-  return 0;
+  CeedScalar Y[5] = {0.};
+  Y[0] = P0 * Pi * pow(Pi, cv/Rd);
```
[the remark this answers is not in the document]
LeilaGhaffari: That makes sense! Just pushed the update!

jedbrown:
Thanks, @LeilaGhaffari. This is great.

PR #1012, opened 2022-06-24
basis - use ncomp from 'from' basis for projection

jeremylt:
@jedbrown @jrwrigh I actually meant for this to skip the need for a dummy basis of the right number of components but reversed the order.

### DIFF
--- a/include/ceed/ceed.h
+++ b/include/ceed/ceed.h
@@ -552,8 +552,8 @@ CEED_EXTERN int CeedBasisCreateHdiv(Ceed ceed, CeedElemTopology topo,
                                     const CeedScalar *div,
                                     const CeedScalar *q_ref,
                                     const CeedScalar *q_weights, CeedBasis *basis);
-CEED_EXTERN int CeedBasisCreateProjection(CeedBasis basis_to, CeedBasis basis_from, CeedBasis *basis_project);
-CEED_EXTERN int CeedBasisCreateProjectionMatrix(CeedBasis basis_to, CeedBasis basis_from, CeedScalar **interp_project);
+CEED_EXTERN int CeedBasisCreateProjection(CeedBasis basis_from, CeedBasis basis_to, CeedBasis *basis_project);
+CEED_EXTERN int CeedBasisCreateProjectionMatrix(CeedBasis basis_from, CeedBasis basis_to, CeedScalar **interp_project);
 CEED_EXTERN int CeedBasisReferenceCopy(CeedBasis basis, CeedBasis *basis_copy);
 CEED_EXTERN int CeedBasisView(CeedBasis basis, FILE *stream);
 CEED_EXTERN int CeedBasisApply(CeedBasis basis, CeedInt num_elem,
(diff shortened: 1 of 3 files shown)


jedbrown:
Can you update the docs to explain this semantic (perhaps via an example)?

jeremylt:
Updated.

jrwrigh:
Also, maybe while the function is new and not super used, but is there any reason the function arguments are in their specific order? I feel like having it be in `from -> to` order makes more intuitive sense than how it is now. Even looking at the use case `PetscCall(CeedBasisCreateProjection(basis_u_face, basis_x_face, &basis_x_to_u_face));`, the name of the projection basis is different than the order of the arguments.

Figured I might as well mention it now before too many things depend on it and it becomes more difficult to change the interface.

jeremylt:
It's new, but its also not. I mirrored the order of the arguments in the multigrid operator interface.
```C
int CeedOperatorMultigridLevelCreate(CeedOperator op_fine,
                                     CeedVector p_mult_fine,
                                     CeedElemRestriction rstr_coarse, CeedBasis basis_coarse,
                                     CeedOperator *op_coarse, CeedOperator *op_prolong,
                                     CeedOperator *op_restrict)
```

jrwrigh:
I'm not sure I see the connection. The setup for multigrid seems like the argument order is ambivalent; you have two operators that you care about going in opposite directions, prolongation and restriction. Where as the projection function is only creating an operator going in one direction, and a left-to-right precedent seems to make more sense.

jeremylt:
`basis_to` comes from `op_fine` and `basis_from` comes from `op_coarse`. The multigrid interface starts with `_to` and builds the `_from`.

jrwrigh:
> I'm just explaining how I set it up to match the multigrid interface

Same here. I recognize I've only been working in the code base for a few months (and only a small fraction of the code base at that). On one hand, I have a perspective of an outsider trying to use the code, which I think can be useful. On the other, I also don't have the experience to recognize a bad design decision, especially in areas of the code I haven't touched (ie. multigrid). 

I'd say I'm still in favor of swapping them, but with the above caveats to my opinion.

jedbrown:
I'm just bikeshedding at this point, but `DMCreateInterpolation` is also of the `(from, to, &result)` pattern and I think that's slightly more intuitive.

include/ceed/ceed.h
```diff
@@ -552,8 +552,8 @@ CEED_EXTERN int CeedBasisCreateHdiv(Ceed ceed, CeedElemTopology topo,
                                     const CeedScalar *div,
                                     const CeedScalar *q_ref,
                                     const CeedScalar *q_weights, CeedBasis *basis);
-CEED_EXTERN int CeedBasisCreateProjection(CeedBasis basis_to, CeedBasis basis_from, CeedBasis *basis_project);
-CEED_EXTERN int CeedBasisCreateProjectionMatrix(CeedBasis basis_to, CeedBasis basis_from, CeedScalar **interp_project);
+CEED_EXTERN int CeedBasisCreateProjection(CeedBasis basis_from, CeedBasis basis_to, CeedBasis *basis_project);
+CEED_EXTERN int CeedBasisCreateProjectionMatrix(CeedBasis basis_from, CeedBasis basis_to, CeedScalar **interp_project);
```
jeremylt, line 556: One of these days we should dump the variable names in the header files since they don't actually matter.

jrwrigh: They can be helpful as function signatures when using a language server. Sometimes it's obvious what the argument should be from the type (`DM`, `Ceed`, etc.), but other times it's nice to know.

jeremylt: But absolutely nothing is enforcing the correctness of the names in the header.

jedbrown: I remember that debate. :laughing:

jedbrown (approved):

PR #1013, opened 2022-06-27
Fix bad indexing in 3d poission det

jeremylt:
Oops


jeremylt:
Exactly. I pulled it over to Ratel and it popped up right away on our Tet tests

PR #1014, opened 2022-06-29
Verify ctx read-only access

jeremylt:
It occurred to me that it would be easy to optionally verify that the context was not written into.


jedbrown (approved):
Thanks for this.

PR #1015, opened 2022-06-30
fluids: Fix post-processing in DMPlexInsertBoundaryValues_NS

jrwrigh:
From this comment: https://github.com/CEED/libCEED/commit/b80a4419c8f00e2f8e9a7afd897117055f7f7a73#r76899317

Basically don't zero the full `Q_loc` vector so that it can be correctly used by `VecView`.

### DIFF
--- a/examples/fluids/src/misc.c
+++ b/examples/fluids/src/misc.c
@@ -98,18 +98,20 @@ PetscErrorCode ICs_FixMultiplicity(DM dm, CeedData ceed_data, User user,
   PetscFunctionReturn(0);
 }
 
-
-// Note: The BCs must be inserted *before* the other values are inserted into Q_loc
 PetscErrorCode DMPlexInsertBoundaryValues_NS(DM dm,
     PetscBool insert_essential, Vec Q_loc, PetscReal time, Vec face_geom_FVM,
     Vec cell_geom_FVM, Vec grad_FVM) {
 
-  Vec            Qbc;
+  Vec            Qbc, boundary_mask;
   PetscErrorCode ierr;
   PetscFunctionBegin;
 
+  // Mask (zero) Dirichlet entries
+  PetscCall(DMGetNamedLocalVector(dm, "boundary mask", &boundary_mask));
+  PetscCall(VecPointwiseMult(Q_loc, Q_loc, boundary_mask));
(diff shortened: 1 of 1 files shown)


examples/fluids/src/misc.c
```diff
@@ -98,18 +98,21 @@ PetscErrorCode ICs_FixMultiplicity(DM dm, CeedData ceed_data, User user,
   PetscFunctionReturn(0);
 }
 
-
 // Note: The BCs must be inserted *before* the other values are inserted into Q_loc
```
jeremylt, line 101: We can delete this note now

LeilaGhaffari (approved):
LGTM! Thanks for the fix!

examples/fluids/src/misc.c
```diff
@@ -260,6 +264,18 @@ PetscErrorCode SetBCsFromICs_NS(DM dm, Vec Q, Vec Q_loc) {
                                     "DMPlexInsertBoundaryValues_C", DMPlexInsertBoundaryValues_NS);
   CHKERRQ(ierr);
 
+  PetscCall(DMHasNamedLocalVector(dm, "boundary mask", &boundary_mask_created));
+  if (!boundary_mask_created) {
```
jeremylt, line 268: We may have to think a bit more about this - In what situation would this mask already exist and be valid for this boundary configuration?

jrwrigh: I mainly put it there for the dirichlet branch, which currently creates it in a different location. Though they shouldn't be called at the same time and `SetBCsFromICs_NS` would overwrite `DMPlexInsertBoundaryValues` from what the dirichlet function would write anyways.

jrwrigh: Arguably `SetBCsFromICs` should either be merged in with the Dirichlet function when that gets merged in anyways.

jrwrigh: Regardless, I'll go ahead and remove that line. I was being a bit over-cautious I think.

examples/fluids/src/misc.c
jrwrigh: Done

examples/fluids/src/misc.c
jrwrigh: Removed the if statement.

jeremylt (approved):
LGTM, thanks

jedbrown:
Thanks!

PR #1016, opened 2022-07-05
examples/fluids: use DMGetCellCoordinateDM to handle periodicity

jedbrown:
Note that without -dm_sparse_localize 0 (default=1), the cell DM will
only have cells with localized coordinates (and there is no vector
representation that contains both).

### DIFF
--- a/examples/fluids/src/setupdm.c
+++ b/examples/fluids/src/setupdm.c
@@ -25,6 +25,7 @@ PetscErrorCode CreateDM(MPI_Comm comm, ProblemData *problem,
 
   // Set Tensor elements
   ierr = PetscOptionsSetValue(NULL, "-dm_plex_simplex", "0"); CHKERRQ(ierr);
+  ierr = PetscOptionsSetValue(NULL, "-dm_sparse_localize", "0"); CHKERRQ(ierr);
   // Set CL options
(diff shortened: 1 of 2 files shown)


jeremylt:
It looks like this requires an update to Noether for the CI to go through. Which version of PETSc main is 'safe' for our use?

jeremylt:
I'm getting this error testing manually on Noether:
```
# Test Name: blasius_STG_strongBC
# TESTARGS: -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-10 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG_strongBC.bin -stg_strong true
ok 7 fluids-navierstokes /cpu/self
not ok 8 fluids-navierstokes /cpu/self stdout
# + Test failed with error norm 0.0125345
ok 9 fluids-navierstokes /cpu/self stderr
```

jedbrown:
That's because we fixed a bug in PETSc in handling essential boundary conditions. @jrwrigh has fixed output in #1009 that matches the new PETSc code (and that's how we discovered the problem in PETSc).

jedbrown:
Rebased, and I think this can be merged once tests pass.

jedbrown:
I updated noether's petsc. @jeremylt Can you update lv?

PR #1018, opened 2022-07-07
fluids: Precompute 1 / Ektot for STG

jrwrigh:
Precompute $( \sum E(\kappa) \Delta \kappa) ^{-1}$ to ease calculation of spectrum coefficients $q^n$.

General framework is also setup to possibly hold other pre-computed values for use in STG.


jedbrown (approved):
I have no obvious concerns. Can you test this at least on Noether with a representative problem size (Noether's GPU has 16 GB of device memory) to see if it has a performance impact?

jrwrigh:
Comparing `SNESFunctionEval` time, we're seeing less than a 10% reduction in time (1.53 vs 1.442 s). However, that only makes up ~1% of the total solve time though, so not a significant performance improvement overall.

jedbrown:
Good. It may be that the inflow boundary isn't even a large fraction of the cost now that it's on the GPU. I still think this change is worthwhile.

jedbrown:
Okay to merge IMO.

PR #1021, opened 2022-07-07
Change the initialization logic for `useCollograd`.

YohannDudouit:
Avoid using the collocated gradient algorithm when we don't compute any gradient.


jedbrown (approved):

jeremylt (approved):
This same change probably also goes in `backends/hip-gen`, right?

YohannDudouit:
> This same change probably also goes in `backends/hip-gen`, right?

Yes, if you're fine with the change I can spend some time doing it in `hip-gen`, it should be quick. @nbeams any objection?

YohannDudouit:
> This same change probably also goes in `backends/hip-gen`, right?

@jeremylt Change propagated to `hip-gen`.

jeremylt:
Thanks - we can merge once CI does its thing

jeremylt:
This change seems to reintroduce the bug we saw in the fluids example, according to the CI

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -1130,16 +1140,14 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
...
       if (useCollograd) {
         //Accumulator for gradient slices
-        code << "    CeedScalar r_tt"<<i<<"[ncomp_out_"<<i<<"*Q1d];\n";
         code << "    for (CeedInt i = 0; i < ncomp_out_"<<i<<"; ++i) {\n";
         code << "      for (CeedInt j = 0; j < Q1d; ++j) {\n";
         code << "        r_tt"<<i<<"[j + i*Q1d] = 0.0;\n";
         code << "      }\n";
         code << "    }\n";
-      } else {
-        code << "    CeedScalar r_tt"<<i<<"[ncomp_out_"<<i<<"*Dim*Q1d];\n";
```
nbeams, line 1142: It looks like these changes have not yet made it into `hip-gen` -- could this be the reason for the failing fluids test on CI?

YohannDudouit: I don't think this should create issues, the size of the array is just bigger than needed if I'm not mistaken. I'll take a closer look at the Hip code.

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -814,12 +814,25 @@ extern "C" int CeedHipGenOperatorBuild(CeedOperator op) {
 
   // Find dim and Q1d
   bool useCollograd = true;
```
YohannDudouit, line 816: That's the error:
```suggestion
  bool useCollograd = false;
```

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
YohannDudouit: I propagated the change in [9781974](https://github.com/CEED/libCEED/pull/1021/commits/97819745621ba5e592629ee16c7baa04faf77f86).

jeremylt:
CI got suck, but the PETSc tests still seem to be failing.

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -1130,16 +1140,14 @@ extern "C" int CeedHipGenOperatorBuild(CeedOperator op) {
...
         //Accumulator for gradient slices
-        code << "    CeedScalar r_tt"<<i<<"[ncomp_out_"<<i<<"*Q1d];\n";
         code << "    for (CeedInt i = 0; i < ncomp_out_"<<i<<"; ++i) {\n";
         code << "      for (CeedInt j = 0; j < Q1d; ++j) {\n";
         code << "        r_tt"<<i<<"[j + i*Q1d] = 0.0;\n";
         code << "      }\n";
         code << "    }\n";
-      } else {
-        code << "    CeedScalar r_tt"<<i<<"[ncomp_out_"<<i<<"*Dim*Q1d];\n";
       }
```
nbeams, line 1151: Glad that CUDA and HIP are now the same, but the more I think about this, the less I understand this change. 
Why does it no longer matter whether we are using collograd or not when determining the size of the `r_tt` array?  If we used to need the extra `Dim` factor in non-collograd case before, why not now?

YohannDudouit: I'm not completely sure about this change either... I'll revert I think.

YohannDudouit: Reverted in [fe667b8](https://github.com/CEED/libCEED/pull/1021/commits/fe667b84b42c45b322bb0a9d91af5da647d51611).

jeremylt:
The CUDA CI flaked, but the HIP CI is happy now.

YohannDudouit:
> The CUDA CI flaked, but the HIP CI is happy now.

Should I run `make test` on Lassen?

nbeams:
Does the setup on Lassen do the PETSc tests?

jeremylt:
This has identical changes in CUDA and HIP - CUDA CI is borked but HIP CI is passing so I think we are ok to merge

YohannDudouit:
> This has identical changes in CUDA and HIP - CUDA CI is borked but HIP CI is passing so I think we are ok to merge

@jeremylt What would I need to do to reproduce the CI tests locally?

jedbrown:
I merged from main and kicked off a pipeline. Assuming it's good, I think we can squash-merge.

PR #1022, opened 2022-07-07
Support non-tensor elements with `cuda/gen` backend.

YohannDudouit:
Try using `dim == 1` for non-tensor basis to add support for non-tensor elements in the `cuda/gen` backend.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -933,7 +933,13 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
     if (emode != CEED_EVAL_WEIGHT) {
       ierr = CeedOperatorFieldGetBasis(opinputfields[i], &basis); CeedChkBackend(ierr);
       if (basis != CEED_BASIS_COLLOCATED) {
-        ierr = CeedBasisGetNumNodes1D(basis, &P1d); CeedChkBackend(ierr);
+        bool isTensor;
+        ierr = CeedBasisIsTensor(basis, &isTensor); CeedChkBackend(ierr); 
+        if (isTensor) {
+          ierr = CeedBasisGetNumNodes1D(basis, &P1d); CeedChkBackend(ierr);
+        } else {
+          ierr = CeedBasisGetNumNodes(basis, &P1d); CeedChkBackend(ierr);
+        }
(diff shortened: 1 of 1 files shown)


jedbrown:
Thanks for this.

So I'd expect this to be tested, but I see the noether-rocm test output is skipping the same number of tests from before
https://gitlab.com/libceed/libCEED/-/pipelines/582787305/test_report
image
as the new https://gitlab.com/libceed/libCEED/-/pipelines/582875782/test_report
image
Does this make sense or should fewer tests be skipped? Cc: @jeremylt

jedbrown:
Oh, probably the issue above is that this change wasn't propagated to HIP (tested on Noether). Meanwhile, lv continuous to have stability issues and we're still waiting on a CUDA GPU for Noether to arrive, at which point we should have more reliable testing.

YohannDudouit:
@jedbrown I would need to work a bit more than that to make this work, it should be doable, but is is not ready (even for cuda-gen).

jedbrown:
Our CUDA CI should be reliable now. What would be needed to finish this branch, at least for cuda/gen?

jeremylt:
I think this might be easier to get to with #1050?

jedbrown:
Perhaps, but #1050 has possibly expansive scope so I don't know if it's feasible to finish for v0.11.

jedbrown:
How much is needed here now that #1050 has merged?

jeremylt:
I think a fair amount still needs to be done.

YohannDudouit:
So, the main issue is with the gradient, it's not just "1D" in that case, technically it's not much work, but it might be better to do something cleaner than this hack.

jeremylt:
With the refactor, we can now write and test the nontensor kernels for the shared backend and then incorporate them into the gen backend.

YohannDudouit:
@jeremylt If we follow this path we will still have to modify the gen backend, the "tensor" assumption is fairly hardcoded there a bit everywhere. I think the easiest would probably be to write a gen path for non-tensor operators (distinct from the current tensor path) using the non-tensor functions.

jeremylt:
Either way we need to write a new basis kernel.  It seems simpler to test that on its own instead of testing the new kernel + the changes to the gen logic all at the same time.

YohannDudouit:
I totally agree.

jeremylt:
Closing in favor of https://github.com/CEED/libCEED/pull/1735 (very similar approach after refactoring the code generation functions)

PR #1025, opened 2022-07-08
basis - fix pointer pointer mistake

jeremylt:


jedbrown (approved):
This kind of error is an argument in favor of using a local variable and assigning it to the output arg in one place (rather than using the output variable throughout).

PR #1029, opened 2022-07-18
fluids: Make Strong STG compatible with primitive variables.

jrwrigh:
Still need to document, but otherwise should be good to go.


jrwrigh:
Left is conservative variables, right is primitive. This is 40 timesteps into the simulation.
```
./navierstokes -options_file blasius.yaml -{ts,snes}_monitor -{snes,ksp}_converged_reason -output_freq 1 -stg_strong -stg_use -stg_mean_only
```
You can see that the inflow is still part way pushing it's way through the domain.

Velocity Y
Pressure
Temperature

Note for pressure, the ranges are different, but they both span approximately 10 Pa

jrwrigh:
Don't merge yet, I discovered some unchanged state functions in PressureOutflow and BoundaryIntegral QFs.

jedbrown:
Good merge when ready.

PR #1032, opened 2022-07-28
Allow empty composite operators

jeremylt:
We allow operators with empty restrictions. Similarly, we should allow composite operators with empty suboperator lists.

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -1006,18 +1006,20 @@ int CeedOperatorCheckReady(CeedOperator op) {
 
   CeedQFunction qf = op->qf;
   if (op->is_composite) {
-    if (!op->num_suboperators)
-      // LCOV_EXCL_START
-      return CeedError(ceed, CEED_ERROR_INCOMPLETE, "No sub_operators set");
-    // LCOV_EXCL_STOP
-    for (CeedInt i = 0; i < op->num_suboperators; i++) {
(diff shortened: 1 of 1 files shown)


interface/ceed-operator.c
```diff
@@ -1006,18 +1006,20 @@ int CeedOperatorCheckReady(CeedOperator op) {
...
-    if (!op->num_suboperators)
-      // LCOV_EXCL_START
-      return CeedError(ceed, CEED_ERROR_INCOMPLETE, "No sub_operators set");
-    // LCOV_EXCL_STOP
-    for (CeedInt i = 0; i < op->num_suboperators; i++) {
-      ierr = CeedOperatorCheckReady(op->sub_operators[i]); CeedChk(ierr);
+    if (!op->num_suboperators) {
+      // Empty operator setup
+      op->input_size = 0;
+      op->output_size = 0;
```
jedbrown, line 1012: I assume this won't cause a size mismatch in a check if applied to a vector with nonzero size?

jedbrown (approved):

interface/ceed-operator.c
jeremylt: Only regular operators check vector sizes when applying the operator. A composite operator only checks for compatibility between sub operator active vector sizes when new operators are added or the composite operator is flagged as 'setup'.

I needed to update these fields because they are also used to create vectors that match the operator input/output vector sizes. The default value when no sub operators have been set is -1.

PR #1034, opened 2022-07-29
Fixed dXdx definition in setupgeo2d.h

rezgarshakeri:


jeremylt:
We really need some text tests for these sorts of issues. Also, we should probably make better use of gallery operators where we can outside of the raw petsc examples.

rezgarshakeri:
> We really need some tet tests for these sorts of issues. Also, we should probably make better use of gallery operators where we can outside of the raw petsc examples.

Yeah, I think 2D example uses uniform mesh that's why they are fine. It would be great to have random mesh (non-affine). I have it in my Hdiv branch.

jedbrown (approved):
In addition to tests, we should have a helper function to compute 2x2 and 3x3 inverses because they show up in enough places and it would be easier to unit test a helper than the full qfunction.

rezgarshakeri:
> 

I have it in qfunction/utils.h in my Hdiv branch, someone can get it from there.

jeremylt:
I think we should package a bunch of QFunction helpers like we do the Gallery QFs and their JiT source. I can dig into this next week.

PR #1035, opened 2022-07-31
fluids: Unify Newtonian Primtive functions

jrwrigh:
- Combines the separate primitive QFunctions in newtonian.h into a single function call. 
- Replaces `StateFromQi*` function pointers with QF helper functions using if statements inside.
- Includes various other refactors and stylings.

Stripped this out from #1031 while that is a bit stalled for the moment. Figured it might as well be it's own PR

### DIFF
--- a/examples/fluids/navierstokes.h
+++ b/examples/fluids/navierstokes.h
@@ -194,7 +194,7 @@ struct Physics_private {
   EulerTestType            euler_test;
   StabilizationType        stab;
   PetscBool                implicit;
-  PetscBool                primitive;
+  PetscBool                use_primitive;
   PetscBool                has_curr_time;
   PetscBool                has_neumann;
   CeedContextFieldLabel    solution_time_label;
--- a/examples/fluids/problems/newtonian.c
+++ b/examples/fluids/problems/newtonian.c
@@ -77,7 +77,7 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
   MPI_Comm          comm = PETSC_COMM_WORLD;
   PetscBool         implicit;
   PetscBool         has_curr_time = PETSC_FALSE,
-                    prim_var, unit_tests;
+                    use_primitive, unit_tests;
   PetscInt          ierr;
   NewtonianIdealGasContext newtonian_ig_ctx;
   CeedQFunctionContext newtonian_ig_context;
@@ -137,8 +137,8 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
                     NULL);
   // -- Conservative vs Primitive variables
   ierr = PetscOptionsBool("-primitive", "Use primitive variables",
-                          NULL, prim_var=PETSC_FALSE, &prim_var, NULL); CHKERRQ(ierr);
-  if (prim_var) {
+                          NULL, use_primitive=PETSC_FALSE, &use_primitive, NULL); CHKERRQ(ierr);
+  if (use_primitive) {
     problem->ics.qfunction                        = ICsNewtonianIG_Prim;
     problem->ics.qfunction_loc                    = ICsNewtonianIG_Prim_loc;
     problem->apply_vol_ifunction.qfunction        = IFunction_Newtonian_Prim;
@@ -231,7 +231,7 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
                        "Warning! Use -stab supg only with -implicit\n");
     CHKERRQ(ierr);
   }
-  if (prim_var && !implicit) {
+  if (use_primitive && !implicit) {
     SETERRQ(comm, PETSC_ERR_ARG_NULL,
             "RHSFunction is not provided for primitive variables (use -primitive only with -implicit)\n");
   }
@@ -279,7 +279,7 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
   // -- Solver Settings
   user->phys->stab          = stab;
   user->phys->implicit      = implicit;
-  user->phys->primitive     = prim_var;
+  user->phys->use_primitive = use_primitive;
   user->phys->has_curr_time = has_curr_time;
 
   // -- QFunction Context
@@ -296,7 +296,7 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
   newtonian_ig_ctx->Ctau_E        = Ctau_E;
   newtonian_ig_ctx->stabilization = stab;
   newtonian_ig_ctx->is_implicit   = implicit;
-  newtonian_ig_ctx->is_primitive  = prim_var;
+  newtonian_ig_ctx->use_primitive = use_primitive;
   ierr = PetscArraycpy(newtonian_ig_ctx->g, g, 3); CHKERRQ(ierr);
 
   CeedQFunctionContextCreate(user->ceed, &problem->ics.qfunction_context);
--- a/examples/fluids/problems/newtonian.c
+++ b/examples/fluids/problems/newtonian.c
@@ -159,10 +159,10 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
     problem->ics.qfunction_loc                    = ICsNewtonianIG_loc;
     problem->apply_vol_rhs.qfunction              = RHSFunction_Newtonian;
     problem->apply_vol_rhs.qfunction_loc          = RHSFunction_Newtonian_loc;
-    problem->apply_vol_ifunction.qfunction        = IFunction_Newtonian;
-    problem->apply_vol_ifunction.qfunction_loc    = IFunction_Newtonian_loc;
-    problem->apply_vol_ijacobian.qfunction        = IJacobian_Newtonian;
-    problem->apply_vol_ijacobian.qfunction_loc    = IJacobian_Newtonian_loc;
+    problem->apply_vol_ifunction.qfunction        = IFunction_Newtonian_Conserv;
+    problem->apply_vol_ifunction.qfunction_loc    = IFunction_Newtonian_Conserv_loc;
+    problem->apply_vol_ijacobian.qfunction        = IJacobian_Newtonian_Conserv;
+    problem->apply_vol_ijacobian.qfunction_loc    = IJacobian_Newtonian_Conserv_loc;
     problem->apply_inflow.qfunction               = BoundaryIntegral_Conserv;
     problem->apply_inflow.qfunction_loc           = BoundaryIntegral_Conserv_loc;
     problem->apply_inflow_jacobian.qfunction      = BoundaryIntegral_Jacobian_Conserv;
--- a/examples/fluids/problems/newtonian.c
+++ b/examples/fluids/problems/newtonian.c
@@ -138,6 +138,7 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
   // -- Conservative vs Primitive variables
   ierr = PetscOptionsBool("-primitive", "Use primitive variables",
                           NULL, use_primitive=PETSC_FALSE, &use_primitive, NULL); CHKERRQ(ierr);
+  // *INDENT-OFF*
   if (use_primitive) {
     problem->ics.qfunction                        = ICsNewtonianIG_Prim;
     problem->ics.qfunction_loc                    = ICsNewtonianIG_Prim_loc;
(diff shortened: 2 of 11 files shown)


jedbrown:
Thanks, this looks good from a coding standpoint. Can you do a basic test to identify whether there is any performance implications to this on CPU or GPU?

examples/fluids/navierstokes.h
```diff
@@ -194,7 +194,7 @@ struct Physics_private {
   EulerTestType            euler_test;
   StabilizationType        stab;
   PetscBool                implicit;
```
jeremylt, line 196: It isn't pressing or anything, but this PR highlights a few places where we aren't using the `PetscBool verb_description` style for our booleans in fluids. I'm sure we're get them all eventually.

jrwrigh: Yeah, I've been trying to fix them when I touch them.

jrwrigh:
We're actually seeing a more significant performance difference that I was expecting. Running
```
./build/fluids-navierstokes -ceed /gpu/hip -options_file examples/fluids/channel.yaml -dm_plex_box_faces 20,20,1 -ts_max_steps 20 -output_freq 0
```
I'm seeing 3% slower for CPU (19.15 -> 19.75 seconds) and 0.5% for HIP on Noether (59.6 -> 59.9). This is trend is consistent after running a few more times.

I figured the branch prediction on the CPU would prevent it from being a significant factor there and any performance difference would be more evident on the GPU (though it very well could just be that at this problem size the consequential speed decrease is masked by other slow-downs).

Any idea why this is consequential by the case statements for the stabilization are (presumably) not? Or is there a way to tell the compiler that branch will be identical through a run (though I'd be surprised if there was a way to do that portably)?

examples/fluids/qfunctions/newtonian.h
```diff
@@ -254,54 +222,48 @@ CEED_QFUNCTION(IFunction_Newtonian)(void *ctx, CeedInt Q,
...
   CeedPragmaSIMD
   // Quadrature Point Loop
   for (CeedInt i=0; i<Q; i++) {
-    CeedScalar U[5];
-    for (CeedInt j=0; j<5; j++) U[j] = q[j][i];
+    CeedScalar qi[5];
+    for (CeedInt j=0; j<5; j++) qi[j] = q[j][i];
     const CeedScalar x_i[3] = {x[0][i], x[1][i], x[2][i]};
-    State s = StateFromU(context, U, x_i);
+    State s = StateFromQi(context, qi, x_i);
```
jedbrown, line 246: An alternative is to make this function (what is now called `IFunction_Newtonian`) into a `CEED_QFUNCTION_HELPER` that takes as additional argument(s) the function pointers for the appropriate version of these operations. Then we'll have an `IFunction_Newtonian_Primitive` and `IFunction_Newtonian_Conservative` that are one-liners, just calling the helper with suitable extra args. This has a similar effect to C++ templates, but if you compile with optimization off, you can still step through the code you wrote in the debugging, including seeing/using the function pointers as defined. Since `CEED_QFUNCTION_HELPER` behaves like `static inline`, we should get both specializations optimized.

jrwrigh: I believe I have implemented this on the `jrwrigh/unite_fluids_pointers` branch, but I'm actually seeing practically identical results to what I was seeing before; the HIP implementation is about the same (or at least within noise of the measurement), but the CPU is consistently about 4% slower.

The branch currently keeps an `if (gas->primitive)` statement for handling the `U_dot` state, but manually removing that actually makes it a little bit slower (maybe 0.5%).

I'm running this on noether. Here's the `make info` for the compilation (in case that may be having an affect somehow):
```
CC            = gcc                                                                                                                                                                  
CXX           = g++                                                                                                                                                                  
FC            = gfortran                                                                                                                                                             
CPPFLAGS      = -I./include  -D__HIP_PLATFORM_HCC__ -D__HIP_PLATFORM_AMD__ -I/opt/rocm-5.1.3/hip/include -I/opt/rocm-5.1.3/llvm/bin/../lib/clang/14.0.0 -I/opt/rocm-5.1.3/hsa/include
CFLAGS        = $(OPT) $(CFLAGS.$(CC_VENDOR)) $(if $(ASAN),$(AFLAGS))                                                                                                                
CXXFLAGS      = $(OPT) $(CXXFLAGS.$(CC_VENDOR))                                                                                                                                      
FFLAGS        = $(OPT) $(FFLAGS.$(FC_VENDOR)) $(if $(ASAN),$(AFLAGS))                                                                                                                
NVCCFLAGS     = -ccbin $(CXX) -Xcompiler "$(OPT)" -Xcompiler -fPIC                                                                                                                   
HIPCCFLAGS    = $(filter-out $(OMP_SIMD_FLAG),$(OPT)) -fPIC -munsafe-fp-atomics                                                                                                      
CEED_LDFLAGS  = $(if $(ASAN),$(AFLAGS))                                                                                                                                              
CEED_LDLIBS   = -lm                                                                                                                                                                  
AR            = ar                                                                                                                                                                   
ARFLAGS       = crD                                                                                                                                                                  
OPT           = -O -g -march=native -ffp-contract=fast -fopenmp-simd                                                                                                                 
AFLAGS        = -fsanitize=address                                                                                                                                                   
```

jrwrigh: I tried with `OPT='-O3 -funsafe-math-optimizations -march=native'`, and the results are:

| Branch | Time (s) |
| ---- | ---- | 
| `main` | ~11.5 |
|  `jrwrigh/unite_fluids_primitive` (current) | ~11.7 |
|  `jrwrigh/unite_fluids_pointers` | ~15.3 |

So the pointers are not having the intended affect at all. I can only guess that they're not being inlined at all, though `-O3` should have the `-findirect-inlining` flag enabled through it. I tried removing the above mentioned if-statement, but the results are more-or-less the same.

jrwrigh: Alright, now I've got it.

A) Using `-fopt-info-inline` to dump out the optimization information, I confirmed that gcc was not inlining the helper function into the wrapper function _NOR_ was it inlining the `StateFrom*` function pointers into the helper function.

B) Prepending `__attribute__((always_inline))` to the beginning of the helper function forced gcc to inline it into the wrapper. Then gcc correctly inlined the `StateFrom*` into the wrapper function (ie. after the helper function was inlined). Doing that, I'm (finally) getting performance parity with the separate QFunctions.

Obviously the `__attribute__` is gcc only. I can't find an equivalent for intel compilers, though it's possible it might automatically inline the helper function into the wrapper. I'd like to know why gcc wasn't inlining the function even though we told it was available.

jrwrigh: >  I'd like to know why gcc wasn't inlining the function even though we told it was available.

Using `-Winline`, I get:
```
inlining failed in call to ‘IFunction_Newtonian_Generic’: --param large-stack-frame-growth limit reached [-Winline]
```

I was able to use `--param large-stack-frame=10000` to get the function to inline as expected. Can't say I really understand what that parameter is or how gcc uses it, but that's what was causing the inlining to fail.

jedbrown: Thanks for working this out. Does `__attribute__((flatten))` applied to `CEED_QFUNCTION` achieve the same result?

We can use `always_inline`, but I'd like to make sure we still have good debugging info at low optimization levels. I see that both gcc and clang `#define __NO_INLINE__ 1` when compiling at `-O0` (default) and drop that, instead `#define __OPTIMIZE__ 1` when compiling with optimization. We could use this if needed to ensure that qfunction helpers are *not* inlined in debug builds, but are always inlined in optimized builds.

jrwrigh: `__attribute__((flatten))` works and actually makes it faster too. Going this route also ropes in intel compilers with `#pragma forceinline`.

> We can use always_inline, but I'd like to make sure we still have good debugging info at low optimization levels.

Yeah, I had the same thought. Would we want to prepend `__attribute__((flatten))` to the `CEED_QFUNCTION` definition depending on `__NO_INLINE__` and/or `__OPTIMIZE__`? Intel specifies a `__NO_INLINE__` macro, but not an `__OPTIMIZE__` macro.

jedbrown: It's possible that the `__attribute__((flatten))` doesn't have an effect at `-O0` so please check that first. But if it still inlines, then yeah, I would
```c
#ifdef __NO_INLINE__ // and maybe compiler identifiers
#  define CEED_QFUNCTION_ATTR __attribute__((flatten))
#else
#  define CEED_QFUNCTION_ATTR
#endif
```
and then use `CEED_QFUNCTION_ATTR` in the `CEED_QFUNCTION()` macro.

jrwrigh: Yeah, `__attribute__((flatten))` does cause inlining at `-O0`.

include/ceed/types.h
```diff
@@ -13,6 +13,16 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#ifndef __NO_INLINE__
```
jeremylt, line 16: I think the discussion in this PR is great. It might be a good idea to put some of that into a comment like we have for the `CeedPragmaSIMD` macro.

Also, this might be worth including in a stand alone PR.

jrwrigh: A separate PR makes sense to me. Working on that now.

jrwrigh:
A couple of code opinion/style questions:
1. I've kept the if statement versions of `StateFromQi`. I think we had agreed that having these as slightly non-optimal for the boundary QFs is ok since they're not the dominant cost. I've still kept them here and only moved the volume QFs to the pointer method. We still good with that?

2. For the function pointers passed into `[IFunction,IJacobian]_Newtonian_Generic`, I've currently named as `pStateFromQi(_fwd)` emphasize that they're different than the `StateFromQi*` helper functions (`p` meaning "pointer"). I originally had the function pointers named as `StateFromQi*` since they have the exact same interface and (intended) behavior as the helper functions. I'm open to switching back or giving them a different name. Thoughts?

examples/fluids/problems/newtonian.c
```diff
@@ -137,8 +137,8 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
                     NULL);
   // -- Conservative vs Primitive variables
   ierr = PetscOptionsBool("-primitive", "Use primitive variables",
-                          NULL, prim_var=PETSC_FALSE, &prim_var, NULL); CHKERRQ(ierr);
-  if (prim_var) {
+                          NULL, use_primitive=PETSC_FALSE, &use_primitive, NULL); CHKERRQ(ierr);
+  if (use_primitive) {
```
jedbrown, line 141: Thinking ahead, should this be `-state_var [conservative|primitive|entropy]`? It'll be easy enough to change later, but independent flags can conflict, as in `-primitive -entropy`.

examples/fluids/qfunctions/newtonian_state.h
```diff
@@ -166,6 +166,24 @@ CEED_QFUNCTION_HELPER State StateFromY_fwd(NewtonianIdealGasContext gas,
   return ds;
 }
 
+// Generic State converters based on gas->use_primitive
+CEED_QFUNCTION_HELPER State StateFromQi(NewtonianIdealGasContext gas,
+                                        const CeedScalar Qi[5], const CeedScalar x[3]) {
+  if (gas->use_primitive) return StateFromY(gas, Qi, x);
+  else                    return StateFromU(gas, Qi, x);
+}
```
jedbrown, line 174: I'm okay with this, though I can't say I love having the two different conventions.

jedbrown (approved):

examples/fluids/qfunctions/newtonian_state.h
jrwrigh: I'm not married to having the different conventions. I honestly don't like having different ones, but I'm being a bit lazy. I can change them to be all pointer based (I'd probably make a typedef for `StateFromQi*` function pointers to ease the verbosity in the helper function definitions).

examples/fluids/problems/newtonian.c
jrwrigh: It should. I was kinda pushing that off until we get a more permanent place for the fluids code (be it ratel or somewhere else).

examples/fluids/qfunctions/newtonian_state.h
jrwrigh: I've just pushed up changes to replace `StateFromQi` with function pointers. Given the large change, I'd appreciate a quick glance to make sure I haven't done anything too nefarious.

examples/fluids/problems/newtonian.c
jrwrigh: Moved to #1052, as it ended up being a bit involved that I had anticipated (mainly multiple uses of the `SetupContext` structs).

examples/fluids/qfunctions/newtonian.h
```diff
@@ -324,12 +313,21 @@ CEED_QFUNCTION(IFunction_Newtonian)(void *ctx, CeedInt Q,
...
-    for (CeedInt j=0; j<5; j++) U_dot[j] = q_dot[j][i];
+    if (context->use_primitive) {
+      CeedScalar qi_dot[5], dx0[3] = {0};
+      for (int j=0; j<5; j++) qi_dot[j] = q_dot[j][i];
+      State s_dot = StateFromY_fwd(context, s, qi_dot, x_i, dx0);
+
+      UnpackState_U(s_dot.U, U_dot);
+    } else {
+      for (CeedInt j=0; j<5; j++) U_dot[j] = q_dot[j][i];
+    }
```
jedbrown, line 327: Should we use `StateFromQi_fwd` here for uniformity? I think the "cost" is negligible in the scheme of things, and should be optimized out after inlining.

jedbrown (approved):
Looks good. I'm assuming you feel comfortable with the test coverage or have run a few things beyond.

examples/fluids/qfunctions/newtonian.h
jrwrigh: I had the same thought going to sleep last night. haha. Yeah, after inlining, the conservative case should effectively break down to copying data into the `State`, then copying it back out.

PR #1036, opened 2022-08-04
QF headers for typedefs and macros

jeremylt:
This PR adds the ability for users to include files from the path `libceed/include/ceed/jit-source/*` inside of QFunctions and have them read during JiT.

This PR also puts common QF typedes in `#include <ceed/types.h>` and uses that file in JiT.


examples/fluids/qfunctions/advection.h
```diff
@@ -11,8 +11,8 @@
 #ifndef advection_h
 #define advection_h
 
+#include <ceed/types.h>
```
jeremylt, line 14: The previous code will (should) work fine here as `<ceed.h>` is still ignored in JiT, but this header has less stuff in it, so I switched all of these `#include <ceed.h>` over to the new `#include <ceed/types.h>`

interface/ceed-jit-tools.c
```diff
@@ -140,23 +141,40 @@ int CeedLoadSourceToInitializedBuffer(Ceed ceed,
...
-      bool is_local_header = is_hash_include && next_quote
-                             && (next_new_line - next_quote > 0);
-      if (is_local_header) {
+      bool is_local_header = is_hash_include && next_quote &&
+                             (next_new_line - next_quote > 0);
+      char *next_left_chevron = strchr(first_hash, '<');
+      bool is_ceed_header = is_hash_include && next_left_chevron &&
+                            (next_new_line - next_left_chevron > 0) &&
+                            !strncmp(next_left_chevron, "<ceed/", 6) &&
+                            strncmp(next_left_chevron, "<ceed/ceed.h>", 13);
```
jeremylt, line 150: I think it might be a good idea to
a) exclude more files here or
b) include only the path `<ceed/jit-source/*>` and the files `<ceed/types.h>`, `<ceed/ceed-f32.h>` and `<ceed/ceed-f64.h>`?

include/ceed/jit-source/cuda/cuda-types.h
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and other CEED contributors.
```
jeremylt, line 1: Woops, stray file from renaming

jeremylt:
Local CUDA:
```
All tests successful.
Files=208, Tests=8502, 65 wallclock secs ( 1.40 usr  0.28 sys + 469.32 cusr 85.93 csys = 556.93 CPU)
Result: PASS
```

Noether ROCm:
```
All tests successful.
Files=208, Tests=8502, 88 wallclock secs ( 2.08 usr  0.66 sys + 2150.97 cusr 4749.35 csys = 6903.06 CPU)
Result: PASS
```

This should be good to merge (probably squash + merge, but I kept the commits separate in case we want to revert one)

jedbrown (approved):
It's been an insane wait for our new GPU for Noether, but it should be delivered in the next week or two.

PR #1037, opened 2022-08-05
Installed QFunction Helpers

jeremylt:
We can start adding helper functions we want to install on this branch.

We should make sure any helper function we add is used somewhere.

I think we need to find a balance between having a large number of files and a large number of unused function warnings.

### DIFF
--- a/include/ceed/jit-source/utils/ceed-mat2.h
+++ b/include/ceed/jit-source/utils/ceed-mat2.h
@@ -1,0 +1,13 @@
+// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and other CEED contributors.
+// All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+//
+// This file is part of CEED:  http://github.com/ceed
+
+/// @file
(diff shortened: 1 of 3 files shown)


jedbrown:
One design decision here is how strongly typed to make these interfaces. If you do something like the following, you can eliminate a class of mistakes by compile-time errors. The downside is that raw access to arrays is harder.
```c
typedef struct {
    CeedScalar values[3][3];
} Mat3x3;

typedef struct {
    CeedScalar values[6];
} Mat3x3KelvinMandel;
```
In C++, you can make types like this that also overload the `[]` operator so you can have standard indexing and strong typing. There is lots of prior art in this space (header-only libraries for small matrix arithmetic) and I'd be reluctant to create something new without a compelling reason after evaluating the prior art.

jrwrigh:
I've added quite a few Vector and matrix related tools in my SGS branch (this commit specifically: 7df3dd064d6625ef5da91a86907791bc5a2a09a0). 

Most of the matrix ones are built with a size-agnostic function, with a wrapper to get the function signature in a more user-friendly format. I figure that the compiler should be able to do all the loop unrolling it wants, especially since the size of the loops are known at runtime. Would we be interested in taking a similar strategy here?

PR #1039, opened 2022-08-07
Fluids - Compressible Blasius boundary layer

LeilaGhaffari:
This is not converging yet and I am not sure where to use the exact temperature in the in/outflow BCs.


examples/fluids/qfunctions/blasius.h
```diff
@@ -50,34 +55,32 @@ CEED_QFUNCTION(ICsBlasius)(void *ctx, CeedInt Q,
...
-    const CeedScalar x[] = {X[0][i], X[1][i], X[2][i]};
+    const CeedScalar x[3] = {X[0][i], X[1][i], 0.};
+    State s = BlasiusSolution(context, x, x0, x_inflow, rho, &t12);
+    CeedScalar q[5] = {0};
+    UnpackState_U(s.U, q);
+    for (CeedInt j=0; j<5; j++) q0[j][i] = q[j];
 
-    BlasiusSolution(context, x[1], x0, x[0] - x_inflow, rho, &u, &v, &T, &t12);
+    // ToDo: Need to drop this line
+    q0[4][i] = rho * (e_internal + 0.5*Dot3(s.Y.velocity, s.Y.velocity));
```
LeilaGhaffari: It diverges without this line.

jedbrown: It looks like `rho` and `e_internal` here are constant. How different are they from the computed values? (Maybe set a breakpoint here to inspect in the debugger, or use printf and filter the output.)

LeilaGhaffari: There was a memory issue that is fixed now. It diverges much later (`t=0.000286`) now. The major difference I am seeing now is that the `f[]` values are much larger than `f`, `fp`, `fpp` on main (incompressible Blasius) which makes the values of u, v, and t12 ~22-27 times bigger. The value of `h[0]` and the computed temperature makes sense though!

jedbrown: So we don't match `Uinf` as specified? How does it compare if you spot-check it versus the compressible Julia code? We spot-checked the PETSc example so it should agree with the Julia, which visually matched diagrams from the literature.

LeilaGhaffari: I spot-checked it versus the Julia code and they were identical. I had to increase the value of `delta0` from `4.2e-4` to `4.2e-3` to have $\eta_{max} = 5$ which helped with converging. 
Screenshot from 2022-08-10 12-31-59
These are taken after 40 time-steps. The left plot is the compressible blasius solution (this branch) and the right one is run on main.

jedbrown: Can you make a line chart at the inflow?

jedbrown: And with the smaller `delta0`, I'm assuming it struggled to converge because $`\eta_{\max} = 50`$? I think in practice, the (laminar) far field is constant (to fairly strict tolerance, though we could quantify) beyond about $`\eta_{\max} = 10`$. So we could solve on that limited domain and extrapolate. Or we can pick a better transform from $`[0,\infty) \to [-1, 1]`$.

LeilaGhaffari: That's right, $\eta_{max}$ was 50.

LeilaGhaffari: At Ma = 0:
Screenshot from 2022-08-25 14-44-36

LeilaGhaffari:
I think I am going to add primitive variables to this problem after  #1035 is merged.

examples/fluids/qfunctions/blasius_exact.h
```diff
@@ -0,0 +1,139 @@
...
+  PetscInt          N = blasius->n_cheb;
+  PetscScalar Ma = Mach(&blasius->newtonian_ctx, blasius->Tinf, blasius->Uinf),
+              Pr = Prandtl(&blasius->newtonian_ctx),
+              gamma = HeatCapacityRatio(&blasius->newtonian_ctx);
+  PetscFunctionBegin;
+  PetscCall(VecGetArrayRead(X, &Tf));
+  Th = Tf + N;
+  PetscCall(VecGetArray(R, &r));
+
+  // Compute function
```
jedbrown, line 79:
```suggestion
  // Left boundary conditions f = f' = 0
```

examples/fluids/qfunctions/blasius_exact.h
```diff
@@ -0,0 +1,139 @@
...
+  r[N+1] = h[0] - 1.;
+
+  // Restore vectors
+  PetscCall(VecRestoreArrayRead(X, &Tf));
+  PetscCall(VecRestoreArray(R, &r));
+  PetscFunctionReturn(0);
+}
+
+// *****************************************************************************
+PetscErrorCode ComputeChebyshevCoefficients(BlasiusContext blasius) {
```
jedbrown, line 111: I would move this and `CompressibleBlasiusResidual` to `blasius.c`. They don't need to be called from qfunctions.

examples/fluids/qfunctions/blasius.h
```diff
@@ -13,107 +13,34 @@
...
-    f   = f_table[nprofs-1];
-    fp  = fp_table[nprofs-1];
-    fpp = fpp_table[nprofs-1];
-    eta = eta_table[nprofs-1];
-  }
-
-  *u = Uinf*fp;
-  *t12 = rho*nu*Uinf*fpp*sqrt(Uinf/(nu*(x0+x)));
-  *v = 0.5*sqrt(nu*Uinf/(x0+x))*(eta*fp - f);
+#include "blasius_exact.h"
```
jedbrown, line 19: I think `ChebyshevEval` could just move here, and the struct can remain so we don't need `blasius_exact.h` at all. (I think it doesn't justify the extra header just for `ChebyshevEval`. This is assuming we move the SNES solver and residual over to `blasius.c`.)

examples/fluids/problems/blasius.c
```diff
@@ -162,31 +162,40 @@ PetscErrorCode NS_BLASIUS(ProblemData *problem, DM dm, void *ctx) {
...
+  problem->ics.qfunction        = ICsBlasius;
+  problem->ics.qfunction_loc    = ICsBlasius_loc;
+
+  CeedScalar Uinf               = 40;          // m/s
+  CeedScalar Tinf               = 288.;        // K
+  CeedScalar T_wall             = 400.;        // K
+  CeedScalar delta0             = 4.2e-3;      // m
+  CeedScalar theta0             = 288.;        // K
+  CeedScalar P0                 = 1.01e5;      // Pa
+  CeedInt    N                  = 10;          // Number of Chebyshev terms
```
jedbrown, line 174: I think this may not be enough, especially at higher Mach or if $\eta_\max$ is larger. I think we were pretty happy with 20 through a reasonable range. Solution evaluation is cheap in any case, but I'd be inclined to bump this up to avoid users being surprised.
```suggestion
  CeedInt    N                  = 20;          // Number of Chebyshev terms
```

jedbrown (approved):
Looks great. The only real thing is that I think the SNES solver can move to `blasius.c`.

LeilaGhaffari:
Thanks @jedbrown . I am happy with a `squash-merge` if this looks good to you now.

examples/fluids/qfunctions/blasius.h
```diff
@@ -320,11 +287,11 @@ CEED_QFUNCTION(Blasius_Inflow_Jacobian)(void *ctx, CeedInt Q,
...
                                };
 
-    const CeedScalar u_normal = Dot3(norm, velocity);
+    const CeedScalar u_normal = Dot3(norm, s.Y.velocity);
 
     v[0][i] = - wdetJb * drho * u_normal;
     for (int j=0; j<3; j++)
-      v[j+1][i] = -wdetJb * (drho * u_normal * velocity[j] + norm[j] * dP);
+      v[j+1][i] = -wdetJb * (drho * u_normal * s.Y.velocity[j] + norm[j] * dP);
     v[4][i] = - wdetJb * u_normal * (dE + dP);
```
jrwrigh, line 295: Similar here for using `FluxTotal_Boundary`

examples/fluids/qfunctions/blasius.h
```diff
@@ -245,17 +215,18 @@ CEED_QFUNCTION(Blasius_Inflow)(void *ctx, CeedInt Q,
...
     // -- Momentum
     for (CeedInt j=0; j<3; j++)
-      v[j+1][i] -= wdetJb * (rho * u_normal * velocity[j] // interior rho
+      v[j+1][i] -= wdetJb * (rho * u_normal * s.Y.velocity[j] // interior rho
                              + norm[j] * P // mixed P
                              + viscous_flux[j]);
 
     // -- Total Energy Density
-    v[4][i] -= wdetJb * (u_normal * (E + P) + Dot3(viscous_flux, velocity));
+    v[4][i] -= wdetJb * (u_normal * (E + P) + Dot3(viscous_flux, s.Y.velocity));
```
jrwrigh, line 223: I added `FluxTotal_Boundary` to the `newtonian_state.h`, so that could be used here to simplify things a decent bit.

jrwrigh (approved):

examples/fluids/qfunctions/blasius.h
LeilaGhaffari: Nice, I did try to use `FluxTotal_Boundary()` but I ended up writing more code since the final state variable is coming from different sources. Can we merge this now and let me think about it more and change it when I add primitive variables?

jrwrigh: Yeah, that's fine!

examples/fluids/problems/blasius.c
```diff
@@ -162,31 +233,40 @@ PetscErrorCode NS_BLASIUS(ProblemData *problem, DM dm, void *ctx) {
...
-  PetscReal  mesh_refine_height = 5.9e-4; // m
-  PetscReal  mesh_growth        = 1.08;   // [-]
-  PetscInt   mesh_Ndelta        = 45;     // [-]
-  PetscReal  mesh_top_angle     = 5;      // degrees
+  problem->ics.qfunction        = ICsBlasius;
+  problem->ics.qfunction_loc    = ICsBlasius_loc;
+
+  CeedScalar Uinf               = 40;          // m/s
+  CeedScalar Tinf               = 288.;        // K
+  CeedScalar T_wall             = 400.;        // K
```
jedbrown, line 241: Should we be consistent here in naming? The last bit is updating `examples/fluids/README.md` and making sure that this code is tested. The three Blasius tests all use `blasius-stgtest.yaml`, which have `stg_use: true`. So I think we need a test with compressible Blasius.

```suggestion
  CeedScalar U_inf              = 40;          // m/s
  CeedScalar T_inf              = 288.;        // K
  CeedScalar T_wall             = 400.;        // K
```

examples/fluids/qfunctions/blasius.h
```diff
@@ -128,35 +109,28 @@ CEED_QFUNCTION(ICsBlasius)(void *ctx, CeedInt Q,
...
+  const CeedScalar delta0     = context->delta0;
+  const CeedScalar Uinf       = context->Uinf;
   const CeedScalar x_inflow   = context->x_inflow;
-
+  const CeedScalar gamma      = HeatCapacityRatio(&context->newtonian_ctx);
   const CeedScalar e_internal = cv * theta0;
   const CeedScalar rho        = P0 / ((gamma - 1) * e_internal);
-  const CeedScalar x0         = Uinf*rho / (mu*25/ (delta0*delta0) );
-  CeedScalar u, v, t12;
+  const CeedScalar x0         = Uinf*rho / (mu*25/(delta0*delta0));
```
jedbrown, line 122: It seems like `theta0` needs to be specified somehow compatible with `T_wall` and `T_inf`. I think it probably shouldn't be used at all. I think it should probably be using `T_inf` https://en.wikipedia.org/wiki/Blasius_boundary_layer#Howarth_transformation

examples/fluids/problems/blasius.c
jedbrown: At least in the user input, should we spell these out, like `velocity_infinity`, `temperature_wall`? Anyway, I don't love ambiguity in T vs theta for temperature. Full names help a user do things like `git grep -i temperature *.yaml`, though this isn't necessarily what we should optimize for.

examples/fluids/qfunctions/blasius.h
LeilaGhaffari: Oops, I meant to fix this but completely forgot about it. Thanks, dropping `theta0` in a new PR.

jedbrown: Cool, sorry about the timing in my review.

PR #1040, opened 2022-08-12
emscripten build

jedbrown:
I can build and run the pure libceed examples, which means we could embed them in a static web page that would run simulations in the user's browser.

This now builds cleanly:

```console
$ emmake make STATIC=1 OPT=-O build/ex2-surface.wasm
[...]
$ wasmer run build/ex2-surface.wasm -- -s 32767
[...]
Exact mesh surface area    :  6
Computed mesh surface area :  5.9548265118955
Surface area error         : -0.04517348810445
```
Increasing the size causes calloc to fail, likely to do some default memory limits that I haven't worked out yet.

Emscripten docs: https://emscripten.org/docs/compiling/Building-Projects.html#building-projects

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -424,10 +424,10 @@ ifneq ($(wildcard $(MAGMA_DIR)/lib/libmagma.*),)
   MAGMA_ARCH=$(shell nm -g $(MAGMA_DIR)/lib/libmagma.* | grep -c "hipblas")
   ifeq ($(MAGMA_ARCH), 0) #CUDA MAGMA
     ifneq ($(CUDA_LIB_DIR),)
-      cuda_link = -Wl,-rpath,$(CUDA_LIB_DIR) -L$(CUDA_LIB_DIR) -lcublas -lcusparse -lcudart
+      cuda_link = $(if $(STATIC),,-Wl,-rpath,$(CUDA_LIB_DIR)) -L$(CUDA_LIB_DIR) -lcublas -lcusparse -lcudart
       omp_link = -fopenmp
       magma_link_static = -L$(MAGMA_DIR)/lib -lmagma $(cuda_link) $(omp_link)
-      magma_link_shared = -L$(MAGMA_DIR)/lib -Wl,-rpath,$(abspath $(MAGMA_DIR)/lib) -lmagma
+      magma_link_shared = -L$(MAGMA_DIR)/lib $(if $(STATIC),,-Wl,-rpath,$(abspath $(MAGMA_DIR)/lib)) -lmagma
       magma_link := $(if $(wildcard $(MAGMA_DIR)/lib/libmagma.${SO_EXT}),$(magma_link_shared),$(magma_link_static))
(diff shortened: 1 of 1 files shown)


interface/ceed-qfunction.c
```diff
@@ -602,7 +602,11 @@ int CeedQFunctionCreateInterior(Ceed ceed, CeedInt vec_length,
...
   (*qf)->function = f;
   (*qf)->user_flop_estimate = -1;
-  if (strlen(source)) {
+  // TODO: This clumsily assumes no JIT on host to avoid searching the file
+  // system. Searching the filesystem fails with WASM (default wasmer run).
+  CeedMemType mem_type;
+  ierr = CeedGetPreferredMemType(ceed, &mem_type); CeedChk(ierr);
+  if (strlen(source) && mem_type != CEED_MEM_HOST) {
     bool is_absolute_path;
     char *absolute_path;
```
jedbrown, line 611: @jeremylt How concerned are you about this hack? It needs a tweak to fix the tests (example below), but I'm not sure whether to add an explicit field to specify *not* to check the filesystem or to infer from the environment/gracefully fall back when it can't.
```diff
--- tests/output/t402-qfunction.out
+++ New
@@ -1,4 +1,4 @@
-User CeedQFunction - setup
+User CeedQFunction - (null)
   1 input field:
     Input field 0:
       Name: "w"
@@ -9,7 +9,7 @@
       Name: "qdata"
       Size: 1
       EvalMode: "none"
-User CeedQFunction - mass
+User CeedQFunction - (null)
   2 input fields:
     Input field 0:
       Name: "qdata"
```

jeremylt: It would be nice to preserve this piece of debugging info if it is reasonably possible.

jedbrown: We have the name (I just need a more careful conditional), but `CeedGetJitAbsolutePath` fails under WASM so we need to skip that. But what should we use to decide whether to try? `mem_type`, some other environment info, or always try but with graceful failure (which might hide a different issue when run in an environment that should be able to find it)?

jeremylt: We could go for a more lazy approach. Always store the full string and the name but only form and store the abspath if the backend requests it?

jedbrown: That sounds good.

jeremylt:
Do we want to include this in the release? The above suggestion would be a ~15 minute commit that I could put together.

jedbrown:
I think it would be wonderful if we can, and then add a cursory emscripten build (just the base library) to CI.

jedbrown:
I don't think it was intended to close this PR, though this PR should be rebased (by someone, maybe me).

jeremylt (approved):

jeremylt:
Did you want to add this to the docs here or in a follow-up?

jedbrown:
Thanks. I think this can be a sleeper feature until we figure out the size limit and get it in CI, at which point we can announce it.

PR #1044, opened 2022-08-22
Makefile: support icc_orig (icc --version on quartz)

jedbrown:
Quartz changed this output after oneAPI compilers came on the scene.


jeremylt (approved):

jedbrown:
There's a docs issue (I think related to version pinning, https://readthedocs.org/projects/libceed/builds/17868027/) that's orthogonal to this so I'll merge now and we can fix that separately.

PR #1048, opened 2022-08-24
Add CEED_QFUNCTION_ATTR for inlining

jrwrigh:
GCC doesn't like to inline all Qfunction helpers, so this forces it do so if inlining is allowed at all.

This was originally apart of #1035, but has been stripped out into this PR. A more detailed discussion is in https://github.com/CEED/libCEED/pull/1035#discussion_r938446943, but to summarize:

GCC (and presumably other compilers) may elect to not inline a function, even with the `inline` modifier specified. This can lead to detrimental performance, particularly if the inlining is required to allow for further optimizations (such as passing a function pointer to a helper function). Forcing a function to always be inlined can however lead to a poor debugging experience. Thus we add a `CEED_QFUNCTION_ATTR` macro that is created based on compiler-set macros (namely `__NO_INLINE__`) which activates forced inlining. This macro is then embedded into the `CEED_QFUNCTION` definition.

### DIFF
--- a/include/ceed/types.h
+++ b/include/ceed/types.h
@@ -15,6 +15,28 @@
 
 /**
   @ingroup CeedQFunction
+  This macro defines compiler attributes to the CEED_QFUNCTION to force inlining
+    for called functions. The `inline` declaration does not necessarily enforce a
+    compiler to inline a function. This can be deterimental to performance, so
+    here we force inlining to occur unless inlining has been forced off (like
+    during debugging).
+**/
+#ifndef CEED_QFUNCTION_ATTR
+#ifndef __NO_INLINE__
+#  if defined(__GNUC__) || defined(__clang__)
+#    define CEED_QFUNCTION_ATTR __attribute__((flatten))
+#  elif defined(__INTEL_COMPILER)
+#    define CEED_QFUNCTION_ATTR _Pragma("forceinline")
+#  else
+#    define CEED_QFUNCTION_ATTR
+#  endif
+#else
+#  define CEED_QFUNCTION_ATTR
+#endif
+#endif
+
+/**
+  @ingroup CeedQFunction
   This macro populates the correct function annotations for User QFunction
     source for code generation backends or populates default values for CPU
     backends. It also creates a variable `name_loc` populated with the correct
@@ -23,7 +45,7 @@
 #ifndef CEED_QFUNCTION
 #define CEED_QFUNCTION(name) \
   static const char name ## _loc[] = __FILE__ ":" #name;        \
-  static int name
+  CEED_QFUNCTION_ATTR static int name
 #endif
 
 /**
@@ -33,7 +55,7 @@
     values for CPU backends.
 **/
 #ifndef CEED_QFUNCTION_HELPER
-#define CEED_QFUNCTION_HELPER static inline
+#define CEED_QFUNCTION_HELPER CEED_QFUNCTION_ATTR static inline
 #endif
 
 /**
--- a/python/build_ceed_cffi.py
+++ b/python/build_ceed_cffi.py
@@ -21,6 +21,7 @@ for header_path in ["include/ceed/types.h", "include/ceed/ceed.h"]:
         lines += [line.strip() for line in f if
                   not (line.startswith("#") and not line.startswith("#include")) and
                   not line.startswith("  static") and
+                  not line.startswith("  CEED_QFUNCTION_ATTR") and
                   "CeedErrorImpl" not in line and
                   "const char *, ...);" not in line and
                   not line.startswith("CEED_EXTERN const char *const") and


include/ceed/types.h
```diff
@@ -23,7 +33,7 @@
 #ifndef CEED_QFUNCTION
 #define CEED_QFUNCTION(name) \
   static const char name ## _loc[] = __FILE__ ":" #name;        \
-  static int name
+  CEED_QFUNCTION_ATTR static int name
```
jrwrigh, line 36: If we want to use intel's `#pragma forceinline`, we'd need to do something like this:
```suggestion
  CEED_QFUNCTION_ATTR 
  static int name
```

I'm not super familiar with C/C++ preprocessing, so I'm not sure if this form is compatible with `__attribute__` and  `_Pragma`.

include/ceed/types.h
```diff
@@ -13,6 +13,22 @@
 #include <stddef.h>
 #include <stdint.h>
 
+/**
+ This macro defines compiler attributes to the CEED_QFUNCTION to force inlining
+ for called functions. The `inline` declaration does not necessarily enforce a
+ compiler to inline a function. This can be deterimental to performance, so
+ here we force inlining to occur.
+**/
```
jrwrigh, line 21: I'm leaving of a `@ingroup` specification here as this macro shouldn't really be used by front-end or back-end developers. Let me know if there's a place for this though.

jrwrigh:
I have no idea why the python install is failing. I can reproduce it on my machine, but it's failing on the (seemingly) unrelated definition of `CeedInt`. Any ideas?

include/ceed/types.h
jeremylt: Does the Python failure stay if you put in an ingroup specification?

jrwrigh: Yep.

jeremylt: That error means that the Python is choking while parsing this header as part of setting up the Python interface

include/ceed/types.h
```diff
@@ -13,6 +13,23 @@
...
+#ifndef __NO_INLINE__
+#  if defined(__GNUC__) || defined(__clang__)
+#    define CEED_QFUNCTION_ATTR __attribute__((flatten))
+#  else
+#    define CEED_QFUNCTION_ATTR
+#  endif
+#else
+#  define CEED_QFUNCTION_ATTR
+#endif
```
jeremylt, line 31: I think we want to let backends override `CEED_QFUNCTION_ATTR`

```suggestion
/**
  @ingroup CeedQFunction
  This macro defines compiler attributes to the CEED_QFUNCTION to force inlining
    for called functions. The `inline` declaration does not necessarily enforce a
    compiler to inline a function. This can be deterimental to performance, so
    here we force inlining to occur unless inlining has been forced off (like
    during debugging).
**/
#ifndef CEED_QFUNCTION_ATTR
#ifndef __NO_INLINE__
#  if defined(__GNUC__) || defined(__clang__)
#    define CEED_QFUNCTION_ATTR __attribute__((flatten))
#  else
#    define CEED_QFUNCTION_ATTR
#  endif
#else
#  define CEED_QFUNCTION_ATTR
#endif
#endif
```

jeremylt:
Do we want to update `CEED_QFUNCTION_HELPER` as well?

include/ceed/types.h
jeremylt: I think this should be fine, but you would need to write
```

  CEED_QFUNCTION_ATTR                                           \
  static int name
```

jrwrigh:
> Do we want to update `CEED_QFUNCTION_HELPER` as well?

I don't see why not.

include/ceed/types.h
jrwrigh: Would that require a change to the python code fix?

jeremylt: Right, you should be fine switching
```

Comment (jrwrigh) [reply include/ceed/types.h:46]: Yeah, that's what I ended up with.

Comment (jrwrigh) [reply include/ceed/types.h:46]: Turns out that putting `CEED_QFUNCTION_ATTR` on a separate line both doesn't do anything and isn't needed.

It doesn't do anything because apparently the preprocessor removes the `\` before it does anything else. They allow for line continutation, but don't affect the actual output.

It's not needed because a `_Pragma()` will automatically insert a newline.

This is confirmed from looking at the output of gcc after preprocessing. Keeping `CEED_QFUNCTION_ATTR` on the same physical line in the macro and setting it equal to `__attribute__((flatten))` gets:
```c
__attribute__((flatten)) static inline State StateFromY(NewtonianIdealGasContext gas,
                                       const CeedScalar Y[5], const CeedScalar x[3]) {
```

while setting it to `_Pragma("message(__FILE__)")` gets:
```c
# 144 "/home/jrwrigh/software/libCEED/examples/fluids/problems/../qfunctions/newtonian_state.h"
#pragma message("/home/jrwrigh/software/libCEED/examples/fluids/problems/../qfunctions/newtonian_state.h")
# 144 "/home/jrwrigh/software/libCEED/examples/fluids/problems/../qfunctions/newtonian_state.h"
 static inline State StateFromY(NewtonianIdealGasContext gas,
                                       const CeedScalar Y[5], const CeedScalar x[3]) {
```

jeremylt (approved):
This looks good to me

PR #1049, opened 2022-08-24
doc: Add CEED_QFUNCTION_ATTR to releasenotes

jrwrigh:
Forgot to add to the release notes for #1048.

Also, didn't realize that the `CEED_QFUNCTION_ATTR` isn't actually showing up in the docs. I'm assuming that's supposed to be happening. I just built the docs locally and it doesn't appear there either.

### DIFF
--- a/doc/sphinx/source/releasenotes.md
+++ b/doc/sphinx/source/releasenotes.md
@@ -15,6 +15,7 @@ On this page we provide a summary of the main API changes, new features and exam
 
 - Update `/cpu/self/memcheck/*` backends to help verify `CeedQFunctionContext` data sizes provided by user.
 - Added `CeedInt_FMT` to support potential future use of larger interger sizes.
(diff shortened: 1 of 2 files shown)


jeremylt (approved):

doc/sphinx/source/releasenotes.md
```diff
@@ -15,6 +15,7 @@ On this page we provide a summary of the main API changes, new features and exam
 
 - Update `/cpu/self/memcheck/*` backends to help verify `CeedQFunctionContext` data sizes provided by user.
 - Added `CeedInt_FMT` to support potential future use of larger interger sizes.
+- Added CEED_QFUNCTION_ATTR for setting compiler attributes/pragmas to CEED_QFUNCTION_HELPER and CEED_QFUNCTION
```
jedbrown, line 18: I'm not sure this is intended as a public symbol. The functional change for users will be more aggressive inlining.

jrwrigh: It's not intended as a public symbol on the user/frontend side, but Jeremy mentioned that it might be useful for backends to be able to overwrite or edit it.

PR #1050, opened 2022-08-25
Refactor `cuda-gen` and `hip-gen` backends.

YohannDudouit:
The purpose of this PR is to propose a refactor of the `cuda-gen` and `hip-gen` backends to be compatible with the device functions of the `cuda-shared` and `hip-shared` backends.

- [x] Move `cuda-gen` and `hip-gen` device functions in jit source folder.
- [x] Make `cuda-shared` and `hip-shared` compatible with `cuda-gen` and `hip-gen` device functions.
- [ ] Factorize `cuda-gen` and `hip-gen` code in functions to help readability and extensibility.


jeremylt:
I'm trying to get #1051 merged first, but I want to jump back to this as soon as that is merged

jeremylt:
I'm going to spend tomorrow on this and see if I can get it in a state where we can test it next week at the hackathon

include/ceed/jit-source/cuda/cuda-shared-basis-tensor-templates.h
```diff
@@ -0,0 +1,419 @@
...
+//------------------------------------------------------------------------------
+// 1D
+//------------------------------------------------------------------------------
+
+//------------------------------------------------------------------------------
+
+//------------------------------------------------------------------------------
+// E-vector -> single element
+//------------------------------------------------------------------------------
+template <int NUM_COMP, int P_1D>
```
jeremylt, line 27: @nbeams, one change I had to make here that these aren't templated on the strides since `num_elem` isn't known at compile time for bases

include/ceed/jit-source/cuda/cuda-shared-basis-tensor.h
```diff
@@ -0,0 +1,396 @@
...
+
+//------------------------------------------------------------------------------
+// 1D interpolate to quadrature points
+//------------------------------------------------------------------------------
+inline __device__ void Interp1d(BackendData &data, const CeedInt num_elem, const CeedInt transpose,
+                                const CeedScalar *c_B,
+                                const CeedScalar *__restrict__ d_U,
+                                CeedScalar *__restrict__ d_V) {
+  CeedScalar r_U[BASIS_BUF_LEN];
+  CeedScalar r_V[BASIS_BUF_LEN];
```
jeremylt, line 19: In general, I wasn't sure what size to make these register arrays

include/ceed/jit-source/cuda/cuda-shared-basis-tensor.h
```diff
@@ -0,0 +1,396 @@
...
+
+  for (CeedInt elem = blockIdx.x*blockDim.z + threadIdx.z; elem < num_elem; elem += gridDim.x*blockDim.z) {
+    if (transpose) {
+      ReadElementStrided1d<BASIS_NUM_COMP, BASIS_P_1D>(data, elem, 1, BASIS_Q_1D*num_elem, BASIS_Q_1D, d_U, &r_U);
```
jeremylt, line 23: Related - I was getting issues calling the templates here

jeremylt:
```
default_program(644): error: no instance of function template "ReadElementStrided1d" matches the argument list
            argument types are: (BackendData, CeedInt, int, int, int, const CeedScalar *__restrict__, CeedScalar (*)[4])
```

jeremylt:
Huzzah, the gen basis kernels work in shared now (minus the collocated, which isn't part of shared currently anyway)

nbeams:
I guess we should probably think of a new name for these basis device functions (the file `ceed-cuda-shared-basis-tensor-templates.h`) -- since they can be used in shared and gen...

jeremylt:
Yeah - though I'm fine with that as a WIP name I suppose

include/ceed/jit-source/cuda/cuda-shared-basis-tensor-templates.h
```diff
@@ -0,0 +1,522 @@
...
+  CeedScalar r_t2[T_1D];
+  for (CeedInt comp = 0; comp < NUM_COMP; comp++) {
+    ContractX3d<NUM_COMP, P_1D, Q_1D>(data, r_U + comp*P_1D, c_B, r_t1);
+    ContractY3d<NUM_COMP, P_1D, Q_1D>(data, r_t1, c_B, r_t2);
+    ContractZ3d<NUM_COMP, P_1D, Q_1D>(data, r_t2, c_B, r_t1);
+    ContractX3d<NUM_COMP, Q_1D, Q_1D>(data, r_t1, c_G, r_V + comp*Q_1D + 0*NUM_COMP*Q_1D);
+    ContractY3d<NUM_COMP, Q_1D, Q_1D>(data, r_t1, c_G, r_V + comp*Q_1D + 1*NUM_COMP*Q_1D);
+    ContractZ3d<NUM_COMP, Q_1D, Q_1D>(data, r_t1, c_G, r_V + comp*Q_1D + 2*NUM_COMP*Q_1D);
+  }
+}
```
jeremylt, line 491: This shows that the "useCollocated" mode in the gen backend is muddled with a change parallelization strategy instead of being solely focused on collocated grad.

jeremylt:
Ok, I have pulled the quoted kernel out of cuda-gen. I'm sure we can refine more, but I would call this branch a win

1) make matching changes in hip-gen and hip-shared
2) profile to make sure we didn't lose (might have gained) perf on *-shared
3) tidy up the PR

In a separate PR we can add non-tensor versions of these bases and then try to use them in *-gen with the "colloGrad = false" path (see notes about about misleading name)

nbeams:
> In a separate PR we can add non-tensor versions of these bases and then try to use them in *-gen with the "colloGrad = false" path

By the way, Ahmad has been working on a specialized GEMM basis action kernel that will get better performance for lower-order tetrahedron elements -- where the sizes can be very highly skewed due to relatively small P and Q compared to number of elements per GPU (for an efficient number of DOFs per GPU).  He's been developing and testing it in MAGMA, but the plan is to eventually make the code available through RTC (like the MAGMA backend tensor kernels) directly in libCEED.

jeremylt:
That would be great - I've got a use case for that with Ratel

nbeams:
> Ok, I think all of the changes we want in this PR of cuda are in the cuda-shared and cuda-gen backends.

Great, I can start working on taking care of the HIP changes, then.

jeremylt:
Sounds good - sorry the history is so messy.

nbeams:
Would you rather start a new branch with a cleaner history (e.g. all changes in cuda-shared, then all changes in cuda-gen) and continue working from there?  Or will we just go with a squash merge in the end?  Unless you mean a cleaner history would help with HIP porting -- but I was planning to roll most changes into just a few commits anyway now that we know what we're doing and aren't trying things and squashing bugs...so no real difference to me either way.

nbeams:
hip-shared should be working now. I thought I noticed some concerning performance regressions for higher orders of basis functions, so in 80267f1 I thought maybe changing the block strategy might help -- not sure yet.  Now I'm wondering if there is/was something else going on with the machine I was testing on, since the hip-gen numbers looked wrong too (and this was before I made any edits to hip-gen).  It definitely warrants some more investigation, though. 
Still need to update hip-gen -- I will go ahead and make the stylistic changes as well, to match cuda-gen.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -758,213 +38,251 @@ extern "C" int CeedCudaGenOperatorBuild(CeedOperator op) {
...
+  for (CeedInt i = 0; i < num_input_fields; i++) {
+    ierr = CeedOperatorFieldGetBasis(op_input_fields[i], &basis); CeedChkBackend(ierr);
     if (basis != CEED_BASIS_COLLOCATED) {
       ierr = CeedBasisGetData(basis, &basis_data); CeedChkBackend(ierr);
-      ierr = CeedQFunctionFieldGetEvalMode(qfinputfields[i], &emode);
+      ierr = CeedQFunctionFieldGetEvalMode(qf_input_fields[i], &eval_mode);
       CeedChkBackend(ierr);
 
       // Check for collocated gradient
-      useCollograd = useCollograd && basis_data->d_collo_grad_1d;
```
jeremylt, line 837: I think this check should be moved to the block checking for collograd and it should only happen with a grad eval mode. I'm seeing ex1/ex2 using the non collograd path when they should and I'm not sure why.

nbeams: I don't have any opinions about where this check goes, but I'm not sure I understand how it could be affecting what you're seeing -- if `d_collo_grad_1d` doesn't exist for the basis, then it can't use collograd, anyway...seems like harmless check in this case?
Do ex1/ex2 never use collograd for dim = 3?  For which parameters do you see the problem?

backends/cuda-shared/ceed-cuda-shared-basis.c
```diff
@@ -271,27 +315,34 @@ int CeedBasisCreateTensorH1_Cuda_shared(CeedInt dim, CeedInt P_1d, CeedInt Q_1d,
...
   ierr = CeedLoadSourceToBuffer(ceed, basis_kernel_path, &basis_kernel_source);
   CeedChkBackend(ierr);
   CeedDebug256(ceed, 2, "----- Loading Basis Kernel Source Complete -----\n");
-  ierr = CeedCompileCuda(ceed, basis_kernel_source, &data->module, 8,
+  ierr = CeedCompileCuda(ceed, basis_kernel_source, &data->module, 9,
                          "BASIS_Q_1D", Q_1d,
                          "BASIS_P_1D", P_1d,
-                         "BASIS_T_1D", CeedIntMax(Q_1d, P_1d),
+                         "T_1D", CeedIntMax(Q_1d, P_1d),
                          "BASIS_BUF_LEN", num_comp * CeedIntPow(Q_1d > P_1d ?
```
nbeams, line 328: I went ahead and removed `BASIS_BUF_LEN` from compile-time definitions in hip-shared -- we don't use it any more in the kernels

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
jeremylt: Logging out the value of `use_collograd_parallelization` showed that it was never set to true due to a minor ordering issue. I simplified in the latest push. I was checking with `./build/ex1-volume -ceed /gpu/cuda/gen -d 3`

nbeams:
I noticed a few stray `isStrided` and `backendstrides` in ceed-cuda-gen-operator-build.cpp -- the preferred versions are now `is_strided` and `has_backend_strides`, correct?

jeremylt:
Woops, yeah. Let me get those.

nbeams:
Still to do: update the release notes.
I didn't add anything about the performance regression for hip-shared yet.  Maybe I should check some orders higher than p = 8, since that's where the trouble starts?

jeremylt:
I can add some release notes.

Getting a handle on the performance impact would be good - I think it could help us get a sense of what orders we can expect reasonable perf for with initial 3d non-tensor elements based on the 1d kernels here

nbeams:
I'm not sure how it would translate to non-tensor elements...  I just want to see if it's a consistent trend, that < 8 is faster than before, and >= 8 is slower (for hip-shared)

nbeams:
Okay, duh, just realized that p = 9 will fail for current hip-shared because the weight kernels uses a threadblock of `Q_1d x Q_1d x Q_1d` = `11 x 11 x 11` = 1331, too big.  This was already true -- not introduced by this PR.

nbeams:
I think 90a7156195b5c9a5622a8d1809434aa9002bf2de should take care of the weight block size issue.  So, we've lost performance for hip-shared when p = 8, but gained the ability to use orders > 8.

nbeams:
I did a quick 3D diffusion test on a V100, and the refactored cuda-shared looked good for p = [1,8], ranging from 10% to 40% speedup over current main.  Just some quick tests on a machine without a queue/ability to guarantee exclusive jobs, though, so not the best data (and it would've been better to do A100, but I already had stuff built on the V100 machine), but nothing concerning yet, at least.

jedbrown:
Cool, thanks for testing. And to confirm, we expect no changes in -gen performance here? Is the PR ready to squash-merge?

jeremylt:
Correct - the only change in -gen should be fixing a regression from a logic error, so we should be good for squash+merge

jedbrown (approved):

PR #1051, opened 2022-08-25
Switch to clang-format

### LINKED ISSUE #1047: clang-format
Ratel and PETSc have both moved to clang-format and I'm starting to feel envious when working on libCEED. Shall we do it? And if so, what config should we use? For reference:
* https://gitlab.com/micromorph/ratel/-/blob/main/.clang-format
* https://gitlab.com/petsc/petsc/-/blob/main/.clang-format

jeremylt:
closes #1047

Squash+merge


Makefile
```diff
@@ -673,19 +674,33 @@ doc-html doc-latexpdf doc-epub doc-livehtml : doc-% : doxygen
...
+	$(CLANG_FORMAT) $(FORMAT_OPTS) $^
+
+format.ch := $(filter-out include/ceedf.h $(wildcard tests/t*-f.h), \
+            $(wildcard include/*.h include/*/*.h include/*/*/*.h include/*/*/*/*.h \
+              interface/*.[ch] tests/*.[ch] backends/*/*.[ch] backends/*/*/*.[ch] backends/*/*.[ch]pp backends/*/*/*.[ch]pp \
+              examples/*/*/*.[ch] examples/*/*.[ch] examples/*/*.[ch]pp gallery/*.[ch] gallery/*/*.[ch]))
+
+format-c  :  $(format.ch:%=%.format)
+
+style-c   : format-c
```
jedbrown, line 692: Let's pick one name (`format` or `style`). I don't think the backward compatibility is worth having multiple aliases at this stage.

Makefile
```diff
@@ -673,19 +674,33 @@ doc-html doc-latexpdf doc-epub doc-livehtml : doc-% : doxygen
...
+
+FORMAT_OPTS += -style=file -i
+
+%.format : %
+	$(CLANG_FORMAT) $(FORMAT_OPTS) $^
+
+format.ch := $(filter-out include/ceedf.h $(wildcard tests/t*-f.h), \
+            $(wildcard include/*.h include/*/*.h include/*/*/*.h include/*/*/*/*.h \
+              interface/*.[ch] tests/*.[ch] backends/*/*.[ch] backends/*/*/*.[ch] backends/*/*.[ch]pp backends/*/*/*.[ch]pp \
+              examples/*/*/*.[ch] examples/*/*.[ch] examples/*/*.[ch]pp gallery/*.[ch] gallery/*/*.[ch]))
```
jedbrown, line 688: Maybe collect using a `git ls-files *.c *.h *.cpp *.hpp`?

Makefile
```diff
@@ -673,19 +674,33 @@ doc-html doc-latexpdf doc-epub doc-livehtml : doc-% : doxygen
...
-          $(filter-out include/ceedf.h $(wildcard tests/t*-f.h), \
-            $(wildcard include/*.h interface/*.[ch] tests/*.[ch] backends/*/*.[ch] \
-              examples/*/*/*.[ch] examples/*/*.[ch] examples/*/*.[ch]pp gallery/*/*.[ch]))
+# Style/Format
+CLANG_FORMAT ?= clang-format
+
+FORMAT_OPTS += -style=file -i
+
+%.format : %
+	$(CLANG_FORMAT) $(FORMAT_OPTS) $^
```
jedbrown, line 683: This is a phony target so it's going to run on every file every time (even if nothing has been edited). Seems to me it may as well be one invocation of `clang-format` rather than one per file.

jedbrown (commented):
Thanks for preparing this. I think we should clear out nearly-ready PRs before merging it.

.github/workflows/c-fortran-test-style.yml
```diff
@@ -15,8 +15,8 @@ jobs:
     steps:
     - name: Environment setup
       uses: actions/checkout@v2
-    - name: Install astyle
-      run: sudo apt-get install astyle
+    - name: Setup clang-format
+      uses: jidicula/clang-format-action@v4.8.0
```
jeremylt, line 19: This isn't quite right. I'll have to mess around with this

Makefile
jeremylt: I have zero preferences. Opinions? Ratel has both. Not sure what PETSc uses, but that could be the tiebreaker

jedbrown: PETSc `make clangformat`
MFEM `make style` (uses astyle)
Apache Arrow `make format` (uses clang-format)

It doesn't matter to me because I'll very rarely run it manually (that's my editor's job).

Makefile
jeremylt: At least on my machine, `make format -j` goes faster with multiple targets, unless I'm missing something.

jedbrown: What does this look like for you?
```console
$ time make -j12 format > /dev/null
4.659 real   11.453 user   6.545 sys   386.38 cpu
$ time clang-format -i $(git ls-files *.[ch]pp *.[ch])
2.293 real   2.256 user   0.036 sys   99.99 cpu
```

jedbrown: We don't want a hard dependency on GNU parallel, but this is faster. I almost never call this directly so I don't have strong opinions.
```console
$ time ( git ls-files *.[ch]pp *.[ch] | parallel -j8 -N 40 clang-format -i )
0.903 real   3.053 user   0.303 sys   371.55 cpu
```

jeremylt: Hmm, I must have mangled something in the attempt

jeremylt:
Any other PRs we want to get in first?

.clang-format
```diff
@@ -0,0 +1,14 @@
+BasedOnStyle: "Google"
+
+AlignAfterOpenBracket:        Align
+AlignArrayOfStructures:       Left
+AlignConsecutiveAssignments:  AcrossComments
```
jedbrown, line 5: Can we restrain ourselves a bit on vertical alignment, at least this much?
```suggestion
AlignConsecutiveAssignments:  Consecutive
```

jedbrown (approved):
I think we can do this and port the in-progress branches up. Possible exceptions

* #944 (@rezgarshakeri ) could be merged at its working-but-not-fast state
* #948 (@nbeams ) touches a lot of code and may be hard to merge/rebase, but it's unclear to me if it can be merged soon without disrupting correctness or performance in the single-precision code paths.

rezgarshakeri (approved):

PR #1052, opened 2022-08-26
fluids: Switch to StateVar enum

jrwrigh:
Specify the state variables to be used by an enum rather than booleans. Makes the code more explicit about it's intent. 

Related: https://github.com/CEED/libCEED/pull/1035#discussion_r954500678

Todo:
- [x] update documentation


jedbrown (approved):
I just see this stray.
```
examples/fluids/problems/newtonian.c
244:            "RHSFunction is not provided for primitive variables (use -primitive only with -implicit)\n");
```

PR #1053, opened 2022-08-26
Fix small memory leaks in JIT source code management

nbeams:
I introduced memory leaks in the MAGMA backend when switching to RTC (did not explicitly free the buffers used in loading source files).  In the course of testing, I found another (very minor) missing `CeedFree` that affects all backends using JIT.
This PR adds these `CeedFree` calls.

I'm not sure I've quite squashed all the potential leaks, though.  Valgrind is still reporting slight differences between `/gpu/hip/ref`, `/gpu/hip/magma`, and `/cpu/self/ref/serial` for some tests that I tried.  Though -- even cpu-ref says some memory is "definitely lost" -- is this something anyone else has seen before?  (E.g., t205-elemrestriction: cpu-ref, 337,984 bytes; gpu-ref, 338,512 bytes; magma, 338,589).
Edit: I see now that linking hipblas and/or hipsparse will create reported memory leaks, even without using the GPU.  Have confirmed through a test that doesn't use libCEED at all.

### DIFF
--- a/backends/magma/ceed-magma-basis.c
+++ b/backends/magma/ceed-magma-basis.c
@@ -757,6 +757,11 @@ int CeedBasisCreateTensorH1_Magma(CeedInt dim, CeedInt P1d, CeedInt Q1d,
                   data->queue);
 
   ierr = CeedBasisSetData(basis, impl); CeedChkBackend(ierr);
+  ierr = CeedFree(&magma_common_path); CeedChkBackend(ierr);
+  ierr = CeedFree(&interp_path); CeedChkBackend(ierr);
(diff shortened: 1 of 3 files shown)


jeremylt:
5061f376ba061780f00b05efb4dc5b34d588e0ef
Should fix the CI issue so this can run on Noether

jeremylt (approved):
I like it so long as we can cherry pick over that CI fix and check that Noether likes the changes

jeremylt:
Can you `--trace-origins=yes` the 205 leak on ref? I'm not seeing an obvious cause.

jeremylt:
sorry, CI is a mess right now. I think we can just remove that cherry-picked commit and merge

PR #1057, opened 2022-08-28
fluids: Fix time label for strong STG

jrwrigh:
The solution time label for the strong STG boundary condition was not created or set. The strong form has it's own operator, and so needs it's own label. 

PR also includes some niceties from Jed to the fluids makefile.


jeremylt (approved):
Looks good to me - I manually ran `$ make prove search=flu -j` on Noether and it passed

PR #1058, opened 2022-08-29
LIBXSMM CI Update

jeremylt:
This fixes CI with LIBXSMM on Noether

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -18,8 +18,8 @@ noether-cpu:
     - echo "-------------- HIPCC ---------------" && $HIPCC --version
     - echo "-------------- GCOV ----------------" && gcov --version
 # Libraries for backends
-# -- LIBXSMM v1.17
-    - cd .. && export XSMM_VERSION=libxsmm-1.17 && { [[ -d $XSMM_VERSION ]] || { git clone --depth 1 --branch 1.17 https://github.com/hfp/libxsmm.git $XSMM_VERSION && make -C $XSMM_VERSION -j$(nproc); }; } && export XSMM_DIR=$PWD/$XSMM_VERSION && cd libCEED
+# -- LIBXSMM 44433be9426eddaed88415646c15b3bcc61afc85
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -18,8 +18,8 @@ noether-cpu:
     - echo "-------------- HIPCC ---------------" && $HIPCC --version
     - echo "-------------- GCOV ----------------" && gcov --version
 # Libraries for backends
-# -- LIBXSMM v1.17
-    - cd .. && export XSMM_VERSION=libxsmm-1.17 && { [[ -d $XSMM_VERSION ]] || { git clone --depth 1 --branch 1.17 https://github.com/hfp/libxsmm.git $XSMM_VERSION && make -C $XSMM_VERSION -j$(nproc); }; } && export XSMM_DIR=$PWD/$XSMM_VERSION && cd libCEED
+# -- LIBXSMM main
+    - cd .. && git clone --depth 1 https://github.com/libxsmm/libxsmm.git && make -C libxsmm -j$(nproc) && export XSMM_DIR=$PWD/libxsmm && cd libCEED
```
jedbrown, line 22: Should we pin a version or do we believe that main is stable? I think we had issues with this some years ago and switched to version pinning for that reason.

jedbrown (approved):

.gitlab-ci.yml
jeremylt: They haven't have a release since last December, so I'm not sure. We could pin to a commit hash?

jedbrown: I think commit hash is better. Note that you can download a specific hash, as in https://github.com/libxsmm/libxsmm/archive/f3e9304cb9c58d5eb6ffd5a06adf41741707abea.tar.gz, without git.

jeremylt: Updated in 1a48f341b49ad324100dfdfeff4f6eec543c3c88

jedbrown (approved):

PR #1056, opened 2022-08-28
Fluids - Some fixes missing from #1039

LeilaGhaffari:
See https://github.com/CEED/libCEED/pull/1039#discussion_r956733785 and https://github.com/CEED/libCEED/pull/1039#discussion_r956734876.

PR #1059, opened 2022-08-30
Fluids - Fix blasius test

LeilaGhaffari:
@jeremylt , does this fix the issue you reported [here](https://github.com/CEED/libCEED/pull/1056#issuecomment-1231842692)?

### DIFF
--- a/examples/fluids/navierstokes.c
+++ b/examples/fluids/navierstokes.c
@@ -22,7 +22,7 @@
 //     ./navierstokes -ceed /cpu/self -problem density_current -degree 1
 //     ./navierstokes -ceed /gpu/cuda -problem advection -degree 1
 //
-//TESTARGS(name="blasius") -ceed {ceed_resource} -test -options_file examples/fluids/blasius.yaml -compare_final_state_atol 2E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius.bin -ts_max_steps 10 -dm_plex_box_faces 5,10,1
+//TESTARGS(name="blasius") -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_test.yaml -compare_final_state_atol 2E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius.bin
 //TESTARGS(name="blasius_STG") -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 2E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG.bin
 //TESTARGS(name="blasius_STG_weakT") -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG_weakT.bin -weakT
 //TESTARGS(name="blasius_STG_strongBC") -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-10 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG_strongBC.bin -stg_strong true
--- a/examples/fluids/tests-output/blasius_test.yaml
+++ b/examples/fluids/tests-output/blasius_test.yaml
@@ -1,0 +1,32 @@
+problem: 'blasius'
+
+implicit: true
+ts:
+  adapt_type: 'none'
+  type: 'beuler'
+  dt: 0.2e-5
+  max_steps: 10
+
+degree: 1
+dm_plex_box_faces: 5,10,1
+platemesh_nDelta: 45
+
+stab: 'supg'
+Ctau_t: 1
+Ctau_v: 36
+Ctau_C: 0.25
+Ctau_M: 0.25
+Ctau_E: 0.125
+
+q_extra: 0
+
+dm_plex_box_lower: 0,0,0
+dm_plex_box_upper: 4.2e-3,4.2e-3,5.e-5
+dm_plex_dim: 3
+# Faces labeled 1=z- 2=z+ 3=y- 4=y+ 5=x+ 6=x-
+bc_slip_z: 1,2
+bc_wall: 3
+wall_comps: 1,2,3
+bc_inflow: 6
+bc_outflow: 5,4
+g: 0,0,0


jeremylt (approved):
LGTM

jedbrown:
Hmm, something is amiss since I still see this on Noether. It's also an issue with avx and xsmm backends, though `/cpu/self/opt/blocked` works. :shrug: 
```console
$ make prove search=fluid PROVE_OPTS=-v BACKENDS=/cpu/self/opt/serial
[...]
# Test Name: blasius
# TESTARGS: -ceed {ceed_resource} -test -options_file examples/fluids/tests-output/blasius_test.yaml -compare_final_state_atol 2E-11 -compare_final_state_filename examples/fluids/tests-outpu
t/fluids-navierstokes-blasius.bin
not ok 1 fluids-navierstokes /cpu/self/opt/serial
ok 2 fluids-navierstokes /cpu/self/opt/serial stdout
not ok 3 fluids-navierstokes /cpu/self/opt/serial stderr
# +[0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
# +[0]PETSC ERROR: TSStep has failed due to DIVERGED_NONLINEAR_SOLVE, increase -ts_max_snes_failures or make negative to attempt recovery
# +[0]PETSC ERROR: See https://petsc.org/release/faq/ for trouble shooting.
# +[0]PETSC ERROR: Petsc Development GIT revision: v3.17.4-1168-g96596328e34  GIT Date: 2022-08-29 13:10:12 +0000
# +[0]PETSC ERROR: build/fluids-navierstokes on a mpich-O named noether by jeka2967 Tue Aug 30 22:31:10 2022
# +[0]PETSC ERROR: Configure options --download-cgns --download-ctetgen --download-exodusii --download-hdf5 --download-hypre --download-kokkos --download-kokkos-kernels --download-med --down
load-metis --download-ml --download-mumps --download-netcdf --download-parmetis --download-pnetcdf --download-scalapack --download-suitesparse --download-superlu --download-superlu_dist --do
wnload-triangle --with-debugging=0 --with-hip --with-hip-arch=gfx906 --with-zlib COPTFLAGS="-O3 -ffp-contract=fast -march=native -g" CXXOPTFLAGS="-O3 -ffp-contract=fast -march=native -g" HIP
OPTFLAGS="-O3 -ffp-contract=fast -march=native" PETSC_ARCH=mpich-O
# +[0]PETSC ERROR: #1 TSStep() at /projects/jed/petsc/src/ts/interface/ts.c:3380
# +[0]PETSC ERROR: #2 TSSolve() at /projects/jed/petsc/src/ts/interface/ts.c:3746
# +[0]PETSC ERROR: #3 TSSolve_NS() at /home/jeka2967/libCEED/examples/fluids/src/setupts.c:569
# +[0]PETSC ERROR: #4 main() at /home/jeka2967/libCEED/examples/fluids/navierstokes.c:292
```

jeremylt:
Odd, either I'm crazy or it seemed to work for me yesterday, but its failing today for me too

jrwrigh:
Whatever is causing the issue here is also causing my production runs with STG to diverge. The STG regression test itself is not failing. 

Manually bisecting, #1056 causes my solution to diverge.

#1039 also causes some odd behavior during the problem setup phase:

<details>
<summary>Details</summary>

```
  0 SNES Function norm 1.982173641657e+00
  0 SNES Function norm 1.982173641657e+00
  0 SNES Function norm 1.982173641657e+00
  0 SNES Function norm 1.982173641657e+00
  0 SNES Function norm 1.982173641657e+00
[....]
Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 2
  2 SNES Function norm 6.608226462156e-01 
  Linear solve converged due to CONVERGED_RTOL iterations 1
Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 2
Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 2
Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 2
  Linear solve converged due to CONVERGED_RTOL iterations 1
  1 SNES Function norm 1.082450171023e+00 
  Linear solve converged due to CONVERGED_RTOL iterations 1
  2 SNES Function norm 6.608226462156e-01 
  Linear solve converged due to CONVERGED_RTOL iterations 1
Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 2
  1 SNES Function norm 1.082450171023e+00 
  Linear solve converged due to CONVERGED_RTOL iterations 1
  2 SNES Function norm 6.608226462156e-01 
  Linear solve converged due to CONVERGED_RTOL iterations 1
Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 2
  Linear solve converged due to CONVERGED_RTOL iterations 1
  1 SNES Function norm 1.082450171023e+00 
  Linear solve converged due to CONVERGED_RTOL iterations 1
  2 SNES Function norm 6.608226462156e-01 
  Linear solve converged due to CONVERGED_RTOL iterations 1
Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 2
PetscViewer Object: 144 MPI processes
  type: binary
  Filename: ./output_files/ns-solution-4700.bin
  Mode: READ (stdio)

  MPI:
    Host Name                          : cc090
    Total ranks                        : 144
  Problem:
  libCEED:
  PETSc:
    Box Faces                          : 389,148,141
  Mesh:
    Global DoFs                        : 40892610
    Owned DoFs                         : 260272
    Global nodes                       : 8178522
    Owned nodes                        : 61798
PetscViewer Object: 144 MPI processes
  type: binary
  Filename: ./output_files/ns-time-4700.bin
  Mode: READ (stdio)
0 TS dt 0.00036 time 1.692
PetscViewer Object: 144 MPI processes
  type: binary
  Filename: ./output_files/ns-solution-4700.bin
  Mode: WRITE (stdio)
PetscViewer Object: 144 MPI processes
  type: binary
  Filename: ./output_files/ns-time-4700.bin
  Mode: WRITE (stdio)
    0 SNES Function norm 2.130358355712e+00 
    Linear solve converged due to CONVERGED_RTOL iterations 4
    1 SNES Function norm 7.843979087704e-03 
    Linear solve converged due to CONVERGED_RTOL iterations 11
    2 SNES Function norm 9.099087042204e-05 
    Linear solve converged due to CONVERGED_RTOL iterations 14
    3 SNES Function norm 1.668392153657e-06 
  Nonlinear solve converged due to CONVERGED_ITS iterations 3
    0 SNES Function norm 2.111441489139e+00 
    Linear solve converged due to CONVERGED_RTOL iterations 4
    1 SNES Function norm 1.147021749319e-02 
    Linear solve converged due to CONVERGED_RTOL iterations 11
    2 SNES Function norm 2.817984609228e-04 
    Linear solve converged due to CONVERGED_RTOL iterations 13
    3 SNES Function norm 7.300919533121e-06 
  Nonlinear solve converged due to CONVERGED_ITS iterations 3
    0 SNES Function norm 4.781029362284e+00 
    Linear solve converged due to CONVERGED_RTOL iterations 4
    1 SNES Function norm 2.936167200644e-01 
    Linear solve converged due to CONVERGED_RTOL iterations 7
    2 SNES Function norm 2.075207891818e-02 
    Linear solve converged due to CONVERGED_RTOL iterations 10
    3 SNES Function norm 1.990857823188e-03 
  Nonlinear solve converged due to CONVERGED_ITS iterations 3
1 TS dt 0.00036 time 1.69236
```

</details>

The first line there is repeated for every MPI rank I'm running on (144 total). Not sure if that helps narrow down the culprit or not.

LeilaGhaffari:
Sorry @jrwrigh this is messing with your runs. 
Regarding the odd behavior you mentioned above, that is happening when the Chebyshev coefficients are computed and you should be able to have it converge by changing the value of  `N_Chebyshev` (It seems that it is diverging now and that could be one of the reasons your runs are blowing up). The default value is 20 right now. Can you try `-N_Chebyshev 10`? I remember 10 was always good with my tests but I changed it to 20 without too many tests. 
The repeated lines shouldn't be causing the issue. That is happening because the Chebyshev snes solve is not in parallel and when you monitor snes, it prints out each line for every MPI rank. Maybe I should fix that but I am positive that that cannot be the issue here.

LeilaGhaffari:
@jeremylt , it was passing for me on Noether too. @jedbrown , do you think I should drop this test for now if it is messing with the CI? (I just can't think of a reason why the test should fail so randomly)

LeilaGhaffari:
Also, @jrwrigh , I don't know if it matters for your runs or not but there is a new variable in blasius.c for wall temperature (`T_wall`) which defaults to 400K but if you don't want it to be different from the freestream temperature, then you should set it to 288.

jrwrigh:
> Should `T_wall` default to the freestream temperature? Would make things less error-prone?

Do you mean that setting default `T_wall` equal to default `T_inf`, or setting `T_wall` to equal `T_inf` if `T_wall` isn't otherwise specified?

I'm down for either, maybe with the added feature that a previously set `T_wall` can be set back to the `T_inf` value by passing a negative `T_wall` (since that's unphysical). That'd be useful when overriding yaml settings files.

So `-T_wall 50 -T_inf 60 -T_wall -1` would result in `T_wall = 60`.

jedbrown:
I only mean that if `T_wall` is not set, it'll be equal to `T_inf`. I'd rather not have an ad-hoc semantic around negative arguments (confusing and hard to maintain, versus the minor convenience).

LeilaGhaffari:
OK, I will change the default value of `T_wall` and set a prefix for the Chebyshev solve.

LeilaGhaffari:
Is there a way to break the code if the snes solve doesn't converge? Right now, if the Chebyshev solve doesn't converge, the code keeps running with wrong coefficients, like [this one](https://github.com/CEED/libCEED/pull/1059#issuecomment-1233560751).

jedbrown:
Usually you would `SNESGetConvergedReason()` and raise an error explaining what's going on if it doesn't converge. There's also `SNESSetErrorIfNotConverged()`, but that could be confusing if it's not obvious to the user that this is the Chebyshev solve.

jedbrown:
(A slightly better initial guess might make the solver a lot more robust.)

jedbrown:
Also, we could experiment on the Julia side first, but we could map the domain $\eta \in (0, \infty)$ to $s \in (-1, 1)$ with a transform like $s = 2 \tanh(\alpha \eta) - 1$ for any chosen $\alpha$. That would be a small code change and avoid needing to specify $\eta_{\max}$. I'd expect $\alpha = 1/3$ would have decent approximation properties and could be chosen universally.

jedbrown:
Let's not go down that path right now. I see the test failing with fb455ff073519dc60531e3d0b72267e590b5c938, which is the first version in main. Adding `-snes_monitor` tells me that the Blasius/Chebyshev solve converges, but the time stepping SNES fails after a few time steps.

PR #1061, opened 2022-09-02
ci: use noether for CUDA

jedbrown:
Disable failing Blasius test; cf. discussion in https://github.com/CEED/libCEED/pull/1059

### DIFF
--- a/tests/t568-operator.c
+++ b/tests/t568-operator.c
@@ -66,8 +66,8 @@ int main(int argc, char **argv) {
 
   // QFunctions
   CeedQFunctionCreateInterior(ceed, 1, setup, setup_loc, &qf_setup);
-  CeedQFunctionAddInput(qf_setup, "weight", 1, CEED_EVAL_WEIGHT);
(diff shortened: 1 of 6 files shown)


tests/t568-operator.c
```diff
@@ -153,8 +152,8 @@ int main(int argc, char **argv) {
                                 comp_in*num_dofs;
           const CeedScalar assembled_value = assembled[index];
           const CeedScalar assembled_true_value = assembled_true[index];
-          if (fabs(assembled_value - assembled_true_value) >
-              100.*CEED_EPSILON)
+          if (!(fabs(assembled_value - assembled_true_value) <
+                100.*CEED_EPSILON))
```
jedbrown, line 156: @nbeams Thanks for pointing this out. The fix (wrong order of arguments) was easy. Building with `-fstack-protector` was very useful and something we should add to CI.

jedbrown:
I'll merge this because it gets CI running again and hopefully will prevent further mistakes. The t568 issue needs to be addressed.

PR #1063, opened 2022-09-02
Fluids - Blasius: Some Minor Modifications

LeilaGhaffari:
Follow up the comments in #1059.


jedbrown:
The CFD solver here fails to converge (not the Chebyshev solve), but this is the mesh. Is this a mesh transform parametrization problem?
image

examples/fluids/tests-output/blasius_test.yaml
```diff
@@ -10,6 +10,7 @@ ts:
 degree: 1
 dm_plex_box_faces: 5,10,1
 platemesh_nDelta: 45
+T_wall: 400
```
jedbrown, line 13:
```suggestion
temperature_wall: 400
```

jeremylt:
By the way...

Congrats on breaking main! 🎉

We all do it sometimes. It means you are moving the code forward.

Thanks for the quick PR to patch things!

jedbrown:
Thanks @jeremylt. Agreed, no shame in getting into this position.

I believe the main reason for breakage here is that the mesh mapping (`platemesh_ndelta` and `platemesh_growth`) need to be adjusted for the mesh resolution. I think we should talk with @jrwrigh sometime about a less error-prone way to parametrize the warping function.

LeilaGhaffari:
> The CFD solver here fails to converge (not the Chebyshev solve), but this is the mesh. Is this a mesh transform parametrization problem?

Yes, you are right. Just used the parameters in `blasius_stgtest.yaml` and it doesn't fail for me anymore. I admit I am not understanding the mesh modification for this problem.

LeilaGhaffari:
> By the way...
> 
> Congrats on breaking main! tada
> 
> We all do it sometimes. (I've lost count of how often I've done it!) It means you are moving the code forward.
> 
> Thanks for the quick PR to patch things!

Haha, thanks @jeremylt . Though it seems the main reason was my lack of understanding of the problem setup :sweat_smile:

jedbrown:
This converges; great! I'm not sure what to expect here, but it seems like the inflow solution perhaps isn't compatible?
image

jedbrown:
Oh, maybe this is just because the wall boundary condition is insulated (rather than specified temperature), thus it departs rapidly from the Blasius solution? Could we run it in primitive variables and match the boundary condition?

jrwrigh:
> I believe the main reason for breakage here is that the mesh mapping (`platemesh_ndelta` and `platemesh_growth`) need to be adjusted for the mesh resolution. I think we should talk with @jrwrigh sometime about a less error-prone way to parametrize the warping function.

Yeah, I'm certainly open to it. We had talked about using an analytic mapping function to handle the warping with `DMPlexRemapGeometry`. That could definitely replace the `platemesh_ndelta` and `platemesh_growth`, though the individual point picking of `ModifyMesh()` would still need to be there for `platemesh_y_node_locs_path`.

jedbrown:
The main flow issue here is that the Blasius solution did not create variable density flow, and that manifested as being a constant temperature solution for inflow. I've fixed that up in my push here. I also notice that `-weakT` has instabilities on this mesh. I'm not sure we have a good reason to keep it around. It would be nice to be able to do Blasius with a strong inflow.

jedbrown (approved):
This works on Noether with CUDA and HIP so I'll merge it now. Thanks.

LeilaGhaffari:
Great! Thanks for the fix!

PR #1065, opened 2022-09-06
Fix t568 bug

### LINKED ISSUE #1064: t568-operator broken on some GPU backends
This will need to be fixed prior to release. https://github.com/CEED/libCEED/pull/1061/commits/963d86364853ef9f95e8fc0edc4b112ec528ac3c

nbeams:
Fixes #1064 

Rather than something GPU-specific, I think the problem was the QFunction that the test was using.  The test itself had `num_comp=2`, but it was using the same QFunction as t534, which has only one component. 

If we want to avoid code duplication, we could probably use the QFunctionContext to set the number of components and then write t534-operator.h in such a way that it would work for t534 and t568.   Would this be preferable?

### DIFF
--- a/tests/junit.py
+++ b/tests/junit.py
@@ -61,7 +61,6 @@ def skip_rule(test, resource):
         test.startswith('t507') and contains_any(resource, ['occa']),
         test.startswith('t318') and contains_any(resource, ['/gpu/cuda/ref']),
         test.startswith('t506') and contains_any(resource, ['/gpu/cuda/shared']),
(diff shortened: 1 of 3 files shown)


jeremylt (approved):
LGTM

jeremylt:
We should also revert this change to junit.py

https://github.com/CEED/libCEED/commit/963d86364853ef9f95e8fc0edc4b112ec528ac3c

jedbrown (approved):

PR #1066, opened 2022-09-12
Lazy access of qf kernel name and path

jeremylt:
This sets up lazy building of the QF kernel name and source path to fix #1040


jeremylt:
Investigating the cargo-doc failure, but I confirmed that this works with #1040

jedbrown (approved):

jeremylt:
Rust is happy now - will merge when CI finishes

PR #1069, opened 2022-09-14
fluids: Fix STG Ek_tot calculation bug

jrwrigh:
When the spectrum is small enough, `Ek_tot` can be zero, causing a divide-by-zero NaN, which then corrupts the calculation of the velocity fluctuations later down the line. No idea why that hasn't be triggered in previous runs with this code, but it didn't. 

Instead, I set `1/Ek_tot` to some random value if `Ek_tot == 0`. The value doesn't matter since `Ek_tot ==0` iff the individual spectrum components are also zero. I'm open to a different random value than `-1111`. I just made it obviously wrong and negative (which isn't physical).

Also refactored the `dw` -> `wall_dist` as I managed to forget what that stood for while debugging, even though I'm the one who coded it.

Originally found during OLCF Crusher Hackathon 4

### DIFF
--- a/examples/fluids/qfunctions/stg_shur14.h
+++ b/examples/fluids/qfunctions/stg_shur14.h
@@ -288,13 +288,14 @@ CEED_QFUNCTION(Preprocess_STGShur14)(void *ctx, CeedInt Q,
     SpectrumConstants(dw, eps, lt, h, nu, &hmax, &ke, &keta, &kcut);
 
     // Calculate total TKE per spectrum
-    stg_data[i] = 0.;
+    CeedScalar Ek_tot=0;
     CeedPragmaSIMD
     for(CeedInt n=0; n<nmodes; n++) {
       const CeedScalar dkappa = n==0 ? kappa[0] : kappa[n] - kappa[n-1];
-      stg_data[i] += Calc_qn(kappa[n], dkappa, keta, kcut, ke, 1.0);
+      Ek_tot += Calc_qn(kappa[n], dkappa, keta, kcut, ke, 1.0);
     }
-    stg_data[i] = 1/stg_data[i];
+    // avoid underflowed and poorly defined spectrum coefficients
+    stg_data[i] = Ek_tot != 0 ? 1/Ek_tot : 0;
   }
   return 0;
 }
--- a/examples/fluids/qfunctions/stg_shur14.h
+++ b/examples/fluids/qfunctions/stg_shur14.h
@@ -27,39 +27,39 @@
 /*
  * @brief Interpolate quantities from input profile to given location
  *
- * Assumed that prof_dw[i+1] > prof_dw[i] and prof_dw[0] = 0
- * If dw > prof_dw[-1], then the interpolation takes the values at prof_dw[-1]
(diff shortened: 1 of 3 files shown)


examples/fluids/qfunctions/stg_shur14.h
```diff
@@ -284,17 +288,17 @@ CEED_QFUNCTION(Preprocess_STGShur14)(void *ctx, CeedInt Q,
...
-    stg_data[i] = 0.;
+    CeedScalar Ek_tot=0;
     CeedPragmaSIMD
     for(CeedInt n=0; n<nmodes; n++) {
       const CeedScalar dkappa = n==0 ? kappa[0] : kappa[n] - kappa[n-1];
-      stg_data[i] += Calc_qn(kappa[n], dkappa, keta, kcut, ke, 1.0);
+      Ek_tot += Calc_qn(kappa[n], dkappa, keta, kcut, ke, 1.0);
     }
-    stg_data[i] = 1/stg_data[i];
+    stg_data[i] = Ek_tot != 0 ? 1/Ek_tot : -1111;
```
jedbrown, line 301: A comment here would be better than a magic number, and it's not clear to me why this structure can't generate denormals so perhaps:
```suggestion
    stg_data[i] = 1 / Max(Ek_tot, 1e-20); // avoid denormals and infinity for vanishing Ek_tot
```

jedbrown (approved):
LGTM

examples/fluids/qfunctions/stg_shur14.h
jrwrigh: > it's not clear to me why this structure can't generate denormals

The reason is that the resulting individual wavemode energies is less than what's representable by denormals. For the case 
the case that I've been looking at in the debugger, the `feta_x_fcut` value is $e^{-729} \approx 10^{-317}$. That is then multiplied by 1e-15 twice, which makes it underflow even denormal numbers.


Regarding the change:
Agree on adding a comment.

I'm not sure how much I like using `Max`, as there isn't a direct connection between the magnitude of `Ek_tot`and the magnitude of the velocity fluctutations. ie. You could have a spectrum with a small `Ek_tot`, but it's corresponding velocity fluctuation magnitude (which is set by the Reynolds stress) could be large. `Ek_tot` is calculated to normalize the model energy spectrum, but doesn't have any physical meaning on it's own.

Would this work for you?

```suggestion
    stg_data[i] = Ek_tot != 0 ? 1/Ek_tot : -1111; // avoid underflow and poorly defined spectrum
```

I'd also be down for changing the magic number to `0`, which would force the fluctuations at that location to be zero, since the spectrum at that location is not representable anyways.

jedbrown: We chatted in the hall today. Simple is good so long as it's documented. And so long as each of the numeric values is non-negative, the above seems fine.

jrwrigh: I've changed the "magic" number after thinking about the discussion in the hall (specifically) the associativity. So in case (for whatever reason) the spectrum coefficients calculated later do not underflow for some reason, they're ignored. Thought being at that point, the spectrum itself is poorly defined.

PR #1077, opened 2022-10-20
CI MacOS 12 Update

jeremylt:


jeremylt:
I want to let CI finish, but this should fix the MacOS error

jedbrown (approved):

jeremylt:
Ready to merge - since this changes the 'required' test runs, it requires admin privileges to squash+merge @jedbrown

.github/workflows/c-fortran-test-hardware.yml
```diff
@@ -7,8 +7,8 @@ jobs:
   test:
     strategy:
       matrix:
-        os: [ubuntu-20.04]
-        compiler: [gcc-9]
+        os: [ubuntu-22.04]
+        compiler: [gcc-11]
         arch: [aarch64, ppc64le]
         distro: [ubuntu20.04]
```
jedbrown, line 13: Seems weird to use distro 20.04 with os 22.04?

jedbrown (commented):
Should we update the python jobs too?

.github/workflows/c-fortran-test-hardware.yml
jeremylt: As I understand it, the parent container will use 22.04 and run the tests in a child image with 20.04

jeremylt:
> Should we update the python jobs too?


Might as well

jedbrown:
Yeah, I was just looking through those required checks and it's a somewhat odd mix. Maybe that's fine/good, but I don't know what it means to use a different `distro` versus `os`.

.github/workflows/c-fortran-test-hardware.yml
jeremylt: Bumping the run-on-hardware action version let me sync those two

jedbrown:
Looks good to squash-merge now.

jeremylt:
Agreed, good to squash+merge

PR #1079, opened 2022-10-26
Update action versions

jeremylt:


jedbrown (approved):
Is there a maintained alternative to `actions-rs/tarpaulin`? Looks like it's the same person and also hasn't been touched in years.

jeremylt:
> Is there a maintained alternative to actions-rs/tarpaulin? Looks like it's the same person and also hasn't been touched in years.

I haven't found one yet

Issue #1081, opened 2022-10-28
examples/fluids: unstructured distance function


jedbrown:
Computing distance from walls is necessary for some turbulence models and to create initial conditions for which the initial transient is not too extreme. The Eikonal equation offers a proper way to solve this equation, but it's difficult to solve scalably using variational methods. [This paper](https://arc.aiaa.org/doi/pdf/10.2514/1.8626) compares several strategies and suggests a simple method:

1. Solve $-\nabla^2 \phi = 1$ with boundary condition $\phi=0$ on no-slip walls.
2. Compute distance as $d = \sqrt{\lVert \nabla \phi \rVert^2 + 2\phi} - \lVert \nabla \phi \rVert$.

This is accurate near walls, but not in the far field (where accurate distance isn't physically important). This requires solving one Poisson system and then evaluating the solution at quadrature points, which is where distance is needed in a turbulence model. If "nodal" distance is required, it can be projected into a nodal space using a (lumped) $L^2$ projection.

jeremylt:
Looks like this has moved to HONEE: https://gitlab.com/phypid/honee/-/merge_requests/159

PR #1082, opened 2022-10-28
Vortex shedding / flow past cylinder 

AdelekeBankole:


jedbrown (commented):
It looks like you still have some compilation errors to go through. Let us know when the code is ready to review or if you're stuck trying to understand a compiler error or warning.

AdelekeBankole:
Yes, somethings are still missing in there. Thank you, I will let you know.

examples/fluids/vortexshedding.yaml
```diff
@@ -33,9 +33,8 @@ dm_plex_box_faces: 50,20,1
...
 # Boundary Settings
-bc_slip_z: 36, 112, 80, 120
-bc_wall: 94, 28, 50, 72
-bc_inflow: 102
-bc_freestream: 116
-wall_comps: 37, 1, 4, 103, 3, 81, 2, 59, 5, 125 #not sure of this!
+bc_slip_z: 6
+bc_wall: 5
+bc_freestream: 1,2,3,4
+wall_comps: 1,2,3
```
jedbrown, line 39: These are the "physical labels" in the mesh. The wall components are those that are constrained at the wall. 1,2,3 means all three components of velocity. If you add `4`, then it'll make it also Dirichlet temperature. You can explain this better in the docs once you have it working. BTW, I notice that your code isn't compiling yet, but not sure if you need help interpreting the compiler errors (we can chat on Zulip).

jedbrown:
I'd suggest merging `main` into this branch (and we'll plan to squash-merge when it's done).

examples/fluids/problems/vortexshedding.c
```diff
@@ -0,0 +1,148 @@
...
+    PetscReal domain_min[3], domain_max[3], domain_size[3];
+    ierr = DMGetBoundingBox(dm, domain_min, domain_max); CHKERRQ(ierr);
+    // compute L and H
+    for (PetscInt i=0; i<3; i++) domain_size[i] = domain_max[i] - domain_min[i];
+    L = domain_size[0]*meter;
+    H = domain_size[1]*meter;
+    center = 0.5 * H; // this should be checked
+  }
+  // Some properties depend on parameters from NewtonianIdealGas
+  CeedQFunctionContextGetData(problem->apply_vol_rhs.qfunction_context,
```
jedbrown, line 110: You're getting an error saying you're trying to access this context while its locked, because you already got access (that you haven't restored yet) a few lines above.

Delete the second occurrence to fix this issue.

jedbrown (commented):
You'll be getting another error because the mesh you produced is only a mesh of the surface, not of the volume. As such, the labels you have are `Cell Sets` (because the topological dimension of the mesh is 2) instead of `Face Sets`. You need to press the `3` key or use `gmsh -3` to create a mesh with 3-dimensional elements. Or label the physical volume. I don't know which it is without seeing your `.geo` file.

jedbrown:
Also, you should merge main into your branch and enable auto-formatting in vscode. It should automatically use clangd/clang-format on save. 
image

AdelekeBankole:
Yes! I will make the mesh with quadratic elements.

jedbrown:
Can you try running with quadratic elements and making a video similar to what I did above?

For the `.geo` file, could you add parametrization like in this example? It'll give a widget on the side of the GUI and make it possible to generate different meshes via different command line arguments (see `-order` and `-setnumber` in the man page). I think we should only commit the `.geo` file in the libCEED repository, but we can have Makefile targets to create the specific meshes. https://gitlab.com/phypid/meshes/-/blob/main/annulus-cylinder.geo

examples/fluids/vortexshedding.yaml
```diff
@@ -0,0 +1,55 @@
...
+dm_plex_filename: examples/fluids/cylinder.msh
+
+# Boundary Settings
+bc_slip_z: 6
+bc_wall: 5
+bc_freestream: 1,2
+bc_slip_y: 3,4
+wall_comps: 1,2,3
+
+# TODO: fix code to convert to the state variables being used
```
jedbrown, line 43: Reminder :point_up:

jedbrown (approved):
There are some headers not normally included in qfunctions that are breaking GPU runs with error messages like
```
home/jed/src/libCEED/interface/ceed-jit-tools.c:91 in CeedLoadSourceToInitializedBuffer(): Couldn't open source file: /home/jed/src/libCEED/examples/fluids/problems/../qfunctions/ceed/ceed-f64.h
```

I don't think we need anything special to this example in a qfunction so deleting the code is probably the simplest way to fix this. The code right now is also weird because it acts like something will be needed and then just initializes zero velocity. In the future, we could have an initial condition that uses a distance function computed on unstructured meshes (see #1081).

examples/fluids/vortexshedding.yaml
AdelekeBankole: So this is the initial condition I have which should go zero at the cylinder walls
```
CeedPragmaSIMD for (CeedInt i = 0; i < Q; i++) {
    CeedScalar       U[5]  = {0.};
    CeedScalar       qi[5] = {0.};
    const CeedScalar x[3]  = {X[0][i], X[1][i], X[2][i]};

    U[0] = S_infty.U.density;
    U[1] = U[0] * tanh(Min(0, fabs(x[0] - center) - radius) / D);
    U[2] = U[0] * tanh(Min(0, fabs(x[1] - center) - radius) / D);
    U[3] = U[0] * tanh(Min(0, fabs(x[2] - center) - radius) / D);
    U[4] = S_infty.Y.pressure / (gamma - 1) + e_kinetic;

    State InitCond = StateFromU(newtonian_ctx, U, x);
    StateToQi(newtonian_ctx, InitCond, qi);

    for (CeedInt j = 0; j < 5; j++) q0[j][i] = qi[j];

  }
```
This initial condition should be converted to conservative variables?

examples/fluids/src/cloptions.c
```diff
@@ -131,9 +131,6 @@ PetscErrorCode ProcessCommandLineOptions(MPI_Comm comm, AppCtx app_ctx, SimpleBC
   // Inflow BCs
   bc->num_inflow = 16;
   PetscCall(PetscOptionsIntArray("-bc_inflow", "Face IDs to apply inflow BC", NULL, bc->inflows, &bc->num_inflow, NULL));
-  // Outflow BCs
-  bc->num_outflow = 16;
-  PetscCall(PetscOptionsIntArray("-bc_outflow", "Face IDs to apply outflow BC", NULL, bc->outflows, &bc->num_outflow, NULL));
```
jedbrown, line 136: Why delete outflow boundaries? I know we like to use `bc_freestream` for convenience and to avoid reflections, but some tests deliberately use outflow.

AdelekeBankole: Alright, I will return the outflow boundaries.

jedbrown: I'm working in your branch and will push shortly with everything working (and these outflows restored). But there's a workflow issue here.
1. You should run tests with every commit. `make test search=fluids BACKENDS=/cpu/self` runs in half a minute or so and would show that deleting the outflows broke a bunch of examples that used outflows. Those examples would need to be updated in the same commit. (I happen to think it's useful to keep around.)
2. The documentation still talks about outflow conditions. We don't make commits that break documentation so if you really were removing it, you'd need to update the documentation.
3. A major functional change like that needs to be explained in a commit message, but the commit message didn't state that.

AdelekeBankole: Ok, I will wait for your push. I will keep in mind to run test before any commit.

examples/fluids/problems/freestream_bc.c
```diff
@@ -104,8 +97,7 @@ PetscErrorCode FreestreamBCSetup(ProblemData *problem, DM dm, void *ctx, Newtoni
   CeedQFunctionContextCreate(user->ceed, &freestream_context);
   CeedQFunctionContextSetData(freestream_context, CEED_MEM_HOST, CEED_USE_POINTER, sizeof(*freestream_ctx), freestream_ctx);
   CeedQFunctionContextSetDataDestroy(freestream_context, CEED_MEM_HOST, FreeContextPetsc);
-  problem->apply_freestream.qfunction_context          = freestream_context;
-  problem->apply_freestream_jacobian.qfunction_context = freestream_context;
```
jedbrown, line 108: @jrwrigh I think this is your double free bug when multiple faces have freestream boundary conditions.

jedbrown:
Next steps @AdelekeBankole before we can merge.
- [ ] Remove `cylinder.msh` (you can keep the `cylinder.geo` if you like, but put it in a subdirectory)
- [ ] Write documentation for this example
- [ ] Run linear and quadratic elements on the coarse and fine meshes and compute the shedding period (or equivalently, Strouhal number) for each case. The Shakib paper does linear elements on the "fine" mesh. You can run these on Noether.
- [ ] Either rebase to eliminate the storage for the `cylinder.msh` or we squash-merge this branch.

examples/fluids/vortexshedding.yaml
```diff
@@ -0,0 +1,54 @@
+problem: newtonian
```
jedbrown, line 1: This now works as a generic model (any domain and BCs), accepting a reference state (initial condition) and freestream state (inherits from reference). In this example, reference is still, but freestream has velocity [1, 0, 0].

AdelekeBankole:
Thank you very much @jedbrown. I will get to these next steps tomorrow morning.

AdelekeBankole:
@jedbrown, in parameterization of my  ```cylinder.geo``` above. My Gmsh application would crash when I pass the variables ```xL = 4.5; xR = 15.5; H = 4.5;``` Any thoughts why this happening?

jedbrown:
If it's with the pushed version, `xL`, `xR`, and `H` are overwritten on line 38. Please give a reproducible test case.

AdelekeBankole:
So, this is what I mean from the pushed version. When I comment out line 38. Gmsh would crash.

jedbrown:
This works fine for me. How were you setting the values? Previous defaults were kinda nonsense.
```diff
diff --git i/examples/fluids/meshes/cylinder.geo w/examples/fluids/meshes/cylinder.geo
index e53dbba8e..81da75efc 100644
--- i/examples/fluids/meshes/cylinder.geo
+++ w/examples/fluids/meshes/cylinder.geo
@@ -10,17 +10,17 @@ Nc  = N/2 + 1; Rc = 1.00;
 zspan = 0.2;
 
 DefineConstant[
-  H = {.5, Min .1, Max 4.5, Step .1,
+  H = {4.5, Min .1, Max 4.5, Step .1,
     Name "Parameters/Height"}
 ];
 
 DefineConstant[
-  xL = {.5, Min .1, Max 4.5, Step .1,
+  xL = {4.5, Min .1, Max 4.5, Step .1,
     Name "Parameters/XL"}
 ];
 
 DefineConstant[
-  xR = {.5, Min .1, Max 15.5, Step .1,
+  xR = {15.5, Min .1, Max 15.5, Step .1,
     Name "Parameters/XR"}
 ];
 
@@ -35,7 +35,6 @@ DefineConstant[
 ];
 
 // Exterior corners
-xL = 4.5; xR = 15.5; H = 4.5;
 Point(1) = {-xL, -H, 0, 1.0};
 Point(2) = {xL, -H, 0, 1.0};
 Point(3) = {xR, -H, 0, 1.0};
```

AdelekeBankole:
Pls, I will get back to you when I get home, just discovered its snowing outside. Hopefully, I can catch my bus home.

examples/fluids/meshes/cylinder.geo
```diff
@@ -0,0 +1,154 @@
...
+    Name "Parameters/XL"}
+];
+
+DefineConstant[
+  xR = {15.5, Min .1, Max 15.5, Step .1,
+    Name "Parameters/XR"}
+];
+
+DefineConstant[
+  cyldiameter = {1.41421356, Min .1, Max 1.41421356, Step .1,
```
jedbrown, line 28: Please fix up how this value is used. It is not the diameter as currently expressed. The default will be 1 once you fix how it gets used.

examples/fluids/meshes/cylinder.geo
```diff
@@ -0,0 +1,154 @@
...
+    Name "Parameters/CylDiameter"}
+];
+
+DefineConstant[
+  resparam = {xL + xR, Min .1, Max xL + xR, Step .1,
```
jedbrown, line 33: This is never used (but should be). Ideally explain what it means.

examples/fluids/meshes/cylinder.geo
```diff
@@ -0,0 +1,154 @@
+// Created Thursday, October 13, 2022.
+// Number of elements
+N = 40;
+Nx1 = N/2 + 1; Rx1 = 1.00;
+Nx2 = N/2 + 1; Rx2 = 1.00;
+Ny  = N/2 + 1; Ry = 2.00;
+Nb  = N/2 + 1; Rb = 0.9;
+Nc  = N/2 + 1; Rc = 1.00;
+
+zspan = 0.2;
```
jedbrown, line 10: Should be a parameter.

examples/fluids/meshes/cylinder.geo
```diff
@@ -0,0 +1,154 @@
...
+// Exterior corners
+Point(1) = {-xL, -H, 0, 1.0};
+Point(2) = {xL, -H, 0, 1.0};
+Point(3) = {xR, -H, 0, 1.0};
+Point(4) = {-xL, H, 0, 1.0};
+Point(5) = {xL, H, 0, 1.0};
+Point(6) = {xR, H, 0, 1.0};
+
+// Coordinates for Cylinder points
+Point(7) = {-cyldiameter/4, -cyldiameter/4, 0, 1.0};
```
jedbrown, line 46: That is, this isn't the formula in terms of a diameter.

examples/fluids/meshes/cylinder.geo
```diff
@@ -0,0 +1,154 @@
...
+    Name "Parameters/ResParam"}
+];
+
+// Exterior corners
+Point(1) = {-xL, -H, 0, 1.0};
+Point(2) = {xL, -H, 0, 1.0};
+Point(3) = {xR, -H, 0, 1.0};
+Point(4) = {-xL, H, 0, 1.0};
+Point(5) = {xL, H, 0, 1.0};
+Point(6) = {xR, H, 0, 1.0};
```
jedbrown, line 43: I think you can remove the fourth argument (characteristic length) because you aren't using a mesh generation algorithm that respects it.

examples/fluids/meshes/cylinder.geo
AdelekeBankole: I divided by 4 to get the four arcs that combine to form the circular cylinder.

examples/fluids/meshes/cylinder.geo
```diff
@@ -7,7 +7,10 @@ Ny  = N/2 + 1; Ry = 2.00;
 Nb  = N/2 + 1; Rb = 0.9;
 Nc  = N/2 + 1; Rc = 1.00;
 
-zspan = 0.2;
+DefineConstant[
+  zspan = {0.2, Min .1, Max 0.2, Step .1,
+    Name "Parameters/Zspan"}
+];
```
jedbrown, line 13: Let's allow it to get big so we can have multiple layers in the z direction, for a more genuinely 3D problem that transitions to turbulence at lower Reynolds number than the 2D problem.

examples/fluids/meshes/cylinder.geo
```diff
@@ -35,23 +38,23 @@ DefineConstant[
...
+Point(7) = {-cyldiameter/4, -cyldiameter/4, 0};
 //+
-Point(8) = {cyldiameter/4, -cyldiameter/4, 0, 1.0};
+Point(8) = {cyldiameter/4, -cyldiameter/4, 0};
 //+
-Point(9) = {-cyldiameter/4, cyldiameter/4, 0, 1.0};
+Point(9) = {-cyldiameter/4, cyldiameter/4, 0};
 //+
-Point(10) = {cyldiameter/4, cyldiameter/4, 0, 1.0};
+Point(10) = {cyldiameter/4, cyldiameter/4, 0};
```
jedbrown, line 55: Basic trigonometry question: If you have a circle of radius $r = d/2$ centered at the origin, what is the coordinate of the point at a 45 degree angle ($\pi/4$ radians) on the unit circle?

examples/fluids/meshes/cylinder.geo
```diff
@@ -35,23 +38,23 @@ DefineConstant[
 ];
```
jedbrown, line 38: `resparam` is still not being used.

examples/fluids/meshes/cylinder.geo
AdelekeBankole: The coordinate at 45 degree on a unit circle will be $(\sqrt{2}/2, \sqrt{2}/2)$

examples/fluids/meshes/cylinder.geo
AdelekeBankole: How big should we get it?

examples/fluids/meshes/cylinder.geo
AdelekeBankole: I think it should be used in computing ```Layers```. But I am not sure what variable ```resparam``` should divide

examples/fluids/meshes/cylinder.geo
jedbrown: You can leave the default, but make `Max 10` (or interpret it as up to 10 times the cylinder diameter). Then populate with however many layers are needed for the elements to have roughly equal aspect to the wake region.

examples/fluids/meshes/cylinder.geo
jedbrown: You need something like `N` as user-visible control on resolution in the plane. Once you have that, you'll have a characteristic element length, I think intended to be $h = 2H/N$, and can calculate layers as `Ceil(zspan / h)`. (I think it was confusing when you made the quadratic mesh that you made everything `N/2 + 1` so there are actually `N/2` elements along each length.)

AdelekeBankole: By "user-visible control" 
Do you mean ```N``` should be a parameter as well?

jedbrown: I did it and fixed other issues like the cylinder being the wrong size because the function [is named `Sqrt`](https://gmsh.info/doc/texinfo/gmsh.html#index-Sqrt-_0028-expression-_0029) not `sqrt`. I think `cylinder.geo` is good. You should take a few minutes to open the mesh and move the sliders to see that it's sensible.

examples/fluids/index.md
```diff
@@ -432,6 +432,9 @@ The simulation produces a strong acoustic wave and leaves behind a cold thermal
...
 This problem can be fixed using a more sophisticated Riemann solver such as HLLC {cite}`toro2009` (option `-freestream_riemann hllc`, which is default), which is a linear constant-pressure wave that transports temperature and transverse momentum at the fluid velocity.
 
+## Vortex Shedding - Flow past Cylinder
+This test case is taken from {cite}`shakib1991femcfd`. At time $t=0$, we introduce a circular cylinder with diameter $D$ centered at $x = y = 0$ in a uniform flow. The viscosity is assumed constant. We use the computational domain $-4.5 \leq x \leq 15.5$, $-4.5 \leq y \leq 4.5$. On the inflow boundary $(x = -4.5)$, pressure, velocity and temperature are prescribed. On the top and bottom boundaries $(y = \pm 4.5)$, symmetry condition, zero normal velocity component, zero heat-flux are imposed. On the outflow boundary $(x = 15.5)$, zero traction is prescribed. On the cylinder wall, a no-slip boundary condition, no heat-flux condition are imposed. Boundary conditions are freestream at inflow and outflow boundaries. As we evolve in time, eddies appear past the cylinder leading to a vortex shedding which is known as the vortex street.
```
jedbrown, line 436: 1. What is the initial condition? It is not the same as the freestream condition.
2. You have described the inflow and outflow conditions incorrectly at length (they're physically invalid subsonic inflow and outflow conditions, but we also aren't using what the code describes as `inflow` and `outflow`) and then contradict what you just wrote by saying they are "freestream" (correct!).
3. State the Reynolds number and Mach number as currently configured.

PR #1083, opened 2022-11-02
ci - update codecov action version

jeremylt:
I thought we had Julia codecov uploaded at some point, but it seems we don't anymore?


jrwrigh (approved):
LGTM

jrwrigh:
Looks like there are some issues with PETSc on Noether.

jeremylt:
I probably need to update PETSc for CUDA and ROCm

jrwrigh:
The issue is that the linker can't find petsc, so it's failing on the CPU tests.

PR #1086, opened 2022-11-05
fluids: Make simulation continuation more flexible

jrwrigh:
This branch does two primary things:
- Allows for the binary file used to continue the simulation to be explicitly specified
- Optionally appends the step number to the output binary files (both `ns-solution` and `ns-time`)

Combined, these allow for more flexible restart options (being able to continue at an arbitrary point in previous history, namely).
This also leaves open the option to select different types of files to restart from in the future (HDF5, CGNS, etc). 

Todo:
- [x] Document new options

### DIFF
--- a/examples/fluids/README.md
+++ b/examples/fluids/README.md
@@ -72,13 +72,25 @@ The following options are common among all problem types:
   - Frequency of output, in number of steps. `0` has no output, `-1` outputs final state only
   - `10`
 
-* - `-continue`
-  - Continue from previous solution
-  - `0`
-
 * - `-output_dir`
   - Output directory
   - `.`
+
+* - `-output_add_stepnum2bin`
+  - Whether to add step numbers to output binary files
+  - `false`
+
+* - `-continue`
+  - Continue from previous solution (input is step number of previous solution)
+  - `0`
+
+* - `-continue_filename`
+  - Path to solution binary file from which to continue from
+  - `[output_dir]/ns-solution.bin`
+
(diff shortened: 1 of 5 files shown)


examples/fluids/README.md
```diff
@@ -72,14 +72,26 @@ The following options are common among all problem types:
...
-  - Continue from previous solution
-  - `0`
-
 * - `-output_dir`
   - Output directory
   - `.`
 
+* - `-output_add_stepnum2bin`
+  - Whether to add step numbers to output binary files
+  - `false`
```
jedbrown, line 81: Thanks for making this PR. Could we simplify life by writing the step number into `ns-time.bin` so that it can just be read, instead of placing semantic meaning on the file name?

jrwrigh: That's definitely a more robust way of doing it. However, we'd still need to write out a solution time file for every restart file, so the name of the `ns-time.bin` file needs to be different to not overwrite the previous one.

Really, I should move over to using hdf5/cgns and store the timestep number and solution time in the restart file itself (unless there's an easy way of doing that with the binary files).

jedbrown: We'd need a bit of code in PETSc to do parallel load from CGNS (I could give you pointers). The native HDF5 format does parallel IO, but the viz workflow is clunky.

jedbrown (approved):
Okay, let's merge this before #1051 and we can revisit later for better usability.

PR #1089, opened 2022-11-11
Update error for multi-field diagonal assembly

zatkins-dev:
This PR corrects the format of the error message for multi-field diagonal assembly on `cuda-ref` and `hip-ref `to start with
> Backend does not implement…

### DIFF
--- a/backends/cuda-ref/ceed-cuda-ref-operator.c
+++ b/backends/cuda-ref/ceed-cuda-ref-operator.c
@@ -799,7 +799,7 @@ static inline int CeedOperatorAssembleDiagonalSetup_Cuda(CeedOperator op,
       if (rstrin && rstrin != rstr)
         // LCOV_EXCL_START
         return CeedError(ceed, CEED_ERROR_BACKEND,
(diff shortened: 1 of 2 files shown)


jeremylt:
This looks good.  Can you update the same message for operator full assembly is those files and double check the cpu ref/blocked/opt backends for the same?

zatkins-dev:
> This looks good. Can you update the same message for operator full assembly is those files and double check the cpu ref/blocked/opt backends for the same?

Should I also do this for the `CEED_EVAL_DIV`, `CEED_EVAL_CURL`, and `Assembling identity Qfunctions` errors across most of the backend files? Those are also backend errors in the style
> ... not supported

jeremylt:
We don't really need them right now.

jeremylt:
I think the CPU version is actually might be right or wrong without telling you which. I can explain tomorrow.

jeremylt (approved):
LGTM, thanks!

jedbrown (approved):

PR #1091, opened 2022-11-17
Update CEED_EVAL_DIV/INTERP

rezgarshakeri:
Updated CEED_EVAL_DIV/INTERP to work for H(div) problem.


jeremylt:
Do we have an operator test with an Hdiv basis using Div and Interp? It would be good to make sure these changes are part of the test suite.

rezgarshakeri:
I have tested before in my Hdiv branch. I just wanted to make my rebase and your review easier, so when we want to merge the Hdiv example there would be no changes in backends.

jeremylt (approved):
Looks good, knowing you manually tested

rezgarshakeri:
Yeah, I did.

PR #1093, opened 2022-11-17
examples/fluids: upgrade to HLLC

jedbrown:
This allows both the shock and thermal bubble to exit pretty cleanly.

[nwave-hllc.webm](https://user-images.githubusercontent.com/3303/202583981-1a99fd81-b987-4da5-9152-d289a12213b1.webm)

@jrwrigh I'm just using HLL for the Jacobian and I didn't refactor the HLL Jacobian yet (it can be simplified in terms of new helpers). Maybe you can look at the code and make a choice of what you think is maintainable and correct.

@AdelekeBankole is traveling right now so unless he comments, we should clean this up and merge it so @jrwrigh's coming tests use this instead of HLL.

For comparison, here is the same setup with HLL.

[nwave-hll.webm](https://user-images.githubusercontent.com/3303/202585139-a5d05e5a-5670-4abb-bdd7-e5f9d261796e.webm)

### DIFF
--- a/examples/fluids/qfunctions/freestream_bc.h
+++ b/examples/fluids/qfunctions/freestream_bc.h
@@ -138,8 +138,8 @@ CEED_QFUNCTION_HELPER StateConservative Harten_Lax_VanLeer_Contact_Flux(
 
   StateConservative flux_left  = FluxInviscidDotNormal(gas, left, normal);
   StateConservative flux_right = FluxInviscidDotNormal(gas, right, normal);
-  StateConservative flux_star_left[5] = {0.};
-  StateConservative flux_star_right[5] = {0.};
+  StateConservative flux_star_left[5];
+  StateConservative flux_star_right[5];
   //StateConservative flux_star_left;
   //StateConservative flux_star_right;
   StateConservative RiemannFlux_HLLC;
--- a/examples/fluids/qfunctions/freestream_bc.h
+++ b/examples/fluids/qfunctions/freestream_bc.h
@@ -12,6 +12,9 @@
 #include "newtonian_state.h"
 #include "newtonian_types.h"
 
+typedef StateConservative (*RiemannFluxFunction)(NewtonianIdealGasContext, State, State, const CeedScalar[3]);
+typedef StateConservative (*RiemannFluxFwdFunction)(NewtonianIdealGasContext, State, State, State, State, const CeedScalar[3]);
+
 typedef struct {
   CeedScalar left, right;
 } RoeWeights;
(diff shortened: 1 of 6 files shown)


examples/fluids/qfunctions/freestream_bc.h
```diff
@@ -81,6 +81,28 @@ CEED_QFUNCTION_HELPER StateConservative Flux_HLL_fwd(State left, State right, St
...
 
+CEED_QFUNCTION_HELPER void ComputeHLLSpeeds_Roe(NewtonianIdealGasContext gas, State left, CeedScalar u_left, State right, CeedScalar u_right,
+                                                CeedScalar *s_left, CeedScalar *s_right) {
+  const CeedScalar gamma = HeatCapacityRatio(gas);
+
+  RoeWeights r = RoeSetup(left.U.density, right.U.density);
+  // Speed estimate
+  // Roe average eigenvalues for left and right non-linear waves
+  // Stability requires that these speed estimates are *at least*
+  // as fast as the physical wave speeds.
```
jrwrigh, line 92: Are these comments relevant anymore? Particularly the stability comment since we're using Roe speeds regardless and give justification for that  on line 101.

examples/fluids/qfunctions/newtonian_state.h
```diff
@@ -132,6 +132,23 @@ CEED_QFUNCTION_HELPER StateConservative StateConservativeFromPrimitive_fwd(Newto
...
 
+CEED_QFUNCTION_HELPER StateConservative StateConservativeAXPBY(CeedScalar a, StateConservative X, CeedScalar b, StateConservative Y) {
+  StateConservative R;
+  R.density = a * X.density + b * Y.density;
+  for (int i = 0; i < 3; i++) R.momentum[i] = a * X.momentum[i] + b * Y.momentum[i];
+  R.E_total = a * X.E_total + b * Y.E_total;
+  return R;
+}
+
+CEED_QFUNCTION_HELPER StateConservative StateConservativeAXPBYPCZ(CeedScalar a, StateConservative X, CeedScalar b, StateConservative Y, CeedScalar c,
```
jrwrigh, line 143: Is there anyway to make the state structs indexable, so this kind of function wouldn't be `StateConservative` specific? So we'd be able to do something like
```c
for (int i=0; i<5; i++) {
  R[i] = a*X[i] + b*Y[i];
}
```
and thus have this be for any state struct? I've been looking into it and there are some weird hacks with `union` operators. Main issue we run into is padding for structs, but that can be overcome with attributes. But I'm not sure if we really want to tie ourselves to a key part of functioning code being formed by compiler-dependent features.

jrwrigh (approved):
Looks good to me.

examples/fluids/qfunctions/freestream_bc.h
jedbrown: Nah, they can be stripped or reorganized.

jrwrigh:
> I'm just using HLL for the Jacobian and I didn't refactor the HLL Jacobian yet (it can be simplified in terms of new helpers). Maybe you can look at the code and make a choice of what you think is maintainable and correct.

I definitely like the helper functions and think the code would be a bit easier to use with it. Looks like most of the work would go into writing up `StateConservativeAXPBY_fwd` and `StateConservativeAXPBYPCZ_fwd` functions, which should be straight forward. Not sure how soon I'd be able to get to it though.

examples/fluids/qfunctions/newtonian_state.h
jedbrown: Not really in C. You can get away with a union or cast on most compilers/opt levels, but it's undefined behavior and the attributes are not really portable. C++ and Rust can do it soundly.

AdelekeBankole:
@jedbrown, @jrwrigh , sorry pls for not being responsive in here. I have been pretty jet-lagged since yesterday I arrived. Will surely get better in the coming days. 

Thanks Jed, it looks good to me.

jedbrown:
Okay, I did the work here and will merge if Noether tests pass. Maybe students can run this during class today. (My morning is going to be complicated with both kids on different late starts.)

PR #1094, opened 2022-11-18
fluids: Fix freestream error typo

jrwrigh:


jedbrown:
@jrwrigh It looks like we haven't tested freestream BC with conservative variables (and it doesn't work currently) because https://github.com/ceed/libceed/commit/89e3cb53fc5f29c44319cd190169331660dcb2de#diff-a791ca288ecb6f3094c6cdcda721a2786d80bf17008a01cdf6f49117586db71bR24-R34 was missing a `break` statement after the conservative case.

jedbrown (approved):
This looks good to me now. Large time steps damp the acoustic waves (because we stop resolving acoustic CFL) so the HLL reflection is less severe. You can always drop the time step a bit to see it more clearly.

jrwrigh:
I'd want to add a quick change to the `FreestreamBCSetup` function and where it's called (move it to `newtonian.c`, similar to how the inflow and outflow BCs are setup automatically) before pushing this up.

jrwrigh:
Alright, good to go now assuming tests pass.

examples/fluids/problems/freestream_bc.c
```diff
@@ -11,20 +11,20 @@
 #include "../qfunctions/freestream_bc.h"
```
jrwrigh, line 12: I tried removing this random extra line, but clang-format seems to want it for some reason. Any ideas why?

jedbrown: I think it's because the `.h` filename matches this `.c` source file. If you include a header like that later, it can hide bugs (where the header doesn't include things it uses, and thus can break if it's included from a different source file). So it's a good rule.

examples/fluids/problems/newtonian.c
```diff
@@ -265,6 +265,8 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx) {
   newtonian_ig_ctx->state_var     = state_var;
   PetscCall(PetscArraycpy(newtonian_ig_ctx->g, g, 3));
 
+  PetscCall(FreestreamBCSetup(problem, dm, ctx, newtonian_ig_ctx));
```
jedbrown, line 268: Tests are failing because `-freestream_velocity` isn't given for the other examples. Should this only be called when there's at least one `-bc_freestream` (`bc->num_freestream > 0`)?

jrwrigh: Yeah, saw that. I was just going to add in the `set` boolean and check there. Running for `bc->num_freestream > 0` would be ideal, but that requires piping `bc` into every problem configuration function (otherwise it's not available).

jrwrigh: I'm open to whichever. You have an opinion either way?

jedbrown: Where can we ensure that if there exists a `-bc_freestream`, that these values have been set? (I don't really care about velocity because it can be zero, but pressure and temperature need to be set. I guess we could have STP defaults.

jrwrigh: I think `newtonian.c` makes the most sense for that location. It's called right after the command line arguments are processed, and the freestream BCs are only defined for Newtonian right now.

jedbrown: Sounds good.

PR #1096, opened 2022-11-18
Fluids-newtonian: Stabilization coefficients dependent on polynomial degree

LeilaGhaffari:
This is a WIP. Need to change `Ctau_*` as well!

CC: @amandashack @jnclement


examples/fluids/problems/newtonian.c
```diff
@@ -98,7 +99,7 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx, SimpleBC
   CeedScalar lambda = -2. / 3.;       // -
   CeedScalar mu     = 1.8e-5;         // Pa s, dynamic viscosity
   CeedScalar k      = 0.02638;        // W/(m K)
-  CeedScalar c_tau  = 0.5;            // -
+  CeedScalar c_tau  = 0.5 / degree;   // -
   CeedScalar Ctau_t = 1.0;            // -
   CeedScalar Ctau_v = 36.0;           // TODO make function of degree
```
jedbrown, line 104: This is what we call `Ctau_v`. I'd rather have a function that works for arbitrary order, or at least theoretical justification for these values, but just indexing this sequence `[36, 60, 128]` would be pragmatic and it's not clear we'll want to go higher than cubic anyway. https://github.com/PHASTA/phasta/blob/master/phSolver/compressible/e3tau.f#L146-L154

jedbrown:
Hi @LeilaGhaffari :wave: I think this doesn't need many changes, but it simplify workflow. Do you envision being able to finish it soon?

LeilaGhaffari:
Yes, I can finish it tomorrow morning.

jnclement:
@LeilaGhaffari Do you want any help? If not, @jedbrown, do you have an idea of where a good starting point for me would be? I won't have a lot of time the next two weeks due to travel and family obilgations, but I'd like to at least get started sometime soon.

jedbrown:
Hi @jnclement. I think what remains to be done is pretty simple. I linked parameters from PHASTA above and some of these are mentioned also in `blasius.yaml`. At first pass, I'd just like to see these set for linear, quadratic, and cubic. Rather than trying to create a function for arbitrary order, higher than cubic could just use the cubic value (and we can revisit them when we have reason to run those higher order cases, as we may have a more principled quantitative analysis framework by then). The goal is really just to have reasonable starting values so that inputs (like in the yaml files) don't have to be tuned separately for different basis order.

examples/fluids/problems/newtonian.c
LeilaGhaffari: Can we automate other `Ctau`s as well or should I drop the TODO comments?

LeilaGhaffari:
Thanks, @jnclement . This branch was simple. #1109 might need more modifications. Let's see what we can do over there.

examples/fluids/problems/newtonian.c
jedbrown: Please just set them all like is done in blasius.yaml and the PHASTA link above, for degree 1, 2, 3. Does that make sense?

examples/fluids/problems/newtonian.c
```diff
@@ -96,18 +96,19 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx, SimpleBC
...
-  CeedScalar Ctau_E = 1.0;            // TODO make function of degree
+  CeedScalar cv         = 717.;           // J/(kg K)
+  CeedScalar cp         = 1004.;          // J/(kg K)
+  CeedScalar g[3]       = {0, 0, -9.81};  // m/s^2
+  CeedScalar lambda     = -2. / 3.;       // -
+  CeedScalar mu         = 1.8e-5;         // Pa s, dynamic viscosity
+  CeedScalar k          = 0.02638;        // W/(m K)
+  CeedScalar c_tau      = 0.5 / degree;   // -
+  CeedScalar Ctau_t     = 1.0;            // -
+  CeedScalar Cv_func[3] = {36, 60, 128};
```
LeilaGhaffari, line 107: Does this look good or were you wanting a separate function?

examples/fluids/problems/newtonian.c
LeilaGhaffari: Hmm, looking at blasius.yaml, I would guess that `Ctau_C = Ctau_M = 0.25 / degree` and `Ctau_E = 0.125`, is that right?

examples/fluids/problems/newtonian.c
```diff
@@ -95,18 +96,19 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx, SimpleBC
...
+  CeedScalar Ctau_v     = Cv_func[degree - 1];
+  CeedScalar Ctau_C     = 0.25 / degree;
+  CeedScalar Ctau_M     = 0.25 / degree;
+  CeedScalar Ctau_E     = 0.125;
```
jedbrown, line 111: @KennethEJansen Is this consistent with what you consider good defaults in PHASTA? I know the 30, 60, 128 sequence is there. Is there a reference/link we could cite for these values?

@LeilaGhaffari The indexing could take `Min(3, degree) - 1` so it doesn't index beyond the array.

examples/fluids/tests-output/blasius_stgtest.yaml
```diff
@@ -13,11 +13,6 @@ platemesh:
   growth: 1.3
 
 stab: 'supg'
-Ctau_t: 1
-Ctau_v: 36
-Ctau_C: 0.25
-Ctau_M: 0.25
-Ctau_E: 0.125
 
 q_extra: 0
```
jedbrown, line 17: It can be a new PR, but shall we make `q_extra = 0` default? It was introduced long ago to handle aliasing as occurs in spectral element methods like Nek. My understanding is that people rarely use over-integration with SUPG because it sufficiently damps the high frequencies. We could preserve the existing unit test output by dropping `-q_extra 2` into their flags.

LeilaGhaffari: I can fix it in this PR really quick, if you don't mind.

LeilaGhaffari: Just changed it. I am not sure why not all tests needed the flag though (at least when testing on my local machine).

jedbrown: It probably doesn't change the answer enough to matter in most cases. Thanks!

examples/fluids/problems/newtonian.c
jedbrown: Resolving per Ken's thumbs up on Zulip.

jedbrown (approved):

jedbrown:
Thanks, team. @LeilaGhaffari @amandashack @jnclement

PR #1097, opened 2022-11-22
minor - missing $ in code samples

jeremylt:


jrwrigh (approved):
LGTM for consistency.

I personally always find the "$" in shell commands annoying when copy/pasting, but it does make it obvious where it's supposed to be run.

PR #1098, opened 2022-11-28
Add method to correct scaling for composite prolong/restrict

jeremylt:
This PR fixes multiplicity across suboperators for composite prolongation/restriction.

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -1962,6 +1962,126 @@ int CeedOperatorMultigridLevelCreateH1(CeedOperator op_fine, CeedVector p_mult_f
 }
 
 /**
+  @brief Scale composite prolongation and restriction operators for multiplicity between suboperators.
+           The prolongation and restriction suboperators must be created by `CeedOperatorMultigridLevelCreate*`.
+           Each suboperator should be a matching pair between prolongation and restriction.
+
+  Note: Calling this function asserts that setup is complete and sets the CeedOperators as immutable.
+
+  @param[in,out] op_prolong    Coarse to fine operator
+  @param[in,out] op_restrict   Fine to coarse operator
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedCompositeOperatorScaleMultigridLevel(CeedOperator op_prolong, CeedOperator op_restrict) {
+  CeedCall(CeedOperatorCheckReady(op_prolong));
+  CeedCall(CeedOperatorCheckReady(op_restrict));
+
+  Ceed                ceed;
+  CeedInt             num_sub_ops;
+  CeedSize            mult_l_vec_len;
+  CeedScalar         *mult;
+  CeedVector          mult_l_vec, ones_l_vec;
+  CeedElemRestriction elem_restr;
+  CeedOperator       *sub_ops;
+  CeedOperatorField  *input_fields;
+
+  CeedCall(CeedOperatorGetCeed(op_prolong, &ceed));
+
+  // Get suboperators
+  CeedCall(CeedOperatorGetNumSub(op_prolong, &num_sub_ops));
+  CeedCall(CeedOperatorGetSubList(op_prolong, &sub_ops));
+  if (num_sub_ops == 0) return CEED_ERROR_SUCCESS;
+
+  // Check for matching operators
+  {
+    char              *field_name;
+    CeedInt            num_sub_ops_restrict;
+    CeedVector         scale_prolong, scale_restrict;
+    CeedOperator      *sub_ops_restrict;
+    CeedOperatorField *input_fields_restrict;
+
+    // -- Check number suboperators
+    CeedCall(CeedOperatorGetNumSub(op_restrict, &num_sub_ops_restrict));
+    CeedCall(CeedOperatorGetSubList(op_restrict, &sub_ops_restrict));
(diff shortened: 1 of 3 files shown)


jeremylt:
If we like this fix, I'll add it to the other language interfaces

interface/ceed-preconditioning.c
```diff
@@ -1961,6 +1961,105 @@ int CeedOperatorMultigridLevelCreateH1(CeedOperator op_fine, CeedVector p_mult_f
...
+  // Compute multiplicity across suboperators
+  for (CeedInt i = 0; i < num_sub_ops; i++) {
+    const CeedScalar *sub_mult;
+    CeedVector        sub_mult_l_vec, ones_e_vec;
+
+    CeedCall(CeedOperatorGetFields(sub_ops[i], NULL, &input_fields, NULL, NULL));
+    CeedCall(CeedOperatorFieldGetElemRestriction(input_fields[1], &elem_restr));
+    CeedCall(CeedElemRestrictionCreateVector(elem_restr, &sub_mult_l_vec, &ones_e_vec));
+    CeedCall(CeedVectorSetValue(sub_mult_l_vec, 0.0));
+    CeedCall(CeedElemRestrictionApply(elem_restr, CEED_NOTRANSPOSE, ones_l_vec, ones_e_vec, CEED_REQUEST_IMMEDIATE));
```
jedbrown, line 2034: Why not `CeedVectorSetValue(ones_e_vec, 1.0)`?

jeremylt: HDiv restrictions prevent this shortcut, right?

jedbrown: Hmm, yeah. I remember that now.

interface/ceed-preconditioning.c
```diff
@@ -1961,6 +1961,105 @@ int CeedOperatorMultigridLevelCreateH1(CeedOperator op_fine, CeedVector p_mult_f
...
+  CeedCall(CeedVectorReciprocal(mult_l_vec));
+
+  // Scale by multiplicity
+  for (CeedInt i = 0; i < num_sub_ops; i++) {
+    CeedVector scale_l_vec;
+
+    CeedCall(CeedOperatorGetFields(sub_ops[i], NULL, &input_fields, NULL, NULL));
+    CeedCall(CeedOperatorFieldGetVector(input_fields[1], &scale_l_vec));
+    CeedCall(CeedVectorPointwiseMult(scale_l_vec, scale_l_vec, mult_l_vec));
+  }
```
jedbrown, line 2054: If I'm reading correctly, this changes the behavior of each suboperator, which could be problematic when working with subdomain problems (e.g., we solve a coupled problem and also subdomain problems). We could check the reference count to confirm that we have exclusive access before mutating (just error unsupported if not) and consider using a nodal scaling vector later.

interface/ceed-preconditioning.c
```diff
@@ -1961,6 +1961,105 @@ int CeedOperatorMultigridLevelCreateH1(CeedOperator op_fine, CeedVector p_mult_f
...
+  Note: Calling this function asserts that setup is complete and sets the CeedOperator as immutable.
+
+  @param[in,out] op_prolong    Coarse to fine operator
+  @param[in,out] op_restrict   Fine to coarse operator
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedCompositeOperatorScaleMultigridLevel(CeedOperator op_prolong, CeedOperator op_restrict) {
```
jedbrown, line 1976: I think the docs need to reference how these operators are created and their required relation.

jedbrown (approved):
Cool. I think we need to be a bit more defensive about misuse and surprise mutation, but it makes sense to me now.

interface/ceed-preconditioning.c
jeremylt: I think the extra information I added might clarify the use case here. This only works with suboperators created via `CeedOperatorMultigridLevelCreate*` functions. In that case, the scaling vector is never available to the user.

jedbrown: My concern is that `CeedOperatorMultigridLevelCreate*` says
>   Note: Calling this function asserts that setup is complete and sets the `CeedOperator` as immutable.

and yet if they put it into a composite operator and call this function, their sub-operator will be mutated.

What if we did multigrid level creation directly for composite operators? Then the proper scaling would be done the first time.

jeremylt: > What if we did multigrid level creation directly for composite operators? Then the proper scaling would be done the first time.

We could do that. We will need to decide what will happen with multiple active bases/restrictions and then pass an array of basis/restriction information. I thought this would be a quick fix, but I'll go tackle that piece of ugliness before trying to fix this bug again.

jedbrown: I don't have a problem with a quick fix if we need to map out the problem space better, but we need to do it in a way that avoids expensive bug hunts in the future because this looks easy to misuse and we should avoid having false statements in documentation.

jeremylt: The documentation is still true as written with the changes here - `CeedOperatorMultigridLevelCreate*` sets the input operator on the fine grid as immutable but does not mark the prolongation, restriction, or coarse grid operators as immutable. Perhaps it should though.

jedbrown: Oh, I missed that distinction. I don't like the late mutation, but I guess we can live with it if a safer interface is too complicated.

PR #1105, opened 2022-11-30
Add CeedCompositeOperatorGetMultiplicity

### LINKED ISSUE #1101: Composite Operator Multigrid
Currently, we have a bug in our composition of prolongation/restriction operators in Ratel - we don't account for multiplicity across suboperators.

1) Fix correctness for single active basis, no overlapping volumes

`CeedCompositeOperatorGetMultiplicity(op, mult)`

The output can be used to scale the `p_mult_fine` vector passed in when creating the restriction/prolongation suboperators.

Looking ahead, it might make sense to include an array of subdomain operators to skip, by index.

`CeedCompositeOperatorGetMultiplicity(op, skip_indices, mult)`

This function internals could be adapted to also handle multiple active bases/restrictions in the future, but restricting to only suboperators with one active basis/restriction would be sufficient for now

2) We could add a utility for building a new composite operator multigrid level

`CeedOperatorMultigridLevelCreateH1(op_fine, p_mult_fine, restrs_coarse, bases_coarse, interps_c_to_f, prolong_skip_indices, op_coarse, op_prolong, op_restrict)`

I'm not sure if I'd use such an interface in Ratel or not.

This also doesn't start to discuss how we'd handle multiple active bases/restrs for a single suboperator. We may need more information on how we want to handle multiple active bases/restrs before we add such an interface, or we may decide to skip such an interface as too messy?

jeremylt:
This tackles #1098 from the other end - get the multiplicity right in the first place

Fixes #1101

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -398,8 +398,8 @@ static inline int CeedCompositeOperatorLinearAssembleAddDiagonal(CeedOperator op
                                                                  CeedVector assembled) {
   CeedInt       num_sub;
   CeedOperator *suboperators;
-  CeedCall(CeedOperatorGetNumSub(op, &num_sub));
-  CeedCall(CeedOperatorGetSubList(op, &suboperators));
+  CeedCall(CeedCompositeOperatorGetNumSub(op, &num_sub));
+  CeedCall(CeedCompositeOperatorGetSubList(op, &suboperators));
   for (CeedInt i = 0; i < num_sub; i++) {
     if (is_pointblock) {
       CeedCall(CeedOperatorLinearAssembleAddPointBlockDiagonal(suboperators[i], assembled, request));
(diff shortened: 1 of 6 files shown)


interface/ceed-preconditioning.c
```diff
@@ -1753,6 +1753,77 @@ int CeedOperatorLinearAssemble(CeedOperator op, CeedVector values) {
...
 }
 
+/**
+  @brief Get the multiplicity of nodes across suboperators in a composite CeedOperator
+
+  Note: Calling this function asserts that setup is complete and sets the CeedOperator as immutable.
+
+  @param[in]  op               Composite CeedOperator
+  @param[in]  num_skip_indices Number of suboperators to skip
+  @param[in]  skip_indices     Array of indices of suboperators to skip
```
jedbrown, line 1763: I don't know if it's easy to say what to keep rather than what to skip, but we can change that later if it becomes an issue.

jedbrown (approved):

PR #1109, opened 2022-12-07
Fluids - Compute component-wise error for NS solver

LeilaGhaffari:
This is part of a course project in collaboration with @jnclement and @amandashack . It adds support for computing the error for each component and converts the errors to the other set of variables (primitive <-> conservative). It currently breaks for the other solvers though (advection and euler).


LeilaGhaffari:
I think this branch was working for our purposes in the course project but is broken for the Advection2D and Euler solvers. A quick fix would be to define a `convert_state` QFunction for each of those solvers. We could, instead, refactor the Euler problems to use `newtonian.c` (could we do the same thing with Advection?). However, the branch would diverge from its original purpose.

jedbrown:
Hmm, I think `State{,Primitive,Conservative}` need not be "Newtonian" things (they have nothing to do with viscosity). Their conversions assume ideal gasses. So we could promote that concept to something that Euler would also use. That's scope creep and I think it would also be fine to just disable your feature for Advection and Euler.

LeilaGhaffari:
Oh, true! I was talking about Newtonian since that is the only solver that uses `NewtonianIdealGasContext`, IIRC. Then, I would prefer disabling the feature for Advection and Euler for now as we have decided to refactor Euler (to use Newtonian) anyway.

jedbrown:
Sounds good to me. Newtonian is what matters for us (and at some point in the next few months, I think we'll add RANS).

examples/fluids/qfunctions/newtonian.h
```diff
@@ -93,6 +93,40 @@ CEED_QFUNCTION(ICsNewtonianIG_Prim)(void *ctx, CeedInt Q, const CeedScalar *cons
...
+    switch (context->state_var) {
+      case STATEVAR_CONSERVATIVE:
+        s = StateFromU(context, q_s, x);
+        StateToY(context, s, q_t);
+        break;
+      case STATEVAR_PRIMITIVE:
+        s = StateFromY(context, q_s, x);
+        StateToU(context, s, q_t);
+        break;
+    }
```
jedbrown, line 124: How are we going to maintain this once entropy variables (#1110) merge? Should it instead be converting from whatever to primitive?

examples/fluids/problems/eulervortex.c
```diff
@@ -49,7 +49,7 @@ PetscErrorCode NS_EULER_VORTEX(ProblemData *problem, DM dm, void *ctx, SimpleBC
   problem->apply_outflow.qfunction_loc       = Euler_Outflow_loc;
   problem->bc                                = Exact_Euler;
   problem->bc_ctx                            = euler_ctx;
-  problem->non_zero_time                     = PETSC_TRUE;
+  problem->non_zero_time                     = PETSC_FALSE;  // TODO: add support for converting state variables
```
jedbrown, line 52: Could this be split into two flags, say `non_zero_time_cons` and `non_zero_time_prim`, so we can still report errors in the state variables, skipping an attempt to convert?

jedbrown (approved):
Looks good. Perhaps a topic for another day, but I wonder if the velocity error norm should be defined using the pointwise relative error $\frac{\lVert \mathbf u_h - \mathbf u_* \rVert}{\lVert \mathbf u_* \rVert}$ (or absolute vector error) so we don't have anomalies for exactly or approximately axis-aligned flows. I think this (from Blasius) would show less than 1% error across all fields if measured in that way.
```
  Primitive variables:
        Component 0: 6.06303e-05 (24.4948, 404002.)
        Component 1: 0.00177814 (0.276762, 155.647)
        Component 2: 5.3239 (0.256163, 0.0481156)
        Component 3: -nan. (0., 0.)
        Component 4: 0.00184015 (2.93409, 1594.48)
```
We could do that in postprocessing by comparing absolute error as a vector to the true solution norm as a vector. (What's happening here is that velocity is very accurate in total, but not in exactly the same direction as the exact solution, thus we get a small y component error that is "big" compared to the exact y component that is almost perfectly aligned with the axis.)

examples/fluids/qfunctions/newtonian.h
LeilaGhaffari: Oh, this was my concern too and tried to create a new restriction to be able to output 10 values instead of 5. However, I didn't find it easy, so gave up due to the timing of the course project. I believe we would need to report the error in all state variables if we want to perform an analysis similar to what we did with the traveling vortex problem in summer. That said, I don't really know how to do it without making the algorithm too complicated.

jedbrown: What's wrong with having `ToPrimitive` and perhaps `ToConservative`, where either can be NULL. Entropy variables can implement both. In practice, I think >90% of comparisons will be comparing primitive variables because they are more intuitive and interpretable, so I think you could just have `ToPrimitive` and call it a day. If someone needs to compare conservative variables directly, they can add that alternative when the time comes.

examples/fluids/problems/eulervortex.c
LeilaGhaffari: Having a second flag makes sense but how about using `problem->convert_state.qfunction` for that purpose?

examples/fluids/qfunctions/newtonian.h
LeilaGhaffari: I guess I was worried about having too many QFunctions and if-statements. Alright, I'll keep `ToPrimitive` for now and output the errors for Euler and Advection2D by using a second flag.

LeilaGhaffari:
@jedbrown , does this look good to you now?

LeilaGhaffari:
@jedbrown , can we merge this branch if there are no other concerns?

examples/fluids/src/setuplibceed.c
```diff
@@ -260,6 +260,16 @@ PetscErrorCode SetupLibceed(Ceed ceed, CeedData ceed_data, DM dm, User user, App
   CeedQFunctionAddInput(ceed_data->qf_ics, "qdata", q_data_size_vol, CEED_EVAL_NONE);
   CeedQFunctionAddOutput(ceed_data->qf_ics, "q0", num_comp_q, CEED_EVAL_NONE);
 
+  // -- Create QFunction for converting the state variable set
+  if (problem->non_zero_time && !user->app_ctx->test_mode && problem->to_prim.qfunction && user->phys->state_var != STATEVAR_PRIMITIVE) {
+    CeedQFunctionCreateInterior(ceed, 1, problem->to_prim.qfunction, problem->to_prim.qfunction_loc, &ceed_data->qf_to_prim);
```
jedbrown, line 265: Does this ever get destroyed?

examples/fluids/src/setuplibceed.c
```diff
@@ -353,6 +363,14 @@ PetscErrorCode SetupLibceed(Ceed ceed, CeedData ceed_data, DM dm, User user, App
   CeedOperatorSetField(ceed_data->op_ics, "q0", ceed_data->elem_restr_q, CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
   CeedOperatorContextGetFieldLabel(ceed_data->op_ics, "evaluation time", &user->phys->ics_time_label);
 
+  // -- Create CEED operator for converting the state variable set
+  if (ceed_data->qf_to_prim) {
+    CeedOperatorCreate(ceed, ceed_data->qf_to_prim, NULL, NULL, &ceed_data->op_to_prim);
```
jedbrown, line 368: Does this get destroyed?

examples/fluids/src/misc.c
```diff
@@ -132,32 +132,126 @@ PetscErrorCode RegressionTests_NS(AppCtx app_ctx, Vec Q) {
...
+  // Get the norm of each component
+  PetscCall(VecStrideNormAll(Q_true_loc, norm_type, norm_true));
+  PetscCallMPI(MPI_Allreduce(MPI_IN_PLACE, norm_true, 5, MPIU_REAL, norm_reduce, PETSC_COMM_WORLD));
+  // Compute the error of each component
+  PetscCall(VecAXPY(Q_soln_loc, -1.0, Q_true_loc));
+  PetscCall(VecStrideNormAll(Q_soln_loc, norm_type, norm_error));
+  PetscCallMPI(MPI_Allreduce(MPI_IN_PLACE, norm_error, 5, MPIU_REAL, norm_reduce, PETSC_COMM_WORLD));
+  for (int i = 0; i < 5; i++) {
+    rel_error[i] = norm_error[i] / norm_true[i];
+  }
```
jedbrown, line 182: I think this can only be correct as written when both norms are `MAX`. So maybe don't bother passing them in. If we want $L^2$ norms, we'd have to do a little work.

jedbrown (commented):
Looks pretty close. Could you test or at least show sample output, perhaps using the Euler vortex?

LeilaGhaffari:
Here is the output with the traveling vortex problem:
```
$ build/fluids-navierstokes -problem euler_vortex -degree 2 -q_extra 2 -dm_plex_box_faces 5,5,1 -dm_plex_box_lower 0,0,0 -dm_plex_box_upper 125,125,250 -dm_plex_dim 3 -mean_velocity 1.4,-2.,0 -bc_inflow 4,6 -bc_outflow 3,5 -bc_slip_z 1,2 -ts_dt 1e-7 -ts_rk_type 5bs -ts_rtol 1e-10 -ts_atol 1e-10 -ts_max_time 1

  MPI:
    Host Name                          : leila-ThinkPad-P53s
    Total ranks                        : 1
  Problem:
    Problem Name                       : euler_vortex
    Test Case                          : isentropic_vortex
    Background Velocity                : 1.400000,-2.000000,0.000000
    Stabilization                      : none
  libCEED:
    libCEED Backend                    : /cpu/self/opt/blocked
  PETSc:
    Box Faces                          : 5,5,1
    Time Stepping Scheme               : explicit
  Mesh:
    Number of 1D Basis Nodes (P)       : 3
    Number of 1D Quadrature Points (Q) : 5
    Global DoFs                        : 1573
    Owned DoFs                         : 1573
    Global nodes (DoFs / 5)            : 314
    Local nodes                        : 363
Time taken for solution (sec): 1.41848

Relative Error (absolute error norm, true solution norm):
  Conservative variables:
        Component 0: 0.0471095 (0.047109, 0.99999)
        Component 1: 0.364438 (0.505128, 1.38605)
        Component 2: 0.133585 (0.299979, 2.24561)
        Component 3: inf. (5.52411e-16, 0.)
        Component 4: 0.0467154 (0.264258, 5.65677)

Time integrator took 245 time steps to reach final time 1.00194
```

examples/fluids/src/misc.c
LeilaGhaffari: Done!

examples/fluids/src/setuplibceed.c
LeilaGhaffari: Done!

jedbrown:
Why does it always say `Conservative variables` (regardless of `-state_var primitive`) and wouldn't we expect this to converge under refinement? Also, the domain shape (`dm_plex_box_upper 125,125,250`) relative to resolution `dm_plex_box_faces 5,5,1` looks weird.

LeilaGhaffari:
> Why does it always say `Conservative variables` (regardless of `-state_var primitive`) ?

It's because traveling vortex doesn't support primitive variables yet.

LeilaGhaffari:
>  wouldn't we expect this to converge under refinement? 

I remember we didn't see convergence with `-degree 2` in the summer. I only presented degree 1 in my talk.

LeilaGhaffari:
> Also, the domain shape (`dm_plex_box_upper 125,125,250`) relative to resolution `dm_plex_box_faces 5,5,1` looks weird.

I took the arguments from the regression tests in `navierstokes.c` (and refined it) to show you the output. Oops, it is indeed weird!

jedbrown:
I wonder what happened. I thought we had made a convergence test (long ago) for this problem at arbitrary order. Anyway, how does isentropic vortex test your `to_prim` conversion if it's using exclusively conservative variables? Are the printed norms before or after conversion?

And with `-options_left`, I see this (which is good).
```
Option left: name:-state_var value: primitive source: command line
```

LeilaGhaffari:
> how does isentropic vortex test your `to_prim` conversion if it's using exclusively conservative variables? Are the printed norms before or after conversion?

I think we decided to report the component-wise norms for isentropic vortex as well. This problem doesn't have `to_prim` conversion and it's printing the norms without any conversion.

jedbrown:
Sounds good. How do we test the new feature?

LeilaGhaffari:
Hmm, the conversion doesn't seem to be right! They did make sense to me when I was running experiments for the course project (although I wasn't running them with the same arguments)!
```
$ build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_stgtest.yaml -state_var primitive -ts_dt 2e-7 
...
Time taken for solution (sec): 0.312749

Relative Error (absolute error norm, true solution norm):
  Primitive variables:
        Component 0: 0.12287 (12409.8, 101000.)
        Component 1: 5.03298 (201.319, 40.)
        Component 2: 73.5649 (11.072, 0.150506)
        Component 3: -nan. (0., 0.)
        Component 4: 0.111034 (31.9779, 288.)

Time integrator took 5 time steps to reach final time 1e-06
```

```
leila@leila-ThinkPad-P53s ~/git/libCEED (leila/fluids-ns-verification)$ build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_stgtest.yaml 

...
Time taken for solution (sec): 0.233436

Relative Error (absolute error norm, true solution norm):
  Conservative variables:
        Component 0: 0.000671353 (0.000820348, 1.22193)
        Component 1: 0.0109931 (0.537314, 48.8773)
        Component 2: 2.20481 (0.405483, 0.183908)
        Component 3: -nan. (0., 0.)
        Component 4: 0.000997987 (252.792, 253302.)
  Primitive variables:
        Component 0: 1.00067 (404276., 404006.)
        Component 1: 93.064 (14890.2, 160.)
        Component 2: 6584.88 (3964.26, 0.602025)
        Component 3: -nan. (0., 0.)
        Component 4: 68.2968 (78679., 1152.02)

Time integrator took 5 time steps to reach final time 1e-05
```

jedbrown (changes requested):
I think this needs a rebase or merge from main (to resolve the conflicts) and a test of some sort. Is that right?

LeilaGhaffari:
This is rebased now but when I try to reproduce [these runs](https://github.com/CEED/libCEED/pull/1109#issuecomment-1373950726), I see a segmentation violation. Same issue exists on `main` as well. I'm investigating.
Also, what sort of test are you thinking of? Something that could reproduce the conversion?

jedbrown:
Probably the issue just fixed by https://github.com/CEED/libCEED/pull/1179.

LeilaGhaffari:
> Probably the issue just fixed by #1179.

Yep! That issue is fixed but all the blasius tests and "turb_spanstats" are failing now. I haven't been following the developments in the fluids mini-app recently. The tests are not broken (which make the debugging more difficult), they are just failing with some norm. Can you think of any recent changes related to the blasius problem that could cause the failure?

jedbrown:
Failing in this branch or on main? @jrwrigh has been dealing with a CI issue.

jrwrigh:
My issues with the CI are limited to the ROCM and float jobs, not to the fluids examples specifically.

As for the failures, the fluids problems aren't failing on `main`, but I also don't see a reason why any changes made in this PR should affect the testing; The binary files and `RegressionTests_NS` are untouched, and it doesn't look like the solution data is being messed with.

LeilaGhaffari:
The tests wouldn't fail until I rebased the branch onto main. It is so confusing 🤷🏼‍♀️ . I will try to look into it more closely today.

examples/fluids/problems/newtonian.c
```diff
@@ -309,8 +309,6 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx, SimpleBC
   CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_vol_ijacobian.qfunction_context);
   CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_inflow.qfunction_context);
   CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_inflow_jacobian.qfunction_context);
-  CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_outflow.qfunction_context);
-  CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_outflow_jacobian.qfunction_context);
```
LeilaGhaffari: I don't know why these weren't dropped while rebasing (They were removed in [this commit](https://github.com/CEED/libCEED/commit/8a94a473032dc6ed59a2cf0afe1d886fbdb591f4)) but the problem is fixed now. @jedbrown , is this good to merge or do we need a test? This post-processing functionality is not supposed to be working with regression tests. I don't know what kind of test would be useful here.

jedbrown: What can I run to see it converging? Can it display errors in primitive variables when the simulation is using conservative variables? (Or is that out of scope? I forget much of the discussion.)

LeilaGhaffari: You can run it with [these arguments](https://github.com/CEED/libCEED/pull/1109#issuecomment-1373950726). Yes, it only displays the errors in primitive variables when the simulation is using conservative variable but not the other way around. That said, the converted errors are too large. This is probably not the best way to compute the errors.

jedbrown: I don't understand what the errors mean for dynamic STG, but if I use `blasius_test.yaml`, it diverges for primitive and blows up for conservative. What I'm looking for is one problem where we can grid converge. I'm okay with that not being tested through CI, but we should have one problem where we can see grid convergence for the two formulations.

LeilaGhaffari: When you say "it blows up", do you mean the convergence test or a single run? Also, I am understanding that grid convergence is the convergence study with mesh refinement, is that right? I remember our convergence study for the final presentation of the class failed as well. I think you suggested that the BC was overshadowing the results and needed a special treatment. Here is the screenshot from our slides:
Screenshot from 2023-03-16 20-55-46

jedbrown:
```
build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_test.yaml -state_var conservative

  MPI:
    Host Name                          : kichatna
    Total ranks                        : 1
  Problem:
    Stabilization                      : none
  libCEED:
  PETSc:
    Box Faces                          : 3,20,1
  Mesh:
    Global DoFs                        : 656
    Owned DoFs                         : 656
    Global nodes (DoFs / 5)            : 131
    Local nodes                        : 168
Time taken for solution (sec): 0.0764067

Relative Error (absolute error norm, true solution norm):
  Conservative variables:
        Component 0: 0.00211302 (0.00258197, 1.22193)
        Component 1: 0.00394782 (0.192959, 48.8773)
        Component 2: 8.16421 (0.130091, 0.0159343)
        Component 3: -nan. (0., 0.)
        Component 4: 0.000187194 (47.4166, 253302.)
  Primitive variables:
        Component 0: 1.00011 (404047., 404002.)
        Component 1: 396.28 (61679.9, 155.647)
        Component 2: 172899. (8319.15, 0.0481156)
        Component 3: -nan. (0., 0.)
        Component 4: 1103.25 (1.75911e+06, 1594.48)
Time integrator CONVERGED_ITS on time step 5 with final time 1e-05
```
So density is nearly constant, but pressure is wrong by order 1 and the x and y velocity are off by orders of magnitude, along with temperature? I think there's still a bug.

jedbrown: I think it's a problem that both fields use `elem_restr_q`.
```c
    CeedOperatorSetField(ceed_data->op_to_prim, "q_cons", ceed_data->elem_restr_q, ceed_data->basis_q, CEED_VECTOR_ACTIVE);
    CeedOperatorSetField(ceed_data->op_to_prim, "q_prim", ceed_data->elem_restr_q, CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
```
The `q_cons` input field is evaluated to quadrature points, but then the `q_prim` output is interpreted as though it was at nodes (and also not corrected for multiplicity). Since you're working with local vectors, you could do a naive CPU traversal running the conversion all at nodes or you could apply a nodal map like James is doing for SGS or you could do a (lumped mass?) projection of the primitive variables into an FE space.

I see `ConvertStateVar_NS` doesn't account for multiplicity in the LocalToGlobal.

LeilaGhaffari: Sorry I've been slow on getting back to this. I think I have fixed multiplicity and dropped `CEED_BASIS_COLLOCATED` above. But the errors haven't changed much.
```
$ build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_test.yaml -state_var conservative

...

Relative Error (absolute error norm, true solution norm):
  Conservative variables:
        Component 0: 0.00211302 (0.00258197, 1.22193)
        Component 1: 0.00394782 (0.192959, 48.8773)
        Component 2: 8.16421 (0.130091, 0.0159343)
        Component 3: -nan. (0., 0.)
        Component 4: 0.000187194 (47.4166, 253302.)
  Primitive variables:
        Component 0: 1.00033 (101034., 101001.)
        Component 1: 485.939 (19295., 39.7065)
        Component 2: 143100. (1824.88, 0.0127525)
        Component 3: -nan. (0., 0.)
        Component 4: 1371.4 (547956., 399.558)
Time integrator CONVERGED_ITS on time step 5 with final time 1e-05
```
Either I am still missing something in `setuplibceed.c`, or this method is basically wrong (I am trying to convert variables at quadrature points and map them back to nodal space and compute the errors).

LeilaGhaffari: I don't see `ComputeL2Projection()` being used in anywhere in the code but can I use that function to try projection instead?

jedbrown: You recall the Blasius case. If you converted conservative to primitive at quadrature points, you would see the "scallops", which would give a bias when projected into a nodal space. However, I think something is wrong with the projection because that doesn't explain the many orders of magnitude difference reported here.

LeilaGhaffari: @jedbrown , I tried to explain the problem in this flowchart. Do you think the second proposal would work?
libCEED_1109 drawio(3)

jedbrown: The quadrature loop in the "current approach" needs a quadrature weight if it's going to be the projection that the math to the right indicates. Sounds like we're just measuring the residual without quadrature weighting (a bug).

The new proposal looks good, and saves the complexity/cost of a consistent projection. The scalloping bias for a nodally-exact field will be present in anything because it's real.

In your last expression, you have 
```
q_source_true = ics.qfunction(final_time)    # this is a `StatePrimitive` per the box above
dQ_source = q_source_soln - q_source_true # can't subtract primitive from conservative
dQ_target = StateFromU(dQ_source, ...)
error_target = w * dQ_target^2
```
However, `StateFromU` needs the full state, not a deviation/difference. I would do the comparison arithmetic in primitive variables, like this.
```
q_source_true_prim = ics.qfunction(final_time)
q_source_soln_prim = StateFromU(q_source_soln_cons, ...)
error_target = w * (q_source_soln_prim - q_source_true_prim)^2
```

LeilaGhaffari: I see that the naming is indeed confusing. In the last box, the simulation is in conservative variables and by "source" I mean "conservative" in that context and before converting the variables, and "target" would mean "primitive" (state variables after the conversion). I was using the same notation as in the existing implementation. Will change them.

Oops, of course conservative and primitive variables are not linearly related. Thanks!

LeilaGhaffari:
Alright, I think this approach is working now! I am trying to perform a convergence study but a quick test shows that at least the conversion is correct:
```
$ mpiexec -n 6 build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_test.yaml -ts_dt 1e-9 -ts_max_steps 20 -state_var conservative

...

L2 Error:
  Conservative variables-Component 0: 3.86413
  Conservative variables-Component 1: 2.57201
  Conservative variables-Component 2: 1.66507
  Conservative variables-Component 3: 2.55199
  Conservative variables-Component 4: 3.7305
  Primitive variables-Component 0: 0.98822
  Primitive variables-Component 1: 1.49868
  Primitive variables-Component 2: 1.54276
  Primitive variables-Component 3: 1.06039
  Primitive variables-Component 4: 0.66525
Time integrator CONVERGED_ITS on time step 20 with final time 2e-08
```
```
$ mpiexec -n 6 build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_test.yaml -ts_dt 1e-9 -ts_max_steps 20 -state_var primitive
...

L2 Error:
  Primitive variables-Component 0: 0.988239
  Primitive variables-Component 1: 1.49869
  Primitive variables-Component 2: 1.54278
  Primitive variables-Component 3: 1.06041
  Primitive variables-Component 4: 0.665282
  Conservative variables-Component 0: 3.86417
  Conservative variables-Component 1: 2.57205
  Conservative variables-Component 2: 1.66509
  Conservative variables-Component 3: 2.552
  Conservative variables-Component 4: 3.73052
Time integrator CONVERGED_ITS on time step 20 with final time 2e-08
```
(Just squashed the commits to make the review easier.)

jedbrown (approved):
Looks good. Just a couple minor suggestions.

LeilaGhaffari:
Thanks, I added comments for `q_true` and changed the names. I am happy with a squash-merge if this last bit looks good to you.

jedbrown (commented):
Can we have one test of this functionality?

LeilaGhaffari:
> Can we have one test of this functionality?

Do you mean a convergence plot?

examples/fluids/qfunctions/newtonian.h
```diff
@@ -63,6 +63,94 @@ CEED_QFUNCTION(ICsNewtonianIG_Conserv)(void *ctx, CeedInt Q, const CeedScalar *c
...
+    for (CeedInt j = 0; j < 5; j++) q_target_error[j][i] = q_data[0][i] * (q_p_t[j] - q_p_s[j]) * (q_p_t[j] - q_p_s[j]);
+  }  // End of Quadrature Point Loop
+  return 0;
+}
+
+// *****************************************************************************
+// This QFunction computes the L2 error in conservative variables if
+//   the simulation is run in primitive variables.
+// *****************************************************************************
+CEED_QFUNCTION(Newtonian_L2ErrorAsConservative)(void *ctx, CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out) {
```
jrwrigh, line 126: We should probably combine all `Newtonian_L2ErrorAs*` into a single helper function that takes `StateVariable` as an argument and then have individual `CEED_QFUNCTION`s for the conservative and primitive case.

I also feel like we could combine the normal `Newtonian_L2Error` function into that same system; looks like the only difference is the lack state conversions at the end. ie. If the desired error and solution states are the same, then nothing would be done.

examples/fluids/qfunctions/newtonian.h
```diff
@@ -63,6 +63,94 @@ CEED_QFUNCTION(ICsNewtonianIG_Conserv)(void *ctx, CeedInt Q, const CeedScalar *c
...
+  CeedScalar(*q_error)[CEED_Q_VLA] = (CeedScalar(*)[CEED_Q_VLA])out[0];
+
+  // Quadrature Point Loop
+  CeedPragmaSIMD for (CeedInt i = 0; i < Q; i++) {
+    CeedScalar q_t[5] = {0}, q_s[5] = {0};
+    for (CeedInt j = 0; j < 5; j++) {
+      q_s[j] = q_soln[j][i];
+      q_t[j] = q_true[j][i];  // Automatically provided in the same variables as the solution.
+    }
+    for (CeedInt j = 0; j < 5; j++) q_error[j][i] = q_data[0][i] * (q_t[j] - q_s[j]) * (q_t[j] - q_s[j]);
```
jrwrigh, line 85:
```suggestion
    for (CeedInt j = 0; j < 5; j++) q_error[j][i] = q_data[0][i] * Square(q_t[j] - q_s[j]);
```

Similar for the other functions as well (though if we combine them into a single helper function, then that becomes a mute point).

examples/fluids/src/misc.c
```diff
@@ -150,32 +136,66 @@ PetscErrorCode RegressionTests_NS(AppCtx app_ctx, Vec Q) {
...
+
+  PetscFunctionReturn(PETSC_SUCCESS);
+};
+
+// Get error for problems with true solutions
+PetscErrorCode GetError_NS(CeedData ceed_data, DM dm, User user, ProblemData *problem, Vec Q, PetscScalar final_time) {
+  Vec         Q_loc;
+  PetscReal   l2_error[5];
+  const char *state_var_source = "Conservative";
   PetscFunctionBegin;
```
jrwrigh, line 161:
```suggestion
  PetscFunctionBeginUser;
```

examples/fluids/src/misc.c
```diff
@@ -150,32 +136,66 @@ PetscErrorCode RegressionTests_NS(AppCtx app_ctx, Vec Q) {
...
 
-  // Get exact solution at final time
-  PetscCall(DMCreateGlobalVector(dm, &Q_exact));
-  PetscCall(DMGetLocalVector(dm, &Q_exact_loc));
-  PetscCall(VecGetSize(Q_exact_loc, &loc_nodes));
-  PetscCall(ICs_FixMultiplicity(dm, ceed_data, user, Q_exact_loc, Q_exact, final_time));
+  // Get the local values of the final solution
+  PetscCall(DMGetLocalVector(dm, &Q_loc));
+  PetscCall(DMGlobalToLocal(dm, Q, INSERT_VALUES, Q_loc));
+  PetscCall(DMPlexInsertBoundaryValues_NS(dm, PETSC_TRUE, Q_loc, final_time, NULL, NULL, NULL));
```
jrwrigh, line 166: Prefer using ` DMPlexInserBoundaryValues` (which `DMPlexInserBoundaryValues_NS` is probably set to execute through anyways) or, probably better, `UpdateBoundaryValues`.

examples/fluids/src/misc.c
```diff
@@ -150,32 +136,66 @@ PetscErrorCode RegressionTests_NS(AppCtx app_ctx, Vec Q) {
...
-  PetscInt  loc_nodes;
-  Vec       Q_exact, Q_exact_loc;
-  PetscReal rel_error, norm_error, norm_exact;
+PetscErrorCode ComputeL2Error(Vec Q_loc, PetscReal l2_error[5], OperatorApplyContext op_error_ctx, CeedContextFieldLabel time_label,
+                              CeedScalar time) {
+  Vec       E;
+  PetscReal l2_norm[5];
+  PetscFunctionBeginUser;
+
+  if (time_label) CeedOperatorSetContextDouble(op_error_ctx->op, time_label, &time);
```
jrwrigh, line 145: Prefer using `UpdateContextLabel`, but that's currently not in `navierstokes.h`; feel free to add the signature there are use it here if you want.

examples/fluids/src/misc.c
```diff
@@ -150,32 +136,66 @@ PetscErrorCode RegressionTests_NS(AppCtx app_ctx, Vec Q) {
...
-  PetscCall(DMGetLocalVector(dm, &Q_exact_loc));
-  PetscCall(VecGetSize(Q_exact_loc, &loc_nodes));
-  PetscCall(ICs_FixMultiplicity(dm, ceed_data, user, Q_exact_loc, Q_exact, final_time));
+  // Get the local values of the final solution
+  PetscCall(DMGetLocalVector(dm, &Q_loc));
+  PetscCall(DMGlobalToLocal(dm, Q, INSERT_VALUES, Q_loc));
+  PetscCall(DMPlexInsertBoundaryValues_NS(dm, PETSC_TRUE, Q_loc, final_time, NULL, NULL, NULL));
+
+  // Compute the L2 error in the source state variables
+  if (user->phys->ics_time_label) CeedOperatorSetContextDouble(ceed_data->op_ics_ctx->op, user->phys->ics_time_label, &final_time);
```
jrwrigh, line 169: Prefer using `UpdateContextLabel`, but that's currently not in `navierstokes.h`; feel free to add the signature there are use it here if you want.

examples/fluids/src/setuplibceed.c
```diff
@@ -264,6 +264,27 @@ PetscErrorCode SetupLibceed(Ceed ceed, CeedData ceed_data, DM dm, User user, App
   CeedQFunctionAddInput(ceed_data->qf_ics, "qdata", q_data_size_vol, CEED_EVAL_NONE);
   CeedQFunctionAddOutput(ceed_data->qf_ics, "q0", num_comp_q, CEED_EVAL_NONE);
 
+  // -- Create QFunction for computing the L2 error
+  if (problem->non_zero_time && user->app_ctx->test_type == TESTTYPE_NONE) {
```
jrwrigh, line 268: Side note; I'm not sure what `non_zero_time` means.

examples/fluids/qfunctions/newtonian.h
jrwrigh: For folding in `Newtonian_L2Error`, you could just have it's `CEED_QFUNCTION` use `context->state_var` for an if statement.

examples/fluids/src/misc.c
```diff
@@ -22,7 +22,7 @@ PetscErrorCode ICs_FixMultiplicity(DM dm, CeedData ceed_data, User user, Vec Q_l
   PetscFunctionBeginUser;
 
   // Update time for evaluation
-  if (user->phys->ics_time_label) CeedOperatorSetContextDouble(ceed_data->op_ics_ctx->op, user->phys->ics_time_label, &time);
+  if (user->phys->ics_time_label) PetscCall(UpdateContextLabel(user->comm, time, ceed_data->op_ics_ctx->op, user->phys->ics_time_label));
```
LeilaGhaffari, line 25: @jrwrigh , would we want this change here as well?

examples/fluids/src/setuplibceed.c
LeilaGhaffari: I don't think that this is a very descriptive name but it is a flag for problems with exact solutions. Maybe I can change it to `has_true_soln`?! I think the channel problem is missing it. Beside blasius and channel (NEWTONIAN), 2D advection and traveling vortex have exact solutions as well.

LeilaGhaffari: Well, I have broken the functionality we previously had for the euler and advection solvers!!!

jrwrigh: If the flag is for problems with exact solutions, I think `has_true_soln` works much better.

examples/fluids/src/misc.c
jrwrigh: Yep, good catch!

examples/fluids/qfunctions/newtonian.h
LeilaGhaffari: Thanks @jrwrigh , good call. I combined `Newtonian_L2ErrorAs*` in a single QFunction and I don't think we would even need a helper function. see https://github.com/CEED/libCEED/pull/1109/commits/81a3eaea3923a79a50cbe4228df7d469c6c0a522. 

Regarding folding everything in `Newtonian_L2Error`, I would need this qfunction for a separate ceed operator in a single simulation. i.e. one operator would compute the L2 errors in the source state variable and another one would take care of the conversion. Since we would need two separate ceed operators, I was thinking that having two distinct qfunctions would make the code cleaner rather than having one qfunction that carries a flag which will be passed to the qfunction depending on which ceed operator is being called. That said, is it even possible for two ceed operators to share a qfunction?
Another reason I am not inclined to combine `Newtonian_L2ErrorAs*` and `Newtonian_L2Error` is that we don't have the primitive variable support for traveling vortex and 2D advection. We need to compute the L2 error for these problems (which I need to fix) but we can't do the conversion. Even if my first argument wasn't valid, I think we would still need to refactor euler (and possibly advection) to use the newton solver.

jedbrown: It is possible for two operators to share a qfunction, but you could move the logic into one helper and have a near-empty qfunction call it with an extra parameter to select prim vs cons.

LeilaGhaffari: I think [this](https://github.com/CEED/libCEED/pull/1109/commits/71306b688a28ee6f36d6d0ce689623e0c4f9b8c0) is what you mean, right? I tried this first but then didn't see any point in keeping the helper function. Could you enlighten me?

examples/fluids/src/setuplibceed.c
LeilaGhaffari: Hmm, the channel problem doesn't work by simply turning on the flag. We could come back to this later!

examples/fluids/qfunctions/newtonian.h
LeilaGhaffari: Done! Thanks for the input.

jedbrown: Looks good to me now.

jrwrigh: Is `Newtonian_L2Error` not just the special case of `Newtonian_ConvertL2Error` where `target_state_var == context->state_var`? 

I guess the compiler wouldn't be able to optimize for the case of them being identical (ie. remove the `StateFromQ` and `StateToQ` statements), but this isn't a performance critical chunk of code (large scale simulations aren't going to have exact solutions anyways). I'd lean towards a more compact code in this case over the small performance uplift.

LeilaGhaffari: > Is `Newtonian_L2Error` not just the special case of `Newtonian_ConvertL2Error` where `target_state_var == context->state_var`?

That's true. After stepping away from my keyboard, I realized I misread your comment. For some reason I though you were talking about a single QFunction while you were talking about a single helper function for all cases. Do you like the last version better?

> I guess the compiler wouldn't be able to optimize for the case of them being identical (ie. remove the `StateFromQ` and `StateToQ` statements), but this isn't a performance critical chunk of code (large scale simulations aren't going to have exact solutions anyways). I'd lean towards a more compact code in this case over the small performance uplift.

True, I am not worried about the performance either (the logic seems a little weird though and I am not sure about its readability).

jrwrigh: Yep! That looks good to me!

jrwrigh (approved):
Thanks for getting this done!

jedbrown:
Cool. Squash or squash-merge?

LeilaGhaffari:
Thanks, @jrwrigh for bearing with me. A lot has changed since the last time I contributed  to this mini-appand.  There are still a couple of remaining tasks before merging:

- [ ] Adding a test that Jed mentioned [above](https://github.com/CEED/libCEED/pull/1109#issuecomment-1629228860)
- [ ] Fixing the broken code for traveling vortex and 2d advection
- [ ] Fixing a minor bug in reporting the error

LeilaGhaffari:
> Cool. Squash or squash-merge?

squash-merge is fine after I am done with the remaining tasks.

LeilaGhaffari:
I am seeing a weird issue. The errors are not correct with rank 1.
```
$ build/fluids-navierstokes -problem euler_vortex -degree 3 -dm_plex_box_faces 1,1,2 -dm_plex_box_lower 0,0,0 -dm_plex
_box_upper 125,125,250 -dm_plex_dim 3 -units_meter 1e-4 -units_second 1e-4 -mean_velocity 1.4,-2.,0 -bc_inflow 4,6 -bc_outflow 3,5 -bc_slip_z 1,2 -vortex_strength 2 -ksp_atol 1e-4 -ksp_rtol 
1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3 -implicit -dm_mat_preallocate_skip 0 -ts_type alpha


L2 Error:
  Conservative variables-Component 0: 0.000741784
  Conservative variables-Component 1: 0.000961908
  Conservative variables-Component 2: 0.00153698
  Conservative variables-Component 3: 0.
  Conservative variables-Component 4: 0.00399707
Time integrator CONVERGED_TIME on time step 50 with final time 0.05
```
```
$ mpiexec -n 4  build/fluids-navierstokes -problem euler_vortex -degree 3 -dm_plex_box_faces 1,1,2 -dm_plex_box_lo
wer 0,0,0 -dm_plex_box_upper 125,125,250 -dm_plex_dim 3 -units_meter 1e-4 -units_second 1e-4 -mean_velocity 1.4,-2.,0 -bc_inflow 4,6 -bc_outflow 3,5 -bc_slip_z 1,2 -vortex_strength 2 -ksp_at
ol 1e-4 -ksp_rtol 1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3 -implicit -dm_mat_preallocate_skip 0 -ts_type alpha

L2 Error:
  Conservative variables-Component 0: 2.66349e-06
  Conservative variables-Component 1: 8.8622e-05
  Conservative variables-Component 2: 6.99578e-05
  Conservative variables-Component 3: 3.72699e-155
  Conservative variables-Component 4: 2.05213e-05
Time integrator CONVERGED_TIME on time step 50 with final time 0.05
```

```
$ mpiexec -n 8  build/fluids-navierstokes -problem euler_vortex -degree 3 -dm_plex_box_faces 1,1,2 -dm_plex_box_lower 0,0,0 -dm_plex_box_upper 125,125,250 -dm_plex_dim 3 -units_meter 1e-4 -units_second 1e-4 -mean_velocity 1.4,-2.,0 -bc_inflow 4,6 -bc_outflow 3,5 -bc_slip_z 1,2 -vortex_strength 2 -ksp_atol 1e-4 -ksp_rtol 1e-3 -ksp_type bcgs -snes_atol 1e-3 -snes_lag_jacobian 100 -snes_lag_jacobian_persists -snes_mf_operator -ts_dt 1e-3 -implicit -dm_mat_preallocate_skip 0 -ts_type alpha

L2 Error:
  Conservative variables-Component 0: 2.66349e-06
  Conservative variables-Component 1: 8.8622e-05
  Conservative variables-Component 2: 6.99578e-05
  Conservative variables-Component 3: 4.56152e-155
  Conservative variables-Component 4: 2.05213e-05
Time integrator CONVERGED_TIME on time step 50 with final time 0.05
```
(same issue with advection2d and blasius)
It could be the logic in [FixMultiplicity_NS()](https://github.com/CEED/libCEED/blob/ec5aba6b589d96ca613bd8a612452e4ebbbf883d/examples/fluids/src/misc.c#L18) but I can't point at anything specific.

LeilaGhaffari:
Testing with different arguments/problems, I see that the results could be different for n>1 as well.

jedbrown (approved):
Cool. Can we test isentropic vortex now or does it need some conversion? Do we have any problems in which we can show convergence under refinement?

LeilaGhaffari:
> Cool. Can we test isentropic vortex now or does it need some conversion? Do we have any problems in which we can show convergence under refinement?

Unfortunately, isentropic vortex doesn't have support for primitive variables. If we want to test the full functionality, we should either make the channel problem compute the errors (I tried but it was more complicated than switching on a flag) or, refactor the euler solver to use newtonian (I think this is the simplest way to make traveling vortex functional for our desired convergence study). 

Or we can just show the convergence with isentropic vortex using conservative variables (without the conversion). This is ready!

However, I am still concerned about [this issue](https://github.com/CEED/libCEED/pull/1109#issuecomment-1634259867). Could you take a look at the functions I defined/modified in `fluids/misc.c` (mainly `FixMultiplicity_NS()` and `ComputeL2Error()`)?

LeilaGhaffari:
@jedbrown , I think I fixed the issues you raised the other day but I think the problem I am seeing with the parallel run could be irrelevant to this PR.
When testing the blasius problem on `main` I get:
### Serial
```
libCEED (main)$ build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_test.yaml -ts_max_steps 10

  MPI:
    Host Name                          : leila-ThinkPad-P53s
    Total ranks                        : 1
  Problem:
    Stabilization                      : none
  libCEED:
    libCEED Backend                    : /cpu/self/opt/blocked
  PETSc:
    Box Faces                          : 3,20,1
  Mesh:
    Global DoFs                        : 656
    Global 5-DoF nodes                 : 131
  Partition:                             (min,max,median,max/median)
    Global Vector 5-DoF nodes          : 131, 131, 131, 1.000000
    Local Vector 5-DoF nodes           : 168, 168, 168, 1.000000
    Ghost Interface 5-DoF nodes        : 0, 0, 0, -nan
    Ghost Interface Ranks              : 0, 0, 0, -nan
    Owned Interface 5-DoF nodes        : 0, 0, 0, -nan
    Owned Interface Ranks              : 0, 0, 0, -nan
Time taken for solution (sec): 0.224105
Time integrator CONVERGED_ITS on time step 10 with final time 2e-05``
```

### Parallel:
```
libCEED (main)$ mpiexec.hydra -n 6 build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_test.yaml -ts_max_steps 10

  MPI:
    Host Name                          : leila-ThinkPad-P53s
    Total ranks                        : 6
  Problem:
    Stabilization                      : none
  libCEED:
    libCEED Backend                    : /cpu/self/opt/blocked
  PETSc:
    Box Faces                          : 3,20,1
  Mesh:
    Global DoFs                        : 656
    Global 5-DoF nodes                 : 131
  Partition:                             (min,max,median,max/median)
    Global Vector 5-DoF nodes          : 17, 28, 20, 1.384615
    Local Vector 5-DoF nodes           : 36, 36, 36, 1.000000
    Ghost Interface 5-DoF nodes        : 0, 8, 8, 1.000000
    Ghost Interface Ranks              : 0, 1, 1, 1.000000
    Owned Interface 5-DoF nodes        : 0, 8, 8, 1.000000
    Owned Interface Ranks              : 0, 1, 1, 1.000000
Time taken for solution (sec): 0.353972
Time integrator DIVERGED_NONLINEAR_SOLVE on time step 0 with final time 0.
```
Basically, the solver is diverging with parallel runs while there is no problem with the serial run. Is this normal? 
(There could still be some issues with my implementation though. Could you give another look at your convenience?)

jedbrown:
Why is the nonlinear solve diverging in parallel? Is the linear solver failing? Is this solely a difference of the preconditioner getting weaker in parallel for the steady problem?

LeilaGhaffari:
The failure is very random and it happens in the linear solver. I ran some tests on Noether with `main`. For example, it failed with 6/8/20 processors but not with 1-5/7/10! 
Here is an example of the failed ones:
```
$ mpiexec -n 8 build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_test.yaml -ts_max_steps 10 -ksp_converged_reason -snes_converged_reason

  MPI:
    Host Name                          : noether
    Total ranks                        : 8
  Problem:
    Stabilization                      : none
  libCEED:
    libCEED Backend                    : /cpu/self/opt/blocked
  PETSc:
    Box Faces                          : 3,20,1
  Mesh:
    Global DoFs                        : 656
    Global 5-DoF nodes                 : 131
  Partition:                             (min,max,median,max/median)
    Global Vector 5-DoF nodes          : 12, 22, 16, 1.400000
    Local Vector 5-DoF nodes           : 28, 32, 28, 1.142857
    Ghost Interface 5-DoF nodes        : 0, 8, 8, 1.000000
    Ghost Interface Ranks              : 0, 1, 1, 1.000000
    Owned Interface 5-DoF nodes        : 0, 8, 8, 1.000000
    Owned Interface Ranks              : 0, 1, 1, 1.000000
    Linear solve did not converge due to DIVERGED_ITS iterations 10000
  Nonlinear solve did not converge due to DIVERGED_LINEAR_SOLVE iterations 0
Time taken for solution (sec): 0.257922
Time integrator DIVERGED_NONLINEAR_SOLVE on time step 0 with final time 0.
```
(I don't see any preconditioning being set in the yaml file.)

jedbrown:
You can use `-ts_view -ksp_converged_reason -ksp_view_singularvalues` to better understand the failure. Does the same failure occur on `main`?

LeilaGhaffari:
Oh, yes. Those runs and the following are all on `main`.

```
libCEED (main)$ mpiexec.hydra -n 6 build/fluids-navierstokes -options_file examples/fluids/tests-output/blasius_test.y
aml -ts_max_steps 10 -ts_view -ksp_converged_reason -ksp_view_singularvalues

  MPI:
    Host Name                          : leila-ThinkPad-P53s
    Total ranks                        : 6
  Problem:
    Stabilization                      : none
  libCEED:
    libCEED Backend                    : /cpu/self/opt/blocked
  PETSc:
    Box Faces                          : 3,20,1
  Mesh:
    Global DoFs                        : 656
    Global 5-DoF nodes                 : 131
  Partition:                             (min,max,median,max/median)
    Global Vector 5-DoF nodes          : 17, 28, 20, 1.384615
    Local Vector 5-DoF nodes           : 36, 36, 36, 1.000000
    Ghost Interface 5-DoF nodes        : 0, 8, 8, 1.000000
    Ghost Interface Ranks              : 0, 1, 1, 1.000000
    Owned Interface 5-DoF nodes        : 0, 8, 8, 1.000000
    Owned Interface Ranks              : 0, 1, 1, 1.000000
    Linear solve did not converge due to DIVERGED_ITS iterations 10000
Iteratively computed extreme singular values: max 482.958 min 1.45981 max/min 330.837
TS Object: 6 MPI processes
  type: beuler
  maximum steps=10
  maximum time=500.
  total number of I function evaluations=1
  total number of I Jacobian evaluations=1
  total number of nonlinear solver iterations=0
  total number of linear solver iterations=10000
  total number of nonlinear solve failures=1
  total number of rejected steps=1
  using relative error tolerance of 0.0001,   using absolute error tolerance of 0.0001
  TSAdapt Object: 6 MPI processes
    type: none
  SNES Object: 6 MPI processes
    type: newtonls
    maximum iterations=50, maximum function evaluations=10000
    tolerances: relative=1e-08, absolute=1e-50, solution=1e-08
    total number of linear solver iterations=10000
    total number of function evaluations=1
    norm schedule ALWAYS
    SNESLineSearch Object: 6 MPI processes
      type: bt
        interpolation: cubic
        alpha=1.000000e-04
      maxstep=1.000000e+08, minlambda=1.000000e-12
      tolerances: relative=1.000000e-08, absolute=1.000000e-15, lambda=1.000000e-08
      maximum iterations=40
    KSP Object: 6 MPI processes
      type: gmres
        restart=30, using Classical (unmodified) Gram-Schmidt Orthogonalization with no iterative refinement
        happy breakdown tolerance 1e-30
      maximum iterations=10000, initial guess is zero
      tolerances:  relative=1e-05, absolute=1e-50, divergence=10000.
      left preconditioning
      using PRECONDITIONED norm type for convergence test
    PC Object: 6 MPI processes
      type: bjacobi
        number of blocks = 6
        Local solver information for first block is in the following KSP and PC objects on rank 0:
        Use -ksp_view ::ascii_info_detail to display information for all blocks
      KSP Object: (sub_) 1 MPI process
        type: preonly
        maximum iterations=10000, initial guess is zero
        tolerances:  relative=1e-05, absolute=1e-50, divergence=10000.
        left preconditioning
        using NONE norm type for convergence test
      PC Object: (sub_) 1 MPI process
        type: ilu
          out-of-place factorization
          0 levels of fill
          tolerance for zero pivot 2.22045e-14
          matrix ordering: natural
          factor fill ratio given 1., needed 1.
            Factored matrix follows:
              Mat Object: (sub_) 1 MPI process
                type: seqaij
                rows=88, cols=88
                package used to perform factorization: petsc
                total: nonzeros=3680, allocated nonzeros=3680
                  using I-node routines: found 26 nodes, limit used is 5
        linear system matrix = precond matrix:
        Mat Object: (sub_) 1 MPI process
          type: seqaij
          rows=88, cols=88
          total: nonzeros=3680, allocated nonzeros=0
          total number of mallocs used during MatSetValues calls=0
            using I-node routines: found 26 nodes, limit used is 5
      linear system matrix = precond matrix:
      Mat Object: Mat_0x84000004_0 6 MPI processes
        type: mpiaij
        rows=656, cols=656
        total: nonzeros=37920, allocated nonzeros=0
        total number of mallocs used during MatSetValues calls=0
          using I-node (on process 0) routines: found 26 nodes, limit used is 5
Time taken for solution (sec): 0.333688
Time integrator DIVERGED_NONLINEAR_SOLVE on time step 0 with final time 0.
```

jeremylt:
Closing in anticipation of moving Honee to a new repo

this seems like a good branch to recreate on the other side of the move! We should make an issue when we stand up the new repo @jrwrigh

PR #1110, opened 2022-12-07
Add Entropy Variables to libCEED fluids NS mini-app

zatkins-dev:
This PR introduces physical entropy variables to libCEED fluids (see [Hughes et al](https://www.sciencedirect.com/science/article/pii/0045782586901271) for full formulation). Support for entropy variables mirrors that of primitive and conservative variables, with conversion and forward physics functions.

Also adds additional unit tests for variable conversion.


zatkins-dev:
Huh, turns out that waiting 7 months and rebasing can sometimes fix issues... I'm no longer seeing strange artifacts with `gaussianwave.yml` or with the density current problem. 

I will add some tests, but I think this MR actually may be pretty much ready.

examples/fluids/README.md
```diff
@@ -644,7 +644,7 @@ For the Density Current, Channel, and Blasius problems, the following common com
   - boolean
 
 * - `-state_var`
-  - State variables to solve solution with. `conservative` ($\rho, \rho \bm{u}, \rho e$) or `primitive` ($P, \bm{u}, T$)
+  - State variables to solve solution with. `conservative` ($\rho, \rho \bm{u}, \rho e$), `primitive` ($P, \bm{u}, T$), or `entropy` ($\frac{\gamma - s}{gamma - 1} - \frac{\rho}{P} (e - \iota), \frac{\rho}{P} \bm{u}, -\frac{\rho}{P}$) where $\iota = c_v T$ and $s = \ln(P\rho^{-\gamma})$
```
jrwrigh, line 647:
```suggestion
  - State variables to solve solution with. `conservative` ($\rho, \rho \bm{u}, \rho e$), `primitive` ($P, \bm{u}, T$), or `entropy` ($\frac{\gamma - s}{\gamma - 1} - \frac{\rho}{P} (e - c_v T),\ \frac{\rho}{P} \bm{u},\ -\frac{\rho}{P}$) where  $s = \ln(P\rho^{-\gamma})$
```

In summary, added missing `\` to gamma, moved the definition of $\iota$ into the expression (cleaner), and added spaces (`\ `) after the commas to (hopefully) make it more legible.

examples/fluids/navierstokes.c
```diff
@@ -22,13 +22,15 @@
...
 //TESTARGS(name="blasius_STG_strongBC") -ceed {ceed_resource} -test_type solver -options_file examples/fluids/tests-output/blasius_stgtest.yaml -compare_final_state_atol 1E-10 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius_STG_strongBC.bin -stg_strong true
 //TESTARGS(name="channel") -ceed {ceed_resource} -test_type solver -options_file examples/fluids/channel.yaml -compare_final_state_atol 2e-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-channel.bin -dm_plex_box_faces 5,5,1 -ts_max_steps 5
 //TESTARGS(name="channel-primitive") -ceed {ceed_resource} -test_type solver -options_file examples/fluids/channel.yaml -compare_final_state_atol 2e-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-channel-prim.bin -dm_plex_box_faces 5,5,1 -ts_max_steps 5 -state_var primitive
+//TESTARGS(name="channel-entropy") -ceed {ceed_resource} -test_type solver -options_file examples/fluids/channel.yaml -compare_final_state_atol 2e-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-channel-entropy.bin -dm_plex_box_faces 5,5,1 -ts_max_steps 5 -state_var entropy
```
jrwrigh, line 33: Any particular reason to have two tests? We already have a lot of tests right now, so I'd like to have changes be minimal. We can keep both if there's reason to (stressing different functions of the new variables).

examples/fluids/problems/freestream_bc.c
```diff
@@ -79,6 +79,23 @@ PetscErrorCode FreestreamBCSetup(ProblemData *problem, DM dm, void *ctx, Newtoni
           break;
       }
       break;
+
+    case STATEVAR_ENTROPY:
+      switch (riemann) {
```
jrwrigh, line 84:
```suggestion
    case STATEVAR_ENTROPY:
      switch (riemann) {
```

examples/fluids/problems/newtonian.c
```diff
@@ -28,38 +46,166 @@ static PetscErrorCode CheckPrimitiveWithTolerance(StatePrimitive sY, StatePrimit
...
-    State          t = StateFromU(gas, dU, dx);
-    StatePrimitive dY;
-    dY.pressure = (t.Y.pressure - s.Y.pressure) / eps;
-    for (int j = 0; j < 3; j++) dY.velocity[j] = (t.Y.velocity[j] - s.Y.velocity[j]) / eps;
-    dY.temperature = (t.Y.temperature - s.Y.temperature) / eps;
-    char buf[128];
-    snprintf(buf, sizeof buf, "StateFromU_fwd i=%d", i);
-    PetscCall(CheckPrimitiveWithTolerance(dY, ds.Y, dY, buf, 5e-6, 1e-6, 1e-6));
+  const CeedScalar T   = 200 * K;
+  const CeedScalar rho = 1.2 * kg / (m * m * m), u_base = 40 * m / sec;
```
jrwrigh, line 89:
```suggestion
  const CeedScalar rho = 1.2 * kg / Cube(m), u_base = 40 * m / sec;
```

examples/fluids/qfunctions/newtonian_types.h
```diff
@@ -50,6 +51,12 @@ typedef struct {
   CeedScalar temperature;
 } StatePrimitive;
 
+typedef struct {
+  CeedScalar chemical_potential;
+  CeedScalar velocity[3];
+  CeedScalar temperature;
```
jrwrigh, line 57: Do we not have more specific names for these?

examples/fluids/src/setupdm.c
```diff
@@ -100,6 +100,14 @@ PetscErrorCode SetUpDM(DM dm, ProblemData *problem, PetscInt degree, SimpleBC bc
       PetscCall(PetscSectionSetComponentName(section, 0, 3, "Velocity Z"));
       PetscCall(PetscSectionSetComponentName(section, 0, 4, "Temperature"));
       break;
+
+    case STATEVAR_ENTROPY:
+      PetscCall(PetscSectionSetComponentName(section, 0, 0, "Chemical Potential"));
+      PetscCall(PetscSectionSetComponentName(section, 0, 1, "Velocity X"));
+      PetscCall(PetscSectionSetComponentName(section, 0, 2, "Velocity Y"));
+      PetscCall(PetscSectionSetComponentName(section, 0, 3, "Velocity Z"));
```
jrwrigh, line 108: I definitely would not call these `Velocity` outright, as they're not just velocity. This breaks CGNS naming convention (see table at bottom of section A.2 in [CGNS standard](https://cgns.github.io/CGNS_docs_current/sids/dataname.html#dataname_flow)).

The spaces also technically break the standard, but you're just following what was done above which is fine. Something to change at somepoint later.

jrwrigh (commented):
Few comments, particularly regarding the naming of the components. Otherwise looks good to me.

examples/fluids/problems/stg_shur14.c
```diff
@@ -371,6 +372,10 @@ PetscErrorCode SetupStrongSTG(DM dm, SimpleBC bc, ProblemData *problem, Physics
       // {1,2,3,4} for u, v, w, T
       for (int i = 0; i < 4; i++) comps[i] = i + 1;
       break;
+
+    case STATEVAR_ENTROPY:
+      for (int i = 0; i < 5; i++) comps[i] = i;
+      break;
```
jrwrigh, line 378: I'm also not sure about this. Setting `comps` to set all values at the boundary would over-constrain the problem setup (this is used to set the components strong boundary conditions). I'm currently refreshing my memory of the strong BC setup right now to see what this should be realistically.

examples/fluids/README.md
zatkins-dev: sounds good to me!

examples/fluids/src/setupdm.c
zatkins-dev: Oh wow, I've never seen that before. I was/am struggling to figure out meaningful names for these quantities.

jrwrigh: Yeah. I think being blunt with it isn't a bad option:
```
EntropyEnergy
EntropyVelocityX
EntropyVelocityY
EntropyVelocityZ
EntropyTemperature
```

examples/fluids/qfunctions/newtonian_types.h
zatkins-dev: I'm not sure. Jed's suggestion in the past was to go the opposite route, and just name them `V1`, `V2`, `V3` or something like  that. I'm starting to think that might be best, since I don't know how to name them otherwise.

e.g. 
- The quantity I'm currently calling `temperature` is $-\frac{\rho}{P}$, which has units $\frac{kg}{J}$ or $\frac{s^2}{m^2}$. 
- The quantity I'm calling `velocity` is `\bm{U}\cdot\frac{\rho}{P}`, which has units $\bm{s}{m}$.

examples/fluids/src/setupdm.c
jedbrown: So these are gradients of entropy with respect to conservative variables and thus `EntropyEnergy` and `EntropyTemperature` feel confusing even though the terms left over with an ideal gas have that structure. I think the velocity-like terms should still be organized as vectors (with `X`, `Y`, `Z` suffixes), but names could even be
```
EntropyDensity
EntropyMomentumX
EntropyMomentumY
EntropyMomentumZ
EntropyEnergy
```
where the qualifier is the conservative variable that entropy is differentiated with respect to. In practice, we'll either need to use a calculator or convert to understandable variables to visualize, but that can be done in a new MR when it's more needed.

zatkins-dev: Thanks Jed -- That makes more sense I think.

examples/fluids/qfunctions/newtonian_types.h
zatkins-dev: I think the units for `V1` are $kg$ (since it has $\frac{\rho}{P}(e_{\text{kinetic}} + e_{\text{potential}})$)

jedbrown: :shrug: Consistency is good so if we're going to name them, may as well make them match. I'm not sure it's easier in the code.

BTW, did you do anything here that may impact performance of the primitive or conservative variable code?

zatkins-dev: Minorly, yes. The `StateFrom*` functions now also compute the entropy variables.

jedbrown: Is that necessary? We added both conservative and primitive to avoid recomputing the equation of state each time that quantity is needed in fluxes, but I don't think entropy variables are used that way so we could probably remove the `V` member from `State`.

examples/fluids/problems/stg_shur14.c
zatkins-dev: I think it should be the last 4 components of the entropy variables, since those involve $T$ and $\mathbf{u}$.

examples/fluids/qfunctions/newtonian_types.h
zatkins-dev: How would we use entropy variables then? We use `StateFromQ` and `StateFromQ_fwd` all over the place, and that relies on the given state variables to be in `State`

zatkins-dev: I don't think that I understand what you mean by "import entropy variables to that representation without making it part of `State`".

Tangentially, one of the major uses for entropy variables is in making the fluxes symmetric, which while not implemented here, would justify storing `V` in `State`.

jrwrigh:
@zatkins-dev Do you mind if I "take over" the PR so we can get it in a mergeable state for the HONEE forking?

zatkins-dev:
@jrwrigh Please do, thanks!

jrwrigh:
I've taken the entropy variables out of the `State` struct. It required two notes:
- I've used primitive variables from the `State` struct to create a `StateEntropy` where necessary. There's an arbitrary choice between whether to use primitive and conservative variables, but I figured primitive would be the most stable due to lack of coupling between different variables. 
  - This is slightly inconvenient when actually using entropy variables, as that means that `StateFromV_fwd` won't actually be using the entropy variable values directly, but have to go through $\mathbf V\rightarrow \mathbf Y\rightarrow \mathbf V$. Not ideal, but the error is minimal to do that jumping. Not sure if it's worth needing to add an extra argument to `StateFromQ_fwd` in order to accommodate this, but I'm leaning towards not right now.
- I added of a `StateToQ_fwd` to handle the fact that `dV` isn't calculated by default in `StateFromQ_fwd`. I didn't both with accompaning `StateToU_fwd`, etc., as they're mostly repeats of the `StateToQ`.


Otherwise, this is good to review now!

jeremylt:
Note - I'd use [co-authored-by](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) so we show the mixed authorship of the commits

examples/fluids/navierstokes.c
```diff
@@ -18,6 +18,7 @@
 //     ./navierstokes -ceed /cpu/self -options_file gaussianwave.yml
 //     ./navierstokes -ceed /gpu/cuda -problem advection -degree 1
 //
+//TESTARGS(name="gaussianwave_idl_entropy") -ceed {ceed_resource} -test_type solver -options_file examples/fluids/gaussianwave.yaml -compare_final_state_atol 2e-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-gaussianwave-IDL-entropy.bin -state_var entropy -dm_plex_box_faces 5,5,1 -ts_max_steps 5 -idl_decay_time 2e-3 -idl_length 0.25 -idl_start 0 -idl_pressure 70 -newtonian_unit_tests
```
jeremylt, line 21:
```suggestion
//TESTARGS(name="Gaussian Wave idl entropy") -ceed {ceed_resource} -test_type solver -options_file examples/fluids/gaussianwave.yaml -compare_final_state_atol 2e-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-gaussianwave-IDL-entropy.bin -state_var entropy -dm_plex_box_faces 5,5,1 -ts_max_steps 5 -idl_decay_time 2e-3 -idl_length 0.25 -idl_start 0 -idl_pressure 70 -newtonian_unit_tests
```
We don't have to use _ in names of test cases instead of spaces anymore

examples/fluids/problems/newtonian.c
```diff
@@ -16,47 +16,138 @@
 #include "../navierstokes.h"
 
 // For use with PetscOptionsEnum
-static const char *const StateVariables[] = {"CONSERVATIVE", "PRIMITIVE", "StateVariable", "STATEVAR_", NULL};
+static const char *const StateVariables[] = {"conservative", "primitive", "entropy", "StateVariable", "STATEVAR_", NULL};
```
jeremylt, line 19: Why not stick with the allcaps? That's what we seem to do with everything else?

jeremylt (approved):
Just a couple of questions/comments

examples/fluids/problems/newtonian.c
jrwrigh: Looks like HONEE has a mix of all-lower-case and mixed capitalization, but this is the only instance of allcaps.

I kind of prefer lower case since that's how most people will be writing out the command line options and it kinda helps with keeping readability in mind.

jrwrigh: I've added a commit to make everything consistently lower case.

jeremylt: Yeah, that's why I made it consistently caps in Ratel so `STATEVAR_` + `CONSERVATIVE` etc is a valid enum value name.

jrwrigh: The documentation for `PetscOptionsEnum` says the array is just a list of options. Following through the code, the connection between the array and the enum is just an ordering thing; the `0` string in the array should refer to the enum value `0`.

I'll go ahead and make them all caps anyways though.

examples/fluids/navierstokes.h
```diff
@@ -33,23 +33,23 @@ typedef enum {
   EULER_TEST_4                 = 4,
   EULER_TEST_5                 = 5,
 } EulerTestType;
-static const char *const EulerTestTypes[] = {"isentropic_vortex", "test_1",      "test_2", "test_3", "test_4", "test_5",
+static const char *const EulerTestTypes[] = {"ISENTROPIC_VORTEX", "TEST_1",      "TEST_2", "TEST_3", "TEST_4", "TEST_5",
                                              "EulerTestType",     "EULER_TEST_", NULL};
```
jeremylt, line 37:
```suggestion
static const char *const EulerTestTypes[] = {"TEST_ISENTROPIC_VORTEX", "TEST_1",      "TEST_2", "TEST_3", "TEST_4", "TEST_5",
                                             "EulerTestType",     "EULER_", NULL};
```
minor, but one way or another these should have a matching pattern

jedbrown (approved):
Looking good. Have you confirmed quadratic Newton convergence if you run Galerkin (so the Jacobian would be exact)? Approved if indeed you have done that.

Perhaps one of the FDSI students will be interested in experimenting on convergence/accuracy studies between the different variable sets.

jrwrigh:
> Have you confirmed quadratic Newton convergence if you run Galerkin (so the Jacobian would be exact)? Approved if indeed you have done that.

Didn't before, but have now. Interestingly, I can't get quadratic convergence from the Gaussian Wave example for any state variable selection, but I can for the channel flow problem.

jedbrown:
It's possible the derivative of the Riemann solver is not exact. If HLL gives quadratic, that would be a smoking gun. I would hope Newton convergence is similar with the different variable sets, even if not quadratic for the Riemann-type conditions. I wish it were easier to unit test this so we would at least have a record of what is quadratic and what is inexact.

jrwrigh:
Either HLL or HLLC doesn't get quadratic convergence.

Convergence between different variable choices varies pretty significantly. Taking the 5th timestep as an example using
```
build/fluids-navierstokes -ceed /cpu/self -options_file examples/fluids/gaussianwave.yaml -state_var conservative -ts_max_steps 5 -ts_monitor -snes_monitor -stab none -snes_rtol 1e-6
```

**Conservative:**
```
4 TS dt 0.002 time 0.008
    0 SNES Function norm 3.866933793909e+00
    1 SNES Function norm 1.080336085412e-01
    2 SNES Function norm 1.281162910157e-02
    3 SNES Function norm 1.730432809866e-03
    4 SNES Function norm 2.430442906523e-04
    5 SNES Function norm 3.448227868442e-05
    6 SNES Function norm 4.990474898545e-06
    7 SNES Function norm 7.529855748397e-07
```

**Primitive:**
```
4 TS dt 0.002 time 0.008
    0 SNES Function norm 3.882167523047e+00
    1 SNES Function norm 2.099883188346e-01
    2 SNES Function norm 4.173983122008e-02
    3 SNES Function norm 7.741725752885e-03
    4 SNES Function norm 1.617289798846e-03
    5 SNES Function norm 3.812815551828e-04
    6 SNES Function norm 9.133517267616e-05
    7 SNES Function norm 2.187075161224e-05
    8 SNES Function norm 5.268857804377e-06
    9 SNES Function norm 1.295585063505e-06
```

**Entropy:**
```
4 TS dt 0.002 time 0.008
    0 SNES Function norm 3.822696166348e+00
    1 SNES Function norm 7.878924784233e-01
    2 SNES Function norm 2.671579538942e-01
    3 SNES Function norm 1.015730388301e-01
    4 SNES Function norm 4.167040137366e-02
    5 SNES Function norm 1.781256117290e-02
    6 SNES Function norm 7.803729063381e-03
    7 SNES Function norm 3.469181893830e-03
    8 SNES Function norm 1.556682257409e-03
    9 SNES Function norm 7.029136648765e-04
   10 SNES Function norm 3.188399923332e-04
   11 SNES Function norm 1.451191310356e-04
   12 SNES Function norm 6.622860601106e-05
   13 SNES Function norm 3.029052811341e-05
   14 SNES Function norm 1.387865628183e-05
   15 SNES Function norm 6.368517859831e-06
   16 SNES Function norm 2.926054422683e-06
```

Comparatively, for the channel flow problem:
```
build/fluids-navierstokes -ceed /cpu/self -options_file examples/fluids/channel.yaml -state_var conservative -ts_max_steps 5 -ts_monitor -snes_monitor -stab none -snes_rtol 1e-6
```

**Conservative:**
```
4 TS dt 5e-06 time 2e-05
    0 SNES Function norm 5.897591477868e-03
    1 SNES Function norm 9.769417352301e-08
    2 SNES Function norm 1.181362654139e-11
```

**Primitive:**
```
4 TS dt 5e-06 time 2e-05
    0 SNES Function norm 5.972752192884e-03
    1 SNES Function norm 3.767480289070e-08
    2 SNES Function norm 1.214635822459e-13
```

**Entropy:**
```
4 TS dt 5e-06 time 2e-05
    0 SNES Function norm 6.225525604360e-03
    1 SNES Function norm 4.735661764131e-06
    2 SNES Function norm 1.997659084392e-10
```

jrwrigh:
Thanks @zatkins-dev for doing all the actually difficult work for this!

zatkins-dev:
Thanks for getting it over the finish line!

knepley:
@jrwrigh For the HLL case, you could try `-snes_fd` to test, which should give you quadratic convergence.

jrwrigh:
Thanks for the tip @knepley. 

I found some errors in the Riemann solver and have pushed a fix in https://github.com/CEED/libCEED/pull/1609, though I'm still not getting quadratic convergence.

PR #1114, opened 2022-12-12
fix(fluids): Change "Owned Nodes" to "Local Nodes"

jrwrigh:
From @jedbrown on Zulip:

> "Owned" should always mean exclusive

"Owned Nodes" are can actually be shared across processes, so it should really be called "Local" to reduce confusion.

### DIFF
--- a/examples/fluids/navierstokes.c
+++ b/examples/fluids/navierstokes.c
@@ -247,15 +247,15 @@ int main(int argc, char **argv) {
     // Mesh
     const PetscInt num_comp_q = 5;
     CeedInt        glob_dofs, owned_dofs;
-    PetscInt       glob_nodes, owned_nodes;
+    PetscInt       glob_nodes, local_nodes;
     const CeedInt  num_P = app_ctx->degree + 1, num_Q = num_P + app_ctx->q_extra;
     // -- Get global size
     PetscCall(VecGetSize(Q, &glob_dofs));
     PetscCall(VecGetLocalSize(Q, &owned_dofs));
     glob_nodes = glob_dofs / num_comp_q;
     // -- Get local size
-    PetscCall(VecGetSize(user->Q_loc, &owned_nodes));
-    owned_nodes /= num_comp_q;
+    PetscCall(VecGetSize(user->Q_loc, &local_nodes));
+    local_nodes /= num_comp_q;
     PetscCall(PetscPrintf(comm,
                           "  Mesh:\n"
                           "    Number of 1D Basis Nodes (P)       : %" CeedInt_FMT "\n"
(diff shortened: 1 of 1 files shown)


examples/fluids/navierstokes.c
```diff
@@ -247,15 +247,15 @@ int main(int argc, char **argv) {
     // Mesh
     const PetscInt num_comp_q = 5;
     CeedInt        glob_dofs, owned_dofs;
-    PetscInt       glob_nodes, owned_nodes;
+    PetscInt       glob_nodes, local_nodes;
     const CeedInt  num_P = app_ctx->degree + 1, num_Q = num_P + app_ctx->q_extra;
     // -- Get global size
     PetscCall(VecGetSize(Q, &glob_dofs));
     PetscCall(VecGetLocalSize(Q, &owned_dofs));
     glob_nodes = glob_dofs / num_comp_q;
```
jedbrown, line 255: Darn, this is also wrong because some of the `global_dofs` are individual components, with other components of the same node eliminated. I'm not aware of a really easy way to get this info. For CGNS, I made calculated the number of global nodes by [reducing](https://gitlab.com/petsc/petsc/-/blob/main/src/dm/impls/plex/cgns/plexcgns2.c#L648) the number of owned nodes, which is [obtained by traversal](https://gitlab.com/petsc/petsc/-/blob/main/src/dm/impls/plex/cgns/plexcgns2.c#L618-630). I'm not wild about using this technique because it doesn't generalized to mixed elements (though we have no intent to use mixed elements, at least not in the near future).

Should we report number of elements? The reason for focusing on nodes is that it normalizes sensibly when varying element order.

jrwrigh: I think number of elements would work ok, but that could also be calculated manually (currently) via the box faces that are reported. 

I think the approximation of dofs/components is ok as long as it's stated that it's an approximation, not the exact number of global nodes.

Does the coordinate dm/vector not store each node exactly? Or would it only do that if we specify isoparametric elements?

jedbrown: Coordinates can be in a different order space. We're running on unstructured meshes sometimes so I think reporting elements is useful. Use `-dm_view` for details.

examples/fluids/navierstokes.c
```diff
@@ -247,15 +247,15 @@ int main(int argc, char **argv) {
     // Mesh
     const PetscInt num_comp_q = 5;
     CeedInt        glob_dofs, owned_dofs;
-    PetscInt       glob_nodes, owned_nodes;
+    PetscInt       glob_nodes, local_nodes;
```
jeremylt, line 250: Side note - what possessed us to call these "glob nodes"?

examples/fluids/navierstokes.c
jedbrown: Are we going to make further changes here or should we merge as is?

jrwrigh: I think we could make it the printed name more clear. But other than that, I think it's good to merge.

jrwrigh: Alright, added a `(DoFs / 5)` print next to the Global Nodes definition. If that's alright with you, then it's good to go for me.

jedbrown (approved):

PR #1115, opened 2022-12-12
fluids: device bind petsc and fix HLLC for CUDA

jedbrown:


examples/fluids/src/cloptions.c
```diff
@@ -98,6 +100,10 @@ PetscErrorCode ProcessCommandLineOptions(MPI_Comm comm, AppCtx app_ctx, SimpleBC
     const char *ceed_resource = "/cpu/self";
     strncpy(app_ctx->ceed_resource, ceed_resource, 10);
   }
+  // If we request a GPU, make sure PETSc has initialized its device (which is
+  // MPI-aware in case multiple devices are available) before CeedInit so that
+  // PETSc and libCEED agree about which device to use.
+  if (strncmp(app_ctx->ceed_resource, "/gpu", 4) == 0) PetscCall(PetscDeviceInitialize(PETSC_DEVICE_DEFAULT()));
```
jeremylt, line 106: Will Ratel need this?

jedbrown: Yeah, on machines like Polaris where the resource manager can't do binding themselves, this would ensure that libCEED stuff is done on the same device that PETSc is using.

Issue #1087, opened 2022-11-10
Multiple Active Basis Diagonal Assembly


jeremylt:
Currently we don't support assembly for a CeedOperator with different bases acting upon the active vector, but we now have a use case for that.

https://gitlab.com/micromorph/ratel/-/merge_requests/233

We can reformulate this specific CeedOperator to use the same bases, though at the expense of efficiency.

It would be good to add the capability to handle different bases acting on the same active vector. We have some other use cases for this in the works as well.

The CPU side should be tedious but straightforward. Once we get the CPU version merged, we can handle the GPU versions.

jeremylt:
The QFunction assembly should be fine.

I think this has 4 parts:

1) CPU diagonal assembly

2) CPU full assembly

3) GPU diagonal assembly

4) GPU diagonal assembly


---

CPU diagonal assembly is the most straightforward and gives us a sense of what we need to do for the rest, so we should start there.

The skeleton of the current diagonal assembly code is

```C
static inline int CeedSingleOperatorAssembleAddDiagonal_Core(CeedOperator op, CeedRequest *request, const bool is_pointblock, CeedVector assembled) {

  // Assemble QFunction
  // -- Should not need changes

  // Get assembly data
  // -- Will need array of active bases here

  // Assemble point block diagonal restriction, if needed
  // -- Will need array of active restrs here, may need multiple point block restrs

  // Create diagonal vector
  // -- Should only need minor tweaks to handle array

  // Assemble element operator diagonals
  // -- Should only need minor tweaks to handle array

  // Basis matrices
  // -- Will need array of interp and grad matrices in/out (eventually will want div too)
  // Compute the diagonal of B^T D B
  // Each element
  for (CeedInt e = 0; e < num_elem; e++) {
     // -- num_elem will be the same across all bases
    // Each basis eval mode pair
    for (CeedInt e_out = 0; e_out < num_eval_mode_out; e_out++) {
      for (CeedInt e_in = 0; e_in < num_eval_mode_in; e_in++) {
        // -- Getting basis matrices for emode should only need minor tweaks
        // Each component
        for (CeedInt c_out = 0; c_out < num_comp; c_out++) {
           // -- num_comp should depend upon the specific basis
          // Each qpoint/node pair
          for (CeedInt q = 0; q < num_qpts; q++) {
            // -- q must always be the same across all bases
            if (is_pointblock) {
              // Point Block Diagonal
[...]

PR #1116, opened 2022-12-12
Multiple Active Basis Diagonal Assembly

rezgarshakeri:
See https://github.com/CEED/libCEED/issues/1087

### DIFF
--- a/include/ceed-impl.h
+++ b/include/ceed-impl.h
@@ -300,7 +300,7 @@ struct CeedQFunctionAssemblyData_private {
 
 struct CeedOperatorAssemblyData_private {
   Ceed          ceed;
-  CeedInt       num_eval_mode_in, num_eval_mode_out;
+  CeedInt       num_active_in, num_active_out;
   CeedEvalMode *eval_mode_in, *eval_mode_out;
   CeedScalar   *B_in, *B_out;
   CeedBasis     basis_in, basis_out;
@@ -349,6 +349,7 @@ struct CeedOperator_private {
   CeedInt                   num_context_labels;
   CeedInt                   max_context_labels;
   CeedContextFieldLabel    *context_labels;
+  CeedBasis                *active_basis_in, *active_basis_out;
 };
 
 CEED_INTERN int CeedOperatorGetFallback(CeedOperator op, CeedOperator *op_fallback);
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -234,6 +234,42 @@ int CeedOperatorGetActiveBasis(CeedOperator op, CeedBasis *active_basis) {
 }
 
 /**
+  @brief Find the active vector bases for a non-composite CeedOperator
+
+  @param[in] op                CeedOperator to find active bases for
+  @param[in] num_active_in     Number of active input CeedEvalModes
+  @param[in] num_active_out    Number of active output CeedEvalModes
+  @param[out] active_basis_in  Bases for active input vectors
+  @param[out] active_basis_out Bases for active output vectors
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ ref Developer
+**/
+int CeedOperatorGetActiveBases(CeedOperator op, CeedInt num_active_in, CeedInt num_active_out, CeedBasis **active_basis_in,
+                               CeedBasis **active_basis_out) {
+  Ceed ceed;
+  CeedCall(CeedOperatorGetCeed(op, &ceed));
+
+  if (op->is_composite) {
+    // LCOV_EXCL_START
+    return CeedError(ceed, CEED_ERROR_MINOR, "composite operator is not supported");
+    // LCOV_EXCL_STOP
+  }
+  CeedCall(CeedCalloc(num_active_in, &op->active_basis_in));
+  CeedCall(CeedCalloc(num_active_out, &op->active_basis_out));
+  for (CeedInt i = 0; i < num_active_in; i++) {
+    op->active_basis_in[i] = op->input_fields[i]->basis;
+  }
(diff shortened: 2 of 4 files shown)


include/ceed-impl.h
```diff
@@ -349,6 +349,7 @@ struct CeedOperator_private {
   CeedInt                   num_context_labels;
   CeedInt                   max_context_labels;
   CeedContextFieldLabel    *context_labels;
+  CeedBasis                *active_basis_in, *active_basis_out;
```
jeremylt, line 352: I'm belatedly realizing that we probably want this as part of 

```
struct CeedOperatorAssemblyData_private {
  Ceed          ceed;
  CeedInt       num_active_in, num_active_out;
  CeedEvalMode *eval_modes_in, *eval_modes_out;
  CeedScalar   *B_in, *B_out;
  CeedBasis     *bases_in, bases_out;
  CeedElemRestriction *elem_restrs_in, *elem_restrs_out;
};
```

interface/ceed-operator.c
```diff
@@ -233,6 +233,43 @@ int CeedOperatorGetActiveBasis(CeedOperator op, CeedBasis *active_basis) {
...
+
+  if (op->is_composite) {
+    // LCOV_EXCL_START
+    return CeedError(ceed, CEED_ERROR_MINOR, "composite operator is not supported");
+    // LCOV_EXCL_STOP
+  }
+  CeedCall(CeedCalloc(op->qf->num_input_fields, &op->active_basis_in));
+  CeedCall(CeedCalloc(op->qf->num_output_fields, &op->active_basis_out));
+  for (CeedInt i = 0; i < op->qf->num_input_fields; i++) {
+    if (op->input_fields[i]->vec == CEED_VECTOR_ACTIVE) {
```
jeremylt, line 259: I don't think we want these NULL spaces in this - it will make these arrays harder to use

include/ceed-impl.h
rezgarshakeri: Yeah, so you mean make `basis_in/out` to array and use this function `CeedOperatorAssemblyDataCreate`?
It seems we get the active basis there.

interface/ceed-operator.c
jeremylt: Right now, for an operator with the inputs

`[interp basis_1, passive qdata, grad basis_1, interp basis_2]`

this array will hold

`[basis_1, NULL, basis_1, basis_2]`

That's sort of awkward compared to how this array will be used. For assembly, we loop over active eval modes like this

`[interp basis_1, grad basis_1, interp basis_2]`

rezgarshakeri: I see, thanks

interface/ceed-preconditioning.c
```diff
@@ -1135,8 +1135,8 @@ int CeedOperatorAssemblyDataCreate(Ceed ceed, CeedOperator op, CeedOperatorAssem
       }
     }
   }
-  (*data)->num_eval_mode_out = num_eval_mode_out;
-  (*data)->eval_mode_out     = eval_mode_out;
+  (*data)->num_active_out = num_active_out;
+  (*data)->eval_mode_out  = eval_mode_out;
   CeedCall(CeedBasisReferenceCopy(basis_out, &(*data)->basis_out));
```
rezgarshakeri, line 1141: @jeremylt I was thing to change `basis_in/out` to array `*basis_in/out` in `CeedOperatorAssemblyData_private`, then put `CeedBasisReferenceCopy` in the loop so we get the active bases, what do you think?

rezgarshakeri: Yeah, I create a loop in `CeedOperatorAssemblyDataDestroy` function to destroy all the active bases.

interface/ceed-preconditioning.c
```diff
@@ -1078,16 +1078,16 @@ int CeedOperatorAssemblyDataCreate(Ceed ceed, CeedOperator op, CeedOperatorAssem
...
         case CEED_EVAL_GRAD:
-          CeedCall(CeedRealloc(num_eval_mode_in + dim, &eval_mode_in));
+          CeedCall(CeedRealloc(num_active_in + dim, &eval_mode_in));
           for (CeedInt d = 0; d < dim; d++) {
-            eval_mode_in[num_eval_mode_in + d] = eval_mode;
+            eval_mode_in[num_active_in + d] = eval_mode;
           }
-          num_eval_mode_in += dim;
+          num_active_in += dim;
           break;
```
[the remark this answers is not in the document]
jeremylt: I think I'd still use 4 as the number of 'active inputs' here. (If you have an idea for a better phrase?) In the assembly, we handle each direction of the gradient separately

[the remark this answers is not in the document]
rezgarshakeri: Yeah, I understand that, but do we need to allocate 2 bases or 4 here? 
```c
CeedCall(CeedCalloc(2, &active_basis_in));
```
The idea was to copy the active basis using `CeedCall(CeedBasisReferenceCopy(basis_out, &(*data)->basis_out));`. I guess we need to copy 2 bases, right?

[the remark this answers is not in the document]
jeremylt: No, that would bring us right back to the problem of messy indexing to figure out what basis were on. 

This seems to be getting a lot messier than I thought it would be.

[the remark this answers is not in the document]
jeremylt: No, that would bring us right back to the problem of messy indexing to figure out what basis were on. 

This seems to be getting a lot messier than I thought it would be.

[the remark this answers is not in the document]
rezgarshakeri: I think I know what's going on, I'll update it next week.

PR #1118, opened 2022-12-20
Mixed-Elasticity

rezgarshakeri:
- [x] Setup snes with zero Dirichlet BC for the linear mixed-elasticity and got desired convergence order for Q2-Q1 element
- [ ] Need to fix the restriction for Q2-P0 element
- [ ] Add boundary conditions
- [ ] Extend to Nonlinear case


jedbrown:
Cool. I think it makes sense to have a basic demo of mixed elasticity here (for pedagogical reasons), but do the nonlinear extensions in Ratel. Even just handling different elements can go there.

rezgarshakeri:
Yeah we can do that. The permutation of Q2-P0 is not correct so need to be fixed in petsc as I discussed with @knepley in this [issue](https://gitlab.com/petsc/petsc/-/issues/1237)

examples/mixedelasticity_KSP/qfunctions/volumetric-geometry-3d.h
```diff
@@ -0,0 +1,44 @@
...
+        {J[0][2][i], J[1][2][i], J[2][2][i]}
+    };
+    const CeedScalar detJ = ComputeDet3(dxdX);
+    CeedScalar       dXdx_voigt[9];
+    MatComputeInverseNonSymmetric3(dxdX, detJ, dXdx_voigt);
+    q_data[0][i] = w[i] * detJ;
+    for (CeedInt j = 0; j < 9; j++) {
+      q_data[j + 1][i] = dXdx_voigt[j];
+    }
+  }  // End of Quadrature Point Loop
```
rezgarshakeri, line 40: @jedbrown @jeremylt I have one question here. Since the MMS elasticity doesn't give the correct order of convergence in Ratel, I copied BP4 from example/petsc then extended to linear elasticity and after that mixed-elasticity to make sure everything is correct. But instead of creating `q_data` as we have in libCEED/Ratel (by creating matrix A[3][3], ...) I tried to use `MatComputeInverseNonSymmetric3` function which I tested for random 3x3 matrix in a notebook.
This setup gives exactly the same L2 error/number of iteration/convergence order as we have in `example/petsc/bps` (bp4) but only for `hex` element not for `tet`, why?
And the only difference is how I created `q_data`, is it wrong to setup `dxdX` like above for `tet` element?

jeremylt: I made a change in Ratel yesterday that may have fixed the order of convergence

rezgarshakeri: No ratel still doesn't give the correct order. My question here is about bp4/tet element. Why when I setup q_data like above works for `hex` but not `tet`?

rezgarshakeri: Elasticity example in this branch works fine, you can run by `./conv_test.sh -o 1` (or 2,3,4) Then run `python conv_rate.py -f conv_test_result.csv`. Results for hex element, order 1,2,3,4
```
   Number of element/direction  convergence order of u  convergence order of p
0                            4                0.000000                     0.0
1                            6                1.812415                     NaN
2                            8                1.929877                     NaN
3                           10                1.965307                     NaN
4                           12                1.979913                     NaN

   Number of element/direction  convergence order of u  convergence order of p
0                            4                0.000000                     0.0
1                            6                2.998729                     NaN
2                            8                2.988485                     NaN
3                           10                2.986055                     NaN
4                           12                2.986057                     NaN

   Number of element/direction  convergence order of u  convergence order of p
0                            4                0.000000                     0.0
1                            6                3.996939                     NaN
2                            8                4.000354                     NaN
3                           10                4.000964                     NaN
4                           12                4.000969                     NaN

   Number of element/direction  convergence order of u  convergence order of p
0                            4                0.000000                     0.0
1                            6                4.985871                     NaN
2                            8                4.991083                     NaN
3                           10                4.993449                     NaN
4                           12                4.994788                     NaN
```

jedbrown: Is pressure convergence not calculated or somehow not converging?

rezgarshakeri: That is elasticity example and pressure is not computed. For mixed Q2-Q1 we have
```
   Number of element/direction  convergence order of u  convergence order of p
0                            4                0.000000                0.000000
1                            6                3.151930                2.076015
2                            8                3.059559                2.041752
3                           10                3.025034                2.026101
```

jeremylt:
Not urgent at this point - but I think there is probably some room for code reuse with the functions copy-pasted from the `examples/petsc` folder?

rezgarshakeri:
> 
In `mixedelasticity_KSP` folder I setup the problem with KSP solver like `example/petsc`. You can run by
```
./main -problem bp4-3d -dm_plex_dim 3 -dm_plex_box_faces 6,6,6 -dm_plex_simplex 0
```
Which gives exactly same results as `./bps -problem bp4` (using same number of elements). Then I extended it to elasticity and mixed-elasticity. I kept this folder to see what is the issue of `tet` element. The only difference with `example/petsc` is the way I setup the `q_data` Qfunction.

In `mixedelasticity` folder I have only mixed problem but with SNES solver. All problem work with `hex` element and give the expected convergence order.

rezgarshakeri:
I don't know why the non-zero dirichlet boundary doesn't work (I don't get the correct convergence order). As a simple test, I added a constant (right now is 0. ) to `u2` in `qfunctions/mixed-linear-elasticity-2d.h` and to the `u[1]` in `BoundaryDirichletMMS` function in `src/setup-boundary.c`.
What am I missing in the BC setup?

jedbrown:
Can you visualize the desired and computed solutions?

rezgarshakeri:
Interesting, I added the constant `2.0` to the y component and it seems the boundary is applied correctly but not the error.
Uy

rezgarshakeri:
BTW, convergence order for pressure is correct
```console
/libCEED/examples/mixedelasticity$ ./conv_test.sh -d 2 -u 2 -p 1
Running convergence test for 2D mixed-linear elasticity with polynomial order u:2, p:1
   Number of element/direction  L2 Error u  Convergence Order of u  L2 Error p  Convergence order of p
0                            4    0.381119                0.000000    0.101004                0.000000
1                            6    0.317647                0.449288    0.043528                2.076015
2                            8    0.277629                0.468063    0.024192                2.041752
3                           10    0.249617                0.476643    0.015393                2.026101
```

jeremylt:
Since this work moved to Ratel, do we want to close this PR?

rezgarshakeri:
> Since this work moved to Ratel, do we want to close this PR?

Yeah we can close it.

PR #1120, opened 2022-12-20
Rust coverage

jedbrown:
actions-rs/tarpaulin seems abandoned and llvm-cov is reportedly more accurate and used by the savvy projects I surveyed. Seems more likely to be maintained.

### DIFF
--- a/.github/workflows/rust-test-with-style.yml
+++ b/.github/workflows/rust-test-with-style.yml
@@ -18,19 +18,19 @@ jobs:
     - name: Rust setup
       uses: dtolnay/rust-toolchain@master
       with:
-        # Note: nightly required for coverage of Doctests with tarpaulin
+        # Note: nightly required for coverage of Doctests
         toolchain: nightly
+    - name: Install cargo-llvm-cov
+      uses: taiki-e/install-action@cargo-llvm-cov
     - name: Rust test with coverage
-      uses: actions-rs/tarpaulin@v0.1
       env:
         CC: ${{ matrix.compiler }}
         FC: gfortran-11
-      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-test-with-style.yml
```diff
@@ -18,19 +18,19 @@ jobs:
...
+    - name: Install cargo-llvm-cov
+      uses: taiki-e/install-action@cargo-llvm-cov
     - name: Rust test with coverage
-      uses: actions-rs/tarpaulin@v0.1
       env:
         CC: ${{ matrix.compiler }}
         FC: gfortran-11
-      with:
-        args: '--run-types Doctests Tests --exclude-files backends/* gallery/* include/* interface/* --out Xml'
+      run: cargo llvm-cov test --doctests --lcov --output-path lcov.info
```
jeremylt, line 29: Will this still exclude the libCEED C files from the coverage?

jedbrown: Yeah, it excludes them, but there is a way to include FFI coverage (that I'm not activating now, but perhaps we would want in the future).

jeremylt: The coverage in the Rust files looks a lot more accurate now

jeremylt (approved):

.github/workflows/rust-test-with-style.yml
jedbrown: Huh, I wouldn't have expected this much change from using `llvm-cov`, but I'll take it.

https://app.codecov.io/gh/CEED/libCEED/pull/1120

image

jeremylt: Its because tarpaulin incorrectly marked a ton of lines as 'uncovered' when one function call split across multiple lines

PR #1128, opened 2023-01-02
Draft: make examples use periodicity and SFC box meshes

jedbrown:
This depends on https://gitlab.com/petsc/petsc/-/merge_requests/5411 so won't be ready for merging soon/ever, but I wanted to put examples out there for discussion.

### DIFF
--- a/examples/fluids/qfunctions/setupgeo.h
+++ b/examples/fluids/qfunctions/setupgeo.h
@@ -77,6 +77,7 @@ CEED_QFUNCTION(Setup)(void *ctx, CeedInt Q, const CeedScalar *const *in, CeedSca
     const CeedScalar A33  = J11 * J22 - J12 * J21;
     const CeedScalar detJ = J11 * A11 + J21 * A12 + J31 * A13;
 
+    if (detJ <= 0) printf("%s: detJ %g\n", __func__, detJ);
     // Qdata
     // -- Interp-to-Interp q_data
     q_data[0][i] = w[i] * detJ;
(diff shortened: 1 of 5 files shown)


examples/fluids/qfunctions/setupgeo.h
```diff
@@ -77,6 +77,7 @@ CEED_QFUNCTION(Setup)(void *ctx, CeedInt Q, const CeedScalar *const *in, CeedSca
     const CeedScalar A33  = J11 * J22 - J12 * J21;
     const CeedScalar detJ = J11 * A11 + J21 * A12 + J31 * A13;
 
+    if (detJ <= 0) printf("%s: detJ %g\n", __func__, detJ);
```
jedbrown, line 80: @jeremylt Do you have thoughts about how we should report such a thing to the user? This is especially noisy (will generate millions of lines on real problems) so not suitable for merging, but it would be good to have a clear message to the user about inverted elements (usually represents a mesh IO problem or issues [like this](https://gitlab.com/petsc/petsc/-/issues/1302) that I'm pretty sure we've stumbled into).

jeremylt: Making a debug macro for qfunction use would be easy enough, but reporting only once would be harder. No obvious solution comes to mind

jeremylt: We could make a mesh checking gallery QFunction that uses context to flag if the mesh has inverted elements and only report once

jrwrigh:
With #1529, we officially have a `zbox` simulation in the test suite.

jedbrown:
Thanks @jrwrigh, it's there anything we should salvage in this branch or just close it as superseded?

jrwrigh:
Not that I can think of, unless you were wanting to keep the conversion of gaussianwave to 3D or the inverted element checks.

PR #1130, opened 2023-01-04
Lift, Drag, Heat Transfer

AdelekeBankole:


AdelekeBankole:
@jedbrown which function calls do we need in here?

jedbrown:
Can you start by trying to get what you have to type-check and calling the new function you wrote at an appropriate place? These functions have documentation and there are examples. I wrote more tips in Zulip last night.

jedbrown (commented):
I pushed a revision of this. It runs in a simple test. Please try running it on the cylinder and validate the results (reference the literature or do other consistency checks). I only tested linear elements. If this works, we'll need to wire in the input and move this calculation to a monitor.

jedbrown:
I'm okay with skipping heat flux in this branch, but we still need to:

- [ ] Test for correctness. You could do this by comparing the lift and drag coefficients in Fig 6.3.4 of Shakib.
- [ ] Make a Python script that creates an equivalent plot.
- [ ] Document this feature in the markdown docs.

AdelekeBankole:
Alright, I will work on these bullet points.

AdelekeBankole:
Please @jedbrown, something to look into, our drag coefficient does not look like  how the profile should be. 
Our profile
Screenshot from 2023-02-13 15-40-43
vs Shakib
Screenshot from 2023-02-13 15-36-49

jedbrown:
@AdelekeBankole Please check the y scale.

AdelekeBankole:
Actually, I checked that y scale, because if we zoom into it in the neighborhood of say ```0.5``` Can we still have that dropping profile in the limit as ```t``` approaches ```40```?

jedbrown:
Here's an updated version that normalizes. I don't know why Shakib's drag plot has a nonsymmetric wiggle, but I don't think it's physical. Ours looks symmetric. It'll be interesting to learn what is grid converged, but this looks close enough for now.

image

jedbrown:
The plotting file needs to be tucked away somewhere specific to this study, but can stay in the repo. The data file can be deleted and squashed out in the merge.

AdelekeBankole:
Thank you @jedbrown. I have seen how the normalization is done using the formula.

jedbrown:
How about we call the script `examples/fluids/postprocess/vortexshedding.py` so it doesn't get lost and so nobody else has to recreate this wheel, perhaps if the test is used in a class or otherwise?

Here's some code for accurately computing the shedding period.
```py
import scipy.signal as sig

sample = df[df['Time'] > 70]    # once the initial transient has passed
peaks, _ = sig.find_peaks(sample['ForceY'])
period = np.diff(sample['Time'].iloc[peaks])
print(period)  # [5.55 5.6  5.55 5.55]
print(period.mean()) # 5.5625
```
The fact that the last few shedding periods are almost identical gives us confidence in using this reported mean. The script should report this shedding period (maybe in the figure title).

AdelekeBankole:
Absolutely, that's perfect that it lives in such directory.

jedbrown (approved):

PR #1132, opened 2023-01-04
Update user manual citation

jeremylt:
I updated the Zenodo citation and put a badge to the link that automatically points to the latest version.

https://github.com/CEED/libCEED/tree/jeremy/cite-minor-fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,6 @@
 [![Code coverage][codecov-badge]][codecov-link]
 [![BSD-2-Clause][license-badge]][license-link]
 [![Documentation][doc-badge]][doc-link]
-[![User manual][zenodo-badge]][zenodo-link]
 [![JOSS paper][joss-badge]][joss-link]
 [![Binder][binder-badge]][binder-link]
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,7 @@
 [![Code coverage][codecov-badge]][codecov-link]
 [![BSD-2-Clause][license-badge]][license-link]
 [![Documentation][doc-badge]][doc-link]
+[![User manual][zenodo-badge]][zenodo-link]
```
jedbrown, line 8: Thanks. My only question here is whether we want to prominently point people to the PDF version of the website docs. Placed in this way, I'd think it's different content. Maybe the documentation site and/or later in the README should just say that equivalent information is available in PDF form at [zenodo link].

jedbrown (approved):

README.md
jeremylt: I was wondering that too. I had it as a badge down with the citations  section and it looked out of place. Maybe a simple link would look less so

jedbrown: Yeah, I think over there with or without badge or clarifying text saying that it's a PDF equivalent of the documentation website.

jeremylt: Updated

jedbrown (approved):

PR #1131, opened 2023-01-04
fluids: checkpoint interval

jedbrown:
* Deprecate `output_freq`, but preserves existing behavior
* Add `checkpoint_interval` for binary checkpoints
* Use `-checkpoint_vtk` to include VTK output alongside binary checkpoints (as with now-deprecated `output_freq`)
* Write step number and time to `ns-solution.bin` checkpoint files; read on restart. So you can now
```yaml
continue_filename: ns-solution-40.bin
```
without any `continue: 40` or `continue_time_filename: ns-time-40.bin` part and it'll pick up correctly.
* If using default (`-output_add_stepnum2bin false`) names, one can use `-continue -1` to read `ns-solution.bin`, picking up the step number and time automatically. If a number is specified in `-continue`, we use it even if it's inconsistent with the value in the file. Note that `-continue 20` and `-continue_time_filename` (whatever the correct step number is) are still needed if reading a legacy checkpoint.

---
## Demo

On main (prior to this PR), I ran this to generate the first 20 time steps and create a (legacy) checkpoint.
```console
$ mpiexec -n 5 build/fluids-navierstokes -options_file examples/fluids/vortexshedding.yaml -{ts,snes}_monitor -ksp_converged_reason  -ts_max_steps 20 -output_dir v1
```

I updated to this branch and recompiled, running the next 20 time steps
```console
$ mpiexec -n 5 build/fluids-navierstokes -options_file examples/fluids/vortexshedding.yaml -{ts,snes}_monitor -ksp_converged_reason  -ts_max_steps 40 -output_dir v2 -continue 20 -continue_filename v1/ns-solution.bin -continue_time_filename v1/ns-time.bin
[...]
20 TS dt 0.05 time 1.
    0 SNES Function norm 1.810353133257e+01
    Linear solve converged due to CONVERGED_RTOL iterations 220
    1 SNES Function norm 2.359853698058e-02
    Linear solve converged due to CONVERGED_RTOL iterations 203
    2 SNES Function norm 5.154421628365e-05
    Linear solve converged due to CONVERGED_RTOL iterations 140
    3 SNES Function norm 3.932162726632e-07
[...]
```

Now I have a new-style checkpoint and can continue with just `-continue_filename`.
```console
$ mpiexec -n 5 build/fluids-navierstokes -options_file examples/fluids/vortexshedding.yaml -{ts,snes}_monitor -ksp_converged_reason  -ts_max_steps 60 -output_dir v3 -continue_filename v2/ns-solution.bin
[...]
[...]

PR #1133, opened 2023-01-06
Switch outflow from pressure to Riemann pressure + temperature

jedbrown:
This BC allows an "outflow" boundary to recirculate stably. Some examples that justify ensuring that "outflow" conditions be stable when not strictly outflow

* strong acoustic waves exiting a low velocity flow cause a transient flow reversal
* far-field boundaries like the y_max boundary for flat plates, if insufficiently angled/far away from the wall, will have moments of inflow.
* If the downstream boundary for a cylinder is too close, the recirculation crosses the boundary. Here is a stable simulation with sustained recirculation through the boundary.

image

* In SRS, we see momentary pockets of negative x velocity near the outflow (black specks are isosurfaces of negative velocity). These cause sudden non-convergence of the linear solve when they intersect the outflow plane.

image

Depends on #1131 .


jedbrown:
@jrwrigh and @KennethEJansen The Blasius tests are failing because this outflow behaves enough different from the purely pressure-based outflow. I think we need to decide whether the behavior in this branch is just better and re-image our tests or to support `-bc_pressure_outflow` (not to be confused with `-outflow_pressure`).

I'd also like to revisit and probably delete `P0`, which is now inherited from `-reference_pressure`. Note that one currently has flexibility
```yaml
reference_pressure: 7143   # defaults
freestream_pressure: 7150  # applied on freestream boundaries
outflow_pressure: 7140     # at outflow boundaries
```
Note that both `freestream` and `output` take their defaults from `reference`.

At some point, I'd like to try this technique for near-wall inflows as well, evaluating a Riemann-based inviscid flux plus a viscous flux. It may be less reflective than the strong inflow we currently use. Another variant (probably more interesting at inflows) is to use a "reaction force" technique to compute viscous flux (which would thus include stabilization) instead of the surface integral that appears in our current weak inflows. Ideally that would render the IDL unnecessary, but our current approach of strong velocity with IDL seems to be working well enough for STG so let's not retread this now.

jedbrown:
I modified this to keep the `outflow_type: pressure` option, though based on what I've seen, I would recommend always using the default (`outflow_type: riemann`). I also updated docs.

PR #1134, opened 2023-01-07
fluids: report TS converged reason and exit cleanly

jedbrown:


examples/fluids/src/misc.c
```diff
@@ -174,8 +175,10 @@ PetscErrorCode PostProcess_NS(TS ts, CeedData ceed_data, DM dm, ProblemData *pro
 
   // Print final time and number of steps
   PetscCall(TSGetStepNumber(ts, &steps));
+  PetscCall(TSGetConvergedReason(ts, &reason));
   if (!user->app_ctx->test_mode) {
-    PetscCall(PetscPrintf(PETSC_COMM_WORLD, "Time integrator took %" PetscInt_FMT " time steps to reach final time %g\n", steps, (double)final_time));
+    PetscCall(PetscPrintf(PETSC_COMM_WORLD, "Time integrator %s on time step %" PetscInt_FMT " with final time %g\n", TSConvergedReasons[reason],
```
jedbrown, line 180: @jrwrigh @jeremylt I think this is more accurate in case of restart. Should we explicitly state the starting time step in this output?

jrwrigh: > Should we explicitly state the starting time step in this output?

I don't think it's necessary per-say since the log itself will effectively say which timestep it started on. But I usually lean towards redundancy for log output as long as it isn't overwelming. And I don't think adding a `"Ran from time steps ____ to ____"` makes it overwhelming. Though it's make sense to do that with the solution time as well.

examples/fluids/src/setupts.c
```diff
@@ -450,6 +450,7 @@ PetscErrorCode TSSolve_NS(DM dm, User user, AppCtx app_ctx, Physics phys, Vec *Q
   }
   PetscCall(TSSetMaxTime(*ts, 500. * user->units->second));
   PetscCall(TSSetExactFinalTime(*ts, TS_EXACTFINALTIME_STEPOVER));
+  PetscCall(TSSetErrorIfStepFails(*ts, PETSC_FALSE));
```
jrwrigh, line 453: Does this also remove the error output if TS exits with a failure? ie:

```
[0]PETSC ERROR: Invalid argument
[0]PETSC ERROR: Wrong type of object: Parameter # 1
[0]PETSC ERROR: WARNING! There are option(s) set that were not used! Could be the program crashed before they were used or a spelling mistake, etc!
[0]PETSC ERROR: Option left: name:-spanstats_dm_sparse_localize value: 0 source: code
   [......]
[0]PETSC ERROR: Petsc Development GIT revision: v3.18.3-328-gecdab737cc0  GIT Date: 2023-01-02 11:45:34 -0700
[0]PETSC ERROR: ./navierstokes on a arch-dev named fermat by jrwrigh Fri Jan  6 20:21:15 2023
[0]PETSC ERROR: Configure options --with-64-bit-indices=0 --download-hdf5 --download-cgns --download-tetgen=1 --download-yaml=1 PETSC_ARCH=arch-dev
[0]PETSC ERROR: #1 PetscFEGetHeightSubspace() at /home/jrwrigh/software/petsc/src/dm/dt/fe/interface/fe.c:1692
   [......]
[0]PETSC ERROR: PETSc Option Table entries:
[0]PETSC ERROR: -bc_inflow 6 (source: file)
```

In the case of a failure, it's probably not bad to still include this information just in case it can help illuminate something (mispelled options, wrong Petsc version, etc.)

examples/fluids/src/misc.c
jedbrown: If you run with no monitors, you get something like this
```console
$ mpiexec -n 5 build/fluids-navierstokes -options_file examples/fluids/newtonianwave.yaml

  MPI:
    Host Name                          : kichatna
    Total ranks                        : 5
  Problem:
    Problem Name                       : newtonian_wave
  libCEED:
  PETSc:
    Box Faces                          : 40,40,1
  Mesh:
    Global DoFs                        : 13448
    Owned DoFs                         : 2616
    Global nodes (DoFs / 5)            : 2689
    Local nodes                        : 736
Time taken for solution (sec): 2.22668
Time integrator CONVERGED_ITS on time step 100 with final time 0.2
```

Perhaps we should announce the starting step/time and intended final step/time before calling `TSSolve`. More wording suggestions welcome.

jrwrigh (approved):
LGTM (reviewing only the TS converged reason parts).

examples/fluids/src/setupts.c
jedbrown: :point_up:  That error hasn't even gotten to TS.

It only prevents TS divergence (perhaps because SNES and KSP diverged, or because an adaptive controller gives up) from raising an error.

jrwrigh: Right, I just posted that error message as an example of useful info that's put into the error message. Particularly, the `option(s) set that were not used` part is particularly useful.

It's not uncommon to run into simple mistakes (such as misspelled options) when bootstrapping a new simulation for the first time and having those breadcrumbs to verify what the code is doing can be quite helpful.

examples/fluids/src/misc.c
jrwrigh: Something at the beginning like: 
```
Restart file found: <path to file>
Loading initial condition from restart at step <step num> and solution time <solution time>
```

I don't think the intended final step/time is necessary. It would actually be nice to be able to change that on the fly, but that's a convenience feature for a (probably much) later time.

examples/fluids/src/setupts.c
jedbrown: Good point. Arguably we should always use `-options_left` (or our code can print the equivalent). It's part of `-log_view`, which is typically always used for provenance in production runs, but noisy for routine testing.

examples/fluids/src/misc.c
jedbrown: Change it by writing to a control file or socket?

jrwrigh: > Change it by writing to a control file or socket?

PHASTA's implementation uses a control file to change the ending timestep. At the beginning of every timestep, it looks for a specific file. If it's present, it reads it, changes the end timestep, and deletes the file. Otherwise it just keeps solving.

jedbrown: Revised to print this
```
  Continue:
    Filename:                          : ./ns-solution.bin
    Step:                              : 10
    Time:                              : 0.02
```

PR #1125, opened 2022-12-25
fluids: Statistics collection

jrwrigh:
Started from @Mateus-A-R-Braga's work at https://github.com/Mateus-A-R-Braga/libCEED/tree/mateus-a-r-braga/reynolds_stress

This will initially focus on global statistics collection, then move onto spanwise-averaged statistics.

Outline:
- Data collection
  - Data stored to new DM object, which contains PetscSection for the statistics items
  - Mass matrix formed for that DM object specifically (necessary until we can get a `VecPointwiseMultiply` going to apply the same $M^{-1}$ scalar to each node)
  - Data is collected to DMLocal vectors at set intervals (not necessarily every timestep)
  - Product statistics ( $\langle u_i u_j \rangle$ ) are $L^2$ projected
- Data Writing
  - Data written to CGNS files from the statistics DM object
  - Any necessary communication should be done upon writing, not when collecting (if possible)

Todo:
- [x] Cleanup commits
- [ ] Create new statistics DM
   - [x] `DMClone`
   - [x] Create new section with CGNS names (if they exist)
   - [x] Create new `CeedElemRestriction` for the projected vectors in the new DM object
   - [x] Create mass matrix for the stats DM object
   - [ ] Translate "normal" BC definitions into statistics BCs
 - [x] Create new statistics struct to store data
    - Includes local vectors, stats DM, CeedOperator, etc.

PR #1135, opened 2023-01-08
fluids: Spanwise Statistics Collection

jrwrigh:
Spanwise statistics collection. Note the method here exploits the tensor-product grid and element structure.

To Do:
- [x] Create statistics `DM`
- [x] ~Set BCs for statistics `DM` to match dirchlet BCs for the projection~ Not done, consistent projection is satisfactory
- [x] Create `PetscSF` for child->parent communication
- [x] Create Child data QF to do quadrature-level products
- [x] Use `PetscSF` to reduce child data to parents
- [x] Create parent data QF to do RHS $L^2$ projection
- [x] Create `KSP` to solve the mass matrix problem
- [x] Create CL options for controlling statistics
- [x] Create TSMonitor for processing and collecting (separately?)
- [x] Implement actual time averaging
- [x] Destroy all ceed and Petsc objects (cleanup at the end)

Code Outline:
- Setup
	- Create 2D DM
	- Create `SF`
		- Use QF to get collocated (quadrature point) coordinates for 2D and 3D
- Collection
   - Every averaging timestep
       - Could be every timestep, or every nth timestep 
	- Average $fw^q$ at collocated quadrature points
	- No communication is done here; just local data averaging
- Processing
   - Every stats file write
	- Use `SF` to get 2D collocated data
	- Use QFunction to compute 2D RHS
		- Input is collocated 2D quadrature points
	- Solve for the $L^2$ projection
	- Write file


jedbrown:
> Set BCs for statistics DM to match dirchlet BCs for the projection

I don't think this is necessary. It's an $L^2$ projection so doesn't *need* boundary conditions, and I suspect not all fields of interest will go to zero or even a known value at the boundary. Unless I'm missing some need, I would just project and we can look at the fields.

In terms of order of operations, I would make the mass matrix projection in the 2D plane first and test that it works (project a known function defined on quadrature points; could even use the gallery qfunctions to start). You should be able to `VecViewFromOptions(solution, NULL, "-stats_vec_view")` and `-stats_vec_view cgns:stats.cgns` to visualize. Getting this working with placeholder data will make debugging the rest go faster.

Then I'd use synthetic 3D data and pass it through the SF to ensure you get the same results. Only after verifying that would I put in actual Reynolds stress, etc.

I think in the end, we want to set this as a `TSMonitor`, so something like `-ts_monitor_turbulence cgns:stats.cgns` with suboptions as needed.

jrwrigh:
> > Set BCs for statistics DM to match dirchlet BCs for the projection
> 
> I don't think this is necessary. It's an L2 projection so doesn't _need_ boundary conditions, and I suspect not all fields of interest will go to zero or even a known value at the boundary. Unless I'm missing some need, I would just project and we can look at the fields.

Not all of the fields will go to zero, but, for example, the velocity products will go to zero at a no-slip wall. Without fixing those values, then the projection will allow them to be non-zero. This is only true for dirichlet BCs. For example, from @Mateus-A-R-Braga's CFD project results, we get non-zero velocity-product projection at the wall for the channel flow:
image

I believe Ken has also said that he doesn't expect there to be a significant difference here, so I'm leaving that part till the end regardless.

> I think in the end, we want to set this as a TSMonitor, so something like -ts_monitor_turbulence cgns:stats.cgns with suboptions as needed.

Can we create multiple `TSMonitors` for a single `TS`? Or would it be just integrating the new functionality into the already existing `TSMonitor_NS`?

jedbrown:
Yes, you can `TSMonitorSet()` any number of monitors. Indeed, `-ts_monitor -ts_monitor_solution cgns:flow.cgns` creates two such monitors.

jedbrown:
Regarding the projection, I think that mismatch will pretty much go away by using a consistent projection instead of lumped. Given the resolution on real problems, I suspect even the lumped projection is not bad.

examples/fluids/navierstokes.c
```diff
@@ -312,18 +329,27 @@ int main(int argc, char **argv) {
...
   CeedBasisDestroy(&ceed_data->basis_q_sur);
   CeedBasisDestroy(&ceed_data->basis_x_sur);
+  CeedBasisDestroy(&ceed_data->sp_stats.basis_q);
+  CeedBasisDestroy(&ceed_data->sp_stats.basis_x);
 
   // -- Restrictions
-  CeedElemRestrictionDestroy(&ceed_data->elem_restr_q);
-  CeedElemRestrictionDestroy(&ceed_data->elem_restr_x);
-  CeedElemRestrictionDestroy(&ceed_data->elem_restr_qd_i);
+  CeedElemRestrictionDestroy(&ceed_data->er_q);
```
jeremylt, line 345: Not to bikeshed too much, but this is the opposite of our usual convention to favor words in variable names instead of short initalizations that need decoder rings

jeremylt: I have been trying to push towards more explicit variable names because 1) its really cheap to buy extra characters to put in variable names now that we don't have punch cards and 2) its my hope that clearer names make it easier to onboard new students. 

I think some of our variable names are a bit too long, especially for the element restrictions, but I try to find a balance between brevity and clarity.

jrwrigh: This is fair. And I did start with keeping the `elem_restr_*` syntax, but then ran into things like `ceed_data->sp_stats.elem_restr_parent_colloc`, which is honestly just as difficult to read.

I initially limited this to just the new statistics one (for which I really did need `{parent,child}_colloc` to accurately describe what it was), but figured consistency was better.

My thought for `er_*` was just that we've already done `CeedQFunction -> qf`, `CeedOperator -> op`, so `CeedElemRestriction -> er` seems fairly natural. 

My other, more broad thought is that the variable name "real estate" shouldn't have a lot go to describing the variable type, leaving more for actually describing non-redundant information.

jeremylt: I've been trying to use local variables to get around some of this issue of a long name for a member of a struct which is a member of a struct. I believe the compiler optimizes out the stack allocation. 

I don't love putting the type in the variable either but I haven't come up with a good way around it yet. I usually do it because I have op_mass and qf_mass or elem_restr_u, basis_u, and u_loc/U_loc.

jedbrown: At some point, we'll remove these things and just have a list/hash of all known qfunctions, then from the input file, match those into `(label, value, qfunction)` for each desired operator, then create the composite operator by turning each of those list entries into a sub-operator. (And this mechanism can largely migrate into PETSc DM.) With this more generic mechanism, we'll be able to ask for a `MatShell` or other operator instead of creating it ourselves. Anyway, I'm not concerned about the length of the names in the struct, just that they follow the same pattern and are uniquely grep'able.

jeremylt: Related but not for this PR - I made a few passes through Ratel to reduce the amount of data carried around in these structs.  That removed a lot of unneeded code. A pass or two for that in the fluids code here might be a good idea. That would help with some of this chain of long names, as a nice added benefit.

examples/fluids/README.md
```diff
@@ -140,6 +140,22 @@ The following options are common among all problem types:
...
+  - `false`
+
+* - `-stats_collect_interval`
+  - Number of timesteps between statistics collection
+  - `1`
+
+* - `-stats_write_interval`
+  - Number of timesteps between statistics file writing (`-1` means only at end of run)
+  - `-1`
+
```
jedbrown, line 158: What do you think of an interface like
```yaml
ts_monitor_turbulence_zspan: cgns:stats.cgns
ts_monitor_turbulence_zspan_collect_interval: 3
ts_monitor_turbulence_zspan_write_interval: 300
```

examples/fluids/README.md
```diff
@@ -140,6 +140,22 @@ The following options are common among all problem types:
   - Use freestream boundary conditions on this list of faces
   -
 
+* - `-stats_enable`
+  - Enable collection of spanwise flow statistics
+  - `false`
+
+* - `-stats_test`
+  - Use manufactured solution for statistics collection (used for testing)
+  - `false`
```
jedbrown, line 149: I don't think this is used in a test? If it's going to exist purely for development (and stay in to test accuracy if this gets futzed with), it'd be better to remove it from docs and from `-help` (by using `PetscOptionsGetBool()`).

examples/fluids/qfunctions/newtonian_types.h
```diff
@@ -32,6 +32,7 @@ struct NewtonianIdealGasContext_ {
   CeedScalar        Ctau_M;
   CeedScalar        Ctau_E;
   CeedScalar        dt;
+  CeedScalar        t;
```
jedbrown, line 35: I'd much prefer long names, like `time`.

examples/fluids/src/matops.c
```diff
@@ -0,0 +1,125 @@
...
+    op_apply_ctx->X_loc = op_apply_ctx->Y_loc = NULL;
+  }
+
+  PetscFunctionReturn(0);
+}
+
+// -----------------------------------------------------------------------------
+// This function returns the computed diagonal of the operator
+// -----------------------------------------------------------------------------
+PetscErrorCode MatGetDiag(Mat A, Vec D) {
```
jedbrown, line 39: As written, this looks like a PETSc function. Note that we have `MatGetDiagonal_NS_IJacobian` already.

examples/fluids/problems/stats.c
```diff
@@ -0,0 +1,558 @@
...
+  PetscCall(PetscSectionSetComponentName(section, 0, 12, "Mean Momentum Z"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 13, "Mean Momentum Flux XX"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 14, "Mean Momentum Flux YY"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 15, "Mean Momentum Flux ZZ"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 16, "Mean Momentum Flux YZ"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 17, "Mean Momentum Flux XZ"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 18, "Mean Momentum Flux XY"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 19, "Mean Velocity X"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 20, "Mean Velocity Y"));
+  PetscCall(PetscSectionSetComponentName(section, 0, 21, "Mean Velocity Z"));
```
jedbrown, line 83: We could have an enum with entries for every variable and use that in the qfunction, as in `v[TURB_MEAN_MOMENTUM_FLUX_YZ][i] = wdetJ * s.U.momentum[1] * s.Y.velocity[2];` so that we don't have so many magic numbers. The last placeholder in the enum can replace the magic number `22` that's currently in a couple places.

jedbrown: I assume you've check that these visualize correctly in Paraview (at least the vectors; I think tensors aren't recognized by the PV/VTK reader for CGNS). Should we cut the spaces (with CamelCase) to match CGNS conventions?

examples/fluids/problems/stats.c
```diff
@@ -0,0 +1,558 @@
...
+  if (steps == user->app_ctx->cont_steps) PetscFunctionReturn(0);
+
+  if (steps % user->app_ctx->stats_collect_interval == 0) {
+    PetscCall(CollectStatistics(user, solution_time, Q));
+  }
+
+  if (steps % user->app_ctx->stats_write_interval == 0 && user->app_ctx->stats_write_interval != -1) {
+    PetscCall(DMGetGlobalVector(user->sp_stats.dm, &stats));
+    PetscCall(ProcessStatistics(user, &stats));
+    PetscCall(VecViewFromOptions(stats, NULL, "-stats_write_view"));
```
jedbrown, line 529: Better to hold the `PetscViewer` in the monitor so this can be a `VecView(stats, viewer)`.

jedbrown (commented):
Great work.

- [ ] I think we need a test of at least one field and an example yaml file. Perhaps Blasius with two elements thick.
- [ ] Misc naming and defensive programming
- [x] Persistent `PetscViewer` to write time series

examples/fluids/problems/stats.c
jrwrigh: Done

examples/fluids/src/matops.c
jrwrigh: So are you thinking to replace the `Mat*_NS_IJacobian` functions with this generic version of MatShell? Looks like they're effectively equivalent. And I think it makes sense to keep it generic.

examples/fluids/README.md
jrwrigh: Yeah, it's not currently used as a test. And it doesn't make for a good regression test either. I'll move it out from help and the docs.

examples/fluids/src/matops.c
jedbrown: I see no harm in merging so long as there are no custom semantics to deal with. But don't use a name that looks like a public PETSc function.

examples/fluids/problems/stats.c
jrwrigh: > We could have an enum with entries for every variable and use that in the qfunction, as in `v[TURB_MEAN_MOMENTUM_FLUX_YZ][i] = wdetJ * s.U.momentum[1] * s.Y.velocity[2];` so that we don't have so many magic numbers. The last placeholder in the enum can replace the magic number `22` that's currently in a couple places.

Yeah, that'd be a much nicer way to go about doing it. It think it'd also be nice for the statistics that are collected to be run-time selectable. That would require that the indexes be actual variables rather than enums. We can leave that for later, as it's not something we really need right now.

examples/fluids/README.md
jedbrown: Or just `ts_monitor_turbulence` since the "zspan" is inferred from the isoperiodic SF. But I like the "monitor" part because it's a thing that is monitoring the transient solution, not just viewing at the end.

examples/fluids/problems/stats.c
jrwrigh: > I assume you've check that these visualize correctly in Paraview (at least the vectors; I think tensors aren't recognized by the PV/VTK reader for CGNS). 

Yep, although it automatically interprets `Mean Momentum Flux X{X,Y,Z}` as a vector rather than leaving it alone.

> Should we cut the spaces (with CamelCase) to match CGNS conventions?

Ah, yep. Had it in my head that they had spaces rather than CamelCase.

examples/fluids/src/matops.c
jrwrigh: Gotcha. I just took it from `examples/petsc`.

jrwrigh: I guess `MatGetDiag_Ceed` would work better.

examples/fluids/README.md
jrwrigh: I've been a bit inconsistent with the flag/prefix naming between `spanstats` and `stats`. I think it makes sense to still reference spanwise since we may want to keep track of the global domain statistics (or at least the mean of the global domain solution). 

`ts_monitor` sounds fine to me. It'll pair well with the `ts_monitor_solution`. 

I'm not sure about `_turbulence` though. It seems both equivalent to "statistics" while being more constraining. We're not necessarily constrained to turbulence problems with the setup. In the distant future, if we implement a RANS model into the code, we still might want to leverage the statistics gathering, but it won't necessarily be turbulence statistics.

So how about:
```yaml
ts_monitor_spanstats: cgns:stats.cgns
ts_monitor_spanstats_collect_interval: 3
ts_monitor_spanstats_write_interval: 300
```

jrwrigh: I guess we could also merge `stats_enable` into `ts_monitor_spanstats`.

examples/fluids/src/matops.c
jrwrigh: I'll leave replacing the IJacobian functions with the generic equivalents for another PR. I've created an issue here #1142.

examples/fluids/README.md
jedbrown: Yeah, I don't think `stats_enable` should be needed. It isn't a separate option for `ts_monitor_solution`, for example.

Your suggestion is fine with me. I suggested `turbulence` because this may be different than stats about entrained particles in a low-Re flow, or about fabric in a granular flow. These option names can evolve, but doing so is a pain for users and causes more errors updating old runs. It's good to look ahead for names that are unambiguous rather than carrying implicit assumptions about the flow regime and thus context of super-generic terms like "stats".

examples/fluids/problems/stats.c
jedbrown: That's a bug in the CGNS reader that ships with VTK. There are a few improvements we should make sometime when our pants aren't on fire.

examples/fluids/README.md
jrwrigh: Gotcha. I was thinking in a single-phase continuum flow mindset. Yeah, I think `turbulence` makes sense in that respect.

So `ts_monitor_turbulence_spanstats_*` then?

jedbrown: Sounds good. That's very clear.

jrwrigh: Done now.

examples/fluids/problems/stats.c
jrwrigh: I've implemented the enum now.

> The last placeholder in the enum can replace the magic number `22` that's currently in a couple places.

So the last placeholder would be `21`, not `22`. Should I create a new const variable of `TURB_MEAN_VELOCITY_Z + 1`, so just do that when setting `user->spanstats.num_comp_stats`?

jrwrigh:
Note that I've changed `-test` to `-test_type`  (bool to enum) to facilitate regression testing of the statistics. Figured this change might also be useful later down the line as we add more features that we would want to do regression tests on, but may not be testable just by the solution output.

I'm not sure if this is the "best" way to do it, so I'm open to other suggestions.

examples/fluids/problems/stats.c
jedbrown: You can add `TURB_NUM_FIELDS` to the end of the enum and use that.

jrwrigh:
I've now pushed the statistics collection weighting into the QFunction itself.

I think the only "loose" thread hanging now is adding the ability to distribute the parent DM across ranks. Would I just need to run `DMPlexDistribute`?

jedbrown:
Yeah, `DMPlexDistribute` should work, and then destroy the old (poorly balanced) DM. It probably won't make much difference at small scale, but should help once there are many MPI ranks. It'll make use of a mesh partitioner (`simple` if configure hasn't added one). For the `zbox`, I think the `simple` partitioner will be acceptable.

Once I have CGNS loading on a different number of ranks, we'll need to decide if we need the output ordering (for instantaneous solution and stats) to match or if it can use the new distribution. Using the new distribution is easier and more efficient for IO.

jedbrown:
This should fix the directory permissions issue in CI. https://github.com/CEED/libCEED/pull/1143

jrwrigh:
Alright, should be all good to go now.

examples/fluids/qfunctions/turb_stats_types.h
```diff
@@ -0,0 +1,46 @@
...
+// SPDX-License-Identifier: BSD-2-Clause
+//
+// This file is part of CEED:  http://github.com/ceed
+
+#ifndef turb_stats_types_h
+#define turb_stats_types_h
+
+#include "./newtonian_types.h"
+
+enum {
```
jedbrown, line 13: Naming it will improve compiler diagnostics.
```suggestion
enum TurbComponent {
```

examples/fluids/src/misc.c
```diff
@@ -257,3 +258,29 @@ int FreeContextPetsc(void *data) {
   if (PetscFree(data)) return CeedError(NULL, CEED_ERROR_ACCESS, "PetscFree failed");
   return CEED_ERROR_SUCCESS;
 }
+
+// Return mass qfunction specification for number of components N
+PetscErrorCode ReturnMassQFunction(CeedInt N, ProblemQFunctionSpec *qf_spec) {
```
jedbrown, line 263: Could we make this return a `CeedQFunction`?
```suggestion
PetscErrorCode CreateMassQFunction(Ceed ceed, CeedInt N, CeedQFunction *qf) {
```

examples/fluids/src/setupts.c
```diff
@@ -144,8 +157,8 @@ PetscErrorCode IFunction_NS(TS ts, PetscReal t, Vec Q, Vec Q_dot, Vec G, void *u
   PetscCall(DMGetLocalVector(user->dm, &G_loc));
 
   // Update time dependent data
+  PetscCall(UpdateBoundaryValues(user, Q_loc, t));
   if (user->time != t) {
-    PetscCall(DMPlexInsertBoundaryValues(user->dm, PETSC_TRUE, Q_loc, t, NULL, NULL, NULL));
     if (user->phys->solution_time_label) {
       CeedOperatorContextSetDouble(user->op_ifunction, user->phys->solution_time_label, &t);
     }
```
jedbrown, line 164: Would it make sense to move this logic and below inside `UpdateBoundaryValues` (perhaps with a new name/args)?

jedbrown (approved):
Looks good to me. A few advisory comments to consider.

examples/fluids/src/misc.c
jrwrigh: Ah, yeah, that'd be cleaner.

examples/fluids/navierstokes.c
```diff
@@ -293,6 +299,9 @@ int main(int argc, char **argv) {
   // ---------------------------------------------------------------------------
   // Destroy libCEED objects
   // ---------------------------------------------------------------------------
+
+  if (app_ctx->turb_spanstats_enable) PetscCall(CleanupStats(user, ceed_data));
```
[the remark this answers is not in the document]
jrwrigh: Can the destroy commands be run on `NULL` values? I believe I ran into issues where it was trying to Destroy non-existent objects and would seg-fault and the end (hence why 7dd241f6 was titled a "fix").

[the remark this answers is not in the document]
jeremylt: I try to make all destroy commands exit early with a null argument. libCEED and PETSc destroy functions should all work that way, though that does require initalizing to NULL any objects that might not be created.

examples/fluids/navierstokes.c
jrwrigh: I'm not sure. I'm good with either:
1) Leaving it as it is (`er_q`, etc)
2) Making the new naming scheme more descriptive 
   - Could be`er_q` -> `er_q_vol` to contrast the `*_sur` used else where in the code
   - Could also change `q` to something else, like `state`. That'd require changing a lot of the other variables names throughout the code to be consistent.
3) Changing back to `elem_restr_*`

jrwrigh: > I don't love putting the type in the variable either but I haven't come up with a good way around it yet. I usually do it because I have op_mass and qf_mass or elem_restr_u, basis_u, and u_loc/U_loc.

I guess my point was less about whether I like the variable type prefix, and more about how much space to dedicate to it. I think having the mix of type-prefix, application-suffix makes for a really useful pattern for understanding the code. I think with `er_*` being used often in that pattern, it becomes straight-forward to remember, similar to `qf`, and `op`.

jrwrigh: Side note: What is the `*_i` supposed to mean for some of the qdata element restriction. (like `elem_restr_qd_i`)?

examples/fluids/navierstokes.h
```diff
@@ -102,21 +111,49 @@ struct AppCtx_private {
...
+  MatopApplyContext     M_ctx, mms_error_ctx;
+  KSP                   ksp;         // For the L^2 projection solve
+  CeedScalar            span_width;  // spanwise width of the child domain
+  PetscScalar           prev_time;
+  PetscBool             do_mms_test;
+  CeedContextFieldLabel solution_time_label, previous_time_label;
+} Span_Stats;
+
 // PETSc user data
 struct User_private {
```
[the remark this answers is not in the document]
jrwrigh: I'm honestly not sure what the logic for what member goes into which struct is. `CeedData` isn't available through the `User` context struct, so everything there isn't accessed directly, but needs to be destroyed at the end.

I believe every member in `Span_Stats` is required to be accessed as the simulation is running, but I'll double check that real quick. Actually, I know that I've make `user->spanstats.prev_time` obsolete, so I can definitely remove that.

As for the additions to `CeedData`, I may be able to remove the element restrictions, and possibly the qfunctions as well.

examples/fluids/navierstokes.c
jeremylt: The _i is actually a legacy convention from when these were called 'identity' restrictions. I've stripped this from Ratel but haven't made a pass over here to do so. We really should, but I haven't made the time. (sort of like how we should reduce the data carried around in the ceed_data struct in a follow up PR, probably need issues for both)

I prefer words in variable names - that's what's in the style guide. The 'elem_restr' prefix, or the op/qf ones don't perfectly match this convention but that's what's used in the rest of the repo. I'd keep the same usage as in the rest of the repo and revisit this when this code moves into another repo or make a separate PR to change the convention in this entire repo all at once. That way we don't mix together functional and purely style issues here.

examples/fluids/src/setupts.c
jrwrigh: Maybe, but I think the flexibility here is nice. For example the `CollectStatistics` function doesn't need to check whether the time has updated to it's label. 

I think what could be nice is having a dedicated function for updating the label? Something like:
```c
PetscErrorCode UpdateContextLabel(PetscScalar *previous_value, PetscScalar update_value, CeedOperator op, CeedContextFieldLabel label) {
  PetscFunctionBeginUser;

  if (*previous_value != update_value) {
    if (label) {
      CeedOperatorContextSetDouble(op, label, &update_value);
    }
    *previous_value = update_value;
  }
  PetscFunctionReturn(0);
}
```

Having a `CeedOperatorContextGetDouble` would be useful here, instead of having to manually track and update the label value.

jeremylt: Adding that (and friends) should be pretty straightforward and seems useful

jrwrigh: Made a new issue for getting label data: #1144

jedbrown:
Thanks, all. I'm happy for this to merge once the open threads are resolved and Jeremy approves. You should decide whether you intend to squash-merge or standard merge.

examples/fluids/src/setupts.c
```diff
@@ -21,23 +20,23 @@ PetscErrorCode ComputeLumpedMassMatrix(Ceed ceed, DM dm, CeedData ceed_data, Vec
...
+  CeedElemRestrictionCreateVector(ceed_data->er_q, &m_ceed, NULL);
+  CeedElemRestrictionCreateVector(ceed_data->er_q, &ones_vec, NULL);
   CeedVectorSetValue(ones_vec, 1.0);
 
   // CEED QFunction
-  CeedQFunctionCreateInterior(ceed, 1, Mass, Mass_loc, &qf_mass);
+  PetscCall(CreateMassQFunction(ceed, num_comp_q, &qf_mass));
   CeedQFunctionAddInput(qf_mass, "q", num_comp_q, CEED_EVAL_INTERP);
   CeedQFunctionAddInput(qf_mass, "qdata", q_data_size, CEED_EVAL_NONE);
   CeedQFunctionAddOutput(qf_mass, "v", num_comp_q, CEED_EVAL_INTERP);
```
jedbrown, line 33: I think setting the inputs and output can be folded into `CreateMassQFunction`, which would make the calls simpler.

examples/fluids/navierstokes.h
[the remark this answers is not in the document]
jrwrigh: What I ended up doing was taking the `ceed_data->spanstats` struct and moving it to a local struct that is created and destroyed during the setup procedure. I think that ends up leading to less fragile code (less worry about whether `x_coord` or `elem_restr_parent_colloc` is created at the correct time) while still leaving it as non-persistent storage. Open to suggestions/comments as always.

examples/fluids/src/setupts.c
jrwrigh: Done!

examples/fluids/src/setupts.c
jrwrigh: Added the `UpdateContextLabel` function

examples/fluids/navierstokes.c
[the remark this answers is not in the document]
jrwrigh: Fixed. The issue I had before was with trying to destroy context members of a context that was never allocated (so pointing from a `NULL` pointer). Solved by actually implementing a destroy command for the context and passing `NULL` through there (borrowing heavily from [Ratel's version of matops](https://gitlab.com/micromorph/ratel/-/blob/main/src/internal/ratel-petsc-ops.c)).

examples/fluids/navierstokes.c
```diff
@@ -205,7 +211,7 @@ int main(int argc, char **argv) {
   // ---------------------------------------------------------------------------
   // Print problem summary
   // ---------------------------------------------------------------------------
-  if (!app_ctx->test_mode) {
+  if (!app_ctx->test_type) {
```
jeremylt, line 214: minor - in other places you have
```suggestion
  if (app_ctx->test_type != TESTTYPE_NONE) {
```

jeremylt (approved):
Looks good to me. Does this close #1125?

jrwrigh:
Nah, it doesn't close #1125. That's for total domain statistics rather than spanwise averaged. Though that's probably going to be a dormant PR/branch for awhile as it's not really needed. Original idea was to use it as a stepping stone before I realized Jed's idea for spanwise didn't actually need domain statistics before projecting to spanwise.

As for merge vs squash-merge, the current commits are more-or-less "clean", but there are a few undo-then-redos in the history (in addition to it being 40+ commits). I'm good with either way, not sure what other's opinions are.

examples/fluids/navierstokes.c
jrwrigh: Good catch!

PR #1138, opened 2023-01-10
Ken/idl

KennethEJansen:
IDL stands for internal damping layer.  Basically, it provides a sponge-like body force to drive the pressure fluctuations towards the reference pressure where the scaling coefficient is greater than zero.  Typically, the scaling coefficient is O(1/dt) at the inflow where the STG boundary condition has been observed to create pressure fluctuations that resonate and eventually globally pollute the solution without this term. The scaling coefficient can be ramped to zero over a stream wise distance of about 2 boundary layer thicknesses (typically known from the STG inflow definition). 

This body force will be applied whenever `IDLramp_amplitude` is greater than zero as this is the value applied for x <= `IDLramp_start` (typically the x coordinate of the STG inflow plane).  The third parameter, `IDLramp_length` is the streamwise distanance of the ramp down to zero.  Beyond, the ramp, the scaling coefficient is zero.


examples/fluids/blasius.yaml
```diff
@@ -10,6 +10,11 @@ ts:
   #monitor_solution_interval: 10
 checkpoint_interval: 10
 
+ramp_amplitude: 1.0
+ramp_start: -3.1
+ramp_length: 0.2
```
jedbrown: I think these need the new option names.

examples/fluids/qfunctions/newtonian.h
```diff
@@ -249,6 +250,16 @@ CEED_QFUNCTION_HELPER int IFunction_Newtonian(void *ctx, CeedInt Q, const CeedSc
...
+    if (context->idl_enable) {
+      const CeedScalar sigma     = LinearRampCoefficient(1 / context->idl_decay_time, context->idl_length, context->idl_start, x_i[0]);
+      const CeedScalar damp_Y[5] = {sigma * (s.Y.pressure - P0), 0, 0, 0, 0};
+      CeedScalar       dx_i[3]   = {0};
+      State            damp_s    = StateFromY_fwd(context, s, damp_Y, x_i, dx_i);
+      v[0][i] += wdetJ * damp_s.U.density;
+      for (int j = 0; j < 3; j++) v[j + 1][i] += wdetJ * damp_s.U.momentum[j];
+      v[4][i] += wdetJ * damp_s.U.E_total;
+    }
+
```
jrwrigh, line 262: Thoughts on moving this (and it's jacobian) into a separate `CeedOperator`? The obvious concern is with performance, but I'm not sure if there's any rules-of-thumb that tamp down those concerns. 

There aren't any intermediate terms used from the `IFunction_Newtonian` other than state `s`. Possibly we could have the separate IDL CeedOperator use the `jac_data` to accelerate that, though I'm not sure if Ceed guarantees order of execution for sub-operators.

examples/fluids/index.md
```diff
@@ -634,19 +634,26 @@ The `STGRand.dat` file is the table of the random number set, $\{\bm{\sigma}^n,
...
+| $\{\kappa^n\}_{n=1}^N$                         | k^n      | No             | Yes       |
+| $h_i$                                          | h_i      | Yes            | No        |
+| $d_w$                                          | d_w      | Yes            | No        |
+
+#### Internal Damping Layer (IDL)
+STG creates spurious acoustic waves during it's development to natural
+turbulent structures. To damp those out, we use an internal damping layer (IDL)
+to prevent them from growing. This implementation was inspired from
+{cite}`shurSTG2014`, but is implemented here as a ramped volumetric forcing
+term in the residual itself.
```
jrwrigh, line 656: I would say more here, but I don't fully grasp the analytical form of the damping (particularly the use of `StateFromY_fwd` instead of just `StateFromY` for `damp_s`).

examples/fluids/problems/newtonian.c
```diff
@@ -202,6 +204,11 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx, SimpleBC
   if (state_var == STATEVAR_PRIMITIVE && !implicit) {
     SETERRQ(comm, PETSC_ERR_ARG_NULL, "RHSFunction is not provided for primitive variables (use -state_var primitive only with -implicit)\n");
   }
+
+  PetscCall(PetscOptionsScalar("-idl_decay_time", "Effectively inverse of strength of IDL. The timestep is good starting point", NULL, idl_decay_time,
```
jrwrigh, line 208: Similarly here, "Effectively inverse of strength of IDL" should be replaced by something more informative/intuitive.

examples/fluids/problems/newtonian.c
```diff
@@ -202,6 +204,11 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx, SimpleBC
...
+                               &idl_decay_time, &idl_enable));
```
jedbrown, line 209: We should explain that `0` is treated like $\infty$. Or perhaps we should use `-1` to mean this is off.

examples/fluids/qfunctions/newtonian.h
```diff
@@ -249,6 +250,16 @@ CEED_QFUNCTION_HELPER int IFunction_Newtonian(void *ctx, CeedInt Q, const CeedSc
     UnpackState_U(s_dot.U, U_dot);
 
     for (CeedInt j = 0; j < 5; j++) v[j][i] = wdetJ * (U_dot[j] - body_force[j]);
+    if (context->idl_enable) {
+      const CeedScalar sigma     = LinearRampCoefficient(1 / context->idl_decay_time, context->idl_length, context->idl_start, x_i[0]);
```
jedbrown, line 254: We probably want to invert as preprocessing so we can handle `-1` as disable. That is, `-idl_decay_time 0` should not cause a divide by zero exception.

examples/fluids/qfunctions/newtonian.h
```diff
@@ -348,6 +359,16 @@ CEED_QFUNCTION_HELPER int IJacobian_Newtonian(void *ctx, CeedInt Q, const CeedSc
...
 
+    if (context->idl_enable) {
+      const CeedScalar sigma      = LinearRampCoefficient(1 / context->idl_decay_time, context->idl_length, context->idl_start, x_i[0]);
+      const CeedScalar damp_dY[5] = {sigma * ds.Y.pressure, 0, 0, 0, 0};
+      CeedScalar       dx_i[3]    = {0};
+      State            damp_ds    = StateFromY_fwd(context, s, damp_dY, x_i, dx_i);
+      v[0][i] += wdetJ * damp_ds.U.density;
+      for (int j = 0; j < 3; j++) v[j + 1][i] += wdetJ * damp_ds.U.momentum[j];
+      v[4][i] += wdetJ * damp_ds.U.E_total;
+    }
```
jedbrown, line 370: Should we move this into a helper since it's near-identical to above (and we'll perhaps need it when we have RANS residuals so best to keep qfunctions slim).

examples/fluids/index.md
jedbrown, line 656: Our formulation is based on Eq 8.32 of this chapter in the book John/Ken/I contributed to. https://doi.org/10.1016/B978-0-32-391144-3.00014-0
It has citations to prior papers, though I didn't read those. We should explain that the damping is toward `reference_pressure` in primitive variables (so no damping of velocity or temperature) linearized around the current state.

jrwrigh (commented):
Summary of updates:
- Renamed Newtonian wave -> Gaussian wave problem
- Extracted out the fluid binary reading code for use in regression tests
- Moved IDL flags to `idl_*` header
- Changed amplitude parameter to decay rate (so the inverse of the amplitude)
- Moved `LinearRampCoefficient` into `utils.h`
- Added IDL test
- Documented IDL (though see comments with questions regarding that)

jedbrown (approved):

examples/fluids/qfunctions/newtonian.h
jedbrown: I'd rather have fewer operators. I don't think this is expensive, though we could do static specialization if we were concerned about the `if` statement.

examples/fluids/problems/newtonian.c
jedbrown: Let $p = p_{ref} + p'$. The deviation will decay like $p'(t) = e^{-t/\tau} p'(0)$ where $\tau$ is the characteristic decay time. It's similar to a "half life" of the pressure anomalies.

examples/fluids/blasius.yaml
jrwrigh: Looks like you started the review before I caught this. I've actually just removed this from `blasius.yaml` all together, as I haven't tested it with the blasius anyways. That work with you?

examples/fluids/index.md
jedbrown: Similar to below, we use
```math
-\sigma(\symbf x) \underbrace{A_0 Y'}_{U'}
```
where $Y' = \\{p - p_{ref}, \symbf 0, 0\\}$.
image

examples/fluids/blasius.yaml
jedbrown: Yup!

knepley:
@jedbrown Can you send me Chapter 13 by John Bell from that volume? UB does not get it.

examples/fluids/problems/newtonian.c
jrwrigh: I think `-1` to mean off (or really any negative number) makes sense. I think we should treat `0` as an invalid input, right?

examples/fluids/index.md
jrwrigh: Should we perhaps denote $\symbf{Y}'$ as $\mathrm{d}\symbf{Y}$ or $\Delta \symbf{Y}$?

Also, I don't think we're exactly doing $\symbf{A}_0 \symbf{Y}$ either, as that would be equivalent to `StateFromY(gas, damp_y, x_i)`. Instead we're doing `StateFromY_fwd(context, s, damp_Y, x_i, dx_i)`, which I don't believe has a representation in our normal notation. Since we're only changing pressure, the two operations (`StateFromY` and `StateFromY_fwd`) are equivalent, but I think the intention here is important (and I'm not sure what the intention is).

examples/fluids/problems/newtonian.c
jedbrown: Sure

examples/fluids/index.md
jedbrown: $A_0$ is $\partial U/\partial Y$ linearized at the current state so I think that's what we're doing. I felt like it was simpler (closer to linear), but probably either would work. I don't have a sufficiently discerning test case.

jrwrigh: My assertion that we're not doing $A_0 \symbf{Y}$ comes from the fact that $A_0$ is used in the literature for $Y_{,t}$, but I see now that we do use `StateFromQi_fwd` for the $Y_{,t}$ term. So I think you're right that $A_0 Y$ is equivalent to `StateFromQi_fwd`.

jrwrigh: Ok, I think I see my confusion now. `StateFromQi` computes $U(Y)$, while `StateFromQi_fwd` computes $\partial U(Y) / \partial Y$. 

For completeness sake, when I'm talking about `StateFrom*`, I'm really talking about `StateConservativeFromPrimitive*`, which is used inside `StateFromY*`.

examples/fluids/qfunctions/newtonian.h
jrwrigh: I think I've done what you suggested, but let me know if I misunderstood.

examples/fluids/qfunctions/newtonian.h
```diff
@@ -348,6 +367,12 @@ CEED_QFUNCTION_HELPER int IJacobian_Newtonian(void *ctx, CeedInt Q, const CeedSc
     UnpackState_U(ds.U, dU);
     for (int j = 0; j < 5; j++) v[j][i] = wdetJ * (context->ijacobian_time_shift * dU[j] - dbody_force[j]);
 
+    if (context->idl_enable) {
+      CeedScalar damp_state[5] = {ds.Y.pressure, 0, 0, 0, 0}, idl_residual[5] = {0.};
+      InternalDampingLayer(context, s, x_i, damp_state, idl_residual);
```
jedbrown, line 372:
```suggestion
      // This is a Picard-type linearization of the damping and could be replaced by an InternalDampingLayer_fwd that uses s and ds.
      InternalDampingLayer(context, s, x_i, damp_state, idl_residual);
```

examples/fluids/index.md
jedbrown: Will you add a couple sentences here?

jrwrigh: Yeah, I started on that, but didn't finish it up.

jrwrigh: Let me know if the current explanation make sense, then I'll go ahead and merge.

examples/fluids/index.md
```diff
@@ -634,19 +634,37 @@ The `STGRand.dat` file is the table of the random number set, $\{\bm{\sigma}^n,
...
+$$
+S(\bm{q}) = -\sigma(\bm{x})\left.\frac{\partial \bm{q}}{\partial \bm{Y}}\right\rvert_{\bm{q}} \bm{Y}'
+$$
+
+where $\bm{Y}' = [P - P_\mathrm{ref}, \bm{0}, 0]^T$, and $\sigma(\bm{x})$ is a
+linear ramp starting at `-idl_start` with length `-idl_length` and an amplitude
+of inverse `-idl_decay_rate`. The damping is defined in pressure-primitive
+variables, but is converted to conservative variables using $\partial
+\bm{q}/\partial \bm{Y}\rvert_{\bm{q}}$, which is linearized about the current
+flow state. $P_\mathrm{ref}$ is defined via the `-refernce_pressure` flag.
```
jedbrown, line 667: Here's an adjustment. I think Shur's IDL is quite different so it at least deserves a citation to the book chapter or one of the mentioned papers.
```suggestion
The STG inflow boundary condition creates large amplitude acoustic waves.
We use an internal damping layer (IDL) to damp them out without disrupting the synthetic structures developing into natural turbulent structures. This implementation was inspired from
{cite}`shurSTG2014`, but is implemented here as a ramped volumetric forcing
term in the residual itself. It takes the following form:

$$
S(\bm{q}) = -\sigma(\bm{x})\left.\frac{\partial \bm{q}}{\partial \bm{Y}}\right\rvert_{\bm{q}} \bm{Y}'
$$

where $\bm{Y}' = [P - P_\mathrm{ref}, \bm{0}, 0]^T$, and $\sigma(\bm{x})$ is a
linear ramp starting at `-idl_start` with length `-idl_length` and an amplitude
of inverse `-idl_decay_rate`. The damping is defined in terms of a pressure-primitive
anomaly $\bm Y'$ converted to conservative source using $\partial
\bm{q}/\partial \bm{Y}\rvert_{\bm{q}}$, which is linearized about the current
flow state. $P_\mathrm{ref}$ is defined via the `-reference_pressure` flag.
```

jrwrigh: Done.

PR #1141, opened 2023-01-12
Icl/nt rtc

abdelfattah83:
This work further improves the performance of the non-tensor MAGMA backend for relatively low-order problems. The backend now has new optimized kernels for the non-tensor interp and grad basis actions. Sample results are attached. 
[summary.pdf](https://github.com/CEED/libCEED/files/10404983/summary.pdf)


abdelfattah83:
One specific change outside the MAGMA backend is here f7cc693. This enables the CeedRunKernelDimSharedCuda function to configure CUDA kernels with more than 48KB of dynamic shared memory.

jeremylt (commented):
I'll do a more in depth look later today, but this looks good at a glance and those performance improvements are really great!


I found a couple of formatting changes in other files that I'd rather omit

examples/petsc/bpsraw.c
```diff
@@ -64,7 +64,7 @@ static void Split3(PetscInt size, PetscInt m[3], bool reverse) {
 static PetscInt Max3(const PetscInt a[3]) { return PetscMax(a[0], PetscMax(a[1], a[2])); }
 static PetscInt Min3(const PetscInt a[3]) { return PetscMin(a[0], PetscMin(a[1], a[2])); }
 static void     GlobalNodes(const PetscInt p[3], const PetscInt i_rank[3], PetscInt degree, const PetscInt mesh_elem[3], PetscInt m_nodes[3]) {
-      for (int d = 0; d < 3; d++) m_nodes[d] = degree * mesh_elem[d] + (i_rank[d] == p[d] - 1);
```
jeremylt, line 67: Our formatting test doesn't like this change

interface/ceed.c
```diff
@@ -27,8 +27,10 @@ static struct {
 } backends[32];
 static size_t num_backends;
 
-#define CEED_FTABLE_ENTRY(class, method) \
-  { #class #method, offsetof(struct class##_private, method) }
+#define CEED_FTABLE_ENTRY(class, method)                     \
+  {                                                          \
+#class #method, offsetof(struct class##_private, method) \
+  }
```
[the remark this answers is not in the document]
abdelfattah83: should be fixed in 4c54a76

include/ceed/khash.h
```diff
@@ -178,11 +178,11 @@ static const double __ac_HASH_UPPER = 0.77;
...
-                kfree(h->flags);                                                                                                              \
-                kfree((void *)h->vals);                                                                                                       \
-                kfree(h);                                                                                                                     \
+    if (h) {                                                                                                                        \
+      kfree((void *)h->keys);                                                                                                       \
+      kfree(h->flags);                                                                                                              \
+      kfree((void *)h->vals);                                                                                                       \
+      kfree(h);                                                                                                                     \
```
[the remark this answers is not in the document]
abdelfattah83: Should be fixed in 4c54a76

examples/petsc/bpsraw.c
abdelfattah83: Should be fixed in 4c54a76

jedbrown:
Wow! :rocket: We were just talking about this in the context of the fluids solver. Even for "simple" geometries, tet or mixed topology meshes greatly reduce the dof count relative to hexes. This PR can be reviewed for correctness/maintainability and merged, but this unlocks an exciting accuracy-efficiency study we'd like to do for turbulence simulation.

Makefile
```diff
@@ -455,23 +455,25 @@ ifneq ($(wildcard $(MAGMA_DIR)/lib/libmagma.*),)
     endif
   else  # HIP MAGMA
     ifneq ($(HIP_LIB_DIR),)
-      hip_link = $(if $(STATIC),,-Wl,-rpath,$(HIP_LIB_DIR)) -L$(HIP_LIB_DIR) -lhipblas -lhipsparse -lamdhip64
+      HIP_LIB_DIR_MAGMA=$(HIP_DIR)/../lib
+      HIP_INC_DIR_MAGMA=$(HIP_DIR)/../include
```
nbeams, line 459: I'm confused why this change is needed.  What were you using as `HIP_DIR`?  I generally use `/opt/rocm` for `HIP_DIR` -- perhaps I should have named it `ROCM_DIR` -- but with my usual build configuration, I get `HIP_LIB_DIR_MAGMA` as `/opt/rocm/../lib` which is not what I want.

jeremylt (approved):
This looks good to me and CI is happy. I think @nbeams makefile question should be adjudicated though

Makefile
abdelfattah83: My `HIP_DIR` points to the "hip dir" :), which is `/opt/rocm-xx/hip`. I got confused that it is not named `ROCM_DIR` or `ROCM_PATH` (the latter is declared by default when load a rocm module on Crusher).

nbeams: Should we change the name?  How many other people's build systems could that interfere with?

I was wondering how you could get all the backends to build at the same time (the "regular" HIP backends and MAGMA), but I see at least some ROCm installs have the basic libraries we need in `/opt/rocm/hip/lib` as well as `/opt/rocm/lib`.  So the current setup will work to build all the backends with `HIP_DIR` as `/opt/rocm/hip`, but only the "standard" HIP backends with `HIP_DIR = /opt/rocm`.

Are there any strong opinions about this?  In the README, we do say "To enable HIP support, add `HIP_DIR=/opt/rocm` or an appropriate directory.", but that could be changed.

jedbrown: I think we should change the makefile to use `ROCM_DIR` to the extent we need that path (e.g., as default for `hipcc`). We're going to need to revisit the HIP build soon to remove AMD-specific assumptions because the CHIP-SPV project is almost usable by libCEED, and will open one (of two) paths to Intel GPUs. The details block below is how one developer hacked the makefile to build with CHIP-SPV, in case that's useful when updating it.

<details>

```diff
diff --git a/Makefile b/Makefile
index 54c99769..7d34364c 100644
--- a/Makefile
+++ b/Makefile
@@ -397,16 +397,21 @@ ifneq ($(CUDA_LIB_DIR),)
 endif

 # HIP Backends
-HIP_LIB_DIR := $(wildcard $(foreach d,lib
lib64,$(HIP_DIR)/$d/libamdhip64.${SO_EXT}))
-HIP_LIB_DIR := $(patsubst %/,%,$(dir $(firstword $(HIP_LIB_DIR))))
+# Probably should extend to look for libCHIP.so?
+# HIP_LIB_DIR := $(wildcard $(foreach d,lib
lib64,$(HIP_DIR)/$d/libamdhip64.${SO_EXT}))
+HIP_LIB_DIR :=
/gpfs/jlse-fs0/users/pvelesko/install/HIP/clang15/chip-spv-develop-debug/lib
 HIP_BACKENDS = /gpu/hip/ref /gpu/hip/shared /gpu/hip/gen
 ifneq ($(HIP_LIB_DIR),)
   $(libceeds) : HIPCCFLAGS += -I./include
   ifneq ($(CXX), $(HIPCC))
-    CPPFLAGS += $(subst =,,$(shell $(HIP_DIR)/bin/hipconfig -C))
+    # removing the equal signs breaks options such as
--target=x86_64-linux-gnu
+    # CPPFLAGS += $(subst =,,$(shell $(HIP_DIR)/bin/hipconfig -C))
+    CPPFLAGS += $(shell $(HIP_DIR)/bin/hipconfig -C)
   endif
   $(libceeds) : CPPFLAGS += -I$(HIP_DIR)/include
-  PKG_LIBS += -L$(abspath $(HIP_LIB_DIR)) -lamdhip64 -lhipblas
+  # extracted from hipconfig
+  PKG_LIBS +=
-L/home/pvelesko/space/install/HIP/clang15/chip-spv-testing/lib -lCHIP
-L/soft/libraries/khronos/loader/master-2022.05.18/lib64 -lOpenCL
-L/soft/restricted/CNDA/emb/intel-gpu-umd/20221031.1-pvc-prq-66/driver/lib64
-lze_loader
-Wl,-rpath,/home/pvelesko/space/install/HIP/clang15/chip-spv-testing/lib
+#   PKG_LIBS += -L$(abspath $(HIP_LIB_DIR)) -lamdhip64 -lhipblas
   LIBCEED_CONTAINS_CXX = 1
   libceed.c     += interface/ceed-hip.c
   libceed.c     += $(hip.c) $(hip-ref.c) $(hip-shared.c) $(hip-gen.c)
@@ -490,7 +495,7 @@ $(libceed.a) : $(call weak_last,$(libceed.o)) |
$$(@D)/.DIR
        $(call quiet,AR) $(ARFLAGS) $@ $^

 $(OBJDIR)/%.o : $(CURDIR)/%.c | $$(@D)/.DIR
-       $(call quiet,CC) $(CPPFLAGS) $(CFLAGS) -c -o $@ $(abspath $<)
+       $(call quiet,CC) $(CFLAGS) -c -o $@ $(abspath $<)

 $(OBJDIR)/%.o : $(CURDIR)/%.cpp | $$(@D)/.DIR
        $(call quiet,CXX) $(CPPFLAGS) $(CXXFLAGS) -c -o $@ $(abspath $<)
@@ -687,7 +692,7 @@ style : style-c style-py
 CLANG_TIDY ?= clang-tidy

 %.c.tidy : %.c
-       $(CLANG_TIDY) $(TIDY_OPTS) $^ -- $(CPPFLAGS) --std=c99
-I$(CUDA_DIR)/include -I$(HIP_DIR)/include
-DCEED_JIT_SOUCE_ROOT_DEFAULT="\"$(abspath ./include)/\""
+       $(CLANG_TIDY) $(TIDY_OPTS) $^ -- --std=c99 -I$(CUDA_DIR)/include
-I$(HIP_DIR)/include -DCEED_JIT_SOUCE_ROOT_DEFAULT="\"$(abspath
./include)/\""

 %.cpp.tidy : %.cpp
        $(CLANG_TIDY) $(TIDY_OPTS) $^ -- $(CPPFLAGS) --std=c++11
-I$(CUDA_DIR)/include -I$(OCCA_DIR)/include -I$(HIP_DIR)/include
```

</details>

abdelfattah83: Please check out 5e5b0c9.

Makefile
```diff
@@ -30,7 +30,7 @@ ifeq (,$(filter-out undefined default,$(origin ARFLAGS)))
 endif
 NVCC ?= $(CUDA_DIR)/bin/nvcc
 NVCC_CXX ?= $(CXX)
-HIPCC ?= $(HIP_DIR)/bin/hipcc
+HIPCC ?= $(ROCM_DIR)/bin/hipcc
```
jeremylt, line 33: This will need to be updated in the README as well

jeremylt: and here: https://github.com/CEED/libCEED/blob/main/.gitlab-ci.yml

abdelfattah83: I've just a quick find/replace in 493cdfe

jedbrown: Any thoughts on how to handle non-ROCm HIP? There is `/opt/rocm/hip/bin/hipcc` in addition to `/opt/rocm/bin/hipcc` on Noether. I'm not sure if `HIP_DIR = $(ROCM_DIR)` or `HIP_DIR = $(ROCM_DIR)/hip` should be used, and then we should use `HIP_DIR` for things that are not specific to ROCm?

abdelfattah83: I am not familiar with Noether, but I found this on Crusher:
For rocm-5.1.0: `hipcc` is under `/opt/rocm-5.1.0/hip/bin`. There is a symlink under `/opt/rocm-5.1.0/bin`, which points to `../hip/bin/hipcc`.

For rocm-5.2.0, 5.3.0, and 5.4.0, it is the opposite scenario, `hipcc` is under `/opt/rocm-5.x.0/bin` and the symlinks are under `/opt/rocm-5.x.0/hip/bin`. I think staying with `ROCM_DIR` is less ambiguous than `HIP_DIR`.

I am not sure what you mean by non-ROCM HIP. Do you mean using HIP to run on NVIDIA for example?

jedbrown: Yeah, HIP to run on Intel GPUs.

abdelfattah83: IMHO, I think using `ROCM_DIR` is fairly more accurate than `HIP_DIR` when targeting AMD GPUs. If there is a plan to run on Intel GPUs using HIP, I think we need a distinctive name like `HIPSYCL_DIR` or something similar. I'm not sure if this has to be resolved in this PR though.

jedbrown: Maybe we can use `ROCM_DIR` to select `HIPCC` and `HIP_DIR`/`HIP_LIB`? I'm not sure if there will be (or should be) a way to build the HIP backend for ROCm and build it for SPV, in which case there would be `/gpu/rocm` and `/gpu/spv` (or `/gpu/intel`, whatever they want to call it). But that's all getting messy/out of scope, so let's just try to make `ROCM_DIR` provide defaults and use the `HIP_*` variables in a vendor-agnostic way.

jedbrown (approved):
I think this is good enough with respect to `ROCM_DIR` handling. I need to revisit this for CHIP-SPV #1146 , but no need to hold up this PR over that.

sebastiangrimberg:
Hi @abdelfattah83, sorry to bring up an old thread here but I was wondering if you had access to scripts or code to perform the automated benchmarking used to generate the offline tuning data for the MAGMA GEMM and RTC code? In the particular case of the RTC code, some of these kernels will change for H(div) and H(curl) support, leading to perhaps new optimal block sizes. Including such a script in the library might be a good idea, especially for new hardware in the future.

abdelfattah83:
@sebastiangrimberg I agree. Unfortunately, the original kernels were developed using ahead-of-time compilation using nvcc/hipcc outside libCEED, and so the scripts we used back then for tuning would not work. With RTC, we need (1) a testing code inside libCEED to run these kernels for specific (P, Q) and blocking size, and (2) a script that would post-process the data generated by tester in order to generate the tuning decisions. It is also possible to combine (1) and (2) into one unified tester.

sebastiangrimberg:
Thanks a lot for the response @abdelfattah83. I guess given no code to start from this won't be so bad. I'm not sure about the non-RTC kernels, though (but maybe those are in less need of up to date tuning? Maybe that tuning comes directly from MAGMA, correct me if I'm wrong).

I'll think about how to put together such a synthetic benchmark for the RTC kernels which constructs a basis for a given P Q, N, and tests out a bunch of NB values, reporting the best. I think this is important to have for the future of the library and `magma` backend. Ideally reusing as much of the libCEED framework to do this as possible would be great, but we'll have to bypass the internal RTC NB selection, obviously. Maybe it can be an external program which builds using some subset of the libCEED source code.

sebastiangrimberg:
See https://github.com/CEED/libCEED/pull/1366!

PR #1143, opened 2023-01-20
gitlab-ci: use safe.directory to describe system directories

jedbrown:
This is a side-effect-free way to describe a repository that may be owned by a different user.

https://github.blog/2022-04-18-highlights-from-git-2-36/#stricter-repository-ownership-checks


jeremylt (commented):
I like this better than what I did in Ratel - I'll adjust over there

PR #1145, opened 2023-01-23
Fix -Wextra-semi -Wmissing-field-initializers

prj-:


jedbrown (approved):
Thanks!

prj-:
No problem. By the way, are you at liberty to share, here on in private, a more production-oriented test case than `build/fluids-navierstokes -options_file examples/fluids/blasius.yaml -{ts,snes}_monitor -ksp_converged_reason` for the application you mentioned in (PETSc) Slack? If we take GPUs out of the equation (I know it is a big if), on my laptop, nothing seems to beat a simple `PCASM`, but I am guessing this won't be the case at scale. I would like to try some alternatives.

jedbrown:
Sure, I'll follow up with you on PETSc Slack to reduce noise over here (but others are ping if interested).

PR #1148, opened 2023-01-24
OperatorContextGet[Double, Int]

### LINKED ISSUE #1144: Add `CeedOperatorContextGetDouble` and friends for label data access
A few questions for `CeedOperatorContextGetDouble` would be:
1) Is getting data access to the label expensive? 
   - ie. should we add a member to `CeedContextFieldLabel` that keeps track of the value to avoid trying to get access to the context (thinking GPU to CPU communication might be at play here)
2) For sublabels, do we trust that all the contexts will be the same value and, if not, what value do we return?

jeremylt:
- [x] Add Context[Get,Restore][Double,Int32]Read
- [x] Add tests

Closes #1144

### DIFF
--- a/include/ceed/ceed.h
+++ b/include/ceed/ceed.h
@@ -509,7 +509,11 @@ CEED_EXTERN int CeedOperatorGetNumQuadraturePoints(CeedOperator op, CeedInt *num
 CEED_EXTERN int CeedOperatorGetFlopsEstimate(CeedOperator op, CeedSize *flops);
 CEED_EXTERN int CeedOperatorContextGetFieldLabel(CeedOperator op, const char *field_name, CeedContextFieldLabel *field_label);
 CEED_EXTERN int CeedOperatorContextSetDouble(CeedOperator op, CeedContextFieldLabel field_label, double *values);
+CEED_EXTERN int CeedOperatorContextGetDoubleRead(CeedOperator op, CeedContextFieldLabel field_label, size_t *num_values, const double **values);
+CEED_EXTERN int CeedOperatorContextRestoreDoubleRead(CeedOperator op, CeedContextFieldLabel field_label, const double **values);
 CEED_EXTERN int CeedOperatorContextSetInt32(CeedOperator op, CeedContextFieldLabel field_label, int *values);
+CEED_EXTERN int CeedOperatorContextGetInt32Read(CeedOperator op, CeedContextFieldLabel field_label, size_t *num_values, const int **values);
+CEED_EXTERN int CeedOperatorContextRestoreInt32Read(CeedOperator op, CeedContextFieldLabel field_label, const int **values);
 CEED_EXTERN int CeedOperatorApply(CeedOperator op, CeedVector in, CeedVector out, CeedRequest *request);
 CEED_EXTERN int CeedOperatorApplyAdd(CeedOperator op, CeedVector in, CeedVector out, CeedRequest *request);
 CEED_EXTERN int CeedOperatorDestroy(CeedOperator *op);
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -270,7 +270,7 @@ int CeedOperatorGetActiveElemRestriction(CeedOperator op, CeedElemRestriction *a
 }
 
 /**
-  @brief Set QFunctionContext field value of the specified type.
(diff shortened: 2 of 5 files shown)


include/ceed/ceed.h
```diff
@@ -508,6 +508,8 @@ CEED_EXTERN int CeedOperatorGetNumElements(CeedOperator op, CeedInt *num_elem);
 CEED_EXTERN int CeedOperatorGetNumQuadraturePoints(CeedOperator op, CeedInt *num_qpts);
 CEED_EXTERN int CeedOperatorGetFlopsEstimate(CeedOperator op, CeedSize *flops);
 CEED_EXTERN int CeedOperatorContextGetFieldLabel(CeedOperator op, const char *field_name, CeedContextFieldLabel *field_label);
+CEED_EXTERN int CeedOperatorContextGetDouble(CeedOperator op, CeedContextFieldLabel field_label, size_t *num_values, const double **values);
+CEED_EXTERN int CeedOperatorContextGetInt32(CeedOperator op, CeedContextFieldLabel field_label, size_t *num_values, const int **values);
```
jrwrigh, line 512: Any reason to return `size_t *num_values`? The user would already need to ensure that `*values` is of the correct size, so they probably already know `num_values` (or could find out from the label itself).

jeremylt: I nearly always send the array length with the array so that each function call can stand alone

jrwrigh: Ah, didn't see that `values` was a double pointer. So my latter point is invalid (user doesn't have to know the length of the array before the call).

examples/fluids/src/turb_spanstats.c
```diff
@@ -25,7 +25,7 @@ typedef struct {
   CeedElemRestriction elem_restr_parent_x, elem_restr_parent_stats, elem_restr_parent_qd, elem_restr_parent_colloc, elem_restr_child_colloc;
   CeedBasis           basis_x, basis_stats;
   CeedVector          x_coord, q_data;
-} * SpanStatsSetupData;
+} *SpanStatsSetupData;
```
jrwrigh, line 28: Yeah, I didn't like that either, but `clang-format` wanted the extra space.

jeremylt: Yeah. I should update to a newer `clang-format` when I get a chance. My local is ahead of the CI version

include/ceed/ceed.h
```diff
@@ -509,7 +509,11 @@ CEED_EXTERN int CeedOperatorGetNumQuadraturePoints(CeedOperator op, CeedInt *num
 CEED_EXTERN int CeedOperatorGetFlopsEstimate(CeedOperator op, CeedSize *flops);
 CEED_EXTERN int CeedOperatorContextGetFieldLabel(CeedOperator op, const char *field_name, CeedContextFieldLabel *field_label);
 CEED_EXTERN int CeedOperatorContextSetDouble(CeedOperator op, CeedContextFieldLabel field_label, double *values);
+CEED_EXTERN int CeedOperatorContextGetDoubleRead(CeedOperator op, CeedContextFieldLabel field_label, size_t *num_values, const double **values);
+CEED_EXTERN int CeedOperatorContextRestoreDoubleRead(CeedOperator op, CeedContextFieldLabel field_label, const double **values);
```
jeremylt, line 513: I specified in the signature that this is read-only and added a restore function as well

jrwrigh (approved):
Looks good to me! Just one small question

interface/ceed-operator.c
```diff
@@ -301,15 +416,15 @@ static int CeedOperatorContextSetGeneric(CeedOperator op, CeedContextFieldLabel
     CeedCall(CeedCompositeOperatorGetSubList(op, &sub_operators));
     if (num_sub != field_label->num_sub_labels) {
       // LCOV_EXCL_START
-      return CeedError(op->ceed, CEED_ERROR_UNSUPPORTED,
-                       "ContextLabel does not correspond to composite operator. Use CeedOperatorGetContextFieldLabel().");
+      return CeedError(op->ceed, CEED_ERROR_UNSUPPORTED, "ContextLabel does not correspond to composite operator");
```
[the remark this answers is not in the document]
jeremylt: Yeah, any context that doesn't have the field will leave a NULL value in this list of labels

PR #1149, opened 2023-01-30
fluids: Data-driven SGS modeling

jrwrigh:
- Implementing the data-driven subgrid-stress modeling [detailed here](http://arxiv.org/abs/2212.00332)
- This should be pushed after #1147

#### Todos
- [x] Add Operator for evaluating stresses at nodes
- [x] Add QFunction for integrating SGS stresses into the newtonian code
- [x] Implement logic/command line arguments for controlling setup (make sure flags are equivalent)
- [x] Enable `*Destroy` functions
- [x] Create DM for SGS stresses
- [x] Separate out anisotropy projection (needed for differential filtering)
- [x] Replace `M_inv` with `ksp` object
- [x] Merge SGS DM creation into SGSSetup
- [x] Store anisotropy data at quadrature points
- [x] Add documentation


examples/fluids/qfunctions/sgs_dd_model.h
```diff
@@ -0,0 +1,170 @@
...
+  const CeedInt     num_neurons = sgsdd_ctx->num_neurons;
+  const CeedInt     num_inputs  = sgsdd_ctx->num_inputs;
+  const CeedInt     num_outputs = sgsdd_ctx->num_outputs;
+  const CeedScalar  alpha       = sgsdd_ctx->alpha;
+  const CeedScalar *bias1       = &sgsdd_ctx->data[sgsdd_ctx->offsets.bias1];
+  const CeedScalar *bias2       = &sgsdd_ctx->data[sgsdd_ctx->offsets.bias2];
+  const CeedScalar *weight1     = &sgsdd_ctx->data[sgsdd_ctx->offsets.weight1];
+  const CeedScalar *weight2     = &sgsdd_ctx->data[sgsdd_ctx->offsets.weight2];
+  CeedScalar        V[num_neurons];
+  SetValueN(V, 0, num_neurons);
```
jrwrigh, line 54: This will fail for GPUs due to variable-length-array incompatabilities. I'm not sure what the "correct way" to allocate a work vector inside a QFunction is. I could obviously allocate something in the context and then access that, but that seems like an abuse of the context.

jedbrown: You can't allocate private workspace in the context because it's shared on the GPU. Do we need this to be dynamic or can it be a static specialization?

jrwrigh: Gotcha. I'm not sure what static specialization is. It doesn't necessarily need to be dynamically allocated; we could just always have an array set to some compile time constant and error out if the number of neurons requested is greater than that constant.

jedbrown: Like you did for the mass projections, for example.

jrwrigh: Note that `V` is only a work vector and doesn't need to be persistent.

> Like you did for the mass projections, for example.

Meaning like doing a `DMClone` and bringing a `Vec` in as a collocated array? I think doing a CeedVector would probably be easier, as we need this vector per quadrature point and can thus just create one from an element restriction or a custom strided vector creation. I guess I'd need to have that be an output for the Operator, as I need to write and read from it.

jedbrown: This is a small workspace, like 20 "neuron" hidden layer, right? You can put the weights/biases in the context and just use a constant size stack array for the hidden layer workspace.

jrwrigh: Yeah, that's right. I could just set a constant size 20 neuron work array and that would be fine for this neural network.

jrwrigh (commented):
Any feedback on these (d00e38d) changes @jedbrown @jeremylt ? Wanted to check in before I continued forward.

Basically took the changes introduced in Ratel here: https://gitlab.com/micromorph/ratel/-/commit/283206d82b406db70993966f9adfff77cab6c05d
and expanded it to allow for CeedVector inputs/outputs and for `CeedOperatorApplyAdd`. Obviously the full tensor-product of global, local, CeedVector, and ApplyAdd would be 18 `ApplyCeedOperator*` functions, but I'd only plan on adding them on an as needed basis.

examples/fluids/include/matops.h
```diff
@@ -22,16 +22,21 @@ struct OperatorApplyContext_ {
...
 PetscErrorCode VecP2C(Vec X_petsc, PetscMemType *mem_type, CeedVector x_ceed);
 PetscErrorCode VecC2P(CeedVector x_ceed, PetscMemType mem_type, Vec X_petsc);
 PetscErrorCode VecReadP2C(Vec X_petsc, PetscMemType *mem_type, CeedVector x_ceed);
 PetscErrorCode VecReadC2P(CeedVector x_ceed, PetscMemType mem_type, Vec X_petsc);
+PetscErrorCode VecCopyP2C(Vec X_petsc, CeedVector x_ceed);
+
+PetscErrorCode ApplyCeedOperatorGlobalToGlobal(Vec X, Vec Y, OperatorApplyContext ctx);
+PetscErrorCode ApplyAddCeedOperatorCeedVectorToLocal(CeedVector x_ceed, Vec Y_loc, OperatorApplyContext ctx);
+PetscErrorCode ApplyCeedOperatorCeedVectorToGlobal(CeedVector x_ceed, Vec Y, OperatorApplyContext ctx);
+PetscErrorCode ApplyAddCeedOperatorGlobalToCeedVector(Vec X, CeedVector y_ceed, OperatorApplyContext ctx);
```
jrwrigh, line 40: Any feedback on these (d00e38d) changes @jedbrown @jeremylt ? Wanted to check in before I continued forward.

Basically took the changes introduced in Ratel here: https://gitlab.com/micromorph/ratel/-/commit/283206d82b406db70993966f9adfff77cab6c05d
and expanded it to allow for CeedVector inputs/outputs and for `CeedOperatorApplyAdd`. Obviously the full tensor-product of global, local, CeedVector, and ApplyAdd would be 18 `ApplyCeedOperator*` functions, but I'd only plan on adding them on an as needed basis.

jrwrigh: Agreed. Most of the use cases for CeedVector inputs/outputs are for short-lived operators (a vector of ones for creation of a lumped mass matrix) or for data that doesn't live at nodes (spanwise statistics collection). 

I can add a comment onto those functions emphasizing the use of `Vec` over `CeedVector` for manipulation of data when possible.

jrwrigh: In your mind, would _every_ `CeedVector` have a corresponding `Vec`? Even stuff like `q_data`? Or just operator inputs/outputs?

I agree with the general principal of putting everything in `Vec`s, but am not sure if I agree with/understand that being a hard rule (which I don't think you're advocating for).

jeremylt: In my perfect world, an application developer would only need to know about PETSc functions and libCEED QFunctions and nothing else about libCEED. 

We aren't there yet though.

If this is already wrapped in a `OperatorApplyContext` can't you bypass this whole issue by `DMGetLocalVector`ing and setting that local Vec to all 1s?

jrwrigh: > If this is already wrapped in a `OperatorApplyContext` can't you bypass this whole issue by `DMGetLocalVector`ing and setting that local Vec to all 1s?

In the case of the lumped mass matrix, the issue is one of laziness. Currently, I have to create a `CeedVector`, regardless of which way I go. If I want to use a local `Vec` as the input method, then I need to create both a `CeedVector` and a `Vec`, which is redundant and (pedantically) annoying.

A potential solution to this would be to generate the required `CeedVector` on the fly via, similar to how we use `DMGetLocalVector` to get local `Vec`s on the fly. 

*Sidenote: I previously thought that `CeedVectors` stored information about their data layout, but now realize they only know the size of the data, which makes things much more straight forward (and I think understand your point better too).*

So maybe something like:
```c
PetscErrorCode OperatorApplyContextCreate(DM dm_x, DM dm_y, Ceed ceed, CeedOperator op_apply, CeedVector x_ceed, CeedVector y_ceed, Vec X_loc,
                                          Vec Y_loc, OperatorApplyContext *op_apply_ctx) {
  CeedSize x_size, y_size;

  PetscFunctionBeginUser;
  PetscCall(PetscNew(op_apply_ctx));

  // Copy PETSc Objects
  if (dm_x) PetscCall(PetscObjectReference((PetscObject)dm_x));
  (*op_apply_ctx)->dm_x = dm_x;
  if (dm_y) PetscCall(PetscObjectReference((PetscObject)dm_y));
  (*op_apply_ctx)->dm_y = dm_y;

  if (X_loc) PetscCall(PetscObjectReference((PetscObject)X_loc));
  (*op_apply_ctx)->X_loc = X_loc;
  if (Y_loc) PetscCall(PetscObjectReference((PetscObject)Y_loc));
  (*op_apply_ctx)->Y_loc = Y_loc;

  // Copy libCEED objects
  CeedOperatorGetActiveVectorLengths(op_apply, &x_size, &y_size);
  if (x_ceed) CeedVectorReferenceCopy(x_ceed, &(*op_apply_ctx)->x_ceed);
  else CeedVectorCreate(ceed, x_size, &(*op_apply_ctx)->x_ceed);
  if (y_ceed) CeedVectorReferenceCopy(y_ceed, &(*op_apply_ctx)->y_ceed);
  else CeedVectorCreate(ceed, y_size, &(*op_apply_ctx)->y_ceed);

  CeedOperatorReferenceCopy(op_apply, &(*op_apply_ctx)->op);
  CeedReferenceCopy(ceed, &(*op_apply_ctx)->ceed);

  PetscFunctionReturn(0);
}
```

We could even throw some checks in there for verifying that the size of the local `Vec`s matches the size for the operator. I think that combined with a version of `RatelCeedOperatorCreateLocalVecs` eliminates any desire to have `CeedVector` inputs/outputs.

jeremylt: I can probably simplify things a tiny bit in Ratel by making the CeedVectors inside of the ContextCreate function. The length checks I really should add too.

examples/fluids/problems/sgs_dd_model.c
```diff
@@ -0,0 +1,365 @@
...
+  // -- Create operator for SGS DD model nodal evaluation
+  switch (user->phys->state_var) {
+    case STATEVAR_PRIMITIVE:
+      CeedQFunctionCreateInterior(ceed, 1, ComputeSGS_DDAnisotropicNodal_Prim, ComputeSGS_DDAnisotropicNodal_Prim_loc, &qf_sgs_dd_nodal);
+      break;
+    case STATEVAR_CONSERVATIVE:
+      CeedQFunctionCreateInterior(ceed, 1, ComputeSGS_DDAnisotropicNodal_Conserv, ComputeSGS_DDAnisotropicNodal_Conserv_loc, &qf_sgs_dd_nodal);
+      break;
+    default:
+      SETERRQ(PetscObjectComm((PetscObject)user->dm), PETSC_ERR_SUP, "No statisics collection available for chosen state variable");
```
jrwrigh, line 116: Reminder fix error message

examples/fluids/include/matops.h
jrwrigh: For this PR, I've stuck with only the `{Global,Local}To{Global,Local}` functions for right now. I'll do the others in a later PR.

examples/fluids/qfunctions/utils_eigensolver_jacobi.h
```diff
@@ -0,0 +1,327 @@
...
+  CeedInt max_num_iters = max_num_sweeps * N * (N - 1) / 2;
+  for (n_iters = 1; n_iters <= max_num_iters; n_iters++) {
+    CeedInt i, j;
+    MaxEntry(A, N, max_idx_row, &i, &j);
+
+    // If A[i][j] is small compared to A[i][i] and A[j][j], set it to 0.
+    if ((A[i * N + i] + A[i * N + j] == A[i * N + i]) && (A[j * N + j] + A[i * N + j] == A[j * N + j])) {
+      A[i * N + j]   = 0.0;
+      max_idx_row[i] = MaxEntryRow(A, N, i);
+    }
```
jrwrigh, line 301: This code currently checks every loop as to the index with the largest entry. I believe this will probably very sub-optimal on GPU. I've left it here for now just to avoid premature optimization, but what I've seen others do is just loop over all the off-diagonal entries (that was for CPU SIMD instructions, but I believe the performance implications are more-or-less the same).

examples/fluids/src/velocity_gradient_projection.c
```diff
@@ -0,0 +1,173 @@
...
+    PetscCall(KSPCreate(comm, &grad_velo_proj->ksp));
+    PetscCall(KSPSetOptionsPrefix(grad_velo_proj->ksp, "velocity_gradient_projection_"));
+    {
+      PC pc;
+      PetscCall(KSPGetPC(grad_velo_proj->ksp, &pc));
+      PetscCall(PCSetType(pc, PCJACOBI));
+      PetscCall(PCJacobiSetType(pc, PC_JACOBI_ROWSUM));
+      PetscCall(KSPSetType(grad_velo_proj->ksp, KSPPREONLY));
+      // TODO Not sure if the option below are necessary
+      PetscCall(KSPSetConvergenceTest(grad_velo_proj->ksp, KSPConvergedSkip, NULL, NULL));
```
jrwrigh, line 146: As comment says, not sure if this is necessary or not. I notice that if I monitor the convergence, it does report residuals before and after preconditioning application, which makes me worried that it's actually calculating the residual here. 

Any ideas how to skip the residual calculation? Or if it's actually happening at all?

examples/fluids/qfunctions/utils.h
```diff
@@ -18,17 +18,128 @@
...
+
+// @brief Curl of vector given its gradient
+CEED_QFUNCTION_HELPER void Curl3(const CeedScalar gradient[3][3], CeedScalar v[3]) {
+  v[0] = gradient[2][1] - gradient[1][2];
+  v[1] = gradient[0][2] - gradient[2][0];
+  v[2] = gradient[1][0] - gradient[0][1];
+}
+
+// @brief Matrix vector product, b = Ax + b. A is NxM, x is M, b is N
+CEED_QFUNCTION_HELPER void MatVecNM(const CeedScalar *A, const CeedScalar *x, const CeedInt N, const CeedInt M, const CeedTransposeMode transpose_A,
```
jrwrigh, line 71: I assume using `CeedTransposeMode` here is ok? I originally made my own `enum`, but then realized that it was functionally identical to this one.

jrwrigh (commented):
I still need to get approval for the model parameter pushing and documentation, but code-wise, this is ready to review.

examples/fluids/qfunctions/utils_eigensolver_jacobi.h
jedbrown: Thanks for pointing this out. I suspect it won't be a significant cost for our use case here, but likely will be in solid mechanics where it's in the critical path of the main solver.

examples/fluids/index.md
```diff
@@ -301,6 +301,18 @@ Note that this wave speed is specific to ideal gases as $\gamma$ is an ideal gas
...
+
+When a fluid simulation is under-resolved (the smallest length scale resolved by the grid is much larger than the Kolmogorov length scale), this is mathematically interpreted as filtering the Navier-Stokes equations. 
+This filtering operation results in an extra stress-like term in the equations, representing the effect of unresolved (or "subgrid" scale) structures in the flow. 
+To close the problem, this term is modeled using a subgrid stress (SGS) model.
+
+#### Data-driven SGS Model
+
+The data-driven SGS model implemented here uses a small neural network (1 layer of 20 neurons) to compute the SGS term.
+The SGS tensor is calculated at nodes using an $L^2$ projection of the velocity gradient and grid anisotropy tensor, and then interpolated onto quadrature points.
+Parameters for the neural network are put into files in a directory found in `-sgs_model_dd_parameter_dir`.
```
jedbrown, line 314: I take it this is multiple files in the directory?

examples/fluids/qfunctions/utils.h
```diff
@@ -40,6 +151,24 @@ CEED_QFUNCTION_HELPER void KMUnpack(const CeedScalar v[6], CeedScalar A[3][3]) {
...
+  v[4]                    = A[2][0] * weight;
+  v[5]                    = A[1][0] * weight;
+}
+
+// @brief Calculate metric tensor from mapping, g_{ij} = xi_{k,i} xi_{k,j} = dXdx^T dXdx
+CEED_QFUNCTION_HELPER void KMMetricTensor(const CeedScalar dXdx[3][3], CeedScalar km_g_ij[6]) {
+  CeedScalar g_ij[3][3] = {{0.}};
+  MatMat3(dXdx, dXdx, CEED_TRANSPOSE, CEED_NOTRANSPOSE, g_ij);
+  KMPack(g_ij, km_g_ij);
+}
```
jedbrown, line 170: Is this a `KM` (Kelvin-Mandel form of a symmetric tensor)?

examples/fluids/src/setupts.c
```diff
@@ -224,11 +224,19 @@ PetscErrorCode IFunction_NS(TS ts, PetscReal t, Vec Q, Vec Q_dot, Vec G, void *u
   // Apply CEED operator
   CeedOperatorApply(user->op_ifunction, user->q_ceed, user->g_ceed, CEED_REQUEST_IMMEDIATE);
 
+  // if (user->app_ctx->sgs_model_type == SGS_MODEL_DATA_DRIVEN) {
+  //   PetscCall(SGS_DD_ModelApplyIFunction(user, Q_loc, user->q_ceed, user->g_ceed));
+  // }
+
```
jedbrown, line 230: Delete or uncomment?

jedbrown (commented):
Looking good. Some modest comments. We need to include example weights so we can run tests and have something to mention in the docs (too opaque now).

examples/fluids/qfunctions/utils_eigensolver_jacobi.h
jrwrigh: Gotcha. I feel like it is a significant cost for the model itself, but I guess until we integrate it into the Jacobian it'll only affect the residual evaluation time.

examples/fluids/src/velocity_gradient_projection.c
jrwrigh: > Also note that ROWSUM is fragile since it won't work for high order tets.

Right. There isn't another lumped-mass matrix option right now, is there? I'm pretty sure "special lumping" would get around that issue if/when we run into it.

examples/fluids/index.md
jrwrigh: Yes. I was planning on mentioning the format and files required explicitly.

examples/fluids/qfunctions/utils.h
jrwrigh: Yep.

examples/fluids/src/setupts.c
jrwrigh: Delete. Thanks for catching.

examples/fluids/index.md
jrwrigh: Done

jrwrigh:
SGS documentation done. 

Note I also have a commit that rearranges the fluids documentation a bit. I think this makes the theoretical section of the docs a little bit more digestable/navigable.

examples/fluids/qfunctions/utils.h
jrwrigh: Looks like there might be a problem for CUDA backends (and probably HIP):
```
/home/jawr8143/software/libCEED/backends/cuda/ceed-cuda-compile.cpp:93 in CeedCompileCuda(): NVRTC_ERROR_COMPILATION
default_program(422): error: identifier "CeedTransposeMode" is undefined

default_program(425): error: identifier "CEED_NOTRANSPOSE" is undefined

default_program(428): error: identifier "CEED_TRANSPOSE" is undefined
[...]
```

If I follow the code correctly, `CeedTransposeMode` is only defined in `<ceed.h>`, but that header gets stripped away when loading in the JIT source. 

Should we just add the `CeedTransposeMode` definition to `ceed/jit-source/cuda/cuda-jit.h` and it's HIP equivalent?

jeremylt: If we want to move it (should be fine to do so), then it should go into `<ceed/types.h>`

jrwrigh: Yep, the following change fixes it:
```diff
diff --git i/include/ceed/ceed.h w/include/ceed/ceed.h
index 233200cf..a838f477 100644
--- i/include/ceed/ceed.h
+++ w/include/ceed/ceed.h
@@ -258,14 +258,7 @@ CEED_EXTERN const CeedElemRestriction CEED_ELEMRESTRICTION_NONE;
 /// @ingroup CeedQFunction
 CEED_EXTERN const CeedQFunction CEED_QFUNCTION_NONE;

-/// Denotes whether a linear transformation or its transpose should be applied
-/// @ingroup CeedBasis
-typedef enum {
-  /// Apply the linear transformation
-  CEED_NOTRANSPOSE,
-  /// Apply the transpose
-  CEED_TRANSPOSE
-} CeedTransposeMode;
+/// Maps enum CeedTransposeMode to strings
 CEED_EXTERN const char *const CeedTransposeModes[];

 /// Argument for CeedElemRestrictionCreateStrided that L-vector is in the Ceed backend's preferred layout.
diff --git i/include/ceed/types.h w/include/ceed/types.h
index b7d32010..650e616a 100644
--- i/include/ceed/types.h
+++ w/include/ceed/types.h
@@ -137,4 +137,13 @@ typedef enum {
   CEED_ERROR_UNSUPPORTED = -3,
 } CeedErrorType;

+/// Denotes whether a linear transformation or its transpose should be applied
+/// @ingroup CeedBasis
+typedef enum {
+  /// Apply the linear transformation
+  CEED_NOTRANSPOSE,
+  /// Apply the transpose
+  CEED_TRANSPOSE
+} CeedTransposeMode;
+
 #endif
```

Have to keep the ` CEED_EXTERN const char *const CeedTransposeModes[];` in `ceed.h` unless we want to move `CEED_EXTERN` inside of `ceed/types.h` (or some other option).

examples/fluids/include/matops.h
jrwrigh: > I'll do the others in a later PR.

Opened up #1185

jrwrigh:
And we (finally) have a test pushed up for this branch. Assuming that passes everything (which is should based on manual testing on noether), this PR should be completely good to go.

examples/fluids/index.md
```diff
@@ -301,6 +307,53 @@ Note that this wave speed is specific to ideal gases as $\gamma$ is an ideal gas
...
+For implicit LES, the subgrid stress is set to zero and the numerical properties of the discretized system are assumed to account for the effect of subgrid scale structures on the filtered solution field.
+For explicit LES, it is defined by a subgrid stress model.
+
+#### Data-driven SGS Model
+
+The data-driven SGS model implemented here uses a small neural network to compute the SGS term.
+The SGS tensor is calculated at nodes using an $L^2$ projection of the velocity gradient and grid anisotropy tensor, and then interpolated onto quadrature points.
+More details regarding the theoretical background of the model can be found in {cite}`prakashDDSGS2022` and {cite}`prakashDDSGSAnisotropic2022`.
+
+The neural network itself consists of 1 layer and 20 neurons, using Leaky ReLU as its activation function.
```
jedbrown, line 344:
```suggestion
The neural network itself consists of 1 hidden layer and 20 neurons, using Leaky ReLU as its activation function.
```

jedbrown (approved):
I think this is really close; sorry about my super slow review. Looks great in general.

Rezgar has a non-iterative method for 3x3 symmetric eigensolves that we can consider later.

examples/fluids/src/velocity_gradient_projection.c
jrwrigh: Yeah, we'll need to revisit more than just here too if we run into lumped mass matrix issues.

We could also do a collocated quadrature approach and then just use `CeedOperatorLinearAssembleDiagonal`, right? Then we'd could just hand off that vector to the KSP as the preconditioning matrix.

jrwrigh: Maybe we should fork this conversation off into a GH issue? Just to have a central location to discuss, when we do get to it.

jedbrown: Yeah, can fork. Collocated quadrature doesn't fix the issue with rowsum lumping.

jrwrigh:
Not sure where the RTD error is coming from. But assuming that's innocuous, feel free to merge (I'm hoping on a plane in 5 minutes).

jeremylt:
RTD issue was fixed on main.

PR #1150, opened 2023-01-31
CPU diagonal assembly for multiple active bases

jeremylt:
I need to fix up all diagonal assembly to work like this and make the fully assembly when num_active_bases > 1 for this initial PR.

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -116,11 +116,11 @@ static int CeedOperatorCreateFallback(CeedOperator op) {
     CeedCall(CeedQFunctionCreateFallback(ceed_fallback, op->dqfT, &dqfT_fallback));
     CeedCall(CeedOperatorCreate(ceed_fallback, qf_fallback, dqf_fallback, dqfT_fallback, &op_fallback));
     for (CeedInt i = 0; i < op->qf->num_input_fields; i++) {
-      CeedCall(CeedOperatorSetField(op_fallback, op->input_fields[i]->field_name, op->input_fields[i]->elem_restr, op->input_fields[i]->basis,
+      CeedCall(CeedOperatorSetField(op_fallback, op->input_fields[i]->field_name, op->input_fields[i]->elem_rstr, op->input_fields[i]->basis,
                                     op->input_fields[i]->vec));
     }
     for (CeedInt i = 0; i < op->qf->num_output_fields; i++) {
-      CeedCall(CeedOperatorSetField(op_fallback, op->output_fields[i]->field_name, op->output_fields[i]->elem_restr, op->output_fields[i]->basis,
+      CeedCall(CeedOperatorSetField(op_fallback, op->output_fields[i]->field_name, op->output_fields[i]->elem_rstr, op->output_fields[i]->basis,
                                     op->output_fields[i]->vec));
     }
     CeedCall(CeedQFunctionAssemblyDataReferenceCopy(op->qf_assembled, &op_fallback->qf_assembled));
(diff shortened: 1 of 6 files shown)


jeremylt:
The test I added works, so @rezgarshakeri you should be able to test Jacobi based PCs with this branch for your mixed elasticity Ratel MR

jeremylt:
As a heads up @nbeams, these updates need to be ported to the GPU backends in another PR and then the full assembly needs to be updated in another couple of PRs. I don't have a timeline for that yet.

nbeams:
> As a heads up @nbeams, these updates need to be ported to the GPU backends in another PR and then the full assembly needs to be updated in another couple of PRs. I don't have a timeline for that yet.

Let me know at what point (if any) you want me to hop in and work on GPU stuff.  I was assuming maybe at the full assembly point, since that was my code originally -- is that what your comment is referring to?

jeremylt:
I don't have a plan of attack yet, I more just wanted to put it on your radar since it will involve code you wrote. 

I think adding this diagonal assembly change to the GPU should be easier but the full assembly update will be harder.

rezgarshakeri:
> The test I added works, so @rezgarshakeri you should be able to test Jacobi based PCs with this branch for your mixed elasticity Ratel MR

Thanks @jeremylt, I'll try it.

interface/ceed-preconditioning.c
```diff
@@ -1160,132 +1262,172 @@ int CeedOperatorAssemblyDataCreate(Ceed ceed, CeedOperator op, CeedOperatorAssem
...
+  @param[out] num_active_bases      Number of active bases
+  @param[out] num_eval_mode_in      Pointer to hold array of numbers of input CeedEvalModes, or NULL
+  @param[out] eval_mode_in          Pointer to hold arrays of input CeedEvalModes, or NULL
+  @param[out] eval_mode_offsets_in  Pointer to hold arrays of input offsets at each quadrature point
+  @param[out] num_eval_mode_out     Pointer to hold array of numbers of output CeedEvalModes, or NULL
+  @param[out] eval_mode_out         Pointer to hold arrays of output CeedEvalModes, or NULL
+  @param[out] eval_mode_offsets_out Pointer to hold arrays of output offsets at each quadrature point
+  @param[out] num_output_components Total number of output components at each quadrature point
 
   @return An error code: 0 - success, otherwise - failure
```
jedbrown, line 1287: I think this man page would benefit from an explanation of the data model. Does `num_active_bases` apply to in or out? How is `num_output_components` counted across the various fields?

jedbrown (approved):

interface/ceed-preconditioning.c
jeremylt: Agreed - the names are less obvious in the light of Monday morning.

interface/ceed-preconditioning.c
```diff
@@ -569,13 +573,20 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
...
+                                                &eval_modes_out, NULL, NULL));
+  CeedBasis *bases;
+  CeedCall(CeedOperatorAssemblyDataGetBases(data, NULL, &bases, NULL, NULL));
+  CeedBasis basis_in = bases[0];
+
+  if (num_active_bases > 1) {
+    // LCOV_EXCL_START
+    return CeedError(ceed, CEED_ERROR_UNSUPPORTED, "Cannot assemble operator with multiple active bases");
+    // LCOV_EXCL_STOP
+  }
```
rezgarshakeri, line 588: This means we still don't support full assembly for multiple bases, right?

jeremylt: Correct. That will be much, much more work

rezgarshakeri (approved):

PR #1151, opened 2023-02-03
operator: Get OperatorField by field name

jrwrigh:
Get a `CeedOperatorField` from a `CeedOperator` by the field's name.

Still need to actually test the code, but the prototype from 9a889ca works correctly.

- [ ] Update Release notes
- [x] Verify error message prints correctly
- [x] Double-check that it's working in this implementation

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -873,6 +873,48 @@ int CeedOperatorGetFields(CeedOperator op, CeedInt *num_input_fields, CeedOperat
 }
 
 /**
+  @brief Get a CeedOperatorField of an CeedOperator from its name
+
+  Note: Calling this function asserts that setup is complete and sets the CeedOperator as immutable.
+
+  @param[in]  op         CeedOperator
+  @param[in]  field_name Name of desired CeedOperatorField
+  @param[out] op_field   CeedOperatorField corresponding to the name
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+int CeedOperatorGetFieldByName(CeedOperator op, const char *field_name, CeedOperatorField *op_field) {
+  CeedInt            num_input_fields, num_output_fields;
+  CeedOperatorField *input_fields, *output_fields;
+  char              *name;
+  CeedCall(CeedOperatorGetFields(op, &num_input_fields, &input_fields, &num_output_fields, &output_fields));
+
+  for (CeedInt i = 0; i < num_input_fields; i++) {
+    CeedCall(CeedOperatorFieldGetName(input_fields[i], &name));
+    if (!strcmp(name, field_name)) {
+      *op_field = input_fields[i];
+      return CEED_ERROR_SUCCESS;
+    }
+  }
+
+  for (CeedInt i = 0; i < num_output_fields; i++) {
+    CeedCall(CeedOperatorFieldGetName(output_fields[i], &name));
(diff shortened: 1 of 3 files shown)


jedbrown (commented):
Do we enforce that names cannot be reused between input and output? Is that a restriction we'll live with forever, or might the caller of this interface want to guarantee that they're getting an input or output field?

jrwrigh:
I don't know if we enforce that, but we implicitly do since `CeedOperatorSetField` doesn't specify inputs or outputs, only the field name. So if a user had the same name for input and output, I'm not sure how `CeedOperatorSetField` would know which one to hook up with.

As for whether that _should_ be the case, I think that's a fine restriction. Otherwise, end users may think that there's a possibility of in/out vectors, which I don't believe we guarantee. If we do want to allow in/out vectors at some point, then we should change this interface and `CeedOperatorSetField`.

jeremylt:
We don't specifically check for name reuse when the field is added to the QF, but we probably should

jrwrigh:
Should be good to go now. Note that I anticipate it'll probably fail code coverage testing again, as this isn't actually tested in the primary library, but will be with the fluids example once #1149 is merged.

jedbrown (approved):

interface/ceed-operator.c
```diff
@@ -872,6 +872,48 @@ int CeedOperatorGetFields(CeedOperator op, CeedInt *num_input_fields, CeedOperat
...
+  for (CeedInt i = 0; i < num_output_fields; i++) {
+    CeedCall(CeedOperatorFieldGetName(output_fields[i], &name));
+    if (!strcmp(name, field_name)) {
+      *op_field = output_fields[i];
+      return CEED_ERROR_SUCCESS;
+    }
+  }
+
+  bool has_name = op->name;
+  // LCOV_EXCL_START
```
jeremylt, line 911: Hey, @jrwrigh can we update this function so that we can let the return value be `NULL` if the field isn't found? That would be very helpful for our plasticity branch in Ratel.

jrwrigh: Hmmm. I had though about returning a `NULL` in case it isn't found, but thought that it would just be more work to handle it on the user's end, at least to get a reasonable error message out of it. But I can imagine returning `NULL` being quite handy.

tl;dr Go for it.

jrwrigh: Maybe a separate `CeedOperator{Has,Check}Field` function could also work here.

jeremylt: I think either is fine - its the same basic logic either way, just a question of one function call or two

jrwrigh: I kinda like `CeedOperator{Has,Check}FieldName` a bit more, mainly for keeping the easy-to-read error message.

jeremylt: I just updated this in Ratel, I use the NULL return there and now use PetscCheck(theThing, ratel->comm, PETSC_ERR_SUP, "the thing not found") or if (theThing) as appropriate

jeremylt: Though I think I might not need this with my new plan

jeremylt: Still need to get back to the issue from this PR about checking for non-dupe fields

jrwrigh: Gotcha. Forgot about `PetscCheck`. That's simple enough.

PR #1152, opened 2023-02-03
Support NULL restrict/prolong operators in multigrid

zatkins-dev:
This MR adds support for NULL `CeedOperator*` arguments for `op_prolong` and `op_restrict` in `CeedSingleOperatorMultigridLevel` in order to facilitate creating multigrid operators for non-volumetric operators.

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -757,6 +757,14 @@ static int CeedSingleOperatorMultigridLevel(CeedOperator op_fine, CeedVector p_m
   CeedCall(CeedOperatorSetName(*op_coarse, op_fine->name));
 
   // Restriction/Prolongation Operators
+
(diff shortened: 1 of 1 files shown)


jeremylt:
This is most of the logic needed, but you also need to make `basis_c_to_f` optional. This is where the factorization error came into play in your branch, I think.

The logic should be something like `if (op_restrict || op_prolong) {create/require basis_c_to_f}` in all of the functions that call this inner function.

zatkins-dev:
Alright, I added logic to only create the `basis_c_to_f` if prolong/restrict operators are non-null. Please let me know if I'm using the correct error type (`CEED_ERROR_INCOMPATIBLE`) or if that should be something else.

jeremylt (approved):
Awesome, thanks for the quick fix.

The only thing I'd add is ", or NULL" to the `op_prolong` and `op_restrict` arguments for the docstrings of the 3 user facing functions

PR #1153, opened 2023-02-03
Update docstrings for NULL arguments

jeremylt:
Also, I noticed that `p_mult_fine` should be optional


zatkins-dev:
Oops, I didn't see your comment on my PR. Looks good!

zatkins-dev (approved):

PR #1155, opened 2023-02-13
Prevent duplicate QFunctino/Operator field names

### LINKED ISSUE #1154: We don't specifically check for name reuse in any way when the field is added to the QF, but we probably should
We don't specifically check for name reuse in any way when the field is added to the QF, but we probably should

_Originally posted by @jeremylt in https://github.com/CEED/libCEED/pull/1151#issuecomment-1416052103_

jeremylt:
Closes #1154


jrwrigh (approved):
Looks good to me

Issue #350, opened 2019-09-17
Non-Tensor Basis Support


jeremylt:
Add support for non-tensor bases
- [x] OCCA
- [ ] CUDA Gen
- [ ] HIP Gen
- [x] MAGMA

nbeams:
Non-tensor capability was added to the MAGMA backend with #424.

jedbrown:
Cc: #1022

jeremylt:
Part of https://github.com/CEED/libCEED/issues/839

PR #1156, opened 2023-02-15
H(curl) basis and `CEED_EVAL_CURL`

sebastiangrimberg:
Preliminary support for H(curl). Changes are very similar to #878.

This handles `CeedBasis` updates for H(curl) and also extends some of the missing H(div) functionality in `ceed-preconditioning.c`, for example. Any changes to element restriction for high-order Nedelec elements are left to a future PR, as is backend support (first non-tensor `cuda/hip-ref`, OCCA, and MAGMA, then later on `cuda/hip-shared/gen`).

Note: I opened this PR from a fork since I didn't have write access to CEED/libCEED, but if it's preferred to create the branch there then I can close and reopen a new PR once I have access.

TODO
- [x] Add tests similar to #878.
- [ ] ~Backend support: So far have only touched `ref-basis`, `ref-operator`, and `blocked-operator`. What do I need to do for CUDA/HIP/MAGMA/OCCA support for non-tensor H(div) and H(curl)? Maybe this is a bigger ask (#350).~
    - Note: Left for a later PR
- [x] Diagonal assembly: Does anything special need to be done here? `CeedOperatorGetBasisPointer`? What about for p-multigrid?
    - Fixes to multigrid and operator assembly completed for CPU, left for later for backends
- [x] Update Julia/Python/Rust bindings. Also Fortran?
    - I updated the Julia bindings which seems to have added a few methods which I didn't touch (`CeedOperatorContextGetDoubleRead`, `CeedOperatorContextRestoreDoubleRead`, etc.). I think this might be just because the generator hadn't been run for a past PR, but let me know if I'm doing something incorrectly.


jedbrown:
Thanks! It looks like maybe your editor formatting configuration is different. Can you have your editor use `clangd` so the auto-formatter works, or use `make format` before committing to avoid whitespace noise?

sebastiangrimberg:
> Thanks! It looks like maybe your editor formatting configuration is different. Can you have your editor use `clangd` so the auto-formatter works, or use `make format` before committing to avoid whitespace noise?

I used `make format` but it might be an issue with my version of `clang-format` (Homebrew uses 15 now vs. Ubuntu 22 default is 14). I'll rerun tomorrow with v14!

jeremylt:
Our pipeline currently uses 14. I haven't had the chance to update yet

jedbrown:
Done. 
https://github.com/CEED/libCEED/pull/1157

backends/ref/ceed-ref-basis.c
```diff
@@ -246,10 +248,19 @@ static int CeedBasisApply_Ref(CeedBasis basis, CeedInt num_elem, CeedTransposeMo
         }
         CeedCallBackend(CeedTensorContractApply(contract, num_comp, P, num_elem, Q, div, t_mode, add, u, v));
       } break;
-      // LCOV_EXCL_START
       // Evaluate the curl to/from the quadrature points
-      case CEED_EVAL_CURL:
-        return CeedError(ceed, CEED_ERROR_BACKEND, "CEED_EVAL_CURL not supported");
+      case CEED_EVAL_CURL: {
+        CeedInt           cdim = (dim < 3) ? 1 : dim;
```
jeremylt, line 253: What does c in `c_dim` mean? It would be nice to have something at the interface level that other backends can use like `CeedBasisGetNumQuadratureComponents`
```suggestion
        CeedInt           c_dim = (dim < 3) ? 1 : dim;
```

include/ceed/ceed.h
```diff
@@ -382,13 +383,18 @@ CEED_EXTERN int CeedBasisGetInterp1D(CeedBasis basis, const CeedScalar **interp_
...
 
 CEED_EXTERN int CeedGaussQuadrature(CeedInt Q, CeedScalar *q_ref_1d, CeedScalar *q_weight_1d);
 CEED_EXTERN int CeedLobattoQuadrature(CeedInt Q, CeedScalar *q_ref_1d, CeedScalar *q_weight_1d);
 CEED_EXTERN int CeedQRFactorization(Ceed ceed, CeedScalar *mat, CeedScalar *tau, CeedInt m, CeedInt n);
 CEED_EXTERN int CeedSymmetricSchurDecomposition(Ceed ceed, CeedScalar *mat, CeedScalar *lambda, CeedInt n);
 CEED_EXTERN int CeedSimultaneousDiagonalization(Ceed ceed, CeedScalar *mat_A, CeedScalar *mat_B, CeedScalar *x, CeedScalar *lambda, CeedInt n);
+CEED_EXTERN int CeedHouseholderApplyQ(CeedScalar *A, const CeedScalar *Q, const CeedScalar *tau, CeedTransposeMode t_mode, CeedInt m, CeedInt n,
+                                      CeedInt k, CeedInt row, CeedInt col);
+CEED_EXTERN int CeedMatrixMatrixMultiply(Ceed ceed, const CeedScalar *mat_A, const CeedScalar *mat_B, CeedScalar *mat_C, CeedInt m, CeedInt n,
+                                         CeedInt kk);
```
jeremylt, line 397: Why the move? These aren't optimized and not intended for external users

interface/ceed-basis.c
```diff
@@ -429,27 +396,28 @@ int CeedBasisGetFlopsEstimate(CeedBasis basis, CeedTransposeMode t_mode, CeedEva
...
-    CeedCall(CeedBasisGetNumQuadratureComponents(basis, &Q_comp));
+    CeedFESpace fe_space;
+    CeedCall(CeedBasisGetFESpace(basis, &fe_space));
     switch (eval_mode) {
       case CEED_EVAL_NONE:
         *flops = 0;
         break;
       case CEED_EVAL_INTERP:
-        *flops = num_nodes * num_qpts * num_comp;
+        *flops = num_nodes * num_qpts * num_comp * (fe_space == CEED_FE_SPACE_H1 ? 1 : dim);
```
jeremylt, line 411: It'd be nice to keep these more generic calculations if we can

interface/ceed-basis.c
```diff
@@ -883,6 +836,86 @@ int CeedBasisCreateHdiv(Ceed ceed, CeedElemTopology topo, CeedInt num_comp, Ceed
...
+  (*basis)->ceed = ceed;
+  CeedCall(CeedReference(ceed));
+  (*basis)->ref_count    = 1;
+  (*basis)->tensor_basis = 0;
+  (*basis)->dim          = dim;
+  (*basis)->topo         = topo;
+  (*basis)->num_comp     = num_comp;
+  (*basis)->P            = P;
+  (*basis)->Q            = Q;
+  (*basis)->fe_space     = 3;  // H(curl) space
```
jeremylt, line 906: Should use the Enum value here instead of hard-coding the magic number.

interface/ceed-operator.c
```diff
@@ -61,7 +62,9 @@ static int CeedOperatorCheckField(Ceed ceed, CeedQFunctionField qf_field, CeedEl
     }
     CeedCall(CeedBasisGetDimension(b, &dim));
     CeedCall(CeedBasisGetNumComponents(b, &num_comp));
-    CeedCall(CeedBasisGetNumQuadratureComponents(b, &Q_comp));
+    CeedCall(CeedBasisGetFESpace(b, &fe_space));
```
jeremylt, line 65: Same here, would be good to actually use something more general like `Q_comp` that was here. And new variables should follow the naming contentions of 1) snake case and 2) using full words as much as reasonable

backends/ref/ceed-ref-basis.c
sebastiangrimberg: `c_dim` is the dimension of the curl, which for 1D and 3D elements is equal to `c_dim = dim` but for 2D elements it is 1D instead of 2D (normal to the reference element plane).

My approach was based on not wanting to keep adding members to the `struct CeedBasis_private` and instead just use information that was already there in the `CeedFESpace` member. I can add back a `CeedBasisGetQuadratureDimension` which returns `q_dim = (fe_space == CEED_FE_SPACE_H1) ? 1 : dim;` and `CeedBasisGetCurlDimension` which returns `c_dim = (dim < 3) ? 1 : dim;` if that's preferable.

As for the naming of `Q_comp` -> `q_dim`, I thought this was just nicer for consistency with `dim` and `c_dim`. I could just as easily have it by `q_comp` and `c_comp` or `curl_comp`, just let me know what you prefer.

include/ceed/ceed.h
sebastiangrimberg: Yeah sorry about this and I'm happy to move them back, this is an artifact of familiarizing myself with the library. I was just trying to make sense of where these methods were being used and defined in `ceed-basis.c`, and thought they could just go alongside the other utility linear algebra methods which these methods which were declared in `ceed.h` as opposed to `backend.h`.

backends/ref/ceed-ref-basis.c
jeremylt: This ultimately calculation needs to go into at 6 other backends and its used a few places in the interface, so I think its better to have a single source of truth for the calculation.

I'm not an expert in HDiv or HCurl elements, but I was under the impression that the number of components changed, not the dimension of the quadrature space.

include/ceed/ceed.h
sebastiangrimberg: Reverted in [063bea1](https://github.com/CEED/libCEED/pull/1156/commits/063bea108b882e83a5ae894d5a49e974101cbd30).

interface/ceed-basis.c
sebastiangrimberg: Resolved in [063bea1](https://github.com/CEED/libCEED/pull/1156/commits/063bea108b882e83a5ae894d5a49e974101cbd30).

interface/ceed-operator.c
sebastiangrimberg: Resolved in [063bea1](https://github.com/CEED/libCEED/pull/1156/commits/063bea108b882e83a5ae894d5a49e974101cbd30).

interface/ceed-basis.c
sebastiangrimberg: Resolved in [063bea1](https://github.com/CEED/libCEED/pull/1156/commits/063bea108b882e83a5ae894d5a49e974101cbd30).

backends/ref/ceed-ref-basis.c
sebastiangrimberg: See [063bea1](https://github.com/CEED/libCEED/pull/1156/commits/063bea108b882e83a5ae894d5a49e974101cbd30).

backends/ref/ceed-ref-basis.c
```diff
@@ -194,11 +195,12 @@ static int CeedBasisApply_Ref(CeedBasis basis, CeedInt num_elem, CeedTransposeMo
     switch (eval_mode) {
       // Interpolate to/from quadrature points
       case CEED_EVAL_INTERP: {
-        CeedInt           P = num_nodes, Q = Q_comp * num_qpts;
+        CeedInt           qdim = (fe_space == CEED_FE_SPACE_H1) ? 1 : dim;
```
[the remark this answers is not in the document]
sebastiangrimberg: Resolved in [063bea1](https://github.com/CEED/libCEED/pull/1156/commits/063bea108b882e83a5ae894d5a49e974101cbd30).

backends/ref/ceed-ref-basis.c
rezgarshakeri: I prefer to use `Q_comp` and `Curl_comp`. @jedbrown what do you prefer?

sebastiangrimberg: Note that this is `q_comp` and `curl_comp` in the most recent commit 063ea1 to match cases better, but happy to adjust to whatever is desired.

rezgarshakeri: That looks good to me, thanks.

tests/t332-basis.c
```diff
@@ -1,37 +1,38 @@
...
-  CeedInt           P        = dim * num_nodes;  // dof per element!
+  const CeedInt     Q = 3, dim = 2, num_qpts = Q * Q, elem_nodes = 4;
+  const CeedInt     P = dim * elem_nodes;
   CeedBasis         b;
   CeedScalar        q_ref[dim * num_qpts], q_weights[num_qpts];
-  CeedScalar        div[P * num_qpts], interp[P * dim * num_qpts];
+  CeedScalar        interp[dim * P * num_qpts], div[P * num_qpts];
   CeedVector        X, Y;
-  const CeedScalar *y, *x;
+  const CeedScalar *y, *x, *div2;
```
jeremylt, line 17:
```suggestion
  const CeedScalar *y, *x, *div_in_basis;
```
maybe? Open to other names, the 2 suffix is just not as clear

tests/t341-basis.c
```diff
@@ -0,0 +1,73 @@
...
+
+int main(int argc, char **argv) {
+  Ceed              ceed;
+  const CeedInt     P = 8, Q = 4, dim = 2;
+  CeedBasis         b;
+  CeedScalar        q_ref[dim * Q], q_weight[Q];
+  CeedScalar        interp[dim * P * Q], curl[P * Q];
+  CeedVector        X, Y;
+  const CeedScalar *y, *x, *interp2;
+  CeedScalar        rowsum[dim * Q], colsum[P];
```
jeremylt, line 17:
```suggestion
  CeedScalar        row_sum[dim * Q], column_sum[P];
```

tests/t342-basis.c
```diff
@@ -0,0 +1,73 @@
...
+#include "t340-basis.h"
+
+int main(int argc, char **argv) {
+  Ceed              ceed;
+  const CeedInt     P = 8, Q = 4, dim = 2;
+  CeedBasis         b;
+  CeedScalar        q_ref[dim * Q], q_weight[Q];
+  CeedScalar        interp[dim * P * Q], curl[P * Q];
+  CeedVector        X, Y;
+  const CeedScalar *y, *x, *curl2;
```
jeremylt, line 16:
```suggestion
  const CeedScalar *y, *x, *curl_in_basis;
```

jeremylt (commented):
Re some of your points at the top

1) I'm fine adding GPU support in a follow-up PR. That way this PR won't get too big. Adding initial `[hip,cuda]/ref` support won't be too bad, but `[hip,cuda]/[shared,gen]` is related to the other issue you pointed out

2) Similarly, I think its good to add Diagonal assembly support in a follow-up PR. The multigrid infrastructure should be largely unaffected, though I'm not sure if our process for automatically building a coarse to fine interpolation works for H(div) and H(curl) elements?

3) Python, Rust, and Julia bindings would be nice. Our Fortran interface is not a high priority but I wouldn't object to adding this there as well. Let me know if you get stuck trying to add any of those and I can help.

tests/t342-basis.c
sebastiangrimberg: [2a6c126](https://github.com/CEED/libCEED/pull/1156/commits/2a6c12603dc224f416582469a85baf5764d87787)

tests/t341-basis.c
sebastiangrimberg: [2a6c126](https://github.com/CEED/libCEED/pull/1156/commits/2a6c12603dc224f416582469a85baf5764d87787)

tests/t332-basis.c
sebastiangrimberg: [2a6c126](https://github.com/CEED/libCEED/pull/1156/commits/2a6c12603dc224f416582469a85baf5764d87787)

sebastiangrimberg:
> Re some of your points at the top
> 
> 1. I'm fine adding GPU support in a follow-up PR. That way this PR won't get too big. Adding initial `[hip,cuda]/ref` support won't be too bad, but `[hip,cuda]/[shared,gen]` is related to the other issue you pointed out

OK, sounds good. I am very unfamiliar with what will be required here, but I don't think it will be overly complex because H(div)/H(curl) interpolation contractions should look the same (or nearly) to H1 gradient.

> 2. Similarly, I think its good to add Diagonal assembly support in a follow-up PR. The multigrid infrastructure should be largely unaffected, though I'm not sure if our process for automatically building a coarse to fine interpolation works for H(div) and H(curl) elements?

For multigrid, the issue I see is that `CeedOperatorMultigridLevelCreate` calls `CeedBasisCreateProjection` when prolongation or restriction operators are requested. This method I think needs fixing for H(div) and H(curl) interpolation.

> 3. Python, Rust, and Julia bindings would be nice. Our Fortran interface is not a high priority but I wouldn't object to adding this there as well. Let me know if you get stuck trying to add any of those and I can help.

I see for Julia that there is an automated generator method in `julia/libCEED.jl/gen` which makes this easy. Is there something similar for Python and Rust or do I have to do it by hand?

jeremylt:
The raw bindings are automatic but the user facing bindings need to be hand-coded in all languages

sebastiangrimberg:
Hi @jedbrown @jeremylt, I just pushed a few commits with a bit of a reworking of this idea. I realized that a general implementation for non-tensor H^1, H(div) and H(curl) basis operations should be possible with the benefit of simplifying backend code a lot. The general idea is that `CEED_EVAL_INTERP` in H(div) and H(curl) looks exactly like `CEED_EVAL_GRAD` for H^1, H(div) `CEED_EVAL_DIV` looks like H^1 `CEED_EVAL_INTERP`, and `CEED_EVAL_CURL` looks again looks like H^1 `CEED_EVAL_GRAD`(in 3D at least). So, I just tried to use the `CeedBasisGetQuadratureNumComponents` more extensively to make all of these operations look the same, with the output of `CeedBasisApply` being a `q_comp x num_comp x num_qpts x num_elems` tensor (where `q_comp = 1` for H^1 interpolation, for example). For now you can check out the CPU implementation in `CeedBasisApply_Ref` and I'd be curious your thoughts.

In this way, the non-tensor `cuda-ref` and `hip-ref` backends should be very easy to port to H(div) and H(curl) with only some minor refactoring, and most of the methods in `ceed-preconditioning.c` should also require only a bit of modification to work with H(div) and H(curl).

rezgarshakeri:
> Hi @jedbrown @jeremylt, I just pushed a few commits with a bit of a reworking of this idea. I realized that a general implementation for non-tensor H^1, H(div) and H(curl) basis operations should be possible with the benefit of simplifying backend code a lot. The general idea is that `CEED_EVAL_INTERP` in H(div) and H(curl) looks exactly like `CEED_EVAL_GRAD` for H^1, H(div) `CEED_EVAL_DIV` looks like H^1 `CEED_EVAL_INTERP`, and `CEED_EVAL_CURL` looks again looks like H^1 `CEED_EVAL_GRAD`(in 3D at least). So, I just tried to use the `CeedBasisGetQuadratureNumComponents` more extensively to make all of these operations look the same, with the output of `CeedBasisApply` being a `q_comp x num_comp x num_qpts x num_elems` tensor (where `q_comp = 1` for H^1 interpolation, for example). For now you can check out the CPU implementation in `CeedBasisApply_Ref` and I'd be curious your thoughts.
> 
> In this way, the non-tensor `cuda-ref` and `hip-ref` backends should be very easy to port to H(div) and H(curl) with only some minor refactoring, and most of the methods in `ceed-preconditioning.c` should also require only a bit of modification to work with H(div) and H(curl).

Not always. When the problem is scalar in H^1, yes `CEED_EVAL_GRAD` is similar to `CEED_EVAL_INTERP` in H(div).

sebastiangrimberg:
> Not always. When the problem is scalar in H^1, yes `CEED_EVAL_GRAD` is similar to `CEED_EVAL_INTERP` in H(div).

Yes that's correct in general. My point was simply that with `num_comp = 1` we should be able to reuse much of the H^1 gradient kernels (for example) and shouldn't need to write much code nor sacrifice performance.

rezgarshakeri:
> > Not always. When the problem is scalar in H^1, yes `CEED_EVAL_GRAD` is similar to `CEED_EVAL_INTERP` in H(div).
> 
> Yep that's correct. My point was simply that with `num_comp = 1` we should be able to reuse much of the H^1 gradient kernels (for example) and shouldn't need to write much code nor sacrifice performance.

Yeah, I got your point, I just wanted to mentioned that is true for scalar case and not using it otherwise.

include/ceed/backend.h
```diff
@@ -173,29 +173,33 @@ CEED_EXTERN int CeedElemRestrictionGetFlopsEstimate(CeedElemRestriction rstr, Ce
...
+CEED_EXTERN int CeedBasisGetFESpace(CeedBasis basis, CeedFESpace *fe_space);
 CEED_EXTERN int CeedBasisGetTopologyDimension(CeedElemTopology topo, CeedInt *dim);
-
 CEED_EXTERN int CeedBasisGetTensorContract(CeedBasis basis, CeedTensorContract *contract);
 CEED_EXTERN int CeedBasisSetTensorContract(CeedBasis basis, CeedTensorContract contract);
+
 CEED_EXTERN int CeedTensorContractCreate(Ceed ceed, CeedBasis basis, CeedTensorContract *contract);
 CEED_EXTERN int CeedTensorContractApply(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *__restrict__ t,
                                         CeedTransposeMode t_mode, const CeedInt Add, const CeedScalar *__restrict__ u, CeedScalar *__restrict__ v);
+CEED_EXTERN int CeedTensorContractStridedApply(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt D, CeedInt J,
```
jeremylt, line 200: Should be CEED_INTERN

include/ceed/backend.h
```diff
@@ -280,7 +284,7 @@ CEED_EXTERN int CeedOperatorSetData(CeedOperator op, void *data);
 CEED_EXTERN int CeedOperatorReference(CeedOperator op);
 CEED_EXTERN int CeedOperatorSetSetupDone(CeedOperator op);
 
-CEED_INTERN int CeedMatrixMatrixMultiply(Ceed ceed, const CeedScalar *mat_A, const CeedScalar *mat_B, CeedScalar *mat_C, CeedInt m, CeedInt n,
+CEED_EXTERN int CeedMatrixMatrixMultiply(Ceed ceed, const CeedScalar *mat_A, const CeedScalar *mat_B, CeedScalar *mat_C, CeedInt m, CeedInt n,
```
jeremylt, line 287: Should be CEED_INTERN (in backend.h)

include/ceed/backend.h
sebastiangrimberg: This was something I was wondering about. Why are all the other functions in `backend.h` not `CEED_INTERN` as well?

jeremylt: Woops, this file is CEED_EXTERN. nevermind

include/ceed/ceed.h
```diff
@@ -382,11 +383,14 @@ CEED_EXTERN int CeedBasisGetInterp1D(CeedBasis basis, const CeedScalar **interp_
 CEED_EXTERN int CeedBasisGetGrad(CeedBasis basis, const CeedScalar **grad);
 CEED_EXTERN int CeedBasisGetGrad1D(CeedBasis basis, const CeedScalar **grad_1d);
 CEED_EXTERN int CeedBasisGetDiv(CeedBasis basis, const CeedScalar **div);
+CEED_EXTERN int CeedBasisGetCurl(CeedBasis basis, const CeedScalar **curl);
 CEED_EXTERN int CeedBasisDestroy(CeedBasis *basis);
 
 CEED_EXTERN int CeedGaussQuadrature(CeedInt Q, CeedScalar *q_ref_1d, CeedScalar *q_weight_1d);
 CEED_EXTERN int CeedLobattoQuadrature(CeedInt Q, CeedScalar *q_ref_1d, CeedScalar *q_weight_1d);
 CEED_EXTERN int CeedQRFactorization(Ceed ceed, CeedScalar *mat, CeedScalar *tau, CeedInt m, CeedInt n);
+CEED_EXTERN int CeedHouseholderApplyQ(CeedScalar *mat, const CeedScalar *Q, const CeedScalar *tau, CeedTransposeMode t_mode, CeedInt m, CeedInt n,
```
jeremylt, line 392: These purely linear algebra functions should probably not be here. I can fix that in a separate PR though.

include/ceed/backend.h
sebastiangrimberg: Ah, got it. That makes more sense that it's file-by-file. Thanks!

jeremylt: Its a weird case where these are intended for purely internal use, but  sometimes consumers of libCEED need just a little bit more options that the standard public header offers

include/ceed/ceed.h
sebastiangrimberg: That's what I figured, but I thought better at least to be all together. I can make a separate MR to move them to backend.h if you want?

jeremylt: Go for it. Some tests will need the `ceed/backend.h` header to compile, but that's not a big fix

sebastiangrimberg: https://github.com/CEED/libCEED/pull/1165

backends/ref/ceed-ref-basis.c
```diff
@@ -191,36 +191,31 @@ static int CeedBasisApply_Ref(CeedBasis basis, CeedInt num_elem, CeedTransposeMo
     }
   } else {
     // Non-tensor basis
+    CeedInt P = num_nodes, Q = num_qpts;
```
nbeams, line 194: I'm curious about this change: how was this able to be pulled out?  Previously, Q was defined differently for `CEED_EVAL_INTERP` (Q_comp * num_qpts).  Also, how do you avoid needing to swap P and Q based on `t_mode`?

sebastiangrimberg: Hi @nbeams, if you look at the new `CeedTensorContractStridedApply` function, this handles both of the comments you have (swaps `P` and `Q` for `t_mode` and also handles `q_comp`). This was introduced to avoid duplicating the same code for all of the `eval_mode` cases.

nbeams: Oh, ha, I didn't even notice the name change 🤦‍♀️ .  But I'm still not sure how `INTERP` and `GRAD` could have the exact same arguments (except for the basis matrix `interp` or `grad`) given to the new function.  How does it know that the gradient matrix is of size `P*dim x Q` while `interp` is `P x Q`?

sebastiangrimberg: For H^1 `INTERP`, `q_comp = 1` versus for `GRAD`, `q_comp = dim`. This for H^1 `INTERP` the "strided apply" doesn't really do anything over the `q_comp` direction. The idea here was to just generalize to a basis tensor of shape `q_comp x num_qpts x num_nodes` to capture both the H^1 `INTERP` and `GRAD` operations, as well as the various shapes of H(div) and H(curl) basis operations.

nbeams: Oh, I see, `q_comp` is no longer a property of the basis object.  I'm not sure how I feel about that change, but if other developers like it, then I won't object. 
The cuda-ref and hip-ref (and MAGMA) non-tensor basis implementations don't use the CeedTensorContractApply setup, but have specialized kernels, so I'm not sure how this refactoring will really help them, but it obviously doesn't hurt anything to have it in ref.

sebastiangrimberg:
I cannot for the life of me figure out what's going on with the Julia API after trying to debug the segfault showing up in the tests. If anyone has any suggestions I'd be very happy to try them out.

jeremylt:
> It looks like the tests were being run twice (once with the development libCEED binary, once with the registered JLL binary)

Not quite. Or at least, that's not the intent and shouldn't be what's happening.

We sometimes make breaking changes that won't be in the Julia package until we make a new release. One run of tests makes sure changes in core libCEED don't accidentally break the current Julia bindings. The second run of tests checks that any new, unreleased bindings function correctly.

Any new or changed tests must go into `julia/LibCEED.jl/rundevtests.jl`

rust/libceed/src/basis.rs
```diff
@@ -238,6 +238,84 @@ impl<'a> Basis<'a> {
         })
     }
 
+    pub fn create_Hdiv(
```
jeremylt, line 241: There need to be corresponding functions and docstrings in the sibling file `lib.rs`

python/ceed_basis.py
```diff
@@ -339,6 +339,88 @@ def __init__(self, ceed, topo, ncomp, nnodes,
 # ------------------------------------------------------------------------------
 
 
+class BasisHdiv(Basis):
```
jeremylt, line 342: Corresponding tests go in `python/tests/test-3-basis.py`

include/ceed/jit-source/cuda/cuda-ref-operator-assemble-diagonal.h
```diff
@@ -10,7 +10,6 @@
 //------------------------------------------------------------------------------
 // Diagonal assembly kernels
 //------------------------------------------------------------------------------
-
```
jeremylt, line 13: This enum probably shouldn't be here. Let me make a quick PR to fix

sebastiangrimberg:
> > It looks like the tests were being run twice (once with the development libCEED binary, once with the registered JLL binary)
> 
> Not quite. Or at least, that's not the intent and shouldn't be what's happening.
> 
> We sometimes make breaking changes that won't be in the Julia package until we make a new release. One run of tests makes sure changes in core libCEED don't accidentally break the current Julia bindings. The second run of tests checks that any new, unreleased bindings function correctly.
> 
> Any new or changed tests must go into `julia/LibCEED.jl/rundevtests.jl`

I think I understand and perhaps to check my understanding I can explain what I think I was seeing happen before:

1.  ```
    echo >> Project.toml
    echo "[preferences.libCEED_jll]" >> Project.toml
    echo "libceed_path = \"$LIBCEED_LIB\"" >> Project.toml
    [[ "$GITHUB_REF" =~ ^refs/(heads/release|tags/).* ]] || julia --project -e 'import Pkg; Pkg.test("LibCEED"; coverage=true, test_args=["--run-dev-tests"])'
    ```
    These tests run the new LibCEED.jl Julia bindings using the new core libCEED library. They were passing fine because the bindings were updated correctly to match the new library, no new tests were added.

2. `git checkout Project.toml && julia --project -e 'import Pkg; Pkg.test("LibCEED")'`
    These tests run with the new LibCEED.jl bindings again, but this time using the registered libCEED_jll packaged binary, which was failing. I think this is expected to fail as the new bindings use new functions from the underlying library (in this case, the definition of `getinterp` changed). I imagine that is not the intention, right? If by "current Julia bindings" you mean the registered LibCEED.jl package, then I think we need to run the tests from a new clean environment using the registered LibCEED.jl but keeping the `preferences.libCEED_jll` line to make it use the local build. Something like:
    ```
    julia --project=../test -e 'import Pkg; Pkg.add("LibCEED")'
    echo >> ../test/Project.toml
    echo "[preferences.libCEED_jll]" >> ../test/Project.toml
    echo "libceed_path = \"$LIBCEED_LIB\"" >> ../test/Project.toml
    julia --project=../test -e 'import Pkg; Pkg.test("LibCEED")'
    ```
    This would ensure the new core library code did not break anything in the registered Julia bindings package.

Do I have the right idea here? I would indeed expect that there will be cases where breaking changes in the libCEED core library will break the registered bindings, and that's where version compatibility constraints come in. So I'm not entirely sure I agree that the second test is necessary but my job here is just to make sure you are happy.

include/ceed/jit-source/cuda/cuda-ref-operator-assemble-diagonal.h
sebastiangrimberg: Thanks, indeed it looks like I missed editing the same duplicated comments in `hip-ref-operator-assemble-diagonal.h` anyway.

jeremylt:
Pinging @pazner as this piece he knows best

rust/libceed/src/basis.rs
sebastiangrimberg: Fixed in [9f5fbf2](https://github.com/CEED/libCEED/pull/1156/commits/9f5fbf2d1bf33955b5a7006814dfefab7b7903f2).

rust/libceed/src/lib.rs
```diff
@@ -707,6 +707,71 @@ impl Ceed {
...
+    /// * `nnodes`  - Total number of nodes
+    /// * `nqpts`   - Total number of quadrature points
+    /// * `interp`  - Row-major `(nqpts * dim * nnodes)` matrix expressing the
+    ///                 values of basis functions at quadrature points
+    /// * `div`     - Row-major `(nqpts * nnodes)` matrix expressing the
+    ///                 divergence of basis functions at quadrature points
+    /// * `qref`    - Array of length `nqpts` holding the locations of quadrature
+    ///                 points on the reference element `[-1, 1]`
+    /// * `qweight` - Array of length `nqpts` holding the quadrature weights on
+    ///                 the reference element
```
jeremylt, line 725: Can we add tests too to verify that the bindings function as intended?

sebastiangrimberg: I imagine you are also looking for Python tests and I should just duplicate the existing simplex H(div) and H(curl) C tests there?

pazner:
> Pinging @pazner as this piece he knows best

I'm just catching up on this discussion, so apologies if I haven't quite understood all the context.

I think you're both largely correct about the purpose of the Julia tests (and why they are run twice).

Most users of the Julia library will use the prebuilt library provided in libCEED_jll. The tests are run using this library to ensure that it is not broken. The tests are _also_ run using the library built from the current source, to make sure that changes in the C library code don't break the Julia bindings.

In the case of this PR, it sounds like there were breaking changes (and the Julia bindings were updated) that mean that the Julia interface is no longer compatible with the old version of the C library. To indicate this, you need to bump the CEED version number (to indicate a breaking change) and bump the required version for Julia compatibility:
```diff
diff --git a/include/ceed/ceed.h b/include/ceed/ceed.h
index a5f1ac15..384440f7 100644
--- a/include/ceed/ceed.h
+++ b/include/ceed/ceed.h
@@ -135,9 +135,9 @@ CEED_EXTERN int CeedResetErrorMessage(Ceed, const char **err_msg);
 /// libCEED library version numbering
 /// @ingroup Ceed
 #define CEED_VERSION_MAJOR 0
-#define CEED_VERSION_MINOR 11
+#define CEED_VERSION_MINOR 12
 #define CEED_VERSION_PATCH 0
-#define CEED_VERSION_RELEASE true
+#define CEED_VERSION_RELEASE false
```
and
```diff
diff --git a/julia/LibCEED.jl/src/LibCEED.jl b/julia/LibCEED.jl/src/LibCEED.jl
index aaa8d5d8..243415c6 100644
--- a/julia/LibCEED.jl/src/LibCEED.jl
+++ b/julia/LibCEED.jl/src/LibCEED.jl
@@ -152,3 +152,3 @@ include("Misc.jl")
 
-const minimum_libceed_version = v"0.10.0"
+const minimum_libceed_version = v"0.12.0"
```

This should then tell LibCEED.jl to skip the tests when using libCEED_jll:
```
┌ Warning: Skipping tests because of incompatible libCEED versions.
```

jeremylt:
I don't think we are ready to increment the minor version to 12 yet but main being marked as release is a bug. Can fix tomorrow

pazner:
> I don't think we are ready to increment the minor version to 12 yet but main being marked as release is a bug. Can fix tomorrow.

Yeah, it should be enough to change `#define CEED_VERSION_RELEASE false` in this PR, since "a non-release version always compares as positive infinity".

The `minimum_libceed_version` will still need to be incremented to some value strictly greater than `v0.11.0` so that it knows that the libCEED_jll version is incompatible.

.github/workflows/julia-test-with-style.yml
```diff
@@ -33,9 +33,9 @@ jobs:
...
           pushd julia/LibCEED.jl
-          echo >> test/Project.toml
-          echo "[preferences.libCEED_jll]" >> test/Project.toml
-          echo "libceed_path = \"$LIBCEED_LIB\"" >> test/Project.toml
+          echo >> Project.toml
+          echo "[preferences.libCEED_jll]" >> Project.toml
+          echo "libceed_path = \"$LIBCEED_LIB\"" >> Project.toml
           [[ "$GITHUB_REF" =~ ^refs/(heads/release|tags/).* ]] || julia --project -e 'import Pkg; Pkg.test("LibCEED"; coverage=true, test_args=["--run-dev-tests"])'
-          git checkout test/Project.toml && julia --project -e 'import Pkg; Pkg.test("LibCEED")'
+          git checkout Project.toml && julia --project -e 'import Pkg; Pkg.test("LibCEED")'
```
pazner, line 40: Are these changes still needed?

sebastiangrimberg: Nope! Fixed in 833e5c2. Thanks @pazner  and @jeremylt!

include/ceed/jit-source/cuda/cuda-ref-operator-assemble-diagonal.h
jeremylt: Change merged to main

jeremylt (commented):
I think the only thing that we need before merging is `create_hdiv_basis` and `create_hcurl_basis` tests for `julia/LibCEED.jl/test/rundevtests.jl`. Currently there are no dev tests, but there are some examples for `create_h1_basis` in ``julia/LibCEED.jl/test/runtests.jl`.

Everything else seems to be all set.

sebastiangrimberg:
> I think the only thing that we need before merging is `create_hdiv_basis` and `create_hcurl_basis` tests for `julia/LibCEED.jl/test/rundevtests.jl`. Currently there are no dev tests, but there are some examples for `create_h1_basis` in ``julia/LibCEED.jl/test/runtests.jl`.
> 
> Everything else seems to be all set.

Done! See [97c1c57](https://github.com/CEED/libCEED/pull/1156/commits/97c1c57a2210bceee970d471329a94c21686be34).

jeremylt (approved):
LGTM

PR #1157, opened 2023-02-15
ci: upgrade to clang-format-15

jedbrown:


jeremylt:
Does Noether have clang-format 15?

jedbrown:
Does now.

PR #1158, opened 2023-02-15
Fix for cubic elements in surface forces for lift/drag coefficients

AdelekeBankole:


examples/fluids/src/setupts.c
```diff
@@ -164,6 +166,7 @@ static PetscErrorCode Surface_Forces_NS(DM dm, Vec G_loc, PetscInt num_walls, co
         const PetscInt           p = points[i];
         const StateConservative *r;
         PetscCall(DMPlexPointLocalRead(dm, p, g, &r));
+        PetscCall(PetscSectionGetDof(s, p, &dof));
```
jedbrown, line 169: Now you know how many dofs, but still need to loop over the nodes.

jedbrown: Once you're pretty sure it's working, can you run a coarse mesh with cubic elements to produce a new lift and drag plot?

AdelekeBankole: Yes gladly! I will produce a new lift/drag plot. Thanks.

jedbrown: Great. You have to fix this bit first (or the results will be quite wrong). Each node has 5 dofs, so loop over the nodes at each point and read out their momentum residuals. The code here reads only the first node, which is correct at vertices, but is only part of a (cubic) edge or face.

examples/fluids/src/setupts.c
```diff
@@ -164,6 +166,10 @@ static PetscErrorCode Surface_Forces_NS(DM dm, Vec G_loc, PetscInt num_walls, co
         const PetscInt           p = points[i];
         const StateConservative *r;
         PetscCall(DMPlexPointLocalRead(dm, p, g, &r));
+        for (PetscInt k = 0; k < num_points; k++) {
+          PetscCall(PetscSectionGetDof(s, points[k], &dof));  // there are 5 dofs
+          // printf("points = %d, num_points = %d\n", points[k], num_points);
+        }
```
jedbrown, line 172: Let's think about what's happening here. `num_points` is the global number of topological points on the wall and we're already looping over those with `i`. The inner loop we need is over the nodes on point `p`. You could `for (PetscInt node=0; node<dof / 5; node++)` and read `r[node].momentum[j]`.

This could be a bit more defensive by checking `PetscSectionGetFieldComponents(s, 0, &num_comp)` to avoid the magic number 5 and help get it correct when we add turbulence variables.

AdelekeBankole: @jedbrown, actually in the loop, sometimes ```dof``` is ```2``` but ```num_comp``` is 5. Please, could you explain why you did this ```dof / 5``` ?

jedbrown: I don't think it should be. There are only 2 global dofs, but the local section should say there are 5 dofs, 3 of which are constrained (or 10 dofs, 6 of which are constrained). Can you give precise instructions to reproduce that `dof = 2`?

AdelekeBankole: Yeah it says 5 dofs but for example if I print ```dof``` before line
```for (PetscInt node = 0; node < dof / num_comp; node++)```
then, it says 2 and 5 (but mostly 5).

File: examples/fluids/src/setupts.c [hunk 3]
```diff
@@ -154,19 +155,25 @@ static PetscErrorCode Surface_Forces_NS(DM dm, Vec G_loc, PetscInt num_walls, co
...
       for (PetscInt i = 0; i < num_points; i++) {
         const PetscInt           p = points[i];
         const StateConservative *r;
         PetscCall(DMPlexPointLocalRead(dm, p, g, &r));
-        if (!r) continue;
-        for (PetscInt j = 0; j < 3; j++) {
-          reaction_force[w * dim + j] -= r->momentum[j];
+        PetscCall(PetscSectionGetFieldComponents(s, 0, &num_comp));
+        for (PetscInt node = 0; node < dof / num_comp; node++) {
+          PetscCall(PetscSectionGetDof(s, points[node], &dof));
```
jedbrown, line 172:
```suggestion
        PetscCall(PetscSectionGetDof(s, p, &dof));
        for (PetscInt node = 0; node < dof / num_comp; node++) {
```

examples/fluids/src/setupts.c
```diff
@@ -154,19 +155,25 @@ static PetscErrorCode Surface_Forces_NS(DM dm, Vec G_loc, PetscInt num_walls, co
...
+          if (!r) continue;
```
jedbrown, line 173: Shouldn't need this any more since you're counting dofs.
```suggestion
```

jedbrown:
@AdelekeBankole I think this is ready to merge, I'd just like to see a lift and drag plot for cubic elements. Can you do that run?

AdelekeBankole:
I will do that run now. I was just having some laptop issues this morning.

jedbrown:
Here's a plot running with Q3 elements on a `cylinder-q1-n08.msh`. You only need to update `vortexshedding.yaml` to use this mesh and degree 3. The question here is what is "correct" (converged) for this flow.

vortexshedding
image

AdelekeBankole:
@jedbrown, thank you! Now, I can run with ```mpiexec``` on my computer, without regurgitating the petsc errors I was having yesterday. I was wondering aside from the update you made in the make file and the .geo file, did you fix any other thing in petsc I am not aware of? Whats the connection among these ```mpiexec, makefile and .geo``` I cant understand why yet

jedbrown:
No changes in PETSc, but I'm using the new coarse mesh. The old output CGNS name is not consistent with the input mesh name (just confusing to name in that way).
```diff
diff --git i/examples/fluids/vortexshedding.yaml w/examples/fluids/vortexshedding.yaml
index 1859c388b..4f5494b21 100644
--- i/examples/fluids/vortexshedding.yaml
+++ w/examples/fluids/vortexshedding.yaml
@@ -12,7 +12,7 @@ ts:
   dt: .05
   max_time: 100
   alpha_radius: 0.5
-  monitor_solution: cgns:vortexshedding-q2-g2-n12-recirc.cgns
+  monitor_solution: cgns:vortexshedding-q3-g1-n08.cgns
   monitor_solution_interval: 5
   monitor_wall_force: ascii:force.csv:ascii_csv

@@ -40,8 +40,8 @@ mu: 0.01
 k: 14.34 # thermal conductivity, Pr = 0.71 typical of air

 ## DM Settings:
-degree: 1
-dm_plex_filename: examples/fluids/meshes/cylinder-q1-n20.msh
+degree: 3
+dm_plex_filename: examples/fluids/meshes/cylinder-q1-n08.msh

 # Boundary Settings
 bc_slip_z: 6
```

AdelekeBankole:
@jedbrown, I discovered the path is not pointing to the right file for python to run. I have pushed the fix.

AdelekeBankole:
@jedbrown, Please, what else do we need to be done in this PR for it to be merged?

jedbrown (approved):
We can squash-merge it since `main` is currently wrong when using cubic elements. For future work
- [ ] the script should take the csv file as an argument and write the output file as an argument. It should use argparse for this. You can see an argparse example in the perf-portable-solids paper repo to get a sense of how it's used for creating figures from batches of files.
- [ ] we need to do a grid refinement study to know what is correct for this domain.
- [ ] we should experiment with bringing the boundary condition in closer. We can compare the effect on lift and drag of using a freestream outflow versus the viscous Riemann outflow.
- [ ] we need to compare high order elements with/without high order geometry.
- [ ] once the problem is well understood (in an honestly 3D domain), we can make an accuracy versus cost diagram like we did in the solids paper.

jedbrown:
(You can make an issue or new draft pull request to work on the above.)

AdelekeBankole:
Thank you very much for these points. I will check the perf-portable-solids paper repo

PR #1164, opened 2023-02-21
Basis Eval at arbitrary points

jeremylt:
This PR adds the ability to interpolate the basis to arbitrary points:

- [x] tests
- [x] Chebyshev interpolating polynomials instead of Legendre
- [x] map from Chebyshev coefficients to arbitrary points
- [x] transpose direction
- [x] create internal `CeedBasis` with "interpolation" from nodes to coefficients (combine interp + map to coeffs to one set of contractions)

### DIFF
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -258,7 +258,6 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
   @ref Backend
 **/
 int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *collo_grad_1d) {
-  int         i, j, k;
   Ceed        ceed;
   CeedInt     P_1d = (basis)->P_1d, Q_1d = (basis)->Q_1d;
   CeedScalar *interp_1d, *grad_1d, *tau;
@@ -274,15 +273,15 @@ int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *collo_grad_1d) {
   CeedCall(CeedQRFactorization(ceed, interp_1d, tau, Q_1d, P_1d));
   // Note: This function is for backend use, so all errors are terminal and we do not need to clean up memory on failure.
 
-  // Apply Rinv, collo_grad_1d = grad_1d Rinv
-  for (i = 0; i < Q_1d; i++) {  // Row i
+  // Apply R_inv, collo_grad_1d = grad_1d R_inv
+  for (CeedInt i = 0; i < Q_1d; i++) {  // Row i
     collo_grad_1d[Q_1d * i] = grad_1d[P_1d * i] / interp_1d[0];
-    for (j = 1; j < P_1d; j++) {  // Column j
+    for (CeedInt j = 1; j < P_1d; j++) {  // Column j
       collo_grad_1d[j + Q_1d * i] = grad_1d[j + P_1d * i];
-      for (k = 0; k < j; k++) collo_grad_1d[j + Q_1d * i] -= interp_1d[j + P_1d * k] * collo_grad_1d[k + Q_1d * i];
+      for (CeedInt k = 0; k < j; k++) collo_grad_1d[j + Q_1d * i] -= interp_1d[j + P_1d * k] * collo_grad_1d[k + Q_1d * i];
       collo_grad_1d[j + Q_1d * i] /= interp_1d[j + P_1d * j];
     }
-    for (j = P_1d; j < Q_1d; j++) collo_grad_1d[j + Q_1d * i] = 0;
+    for (CeedInt j = P_1d; j < Q_1d; j++) collo_grad_1d[j + Q_1d * i] = 0;
   }
 
   // Apply Q^T, collo_grad_1d = collo_grad_1d Q^T
@@ -950,7 +949,7 @@ int CeedBasisCreateTensorH1(Ceed ceed, CeedInt dim, CeedInt num_comp, CeedInt P_
 **/
 int CeedBasisCreateTensorH1Lagrange(Ceed ceed, CeedInt dim, CeedInt num_comp, CeedInt P, CeedInt Q, CeedQuadMode quad_mode, CeedBasis *basis) {
   // Allocate
-  int        ierr = CEED_ERROR_SUCCESS, i, j, k;
+  int        ierr = CEED_ERROR_SUCCESS;
   CeedScalar c1, c2, c3, c4, dx, *nodes, *interp_1d, *grad_1d, *q_ref_1d, *q_weight_1d;
 
   CeedCheck(dim > 0, ceed, CEED_ERROR_DIMENSION, "Basis dimension must be a positive value");
@@ -977,15 +976,15 @@ int CeedBasisCreateTensorH1Lagrange(Ceed ceed, CeedInt dim, CeedInt num_comp, Ce
 
   // Build B, D matrix
   // Fornberg, 1998
-  for (i = 0; i < Q; i++) {
+  for (CeedInt i = 0; i < Q; i++) {
     c1                   = 1.0;
     c3                   = nodes[0] - q_ref_1d[i];
     interp_1d[i * P + 0] = 1.0;
-    for (j = 1; j < P; j++) {
+    for (CeedInt j = 1; j < P; j++) {
       c2 = 1.0;
       c4 = c3;
       c3 = nodes[j] - q_ref_1d[i];
-      for (k = 0; k < j; k++) {
+      for (CeedInt k = 0; k < j; k++) {
         dx = nodes[j] - nodes[k];
         c2 *= dx;
         if (k == j - 1) {
@@ -1353,9 +1352,7 @@ int CeedBasisApply(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode, 
   CeedCall(CeedBasisGetNumNodes(basis, &num_nodes));
   CeedCall(CeedBasisGetNumQuadraturePoints(basis, &num_qpts));
   CeedCall(CeedVectorGetLength(v, &v_length));
-  if (u) {
-    CeedCall(CeedVectorGetLength(u, &u_length));
-  }
+  if (u) CeedCall(CeedVectorGetLength(u, &u_length));
 
   CeedCheck(basis->Apply, basis->ceed, CEED_ERROR_UNSUPPORTED, "Backend does not support BasisApply");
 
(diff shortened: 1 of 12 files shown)


interface/ceed-basis.c
```diff
@@ -1085,6 +1085,29 @@ int CeedBasisApply(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode,
...
+  @param[in]  u        Input CeedVector, of length `num_points * num
+  @param[out] v        Output CeedVector, of length `num_points * num_comp` for `CEED_NOTRANSPOSE` with `CEED_EVAL_INTERP`
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMode t_mode, CeedEvalMode eval_mode, CeedVector x_ref, CeedVector u,
+                           CeedVector v) {
+  return CEED_ERROR_SUCCESS;
```
jedbrown, line 1108: I think `u` is in the FE basis (does not depend on `num_points`). We probably don't need `num_points` to be passed explicitly because it's implied by the shape of `x_ref` and `v`.

jeremylt: My thought here was that we could have a oversized vector that holds the maximum number of points for any element in the mesh rather than creating appropriately sized vectors for every element

jedbrown: That could work, but I don't know how to set a safe upper bound that isn't kinda huge. I'm not too concerned at this stage because we need to avoid premature optimization. At some point we'll have a spatial tree structure for locating points and will want it to run on the GPU. Pragmatically, I think we should just let it run on the CPU (even when doing implicit solves on GPU) through 2023 and plan to assess performance and discuss GPU-friendly data structures after that.

jeremylt: Is it premature optimization? We know that creating vectors on GPU backends is expensive because of the amount of JiT involved, and this happens even if only the CPU buffer is used.

To me this mirrors our passing of `num_elem` to `CeedBasisApply` even though we could probably guess the `num_elem` from the vector sizes passed.

jedbrown: What JiT is needed to create a vector? (I'm okay with passing this as an extra argument. Perhaps we should make a `ceed/experimental.h` header to make it clear that some new things can be expected to change.)

jeremylt: We ran into this with QFunction/Diagonal assembly, though we've changed somewhat how the GPU backends are organized so I'm trying to pin down where we would see that JiT now. It might have changed a bit but we were seeing the kernels for manipulating the data on the device had some sort of creation cost which is why we started caching those vectors.

jedbrown: It's not JIT level of cost, but the cost of allocating on device is higher than on CPU. We try to avoid any allocations in the critical path. One thing we could handle simply is make it possible for vector length to be less than allocated length, so that we could have a cheap resize operation.

interface/ceed-basis.c
```diff
@@ -1085,6 +1083,110 @@ int CeedBasisApply(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode,
...
+  // Default implementation
+  if (!basis->legendre_coeffs_1d) {
+    // Build matrix mapping from quadrature point values to Legendre coefficients
+  }
+  if (t_mode == CEED_NOTRANSPOSE) {
+    // Nodes to arbitrary points
+    CeedVector q;
+
+    // -- Interpolate to quadrature points
+    CeedCall(CeedVectorCreate(basis->ceed, basis->Q * basis->Q_comp, &q));
```
jeremylt, line 1170: Probably will want to cache this vector - GPU backends will do this work on the device

interface/ceed-basis.c
```diff
@@ -1393,6 +1391,203 @@ int CeedBasisApply(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode,
...
+
+        CeedCall(CeedVectorGetArrayRead(q, CEED_MEM_HOST, &q_array));
+        for (CeedInt d = 0; d < dim; d++) {
+          CeedCall(CeedTensorContractApply(basis->contract, pre, Q, post, Q, basis->legendre_coeffs_1d, CEED_NOTRANSPOSE, d > 0,
+                                           d == 0 ? q_array : tmp[d % 2], d == dim - 1 ? coeffs : tmp[(d + 1) % 2]));
+          pre /= Q;
+          post *= Q;
+        }
+        CeedCall(CeedVectorRestoreArrayRead(q, &q_array));
+      }
```
jeremylt, line 1540: I should replace this with a "basis" that just interpolates since I'm repeating logic here

jeremylt: @jedbrown feel free to fiddle with this branch while I'm gone if you are so inclined.

tests/t350-basis.c
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  // Interpolate to arbitrary points
+  CeedBasisCreateTensorH1Lagrange(ceed, 1, 1, p, q, CEED_GAUSS, &basis_u);
+  {
+    CeedScalar x_array[4] = {-0.33, -0.65, .16, 0.99};
+
+    CeedVectorSetArray(x_points, CEED_MEM_HOST, CEED_COPY_VALUES, x_array);
+  }
+  CeedBasisApplyAtPoints(basis_u, num_points, CEED_NOTRANSPOSE, CEED_EVAL_INTERP, x_points, u, v);
```
jeremylt, line 59: It is with great relief that I report that we can interpolate from the background mesh to material points*!

*in 1D with 1 component.... further tests to come

interface/ceed-basis.c
```diff
@@ -1386,6 +1383,198 @@ int CeedBasisApply(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode,
...
+    for (CeedInt i = 0; i < Q_1d; i++) {  // Row i
+      chebyshev_coeffs_1d[Q_1d * i] = I[Q_1d * i] / C[0];
+      for (CeedInt j = 1; j < Q_1d; j++) {  // Column j
+        chebyshev_coeffs_1d[j + Q_1d * i] = I[j + Q_1d * i];
+        for (CeedInt k = 0; k < j; k++) chebyshev_coeffs_1d[j + Q_1d * i] -= C[j + Q_1d * k] * chebyshev_coeffs_1d[k + Q_1d * i];
+        chebyshev_coeffs_1d[j + Q_1d * i] /= C[j + Q_1d * j];
+      }
+    }
+    // ---- Apply Q^T, chebyshev_coeffs_1d = R_inv Q^T
+    CeedCall(CeedHouseholderApplyQ(chebyshev_coeffs_1d, C, tau, CEED_NOTRANSPOSE, Q_1d, Q_1d, Q_1d, 1, Q_1d));
```
jedbrown, line 1490: Maybe we should just have a pseudoinverse utility, or at least a triangular solve utility so we aren't dealing with this here.

interface/ceed-basis.c
```diff
@@ -1386,6 +1383,198 @@ int CeedBasisApply(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode,
...
+                for (CeedInt b = 0; b < Q_1d; b++) {
+                  for (CeedInt c = 0; c < post; c++) out[a * post + c] += chebyshev_x[b] * in[(a * Q_1d + b) * post + c];
+                }
+              }
+            }
+            pre /= Q_1d;
+            post *= 1;
+          }
+        }
+      }
```
jedbrown, line 1565: We can presumably call existing `TensorContract` here. I assume this will be the "hot" spot we'll seek to optimize eventually.

jedbrown (commented):
Cool! It may be worth sketching out the implementation strategy in the documentation. It's something we'll be working on for GPU backends at some point so the math explanation won't be a waste. There will be choices about granularity. For example, the first tensor contraction could evaluate to the y-z planes of all the particles, or of a batch of particles, instead of just one particle at a time. I don't know if that's really a win.

Have you tested the pseudo-inverse approach with Q>P?

interface/ceed-basis.c
jeremylt: For the CPU, yeah, I plan on using the CeedBasis's CeedTensorContract object. On any backend that doesn't have a CeedTensorContract object, I think I'll make a `/cpu/self` ceed and then use that to make a CeedTensorContract (until we get GPU impls)

jeremylt: See the latest push (also using the `TensorContract` object makes clang-tidy happier about buffers it couldn't verify were clean, which probably should give me some pause about clang-tidy's ability to reason through function calls)

interface/ceed-basis.c
jeremylt: Yeah, that would probably be nice to put in a separate PR. I'll put in on the ToDo list.

tests/t352-basis.c
```diff
@@ -0,0 +1,105 @@
...
+
+      CeedVectorGetArrayRead(x_points, CEED_MEM_HOST, &x_array);
+      CeedVectorGetArrayRead(v, CEED_MEM_HOST, &v_array);
+      for (CeedInt i = 0; i < num_points; i++) {
+        CeedScalar coord[dim];
+
+        for (CeedInt d = 0; d < dim; d++) coord[d] = x_array[d + i * dim];
+        for (CeedInt c = 0; c < num_comp; c++) {
+          CeedScalar fx = Eval(dim, c, coord);
+          if (fabs(v_array[c + i * num_comp] - fx) > 1E-4) {
```
jeremylt, line 81: Note, components are contiguous per point, which is a byproduct of how we have the tensor contractions right now

interface/ceed-basis.c
jeremylt: see https://github.com/CEED/libCEED/issues/1250

interface/ceed-basis.c
```diff
@@ -1386,6 +1383,237 @@ int CeedBasisApply(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode,
...
+      CeedCall(CeedVectorRestoreArrayRead(basis->vec_chebyshev, &chebyshev_coeffs));
+      CeedCall(CeedVectorRestoreArrayRead(x_ref, &x_array_read));
+      CeedCall(CeedVectorRestoreArray(v, &v_array));
+      break;
+    }
+    case CEED_TRANSPOSE: {
```
jeremylt, line 1571: I don't have a good test in mind yet, but I have what I think should be the transpose here

jedbrown: So our evaluation to a single point is effectively a row vector $B := b^T$ (implemented by tensor product). We can apply $B$ to columns of the identity to fill up the explicit representation $b^T$. Or we can apply $B^T$ to the value 1.0 at the particle to produce the column vector $b$. Then we check that they're the same.

tests/t353-basis.c
```diff
@@ -0,0 +1,95 @@
+/// @file
+/// Test polynomial interpolation transpose from arbirtary points in 1D
+/// \test Test polynomial interpolation transpose from arbitrary points in 1D
```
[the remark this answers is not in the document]
jeremylt: Note - the LIBXSMM overhaul branch fixes the failure I'm seeing. Alternatively we could merge that branch in anticipation of a "soon" LIBXSMM release and then add the multidimensional transpose test here.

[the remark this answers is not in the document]
jeremylt: @sebastiangrimberg @hfp is there an anticipated timeline for the next LIBXSMM release?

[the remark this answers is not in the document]
jedbrown: If this simplifies our code and fixes a correctness issue for you, I think we can start depending on libxsmm main. If there sitll hasn't been a release when we release v0.12, we can depend on `libxsmm@main` in Spack.

jeremylt:
All backends pass now - ready to merge

jedbrown (approved):
woohoo

PR #1165, opened 2023-02-25
Move some linear algebra/matrix operations to internal API

sebastiangrimberg:
Addresses https://github.com/CEED/libCEED/pull/1156#discussion_r1117813181


jeremylt:
For some reason not all CI tests are running, but I think some of the t30* tests now need ceed/backend.h added

sebastiangrimberg:
> For some reason not all CI tests are running, but I think some of the t30* tests now need ceed/backend.h added

They aren't running because all of your Github Actions workflows are marked as `on: push` and so PRs from forks won't trigger jobs. You can either modify them to have a header which looks like:

```
on:
  push:
    branches:
      - main
  pull_request:
```

or, I'd love to develop right from `CEED/libCEED` in which case none of that is necessary. But I need write access to the repo to do that.

jeremylt:
Huh, I thought that on pull_request also covered forks?

I don't have perms for adding people I don't think

sebastiangrimberg:
`on: pull_request` does cover forks but most tests in libCEED are only `on: push` which does not cover forks. See for example https://github.com/CEED/libCEED/blob/main/.github/workflows/c-fortran-test-linux-osx.yml#L3.

jeremylt:
Ahh. I can fix that on Monday

sebastiangrimberg:
No problem! Another question for this is should these functions get pulled out of the Python/Fortran APIs? Looks like they are already not in the Rust one.

sebastiangrimberg:
Awesome I agree! Thanks

jedbrown:
I added you to the repo so you can supersede with an in-repo branch if you'd like.

sebastiangrimberg:
Thank you!

jeremylt:
See https://github.com/CEED/libCEED/pull/1166 for the CI fix

jeremylt:
I think this is ready for squash+merge but wanted to check if you had anything else to do on this branch first.

sebastiangrimberg:
All good from my end, do what you need to do. I'm happy for you to rebase, squash, and merge as you'd like.

PR #1166, opened 2023-02-27
Run Actions or push to main or PR

jeremylt:
This should fix our CI not running on PRs from forks

### DIFF
--- a/.github/workflows/c-fortran-test-hardware.yml
+++ b/.github/workflows/c-fortran-test-hardware.yml
@@ -2,6 +2,9 @@ name: ARM and IBM Power
 
 on:
   push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   test:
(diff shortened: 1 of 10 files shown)


jeremylt:
Yeah, these is an issue with ROCm on our server that I haven't traced yet

jeremylt:
Instead of dropping the `strip()`, I added a default message if those parts of the test case output are not found. Python really could use optional chaining

jeremylt:
@jedbrown can we try restarting Noether? With this fix to junit.py I'm seeing a bunch of funky looking errors like
```
Test: t206-elemrestriction 
ERROR: returncode = -11
Output: NO OUTPUT
```
```
Test: t201-elemrestriction 
FAIL: stderr
Output: /var/lib/gitlab-runner/builds/N8GCsqus/0/libceed/libCEED/backends/hip/ceed-hip-compile.cpp:114 in CeedCompileHip(): hipErrorSharedObjectInitFailed
```

jedbrown (approved):

PR #1167, opened 2023-03-02
Consistency in c tests

jeremylt:
I use the label 'minor' since this is consistency changes for our tests.

People seem to use the tests as examples of how the library works, but our tests have drifted apart in style as they have been created over the last 5 years. This PR tries to pull the style all in the same direction.

The goal here is to make these tests easier to read and more consistent with the patterns we use in our bigger example/application codes. If we like and merge these changes, we can homogenize some of the other examples down the road.

### DIFF
--- a/tests/t102-vector.c
+++ b/tests/t102-vector.c
@@ -4,23 +4,25 @@
 #include <ceed.h>
 
 int main(int argc, char **argv) {
-  Ceed              ceed;
-  CeedVector        x;
-  CeedInt           n;
-  const CeedScalar *a, *b;
+  Ceed       ceed;
+  CeedVector x;
+  CeedInt    len = 10;
 
   CeedInit(argv[1], &ceed);
 
-  n = 10;
-  CeedVectorCreate(ceed, n, &x);
+  CeedVectorCreate(ceed, len, &x);
   CeedVectorSetValue(x, 0.0);
 
   // Two read accesses should not generate an error
-  CeedVectorGetArrayRead(x, CEED_MEM_HOST, &a);
-  CeedVectorGetArrayRead(x, CEED_MEM_HOST, &b);
+  {
+    const CeedScalar *a, *b;
 
-  CeedVectorRestoreArrayRead(x, &a);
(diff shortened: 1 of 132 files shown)


tests/t102-vector.c
```diff
@@ -4,23 +4,25 @@
...
+    const CeedScalar *a, *b;
 
-  CeedVectorRestoreArrayRead(x, &a);
-  CeedVectorRestoreArrayRead(x, &b);
+    CeedVectorGetArrayRead(x, CEED_MEM_HOST, &a);
+    CeedVectorGetArrayRead(x, CEED_MEM_HOST, &b);
+
+    CeedVectorRestoreArrayRead(x, &a);
+    CeedVectorRestoreArrayRead(x, &b);
+  }
```
jrwrigh, line 25: Is there any logic behind when to enclose locally-needed variables in long functions, like you've done here? I'm always torn on whether to do it or not, and much of the time lean on not unless it's a large segment of code.

jrwrigh (approved):
Looks good to me, just a question about general formatting things.

(also, did you manually go through this or use clang-tidy/clang-format to handle this?)

tests/t102-vector.c
jeremylt: I've gone back and forth too. I've started leaning towards restricting scope, especially for access to raw array pointers for vectors, so that access outside of my intended scope becomes a compile time error.

Since we null array pointers (and maybe we also should for Ceed object pointers even when the recount isn't zero) unintentional access only produces a segfault which is easy to debug, but I've been trying to push more logic issues towards compile errors than segfaults where able. 

At the end of the day it compiles to the same machine code more or less but I'm trying to code more defensively in general.

jeremylt:
It was a combo of VSCode tools and manual work to get these changes

tests/t102-vector.c
jedbrown: Yup, I also prefer smaller scope so long as it's not too distracting. I miss syntax of this form in C:
```rs
let x = {
    let mut a = Thing::new();
    a.frob_in_place();
    a
};
```

PR #889, opened 2022-02-01
Element Restriction Oriented

rezgarshakeri:
Added `CeedElemRestrictionCreateOriented` to flip  dofs (multiply by -1) if needed for H(div) implementation.

PR #1168, opened 2023-03-02
`CeedElemRestriction` for H(curl)

sebastiangrimberg:
High-order H(curl) spaces (`p > 1`) in 3D involve pairs of element face dofs which must be treated specially in the `L`-vector to `E`-vector transformation. These transformations are more complicated than the simple sign flip accounted for in https://github.com/CEED/libCEED/pull/889. In general, the pairs of dofs are transfomed by 2x2 blocks with entries {-1, 0, 1} to write the transformed `E`-vector dofs as linear combinations of the `L`-vector dofs. Thus, if each pair of face dofs are arranged to be consecutive in the element dof ordering, we can represent the transformation required as a tridiagonal matrix with entries {-1, 0, 1} applied to the `E`-vector. Note this is a (more expensive) generalization of the simple orientation flip case where the transformation matrix is just diagonal with entries {-1, 1}. Note that an alternative approach is to apply some global ordering of the nodes of the mesh as a preprocessing step after which a simple signed restriction suffices (like in https://epubs.siam.org/doi/10.1137/08073901X). However this strategy has its own shortcomings in particular when doing things like AMR.

More information can be found in this paper: https://dl.acm.org/doi/pdf/10.1145/3524456 (section 4.1.4, for example). See also https://github.com/mfem/mfem/pull/1046 and https://github.com/mfem/mfem/blob/master/fem/doftrans.hpp for examples of how this transformation is handled in MFEM.

This PR adds `CeedElemRestrictionCreateCurlOriented`, which is similar to the `CeedElemRestrictionCreateOriented` constructor introduced in https://github.com/CEED/libCEED/pull/889. Currently only implemented for CPU backends, with the main change coming in `CeedElemRestrictionApply_Ref_Core`.

Also adds some missing methods which should make `blocked` and `opt` backends work for oriented restrictions.

TODO

- [x] Open questions:
    - The tridiagonal matrix-vector product in `CeedElemRestrictionApply_Ref_Core` is kind of ugly and likely slow. Simplifying this might make the eventual GPU backends for oriented and curl-oriented element restriction easier.
    - The storage of `curl_orients` could be optimized because we know the entries are only {-1, 0, 1}. So we could pack the tridiagonal matrix into an array of `char`, for example, where three pairs of bits starting from the least significant represent `T[i-1,i]`, `T[i,i]`, and `T[i+1,i]`.
- [x] Add unit tests
- [x] Update Python/Julia/Fortran/Rust APIs


backends/ref/ceed-ref-restriction.c
```diff
@@ -373,19 +458,15 @@ int CeedElemRestrictionCreate_Ref(CeedMemType mem_type, CeedCopyMode copy_mode,
     case 381:
       impl->Apply = CeedElemRestrictionApply_Ref_381;
       break;
-    // LCOV_EXCL_START
```
jeremylt, line 376: oops, those exclusion statements can probably be combined but should be there

backends/opt/ceed-opt-operator.c
```diff
@@ -59,12 +59,27 @@ static int CeedOperatorSetupFields_Opt(CeedQFunction qf, CeedOperator op, bool i
...
+        const CeedInt *offsets      = NULL;
+        const bool    *orients      = NULL;
+        const CeedInt *curl_orients = NULL;
         CeedCallBackend(CeedElemRestrictionGetOffsets(r, CEED_MEM_HOST, &offsets));
+        CeedCallBackend(CeedElemRestrictionGetOrientations(r, CEED_MEM_HOST, &orients));
+        CeedCallBackend(CeedElemRestrictionGetCurlOrientations(r, CEED_MEM_HOST, &curl_orients));
         CeedCallBackend(CeedElemRestrictionGetCompStride(r, &comp_stride));
-        CeedCallBackend(CeedElemRestrictionCreateBlocked(ceed, num_elem, elem_size, blk_size, num_comp, comp_stride, l_size, CEED_MEM_HOST,
-                                                         CEED_COPY_VALUES, offsets, &blk_restr[i + start_e]));
+        if (!orients && !curl_orients) {
```
jeremylt, line 69: This makes me wonder if we need an enum for restriction types to make sure backends don't miss a case?

backends/ref/ceed-ref-restriction.c
sebastiangrimberg: Ah yes, I meant to ask about this. I was confused why certain cases were fine without the exclusions and others were not. Is it just to simplify coverage and exclude some sizes?

backends/opt/ceed-opt-operator.c
sebastiangrimberg: That was another approach, similar to the `CeedFESpace` variable for a basis. What do you think? I'm not sure how much more I'd expect things to expand in terms of variants of restriction types.

backends/ref/ceed-ref-restriction.c
jeremylt: Right, there are a few cases that list that are helpful for work in other places but don't want to test here. But that 'uncovered' code just adds noise to the code coverage reports

backends/opt/ceed-opt-operator.c
jeremylt: I could go either way. I like the enums because we have different impls of the same logic and I like the compiler yelling at me when I forget a case. I would picture this being used internally only.

backends/ref/ceed-ref-restriction.c
sebastiangrimberg: Sounds good, I'll add them back

nbeams:
Hi @sebastiangrimberg , I was hoping you could help me understand what exactly you meant by 

> pairs of element face dofs which must be treated specially in the L-vector to E-vector transformation.

I was thinking you meant the pairs of face dofs which are located at the same place (the two tangential components at that location on the face).  But what about orders higher than 2?  I read through the Scroggs et al. paper you linked above, and based on my understanding of that paper, wouldn't having 3 or more edge dofs on an edge mean that if we need to flip the edge, we may need to swap dofs which are not necessarily consecutively ordered?  Did you mean that dofs on edges would also be paired in ordering based on how they would be swapped if the edge were flipped, rather than going "in order" along the edge? But also, what about rotations/reflections of a 3rd-or-higher order face?  What could the transformation matrix look like there?

(In my head I was picturing elements like figure 3.10 in http://www.logg.org/anders/pub/papers/KirbyLoggEtAl2012a.pdf, 
then thinking about how the dofs would be ordered for 3rd order or higher)

sebastiangrimberg:
Hi @nbeams, I'm sorry for the confusing explanation and let me see if I can make my thinking more clear in another go.

In the context of Figure 3.10 in the reference you provided, you can see for order > 3 we have multiple co-located face dofs for each triangular face, but they still come in pairs. The transformation matrix for each face will include a 2x2 block for each pair of face dofs. These transformations only swap each pair of face dofs, there are no couplings between pairs so that's why I say consecutive ordering in each pair will lead the entire transformation matrix for the element to appear with 2x2 blocks. For the edge degrees of freedom, each can be oriented with a simple sign flip.

Maybe specifically in the case of Figure 3.10, each triangular face of the tetrahedon has 15 dofs (9 edge dofs, 6 face dofs). If we order the element dofs as edge dofs followed by face dofs, I believe the transformation matrix would contain a 9x9 block which is diagonal with entries {-1, +1}, followed by 3, 2x2 blocks with entries {-1, 0, 1} orienting each pair of face dofs. The transformation for the entire element would combine the transformation for all faces.

To see this in code, [this method](https://github.com/mfem/mfem/blob/master/fem/doftrans.cpp#LL271C23-L271C23) performs the transformation for triangular elements in MFEM. You can see that for a triangle element of order `p`, we have `p*(p-1)` face dofs which are treated in pairs, multiplied by 2x2 dense matrices in the transformation depending on the orientation type of the face (which can take on one of 6 values depending on the ordering of the nodes).

nbeams:
Thanks for the detailed response. 

>  you can see for order > 3 we have multiple co-located face dofs for each triangular face, but they still come in pairs

Yes, I understand that part.  My confusion is in lining this up with what is presented in the Scroggs paper, unless you do not mean for this PR to be an implementation of exactly that method, and/or I do not understand the paper fully.  E.g., if we have 4 dofs on an edge, numbered 1-2-3-4, and we flip that edge, wouldn't that swap dofs 1 and 4 according to the paper? Even though they both just flip in direction.  Or if we "reflected" the face for a high-order element, then wouldn't some pairs be swapped with each other (but of course they still remain "partnered" in their original pairs).

The linked code for MFEM does indeed seem to only be considering the face dofs, and skipping the edge dofs (everything starts after `[number of edges] * nedofs`).

sebastiangrimberg:
Ah, I think I understand now. Indeed I don't mean to literally implement the exact matrix `M` from the paper as it is generally dense in their implementation (as you point out). But what I'm recognizing is that you can separate into two parts, the permutation of numbering (which is certainly not tridiagonal) followed but the orientation changes and rotations affecting pairs of face dofs (which is tridiagonal). Thus the same effective `M` as provided by a library like Basix should be compatible with this interface in libCEED by providing the correct permutations and tridiagonal transformation matrix.

Also, you are right that in MFEM the orientation changes are handled outside of this `DofTransformation` class, which was added relatively recently. Hence the reason why the code I pointed to is only touching face dofs. I do understand this is slightly confusing but has to do with the order in which things were added to the library I suppose. In the case where you have a nice tensor product element mesh with no triangular faces, you no longer need the general face dof transformations and can make do with only +/- sign flips of your dofs in the element restriction.

interface/ceed-elemrestriction.c
```diff
@@ -440,8 +538,80 @@ int CeedElemRestrictionCreateOriented(Ceed ceed, CeedInt num_elem, CeedInt elem_
...
+  @param[in]  num_comp     Number of field components per interpolation node (1 for scalar fields)
+  @param[in]  comp_stride  Stride between components for the same L-vector "node". Data for node i, component j, element k can be found in the
+L-vector at index offsets[i + k*elem_size] + j*comp_stride.
+  @param[in]  l_size       The size of the L-vector. This vector may be larger than the elements and fields given by this restriction.
+  @param[in]  mem_type     Memory type of the @a offsets array, see CeedMemType
+  @param[in]  copy_mode    Copy mode for the @a offsets array, see CeedCopyMode
+  @param[in]  offsets      Array of shape [@a num_elem, @a elem_size]. Row i holds the ordered list of the offsets (into the input CeedVector) for the
+unknowns corresponding to element i, where 0 <= i < @a num_elem. All offsets must be in the range [0, @a l_size - 1].
+  @param[in]  curl_orients Array of shape [@a num_elem, @a 3 * elem_size] representing a row-major tridiagonal matrix (curl_orients[0] =
+curl_orients[(i + 1) * 3 * elem_size - 1] = 0, where 0 <= i < @a num_elem) which is applied to the element unknowns upon restriction.
```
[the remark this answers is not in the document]
sebastiangrimberg: I would be open to any suggestions you have for improving the docstring. I agree it is hard to explain without giving a lot of background information, but I think users can resolve any further misunderstanding by just looking at the unit tests for example usage.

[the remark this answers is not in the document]
nbeams: Why not go ahead and provide some background information?  It's my understanding that there's not necessarily one "standard" way to handle this among libraries that can do H(curl) elements (or perhaps because it would normally be handled completely inside the library, and not require a parameter from the user), so that's why I think we need to be very clear about what the input needs to be. I don't really like the idea of asking users to parse out the proper way to use a parameter by reading the tests.  Though I suppose adding some more detailed comments to the tests would be better than nothing, and make them even more useful for helping users understand. Even just mentioning a bit of context about the tridiagonal format coming from the pairs of face DOFs would be helpful, IMO.

Does anyone else have any thoughts about whether we should have some more detailed information about element restrictions (basic, oriented, curl oriented) in the higher-level docs (like in doc/sphinx/source/libCEEDapi.md or elsewhere)?  Or is it unnecessary?  If we're going to add that in the future it wouldn't need to be in the PR necessarily.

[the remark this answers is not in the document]
sebastiangrimberg: Hi @nbeams, check out 268839c. Is this satisfactory? I referenced the above paper as well.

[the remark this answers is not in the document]
nbeams: Thanks, I think that's really helpful to have 👍

backends/opt/ceed-opt-operator.c
sebastiangrimberg: Check out 3b8f45e for a suggestion of how this might look with enums. I'm not sure which is better, open to input.

jeremylt: I think it makes more sense to me that way, personally? I'd be curious to see what others think.

jeremylt: Note - small consistency update for creating Blocked restrictions here: https://github.com/CEED/libCEED/pull/1205

sebastiangrimberg: I agree, it it nice this way with the enum. Consistency update fixed after rebase in [27f40bc](https://github.com/CEED/libCEED/pull/1168/commits/27f40bc230702d5d2f53964f310f001318da0e54).

sebastiangrimberg: I don't think the unsigned apply should add new type of restriction to the enum, and instead think it as the following (this is what I've been working with locally where I've merged this branch with #1196):
- `CEED_RESTRICTION_DEFAULT` and `CEED_RESTRICTION_STRIDED`: Have apply, unsigned apply methods which are identical
- `CEED_RESTRICTION_ORIENTED`: Has a special unsigned apply which ignores the orientations and performs the same apply as `CEED_RESTRICTION_DEFAULT`
- `CEED_RESTRICTION_CURL_ORIENTED`: Has a special unsigned apply which applies |E| or |E|^T, this results in a convergent diagonal and is similar to what MFEM does for non-conforming meshes where the P (parallel prolongation) matrix has potentially multiple entries per row (https://github.com/mfem/mfem/blob/master/fem/pbilinearform.cpp#L309).

jeremylt: Maybe its the word "Default" that feels strange to me. If the restriction is set up as oriented, the default apply method should be "CEED_RESTRICTION_ORIENTED"

interface/ceed-preconditioning.c
```diff
@@ -253,6 +254,71 @@ static int CeedOperatorCreateActivePointBlockRestriction(CeedElemRestriction rst
...
+  @brief Create an unsigned restriction for active operator field
+
+  @param[in]  rstr          Original CeedElemRestriction for active field
+  @param[out] unsigned_rstr Address of the variable where the newly created CeedElemRestriction will be stored
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Developer
+**/
+static int CeedOperatorCreateUnsignedRestriction(CeedElemRestriction rstr, CeedElemRestriction *unsigned_rstr) {
```
jeremylt, line 267: So we're undoing the PR just merged? I'm very confused

sebastiangrimberg: Yes give me a few minutes and I'll try to write up an explanation...

Sorry I did not mean to click on your name to re-request review.

jeremylt: I'm curious to see. I don't love the idea of double storing the exact same indices since that's a non trivial amount of memory.

sebastiangrimberg: OK here it goes. I fully agree with you so let me try to describe...

I am only realizing this after #1196 was merged, of course, but doing some more work I realized that p-multigrid will also require an unsigned element restriction based on how the dof multiplicity vector `mult_vec` is computed. Namely the scaling should be (I think) `|E| (E^T E 1)^-1` where I am abusing notation a bit: `| |` is the entry-wise absolute value and `( )^1` is a component-wise reciprocal. The issue with the `ApplyUnsigned` method is that as far as I could tell, there isn't a way to call `CeedOperatorSetField` providing a restriction and tell it to use `ApplyUnsigned` instead of `Apply`. So my thought was just to create a new restriction object sharing the `offsets` memory from the old one but without the signs. This is what I am attempting in [0724f62](https://github.com/CEED/libCEED/pull/1168/commits/0724f6211a6cbff62e9a60e24f9d1574ba274d56), yes it undoes #1196 as you noticed but would solve both cases where an unsigned element restriction is required.

In the case of `CEED_RESTRICTION_ORIENTED`, I am using `CEED_USE_POINTER` to avoid duplication of the indices. However I can't do this for the case of `CEED_RESTRICTION_CURL_ORIENTED` and so the copy of the `curl_orients` array is not nice. I thought about  creating another enum entry for `CEED_RESTRICTION_UNSIGNED_CURL_ORIENTED` which would allow us to avoid the memory duplication and then just applying the `abs` inside of `ElemRestrictionApply` for that restriction type. I wasn't sure if that is a better idea so was going to ask.

Anyway, maybe I am missing something and we can use the `ApplyUnsigned` approach of #1196 even for the multiplicity issue, so I'm curious what you and others think here. Thank you!

jeremylt: If the issue is with multiplicity, we should target that directly. Creating restrictions on the GPU without extra memcopies and allocations is going to be harder than on the CPU.

We could modify the signature of `CeedOperatorMultigridLevelCreate*` to take an optional `rstr_p_mult_fine`.

Another option - We could instead add a `CeedElemRestrictionCreateUnsigned(rstr_old, &rstr_unsigned)` that uses the same backend data but swaps `Apply` to be `ApplyUnsigned`. That might be the cleanest approach. There are some details to iron out, but I could dummy up a version tomorrow morning pretty quickly.

sebastiangrimberg: > Another option - We could instead add a `CeedElemRestrictionCreateUnsigned(rstr_old, &rstr_unsigned)` that uses the same backend data but swaps `Apply` to be `ApplyUnsigned`. That might be the cleanest approach. There are some details to iron out, but I could dummy up a version tomorrow morning pretty quickly.

This seems like a nice way of solving this, I agree! Would be very helpful if you can come up with a demo implementation of this.

jeremylt: This? https://github.com/CEED/libCEED/pull/1206

sebastiangrimberg: Sorry for the delay but this looks good. I'm trying to straighten out if this multiplicity calculation works or what it should look like in the case of the changes provided in this branch, and it isn't super straightforward to me. If I think of `\mathcal{E} = TE` where `E` is the standard restriction of L-dofs to E-dofs and `T` is the block 2x2 matrix with orientations and transformations like `[-1 0; -1 1]` (for example), then I don't know if `|\mathcal{E}| (|\mathcal{E}|^T|\mathcal{E}| 1)^-1` is actually the right multiplicity to use to get an interpolating operator. It's almost like you always want to use only `E (E^T E 1)^-1` and ignore the auxiliary orientations and transformations data altogether. Though I'm also not sure about this.

jeremylt: The H^Div and H^Curl interpolation operations are a bit of a mystery to me currently, so I'm not quite sure.

The multiplicity should scale for duplication across element and process boundaries, but I'm not sure what that is for those bases

sebastiangrimberg:
A question (for @jeremylt  and others) as I work on rebasing this branch now that https://github.com/CEED/libCEED/pull/1206 is merged:

If there are 4 public interface methods for `ElemRestriction` creation: `CeedElemRestrictionCreate`, `...CreateOriented`, and `...CreateCurlOriented`, `...CreateStrided`, plus 4 more for the `...CreateBlocked...` variants of each, I see two options for backend implementations:

1. Have a 1:1 correspondence for backend functions to the interface (8 total), that is there are lookup table entries for all of `ElemRestrictionCreate`, `...CreateOriented`, ..., `...CreateStridedBlocked`. Each backend will call `CeedSetBackendFunction` for the subset of the 8 total that it implements (right now, only `ref` would implement the full set).
2. Alternatively, as is done now for `ElemRestrictionCreateStrided` and `...CreateBlockedStrided`, we could have only 2 variants of backend functions in the table: `ElemRestrictionCreate` and `ElemRestrictionCreateBlocked`, where the backend registered function handles the type of restriction correctly using a call to `CeedElemRestrictionGetType`. This would mean the signature for these backend functions would need to be expanded to handle `offsets`, `orients`, and `curl_orients` (some of which may be NULL), which is less appealing than option (1) but on the other hand it might be nicer as we can avoid needing 8 separate backend function table entries for element restriction creation.

Do you have a preference here? I'm sorry if I am not doing a good job at explaining what I see as the alternative options, just let me know and I can do a better job.

jeremylt:
I think that I like 2) for internal interfaces. I generally like more specific functions for the user facing API but more general function that support less overall repetition for internal facing APIs.

Perhaps we should pass an additional argument to the backend function indicating which of `offsets`, `orients`, and `curl_orients` should be expected?

sebastiangrimberg:
> I think that I like 2) for internal interfaces. I generally like more specific functions for the user facing API but more general function that support less overall repetition for internal facing APIs.
> 
> Perhaps we should pass an additional argument to the backend function indicating which of offsets, orients, and curl_orients should be expected?

Cool, I'll work on this. Thank you!

interface/ceed-preconditioning.c
sebastiangrimberg: Resolving for now due to the discussion [here](https://github.com/CEED/libCEED/pull/1206#discussion_r1184507683). The case of curl-oriented element restrictions is not accounted for yet but will be left to resolve in the future outside the scope of this PR.

sebastiangrimberg:
> Perhaps we should pass an additional argument to the backend function indicating which of offsets, orients, and curl_orients should be expected?

For now (in [a32a7e7](https://github.com/CEED/libCEED/pull/1168/commits/a32a7e76f6c756eedceaa2617be4b2c575023641)), I have just checked that the provided arguments are consistent with the `rstr_type` as supported by the backend.

> I think that I like 2) for internal interfaces. I generally like more specific functions for the user facing API but more general function that support less overall repetition for internal facing APIs.

Regarding this, we could go a bit further than I have by removing the backend `ElemRestrictionCreateBlocked` function entry altogether as the only backend which defines it is `ref` at least as far as I can tell, and there it is the same as the function for `ElemRestrictionCreate`. If you'd like I can add that change, or we can leave for later.

jedbrown:
Sorry about the slow movement on this. Is it just awaiting review/merging at this point?

sebastiangrimberg:
No worries at all, thanks for following up @jedbrown. I've tested and been working with this branch through MFEM and in our application code and as far as correctness goes everything is checking out, with complete partial assembly support for non-tensor H(curl) elements and even mixed H1-H(curl) and H(curl)-H(div). However in profiling I'm seeing that the element restriction is expensive relative to basis and quadrature functions, especially at low orders, so I think there is still work to be done improving this implementation before merging. There's a lot of structure to the actual linear combinations of face dofs in the element restriction that isn't taken advantage of in the general `int`-valued tridiagonal form (values are only -1, 0, 1, and off diagonals only appear in a few 2x2 blocks), so I think there's room for improvement.

I'd be open to suggestions if anything stands out from those with more experience than me, but I think I will be able to experiment a bit further by next week.

backends/opt/ceed-opt-operator.c
```diff
@@ -42,29 +42,52 @@ static int CeedOperatorSetupFields_Opt(CeedQFunction qf, CeedOperator op, bool i
 
     if (eval_mode != CEED_EVAL_WEIGHT) {
       CeedCallBackend(CeedOperatorFieldGetElemRestriction(op_fields[i], &r));
-      Ceed ceed;
```
jeremylt, line 45: This should be here because this `ceed` may be different than the `ceed` for the whole operator

sebastiangrimberg: Ah I see, thanks! [f52c7b8](https://github.com/CEED/libCEED/pull/1168/commits/f52c7b8346769c8482ac5c1fee869e66a09109bc) should fix this and I think the renaming should clarify this for other developers hopefully.

backends/opt/ceed-opt-operator.c
jedbrown: Is this resolved?

jedbrown (approved):
Thanks for this major feature and careful testing. It looks good to me, though

- [ ] Would you prefer to squash-merge or to rebase-squash the WIP/revert commits.
- [ ] GitLab pipeline didn't run. Hopefully it's just been a while and will run if you update/rebase the branch.

sebastiangrimberg:
> * [ ]  Would you prefer to squash-merge or to rebase-squash the WIP/revert commits.

Totally up to you as the project maintainers! Do whatever you'd like to the PR or if you need me to squash/rebase I can do that.

> * [ ]  GitLab pipeline didn't run. Hopefully it's just been a while and will run if you update/rebase the branch.

I think this is unfortunately because this PR is from a fork (since it's from before I had write access to CEED/libCEED). One option is to close this PR and open a new one from a new branch inside of CEED/libCEED which will trigger the GitLab PR. Let me know if/when I should go ahead and do this.

backends/opt/ceed-opt-operator.c
sebastiangrimberg: @jeremylt if you are happy with the current implementation, then it's resolved. Just let me know if you have any questions about it or see things I should change. The enum values are still:

```
/// Type of element restriction;
typedef enum {
  /// Default element restriction with offsets
  CEED_RESTRICTION_DEFAULT = 1,
  /// Oriented element restriction
  CEED_RESTRICTION_ORIENTED = 2,
  /// Curl-oriented element restriction
  CEED_RESTRICTION_CURL_ORIENTED = 3,
  /// Strided element restriction
  CEED_RESTRICTION_STRIDED = 4,
} CeedRestrictionType;
```

but hopefully the reorganization of the apply methods in `ceed-ref-restriction.c` has helped make the organization and naming more clear.

jeremylt: I don't love `DEFAULT` as "not strided, oriented, or curl oriented", but I don't have a better name. It sounds funny in my mind because different `CeedElemRestriction*Create` have different "default" application techniques. Maybe `CEED_RESTRICTION_UNSIGNED`? I dunno, maybe its just a mild :bike: :house: on how the words sound in my mind against the API

I think the organization is clear though.

sebastiangrimberg: I guess the best argument for `DEFAULT` is the public API. We have `CeedElemRestrictionCreate`, `...CreateOriented`, `...CreateCurlOriented`, and `...CreateStrided`, so the "default" case really is defined there by "not strided, oriented, or curl oriented" as you say. But I'm in favor of using something more descriptive. `"OFFSETTED"` is a pretty ugly past-participle, but how about `"STANDARD"` to avoid the use of "default"?

jeremylt: `STANDARD` does work in my brain better. Thanks for the flexibility!

rezgarshakeri: When you call one of them `CurlOriented` why not just call the others by the FE Space that we are using? maybe`CreatedH1`, `CreatedHdiv` and `CreatedHcurl` ?

nbeams: Couldn't we just use `_OFFSET` as the type? Isn't it its own past participle? :)  
I think the argument for "standard" from function names makes a lot of sense, though.

sebastiangrimberg: Nice, done.

> When you call one of them CurlOriented why not just call the others by the FE Space that we are using? maybeCreatedH1, CreatedHdiv and CreatedHcurl ?

I think it's a good idea to decouple the element restriction from the FE space and basis as much as possible. For example, we use `CreateStrided`, not `CreateL2`. Also this would involve some breaking changes to the public API that would be good to avoid.

> Couldn't we just use `_OFFSET` as the type? Isn't it its own past participle? :) I think the argument for "standard" from function names makes a lot of sense, though.

This would work too. To avoid getting too far into the bike shed regime I just went with `STANDARD` in 
82a976c. If there are no objections I'll consider this resolved.

sebastiangrimberg:
Regarding GitLab CI failures, correct me if I'm wrong but I think we need a way to opt-out of certain `CeedElemRestrictionCreate*` methods from backends which don't support it. All of the public API variants map to `ElemRestrictionCreate`. Should there be a check inside of the various backends to ensure that the `CeedRestrictionType` is supported? And in this case, just to confirm, I believe the check and error statement should be something like:

`CeedCheck(rstr_type != CEED_RESTRICTION_ORIENTED && rstr_type != CEED_RESTRICTION_CURL_ORIENTED, CEED_ERROR_BACKEND, "Backend does not implement CeedElemRestrictionCreateOriented or CeedElemRestrictionCreateCurlOriented");`

Thanks again for the help.

jeremylt:
I'll point you to the specific wording when I'm back at the desk in the morning. The convention is captured in tests/junit.py but that's not easy to parse at first glance

backends/opt/ceed-opt-operator.c
jedbrown: My only question is whether `STRIDED` is semantically an alternative to `ORIENTED`, versus orthogonal to? I probably need to read more carefully, but that part read strangely to me.

jedbrown:
Could we delegate on those cases instead of erroring?

jeremylt:
Crossing delegation between GPU and CPU impls won't work because of evec ordering

backends/opt/ceed-opt-operator.c
sebastiangrimberg: Interesting and good point that I guess we could have an oriented, strided restriction if I understand what you are saying. But the way things are implemented they are orthogonal and with increasing levels of generality between the different types. By that I mean that a `STANDARD` restriction captures all the functionality of the (more optimized) `STRIDED` one, adding the ability to overlap points between elements. Likewise `ORIENTED` augments `STANDARD` by adding orientations (if these are all `false` we have `STRIDED` behavior), and finally `CURL_ORIENTED` augments `ORIENTED` by allowing more general element-wise transformations. So even if `STRIDED` and `ORIENTED` are not really orthogonal but could be composed, the user should just use `CeedElemRestrictionCreateOriented` with a non-overlapping `offsets` array. I hope I understood your comment correctly, let me know if this makes sense or not.

jedbrown: Yeah, that was kind of my question. An enum is fine for a hierarchy of generality, but not for what may eventually become the cross product of orthogonal knobs. If this is a hierarchy, then there will be no need to later add other entries to complete a cross product.

> But the way things are implemented they are orthogonal ...

An enum (mutually exclusive options) doesn't sound like orthogonality, though I'm not sure you intended it that way.

jeremylt: At least how we currently use strided, its only for when the evec the same length as the lvec and is strictly a permutation of the lvec (And I think our only current usage is for QData). (maybe we should have called it a permutation - names are hard).

Is there a use case where we will want to permute and orient? It feels like in that case we'd be building the orientation array anyways so using the current implementation wouldn't be a big difference in performance with making a special strided + oriented implementation.

sebastiangrimberg:
Once I get the go ahead and this is approved for merging, I will squash WIP/revert commits and force push if that sounds like a good plan.

backends/opt/ceed-opt-operator.c
jedbrown: Thanks, both, for this discussion.. The present design sounds good to me.

jedbrown:
Your squash plan sounds good. I think we also need
- [ ] add entry to `doc/sphinx/source/releasenotes.md`
- [ ] add your name to `AUTHORS`

sebastiangrimberg:
See #1265 for the rebased + squashed branch, to be merged.

jedbrown:
Thanks! What do you think of the claimed code coverage? https://app.codecov.io/gh/CEED/libCEED/pull/1168/blob/backends/ref/ceed-ref-restriction.c

There are sizable loop/indexing blocks that are purportedly not tested. Do you think the tests cover those or would you agree they are not covered? (If it's a codecov issue, we can fix later, but those are pretty big blocks to actually be untested/unreachable.)

sebastiangrimberg:
I think this is right. Let me see if I can improve this by playing around with the unit tests a bit.

sebastiangrimberg:
> There are sizable loop/indexing blocks that are purportedly not tested. Do you think the tests cover those or would you agree they are not covered? (If it's a codecov issue, we can fix later, but those are pretty big blocks to actually be untested/unreachable.)

I think ae4e5d7 should offer a lot of coverage improvement for those loops, though I have to admit I'm not fully sure how to check. Looking at https://app.codecov.io/gh/CEED/libCEED/pull/1168/blob/backends/ref/ceed-ref-restriction.c doesn't show any data.

jedbrown:
Coverage looks off in this PR, but it looks reasonable (and much improved) for #1265: https://app.codecov.io/gh/CEED/libCEED/pull/1265

What do you think of the remaining red blocks? I'm not too concerned about coverage of the Fortran interface, but there's still a couple functions and loop nests

sebastiangrimberg:
> Coverage looks off in this PR, but it looks reasonable (and much improved) for #1265: https://app.codecov.io/gh/CEED/libCEED/pull/1265
> 
> What do you think of the remaining red blocks? I'm not too concerned about coverage of the Fortran interface, but there's still a couple functions and loop nests

Nice, I added a new test `t571` which should hopefully fix the remaining untested code paths for curl-oriented element restrictions (fingers crossed).

backends/opt/ceed-opt-operator.c
```diff
@@ -41,30 +41,54 @@ static int CeedOperatorSetupFields_Opt(CeedQFunction qf, CeedOperator op, bool i
...
+          const CeedInt  *offsets      = NULL;
+          const CeedInt8 *curl_orients = NULL;
+          CeedCallBackend(CeedElemRestrictionGetOffsets(r, CEED_MEM_HOST, &offsets));
+          CeedCallBackend(CeedElemRestrictionGetCurlOrientations(r, CEED_MEM_HOST, &curl_orients));
+          CeedCallBackend(CeedElemRestrictionCreateBlockedCurlOriented(ceed_rstr, num_elem, elem_size, blk_size, num_comp, comp_stride, l_size,
+                                                                       CEED_MEM_HOST, CEED_COPY_VALUES, offsets, curl_orients,
+                                                                       &blk_restr[i + start_e]));
+          CeedCallBackend(CeedElemRestrictionRestoreOffsets(r, &offsets));
+          CeedCallBackend(CeedElemRestrictionRestoreCurlOrientations(r, &curl_orients));
+        } break;
```
jeremylt, line 85: Same comment here

backends/blocked/ceed-blocked-operator.c
```diff
@@ -41,29 +41,54 @@ static int CeedOperatorSetupFields_Blocked(CeedQFunction qf, CeedOperator op, bo
...
+          const CeedInt  *offsets      = NULL;
+          const CeedInt8 *curl_orients = NULL;
+          CeedCallBackend(CeedElemRestrictionGetOffsets(r, CEED_MEM_HOST, &offsets));
+          CeedCallBackend(CeedElemRestrictionGetCurlOrientations(r, CEED_MEM_HOST, &curl_orients));
+          CeedCallBackend(CeedElemRestrictionCreateBlockedCurlOriented(ceed_rstr, num_elem, elem_size, blk_size, num_comp, comp_stride, l_size,
+                                                                       CEED_MEM_HOST, CEED_COPY_VALUES, offsets, curl_orients,
+                                                                       &blk_restr[i + start_e]));
+          CeedCallBackend(CeedElemRestrictionRestoreOffsets(r, &offsets));
+          CeedCallBackend(CeedElemRestrictionRestoreCurlOrientations(r, &curl_orients));
+        } break;
```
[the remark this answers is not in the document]
sebastiangrimberg: I think I see why these aren't being tested. `t571` creates a `CeedOperator` which as I understand it, should dispatch to `CeedOperatorLinearAssembleQFunctionCore_Opt` and `CeedOperatorLinearAssembleQFunctionCore_Blocked` when we call `CeedOperatorLinearAssemble` and thus cover these code blocks. It looks like it isn't though, and every call in the test falls back to `CeedOperatorLinearAssembleQFunctionCore_Ref` in the `ref/serial` backend. I think this is because in `CeedSingleOperatorAssemble` we don't have an `op->LinearAssembleSingle` entry and thus we fall back to `ref/serial` before we actually need to. Is this intended/is there a way to improve this to make use of the `CeedOperatorLinearAssembleQFunctionCore` specializations for the `opt`/`blocked` backends?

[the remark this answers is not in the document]
jeremylt: Yeah, that's by design. We only have backend specific variants of `CeedOperatorLinearAssembleQFunction`, not of the full operator assembly. It hasn't bottlenecked us to not have some special version of assembly for the opt operator evaluation.

So probably a separate test that directly calls `CeedOperatorApply` will fix this?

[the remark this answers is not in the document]
sebastiangrimberg: Sorry I meant "resolve the issue" as "resolve the CPU issue" regarding prematurely falling back for `opt` and `ref/blocked`. But yes I see the problem with that simple approach now. Maybe I can think about this a bit more about it of how to make use of the `CeedOperatorLinearAssembleQFunctionCore_Opt/Blocked` here still.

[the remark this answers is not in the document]
jeremylt: `CeedOperatorLinearAssembleQFunctionCore_Opt/Blocked` is being used. The missing lines of coverage are in the `CeedOperatorApply_Opt/Blocked` related code.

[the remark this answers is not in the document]
sebastiangrimberg: If you add print statements to `CeedOperatorLinearAssembleQFunctionCore_Opt/Blocked`, you can see they are not being used by `t571`. This is because these backends don't have a `LinearAssembleSingle` specialization so we fall back to `ref/serial` before we get the chance to call `LinearAssembleQFunction`/`LinearAssembleQFunctionUpdate`, which they do have specializations for. If we were in fact calling the specialized `CeedOperatorLinearAssembleQFunctionCore` methods, then we would get calls to `CeedOperatorSetup_Opt/Blocked` and the test coverage would be fixed.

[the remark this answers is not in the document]
sebastiangrimberg: 👍 Agreed, done in 726c2de so that should hopefully squash the remaining coverage issues.

tests/t571-operator.c
```diff
@@ -0,0 +1,263 @@
...
+
+  CeedOperatorCreate(ceed, qf_mass, CEED_QFUNCTION_NONE, CEED_QFUNCTION_NONE, &op_mass_oriented);
+  CeedOperatorSetField(op_mass_oriented, "rho", elem_restriction_q_data, CEED_BASIS_COLLOCATED, q_data);
+  CeedOperatorSetField(op_mass_oriented, "u", oriented_elem_restriction_u, basis_u, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(op_mass_oriented, "v", oriented_elem_restriction_u, basis_u, CEED_VECTOR_ACTIVE);
+
+  CeedOperatorCreate(ceed, qf_mass, CEED_QFUNCTION_NONE, CEED_QFUNCTION_NONE, &op_mass_curl_oriented);
+  CeedOperatorSetField(op_mass_curl_oriented, "rho", elem_restriction_q_data, CEED_BASIS_COLLOCATED, q_data);
+  CeedOperatorSetField(op_mass_curl_oriented, "u", curl_oriented_elem_restriction_u, basis_u, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(op_mass_curl_oriented, "v", curl_oriented_elem_restriction_u, basis_u, CEED_VECTOR_ACTIVE);
```
jeremylt, line 111: It would be helpful for future development to test these two new operators separately before diving into full assembly tests. Maybe 
```
    5.7. CeedOperator extra diagonal and full assembly tests
```
should become
```
    5.7. CeedOperator Hdiv and Hcurl tests
```
and we add a simpler 571/572 for testing `CeedOperatorApply` for these two mass operators and rename this 573?

sebastiangrimberg: See d47b7a3.

python/ceed_basis.py
```diff
@@ -45,7 +45,7 @@ def apply(self, nelem, emode, u, v, tmode=NOTRANSPOSE):
            Args:
              nelem: the number of elements to apply the basis evaluation to;
                       the backend will specify the ordering in a
-                      BlockedElemRestriction
```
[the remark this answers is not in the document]
sebastiangrimberg: On `main`, there are two uses one of `Blocked ElemRestriction` and one of `BlockedElemRestriction` in ceed_basis.py. Your previous comment https://github.com/CEED/libCEED/pull/1265#discussion_r1275163858 (correctly, in my opinion) identified the need to change the `BlockedElemRestriction` into `Blocked ElemRestriction` for consistency, which matches usage in `ceed_elemrestriction.py`. Do you want me to change it back now?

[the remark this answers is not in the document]
jeremylt: I don't care that much.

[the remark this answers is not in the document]
sebastiangrimberg: Sounds good, I'll leave as is.

backends/blocked/ceed-blocked-operator.c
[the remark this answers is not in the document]
sebastiangrimberg: Nice, I will rebase #1265 and double check with you once that is done before merging.

sebastiangrimberg:
Closed in favor of #1265.

PR #1170, opened 2023-03-06
Adding CeedVectorCopy() and CeedVectorAXPBY()

karenlstengel:
Includes Rust, Cuda, and HIP hooks.


interface/ceed-vector.c
```diff
@@ -220,6 +220,46 @@ int CeedVectorReferenceCopy(CeedVector vec, CeedVector *vec_copy) {
...
+
+  // Get the preferred memory type
+  CeedVectorGetCeed(*vec_copy, &ceed);
+  CeedGetPreferredMemType(ceed, &mem_type_copy);
+
+  // Check that both have same memory type
+
+  if (mem_type != mem_type_copy) {
+    mem_type = CEED_MEM_HOST;
+  }
```
jeremylt, line 253: minor
```suggestion
  // Check that both have same memory type
  if (mem_type != mem_type_copy) mem_type = CEED_MEM_HOST;
```

rust/libceed/src/vector.rs
```diff
@@ -636,6 +636,40 @@ impl<'a> Vector<'a> {
...
+    ///
+    /// y = y.axpby(-0.5, 1.0, &x)?;
+    /// for (i, y) in y.view()?.iter().enumerate() {
+    ///     assert_eq!(*y, (i as Scalar) / 2.0, "Value not set correctly");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[allow(unused_mut)]
+    pub fn axpby(
```
jeremylt, line 662: Can we add copy here too?

I think adding these functions in `python/ceed_vector.py` (and `python/tests/test-1-vector.py`) should be quick?

interface/ceed-vector.c
```diff
@@ -695,6 +735,83 @@ int CeedVectorAXPY(CeedVector y, CeedScalar alpha, CeedVector x) {
...
+  CeedCall(CeedVectorGetArray(y, CEED_MEM_HOST, &y_array));
+  CeedCall(CeedVectorGetArrayRead(x, CEED_MEM_HOST, &x_array));
+
+  assert(x_array);
+  assert(y_array);
+
+  for (CeedInt i = 0; i < n_y; i++) {
+    y_array[i] = beta * y_array[i];
+    y_array[i] += alpha * x_array[i];
+  }
```
jeremylt, line 807: Why not all at once like
```suggestion
  for (CeedInt i = 0; i < n_y; i++) y_array[i] = alpha * x_array[i] + beta * y_array[i];
```
so it matches the docstring more closely?

jeremylt (commented):
Looks pretty good - I just had a couple of small comments. 

Can we add some tests to make sure everything works right?

interface/ceed-vector.c
```diff
@@ -220,6 +220,43 @@ int CeedVectorReferenceCopy(CeedVector vec, CeedVector *vec_copy) {
...
+          This CeedVector will be destroyed if `*vec_copy` is the only reference to this CeedVector.
+
+  @param[in]     vec      CeedVector to copy
+  @param[in,out] vec_copy Variable to store copied CeedVector to
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedVectorCopy(CeedVector vec, CeedVector *vec_copy) {
```
jeremylt, line 236: I think this is throwing you here:
```suggestion
int CeedVectorCopy(CeedVector vec, CeedVector vec_copy) {
```

(This change will cause other changes)

python/ceed_vector.py
```diff
@@ -58,6 +58,17 @@ def __str__(self):
...
+    def copy(self, vec_copy):
+        """Copies the array of a self into the array of vec_copy.
+
+           Args:
+             *vector: the Vector to copy to"""
+
+        # libCEED call
+        err_code = lib.CeedVectorCopy(self._pointer[0], vec_copy._pointer)
+        self._ceed._check_error(err_code)
+
```
jeremylt, line 71: Then you can do this
```suggestion
             *vector: the Vector to copy"""

        # libCEED call
        err_code = lib.CeedVectorCopy(self._pointer[0], vec_copy._pointer[0])
        self._ceed._check_error(err_code)

```

rust/libceed/src/vector.rs
```diff
@@ -269,6 +269,21 @@ impl<'a> Vector<'a> {
...
+    /// * `vec_copy` - vector to copy array values to
+    #[allow(unused_mut)]
+    fn vec_copy_mut(
+        mut self,
+        vec_copy: crate::Vector,
+    ) -> crate::Result<bind_ceed::CeedVector> {
+        let ierr = unsafe { bind_ceed::CeedVectorCopy(self, vec_copy) };
+        self.check_error(ierr)?;
+        Ok(vec_copy)
+    }
```
jeremylt, line 285: Hmm, this looks like a situation where you may be fighting the borrow checker. I think this would prevent some ownership headaches:
```suggestion
    fn copy_from(
        mut self,
        vec_source: &crate::Vector,
    ) -> crate::Result<bind_ceed::CeedVector> {
        let ierr = unsafe { bind_ceed::CeedVectorCopy(vec_source.ptr, self.ptr) };
        self.check_error(ierr)?;
        Ok(self)
    }
```

tests/t125-vector.c
```diff
@@ -0,0 +1,45 @@
+/// @file
+/// Test summing of a pair of vectors
+/// \test Test y = alpha x + beta y
+#include <ceed.h>
+#include <math.h>
+
+int main(int argc, char **argv) {
+  Ceed              ceed;
+  CeedVector        x, y;
+  CeedInt           n;
```
jeremylt, line 10: See main for a few similar changes to these tests:
```suggestion
  CeedInt           len = 10;
```

python/tests/test-1-vector.py
```diff
@@ -208,6 +208,29 @@ def test_108(ceed_resource, capsys):
 
     assert abs(norm - 9.) < TOL
 
+# -------------------------------------------------------------------------------
+# Test vector copy
+# -------------------------------------------------------------------------------
+
+
+def test_109(ceed_resource, capsys):
```
jeremylt, line 216: Test numbering should match the numbering of the new C tests

jeremylt (commented):
couple of thoughts

python/tests/test-1-vector.py
```diff
@@ -366,6 +343,30 @@ def test_125(ceed_resource, capsys):
...
+    n = 10
+
+    x = ceed.Vector(n)
+    y = ceed.Vector(n)
+
+    a = np.arange(10, 10 + n, dtype=ceed.scalar_type())
+    x.set_array(a, cmode=libceed.USE_POINTER)
+
+    with x.array() as x_array:
+        y.copy(x_array)
```
jeremylt, line 363: I think the test is upset because it wants a libCEED vector?
```
y.copy(x)
```
I wonder if it might make sense to replicate the Rust interface change?

rust/libceed/src/vector.rs
```diff
@@ -273,15 +273,15 @@ impl<'a> Vector<'a> {
...
     /// # arguments
     ///
-    /// * `vec_copy` - vector to copy array values to
+    /// * `vec_source` - vector to copy array values from
     #[allow(unused_mut)]
-    fn vec_copy_mut(
+    fn copy_from(
         mut self,
-        vec_copy: crate::Vector,
+        vec_source: &crate::Vector,
```
jeremylt, line 280: I think I missed below, return type of `crate::Result<Self>` now, like with `from_slice`?

tests/t126-vector.c
```diff
@@ -6,35 +6,43 @@
...
 
-  CeedSize len;
-  CeedVectorGetLength(x_copy, &len);
-  if (len != n) printf("Error copying CeedVector: %td\n", len);
+  {
+    CeedSize len_2;
+
+    CeedVectorGetLength(x_copy, &len_2);
+    if (len != len_2) printf("Error copying CeedVector\n");
+  }
```
jeremylt, line 35: These C tests look pretty good now. I think you can skip the length check since `CeedVectorCopy` cannot change the length of the underlying array

rust/libceed/src/vector.rs
karenlstengel: ah, that makes sense.

python/ceed_vector.py
```diff
@@ -58,6 +58,17 @@ def __str__(self):
 
         return out_string
 
+    # Copy the array from a vector into self
+    def copy(self, vec_source):
```
jeremylt, line 62: I'd change the method name to copy_from, but otherwise I think this is ready to merge!

jeremylt (approved):

jedbrown (approved):

jeremylt:
I think this is ready for squash+merge @karenlstengel

PR #1173, opened 2023-03-07
Add CeedOperatorGetContext

jeremylt:
Adding `CeedOperatorGetContext` provides more ways to get at the context data without dragging around a big struct

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -1426,6 +1426,30 @@ int CeedOperatorGetFlopsEstimate(CeedOperator op, CeedSize *flops) {
 }
 
 /**
+  @brief Get CeedQFunction global context for a CeedOperator.
+           Note: The caller is responsible for destroying `ctx` returned from this function via `CeedQFunctionContextDestroy()`.
+           Note: If the value of `ctx` passed into this function is non-NULL, then it is assumed that `ctx` is a valid pointer to a
+             CeedQFunctionContext. This CeedQFunctionContext will be destroyed if `ctx` is the only reference to this CeedQFunctionContext.
+
+  @param[in]  qf  CeedQFunction
+  @param[out] ctx Variable to store CeedQFunctionContext
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedOperatorGetContext(CeedOperator op, CeedQFunctionContext *ctx) {
+  bool is_composite;
+
+  CeedCall(CeedOperatorIsComposite(op, &is_composite));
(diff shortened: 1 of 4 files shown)


jeremylt:
Side note - for `CeedOperatorContextGetFieldLabel` and similar, technically a `CeedOperatorContext` isn't an actual thing. Perhaps these should actually be named `CeedOperatorGetContextFieldLabel` and similar.

interface/ceed-operator.c
```diff
@@ -1425,6 +1425,30 @@ int CeedOperatorGetFlopsEstimate(CeedOperator op, CeedSize *flops) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Get CeedQFunction global context for a CeedOperator.
+           Note: The caller is responsible for destroying `ctx` via `CeedQFunctionContextDestroy()`.
+           Note: If `*ctx` is non-NULL, then it is assumed that `*ctx` is a pointer to a CeedQFunctionContext.
+             This CeedQFunctionContext will be destroyed if `*ctx` is the only reference to this CeedQFunctionContext.
```
jrwrigh, line 1432: This seems redundant and a little confusing. I see that the second note is posted on other functions, but usually by itself. Having both kinda confused me here that there might be some special rules with the function, but they're both talking about the same idea more-or-less. I think the first one is more direct (the second basically explains the concept of the `*Destroy` functions).

jrwrigh (approved):
LGTM, pending the function documentation header

interface/ceed-operator.c
jeremylt: I think there are three ideas here:

1) If the `ctx` passed in is non-null, then its assumed to be actually be a `ctx` and it will be destroyed as part of this function call (so `ctx` passed in should hold an actual QFunctionContext or should be initialized to NULL)

2) If that `ctx` is the last reference to a CeedQFunctionContext, then that memory will be freed

3) The `ctx` that is passed back out does need to be destroyed

jrwrigh: Oooh, gotcha. I thought `*ctx` was referring to the state after the function execution, not the state being passed into the function.

jeremylt: Tweaked the wording a bit to hopefully make this clearer

jrwrigh: Yep, reads clear to me now!

jrwrigh: Might be worth changing the other uses of that phrase too, assuming it also applies. Grepping `non-NULL` did the trick for me bringing them all up.

PR #1174, opened 2023-03-07
CeedOperatorContext* -> CeedOperator*Context*

jeremylt:
Rename for consistency as `CeedOperatorContext` isn't a real thing


jrwrigh:
Is there any reason to not just use `CeedOperatorGetFieldLabel`? I haven't looked around the rest of the API, but the "context" keyword is most familiar to me in user-defined structs.

jeremylt:
We have `OperatorField`s and `QFunctionContextField`s, so I think we want to be clear that we are talking about a ContextField instead of an OperatorField

jrwrigh (approved):

PR #1177, opened 2023-03-11
fluids: Use ContextGetDouble for time and dt label updates

jrwrigh:
Instead of using `user->{time,dt}` to keep track of what data is set in the QFContext labels (used for determining when to update the labels), use the `CeedOperatorGetContextDoubleRead` and friends to get the label data directly.

### DIFF
--- a/examples/fluids/navierstokes.h
+++ b/examples/fluids/navierstokes.h
@@ -168,7 +168,7 @@ struct User_private {
   CeedVector   q_ceed, q_dot_ceed, g_ceed, coo_values_amat, coo_values_pmat, x_ceed;
   CeedOperator op_rhs_vol, op_rhs, op_ifunction_vol, op_ifunction, op_ijacobian, op_dirichlet;
   bool         matrices_set_up;
-  CeedScalar   time, dt, time_bc_set;
+  CeedScalar   time_bc_set;
   Span_Stats   spanstats;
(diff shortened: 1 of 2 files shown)


jedbrown (commented):
Is this going to cause a device->host transfer? We could also get it from `TS`.

jeremylt:
This should not create an extra device->host transfer since the QFunctions never update this value (I think). I don't think getting this value from the TS will work, since that would tell us what the TS has as the time, not what the CeedOperator has as the time.

jrwrigh:
We discussed the host-device transfer a bit in https://github.com/CEED/libCEED/issues/1144.

But ditto what Jeremy just wrote

jrwrigh:
I guess alternatively we could just update the context field label regardless of the contents. Won't do harm datawise, but it will result in multiple unnecessary writes.

jeremylt:
Specifically it will result in extra host->device

jrwrigh:
So the read doesn't require device->host transfer, but the write requires host->device? (the latter is obvious, but the former seems odd, though I'm not familiar with the details of device <-> host memory transfers).

jeremylt:
The write marks the device data as stale so an extra host->device is needed on the next QFunction evaluation.

jedbrown (approved):

PR #1178, opened 2023-03-13
Add CeedVectorViewRange utility

jeremylt:
@karenlstengel I think this is a version of what we were talking about on Friday?

### DIFF
--- a/interface/ceed-vector.c
+++ b/interface/ceed-vector.c
@@ -940,6 +940,45 @@ int CeedVectorReciprocal(CeedVector vec) {
 /**
   @brief View a CeedVector
 
+          Note: It is safe to use any unsigned values for `start` or `stop` and any nonzero integer for `step`.
+                Any portion of the provided range that is outside the range of valid indices for the CeedVector will be ignored.
+
+  @param[in] vec    CeedVector to view
+  @param[in] start  Index of first CeedVector entry to view
+  @param[in] stop   Index of last CeedVector entry to view
+  @param[in] step   Step between CeedVector entries to view
+  @param[in] fp_fmt Printing format
+  @param[in] stream Filestream to write to
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedVectorViewRange(CeedVector vec, CeedSize start, CeedSize stop, CeedInt step, const char *fp_fmt, FILE *stream) {
+  const CeedScalar *x;
+  char              fmt[1024];
+
+  if (step == 0) return CeedError(vec->ceed, CEED_ERROR_MINOR, "View range 'step' must be nonzero");
+
+  fprintf(stream, "CeedVector length %ld\n", (long)vec->length);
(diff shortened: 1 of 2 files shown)


interface/ceed-vector.c
```diff
@@ -940,27 +940,46 @@ int CeedVectorReciprocal(CeedVector vec) {
 /**
   @brief View a CeedVector
 
-  @param[in] vec    CeedVector to view
-  @param[in] fp_fmt Printing format
-  @param[in] stream Filestream to write to
+  @param[in] vec         CeedVector to view
+  @param[in] num_entries Number of CeedVector entries to view
```
jedbrown, line 944: Maybe `CeedVectorViewRange()` taking a range (start, end, step)?

karenlstengel: I like this a lot. 
I think this will help a lot with mid-development debugging/sanity checking

jeremylt: Good idea, I like that better

interface/ceed-vector.c
```diff
@@ -940,27 +940,49 @@ int CeedVectorReciprocal(CeedVector vec) {
...
+  @param[in] fp_fmt      Printing format
+  @param[in] stream      Filestream to write to
 
   @return An error code: 0 - success, otherwise - failure
 
   @ref User
 **/
-int CeedVectorView(CeedVector vec, const char *fp_fmt, FILE *stream) {
+int CeedVectorViewRange(CeedVector vec, CeedSize start, CeedSize stop, CeedSize step, const char *fp_fmt, FILE *stream) {
```
jedbrown, line 952: Looks like docs need an update.

interface/ceed-vector.c
```diff
@@ -940,27 +940,49 @@ int CeedVectorReciprocal(CeedVector vec) {
...
   const CeedScalar *x;
   CeedCall(CeedVectorGetArrayRead(vec, CEED_MEM_HOST, &x));
 
   char fmt[1024];
   fprintf(stream, "CeedVector length %ld\n", (long)vec->length);
+  if (start != 0 || stop != vec->length || step != 1) {
+    fprintf(stream, "  start: %ld\n  stop:  %ld\n  step:  %ld\n", (long)start, (long)stop, (long)step);
+  }
+  if (stop > vec->length) stop = vec->length;
```
jedbrown, line 961: Will we allow negative steps? Should we document that one can safely use `INT_MAX`?

interface/ceed-vector.c
jeremylt: oops, fixed

interface/ceed-vector.c
jeremylt: I added a note that only the subset of the range that's valid for the vector will be printed.

I wasn't thinking of using negative steps but we can add it if you anticipate it being a used.

jeremylt: Good point wrt Python's behavior. Adjusted

jrwrigh (approved):

PR #1179, opened 2023-03-14
fix(fluids): Destroy context after call to print_info

jrwrigh:
Currently most fluids example problems seg fault as the `print_info` function often reads QFunctionContexts. Reason it doesn't show up in tests is that `print_info` (eg `PRINT_NEWTONIAN`) isn't run in test mode.

This is just kicking the refactoring of the fluid structs down the road, but this currently causes (nearly) all fluids examples to seg fault, figured quick fix is good for now.

### DIFF
--- a/examples/fluids/navierstokes.c
+++ b/examples/fluids/navierstokes.c
@@ -159,23 +159,6 @@ int main(int argc, char **argv) {
     PetscCall(SetupStatsCollection(ceed, user, ceed_data, problem));
   }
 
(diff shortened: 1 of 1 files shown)


jedbrown (approved):
Didn't these move recently?

jrwrigh:
Yeah, they were moved in #1169, but leaving them in their original place would've also resulted in issues.

Also, looks like Noether CI for HIP is broken again.

PR #1181, opened 2023-03-17
doc: Fixes, Various formatting updates

jrwrigh:
Handful of documentation fixes/ formatting updates.


include/ceed/ceed.h
```diff
@@ -199,11 +199,11 @@ CEED_EXTERN const char *const CeedCopyModes[];
...
 /// @ingroup CeedVector
 typedef enum {
-  /// L_1 norm: sum_i |x_i|
+  /// \f$L_1\f$ norm: \f$\sum_i \vert x_i\vert\f$
   CEED_NORM_1,
-  /// L_2 norm: sqrt(sum_i |x_i|^2)
+  /// \f$L_2\f$ norm: \f$\sqrt{\sum_i \vert x_i \vert^2}\f$
   CEED_NORM_2,
-  /// L_Infinity norm: max_i |x_i|
+  /// \f$L_\infty\f$ norm: \f$\max_i \vert x_i \vert\f$
```
jedbrown, line 206: These are discrete 1-norm, 2-norm, $\infty$ norms. The [usual notation](https://en.wikipedia.org/wiki/Lp_space) is $L^p$ rather than $L_p$, though it's sometimes useful to distinguish the continuous norms (integrals) from these discrete norms.

interface/ceed-basis.c
```diff
@@ -1739,9 +1739,9 @@ CeedPragmaOptimizeOn
 
     /**
       @brief Return Simultaneous Diagonalization of two matrices.
-               This solves the generalized eigenvalue problem A x = lambda B x, where A and B are symmetric and B is positive definite.
-               We generate the matrix X and vector Lambda such that X^T A X = Lambda and X^T B X = I.
-               This is equivalent to the LAPACK routine 'sygv' with TYPE = 1.
+               This solves the generalized eigenvalue problem \f$A x = \lambda B x\f$, where \f$A\f$ and \f$B\f$ are symmetric and \f$B\f$ is positive
+    definite. We generate the matrix \f$X\f$ and vector \f$\lambda\f$ such that \f$X^T A X = \lambda\f$ and \f$X^T B X = I\f$. This is equivalent to
```
jedbrown, line 1743:
```suggestion
    definite. We generate the matrix \f$X\f$ and diagonal \f$\Lambda\f$ such that \f$X^T A X = \Lambda\f$ and \f$X^T B X = I\f$. This is equivalent to
```

interface/ceed-preconditioning.c
```diff
@@ -1537,8 +1537,10 @@ int CeedOperatorAssemblyDataDestroy(CeedOperatorAssemblyData *data) {
   @brief Assemble a linear CeedQFunction associated with a CeedOperator
 
   This returns a CeedVector containing a matrix at each quadrature point providing the action of the CeedQFunction associated with the CeedOperator.
-    The vector 'assembled' is of shape [num_elements, num_input_fields, num_output_fields, num_quad_points] and contains column-major matrices
-representing the action of the CeedQFunction for a corresponding quadrature point on an element. Inputs and outputs are in the order provided by the
+    The vector `assembled` is of shape [num_elements, num_input_fields, num_output_fields, num_quad_points] and contains column-major matrices
```
jedbrown, line 1540:
```suggestion
    The vector `assembled` is of shape `[num_elements, num_input_fields, num_output_fields, num_quad_points]` and contains column-major matrices
```

jedbrown (approved):

include/ceed/ceed.h
jrwrigh: Yeah, not sure how to distinguish the continuous from the discrete form other than writing out the definition there. I don't think there's a common notation to distinguish the difference (I thought it was $L^p$ vs $\ell^p$, but according to wikipedia that's not necessarily true).

jedbrown: Just saying "1-norm" or writing $\lVert x \rVert_1$ is common. The $\ell^1$ and $L^1$ and notation is usually for countably and uncountably infinite dimensional $x$.

interface/ceed-basis.c
```diff
@@ -1739,9 +1739,9 @@ CeedPragmaOptimizeOn
 
     /**
       @brief Return Simultaneous Diagonalization of two matrices.
-               This solves the generalized eigenvalue problem A x = lambda B x, where A and B are symmetric and B is positive definite.
-               We generate the matrix X and vector Lambda such that X^T A X = Lambda and X^T B X = I.
-               This is equivalent to the LAPACK routine 'sygv' with TYPE = 1.
+               This solves the generalized eigenvalue problem \f$A x = \lambda B x\f$, where \f$A\f$ and \f$B\f$ are symmetric and \f$B\f$ is positive
+    definite. We generate the matrix \f$X\f$ and diagonal \f$\Lambda\f$ such that \f$X^T A X = \Lambda\f$ and \f$X^T B X = I\f$. This is equivalent to
+    the LAPACK routine 'sygv' with `TYPE = 1`.
```
jeremylt, line 1744: Note - I try to use one sentence per line with the documentation so we can change one sentence and have it only effect that one line. The `clang-tidy` PR did break some of the docstrings wrt this convention though and I haven't had a chance to go fix them all.

jrwrigh: Yeah, that makes sense. I usually do that for my markdown/LaTeX files for the same reason.

jrwrigh: Speaking of the developer notes, I notice that include-what-you-use doesn't seem to be used. At least it complains about a lot of core files not having the correct includes. I know fluids is quite guilty of it too.

jeremylt: Its been a while since I ran IWYU. It produces some silly results sometimes, so its something I do manually.

PR #1182, opened 2023-03-17
IWYU fixes

jeremylt:
@jrwrigh I added some iwyu fixes in this PR. Here is a good example of flaky results:
```
          CC build/interface/ceed-basis.o

/home/jeremy/Dev/libCEED/interface/ceed-basis.c should add these lines:
#include "ceed/ceed-f64.h"  // for CeedScalar, CEED_EPSILON
#include "ceed/types.h"     // for CeedInt, CEED_ERROR_SUCCESS, CEED_ERROR_D...

/home/jeremy/Dev/libCEED/interface/ceed-basis.c should remove these lines:

The full include-list for /home/jeremy/Dev/libCEED/interface/ceed-basis.c:
#include <ceed-impl.h>      // for CeedBasis_private, Ceed_private, CeedVect...
#include <ceed/backend.h>   // for CeedCall, CeedFree, CeedCalloc, CeedMalloc
#include <ceed/ceed.h>      // for CeedBasis, CeedError, CeedIntPow, Ceed
#include <math.h>           // for fabs, sqrt, copysign, atan, cos
#include <stdbool.h>        // for bool, false
#include <stdio.h>          // for fprintf, NULL, fputs, FILE
#include <string.h>         // for memcpy
#include "ceed/ceed-f64.h"  // for CeedScalar, CEED_EPSILON
#include "ceed/types.h"     // for CeedInt, CEED_ERROR_SUCCESS, CEED_ERROR_D...
```
We don't actually want to directly include `"ceed\ceed-f64.h"` or `"ceed\ceed-types.h"` in this case, despite what IWYU says. Its intended that you get at those definitions via `<ceed.h>`.

---

I made a quick pass at updating the tests and examples for includes, but I didn't double check everything. Over in Ratel I switched to only including `<ceed.h>` and `<petsc.h>` since those contain almost every definition we need. The only exception is the occasional need for `<ceed-backend.h>`.


include/ceed/ceed.h
```diff
@@ -55,7 +55,7 @@
 #include <stdio.h>
 
 /// Typedefs and macros used in public interfaces and user QFunction source
-#include "types.h"
+#include "types.h"  // IWYU pragma: export
```
jeremylt, line 58: this should silence those warnings and make better output

jeremylt:
We can do more for the examples/tests in a follow-up, but this fixes all the IWYU issues in the core library.

jrwrigh:
> Over in Ratel I switched to only including <ceed.h> and <petsc.h> since those contain almost every definition we need.

Yeah, that's what I tried to do for the fluids example, but for awhile I didn't realize that `clangd` was adding in includes based on IWYU rather than what was strictly necessary.

jrwrigh (approved):

jedbrown:
Note that `petsc.h` is rather heavy since it sucks in *a lot*. In PETSc, we generally include the highest level thing that we actually need, and never include the very blunt `petsc.h`. I'd be fine with IWYU-style including every "package" that's actually needed (though the highest level brings already brings it in). Is there something we should do in PETSc to better support IWYU?

jeremylt:
My biggest annoyance with PETSc + IWYU is that IWYU recommends a half dozen or more PETSc includes per file and I haven't dug around to determine which ones include the others.

Adding the include pragma I am using here in ceed.h in the include files PETSc intends people to use might help identify which ones are intended to be top level includes.

jedbrown:
Is this anywhere close to clean for you? I got lots of noise for almost every file, including some wrong like '"cuda.h"` instead of `<cuda.h>`.
```console
$  make iwyu IWYU_CC=include-what-you-use
make -B CC=include-what-you-use
make[1]: Entering directory '/home/jed/src/libCEED'
make: 'lib' with optional backends: /cpu/self/memcheck/serial /cpu/self/memcheck/blocked /cpu/self/avx/serial /cpu/self/avx/blocked /cpu/self/xsmm/serial /cpu/self/xsmm/blocked /gpu/cuda/ref
 /gpu/cuda/shared /gpu/cuda/gen
          CC build/interface/ceed-cuda.o
         CXX build/backends/cuda/ceed-cuda-compile.o
         CXX build/backends/cuda-ref/ceed-cuda-ref-qfunction-load.o
         CXX build/backends/cuda-gen/ceed-cuda-gen-operator-build.o
          CC build/interface/ceed-basis.o
          CC build/interface/ceed.o
          CC build/interface/ceed-elemrestriction.o

/home/jed/src/libCEED/interface/ceed-cuda.c should add these lines:
#include <ceed.h>          // for CeedQFunctionGetCeed, CEED_ERROR_SUCCESS
#include "cuda.h"          // for CUfunc_st, CUfunction

/home/jed/src/libCEED/interface/ceed-cuda.c should remove these lines:
- #include <ceed/ceed.h>  // lines 10-10

The full include-list for /home/jed/src/libCEED/interface/ceed-cuda.c:
#include <ceed-impl.h>     // for CeedQFunction_private, Ceed_private
#include <ceed.h>          // for CeedQFunctionGetCeed, CEED_ERROR_SUCCESS
#include <ceed/backend.h>  // for CeedCall, CeedDebug
#include <ceed/cuda.h>     // for CeedQFunctionSetCUDAUserFunction
#include "cuda.h"          // for CUfunc_st, CUfunction
---
[...]
```

jeremylt:
Outside of CUDA and HIP, yes. I honestly forgot to check with CUDA/HIP since it gives so much false crap.

jedbrown (approved):
LGTM. I don't know how we'll enforce this if we don't have people add it to their editor/clangd profiles and confirm that in CI. I currently see a little noise about libxsmm. But I think there's no harm in merging this and perhaps we can enforce more once PETSc annotates its includes.

jeremylt:
Yeah, I've been hesitant to turn this into a CI job because of some of that noise.

PR #1184, opened 2023-03-24
Update op check for a context label

jeremylt:


include/ceed-impl.h
```diff
@@ -280,6 +280,8 @@ struct CeedContextFieldLabel_private {
   size_t                 offset;
   CeedInt                num_sub_labels;
   CeedContextFieldLabel *sub_labels;
+  CeedInt                num_operators;
+  CeedOperator          *op;
```
jeremylt, line 284: I should perhaps comment here in the code that this is only for checking compatibility and not in any way for calling any op methods

interface/ceed-operator.c
```diff
@@ -1541,7 +1559,16 @@ int CeedOperatorGetContextFieldLabel(CeedOperator op, const char *field_name, Ce
 
     return CEED_ERROR_SUCCESS;
   } else {
-    return CeedQFunctionContextGetFieldLabel(op->qf->ctx, field_name, field_label);
+    CeedCall(CeedQFunctionContextGetFieldLabel(op->qf->ctx, field_name, field_label));
+    if (*field_label) {
+      CeedInt num_operators = (*field_label)->num_operators + 1;
+
+      if (num_operators == 1) CeedCall(CeedCalloc(num_operators, &(*field_label)->op));
```
jrwrigh, line 1566: nit-pick, but
```suggestion
      if ((*field_label)->num_operators == 0) CeedCall(CeedCalloc(num_operators, &(*field_label)->op));
```

Would work right? Seems a bit more straight forward to me.

jrwrigh (approved):

interface/ceed-operator.c
```diff
@@ -1487,6 +1500,9 @@ int CeedOperatorGetContextFieldLabel(CeedOperator op, const char *field_name, Ce
     CeedCall(CeedCompositeOperatorGetSubList(op, &sub_operators));
     CeedCall(CeedCalloc(num_sub, &new_field_label->sub_labels));
     new_field_label->num_sub_labels = num_sub;
+    new_field_label->num_operators  = 1;
+    CeedCall(CeedCalloc(1, &new_field_label->op));
+    new_field_label->op[0] = op;
```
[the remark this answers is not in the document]
jrwrigh: Is there not a reason to just check that the operator contains the label? ie. that `field_label` is in `op->context_labels[]`? If I understand correctly, the loop is coming from the fact that Operator's already track their associated context labels.

interface/ceed-operator.c
jeremylt: six of one, half dozen of the other since `num_operators = (*field_label)->num_operators`

interface/ceed-operator.c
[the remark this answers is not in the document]
jeremylt: The reason is I forgot that I did that, lol. Let me look again this afternoon

interface/ceed-operator.c
```diff
@@ -1804,9 +1852,11 @@ int CeedOperatorDestroy(CeedOperator *op) {
   CeedCall(CeedQFunctionDestroy(&(*op)->dqf));
   CeedCall(CeedQFunctionDestroy(&(*op)->dqfT));
   // Destroy any composite labels
-  for (CeedInt i = 0; i < (*op)->num_context_labels; i++) {
-    CeedCall(CeedFree(&(*op)->context_labels[i]->sub_labels));
-    CeedCall(CeedFree(&(*op)->context_labels[i]));
+  if ((*op)->is_composite) {
```
jrwrigh, line 1855: Not sure why this would only apply to composite operators. Non-composite operators won't have labels with sub_labels, but they could have multiple labels, right?

jrwrigh (approved):

interface/ceed-operator.c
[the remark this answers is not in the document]
jrwrigh: > We may wish to make a ContextFieldLabel a proper object in a followup?

In what way is it not a proper object now?

interface/ceed-operator.c
jeremylt: It only applies to composite operators because non-composite labels are owned by the QFunctionContext

interface/ceed-operator.c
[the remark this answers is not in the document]
jeremylt: No constructor, destructor, or recounting currently

interface/ceed-operator.c
jrwrigh: Ah, gotcha. Makes sense.

PR #1189, opened 2023-04-06
fluids: Add differential filtering capability

jrwrigh:
Adds the ability to filter the solution field by solving a differential equation. Particularly useful for doing data-driven subgrid-stress model training.

This includes isotropic, homogenous filtering (apply the same isotropic filter kernel to the entire domain) and anisotropic inhomogenous filtering (applies an filter based on the element sizes of the mesh).

TODO:
- [x] Document options
- [x] Document theory


jrwrigh:
Only change here was to rebase onto main in order to get the updates in #1225 . Still good for review.

jedbrown (approved):
LGTM, sorry about my slow review.

PR #1191, opened 2023-04-07
fix: Update DMGetRegionDS calls in examples/petsc

jrwrigh:
PETSc 3.19 introduced a changed function signature for `DMGetRegionDS`, which causes the petsc examples to fail on compile. This updates the required petsc version as well.


jeremylt:
If I recall correctly, we may have some other compatibility marcos that we don't need if we set 3.19 as a minimum

jrwrigh:
Yep, removed those now

doc/sphinx/source/releasenotes.md
```diff
@@ -17,6 +17,12 @@ For example, `CeedOperatorContextGetFieldLabel` was renamed to `CeedOperatorGetC
 - Update `/cpu/self/memcheck/*` backends to help verify `CeedVector` array access assumptions and `CeedQFunction` user output assumptions.
 - Update {c:func}`CeedOperatorLinearAssembleDiagonal` to provide default implementation that supports `CeedOperator` with multiple active bases.
 
+### Examples
+
+#### {ref}`example-petsc-bps`
+
+- Requires PETSc version >3.19
```
jeremylt, line 24: I think this is usually how we phrase this
```suggestion
- Requires PETSc version 3.19 or later
```

jrwrigh:
Can't say for `examples/solids/`, but `examples/fluids/` runs fine without on 3.19 without changes. Not sure if we want to be overly restrictive on that.

jeremylt:
Generally my philosophy has been to keep the same min version for every PETSc based example

jrwrigh:
Gotcha. That'll let me move to using `PETSC_SUCCESS`, so I'm not unopposed. Might as well do that in this PR?

PR #1194, opened 2023-04-12
CI: Run apt update before installing clang-format

jrwrigh:


.github/workflows/c-fortran-test-style.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: Environment setup
       uses: actions/checkout@v3
     - name: Install clang-format
-      run: sudo apt install clang-format-15
+      run: sudo apt update && sudo apt install clang-format-15
```
jedbrown, line 22: Was it failing without this? How did it pass before? (Seems weird for it to stop working.)

jrwrigh: Yep, it was failing before adding that. See here for example: https://github.com/CEED/libCEED/actions/runs/4680826007/jobs/8294412612

As for why it was passing before and not now, I'm guessing that ubuntu (or whoever controls the package repository) updated the clang-format package and the docker image has the wrong package database information.

jedbrown: Oh, I see. The current version is `15.0.7-0ubuntu0.22.04.1` and servers are no longer mirroring the old package versions.

jedbrown (approved):

PR #1196, opened 2023-04-17
Unsigned element restriction for correct operator diagonal assembly

### LINKED ISSUE #1195: Operator diagonal assembly after https://github.com/CEED/libCEED/pull/889
https://github.com/CEED/libCEED/pull/889 introduced `CeedElemRestrictionCreateOriented` in order to support H(div) and low-order `p=1` H(curl) elements. I ran into an issue realizing that this breaks `CeedOperatorLinearAssembleDiagonal`, due to the final `CeedElemRestrictionApply` (transpose) call at the end of `CeedSingleOperatorAssembleAddDiagonal_Core`. For an oriented basis, this needs to be an "unsigned" apply (transpose) in order to give the correct diagonal entries. Likewise signs on the element restriction need to be handled elsewhere in `ceed-preconditioning.c` like in full assembly.

This will be even further complicated by a solution like https://github.com/CEED/libCEED/pull/1168 for high-order H(curl) support and I haven't quite figured out the best way to get the operator diagonal in this case. In this case, off-diagonal entries of `B^T D B` are going to contribute to the operator diagonal due to pairs of triangular face dofs for Nedelec spaces of order `p > 1`.

For now, I have two questions:

1. For the oriented element restriction case, should we add an `CeedElemRestrictionApplyUnsigned` method which defaults to just `CeedElemRestrictionApply` but for `CeedElemRestrictionCreateOriented` will perform an apply without the orientation signs? This will touch a lot of code and so I wanted to see if this is a good approach before starting. An alternative I see would be to modify the function signature of `CeedElemRestrictionApply` to include an `unsigned` flag but this might be worse.
2. For high-order H(curl) support, do you have any ideas of a best way to approach the issue of diagonal assembly? It seems as though in general we will need to assemble a matrix product `T^T (B^T D B) T` where `T` is the tri-diagonal part of the element restriction accounting for orientations and coupling of pairs of face dofs. I think because of the block structure of `T` (not only is it tri-diagonal, but actually it is diagonal + block 2x2 for the pairs of face degrees of freedom), this means we can get away with constructing only the tri-diagonal part of `B^T D B`. But I don't know how such an interface might be handled such that it can be abstracted away easily enough for the user who just wants to call `CeedOperatorLinearAssembleDiagonal`. I'd appreciate any suggestions or input here.

Thank you!

sebastiangrimberg:
This PR attempts to resolve #1195 for an element restriction constructed using `CeedElemRestrictionCreateOriented`, and adds a test for H(div) mass from [this branch](https://github.com/CEED/libCEED/tree/rezgar/Hdiv-diagonal-assembly) (thank you @rezgarshakeri) 

Correctly handling #1168 is more complicated and likely should only allow for approximate diagonal assembly using this same `CeedElemRestrictionApplyUnsigned` approach. This is the same strategy used, for example, in MFEM for parallel non-conforming meshes where the `P` matrix has a form more complicated than being a simple mapping of local dof to parallel true dof.

### DIFF
--- a/backends/ref/ceed-ref-restriction.c
+++ b/backends/ref/ceed-ref-restriction.c
@@ -16,8 +16,8 @@
 // Core ElemRestriction Apply Code
 //------------------------------------------------------------------------------
 static inline int CeedElemRestrictionApply_Ref_Core(CeedElemRestriction r, const CeedInt num_comp, const CeedInt blk_size, const CeedInt comp_stride,
-                                                    CeedInt start, CeedInt stop, bool use_orient, CeedTransposeMode t_mode, CeedVector u, CeedVector v,
-                                                    CeedRequest *request) {
+                                                    CeedInt start, CeedInt stop, bool use_orient, CeedTransposeMode t_mode, CeedVector u,
+                                                    CeedVector v, CeedRequest *request) {
   CeedElemRestriction_Ref *impl;
   CeedCallBackend(CeedElemRestrictionGetData(r, &impl));
   const CeedScalar *uu;
@@ -81,8 +81,7 @@ static inline int CeedElemRestrictionApply_Ref_Core(CeedElemRestriction r, const
           CeedPragmaSIMD for (CeedInt i = 0; i < elem_size * blk_size; i++) {
             if (!use_orient || !impl->orient) {
               // Unsigned restriction
-              vv[elem_size * (k * blk_size + num_comp * e) + i - v_offset] =
-                  uu[impl->offsets[i + elem_size * e] + k * comp_stride];
+              vv[elem_size * (k * blk_size + num_comp * e) + i - v_offset] = uu[impl->offsets[i + elem_size * e] + k * comp_stride];
             } else {
               // Signed restriction
               vv[elem_size * (k * blk_size + num_comp * e) + i - v_offset] =
@@ -138,8 +137,7 @@ static inline int CeedElemRestrictionApply_Ref_Core(CeedElemRestriction r, const
             for (CeedInt j = i; j < i + CeedIntMin(blk_size, num_elem - e); j++) {
               if (!use_orient || !impl->orient) {
                 // Unsigned restriction
-                vv[impl->offsets[j + e * elem_size] + k * comp_stride] +=
-                    uu[elem_size * (k * blk_size + num_comp * e) + j - v_offset];
+                vv[impl->offsets[j + e * elem_size] + k * comp_stride] += uu[elem_size * (k * blk_size + num_comp * e) + j - v_offset];
               } else {
                 // Signed restriction
                 vv[impl->offsets[j + e * elem_size] + k * comp_stride] +=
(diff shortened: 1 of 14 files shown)


backends/ref/ceed-ref-restriction.c
```diff
@@ -79,8 +79,14 @@ static inline int CeedElemRestrictionApply_Ref_Core(CeedElemRestriction r, const
...
-            vv[elem_size * (k * blk_size + num_comp * e) + i - v_offset] =
-                uu[impl->offsets[i + elem_size * e] + k * comp_stride] * (is_oriented && impl->orient[i + elem_size * e] ? -1. : 1.);
+            if (!use_orient || !impl->orient) {
+              // Unsigned restriction
+              vv[elem_size * (k * blk_size + num_comp * e) + i - v_offset] = uu[impl->offsets[i + elem_size * e] + k * comp_stride];
+            } else {
+              // Signed restriction
+              vv[elem_size * (k * blk_size + num_comp * e) + i - v_offset] =
+                  uu[impl->offsets[i + elem_size * e] + k * comp_stride] * (impl->orient[i + elem_size * e] ? -1.0 : 1.0);
+            }
```
jeremylt, line 89: We should double check for a performance impact here. This tends to be an operation that we spend a fair amount of time in and I don't think we've dug into making sure this code builds the best assembly it can.

sebastiangrimberg: Here's a small example which (if I am reading things correctly) shows this test being optimized away around optimization -O2: https://godbolt.org/z/9cPfEGqzr.

Of course I can run benchmarks too. Are there any in particular you'd like me to run?

jeremylt: Thanks for so thoroughly checking - looks like any difference is negligible to me

interface/ceed-elemrestriction.c
```diff
@@ -630,6 +630,52 @@ int CeedElemRestrictionApply(CeedElemRestriction rstr, CeedTransposeMode t_mode,
...
+                     n);
+    // LCOV_EXCL_STOP
+  }
+  if (m != ru->length) {
+    // LCOV_EXCL_START
+    return CeedError(rstr->ceed, CEED_ERROR_DIMENSION,
+                     "Output vector size %" CeedInt_FMT " not compatible with element restriction (%" CeedInt_FMT ", %" CeedInt_FMT ")", ru->length,
+                     m, n);
+    // LCOV_EXCL_STOP
+  }
```
jeremylt, line 671: New approach with `CeedCheck` macro:
```suggestion
   CeedCheck(n == u->length, rstr->ceed, CEED_ERROR_DIMENSION,
                     "Input vector size %" CeedInt_FMT " not compatible with element restriction (%" CeedInt_FMT ", %" CeedInt_FMT ")", u->length, m,
                     n);
  CeedCheck(m == ru->length, rstr->ceed, CEED_ERROR_DIMENSION,
                     "Output vector size %" CeedInt_FMT " not compatible with element restriction (%" CeedInt_FMT ", %" CeedInt_FMT ")", ru->length,
                     m, n);
```

interface/ceed-elemrestriction.c
```diff
@@ -630,6 +630,52 @@ int CeedElemRestrictionApply(CeedElemRestriction rstr, CeedTransposeMode t_mode,
...
+  if (rstr->num_elem > 0) {
+    if (rstr->ApplyUnsigned) CeedCall(rstr->ApplyUnsigned(rstr, t_mode, u, ru, request));
+    else CeedCall(rstr->Apply(rstr, t_mode, u, ru, request));
+  }
+  return CEED_ERROR_SUCCESS;
+}
```
jeremylt, line 677: I'm not sure we want to fallback to `Apply` here? `ApplyUnsigned` is more specific of a request from the caller than `Apply`, so I think we need to make sure we fulfill that specific request. I think we want this fallback to work the other way around:

In that case, by default every backend would implement `ApplyUnsigned`. If `Apply` is not implemented, then `CeedElemRestrictionApply` would fall back to using `ApplyUnsigned`.

sebastiangrimberg: This is an interesting point. I did not think about this much but my logic for the current fallback structure is the following used in diagonal assembly: User calls `ApplyUnsigned` in this specific case (and nowhere else in the library currently), but since most methods don't have a signed vs. unsigned distinction this just falls back to `Apply`. I imagined people writing new backends would see `ApplyUnsigned` as a special case which typically doesn't require special handling and thus should fall back to the default `Apply`.

Of course I guess the reverse which you propose is the same thing which may make more sense in the literal sense of the methods. Let me know and I'll change it throughout the code if desired.

interface/ceed-elemrestriction.c
sebastiangrimberg: Missed that one during rebase, fixed in [57d0363](https://github.com/CEED/libCEED/pull/1196/commits/57d03632a4275bf84e2c35de7a183a4fc8e95d18).

interface/ceed-elemrestriction.c
jeremylt: A simpler option that's in the same spirit as my thoughts above would be to delete the else line, add a check for `ApplyUnsigned` at the start of the  function like `Apply` has, and then add lines like
```
  CeedCallBackend(CeedSetBackendFunction(ceed, "ElemRestriction", r, "ApplyUnsigned", CeedElemRestrictionApply_Cuda));
```
to the GPU backends.

This way the backend developer is making an affirmative statement about what their unsigned implementation is rather than letting the interface infer. That would be simpler than my first suggestion while still remaining mildly paranoid about the potential for future errors.

sebastiangrimberg: Nice. Done in [b17517e](https://github.com/CEED/libCEED/pull/1196/commits/b17517ee98e70bbde11768f0b2de0644e9c68f9a), let me know if I got that right.

jeremylt (approved):
LGTM

PR #1199, opened 2023-04-19
Fix HIP include order

jeremylt:
This out of order include was triggering the warning for the header `hipblas.h`


jrwrigh (approved):
Nice. Odd how the include order would control whether a deprecation warning pops up.

jeremylt:
`HIP_VERSION` wasn't defined yet before the `#if` checking its value with the previous order

PR #1202, opened 2023-04-20
Clean up backend headers for const and argument names

jeremylt:
Small consistency changes


backends/cuda-ref/ceed-cuda-ref.h
```diff
@@ -115,19 +115,19 @@ CEED_INTERN int CeedVectorCreate_Cuda(CeedSize n, CeedVector vec);
 
 CEED_INTERN int CeedElemRestrictionCreate_Cuda(CeedMemType mem_type, CeedCopyMode copy_mode, const CeedInt *indices, CeedElemRestriction r);
 
-CEED_INTERN int CeedElemRestrictionCreateBlocked_Cuda(const CeedMemType mem_type, const CeedCopyMode copy_mode, const CeedInt *indices,
-                                                      const CeedElemRestriction res);
+CEED_INTERN int CeedElemRestrictionCreateBlocked_Cuda(const CeedMemType mem_type, CeedCopyMode copy_mode, const CeedInt *indices,
```
jrwrigh, line 118: Not sure I understand why `mem_type` is `const`ed, but `copy_mode` isn't. Shouldn't they both be not modified by the function?

backends/hip-ref/ceed-hip-ref.h
```diff
@@ -116,19 +116,18 @@ CEED_INTERN int CeedVectorCreate_Hip(CeedSize n, CeedVector vec);
 
 CEED_INTERN int CeedElemRestrictionCreate_Hip(CeedMemType mtype, CeedCopyMode cmode, const CeedInt *indices, CeedElemRestriction r);
 
-CEED_INTERN int CeedElemRestrictionCreateBlocked_Hip(const CeedMemType mtype, const CeedCopyMode cmode, const CeedInt *indices,
-                                                     const CeedElemRestriction res);
+CEED_INTERN int CeedElemRestrictionCreateBlocked_Hip(CeedMemType mtype, CeedCopyMode cmode, const CeedInt *indices, CeedElemRestriction res);
```
jrwrigh, line 119: Might as well change `mtype`->`mem_type` and `cmode` -> `copy_mode` for consistencies sake?

backends/cuda-ref/ceed-cuda-ref.h
jeremylt: woops, mem_type shouldn't be consted, thanks

backends/hip-ref/ceed-hip-ref.h
jeremylt: updated

backends/ref/ceed-ref.h
```diff
@@ -30,8 +30,7 @@ typedef struct {
   // Orientation, if it exists, is true when the face must be flipped (multiplies by -1.).
   const bool *orient;
   bool       *orient_allocated;
-  int (*Apply)(CeedElemRestriction, const CeedInt, const CeedInt, const CeedInt, CeedInt, CeedInt, CeedTransposeMode, CeedVector, CeedVector,
-               CeedRequest *);
+  int (*Apply)(CeedElemRestriction, CeedInt, CeedInt, CeedInt, CeedInt, CeedInt, CeedTransposeMode, CeedVector, CeedVector, CeedRequest *);
```
sebastiangrimberg, line 33: I must be missing something here so I am looking for some help, but why does this work when the definition for `CeedElemRestrictionApply_Ref_Core` is still using the previous signature with `const CeedInt`s in `backends/ref/ceed-ref-restriction.c`?

jrwrigh: @jeremylt can probably explain it better than I can, but I believe the `const` in the function definition only applies to within the function context itself. Since the `CeedInt` parameters are pass-by-value, having a `const` on their signature is meaningless; the value is copied and independent of the variable passed to the function when called. This [SO](https://softwareengineering.stackexchange.com/a/204720/373632) answer covers it fairly well I think.

sebastiangrimberg: This makes sense, `CeedInt` -> `const CeedInt` is a valid compatible conversion on call to the function. Thanks!

PR #1206, opened 2023-04-28
Add CeedElemRestrictionCreateUnsignedCopy

jeremylt:
I need to double check the design, but I think this should do what we need.


jeremylt:
Note to self: I may have a leek on this branch. Need to investigate.

jeremylt:
Ok, "leek" found and fixed

sebastiangrimberg:
I like this idea and wonder if it can be used to simplify #1196 by removing the `CeedElemRestrictionApplyUnsigned` interface function added there, having the user instead create a copy (`CeedElemRestrictionCreateUnsignedCopy`) and then just use `CeedElemRestrictionApply`. This would remove the additions to the API which are now unnecessary.

jeremylt:
I'd be inclined to move the function to the `ceed/backend.h` interface if we don't intend it to be used in general anymore.

jeremylt:
Ok, moved. Really, I should cook up a test for `CeedElemRestrictionApplyUnsigned` so that there is an easy point of debugging when the GPU backends start adding signed restrictions.

include/ceed-impl.h
```diff
@@ -149,7 +149,8 @@ struct CeedVector_private {
 };
 
 struct CeedElemRestriction_private {
-  Ceed ceed;
+  Ceed                ceed;
+  CeedElemRestriction rstr_signed;
```
jrwrigh, line 153: If I follow correctly, `rstr_signed` will be `NULL` by default, but will be filled iff we ask for an unsigned copy. In that case, the unsigned copy has a reference to its signed copy pair. 

So are we effectively using `rstr_signed == NULL` as the restriction being "unsigned"? Or how else might someone differentiate between an "signed" and "unsigned" restriction?

Also, if there's a `CeedElemRestriction->ApplyUnsigned` method, why is there a need for creating the copy with  `CeedElemRestriction->ApplyUnsigned`? I'm guessing this was hashed out in a separate PR/thread somewhere, but just throwing the question out there more than anything else.

jeremylt: The idea here is that this is mostly used internally as a way to simplify diagonal assembly and multigrid operators.

H1 uses unsigned restrictions, but Hdiv and Hcurl need signed/oriented restrictions. But its messy to special case the logic for applying these restrictions unsigned where needed for the preconditioning ingredients. This lets us create a shallow clone of the restriction that acts like a H1 restriction instead of Hdiv or Hcurl.

jrwrigh: Ah, ok. So there are cases where we want both `rstr->Apply` and `rstr->ApplyUnsigned` to be used, but in different contexts?

jeremylt: Yeah, usually we want Apply, but sometimes we need to temporarily set the default to ApplyUnsigned. The alternative would be to somehow make operator fields know which to use.

sebastiangrimberg: I have been thinking about this a lot and one way to get around the only use of this unsigned clone approach (which is really only needed for calculating the E-vector multiplicity for used multigrid prolongation and restriction) is to instead just have the user construct this multiplicity vector as an L-vector themselves (outside of libCEED) and apply it outside of the `CeedOperator`. That is, the prolongation operator is simply `P = diag(m)^-1 P_ceed` where `m` is the vector of multiplicities of the L-vector entries.

So, we could just throw a warning or error in `CeedSingleOperatorMultigridLevel` if the user has provided a fine `ElemRestriction` object that has orientations that this only works for standard H1 or L2 unsigned restrictions and be on our way without this PR.

jeremylt: Yeah, D'd like to avoid that if we can do so cleanly. The goal of the multigrid set-up functions was to avoid exactly that sort of boilerplate on the users end.

sebastiangrimberg: Fair enough, I agree but I really do not see how this might be handled in the framework of https://github.com/CEED/libCEED/pull/1168 for high-order H(curl) restrictions. In this case even the unsigned `|E||E|^T|E|` will give an incorrect multiplicity for each E-vector dof. It saved me a lot of headache to just write the very small function inside MFEM to compute the L-vector dof multiplicities and just apply the scaling on the outside of the `CeedOperator` for the prolongation and restriction.

jeremylt: We can also use this fix for the situations where it helps and use more complicated fixes as needed in the future.

sebastiangrimberg: Indeed this will fix the issue for H(div) and p = 1 H(curl) which just use `CeedElemRestrictionCreateOriented`.

tests/t221-elemrestriction.c
```diff
@@ -0,0 +1,69 @@
...
+    ind[2 * i + 0] = i;
+    ind[2 * i + 1] = i + 1;
+    // flip the dofs on element 1,3,...
+    orient[2 * i + 0] = (i % (2)) * -1 < 0;
+    orient[2 * i + 1] = (i % (2)) * -1 < 0;
+  }
+  CeedElemRestrictionCreateOriented(ceed, num_elem, p, dim, 1, num_elem + 1, CEED_MEM_HOST, CEED_USE_POINTER, ind, orient, &elem_restriction);
+
+  CeedElemRestrictionApply(elem_restriction, CEED_NOTRANSPOSE, x, y_oriented, CEED_REQUEST_IMMEDIATE);
+  CeedElemRestrictionApplyUnsigned(elem_restriction, CEED_NOTRANSPOSE, x, y_unsigned, CEED_REQUEST_IMMEDIATE);
```
sebastiangrimberg, line 36: My only comment before approving is that this test could also check `CeedElemRestrictionCreateUnsignedCopy` by copying the original `elem_restriction` and ensuring the result of `CeedElemRestrictionApply` matches the unsigned apply? If that is superfluous just say so and I will approve.

jeremylt: I'm a fan of belt and suspenders! Done

sebastiangrimberg (approved):

PR #1210, opened 2023-05-04
Fix bug in `CeedBasisCreateProjectionMatrices` for non-H^1 basis

sebastiangrimberg:
This was a mistake from https://github.com/CEED/libCEED/pull/1156. When the input coarse and fine bases are not H^1, the resulting prolongation basis should still be (non-tensor) H^1 so that `CEED_EVAL_INTERP` treats `P_fine x P_coarse` dimensional `B` matrix correctly.


jeremylt:
This looks good to me, but @jedbrown and @rezgarshakeri have more experience with these spaces, so I'd like to have one of them :+1: as well.

rezgarshakeri:
> This looks good to me, but @jedbrown and @rezgarshakeri have more experience with these spaces, so I'd like to have one of them +1 as well.

Yeah, that looks good, thanks.

jeremylt (approved):

PR #1211, opened 2023-05-08
Fix OperatorApply Bug

jeremylt:
Fixes the bug identified in Ratel. See discussion here: https://gitlab.com/micromorph/ratel/-/merge_requests/464

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -1469,24 +1469,7 @@ active outputs
 int CeedOperatorApply(CeedOperator op, CeedVector in, CeedVector out, CeedRequest *request) {
   CeedCall(CeedOperatorCheckReady(op));
 
-  if (op->num_elem) {
-    // Standard Operator
-    if (op->Apply) {
-      CeedCall(op->Apply(op, in, out, request));
-    } else {
-      // Zero all output vectors
-      CeedQFunction qf = op->qf;
-      for (CeedInt i = 0; i < qf->num_output_fields; i++) {
-        CeedVector vec = op->output_fields[i]->vec;
-        if (vec == CEED_VECTOR_ACTIVE) vec = out;
-        if (vec != CEED_VECTOR_NONE) {
-          CeedCall(CeedVectorSetValue(vec, 0.0));
-        }
-      }
-      // Apply
-      CeedCall(op->ApplyAdd(op, in, out, request));
-    }
-  } else if (op->is_composite) {
+  if (op->is_composite) {
     // Composite Operator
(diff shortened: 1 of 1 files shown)


interface/ceed-operator.c
```diff
@@ -1513,6 +1494,21 @@ int CeedOperatorApply(CeedOperator op, CeedVector in, CeedVector out, CeedReques
...
+    } else {
+      // Zero all output vectors
+      CeedQFunction qf = op->qf;
+      for (CeedInt i = 0; i < qf->num_output_fields; i++) {
+        CeedVector vec = op->output_fields[i]->vec;
+        if (vec == CEED_VECTOR_ACTIVE) vec = out;
+        if (vec != CEED_VECTOR_NONE) CeedCall(CeedVectorSetValue(vec, 0.0));
+      }
+      // Apply
+      if (op->num_elem) CeedCall(op->ApplyAdd(op, in, out, request));
```
jrwrigh, line 1510: Do we want to catch if we get to this point and `op->num_elem` is NULL? Seems weird that we might get this far, not actually apply the operator, but not return an error or something.

interface/ceed-operator.c
```diff
@@ -1552,6 +1545,9 @@ int CeedOperatorApplyAdd(CeedOperator op, CeedVector in, CeedVector out, CeedReq
         CeedCall(CeedOperatorApplyAdd(sub_operators[i], in, out, request));
       }
     }
+  } else if (op->num_elem) {
+    // Standard Operator
+    CeedCall(op->ApplyAdd(op, in, out, request));
```
jrwrigh, line 1550: Similar here. Not sure the likelihood of both `op->is_composite` and `op->num_elem` being NULL though.

jrwrigh (approved):

interface/ceed-operator.c
jeremylt: `op->num_elem` is set to `-1` initially. Having `op->num_elem == 0` means that the operator has zero elements on the current process (think like a boundary condition operator)

jeremylt: By accepting 0 elements on the current process as valid, we need a small amount of special case logic in libCEED but prevent the caller from having to wrap similar special case logic around all of their operators.

jeremylt: Found it! Here's the original bugfix that added this logic: https://github.com/CEED/libCEED/pull/383

jrwrigh: Digging that PR from way back. Haha.

But gotcha, makes sense.

PR #1218, opened 2023-05-20
ci: update to rocm-5.5

jedbrown:


jedbrown:
CI passed even without MAGMA, but codecov went down. When I click through to the branch, it shows high coverage everywhere except occa, but reports reduced coverage here. If you don't immediately know why, we could merge this so other CI can run and revisit occa coverage another time.

jeremylt:
I think there is some strangeness going on with CodeCov - it often reports higher or lower coverage on branches that have no changes in lines covered

jeremylt (approved):

PR #1221, opened 2023-05-25
No ctx access after restoring

jeremylt:
Bug found by @rezgarshakeri - QFunctionContext data should not be accessible after restoring


rezgarshakeri (approved):
Thanks

PR #1203, opened 2023-04-20
fluids: SmartSim Online Data-Driven SGS training

jrwrigh:
Todo:
- [x] Documentation
- [x] Send subgrid stresses (not just filtered velocity products)
    - [x] Add SGS to dd training input DM
    - [x] Add calculation of subgrid stress to training input calculation
    - [x] Verify DM Vec data placement (if true, remove IS and copying mechanics)
- [x] Add test

Depends on #1223

PR #1223, opened 2023-05-25
Fluids: Decompose SGS DD input and output processing

jrwrigh:
- Extract out the SGS DD input and output processing (ie. the rotation to/from S-frame and un-dimensionalizing)
    - This is particularly useful for _in situ_ ML training and for incorporating external inference libraries (rather than computing the neural network manually).
    
#1203 depends on this PR

### DIFF
--- a/examples/fluids/qfunctions/sgs_dd_utils.h
+++ b/examples/fluids/qfunctions/sgs_dd_utils.h
@@ -1,0 +1,101 @@
+// Copyright (c) 2017-2023, Lawrence Livermore National Security, LLC and other CEED contributors.
+// All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+//
+// This file is part of CEED:  http://github.com/ceed
+
+/// @file
+/// Structs and helper functions for data-driven subgrid-stress modeling
+/// See 'Invariant data-driven subgrid stress modeling in the strain-rate eigenframe for large eddy simulation' 2022 and 'S-frame discrepancy
+/// correction models for data-informed Reynolds stress closure' 2022
+
+#ifndef sgs_dd_utils_h
+#define sgs_dd_utils_h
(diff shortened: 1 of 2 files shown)


jeremylt (approved):
This looks reasonable to me. Not sure if there is someone else we should have review?

examples/fluids/qfunctions/sgs_dd_utils.h
```diff
@@ -0,0 +1,145 @@
...
+#include "utils_eigensolver_jacobi.h"
+
+// @brief Calculate the inverse of the multiplicity, reducing to a single component
+CEED_QFUNCTION(InverseMultiplicity)(void *ctx, CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out) {
+  const CeedScalar(*multiplicity)[CEED_Q_VLA] = (const CeedScalar(*)[CEED_Q_VLA])in[0];
+  CeedScalar(*inv_multiplicity)               = (CeedScalar(*))out[0];
+
+  CeedPragmaSIMD for (CeedInt i = 0; i < Q; i++) inv_multiplicity[i] = 1.0 / multiplicity[0][i];
+  return 0;
+}
```
jeremylt, line 30: It feels like a lot of work to spin up an operator just to invert a vector's contents?

jrwrigh:
> This looks reasonable to me. Not sure if there is someone else we should have review?

This is mostly a refactor, so I think it's fine. The SmartSim PR should have more eyes on it though.

examples/fluids/qfunctions/sgs_dd_utils.h
jrwrigh: I'll go ahead and merge this, but let me know if you thing of a way to get out  a vector of single components.

jeremylt: We can probably add this to the interface - I'll think on the best way to do so

PR #1225, opened 2023-05-29
Fluids: Move to state enum for state functions

jrwrigh:
Instead of using function pointers to convert generic state vectors to `State` structs, use enums passed to a function. This is done to increase backend compatibility (particularly with SYCL/OpenCL).


jedbrown (approved):
Looks like this was based on my earlier patch and has (at least nominally) been tested with SYCL/OpenCL (this is technically an OpenCL limitation). I reset the GPU on Noether so it passes now.

jrwrigh:
Yep. This is primarily to ease running (ie. merging) new fluids developments as the SYCL backend is updated.

PR #1227, opened 2023-06-13
make: pass CPPFLAGS to HIPCC

jedbrown:
Otherwise users setting `HIPCCFLAGS` overrides the necessary directories that must appear in `-I` parameters.


jedbrown:
@nbeams @jeremylt PETSc sets `HIPCCFLAGS` based on its configure and user-specified options, thus was losing the `-I` flags that were being put into that variable. Preprocessor defines are intended to go in `CPPFLAGS` partly for this reason so that's what this PR does. It also removes `-I/opt/rocm/include` from the command-line for files that don't need HIP headers.

nbeams (approved):

PR #1229, opened 2023-06-13
Add sycl/ref and sycl/shared backends

jrwrigh:
- Adds backends for using SYCL programming environment
- Development done in from https://github.com/argonne-lcf/libCEED repo (specifically the `alcf-fork/sycl-backend`)
    - Work actually done predominantly by @kris-rowe @uumesh
    
    
TODOs:
- [x] Check for `CeedCheck` replacements
- [x] Add build documentation
- [ ] Verify tests pass on Sunspot


examples/fluids/navierstokes.c
```diff
@@ -45,11 +45,14 @@
...
 
 #include "navierstokes.h"
+#include <ceed/sycl.h>
+#include <petscdevice.h>
 
 #include <ceed.h>
 #include <petscdmplex.h>
 #include <petscts.h>
 
+int CeedSetStream_Sycl(Ceed ceed, void* handle) __attribute__((weak));
```
jrwrigh, line 55: I think it makes sense to put this somewhere in the backend itself. I'm not sure where though.

jrwrigh:
Partially also, there were other developments as "niceties" (such as `CeedCheck`). Do we want to comb through the code not to replace them (possibly error-prone), or leave that for latter? Part of me sides with fixing that up now, but I also know that we're trying to get this moving quickly.

examples/fluids/navierstokes.c
jeremylt: If this needs to be publicly viewable, I'd put it in `include/ceed/sycl.h`, like `include/ceed/cuda.h` and `include/ceed/hip.h`

The name isn't quite our convention. Maybe `CeedSetSyclStream`?

Makefile
```diff
@@ -438,6 +451,19 @@ ifneq ($(HIP_LIB_DIR),)
   BACKENDS_MAKE += $(HIP_BACKENDS)
 endif
 
+# SYCL Backends
```
jeremylt, line 454: Don't forget to add the magic decoder ring for user facing make options to the README

jeremylt:
Overall looks good - do we have a plan for CI yet?

jeremylt:
> Partially also, there were other developments as "niceties" (such as CeedCheck). Do we want to comb through the code not to replace them (possibly error-prone), or leave that for latter? Part of me sides with fixing that up now, but I also know that we're trying to get this moving quickly.

I replaced all them at once by replacing `ierr\s*=\s*` with `CeedCallBackend(` and `\);\s*CeedChkBackend\(ierr\);` with `));` in VSCode. It didn't take too long - minor clean-up after that. I'd do it in this MR or a follow-up before the end of the week.

examples/fluids/navierstokes.c
```diff
@@ -45,11 +45,14 @@
...
 
 #include "navierstokes.h"
+#include <ceed/sycl.h>
+#include <petscdevice.h>
 
 #include <ceed.h>
 #include <petscdmplex.h>
 #include <petscts.h>
 
+int CeedSetStream_Sycl(Ceed ceed, void *handle) __attribute__((weak));
```
jeremylt, line 55: You should be able to remove this since you have `ceed/sycl.h`

examples/fluids/navierstokes.c
```diff
@@ -45,11 +45,14 @@
 const char help[] = "Solve Navier-Stokes using PETSc and libCEED\n";
 
 #include "navierstokes.h"
+#include <ceed/sycl.h>
```
jeremylt, line 48: Probably should guard this include in case it requires sycl specific headers in the future

backends/sycl-ref/ceed-sycl-ref.hpp
```diff
@@ -0,0 +1,141 @@
...
+  CeedVector                *qvecsout;  // Output Q-vectors needed to apply operator
+  CeedInt                    numein;
+  CeedInt                    numeout;
+  CeedInt                    qfnumactivein, qfnumactiveout;
+  CeedVector                *qfactivein;
+  CeedOperatorDiag_Sycl     *diag;
+  CeedOperatorAssemble_Sycl *asmb;
+} CeedOperator_Sycl;
+
+// CEED_INTERN int CeedSyclGetCublasHandle(Ceed ceed, cublasHandle_t *handle);
```
jeremylt, line 115: Dead code

jedbrown:
I requested a JLSE GitLab-CI project for libceed. It lacks the C in CI, but those people with access will be able to bump it and get a report. From my perspective, once this looks reasonable in review, passes our existing CI, and someone can certify that the tests (including fluids) pass on Sunspot, we can merge. We'll add a `.gitlab-jlse-ci.yml` when that project is available.

backends/sycl/ceed-sycl-common.hpp
```diff
@@ -0,0 +1,50 @@
...
+  return CeedSetBackendFunction(ceed, type, object, fname, bf);
+}
+
+typedef struct {
+  sycl::context sycl_context;
+  sycl::device  sycl_device;
+  sycl::queue   sycl_queue;
+} Ceed_Sycl;
+
+CEED_INTERN int CeedSyclGetResourceRoot(Ceed ceed, const char *resource, char **resource_root);
```
jeremylt, line 42: Probably `CeedGetResourceRoot` should be a `ceed/backend.h` function in `interface/ceed.c`

backends/sycl-ref/ceed-sycl-ref.sycl.cpp
```diff
@@ -0,0 +1,66 @@
...
+// Backend Init
+//------------------------------------------------------------------------------
+static int CeedInit_Sycl(const char *resource, Ceed ceed) {
+  char *resource_root;
+  CeedCallBackend(CeedSyclGetResourceRoot(ceed, resource, &resource_root));
+  if (std::strcmp(resource_root, "/gpu/sycl/ref") && std::strcmp(resource_root, "/cpu/sycl/ref")) {
+    // LCOV_EXCL_START
+    return CeedError(ceed, CEED_ERROR_BACKEND, "Sycl backend cannot use resource: %s", resource);
+    // LCOV_EXCL_STOP
+  }
```
jeremylt, line 35:
```suggestion
  CeedCheck(!std::strcmp(resource_root, "/gpu/sycl/ref") || !std::strcmp(resource_root, "/cpu/sycl/ref"), ceed, CEED_ERROR_BACKEND, "Sycl backend cannot use resource: %s", resource);
```

backends/sycl-ref/ceed-sycl-restriction.sycl.cpp
```diff
@@ -0,0 +1,477 @@
...
+// Blocked not supported
+//------------------------------------------------------------------------------
+int CeedElemRestrictionApplyBlock_Sycl(CeedElemRestriction r, CeedInt block, CeedTransposeMode t_mode, CeedVector u, CeedVector v,
+                                       CeedRequest *request) {
+  // LCOV_EXCL_START
+  Ceed ceed;
+  CeedCallBackend(CeedElemRestrictionGetCeed(r, &ceed));
+  return CeedError(ceed, CEED_ERROR_BACKEND, "Backend does not implement blocked restrictions");
+  // LCOV_EXCL_STOP
+}
```
jeremylt, line 209: Should be removed - this check moved to the interface level

backends/sycl-ref/ceed-sycl-restriction.sycl.cpp
```diff
@@ -0,0 +1,477 @@
...
+
+//------------------------------------------------------------------------------
+// Blocked not supported
+//------------------------------------------------------------------------------
+int CeedElemRestrictionCreateBlocked_Sycl(const CeedMemType m_type, const CeedCopyMode copy_mode, const CeedInt *indices, CeedElemRestriction r) {
+  Ceed ceed;
+  CeedCallBackend(CeedElemRestrictionGetCeed(r, &ceed));
+  return CeedError(ceed, CEED_ERROR_BACKEND, "Backend does not implement blocked restrictions");
+}
```
jeremylt, line 476: See above, should be removed

backends/sycl-ref/ceed-sycl-vector.sycl.cpp
```diff
@@ -0,0 +1,760 @@
...
+  CeedVector_Sycl *impl;
+  CeedCallBackend(CeedVectorGetData(vec, &impl));
+  Ceed_Sycl *data;
+  CeedCallBackend(CeedGetData(ceed, &data));
+
+  if (!impl->d_array) {
+    // LCOV_EXCL_START
+    return CeedError(ceed, CEED_ERROR_BACKEND, "No valid device data to sync to host");
+    // LCOV_EXCL_STOP
+  }
```
jeremylt, line 91:
```suggestion
  CeedCheck(impl->d_array, ceed, CEED_ERROR_BACKEND, "No valid device data to sync to host");
```

backends/sycl-shared/ceed-sycl-shared-basis.sycl.cpp
```diff
@@ -0,0 +1,265 @@
...
+// Compute the local range of for basis kernels
+//------------------------------------------------------------------------------
+static int ComputeLocalRange(Ceed ceed, CeedInt dim, CeedInt thread_1d, CeedInt *local_range, CeedInt max_group_size = 256) {
+  local_range[0] = thread_1d;
+  local_range[1] = (dim > 1) ? thread_1d : 1;
+
+  const CeedInt min_group_size = local_range[0] * local_range[1];
+  if (min_group_size > max_group_size) {
+    return CeedError(ceed, CEED_ERROR_BACKEND, "Requested group size is smaller than the required minimum.");
+  }
```
jeremylt, line 29:
```suggestion
  CeedCheck(min_group_size <= max_group_size, ceed, CEED_ERROR_BACKEND, "Requested group size is smaller than the required minimum.");
```

backends/sycl/ceed-sycl-compile.sycl.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "ceed-sycl-compile.hpp"
+
+#include <ceed/backend.h>
+#include <ceed/ceed.h>
+#include <ceed/jit-tools.h>
+#include <level_zero/ze_api.h>
+
+#include <map>
+#include <sstream>
+// #include <sycl/ext/intel/experimental/online_compiler.hpp>
```
jeremylt, line 17: Dead line

backends/sycl/ocloc_api.h
```diff
@@ -0,0 +1,103 @@
+//===------- ocloc_api.h --------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// This file is copied from
+// https://github.com/intel/compute-runtime/blob/master/shared/offline_compiler/source/ocloc_api.h
```
jeremylt, line 10: How does this square with our license?

backends/sycl-shared/ceed-sycl-shared.sycl.cpp
```diff
@@ -0,0 +1,70 @@
...
+// Backend init
+//------------------------------------------------------------------------------
+static int CeedInit_Sycl_shared(const char *resource, Ceed ceed) {
+  char *resource_root;
+  CeedCallBackend(CeedSyclGetResourceRoot(ceed, resource, &resource_root));
+  if (std::strcmp(resource_root, "/gpu/sycl/shared") && std::strcmp(resource_root, "/cpu/sycl/shared")) {
+    // LCOV_EXCL_START
+    return CeedError(ceed, CEED_ERROR_BACKEND, "Sycl backend cannot use resource: %s", resource);
+    // LCOV_EXCL_STOP
+  }
```
jeremylt, line 27:
```suggestion
  CeedCheck(!std::strcmp(resource_root, "/gpu/sycl/shared") || !std::strcmp(resource_root, "/cpu/sycl/shared"), ceed, CEED_ERROR_BACKEND, "Sycl backend cannot use resource: %s", resource);
```

backends/sycl/ceed-sycl-common.hpp
jeremylt: Maybe with the function signature `CeedGetResourceRoot(Ceed ceed, char **resource_root)` or `CeedGetResourceRoot(Ceed ceed, const char *deliminator, char **resource_root)`

backends/sycl/ocloc_api.h
jedbrown: Apache-2.0 with LLVM-exception is functionally equivalent to Rust's MIT or Apache-2.0. MIT and BSD-2-Clause are also compatible so I think it's fine (we must preserve the statement).

backends/sycl-ref/ceed-sycl-ref.sycl.cpp
jrwrigh: Yeah, these `CeedCheck` replacements is what I was referring to earlier, not `CeedCallBackend`. I think that's being used everywhere (or at least the `ierr = ...; CEEDCHK` isn't being used.

jeremylt: Yeah, I clued in belatedly to that. Looks like otherwise its all pretty consistent with the other backends

backends/sycl-ref/ceed-sycl-ref-qfunction-load.sycl.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "ceed-sycl-ref.hpp"
+
+#define SUB_GROUP_SIZE_QF 16
+
+//------------------------------------------------------------------------------
+// Build QFunction kernel
+//
+// TODO: Refactor
+//------------------------------------------------------------------------------
+extern "C" int CeedSyclBuildQFunction(CeedQFunction qf) {
```
jeremylt, line 29: This name we should fix after merging - its consistent with the CUDA and HIP backends, which both violate our usual convention, which would be `CeedQFunctionBuild_Sycl` or similar

jrwrigh:
> someone can certify that the tests (including fluids) pass on Sunspot

I can do that as soon as nodes on SunSpot frees up. Do we care about the other example tests passing? I can obviously check to see if they will, but I'd anticipate needing to make changes to solids that are similar to the changes for fluids.

jeremylt:
I wouldn't worry too much about solids but its worth it if the fixes are simple - we really should simplify that code now that most of it is in Ratel. Adding the solids fixes here or in a follow-up would let us know how much work we'd need to do to run Ratel on Intel cards

jedbrown:
It'll need to pass when we add JLSE CI, but I'm inauguration with cutting that corner for now. It will also be interesting to get Ratel running there.

examples/fluids/navierstokes.c
jrwrigh: Wouldn't that contradict with putting `CeedGetStream_Sycl ... __attribute__((weak))` into that header then?

jrwrigh: `CeedGetStream_Sycl ...__attribute__((weak))` is there to prevent errors at link time when the Sycl backend isn't compiled (and thus the actual implementation for `CeedGetStream_Sycl` doesn't exist). If we put `CeedGetStream_Sycl ...__attribute__((weak))` into `ceed/sycl.h` and include guards, then instances without a sycl backend will still have linking errors (because `CeedGetStream_Sycl` wouldn't be defined anywhere anymore).

jeremylt: Ah, I see. Skip the guard and we'll handle the future when it gets here?

jrwrigh: Sounds fine by me. Not sure if this can/should be generalized to include other backends (if the concept exists of course).

backends/sycl/ceed-sycl-common.hpp
jrwrigh: Ok if I leave that for a later PR?

jeremylt: Sure, if we need to merge this today

jrwrigh: IDK about today, but soon. I'm guessing by your proposal that other backends would also need to implement a `CeedGetResourceRoot` method?

jeremylt: I don't see this as a backend specific thing? That's why I suggested moving the entire function to ceed.c instead of making it only a sycl backend function. I don't see any special logic that only applies to sycl in this function, except maybe the specific delineator, which can be passed as an argument.

This is something other GPU backbends have discussed wanting

backends/sycl-ref/ceed-sycl-restriction.sycl.cpp
jrwrigh: When you say "this check moved to the interface level", do you mean that if `ApplyBlock` isn't set via `CeedSetBackendFunctionCpp`, it'll auto error-out saying that it can't do that?

jeremylt: Exactly, with an appropriate message that CI interprets as SKIP

backends/sycl/ceed-sycl-common.hpp
jrwrigh: It looks like all the GPU backends have their own version of `Ceed[ARCH]GetResourceRoot`, which is kind of why I suggested another PR (let this one be just SYCL, then we can merged all the GPU backends together in a separate change).

jeremylt: Ah, I hadn't realized the idea had already invaded in this fractured fashion

Makefile
```diff
@@ -95,7 +96,7 @@ endif
 AFLAGS = -fsanitize=address #-fsanitize=undefined -fno-omit-frame-pointer
 
 # Note: Intel oneAPI C/C++ compiler is now icx/icpx
-CC_VENDOR := $(subst icc_orig,icc,$(subst oneAPI,icc,$(firstword $(filter gcc clang icc icc_orig oneAPI XL emcc,$(subst -, ,$(shell $(CC) --version))))))
+CC_VENDOR := $(firstword $(filter gcc clang icc oneAPI XL emcc,$(subst -, ,$(shell $(CC) --version))))
```
jrwrigh, line 99: I'm not sure what to do about the bludgeoning of this line. The commit from the alcf-fork has a diff of:

```diff
-CC_VENDOR := $(subst icc_orig,icc,$(subst oneAPI,icc,$(firstword $(filter gcc clang icc icc_orig oneAPI XL emcc,$(subst -, ,$(shell $(CC) --version))))))
+CC_VENDOR := $(firstword $(filter gcc clang icc oneAPI XL emcc,$(subst -, ,$(shell $(CC) --version))))
```

which seemingly doesn't add anything? But I'm also not wise in the ways of `Make`, so I'm not sure if there was something actually wrong with the original line.

@kris-rowe do you remember why you made this change? (commit f9ab9a2a1 in the ALCF fork)

jedbrown (approved):
Looks good. I agree with Jeremy's plan for merging.

- [x] Update `releasenotes.md`.

jeremylt:
Oh, don't forget the README.md at the top level too

examples/fluids/navierstokes.c
jeremylt: Ah, I see. Maybe we put that on the list to tweek in a follow up.

James, can you make an issue with all the things we want to follow up on?

Makefile
jedbrown: It looks like we lost translation of `icc_orig`, which we've seen at some sites for their icc (classic, not icx) compilers. Can we restore that?

Does icx use clang-style defaults or is `-fp-model=fast` default? Is it okay to use clang(/gcc) options syntax (`-ffp-model`) instead of icc's?

examples/fluids/navierstokes.c
jrwrigh: I had the thought that this might be with intention of having other GPU backends recreate the same functionality, but wasn't sure if the concept translated across hardware. 

That being said, I actually already changed the name over, but we can change it back if/when we make the global interface.

jrwrigh: > James, can you make an issue with all the things we want to follow up on?

https://github.com/CEED/libCEED/issues/1230

README.md
```diff
@@ -157,6 +158,10 @@ There are multiple supported backends, which can be selected at runtime in the e
 | `/gpu/hip/shared`          | Optimized pure HIP kernels using shared memory    | Yes                   |
 | `/gpu/hip/gen`             | Optimized pure HIP kernels using code generation  | No                    |
 ||
+| **SYCL Native**            |
+| `/gpu/sycl/ref`            | Reference pure SYCL kernels                       | Yes                   |
+| `/gpu/sycl/shared`         | Optimized pure SYCL kernels using shared memory   | Yes                   |
+||
 | **MAGMA**                  |
```
jeremylt, line 165: Should also add a brief description down on line 196 or so

Makefile
jrwrigh: > It looks like we lost translation of `icc_orig`, which we've seen at some sites for their icc (classic, not icx) compilers. Can we restore that?

Yeah, that should be fixed now.

> Does icx use clang-style defaults or is `-fp-model=fast` default? Is it okay to use clang(/gcc) options syntax (`-ffp-model`) instead of icc's?

Not sure about that.

README.md
```diff
@@ -157,6 +158,10 @@ There are multiple supported backends, which can be selected at runtime in the e
 | `/gpu/hip/shared`          | Optimized pure HIP kernels using shared memory    | Yes                   |
 | `/gpu/hip/gen`             | Optimized pure HIP kernels using code generation  | No                    |
 ||
+| **SYCL Native**            |
+| `/gpu/sycl/ref`            | Reference pure SYCL kernels                       | Yes                   |
+| `/gpu/sycl/shared`         | Optimized pure SYCL kernels using shared memory   | Yes                   |
```
jeremylt, line 163: Are there also CPU impls? `ceed-backend-list.h` says yes?

jrwrigh: Oh, I guess there are? I haven't heard anything about them. They're also not built when building with SYCL.

backends/ceed-backend-list.h
```diff
@@ -19,6 +19,8 @@ CEED_BACKEND(CeedRegister_Cuda_Shared, 1, "/gpu/cuda/shared")
 CEED_BACKEND(CeedRegister_Hip, 1, "/gpu/hip/ref")
 CEED_BACKEND(CeedRegister_Hip_Gen, 1, "/gpu/hip/gen")
 CEED_BACKEND(CeedRegister_Hip_Shared, 1, "/gpu/hip/shared")
+CEED_BACKEND(CeedRegister_Sycl, 2, "/gpu/sycl/ref", "/cpu/sycl/ref")
+CEED_BACKEND(CeedRegister_Sycl_Shared, 2, "/gpu/sycl/shared", "/cpu/sycl/shared")
```
jeremylt, line 23: ah, then we should update here to drop that for now if we don't have those as options yet

backends/ceed-backend-list.h
```diff
@@ -19,6 +19,8 @@ CEED_BACKEND(CeedRegister_Cuda_Shared, 1, "/gpu/cuda/shared")
 CEED_BACKEND(CeedRegister_Hip, 1, "/gpu/hip/ref")
 CEED_BACKEND(CeedRegister_Hip_Gen, 1, "/gpu/hip/gen")
 CEED_BACKEND(CeedRegister_Hip_Shared, 1, "/gpu/hip/shared")
+CEED_BACKEND(CeedRegister_Sycl, 2, "/gpu/sycl/ref")
+CEED_BACKEND(CeedRegister_Sycl_Shared, 2, "/gpu/sycl/shared")
```
jeremylt, line 23:
```suggestion
CEED_BACKEND(CeedRegister_Sycl, 1, "/gpu/sycl/ref")
CEED_BACKEND(CeedRegister_Sycl_Shared, 1, "/gpu/sycl/shared")
```

jrwrigh: I'm not great at multi-tasking...

backends/ceed-backend-list.h
jedbrown: Do we actually not have those? The dispatch code for `/cpu/sycl` stuff is still present and I recall we tested it early in this development process. I'd hope it still runs, and in fact, we might be able to test it on noether (using the `/opt/intel/oneapi` stack).

jrwrigh: They're not present when I build on Sunspot. I'm not sure if there's just a single line that got commented out/deleted to remove that functionality.

jrwrigh: Would I just need to add the cpu backends to the `SYCL_BACKENDS` list in the Makefile to get it?

jedbrown: What do you mean by "not present"? Not reported in `make info` or fails saying there is no such backend when running with `-ceed /cpu/sycl`?

jrwrigh: Not reported by make info.

jedbrown: Let's try dropping that last commit and see if it runs. Could also test on noether. Then we can add it to `SYCL_BACKENDS` so it shows up in `make info`.

jrwrigh: libCEED tests with `/cpu/sycl/*` backends report:

```
SKIP: sycl device type not available fluids-navierstokes /cpu/sycl/shared
```

Attempting to run a test outside of the testing framework (ie. `build/t570-operator /cpu/sycl/shared`) results in:
```
/home/jrwrigh/software/libCEED/backends/sycl/ceed-sycl-common.sycl.cpp:46 in CeedSyclInit(): No SYCL devices of the requested type are available
```

So it still might work, the backend is just blocking itself from trying.

*Side note:* Should that be `/cpu/self/sycl/*`? I see some cpu backends including `self` and others not. Can't quite see the pattern between the two.

kris-rowe: Currently the SYCL runtime compilation (even for qfunctions in sycl-ref) is hardwired for use with the Level Zero runtime, so the `/cpu/sycl/` backends should definitely be dropped from the Makefile and registration now. This can be relaxed in the future.

README.md
```diff
@@ -63,6 +63,7 @@ if your compiler does not support gcc-style options, if you are cross compiling,
 
 To enable CUDA support, add `CUDA_DIR=/opt/cuda` or an appropriate directory to your `make` invocation.
 To enable HIP support, add `ROCM_DIR=/opt/rocm` or an appropriate directory.
+To enable SYCL support, add `SYCL_DIR=/opt/sycl` or an appropriate directory.
```
jedbrown, line 66: Don't we need to explicitly set compilers? This links for me on noether. In the other backends, we work out any necessary includes so that `CC` and `CXX` don't need to be special, but it's needed here. That's okay for this PR, but we should state that oneAPI compilers should be used for everything when building a SYCL backend.
```console
$ . /opt/intel/oneapi/setvars.sh
$ make SYCL_DIR=/opt/intel/oneapi/compiler/latest/linux SYCLCXX=icpx CC=icx CXX=icpx
```

jedbrown (approved):
Looks good to merge with a better statement in the readme.

README.md
jrwrigh: I haven't been explicitly setting compilers on Sunspot builds, but that's probably because `CC` and `CXX` are set correctly already.

PR #1233, opened 2023-06-18
Add CeedOperatorLinearAssemblePointBlockDiagonalSymbolic

### LINKED ISSUE #1232: Create `CeedOperatorLinearAssemblePointBlockDiagonalSymbolic`
*Mostly just notes for myself as I go about trying to make this.*

Goal is to make 

#### Role of `CeedOperatorCreateActivePointBlockRestriction`
This just creates an element restriction that has `num_comp**2` components at each node for storing the matrix at each node. This is used in `CeedSingleOperatorAssembleAddDiagonal_Core` primarily to assemble the element-level point-block-diagonals into the global point-block diagonals. It's also used to create a `CeedVector` of the correct size to store element-level pb-diagonals.

#### Breakdown of what `CeedOperatorLinearAssemblePointBlockDiagonal` does
*Assume that backend implementation is not set, use composite operator*

`CeedOperatorLinearAssemblePointBlockDiagonal` -> `CeedOperatorLinearAssembleAddPointBlockDiagonal` ->  `CeedCompositeOperatorLinearAssembleAddDiagonal` -> Loop over sub operators -> `CeedOperatorLinearAssembleAddPointBlockDiagonal` -> Operator is not longer composite -> `CeedSingleOperatorAssembleAddDiagonal_Core`

#### Breakdown of what `CeedSingleOperatorAssembleAddDiagonal_Core` does
*Assume that we assemble the point-block representation*

- Computes assembled QFunction as matrix $D_e$ for ever $e$ element
- Loop over all the active bases
    - Create point-block element restriction (`pointblock_rstr`) for the active element restriction. This has the same number of elements, number of nodes per element, but changes the number of components to be the square of that value.
    - Computes point-block of $B^T D_e B$ (the element matrix) and stores the result at each node as `num_comps**2` components.
    - Assembles the element-matrix components into the local-vector using the `pointblock_rstr`.

jrwrigh:
Create COO mapping for the result from `CeedOperatorLinearAssemblePointBlockDiagonal`.

Closes #1232


interface/ceed-preconditioning.c
```diff
@@ -518,6 +518,77 @@ static int CeedSingleOperatorAssembleSymbolic(CeedOperator op, CeedInt offset, C
...
+
+  CeedInt num_nodes = input_size / num_output_components;
+  for (CeedSize n = 0; n < num_nodes; n++) {
+    for (CeedInt i = 0; i < num_output_components; i++) {
+      for (CeedInt j = 0; j < num_output_components; j++) {
+        (*rows)[(n * num_output_components + i) * num_output_components + j] = n * num_output_components + i;
+        (*cols)[(n * num_output_components + i) * num_output_components + j] = n * num_output_components + j;
+      }
+    }
+  }
```
jrwrigh, line 588: Note that this may not be accurate for element restrictions with component stride != 1. Currently, the output of `CeedOperatorLinearAssemblePointBlockDiagonal` has components collapsed together, regardless of the component stride for the active element restriction.

I'm not sure if `CeedOperatorLinearAssemble` and `CeedOperatorLinearAssemble` does the same, or if it tried to assemble it into a matrix that would operate directly on the L-vector's ordering.

jedbrown: We need it to either be calculated correctly for general stride or to error with an explanation that it's not currently supported (and we'll add it if/when there is demand).

jrwrigh: I think for right now, erroring out with an explanation is probably the way to go until I get time to do a full implementation.

Thinking outloud, wouldn't this also assume successive offsets as well? I guess I should probably check that `offsets[i+1] > offsets[i]` in addition to component stride length.

jedbrown: I don't think we require positive stride, but do you mean `offsets[n+1] > offsets[n]` (index by node)? That needn't be guaranteed either.

jrwrigh:
I've gone ahead and made the general implementation.

I've also left in the check for `comp_stride` to verify that all active element restrictions have the same one (in a separate commit). I'm not sure if that's actually necessary though; seems like something that's required for the active fields of an CeedOperator and is checked there.

For documentation's sake, here's the document where I worked out the COO mapping:
[Point-block diagonal symbolic assembly.pdf](https://github.com/CEED/libCEED/files/11887170/Point-block.diagonal.symbolic.assembly.pdf)

jrwrigh:
Looking at `CeedOperatorSetField`, we only verify that that L-vector sizes are the same between active element restrictions, nothing about component strides. Should we try to enforce that? (in another PR)

jedbrown:
I'm not sure how because you can have different fields with different strides.

jrwrigh:
Even for active fields? Those should all be operating on the same L-vector data. Possibly subsets of that L-vector, but the vector will have the same structure. Different active fields can have different number of components, or even index to a subset of components, but the component stride will always be the same.

jedbrown:
Yeah, it may be unnatural (and not what `DMPlex` chooses), but I don't think libCEED should prevent having one field interlaced and the other blocked.

jrwrigh:
Gotcha. I can't decide whether or not active element restrictions with different numbers of components/component stride breaks the diagonal or point-block assembly code or not, though I lean towards that it probably would. I think addressing that would require a much larger overhaul of both functions to get that working though.

jedbrown:
It depends how the linear algebra defines "blocks". In any case, it's a permutation away from a "classic" block diagonal. I don't think it's libCEED's business to say.

jrwrigh:
Sure. My point is that currently the point-block diagonal assembly code puts all the components together into a single contiguous "block" without maintaining any sense of which components in a "limited" element restriction (e.g. skipping every other component, thus having half the number of components) correspond to the "full" L vector. So when doing the `CeedElemRestrictionApplyUnsigned` to assemble the blocks together, the correct elements of the point block won't correspond to each other. So I think it will fail to build that correctly.

I know that's a bit hand-wavy/vague. I can write it out if you'd like. Main point is that I think the current code will handle all the common cases, but some of the potential edge cases aren't handled.

jrwrigh:
Review of where this PR stands.

The current implementation of `CeedOperatorLinearAssemblePointBlockDiagonalSymbolic` cannot handle every possible case for L-vector representations. It does however verify that it's implementation will work with a given element restriction and will error out if it does not work. 

I think the conclusion that we came to was that the best way of having a general symbolic calculation was to actually change how the actual PBDiag assembly is done. 

Currently, the assembly outputs the resulting data in a point-major form (all the values of a block are contiguous) while also doing the restriction transpose (ie. element assembly). Comparatively, the `CeedOperatorLinearAssemble` assembles the element matrices, but does _not_ do the restriction transpose. Instead, the reduction of the restriction transpose is performed on the user side using the COO representation given by `CeedOperatorLinearAssembleSymbolic`. 

I think we'd want to move PBDiag assembly to something closer to the `CeedOperatorLinearAssemble`, but that would require changing all the backends code to reflect the change in format.

jeremylt:
I'd be fine merging this fix for the current impl and then going back and expanding the functionality of the current impl.

jrwrigh:
Ok. I think that'll also (kind of) make the transition to the new implementation a bit cleaner too. Since we don't have a symbolic assembly, any change to the actual assembly is necessarily a breaking change. It'll still be a breaking change for anyone who can't use this version of the symbolic assembly, but I expect that to be a very limited subset of users.

jrwrigh:
This currently isn't working right now for the fluids example for some reason. Possibly my checks are too strict.

- [ ] Also, note to self, I should refactor the checks for composite vs non-composite into a single set via pointer dereferencing

jrwrigh:
Ok, this is working fine now. I've also added a fluids test with `-amat_type shell -pmat_diagonal -pc_type vpbjacobi`, which wasn't tested before and should add an extra layer of testing for `CeedOperatorLinearAssemblePointBlockDiagonalSymbolic`.

tests/t537-operator.c
```diff
@@ -87,11 +90,17 @@ int main(int argc, char **argv) {
...
   CeedOperatorLinearAssemblePointBlockDiagonal(op_mass, assembled, CEED_REQUEST_IMMEDIATE);
+  CeedOperatorLinearAssemblePointBlockDiagonalSymbolic(op_mass, &num_entries, &rows, &cols);
 
-  // Manually assemble diagonal
+  // Manually assemble point-block diagonal
   CeedVectorCreate(ceed, num_comp * num_dofs, &u);
   CeedVectorSetValue(u, 0.0);
   CeedVectorCreate(ceed, num_comp * num_dofs, &v);
+  for (int i = 0; i < num_comp * num_dofs; i++) {
+    for (int j = 0; j < num_comp * num_dofs; j++) {
```
jeremylt, line 100:
```suggestion
  for (CeedInt i = 0; i < num_comp * num_dofs; i++) {
    for (CeedInt j = 0; j < num_comp * num_dofs; j++) {
```

jeremylt (approved):
Looks good to me!

tests/t537-operator.c
jeremylt: Currently every `t*` file uses `CeedInt`. We can change them all to use `int` instead, but that should be a separate PR if we want to do that.

jrwrigh: Whoops, I answered this through my phone and thought this was a fluids PR, not for the library. I'll accept the changes when I get back to my computer.

PR #1237, opened 2023-06-20
examples: Use PETSC_SUCCESS in PetscFunctionReturn

jrwrigh:


jeremylt (approved):
LGTM - can we also apply to `examples/petsc`?

jeremylt:
Cool, I think we can merge once Noether's done its thing

PR #1247, opened 2023-07-03
fluids: Add partition statistics to standard output

jrwrigh:
Add partition statistics to the normal fluids output. Effectively just calculates the min/max global/local `Vec` size associated with the DM. 

TODO:
- [x] Fix shared nodes
- [x] Add in median and/or mean statistics (probably using `PetscSortMPIInt` for the median)

### DIFF
--- a/examples/fluids/navierstokes.c
+++ b/examples/fluids/navierstokes.c
@@ -42,10 +42,11 @@
 /// @file
 /// Navier-Stokes example using PETSc
 
+#include "petscsf.h"
 const char help[] = "Solve Navier-Stokes using PETSc and libCEED\n";
 
-#include "navierstokes.h"
 #include <petscdevice.h>
+#include "navierstokes.h"
 
 #include <ceed.h>
 #include <petscdmplex.h>
@@ -280,26 +281,83 @@ int main(int argc, char **argv) {
     }
     // Mesh
     const PetscInt num_comp_q = 5;
-    PetscInt       glob_dofs, owned_dofs;
-    PetscInt       glob_nodes, local_nodes;
+    PetscInt       glob_dofs, owned_dofs, local_dofs;
     const CeedInt  num_P = app_ctx->degree + 1, num_Q = num_P + app_ctx->q_extra;
     // -- Get global size
-    PetscCall(VecGetSize(Q, &glob_dofs));
-    PetscCall(VecGetLocalSize(Q, &owned_dofs));
-    glob_nodes = glob_dofs / num_comp_q;
+    PetscCall(DMGetGlobalVectorInfo(user->dm, &owned_dofs, &glob_dofs, NULL));
     // -- Get local size
-    PetscCall(VecGetSize(user->Q_loc, &local_nodes));
-    local_nodes /= num_comp_q;
+    PetscCall(DMGetLocalVectorInfo(user->dm, &local_dofs, NULL, NULL));
     PetscCall(PetscPrintf(comm,
                           "  Mesh:\n"
                           "    Number of 1D Basis Nodes (P)       : %" CeedInt_FMT "\n"
                           "    Number of 1D Quadrature Points (Q) : %" CeedInt_FMT "\n"
                           "    Global DoFs                        : %" PetscInt_FMT "\n"
-                          "    Owned DoFs                         : %" PetscInt_FMT "\n"
                           "    DoFs per node                      : %" PetscInt_FMT "\n"
-                          "    Global nodes (DoFs / %" PetscInt_FMT ")            : %" PetscInt_FMT "\n"
-                          "    Local nodes                        : %" PetscInt_FMT "\n",
-                          num_P, num_Q, glob_dofs, owned_dofs, num_comp_q, num_comp_q, glob_nodes, local_nodes));
+                          "    Global nodes                       : %" PetscInt_FMT "\n",
+                          num_P, num_Q, glob_dofs, num_comp_q, glob_dofs / num_comp_q));
+    // -- Get Partition Statistics
+    PetscCall(PetscPrintf(comm, "  Partition:                             (min,max,median,ratio)\n"));
+    {
+      PetscInt    *gather_buffer = NULL;
+      PetscInt     part_owned_dofs[3], part_local_dofs[3], part_shared_dofs[3];
+      PetscInt     median_index = comm_size % 2 ? comm_size / 2 : comm_size / 2 - 1;
+      MPI_Datatype PetscInt_MPI;
+      PetscCall(PetscDataTypeToMPIDataType(PETSC_INT, &PetscInt_MPI));
+      if (!rank) PetscCall(PetscMalloc1(comm_size, &gather_buffer));
+
+      PetscCallMPI(MPI_Gather(&owned_dofs, 1, PetscInt_MPI, gather_buffer, 1, PetscInt_MPI, 0, comm));
+      if (!rank) {
+        PetscCall(PetscSortInt(comm_size, gather_buffer));
+        part_owned_dofs[0]             = gather_buffer[0];              // min
+        part_owned_dofs[1]             = gather_buffer[comm_size - 1];  // max
+        part_owned_dofs[2]             = gather_buffer[median_index];   // median
+        PetscReal part_owned_dof_ratio = (PetscReal)part_owned_dofs[1] / (PetscReal)part_owned_dofs[2];
+        PetscCall(PetscPrintf(comm, "    Owned free nodes                   : %" PetscInt_FMT ", %" PetscInt_FMT ", %" PetscInt_FMT ", %f\n",
+                              part_owned_dofs[0] / num_comp_q, part_owned_dofs[1] / num_comp_q, part_owned_dofs[2] / num_comp_q,
+                              part_owned_dof_ratio));
+      }
+
+      PetscCallMPI(MPI_Gather(&local_dofs, 1, PetscInt_MPI, gather_buffer, 1, PetscInt_MPI, 0, comm));
+      if (!rank) {
+        PetscCall(PetscSortInt(comm_size, gather_buffer));
+        part_local_dofs[0]             = gather_buffer[0];              // min
+        part_local_dofs[1]             = gather_buffer[comm_size - 1];  // max
+        part_local_dofs[2]             = gather_buffer[median_index];   // median
+        PetscReal part_local_dof_ratio = (PetscReal)part_local_dofs[1] / (PetscReal)part_local_dofs[2];
+        PetscCall(PetscPrintf(comm, "    Local nodes                        : %" PetscInt_FMT ", %" PetscInt_FMT ", %" PetscInt_FMT ", %f\n",
+                              part_local_dofs[0] / num_comp_q, part_local_dofs[1] / num_comp_q, part_local_dofs[2] / num_comp_q,
+                              part_local_dof_ratio));
+      }
+
+      PetscInt num_remote_roots_total = 0;
+      {
+        PetscSF            sf;
+        PetscInt           nrranks;
+        const PetscInt    *roffset, *rmine, *rremote;
+        const PetscMPIInt *rranks;
+        PetscCall(DMGetSectionSF(user->dm, &sf));
+        PetscCall(PetscSFGetRootRanks(sf, &nrranks, &rranks, &roffset, &rmine, &rremote));
+        for (PetscInt i = 0; i < nrranks; i++) {
+          if (rranks[i] == rank) continue;  // Ignore same-part global->local transfers
+          num_remote_roots_total += roffset[i + 1] - roffset[i];
+        }
+      }
+      PetscCallMPI(MPI_Gather(&num_remote_roots_total, 1, PetscInt_MPI, gather_buffer, 1, PetscInt_MPI, 0, comm));
+      if (!rank) {
+        PetscCall(PetscSortInt(comm_size, gather_buffer));
+        part_shared_dofs[0]             = gather_buffer[0];              // min
+        part_shared_dofs[1]             = gather_buffer[comm_size - 1];  // max
+        part_shared_dofs[2]             = gather_buffer[median_index];   // median
+        PetscReal part_shared_dof_ratio = (PetscReal)part_shared_dofs[1] / (PetscReal)part_shared_dofs[2];
+        PetscCall(PetscPrintf(comm, "    Shared nodes                       : %" PetscInt_FMT ", %" PetscInt_FMT ", %" PetscInt_FMT ", %f\n",
+                              part_shared_dofs[0] / num_comp_q, part_shared_dofs[1] / num_comp_q, part_shared_dofs[2] / num_comp_q,
+                              part_shared_dof_ratio));
+      }
+
+      if (!rank) PetscCall(PetscFree(gather_buffer));
+    }
+
+    PetscCall(PetscPrintf(comm, "(nodes == DoFs / %" PetscInt_FMT ")\n", num_comp_q));
   }
   // -- Zero Q_loc
   PetscCall(VecZeroEntries(user->Q_loc));
--- a/examples/fluids/navierstokes.c
+++ b/examples/fluids/navierstokes.c
@@ -42,11 +42,10 @@
 /// @file
 /// Navier-Stokes example using PETSc
 
-#include "petscsf.h"
 const char help[] = "Solve Navier-Stokes using PETSc and libCEED\n";
 
-#include <petscdevice.h>
 #include "navierstokes.h"
+#include <petscdevice.h>
 
 #include <ceed.h>
 #include <petscdmplex.h>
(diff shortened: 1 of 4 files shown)


examples/fluids/navierstokes.c
```diff
@@ -280,26 +281,50 @@ int main(int argc, char **argv) {
...
+      PetscCall(DMGetSectionSF(user->dm, &sf));
+      PetscCall(PetscSFGetLeafRanks(sf, &niranks, &iranks, &ioffset, &irootloc));
+      PetscInt shared_dofs = ioffset[niranks];
+      PetscInt part_shared_dofs[2];
+      sendbuf[0] = sendbuf[1] = shared_dofs;
+      PetscCall(PetscGlobalMinMaxInt(comm, sendbuf, part_shared_dofs));
+      PetscReal part_shared_dof_ratio = (PetscReal)part_shared_dofs[1] / (PetscReal)part_shared_dofs[0];
+      PetscCall(PetscPrintf(comm, "    Shared nodes (min,max,ratio)       : %" PetscInt_FMT ", %" PetscInt_FMT ", %f\n",
+                            part_shared_dofs[0] / num_comp_q, part_shared_dofs[1] / num_comp_q, part_shared_dof_ratio));
+    }
```
jrwrigh, line 327: Note that this currently doesn't work as expected. I'd think that the length of `irootloc` would be the number of remote leaves referencing local roots (which I'm fairly certain would be shared dofs), but the result of that has the values seemingly much too large:

```
 ❯ mpirun -n 2 ./navierstokes -options_file blasius.yaml -ts_monitor -snes_monitor -ts_max_steps 1 -petscpartitioner_view -dm_view -dm_plex_box_faces 41,61,1
Graph Partitioner: 2 MPI Processes
  type: simple
  edge cut: 0
  balance: 0
  use vertex weights: 1
DM Object: box 2 MPI processes
  type: plex
box in 3 dimensions:
  Number of 0-cells per rank: 2648 2646
  Number of 1-cells per rank: 6472 6467
  Number of 2-cells per rank: 5076 5072
  Number of 3-cells per rank: 1251 1250
Labels:
  depth: 4 strata with value/size (0 (2648), 1 (6472), 2 (5076), 3 (1251))
  marker: 1 strata with value/size (1 (10503))
  Face Sets: 5 strata with value/size (1 (1251), 2 (1251), 3 (41), 5 (30), 6 (31))
  celltype: 4 strata with value/size (0 (2648), 1 (6472), 4 (5076), 7 (1251))

  MPI:
    Host Name                          : fermat
    Total ranks                        : 2
  Problem:
  libCEED:
  PETSc:
    Box Faces                          : 41,61,1
  Mesh:
    Global DoFs                        : 20664
    Global nodes                       : 4132
  Partition:
    Owned free nodes (min,max,ratio)   : 2016, 2116, 1.050000
    Local nodes (min,max,ratio)        : 2646, 2648, 1.000756
(nodes == DoFs / 5)
    Shared nodes (min,max,ratio)       : 2016, 2185, 1.084127
```
Does the section `SF` map *all* non-fixed dofs rather than just the "necessary" ones?

Reference: [`PetscSFGetLeafRanks`](https://petsc.org/release/manualpages/PetscSF/PetscSFGetLeafRanks/)

jrwrigh: Actually, I think I'm overlooking that the section `SF` maps global `Vec`s to local `Vec`s rather than just the remote nodes. So I should discount the `ioffset` value that corresponds to the local rank?

examples/fluids/navierstokes.c
```diff
@@ -280,26 +281,50 @@ int main(int argc, char **argv) {
...
+                          "(nodes == DoFs / %" PetscInt_FMT ")\n",
+                          part_owned_dofs[0] / num_comp_q, part_owned_dofs[1] / num_comp_q, part_owned_dof_ratio, part_local_dofs[0] / num_comp_q,
+                          part_local_dofs[1] / num_comp_q, part_local_dof_ratio, num_comp_q));
+    {
+      PetscSF            sf;
+      PetscInt           niranks;
+      const PetscInt    *ioffset, *irootloc;
+      const PetscMPIInt *iranks;
+      PetscCall(DMGetSectionSF(user->dm, &sf));
+      PetscCall(PetscSFGetLeafRanks(sf, &niranks, &iranks, &ioffset, &irootloc));
```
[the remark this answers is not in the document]
jrwrigh: I was going for "what do I have that is owned by someone else", which I think is the same as "who owns stuff that I need".

I'm not sure if we want both though. I assume that the primary latency for MPI communication is "waiting to recieve" rather than "waiting to send", but I suppose if there's a significant concentration of sends in on a specific network switch, that could cause bottlenecks.

jrwrigh:
Fixed the shared nodes and added in the median statistic for everything too.

jrwrigh:
I'm also tempted extract out the printing stuff into a separate function inside `misc.c` since it's currently 360/490 lines of `navierstokes.c`.

jedbrown:
No objections from me on that plan.

examples/fluids/src/misc.c
```diff
@@ -465,3 +466,136 @@ PetscErrorCode IntArrayP2C(PetscInt num_entries, PetscInt **array_petsc, CeedInt
...
+                        "    Global nodes                       : %" PetscInt_FMT "\n",
+                        num_P, num_Q, glob_dofs, num_comp_q, glob_dofs / num_comp_q));
+  // -- Get Partition Statistics
+  PetscCall(PetscPrintf(comm, "  Partition:                             (min,max,median,max/median)\n"));
+  {
+    PetscInt    *gather_buffer = NULL;
+    PetscInt     part_owned_dofs[3], part_local_dofs[3], part_shared_dofs[3];
+    PetscInt     median_index = comm_size % 2 ? comm_size / 2 : comm_size / 2 - 1;
+    MPI_Datatype PetscInt_MPI;
+    PetscCall(PetscDataTypeToMPIDataType(PETSC_INT, &PetscInt_MPI));
```
jedbrown, line 545: This works. You can also just use `MPIU_INT`, which is defined appropriately in `petscsys.h` (which is always included).

examples/fluids/src/misc.c
```diff
@@ -465,3 +466,136 @@ PetscErrorCode IntArrayP2C(PetscInt num_entries, PetscInt **array_petsc, CeedInt
...
+    if (!rank) PetscCall(PetscMalloc1(comm_size, &gather_buffer));
+
+    PetscCallMPI(MPI_Gather(&owned_dofs, 1, PetscInt_MPI, gather_buffer, 1, PetscInt_MPI, 0, comm));
+    if (!rank) {
+      PetscCall(PetscSortInt(comm_size, gather_buffer));
```
jedbrown, line 550: This is not expensive for GPU runs because number of ranks is not huge. For >1M ranks, it's not trivial, but I suppose still not going to be big compared to other costs.

examples/fluids/src/misc.c
```diff
@@ -465,3 +466,136 @@ PetscErrorCode IntArrayP2C(PetscInt num_entries, PetscInt **array_petsc, CeedInt
...
+    PetscCallMPI(MPI_Gather(&num_remote_roots_total, 1, PetscInt_MPI, gather_buffer, 1, PetscInt_MPI, 0, comm));
+    if (!rank) {
+      PetscCall(PetscSortInt(comm_size, gather_buffer));
+      part_shared_dofs[0]             = gather_buffer[0];              // min
+      part_shared_dofs[1]             = gather_buffer[comm_size - 1];  // max
+      part_shared_dofs[2]             = gather_buffer[median_index];   // median
+      PetscReal part_shared_dof_ratio = (PetscReal)part_shared_dofs[1] / (PetscReal)part_shared_dofs[2];
+      PetscCall(PetscPrintf(comm, "    Shared nodes                       : %" PetscInt_FMT ", %" PetscInt_FMT ", %" PetscInt_FMT ", %f\n",
+                            part_shared_dofs[0] / num_comp_q, part_shared_dofs[1] / num_comp_q, part_shared_dofs[2] / num_comp_q,
+                            part_shared_dof_ratio));
```
jedbrown, line 592: A typical run on two processes.
```
  Partition:                             (min,max,median,max/median)
    Owned free nodes                   : 41859, 43941, 41859, 1.049738
    Local nodes                        : 46624, 48112, 46624, 1.031915
    Shared nodes                       : 0, 483, 0, inf
```
The `inf` is a little weird, but I guess not a problem.

I think we need better terminology than "Shared nodes" since that's ambiguous as to whether those are unowned halo nodes or nodes that I own and are referenced by another process. Those have similar but not identical information.

examples/fluids/src/misc.c
```diff
@@ -465,3 +466,136 @@ PetscErrorCode IntArrayP2C(PetscInt num_entries, PetscInt **array_petsc, CeedInt
...
+    }
+
+    if (!rank) PetscCall(PetscFree(gather_buffer));
+  }
+
+  PetscCall(PetscPrintf(comm, "(nodes == DoFs / %" PetscInt_FMT ")\n", num_comp_q));
```
jedbrown, line 598: This is just dangling here in the otherwise pseudo-yaml table.

jedbrown (commented):
I think we should add number of neigbors. This could go both ways or just report number of remote root ranks referenced by my leaves. Our partitioners should keep this bounded, but it's not directly measured by the other stats and parallel congestion is often caused by one-to-many communication graphs that are not coerced into using collectives for that pattern. We could put such partition stats into PETSc output instead.

examples/fluids/src/misc.c
jrwrigh: Yeah, I found quickselect would probably be faster, but I'd have to implement that. I figured that our problem size was small enough that it would really matter (plus it's only run once). 

Testing using numpy (used here as an upper bound), it takes about 326ms to sort 5 million 64bit ints. I'd say that's plenty fine for this usage.

examples/fluids/src/misc.c
jrwrigh: Fair. I had intended that each local partition could be evenly divided between Owned "free" nodes, Dirichlet nodes, and non-owned nodes. I guess "non-owned" nodes would be more clear than "shared" nodes?

examples/fluids/src/misc.c
jrwrigh: Yeah, I intentionally put that there. Rather than having to describe our ad-hoc definition of "nodes" multiple times. Any suggestions on a better place to put it?

examples/fluids/src/misc.c
jedbrown: Right, you have 3 sorts and 1 second would be too much for routine small runs, but in the noise for a truly huge run like that.

examples/fluids/src/misc.c
jedbrown: Perhaps halo nodes or ghost nodes?

examples/fluids/src/misc.c
jedbrown: I think it messes up the parse. We could call all the things `5-DoF nodes`, which is concise and still easy enough to parse when we have turbulence variables.

examples/fluids/src/misc.c
jrwrigh: I went with Ghost.

examples/fluids/src/misc.c
```diff
@@ -465,3 +466,151 @@ PetscErrorCode IntArrayP2C(PetscInt num_entries, PetscInt **array_petsc, CeedInt
...
+      for (PetscInt i = 0; i < nrranks; i++) {
+        if (rranks[i] == rank) continue;  // Ignore same-part global->local transfers
+        num_remote_roots_total += roffset[i + 1] - roffset[i];
+        num_neighbor_ranks++;
+      }
+      for (PetscInt i = 0; i < niranks; i++) {
+        if (iranks[i] == rank) continue;
+        num_neighbor_ranks++;
+      }
+    }
```
jrwrigh, line 588: I assume you were wanting all neighbors, right? Not just the number of neighbors owning ghost nodes?

jedbrown: We don't want to double-count. We have neighbors that we receive from in the Bcast (global-to-local) and those we receive from in the Gather (local-to-global). Those numbers are different and it's possible for one to be tiny while the other is huge (but a good partitioner shouldn't leave you that way). It wouldn't hurt to print both, or if we want to save lines, just print the larger of the two.

jrwrigh: Ah, I guess I assumed that data would waterfall rather than be. I'll do both.

jrwrigh: Is it ok to assume that the current rank is included? Such that I can just use `nrranks`/`niranks` without the for loop?

jedbrown: I think we want max and mean sizes so you'd have the loop anyway.

jrwrigh: I don't need a loop to get the per-rank statistics, just to get local value. The statistics are handled by the `MPI_Gather` and `PetscSortInt`.  The only reason I have a for loop right now is to exclude the remote roots on the current rank.

jedbrown: We have several messages per rank so you must need the loop of you want per-message stats, not just per-rank.

jrwrigh: I'm not sure I follow. Assuming the commit I just pushed is correct, I don't see a reason why I couldn't do
```diff
-      for (PetscInt i = 0; i < niranks; i++) {
-        if (iranks[i] == rank) continue;
-        num_L2Gneighbor_ranks++;
-      }
+ num_L2Gneighbor_ranks = niranks - 1;
```

if `rank` is always present in `iranks`.

jedbrown: That's just counting ranks, but I was thinking of message sizes (`roffset[i+1] - roffset[i]`).

jrwrigh: Ah, gotcha. Would we want to label them both as GlobalToLocal and LocalToGlobal directly then? Otherwise I can't think of an inverse name of "ghost".

jedbrown: Naming things is always the hardest. What about "Owned messages" and "Ghost messages"? Or "Messages as root" and "Messages as leaf"?

jrwrigh: I feel like "root" and "leaf" are less intuitive unless you're really familiar with SF (and I'm not sure if I'd put myself in that camp).

How about "Owned boundary nodes" and "Ghost boundary nodes", relying on the notion of partition boundaries (which is really what we're trying to quantify/communicate here).

jedbrown: s/boundary/interface/ is fine with me.

jrwrigh:
For `mpirun -n 4 ./navierstokes -options_file blasius.yaml -ts_monitor -snes_monitor -ts_max_steps 1 -dm_plex_box_faces 40,60,2` I now get:

```
  MPI:
    Host Name                          : fermat
    Total ranks                        : 4
  Problem:
  libCEED:
  PETSc:
    Box Faces                          : 40,60,2
  Mesh:
    Global DoFs                        : 32226
    Global 5-DoF nodes                 : 6445
  Partition:                             (min,max,median,max/median)
    Owned free 5-DoF nodes             : 1536, 1766, 1538, 1.148505
    Local 5-DoF nodes                  : 1920, 2064, 1938, 1.065015
    Ghost Interface 5-DoF nodes        : 0, 163, 59, 2.739130
    Ghost Interface Ranks              : 0, 3, 1, 3.000000
    Owned Interface 5-DoF nodes        : 0, 185, 44, 4.203620
    Owned Interface Ranks              : 0, 3, 1, 3.000000
```

(this is using Parmetis as the partitioner). All good now?

examples/fluids/src/misc.c
```diff
@@ -465,3 +466,175 @@ PetscErrorCode IntArrayP2C(PetscInt num_entries, PetscInt **array_petsc, CeedInt
...
+
+    PetscCallMPI(MPI_Gather(&local_dofs, 1, MPIU_INT, gather_buffer, 1, MPIU_INT, 0, comm));
+    if (!rank) {
+      PetscCall(PetscSortInt(comm_size, gather_buffer));
+      part_local_dofs[0]             = gather_buffer[0];              // min
+      part_local_dofs[1]             = gather_buffer[comm_size - 1];  // max
+      part_local_dofs[2]             = gather_buffer[median_index];   // median
+      PetscReal part_local_dof_ratio = (PetscReal)part_local_dofs[1] / (PetscReal)part_local_dofs[2];
+      PetscCall(PetscPrintf(
+          comm, "    Local %" PetscInt_FMT "-DoF nodes                  : %" PetscInt_FMT ", %" PetscInt_FMT ", %" PetscInt_FMT ", %f\n", num_comp_q,
```
jrwrigh, line 566: Actually, not to keep bikeshedding, but maybe this would be better as "Partition nodes"? Feel like that might be more clear than "local".

jedbrown: `LocalVector` has a precise meaning in PETSc and I think it's okay to use that. "Owned free" could be "Global" for consistency with `GlobalVector`.

jedbrown: That would be fine with me.

jrwrigh:
Alright, should be good to go now. Same commands as before gives:
```
  MPI:
    Host Name                          : fermat
    Total ranks                        : 4
  Problem:
  libCEED:
  PETSc:
    Box Faces                          : 40,60,2
  Mesh:
    Global DoFs                        : 32226
    Global 5-DoF nodes                 : 6445
  Partition:                             (min,max,median,max/median)
    Owned free 5-DoF nodes             : 1536, 1766, 1538, 1.148505
    Local 5-DoF nodes                  : 1920, 2064, 1938, 1.065015
    Ghost Boundary 5-DoF nodes         : 0, 163, 59, 2.739130
    Ghost Boundary Ranks               : 0, 3, 1, 3.000000
    Owned Boundary 5-DoF nodes         : 0, 185, 44, 4.203620
    Owned Boundary Ranks               : 0, 3, 1, 3.000000
```

jedbrown (approved):
LGTM

PR #1248, opened 2023-07-03
Update LIBXSMM backend

sebastiangrimberg:
Provide a few updates to LIBXSMM to work with recent `main_stable` and `main` (broken on since May 25, 2023):

- LIBXSMM commit https://github.com/libxsmm/libxsmm/commit/ccc537389e63ddd75b6f9f0013146c642f3e9348 on `main_stable` on 05/25 removed `libxsmm_dmmdispatch` and `libxsmm_smmdispatch`, which breaks the libCEED interface. These are superceded with `libxsmm_dispatch_gemm_v2`, available on `main_stable` since Jan. 2022 (https://github.com/libxsmm/libxsmm/commit/c7229010f932ec00b9f66f9f3f662c9f75103609).
- There is no need to call `libxsmm_release_kernel` on the JIT'd kernels from the hash tables in `CeedTensorContractDestroy_Xsmm` (see [this comment](https://github.com/libxsmm/libxsmm/issues/783#issuecomment-1596655284)).
- Allow the user to specify and override `BLAS_LIB` in the Makefile for a known LIBXSMM configuration (for example, with `BLAS=0` we can set `BLAS_LIB=`).

TODO:
- [ ] Backward compatibility (?)


jeremylt:
Note - I've been waiting for libXSMM to make a new release. Do you know if one is coming?

jeremylt:
This mix of commits from other branches makes this branch hard for me to follow.

sebastiangrimberg:
> This mix of commits from other branches makes this branch hard for me to follow.

Indeed, hence the draft state. I will clean this up and mark as ready shortly.

sebastiangrimberg:
> Note - I've been waiting for libXSMM to make a new release. Do you know if one is coming?

I do not know a timeline unfortunately. There's some information on compatibility here: https://github.com/libxsmm/libxsmm/wiki/Compatibility#libxsmm-2x, which recommends moving to the `v2` interface for now as I understand it.

jeremylt:
> Indeed, hence the draft state. I will clean this up and mark as ready shortly.

Cool, you did some linked PRs earlier so I wasn't sure if this was a new strategy you were adopting. I've tried to move away from linked PRs as much as I can since they tend to get confusing for others.

> recommends moving to the v2 interface for now as I understand it.

Hmm. I read it as 'new work should use the new interface'. It's not a huge deal if they super promise this is close to v2, but it's definitely easier to communicate libCEED dependencies we know will work if we stick to releases. We've had issues in the past tracking main branches so I'm a bit hesitant to re-adopt a strategy we intentionally moved away from.

jeremylt:
Re kernel cleanup: Does libXSMM now automatically clean those up on its own? libCEED tries to be very strict about cleaning up absolutely everything we allocate.

sebastiangrimberg:
> Re kernel cleanup: Does libXSMM now automatically clean those up on its own? libCEED tries to be very strict about cleaning up absolutely everything we allocate.

This is my impression after asking in LIBXSMM here: https://github.com/libxsmm/libxsmm/issues/783#issuecomment-1595337903. I noticed this running on libCEED `main` when using the commit of LIBXSMM tested in the CI: https://github.com/CEED/libCEED/blob/main/.gitlab-ci.yml#L22. When running with `LIBXSMM_VERBOSE`, I saw errors like `LIBXSMM ERROR: failed to release kernel!`. I asked in LIBXSMM if we needed to free the kernels explicitly and was told no, unless you build LIBXSMM in a specific way.

sebastiangrimberg:
I agree that it is preferable to avoid the latter behavior, and there might be a way to do so though I do not know. But at least the current behavior of calling `libxsmm_release_kernel` resulting in error messages seems like it isn't the right way to use the API and should be fixed.

sebastiangrimberg:
OK, here are a few results, running `LIBXSMM_VERBOSE=3 ./bps -ceed /cpu/self/xsmm/serial,/cpu/self/xsmm/blocked -problem bp1,bp3 -degree 3 -q_extra 1 -simplex -local_nodes 100000 -ksp_max_it 200` (M1 Max). I ran with `-simplex` to put a bit more stress on the basis application, but would expect the changes to be the same for tensor-product elements:

- `main` (with LIBXSMM from 44433be9426eddaed88415646c15b3bcc61afc85):
    - bp1, serial: 4.7 MDoFs/s
    - bp1, blocked: 25.9 MDoFs/s
    - bp3, serial: 1.7 MDoFs/s
    - bp3, blocked: 9.0 MDoFs/s
    - JIT (small/medium/large): 7/8/4

- 1300445 (using libCEED hash tables to cache kernels):
    - bp1, serial: 4.7 MDoFs/s
    - bp1, blocked: 25.1 MDoFs/s
    - bp3, serial: 1.7 MDoFs/s
    - bp3, blocked: 9.0 MDoFs/s
    - JIT: 15/8/0

- 1962c6df (using LIBXSMM's internal kernel lookup):
    - bp1, serial: 4.8 MDoFs/s
    - bp1, blocked: 26.3 MDoFs/s
    - bp3, serial: 1.7 MDoFs/s
    - bp3, blocked: 9.2 MDoFs/s
    - JIT: 4/3/0

Using a more recent LIBXSMM (`main_stable` from July 1):

- 1300445 (using libCEED hash tables to cache kernels): Does not work for `xsmm/serial` (`libxsmm_[sd]gemm` segfault?)

- 1962c6df (using LIBXSMM's internal kernel lookup):
    - bp1, serial: 4.7 MDoFs/s
    - bp1, blocked: 26.0 MDoFs/s
    - bp3, serial: 1.7 MDoFs/s
    - bp3, blocked: 9.0 MDoFs/s
    - JIT: 2/2/0

The changes all follow from LIBXSMM's simple example: https://github.com/libxsmm/libxsmm/blob/main_stable/samples/hello/hello.c.

hfp:
> It is not clear to me if that pattern of kernel management results in libXSMM cleaning up afterwards or using the 'it all goes away when the application ends' memory management. We try to avoid the second.

LIBXSMM performs explicit cleanup (not left-up to OS or such). This cleanup might happen "late", i.e, when your application terminates.

hfp:
> Re kernel cleanup: Does libXSMM now automatically clean those up on its own?

This has always been the case. We never had a version not attempting to cleanup the code registry. We always had the same lifetime policy to match normal function pointers, and the function `libxsmm_release_kernel` only exists "for symmetry" and completeness.

jeremylt:
Looks like letting libXSMM cache the kernels itself makes good sense and has basically the same performance.

> LIBXSMM performs explicit cleanup (not left-up to OS or such). This cleanup might happen "late", i.e, when your application terminates.

Huzzah, one less thing to worry about.

Seems like we should just let libXSMM manage the caching and cleanup itself.

sebastiangrimberg:
It seems, then, that the only question with this PR would remain to check compatibility? `libxsmm_dispatch_gemm_v2` has existed since `master-1.17-2069`, which is from Jan. 2022. If a user were to install with Spack by default they would get v1.17 (unless using `@develop`) which predates this by about a month. So, for the user's sake it does seem like it would be irresponsible to merge this change until a new version of LIBXSMM is tagged...

jeremylt:
Yeah, I say we make this change as soon as libXSMM has tagged the new release

jedbrown:
Yep, sounds good. Thanks for doing this study @sebastiangrimberg. Do you think the very slight regression in the most recent version is real or within the noise?

sebastiangrimberg:
I ran these tests on my laptop, averaging results over 10 applications of `KSPSolve`. I would have to think these small differences are just noise, but unfortunately I don't have a good enough familiarity with LIBXSMM to comment on any possible performance regressions. I can try running the benchmarks on a server instead to try to get more consistent performance and see if there is any noticable difference in performance.

I would have to think that if we are using the library in the supported/recommended way to compile/query/apply the GEMM kernels, I don't see how we could achieve better performance or how we could be doing something which is hurting our performance. But again, I'd love to defer to the expertise of @hfp  here to ensure there is not something we could be doing better for the types of GEMM/matrix sizes we are operating on.

include/ceed/hash.h
```diff
@@ -1,103 +0,0 @@
-// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and other CEED contributors.
```
jeremylt, line 1: Note - do we want to delete these? No harm in keeping them and I'm not sure if anyone using libCEED is using this hashing or not

jedbrown: I think this is okay to consider as private, but I think there's a decent chance we'll want to use this code again. BTW, the copyright header is incorrect since the copyright statement from PETSc was supposed to be preserved (`UChicago Argonne, LLC and the PETSc Development Team`). It is also BSD-2-Clause so that part is correct.

jeremylt:
Per the discussion in #1164, it would be good to merge this branch now in anticipation of a "soon" LIBXSMM release.

If you let me know when this is PR just about ready, I can update the version of LIBXSMM that CI uses and we can run CI to verify before merging.

Makefile
```diff
@@ -698,8 +697,6 @@ install : $(libceed) $(OBJDIR)/ceed.pc
 	$(INSTALL_DATA) include/ceed/backend.h "$(DESTDIR)$(includedir)/ceed/"
 	$(INSTALL_DATA) include/ceed/cuda.h "$(DESTDIR)$(includedir)/ceed/"
 	$(INSTALL_DATA) include/ceed/hip.h "$(DESTDIR)$(includedir)/ceed/"
-	$(INSTALL_DATA) include/ceed/hash.h "$(DESTDIR)$(includedir)/ceed/"
-	$(INSTALL_DATA) include/ceed/khash.h "$(DESTDIR)$(includedir)/ceed/"
```
jeremylt, line 702: We'd drop these two lines to keep these headers private

backends/xsmm/ceed-xsmm-tensor-f64.c
```diff
@@ -7,148 +7,65 @@
...
 
-  // Run kernel or fallback to default implementation
-  if (C != 1) {
-    for (CeedInt a = 0; a < A; a++) LIBXSMM_MMFUNCTION_KERNEL(&u[a * B * C], &t[0], &v[a * J * C]);
+  if (C == 1) {
+    // Build or query the required kernel
+    const int                  flags_t    = LIBXSMM_GEMM_FLAGS(!t_mode ? 'T' : 'N', 'N');
+    const int                  flags_ab   = (!add) ? LIBXSMM_GEMM_FLAG_BETA_0 : 0;
+    const int                  flags      = (flags_t | flags_ab);
+    const libxsmm_gemm_shape   gemm_shape = libxsmm_create_gemm_shape(J, A, B, !t_mode ? B : J, B, J, LIBXSMM_DATATYPE_F64, LIBXSMM_DATATYPE_F64,
```
jeremylt, line 27: Can we combine the two `ceed-xsmm-tensor` files now since the only difference is the `LIBXSMM_DATATYPE_*`, which we can select based upon the value of `CEED_SCALAR_TYPE`?

sebastiangrimberg: Nice, see [fe99f24](https://github.com/CEED/libCEED/pull/1248/commits/fe99f2494c1cc00aebca1d5c9b9672a3607e535b).

include/ceed/hash.h
sebastiangrimberg: I would say remove for now and easy enough to restore them back into the repo (as private) once they are needed again? I don't see the benefit in keeping around the files if they aren't used anywhere for now.

jeremylt: cool!

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -0,0 +1,79 @@
...
+    gemm_param.a.primary = (CeedScalar *)&t[0];
+    gemm_param.b.primary = (CeedScalar *)&u[0];
+    gemm_param.c.primary = (CeedScalar *)&v[0];
+    kernel(&gemm_param);
+  } else {
+    // Build or query the required kernel
+    const int                  flags_t    = LIBXSMM_GEMM_FLAGS('N', t_mode ? 'T' : 'N');
+    const int                  flags_ab   = (!add) ? LIBXSMM_GEMM_FLAG_BETA_0 : 0;
+    const int                  flags      = (flags_t | flags_ab);
+    const libxsmm_gemm_shape   gemm_shape = (CEED_SCALAR_TYPE != CEED_SCALAR_FP32)
```
jeremylt, line 46: minor - I generally find "if this then that else other" easier to read than "if not this then other else that"
```suggestion
    const libxsmm_gemm_shape   gemm_shape = (CEED_SCALAR_TYPE == CEED_SCALAR_FP64)
```

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -0,0 +1,79 @@
...
+                                        CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
+  Ceed ceed;
+  CeedCallBackend(CeedTensorContractGetCeed(contract, &ceed));
+
+  if (C == 1) {
+    // Build or query the required kernel
+    const int                  flags_t    = LIBXSMM_GEMM_FLAGS(!t_mode ? 'T' : 'N', 'N');
+    const int                  flags_ab   = (!add) ? LIBXSMM_GEMM_FLAG_BETA_0 : 0;
+    const int                  flags      = (flags_t | flags_ab);
+    const libxsmm_gemm_shape   gemm_shape = (CEED_SCALAR_TYPE != CEED_SCALAR_FP32)
```
jeremylt, line 27:
```suggestion
    const libxsmm_gemm_shape   gemm_shape = (CEED_SCALAR_TYPE == CEED_SCALAR_FP64)
```

backends/xsmm/ceed-xsmm-tensor.c
```diff
@@ -0,0 +1,79 @@
...
+//------------------------------------------------------------------------------
+static int CeedTensorContractApply_Xsmm(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
+                                        CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
+  Ceed ceed;
+  CeedCallBackend(CeedTensorContractGetCeed(contract, &ceed));
+
+  if (C == 1) {
+    // Build or query the required kernel
+    const int                  flags_t    = LIBXSMM_GEMM_FLAGS(!t_mode ? 'T' : 'N', 'N');
+    const int                  flags_ab   = (!add) ? LIBXSMM_GEMM_FLAG_BETA_0 : 0;
```
jeremylt, line 25: small question - is there a named enum/macro (not sure which `LIBXSMM_GEMM_FLAG_BETA` is) for 0 here?

sebastiangrimberg: This is from https://github.com/libxsmm/libxsmm/blob/main_stable/samples/hello/hello.c. I think the alternative would be to replace `0` by `LIBXSMM_BASIC_GEMM_FLAG_NONE`.

jeremylt: My preference is to use words if we have them.

backends/xsmm/ceed-xsmm-tensor-f64.c
sebastiangrimberg: I have a similar update for the AVX backend. Can I add this here or should I make a new PR?

backends/xsmm/ceed-xsmm-tensor.c
sebastiangrimberg: Done: 684fdff.

backends/xsmm/ceed-xsmm-tensor.c
sebastiangrimberg: Done: 684fdff.

backends/xsmm/ceed-xsmm-tensor.c
sebastiangrimberg: Done: 684fdff.

jeremylt:
I'll update CI right after my lunch and then I think we can merge

jeremylt:
Ok, if you apply this diff, then we can merge this branch:

```diff
$ git diff
diff --git a/.gitlab-ci.yml b/.gitlab-ci.yml
index 11b5881c..dcafb306 100644
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -18,8 +18,8 @@ noether-cpu:
     - echo "-------------- HIPCC ---------------" && $HIPCC --version
     - echo "-------------- GCOV ----------------" && gcov --version
 # Libraries for backends
-# -- LIBXSMM 44433be9426eddaed88415646c15b3bcc61afc85
-    - cd .. && export XSMM_HASH=44433be9426eddaed88415646c15b3bcc61afc85 && { [[ -d libxsmm-$XSMM_HASH ]] || { curl -L https://github.com/libxsmm/libxsmm/archive/$XSMM_HASH.tar.gz -o xsmm.tar.gz && tar zvxf xsmm.tar.gz && rm xsmm.tar.gz && make -C libxsmm-$XSMM_HASH -j$(nproc); }; } && export XSMM_DIR=$PWD/libxsmm-$XSMM_HASH && cd libCEED
+# -- LIBXSMM 2c145a109b5a8ad4e15f60ea42a86b9056bdc8b8
+    - cd .. && export XSMM_HASH=2c145a109b5a8ad4e15f60ea42a86b9056bdc8b8 && { [[ -d libxsmm-$XSMM_HASH ]] || { curl -L https://github.com/libxsmm/libxsmm/archive/$XSMM_HASH.tar.gz -o xsmm.tar.gz && tar zvxf xsmm.tar.gz && rm xsmm.tar.gz && make -C libxsmm-$XSMM_HASH -j$(nproc); }; } && export XSMM_DIR=$PWD/libxsmm-$XSMM_HASH && cd libCEED
     - echo "-------------- LIBXSMM -------------" && basename $XSMM_DIR
 # -- OCCA v1.1.0
     - cd .. && export OCCA_VERSION=occa-1.4.0 && { [[ -d $OCCA_VERSION ]] || { git clone --depth 1 --branch v1.4.0 https://github.com/libocca/occa.git $OCCA_VERSION && cd $OCCA_VERSION && export ENABLE_OPENCL="OFF" ENABLE_DPCPP="OFF" ENABLE_HIP="OFF" ENABLE_CUDA="OFF" && ./configure-cmake.sh && cmake --build build --parallel $NPROC_CPU && cmake --install build && cd ..; }; } && export OCCA_DIR=$PWD/$OCCA_VERSION/install && cd libCEED
@@ -162,8 +162,8 @@ noether-float:
 # -- MAGMA from dev branch
     - echo "-------------- MAGMA ---------------"
     - export MAGMA_DIR=/projects/hipMAGMA && git -C $MAGMA_DIR -c safe.directory=$MAGMA_DIR describe
-# -- LIBXSMM 44433be9426eddaed88415646c15b3bcc61afc85
-    - cd .. && export XSMM_HASH=44433be9426eddaed88415646c15b3bcc61afc85 && { [[ -d libxsmm-$XSMM_HASH ]] || { curl -L https://github.com/libxsmm/libxsmm/archive/$XSMM_HASH.tar.gz -o xsmm.tar.gz && tar zvxf xsmm.tar.gz && rm xsmm.tar.gz && make -C libxsmm-$XSMM_HASH -j$(nproc); }; } && export XSMM_DIR=$PWD/libxsmm-$XSMM_HASH && cd libCEED
+# -- LIBXSMM 2c145a109b5a8ad4e15f60ea42a86b9056bdc8b8
+    - cd .. && export XSMM_HASH=2c145a109b5a8ad4e15f60ea42a86b9056bdc8b8 && { [[ -d libxsmm-$XSMM_HASH ]] || { curl -L https://github.com/libxsmm/libxsmm/archive/$XSMM_HASH.tar.gz -o xsmm.tar.gz && tar zvxf xsmm.tar.gz && rm xsmm.tar.gz && make -C libxsmm-$XSMM_HASH -j$(nproc); }; } && export XSMM_DIR=$PWD/libxsmm-$XSMM_HASH && cd libCEED
     - echo "-------------- LIBXSMM -------------" && basename $XSMM_DIR
   script:
     - rm -f .SUCCESS
```

jeremylt (approved):

Issue #1250, opened 2023-07-07
PesudoInverse Utility


jeremylt:
We use the QR factorization to create pesudoinverses in several places in our code. We should create a single pesudoinverse utility to reduce the duplicated code. This utility should be part of `ceed/backend.h` as this is not intended as a performant option for users.

Proposed function signature:
```C
CEED_INTERN int CeedMatrixPesudoinverse(Ceed ceed, const CeedScalar *mat, CeedInt m, CeedInt n, CeedScalar *mat_pinv);
```

rezgarshakeri:
@jeremylt when I look at these two loops (one for chebyshev and one for collocated grad)
```c
     // Inverse of coefficient matrix
    CeedCall(CeedCalloc(Q_1d * Q_1d, &chebyshev_coeffs_1d));
    CeedCall(CeedCalloc(Q_1d * Q_1d, &I));
    CeedCall(CeedCalloc(Q_1d, &tau));
    // -- QR Factorization, C = Q R
    CeedCall(CeedQRFactorization(basis->ceed, C, tau, Q_1d, Q_1d));
    // -- chebyshev_coeffs_1d = R_inv Q^T
    for (CeedInt i = 0; i < Q_1d; i++) I[i * Q_1d + i] = 1.0;
    // ---- Apply R_inv, chebyshev_coeffs_1d = I R_inv
    for (CeedInt i = 0; i < Q_1d; i++) {  // Row i
      chebyshev_coeffs_1d[Q_1d * i] = I[Q_1d * i] / C[0];
      for (CeedInt j = 1; j < Q_1d; j++) {  // Column j
        chebyshev_coeffs_1d[j + Q_1d * i] = I[j + Q_1d * i];
        for (CeedInt k = 0; k < j; k++) chebyshev_coeffs_1d[j + Q_1d * i] -= C[j + Q_1d * k] * chebyshev_coeffs_1d[k + Q_1d * i];
        chebyshev_coeffs_1d[j + Q_1d * i] /= C[j + Q_1d * j];
      }
    }
    // ---- Apply Q^T, chebyshev_coeffs_1d = R_inv Q^T
    CeedCall(CeedHouseholderApplyQ(chebyshev_coeffs_1d, C, tau, CEED_NOTRANSPOSE, Q_1d, Q_1d, Q_1d, 1, Q_1d));
```

```c
   // QR Factorization, interp_1d = Q R
  CeedCall(CeedBasisGetCeed(basis, &ceed));
[...]

PR #1251, opened 2023-07-10
CeedMatrixPseudoinverse utility function

rezgarshakeri:
See https://github.com/CEED/libCEED/issues/1250

### DIFF
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -271,6 +271,10 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
     // output_project = interp_to^+ * interp_from
     memcpy(interp_from, input_from[m], Q * P_from * q_comp * sizeof(input_from[m][0]));
     CeedCall(CeedMatrixMatrixMultiply(ceed, interp_to_inv, input_from[m], output_project[m], P_to, P_from, Q * q_comp));
+    // Round zero to machine precision
+    for (CeedInt i = 0; i < P_to * P_from; i++) {
+      if (fabs(output_project[m][i]) < 10 * CEED_EPSILON) output_project[m][i] = 0.0;
+    }
   }
 
   // Cleanup
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -302,12 +302,11 @@ int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *collo_grad_1d) {
   CeedInt     P_1d, Q_1d;
   CeedScalar *interp_1d_pinv;
 
-  // QR Factorization, interp_1d = Q R
   CeedCall(CeedBasisGetCeed(basis, &ceed));
   CeedCall(CeedBasisGetNumNodes1D(basis, &P_1d));
   CeedCall(CeedBasisGetNumQuadraturePoints1D(basis, &Q_1d));
 
-  // QR Factorization, interp_1d = Q R
+  // Compute interp_1d^+, pseudoinverse of interp_1d
   CeedCall(CeedCalloc(P_1d * Q_1d, &interp_1d_pinv));
 
   CeedCall(CeedMatrixPseudoinverse(ceed, basis->interp_1d, Q_1d, P_1d, interp_1d_pinv));
@@ -1534,7 +1533,7 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
   }
   if (!basis->basis_chebyshev) {
     // Build matrix mapping from quadrature point values to Chebyshev coefficients
-    CeedScalar       *C, *chebyshev_coeffs_1d;
+    CeedScalar       *C, *chebyshev_coeffs_1d_inv;
     const CeedScalar *q_ref_1d;
 
     // Build coefficient matrix
@@ -1544,10 +1543,9 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
     CeedCall(CeedBasisGetQRef(basis, &q_ref_1d));
     for (CeedInt i = 0; i < Q_1d; i++) CeedCall(CeedChebyshevPolynomialsAtPoint(q_ref_1d[i], Q_1d, &C[i * Q_1d]));
 
-    // Inverse of coefficient matrix
-    // -- QR Factorization, C = Q R
-    CeedCall(CeedCalloc(Q_1d * Q_1d, &chebyshev_coeffs_1d));
-    CeedCall(CeedMatrixPseudoinverse(basis->ceed, C, Q_1d, Q_1d, chebyshev_coeffs_1d));
+    // Compute C^+, pseudoinverse of coefficient matrix
+    CeedCall(CeedCalloc(Q_1d * Q_1d, &chebyshev_coeffs_1d_inv));
+    CeedCall(CeedMatrixPseudoinverse(basis->ceed, C, Q_1d, Q_1d, chebyshev_coeffs_1d_inv));
 
     // Build basis mapping from nodes to Chebyshev coefficients
     CeedScalar       *chebyshev_interp_1d, *chebyshev_grad_1d, *chebyshev_q_weight_1d;
@@ -1557,7 +1555,7 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
     CeedCall(CeedCalloc(P_1d * Q_1d, &chebyshev_grad_1d));
     CeedCall(CeedCalloc(Q_1d, &chebyshev_q_weight_1d));
     CeedCall(CeedBasisGetInterp1D(basis, &interp_1d));
-    CeedCall(CeedMatrixMatrixMultiply(basis->ceed, chebyshev_coeffs_1d, interp_1d, chebyshev_interp_1d, Q_1d, P_1d, Q_1d));
+    CeedCall(CeedMatrixMatrixMultiply(basis->ceed, chebyshev_coeffs_1d_inv, interp_1d, chebyshev_interp_1d, Q_1d, P_1d, Q_1d));
 
     CeedCall(CeedVectorCreate(basis->ceed, num_comp * CeedIntPow(Q_1d, dim), &basis->vec_chebyshev));
     CeedCall(CeedBasisCreateTensorH1(basis->ceed, dim, num_comp, P_1d, Q_1d, chebyshev_interp_1d, chebyshev_grad_1d, q_ref_1d, chebyshev_q_weight_1d,
@@ -1565,7 +1563,7 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
 
     // Cleanup
     CeedCall(CeedFree(&C));
-    CeedCall(CeedFree(&chebyshev_coeffs_1d));
+    CeedCall(CeedFree(&chebyshev_coeffs_1d_inv));
     CeedCall(CeedFree(&chebyshev_interp_1d));
     CeedCall(CeedFree(&chebyshev_grad_1d));
     CeedCall(CeedFree(&chebyshev_q_weight_1d));
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -1550,7 +1550,7 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
   }
   if (!basis->basis_chebyshev) {
     // Build matrix mapping from quadrature point values to Chebyshev coefficients
-    CeedScalar       *tau, *C, *I, *chebyshev_coeffs_1d;
+    CeedScalar       *C, *chebyshev_coeffs_1d;
     const CeedScalar *q_ref_1d;
 
     // Build coefficient matrix
@@ -1561,24 +1561,9 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
     for (CeedInt i = 0; i < Q_1d; i++) CeedCall(CeedChebyshevPolynomialsAtPoint(q_ref_1d[i], Q_1d, &C[i * Q_1d]));
 
     // Inverse of coefficient matrix
-    CeedCall(CeedCalloc(Q_1d * Q_1d, &chebyshev_coeffs_1d));
-    CeedCall(CeedCalloc(Q_1d * Q_1d, &I));
-    CeedCall(CeedCalloc(Q_1d, &tau));
     // -- QR Factorization, C = Q R
-    CeedCall(CeedQRFactorization(basis->ceed, C, tau, Q_1d, Q_1d));
-    // -- chebyshev_coeffs_1d = R_inv Q^T
-    for (CeedInt i = 0; i < Q_1d; i++) I[i * Q_1d + i] = 1.0;
-    // ---- Apply R_inv, chebyshev_coeffs_1d = I R_inv
-    for (CeedInt i = 0; i < Q_1d; i++) {  // Row i
-      chebyshev_coeffs_1d[Q_1d * i] = I[Q_1d * i] / C[0];
-      for (CeedInt j = 1; j < Q_1d; j++) {  // Column j
-        chebyshev_coeffs_1d[j + Q_1d * i] = I[j + Q_1d * i];
-        for (CeedInt k = 0; k < j; k++) chebyshev_coeffs_1d[j + Q_1d * i] -= C[j + Q_1d * k] * chebyshev_coeffs_1d[k + Q_1d * i];
-        chebyshev_coeffs_1d[j + Q_1d * i] /= C[j + Q_1d * j];
-      }
-    }
-    // ---- Apply Q^T, chebyshev_coeffs_1d = R_inv Q^T
-    CeedCall(CeedHouseholderApplyQ(chebyshev_coeffs_1d, C, tau, CEED_NOTRANSPOSE, Q_1d, Q_1d, Q_1d, 1, Q_1d));
+    CeedCall(CeedCalloc(Q_1d * Q_1d, &chebyshev_coeffs_1d));
+    CeedCall(CeedMatrixPseudoinverse(basis->ceed, C, Q_1d, Q_1d, chebyshev_coeffs_1d));
 
     // Build basis mapping from nodes to Chebyshev coefficients
     CeedScalar       *chebyshev_interp_1d, *chebyshev_grad_1d, *chebyshev_q_weight_1d;
@@ -1597,8 +1582,6 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
     // Cleanup
     CeedCall(CeedFree(&C));
     CeedCall(CeedFree(&chebyshev_coeffs_1d));
-    CeedCall(CeedFree(&I));
-    CeedCall(CeedFree(&tau));
     CeedCall(CeedFree(&chebyshev_interp_1d));
     CeedCall(CeedFree(&chebyshev_grad_1d));
     CeedCall(CeedFree(&chebyshev_q_weight_1d));
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -227,7 +227,7 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
 
   // Get source matrices
   CeedInt           dim, q_comp = 1;
-  CeedScalar       *interp_to, *interp_from, *tau;
+  CeedScalar       *interp_to_inv, *interp_from;
   const CeedScalar *interp_to_source = NULL, *interp_from_source = NULL, *grad_from_source = NULL;
 
   CeedCall(CeedBasisGetDimension(basis_to, &dim));
@@ -240,9 +240,7 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
     CeedCall(CeedBasisGetInterp(basis_from, &interp_from_source));
   }
   CeedCall(CeedMalloc(Q * P_from * q_comp, &interp_from));
-  CeedCall(CeedMalloc(Q * P_to * q_comp, &interp_to));
   CeedCall(CeedCalloc(P_to * P_from, interp_project));
-  CeedCall(CeedMalloc(Q * q_comp, &tau));
 
   // `grad_project = interp_to^+ * grad_from` is computed for the H^1 space case so the
   // projection basis will have a gradient operation (allocated even if not H^1 for the
@@ -256,10 +254,9 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
   }
   CeedCall(CeedCalloc(P_to * P_from * (is_tensor_to ? 1 : dim), grad_project));
 
-  // QR Factorization, interp_to = Q R
-  memcpy(interp_to, interp_to_source, Q * P_to * q_comp * sizeof(interp_to_source[0]));
-  CeedCall(CeedQRFactorization(ceed, interp_to, tau, Q * q_comp, P_to));
-
+  // Compute interp_to^+, pseudoinverse of interp_to
+  CeedCall(CeedCalloc(Q * q_comp * P_to, &interp_to_inv));
+  CeedCall(CeedMatrixPseudoinverse(ceed, (CeedScalar *)interp_to_source, Q * q_comp, P_to, interp_to_inv));
   // Build matrices
   CeedInt     num_matrices = 1 + (fe_space_to == CEED_FE_SPACE_H1) * (is_tensor_to ? 1 : dim);
   CeedScalar *input_from[num_matrices], *output_project[num_matrices];
@@ -271,26 +268,13 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
     output_project[m] = &((*grad_project)[(m - 1) * P_to * P_from]);
   }
   for (CeedInt m = 0; m < num_matrices; m++) {
-    // Apply Q^T, interp_from = Q^T interp_from
+    // output_project = interp_to^+ * interp_from
     memcpy(interp_from, input_from[m], Q * P_from * q_comp * sizeof(input_from[m][0]));
-    CeedCall(CeedHouseholderApplyQ(interp_from, interp_to, tau, CEED_TRANSPOSE, Q * q_comp, P_from, P_to, P_from, 1));
-
-    // Apply Rinv, output_project = Rinv interp_from
-    for (CeedInt j = 0; j < P_from; j++) {  // Column j
-      output_project[m][j + P_from * (P_to - 1)] = interp_from[j + P_from * (P_to - 1)] / interp_to[P_to * P_to - 1];
-      for (CeedInt i = P_to - 2; i >= 0; i--) {  // Row i
-        output_project[m][j + P_from * i] = interp_from[j + P_from * i];
-        for (CeedInt k = i + 1; k < P_to; k++) {
-          output_project[m][j + P_from * i] -= interp_to[k + P_to * i] * output_project[m][j + P_from * k];
-        }
-        output_project[m][j + P_from * i] /= interp_to[i + P_to * i];
-      }
-    }
+    CeedCall(CeedMatrixMatrixMultiply(ceed, interp_to_inv, input_from[m], output_project[m], P_to, P_from, Q * q_comp));
   }
 
   // Cleanup
-  CeedCall(CeedFree(&tau));
-  CeedCall(CeedFree(&interp_to));
+  CeedCall(CeedFree(&interp_to_inv));
   CeedCall(CeedFree(&interp_from));
   return CEED_ERROR_SUCCESS;
 }
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -315,37 +315,21 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
 **/
 int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *collo_grad_1d) {
   Ceed        ceed;
-  CeedInt     P_1d = (basis)->P_1d, Q_1d = (basis)->Q_1d;
-  CeedScalar *interp_1d, *grad_1d, *tau;
-
-  CeedCall(CeedMalloc(Q_1d * P_1d, &interp_1d));
-  CeedCall(CeedMalloc(Q_1d * P_1d, &grad_1d));
-  CeedCall(CeedMalloc(Q_1d, &tau));
-  memcpy(interp_1d, (basis)->interp_1d, Q_1d * P_1d * sizeof(basis)->interp_1d[0]);
-  memcpy(grad_1d, (basis)->grad_1d, Q_1d * P_1d * sizeof(basis)->interp_1d[0]);
+  CeedInt     P_1d, Q_1d;
+  CeedScalar *interp_1d_pinv;
 
   // QR Factorization, interp_1d = Q R
   CeedCall(CeedBasisGetCeed(basis, &ceed));
-  CeedCall(CeedQRFactorization(ceed, interp_1d, tau, Q_1d, P_1d));
-  // Note: This function is for backend use, so all errors are terminal and we do not need to clean up memory on failure.
+  CeedCall(CeedBasisGetNumNodes1D(basis, &P_1d));
+  CeedCall(CeedBasisGetNumQuadraturePoints1D(basis, &Q_1d));
 
-  // Apply R_inv, collo_grad_1d = grad_1d R_inv
-  for (CeedInt i = 0; i < Q_1d; i++) {  // Row i
-    collo_grad_1d[Q_1d * i] = grad_1d[P_1d * i] / interp_1d[0];
-    for (CeedInt j = 1; j < P_1d; j++) {  // Column j
-      collo_grad_1d[j + Q_1d * i] = grad_1d[j + P_1d * i];
-      for (CeedInt k = 0; k < j; k++) collo_grad_1d[j + Q_1d * i] -= interp_1d[j + P_1d * k] * collo_grad_1d[k + Q_1d * i];
-      collo_grad_1d[j + Q_1d * i] /= interp_1d[j + P_1d * j];
-    }
-    for (CeedInt j = P_1d; j < Q_1d; j++) collo_grad_1d[j + Q_1d * i] = 0;
-  }
+  // QR Factorization, interp_1d = Q R
+  CeedCall(CeedCalloc(P_1d * Q_1d, &interp_1d_pinv));
 
-  // Apply Q^T, collo_grad_1d = collo_grad_1d Q^T
-  CeedCall(CeedHouseholderApplyQ(collo_grad_1d, interp_1d, tau, CEED_NOTRANSPOSE, Q_1d, Q_1d, P_1d, 1, Q_1d));
+  CeedCall(CeedMatrixPseudoinverse(ceed, basis->interp_1d, Q_1d, P_1d, interp_1d_pinv));
+  CeedCall(CeedMatrixMatrixMultiply(ceed, basis->grad_1d, (const CeedScalar *)interp_1d_pinv, collo_grad_1d, Q_1d, Q_1d, P_1d));
 
-  CeedCall(CeedFree(&interp_1d));
-  CeedCall(CeedFree(&grad_1d));
-  CeedCall(CeedFree(&tau));
+  CeedCall(CeedFree(&interp_1d_pinv));
   return CEED_ERROR_SUCCESS;
 }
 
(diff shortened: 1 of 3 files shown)


jeremylt:
This is looking good. We now need to use this utility in a few places to cut repeated code. I think the two places to update are `CeedBasisCreateProjectionMatrices` and `CeedBasisGetCollocatedGrad`.

interface/ceed-basis.c
```diff
@@ -636,6 +636,44 @@ int CeedHouseholderApplyQ(CeedScalar *mat_A, const CeedScalar *mat_Q, const Ceed
...
+      mat_pinv[j + m * i] = I[j + m * i];
+      for (CeedInt k = 0; k < j; k++) mat_pinv[j + m * i] -= mat[j + m * k] * mat_pinv[k + m * i];
+      mat_pinv[j + m * i] /= mat[j + n * j];
+    }
+  }
+  // ---- Apply Q^T, mat_pinv = R_inv Q^T
+  CeedCall(CeedHouseholderApplyQ(mat_pinv, mat, tau, CEED_NOTRANSPOSE, m, n, n, 1, m));
+
+  return CEED_ERROR_SUCCESS;
+}
```
rezgarshakeri, line 675: Am I using the indices in a correct way here? I tried to use this function in `CeedBasisGetCollocatedGrad` as
```c
CeedCall(CeedMatrixPseudoinverse(ceed, interp_1d, Q_1d, P_1d, interp_1d_pinv));
  CeedCall(CeedMatrixMatrixMultiply(ceed, (const CeedScalar *)interp_1d_pinv, (const CeedScalar *)grad_1d, collo_grad_1d, Q_1d, P_1d, P_1d));
```
but breaks some tests.

jeremylt: I think you want to swap the order of the MatMatMult - grad_1d * interp_1d_pinv = collograd_1d

rezgarshakeri: Yeah, I tried that, didn't work. In last commit I changed the indices and I tested t302-basis.c , still doesn't work but I don't get memory error at least.

jeremylt: I remembered why I don't like this code, lol. I did some fiddling and this might work?

```diff
diff --git a/interface/ceed-basis.c b/interface/ceed-basis.c
index 60c7a6e9..8e5d304a 100644
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -259,26 +259,21 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
 **/
 int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *collo_grad_1d) {
   Ceed        ceed;
-  CeedInt     P_1d = (basis)->P_1d, Q_1d = (basis)->Q_1d;
-  CeedScalar *interp_1d, *interp_1d_pinv, *grad_1d, *tau;
+  CeedInt     P_1d, Q_1d;
+  CeedScalar *interp_1d_pinv;
 
-  CeedCall(CeedMalloc(Q_1d * P_1d, &interp_1d));
-  CeedCall(CeedMalloc(P_1d * Q_1d, &interp_1d_pinv));
-  CeedCall(CeedMalloc(Q_1d * P_1d, &grad_1d));
-  CeedCall(CeedMalloc(Q_1d, &tau));
-  memcpy(interp_1d, (basis)->interp_1d, Q_1d * P_1d * sizeof(basis)->interp_1d[0]);
-  memcpy(grad_1d, (basis)->grad_1d, Q_1d * P_1d * sizeof(basis)->interp_1d[0]);
-
-  // QR Factorization, interp_1d = Q R
   CeedCall(CeedBasisGetCeed(basis, &ceed));
+  CeedCall(CeedBasisGetNumNodes1D(basis, &P_1d));
+  CeedCall(CeedBasisGetNumQuadraturePoints1D(basis, &Q_1d));
 
-  CeedCall(CeedMatrixPseudoinverse(ceed, interp_1d, Q_1d, P_1d, interp_1d_pinv));
-  CeedCall(CeedMatrixMatrixMultiply(ceed, (const CeedScalar *)interp_1d_pinv, (const CeedScalar *)grad_1d, collo_grad_1d, Q_1d, P_1d, P_1d));
+  // QR Factorization, interp_1d = Q R
+  CeedCall(CeedCalloc(P_1d * Q_1d, &interp_1d_pinv));
+  CeedCall(CeedMatrixPseudoinverse(ceed, basis->interp_1d, Q_1d, P_1d, interp_1d_pinv));
+  CeedCall(CeedMatrixMatrixMultiply(ceed, basis->grad_1d, (const CeedScalar *)interp_1d_pinv, collo_grad_1d, Q_1d, Q_1d, P_1d));
 
-  CeedCall(CeedFree(&interp_1d));
+  // Cleanup
   CeedCall(CeedFree(&interp_1d_pinv));
-  CeedCall(CeedFree(&grad_1d));
-  CeedCall(CeedFree(&tau));
+
   return CEED_ERROR_SUCCESS;
 }
 
@@ -627,36 +622,45 @@ int CeedHouseholderApplyQ(CeedScalar *mat_A, const CeedScalar *mat_Q, const Ceed
 /**
   @brief Return pseudoinverse of a matrix
 
-  @param[in]     ceed Ceed context for error handling
-  @param[in]     mat  Row-major matrix to be factorized in place
-  @param[in]     m    Number of rows
-  @param[in]     n    Number of columns
-  @param[out]    mat_pinv  Row-major pseudoinverse matrix
+  @param[in]  ceed     Ceed context for error handling
+  @param[in]  mat      Row-major matrix to be factorized in place
+  @param[in]  m        Number of rows
+  @param[in]  n        Number of columns
+  @param[out] mat_pinv Row-major pseudoinverse matrix
 
   @return An error code: 0 - success, otherwise - failure
 
   @ref Utility
 **/
 int CeedMatrixPseudoinverse(Ceed ceed, CeedScalar *mat, CeedInt m, CeedInt n, CeedScalar *mat_pinv) {
-  CeedScalar *tau, *I;
+  CeedScalar *tau, *I, *mat_copy;
 
-  CeedCall(CeedCalloc(m * m, &I));
   CeedCall(CeedCalloc(m, &tau));
-  // -- QR Factorization, mat = Q R
-  CeedCall(CeedQRFactorization(ceed, mat, tau, m, n));
-  // -- mat_pinv = R_inv Q^T
-  for (CeedInt i = 0; i < m; i++) I[i * m + i] = 1.0;
-  // ---- Apply R_inv, mat_pinv = I R_inv
-  for (CeedInt i = 0; i < m; i++) {  // Row i
-    mat_pinv[m * i] = I[n * i] / mat[0];
-    for (CeedInt j = 1; j < n; j++) {  // Column j
-      mat_pinv[j + m * i] = I[j + n * i];
-      for (CeedInt k = 0; k < j; k++) mat_pinv[j + m * i] -= mat[j + n * k] * mat_pinv[k + m * i];
-      mat_pinv[j + m * i] /= mat[j + n * j];
+  CeedCall(CeedCalloc(n * n, &I));
+  CeedCall(CeedCalloc(m * n, &mat_copy));
+  memcpy(mat_copy, mat, m * n * sizeof mat[0]);
+
+  // QR Factorization, mat = Q R
+  CeedCall(CeedQRFactorization(ceed, mat_copy, tau, m, n));
+
+  // mat_pinv = R_inv Q^T
+  for (CeedInt i = 0; i < n; i++) I[i * n + i] = 1.0;
+  // -- Apply R_inv, mat_pinv = I R_inv
+  for (CeedInt i = 0; i < n; i++) {  // Row i
+    mat_pinv[n * i] = I[n * i] / mat_copy[0];
+    for (CeedInt j = 1; j < m; j++) {  // Column j
+      mat_pinv[j + n * i] = I[j + n * i];
+      for (CeedInt k = 0; k < j; k++) mat_pinv[j + n * i] -= mat_copy[j + m * k] * mat_pinv[k + n * i];
+      mat_pinv[j + n * i] /= mat_copy[j + n * j];
     }
   }
-  // ---- Apply Q^T, mat_pinv = R_inv Q^T
-  CeedCall(CeedHouseholderApplyQ(mat, mat_pinv, tau, CEED_NOTRANSPOSE, m, n, n, 1, n));
+  // -- Apply Q^T, mat_pinv = R_inv Q^T
+  CeedCall(CeedHouseholderApplyQ(mat_pinv, mat_copy, tau, CEED_NOTRANSPOSE, n, m, n, 1, n));
+
+  // Cleanup
+  CeedCall(CeedFree(&I));
+  CeedCall(CeedFree(&tau));
+  CeedCall(CeedFree(&mat_copy));
 
   return CEED_ERROR_SUCCESS;
 }
```

I ended up pulling out pencil and paper and checking matrix dimensions.

If this works, then next fix would be for the projection function.

rezgarshakeri: Thanks, yeah it works.

julia/LibCEED.jl/src/generated/libceed_bindings.jl
```diff
@@ -1350,6 +1350,10 @@ function CeedHouseholderApplyQ(mat_A, mat_Q, tau, t_mode, m, n, k, row, col)
     ccall((:CeedHouseholderApplyQ, libceed), Cint, (Ptr{CeedScalar}, Ptr{CeedScalar}, Ptr{CeedScalar}, CeedTransposeMode, CeedInt, CeedInt, CeedInt, CeedInt, CeedInt), mat_A, mat_Q, tau, t_mode, m, n, k, row, col)
 end
 
+function CeedMatrixPseudoinverse(ceed, mat, m, n, mat_pinv)
```
jeremylt, line 1353: I've actually been trying to get rid of the other language bindings for utility functions like this - in Julia, Python, and Rust, you have better ways to do this

rezgarshakeri: Hmm, okay I'll remove it, how can I do that in a better way?

jeremylt: In Julia, I'd use  backslash, for example, instead of anything libCEED offers

interface/ceed-basis.c
rezgarshakeri: In projection we ApplyQ with transpose mode in the beginning
```c
  for (CeedInt m = 0; m < num_matrices; m++) {
    // Apply Q^T, interp_from = Q^T interp_from
    memcpy(interp_from, input_from[m], Q * P_from * q_comp * sizeof(input_from[m][0]));
    CeedCall(CeedHouseholderApplyQ(interp_from, interp_to, tau, CEED_TRANSPOSE, Q * q_comp, P_from, P_to, P_from, 1));

    // Apply Rinv, output_project = Rinv interp_from
    for (CeedInt j = 0; j < P_from; j++) {  // Column j
      output_project[m][j + P_from * (P_to - 1)] = interp_from[j + P_from * (P_to - 1)] / interp_to[P_to * P_to - 1];
      for (CeedInt i = P_to - 2; i >= 0; i--) {  // Row i
        output_project[m][j + P_from * i] = interp_from[j + P_from * i];
        for (CeedInt k = i + 1; k < P_to; k++) {
          output_project[m][j + P_from * i] -= interp_to[k + P_to * i] * output_project[m][j + P_from * k];
        }
        output_project[m][j + P_from * i] /= interp_to[i + P_to * i];
      }
    }
  }
```

It is like a reverse of what we wrote in `CeedMatrixPseudoinverse` since we ApplyQ with no transpose mode at the end.

jeremylt: ASAN is saying that this fix is still wrong

jeremylt: Yeah, that patch is wrong and makes NaNs

rezgarshakeri: Yeah, I just checked `t302-basis.c` which was fine I need to write it down on paper.

jeremylt: Alternatively, we could just swap to your previous suggestion
```
/**
  @brief Compute A B_pinv = C

  @param[in]  ceed   Ceed context for error handling
  @param[in]  A      First matrix
  @param[in]  B      Matrix to apply pesudoinverse of
  @param[in]  t_mode A B_pinv (NOTRANSPOSE) or B_pinv A (TRANSPOSE)
  @param[in]  m      Number of rows in C
  @param[in]  n      Number of columns in C
  @param[in]  k      Number of colums in A and B (NOTRANSPOSE)
  @param[out] C      Row-major pseudoinverse matrix

  @ref Utility
**/
int CeedMatrixMatrixPseudoinverseMultiply(Ceed ceed, CeedScalar *A, CeedScalar *B, CeedTransposeMode t_mode, CeedInt m, CeedInt n, CeedInt k, CeedScalar *C) {
```
or similar?

jeremylt: I don't see an obvious reason why this isn't working but it seems to be some stupid indexing issue.

rezgarshakeri: Yeah there is an issue with indices. At the beginning I considered `interp_1d` with size `QxP` and its inverse `interp_1d_pinv` size is `PxQ` so I said if we want to create `grad_1d` with size `QxP` we need `interp_1d x interp_1d_pinv`

jeremylt: The collocated gradient has size QxQ

rezgarshakeri: Ah, yeah. I'll work on it.

interface/ceed-basis.c
```diff
@@ -691,37 +686,45 @@ int CeedHouseholderApplyQ(CeedScalar *mat_A, const CeedScalar *mat_Q, const Ceed
...
+  for (CeedInt i = 0; i < n; i++) I[i * n + i] = 1.0;
+  // -- Apply R_inv, mat_pinv = I R_inv
+  for (CeedInt i = 0; i < n; i++) {  // Row i
+    mat_pinv[n * i] = I[n * i] / mat_copy[0];
+    for (CeedInt j = 1; j < m; j++) {  // Column j
+      mat_pinv[j + n * i] = I[j + n * i];
+      for (CeedInt k = 0; k < j; k++) mat_pinv[j + n * i] -= mat_copy[j + m * k] * mat_pinv[k + n * i];
+      mat_pinv[j + n * i] /= mat_copy[j + n * j];
     }
   }
```
rezgarshakeri: @jeremylt the issue is in `j=1:m` loop. Right now with this structure apparently we have only issues on some qfunction tests with this message
```
/home/rezgar/libCEED/backends/memcheck/ceed-memcheck-qfunction.c:60 in CeedQFunctionApply_Memcheck(): QFunction output 0 entry 0 is NaN after restoring write-only access
```
and basis tests look fine, for example
```
build/t311-basis /cpu/self/ref/serial
(base) rezgar@rezgar:~/libCEED$ build/t311-basis /cpu/self/memcheck/serial
```
but actually `valgrind` shows a leak in `t311-basis`  like
```
==6684== Invalid read of size 8
==6684==    at 0x4869D59: CeedMatrixPseudoinverse (ceed-basis.c:718)
==6684==    by 0x486B578: CeedBasisGetCollocatedGrad (ceed-basis.c:332)
==6684==    by 0x4889957: CeedBasisCreateTensorH1_Ref (ceed-ref-basis.c:297)
==6684==    by 0x486A20E: CeedBasisCreateTensorH1 (ceed-basis.c:1018)
==6684==    by 0x4869F3A: CeedBasisCreateTensorH1 (ceed-basis.c:987)
==6684==    by 0x486D1A5: CeedBasisCreateTensorH1Lagrange (ceed-basis.c:1089)
==6684==    by 0x109469: main (t311-basis.c:39)
==6684==  Address 0x4f78c20 is 0 bytes after a block of size 96 alloc'd
==6684==    at 0x484DA83: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==6684==    by 0x486F056: CeedCallocArray (ceed.c:229)
==6684==    by 0x4869C1F: CeedMatrixPseudoinverse (ceed-basis.c:704)
==6684==    by 0x486B578: CeedBasisGetCollocatedGrad (ceed-basis.c:332)
==6684==    by 0x4889957: CeedBasisCreateTensorH1_Ref (ceed-ref-basis.c:297)
==6684==    by 0x486A20E: CeedBasisCreateTensorH1 (ceed-basis.c:1018)
==6684==    by 0x4869F3A: CeedBasisCreateTensorH1 (ceed-basis.c:987)
==6684==    by 0x486D1A5: CeedBasisCreateTensorH1Lagrange (ceed-basis.c:1089)
==6684==    by 0x109469: main (t311-basis.c:39)
```
and I printed the indices and values of `mat_copy`  and `mat_pinv` which shows we are out of range/size m x n and also have some nan/inf values. I don't know how with these values in `mat_pinv` after applying Q we have correct values for collocated grad.

jeremylt: That's not quite a leak - a leak would be unfreed memory - but its definitely an issue. `mat_pinv[j + n * i]` would be assuming the matrix is NxN, so I agree with Valgrind that's a dimensions mismatch if N > M

julia/LibCEED.jl/src/generated/libceed_bindings.jl
```diff
@@ -1410,6 +1410,10 @@ function CeedHouseholderApplyQ(mat_A, mat_Q, tau, t_mode, m, n, k, row, col)
...
+    ccall((:CeedMatrixPseudoinverse, libceed), Cint, (Ceed, Ptr{CeedScalar}, CeedInt, CeedInt, Ptr{CeedScalar}), ceed, mat, m, n, mat_pinv)
+end
+
```
jeremylt, line 1416: Oops, this snuck back in. We shouldn't be adding our internal linear algebra routines to any external packages.

rezgarshakeri: Hmm, I'll remove it.

interface/ceed-basis.c
jeremylt:
```
/var/lib/gitlab-runner/builds/N8GCsqus/0/libceed/libCEED/backends/memcheck/ceed-memcheck-qfunction.c:60 in CeedQFunctionApply_Memcheck(): QFunction output 0 entry 0 is NaN after restoring write-only access
```

this means that either the QFunction is not covering all qpts (which we know isn't the case) or the QFunction is getting NaNs in as input and then writing them out. If you run with  `CEED_DEBUG=1`, you should see an error about the write only access still having a NaN earlier in the OperatorApply, after `CeedBasisApply` for the input, I'd bet

rezgarshakeri:
I don't understand the rust failure. How can I fix it?

jeremylt:
The rust test is saying that the incorrect numerical result was computed, so the pesudoinverse probably isn't right in that case

rezgarshakeri:
> The rust test is saying that the incorrect numerical result was computed, so the pesudoinverse probably isn't right in that case

Hmm, so how the other tests are fine and rust is not? I checked only with `make prove` in my machine and I got 
```
All tests successful.
Files=232, Tests=2016,  6 wallclock secs ( 1.10 usr  0.34 sys + 44.30 cusr 11.80 csys = 57.54 CPU)
Result: PASS
```

jeremylt:
I'd start by seeing if there are any differences between how the Rust test t501 is written vs the C version of the test. It's possible it has a slightly different setup.

The Rust tests allow less memory misbehavior on our part, so I'm inclined to trust them when they say something went wrong.

interface/ceed-basis.c
```diff
@@ -318,37 +318,21 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
...
 
-  // Apply Q^T, collo_grad_1d = collo_grad_1d Q^T
-  CeedCall(CeedHouseholderApplyQ(collo_grad_1d, interp_1d, tau, CEED_NOTRANSPOSE, Q_1d, Q_1d, P_1d, 1, Q_1d));
+  // QR Factorization, interp_1d = Q R
+  CeedCall(CeedCalloc(P_1d * Q_1d, &interp_1d_pinv));
 
-  CeedCall(CeedFree(&interp_1d));
-  CeedCall(CeedFree(&grad_1d));
-  CeedCall(CeedFree(&tau));
+  CeedCall(CeedMatrixPseudoinverse(ceed, basis->interp_1d, Q_1d, P_1d, interp_1d_pinv));
```
jeremylt, line 332: Won't this destroy interp_1d? And we really, really need interp_1d not destroyed

interface/ceed-basis.c
```diff
@@ -699,6 +683,50 @@ int CeedHouseholderApplyQ(CeedScalar *mat_A, const CeedScalar *mat_Q, const Ceed
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Return pseudoinverse of a matrix
+
+  @param[in]     ceed      Ceed context for error handling
+  @param[in]     mat       Row-major matrix to be factorized in place
```
jeremylt, line 690: Ah, no, you make a copy, so it isn't factored in place (thus destroying the original data)

jeremylt:
```suggestion
  @param[in]     mat       Row-major matrix to compute pseudoinverse of
```

rezgarshakeri:
> I'd start by seeing if there are any differences between how the Rust test t501 is written vs the C version of the test. It's possible it has a slightly different setup.
> 
> The Rust tests allow less memory misbehavior on our part, so I'm inclined to trust them when they say something went wrong.

I see, I checked the pseudoinverse with python. For example `t301-basis`
```c
  CeedScalar A[12]    = {1, -1, 4, 1, 4, -2, 1, 4, 2, 1, -1, 0};
  CeedScalar A_inv[12] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
  CeedCall(CeedMatrixPseudoinverse(ceed, A, 4, 3, A_inv));
  for (CeedInt i =0; i < 12; i++) printf("A_inv[%d]=%f\n", i, A_inv[i]);
```
gives
```
A_inv[0]=0.200000
A_inv[1]=0.300000
A_inv[2]=-0.100000
A_inv[3]=0.600000
A_inv[4]=-0.050000
A_inv[5]=0.050000
A_inv[6]=0.150000
A_inv[7]=-0.150000
A_inv[8]=0.125000
A_inv[9]=-0.125000
A_inv[10]=0.125000
A_inv[11]=-0.125000
```
which is exactly the same as python
```
array([[ 0.2  ,  0.3  , -0.1  ,  0.6  ],
       [-0.05 ,  0.05 ,  0.15 , -0.15 ],
       [ 0.125, -0.125,  0.125, -0.125]])
```
so maybe we have memory/size issue in the middle of the code but result is correct.

rezgarshakeri:
Now the error is clear
```
Incorrect interval length computed. Expected: 2.0, Found: 1.999999999999999, Error: 1.110223024625e-15
```

jeremylt:
```
Incorrect interval length computed. Expected: 2.0, Found: 1.999999999999999, Error: 1.110223024625e-15
```
So this gave us a mild degradation in accuracy on this test case, but not enough to be worried. I'd say we bump the tolerance to `5e-15` for this test.

rezgarshakeri:
> ```
> Incorrect interval length computed. Expected: 2.0, Found: 1.999999999999999, Error: 1.110223024625e-15
> ```
> 
> So this gave us a mild degradation in accuracy on this test case, but not enough to be worried. I'd say we bump the tolerance to `5e-15` for this test.

Yeah. I'll update that.

rezgarshakeri:
@jeremylt  Do I need to update
```
    /// // Check
    /// let sum: Scalar = v.view()?.iter().sum();
    /// assert!(
    ///     (sum - 2.0).abs() < 10.0 * libceed::EPSILON,
    ///     "Incorrect interval length computed"
    /// );
```
in operator.rs to 
```
    /// // Check
    /// let sum: Scalar = v.view()?.iter().sum();
    /// assert!(
    ///     (sum - 2.0).abs() < 50.0 * libceed::EPSILON,
    ///     "Incorrect interval length computed. Expected: 2.0, Found: {}, Error: {:.12e}"
    /// );
```
?

jeremylt:
Probably. And the t501 in main should probably use CeedEpsilon too

rezgarshakeri:
> Probably. And the t501 in main should probably use CeedEpsilon too

I'll make small PR for that.

interface/ceed-basis.c
```diff
@@ -318,37 +302,20 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
...
-  CeedCall(CeedMalloc(Q_1d, &tau));
-  memcpy(interp_1d, (basis)->interp_1d, Q_1d * P_1d * sizeof(basis)->interp_1d[0]);
-  memcpy(grad_1d, (basis)->grad_1d, Q_1d * P_1d * sizeof(basis)->interp_1d[0]);
+  CeedInt     P_1d, Q_1d;
+  CeedScalar *interp_1d_pinv;
 
-  // QR Factorization, interp_1d = Q R
   CeedCall(CeedBasisGetCeed(basis, &ceed));
-  CeedCall(CeedQRFactorization(ceed, interp_1d, tau, Q_1d, P_1d));
-  // Note: This function is for backend use, so all errors are terminal and we do not need to clean up memory on failure.
```
jeremylt, line 333: We should still keep this note - it explains why we don't  have  a cleanup section for this function

jeremylt (approved):
:tada: looks like all the checks are happy!

interface/ceed-basis.c
rezgarshakeri: Should I add that comments under `CeedQRFactorization` used in `CeedMatrixPseudoinverse` function?

jeremylt: Probably just above this block of code

jedbrown (approved):
Nice!

rezgarshakeri:
@jeremylt feel free to squash+merge if you think we can ignore `ci` failure.

jeremylt:
I'd try rebasing on main and seeing if that fixes the CI issues

rezgarshakeri:
So it seems this PR breaks some fluid tests I don't know why.

jrwrigh:
Unfortunately I don't have quick ~answers~ solutions here. And it'll be a little while before I can get back to this. 

I'm guessing this is probably something to do with the `basis_x_to_q`/`basis_x_to_q_sur`, as that's the only place that the pseudo-inverse code would be called (via `CeedBasisCreateProjection`). Obviously your code has been validated in other tests (`t319` and the multi-grid ones too), so it's probably the coordinate locations getting slightly fuzzy. Should the result of the inverse be exactly the same as before, or is it prone to being slightly different? Assuming the latter, any ideas on how much different?

This seems like similar issues with tests we ran into in #1264 (which incidentally introduced `CeedBasisCreateProjection` in the first place). Though those were only solution tolerance failures, not the solver divergence that it's showing for some of the tests.

rezgarshakeri:
> Unfortunately I don't have quick ~answers~ solutions here. And it'll be a little while before I can get back to this.
> 
> I'm guessing this is probably something to do with the `basis_x_to_q`/`basis_x_to_q_sur`, as that's the only place that the pseudo-inverse code would be called (via `CeedBasisCreateProjection`). Obviously your code has been validated in other tests (`t319` and the multi-grid ones too), so it's probably the coordinate locations getting slightly fuzzy. Should the result of the inverse be exactly the same as before, or is it prone to being slightly different? Assuming the latter, any ideas on how much different?
> 
> This seems like similar issues with tests we ran into in #1264 (which incidentally introduced `CeedBasisCreateProjection` in the first place). Though those were only solution tolerance failures, not the solver divergence that it's showing for some of the tests.

One of the rust test had issue and we update the tolerance. The difference is really small, for example adding the inverse code produced this failure
```
Incorrect interval length computed. Expected: 2.0, Found: 1.999999999999999, Error: 1.110223024625e-15
```
and we updated the tolerance to `50 * machine_epsilon`

rezgarshakeri:
> I agree that a diverging linear solve is an indication of a problem, though why is it only showing on GPU and not CPU?

Any idea how we are going to fix ci here?

jeremylt:
I don't have a good guess at what might be causing this, no. I'd start looking at what's different about that test case between CPU and GPU

jrwrigh:
Same. I'd start with trying to see how small you can make the problem while reproducing the issue, then dig down into comparing intermediate states.

jeremylt:
It looks like those fluids tests aren't even reaching the same solution now, which is bizarre
```
1..1
Test: fluids-navierstokes Blasius, SGS DataDriven
  $ build/fluids-navierstokes -ceed /gpu/cuda/ref -options_file examples/fluids/tests-output/blasius_stgtest.yaml -sgs_model_type data_driven -sgs_model_dd_leakyrelu_alpha 0.3 -sgs_model_dd_parameter_dir examples/fluids/dd_sgs_data -ts_dt 2e-9 -state_var primitive -ksp_rtol 1e-12 -snes_rtol 1e-12 -stg_mean_only -stg_fluctuating_IC -test_type solver -compare_final_state_atol 1e-10 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius-sgs-data-driven.bin
FAIL: stdout
Output: 
Test failed with error norm 0.00107214
```

jeremylt:
The issues are with `Spanwise Turbulence Statistics` and `Blasius` tests, so what do those have in common? And how do they differ from the rest of the fluids code?

jrwrigh:
> The issues are with `Spanwise Turbulence Statistics` and `Blasius` tests, so what do those have in common? And how do they differ from the rest of the fluids code?

They're all using the STG boundary condition, fluctuating or not. As I [mentioned earlier](https://github.com/CEED/libCEED/pull/1251#issuecomment-1769020943), the only usage of the PseudoInverse function is for the `CeedBasisCreateProjection` for `basis_x_to_q`/`basis_x_to_q_sur`. Given the boundary conditions connection, it's probably for `basis_x_to_q_sur`. 

Since even some of these include `stg_mean_only: true`, then that narrows it down to the interpolation of the inflow profile onto the nodes in `qfunctions/stg_shur14.h:InterpolateProfile` (which is used in `StgShur14Inflow` and `StgShur14InflowStrongQF` in the same file).

Given that the CPU tests are passing, it's probably some fuzziness with the coordinates.

jeremylt:
Ah, that's right. *Read the thread more carefully Jeremy*

The error being E-3 instead of E-10 feels huge though

jrwrigh:
Agreed. I'd believe that kind of error from coordinate "fuzziness" for the SGS DataDriven problems (which have a neural network eval in them, so all kinds of weird non-linearities in that), but the rest of them make me think something else is very wrong. 
Somewhat similar for the problems with `stg_mean_only: false`, as the boundary condition they're interpolating isn't smooth to the mesh, so I could believe small changes to the coordinates could result in larger changes to the end result.

But there are failing tests without the NN evaluation or `stg_mean_only: false`, so I'm just as lost.

The `basis_x_to_q_sur` for the strong BCs should be identity (projecting GLL nodes to GLL quadrature points). But there's also weak BC failures which would be GLL nodes to Gauss quadrature points.

I'd be curious to see the differences in the basis operator matrices between main and this branch for `basis_x_to_q_sur`. At least that'd be the first place I'd check.

jeremylt:
The fact that its only on GPU also has me confused because if there is a problem with the basis operator matrix itself, it should be in all backends since this change is in the core interface code. Maybe some core issue in the CUDA backend coping over the right data?

jeremylt:
Ok, I don't think I love this per sey, but this "fixes" the failures for me locally
```diff
$ git diff
diff --git a/interface/ceed-basis.c b/interface/ceed-basis.c
index 829ab543..31108865 100644
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -271,6 +271,10 @@ static int CeedBasisCreateProjectionMatrices(CeedBasis basis_from, CeedBasis bas
     // output_project = interp_to^+ * interp_from
     memcpy(interp_from, input_from[m], Q * P_from * q_comp * sizeof(input_from[m][0]));
     CeedCall(CeedMatrixMatrixMultiply(ceed, interp_to_inv, input_from[m], output_project[m], P_to, P_from, Q * q_comp));
+    // Round zero to machine precision
+    for (CeedInt i = 0; i < P_to * P_from; i++) {
+      if (fabs(output_project[m][i]) < 10 * CEED_EPSILON) output_project[m][i] = 0.0;
+    }
   }
 
   // Cleanup
```

I think this means the projection is really sensitive to numerical error. Basically, I discovered that the entries of the projection basis interp matrix were different from main by ~e-16 or so, and that mattered more to the GPU backends than the CPU backends for some reason.

interface/ceed-basis.c
```diff
@@ -605,7 +572,7 @@ int CeedMatrixMatrixMultiply(Ceed ceed, const CeedScalar *mat_A, const CeedScala
       CeedScalar sum = 0;
 
       for (CeedInt k = 0; k < kk; k++) sum += mat_A[k + i * kk] * mat_B[j + k * n];
-      mat_C[j + i * n] = sum;
+      mat_C[j + i * n] = (fabs(sum) < 5 * CEED_EPSILON) ? 0.0 : sum;
```
jeremylt, line 575: I'm not sure if we should mess around with how every matrix-matrix product works to fix one flakey test?

jeremylt: Also, 5*eps I don't think was enough for me locally.

rezgarshakeri: Yeah, I was thinking to avoid an extra for loop for rounding in `CeedBasisCreateProjectionMatrices`. We can try that.

jeremylt: The projection creation is really infrequent, so we don't have to worry about high efficiency there.

rezgarshakeri: Same error. Can we change the tolerance for that example?

jeremylt: The GPU job hasn't finished yet. The rust error is something else that I'm looking at

jeremylt: Yeah, the last commit is good. The rust issue is unrelated

jeremylt:
This is needed but may not fix the Rust issue. There shouldn't be any problem with any slice from raw parts and I can't reproduce that locally
```diff
$ git diff
diff --git a/examples/rust/ex1-volume/src/main.rs b/examples/rust/ex1-volume/src/main.rs
index ca755438..ed22b9fb 100644
--- a/examples/rust/ex1-volume/src/main.rs
+++ b/examples/rust/ex1-volume/src/main.rs
@@ -247,7 +247,7 @@ fn example_1(options: opt::Opt) -> libceed::Result<()> {
         );
     }
     let tolerance = match dim {
-        1 => 100.0 * libceed::EPSILON,
+        1 => 500.0 * libceed::EPSILON,
         _ => 1E-5,
     };
     let error = (volume - exact_volume).abs();
```

examples/fluids/tests-output/blasius_stgtest.yaml
```diff
@@ -6,6 +6,7 @@ ts:
   type: 'beuler'
   dt: 2e-6
   max_steps: 5
+  max_snes_failures: 40
```
jeremylt, line 9: We can drop these changes now

examples/fluids/tests-output/blasius_test.yaml
```diff
@@ -6,6 +6,7 @@ ts:
   type: 'beuler'
   dt: 2e-6
   max_steps: 5
+  max_snes_failures: 40
```
jeremylt, line 9: here

examples/fluids/tests-output/stats_test.yaml
```diff
@@ -6,6 +6,7 @@ ts:
   type: 'beuler'
   dt: 2e-6
   max_steps: 10
+  max_snes_failures: 40
```
jeremylt, line 9: here

jeremylt:
Well, Rust is doing some crappy misbehavior I haven't been able to identify yet

jeremylt:
Its happening on a different branch too, so the Rust issue isn't related to this branch and is something that happened to the nightly toolchain

jrwrigh:
Possibly another solution (though I'm not sure how feasible it would be) is to detect whether the projection should result in an identity matrix and, if it should, just return an identity matrix. Determining that ahead of time while not limiting ourselves to nodal basis functions might be a bit challenging. The other option would be verifying identity after matrix-matrix multiply, but that might be just as ad-hoc as forcing small values to zero.

jrwrigh:
> Basically, I discovered that the entries of the projection basis interp matrix were different from main by ~e-16 or so,

I'm a bit surprised by that. Aren't those matrices created by CPU code, not the GPU? Or is the PseudoInverse code also used for generating those projection matrices?

rezgarshakeri:
> This is needed but may not fix the Rust issue. There shouldn't be any problem with any slice from raw parts and I can't reproduce that locally
> 
> ```diff
> $ git diff
> diff --git a/examples/rust/ex1-volume/src/main.rs b/examples/rust/ex1-volume/src/main.rs
> index ca755438..ed22b9fb 100644
> --- a/examples/rust/ex1-volume/src/main.rs
> +++ b/examples/rust/ex1-volume/src/main.rs
> @@ -247,7 +247,7 @@ fn example_1(options: opt::Opt) -> libceed::Result<()> {
>          );
>      }
>      let tolerance = match dim {
> -        1 => 100.0 * libceed::EPSILON,
> +        1 => 500.0 * libceed::EPSILON,
>          _ => 1E-5,
>      };
>      let error = (volume - exact_volume).abs();
> ```

Do I still need this change?

jeremylt:
The pesudoinverse is use in the core of the library to create the projection matrices. The new computations are in this weird place where they are not different enough to cause appreciable differences on the CPU but they do create differences on the GPU, likely due to slightly different algebra on the machine to apply the basis matrices

jeremylt:
> Do I still need this change?

I think so, yea

rezgarshakeri:
I don't know why `codecov` breaks.

jeremylt:
Its a minor change and codecov is flaky, so I think we can ignore it

rezgarshakeri:
Cool, so I merge it.

jrwrigh:
> The pesudoinverse is use in the core of the library to create the projection matrices.

Oh, I misread the earlier message to be that the interpolation matrices themselves had slightly different values. This makes more sense.

PR #1254, opened 2023-07-12
Unify single and double precision `avx` backend code

sebastiangrimberg:
Similar to the `xsmm` backend done in https://github.com/CEED/libCEED/pull/1248, this removes duplicate code between the `fp64` and `fp32` implementations.


jeremylt:
Unfortunately, since this is a branch on your fork, our GitLab CI cannot run in this PR. But that's where the single precision tests occur.

backends/ref/ceed-ref-tensor.c
```diff
@@ -36,11 +36,6 @@ static int CeedTensorContractApply_Ref(CeedTensorContract contract, CeedInt A, C
   return CEED_ERROR_SUCCESS;
 }
 
-//------------------------------------------------------------------------------
```
jeremylt, line 39: This is here intentionally - this isn't needed for functionality but the `ref` backend is the "example" other people can use to make their own backend, so I kept this in as in example.

sebastiangrimberg:
Closing in favor of #1255.

PR #1255, opened 2023-07-12
Unify single and double precision avx backend code

sebastiangrimberg:
Reopening #1254 from within CEED/libCEED for GitLab CI tests.

Similar to the xsmm backend done in https://github.com/CEED/libCEED/pull/1248, this removes duplicate code between the fp64 and fp32 implementations.


backends/avx/ceed-avx-tensor.c
```diff
@@ -35,43 +52,43 @@ static inline int CeedTensorContract_Avx_Blocked(CeedTensorContract contract, Ce
     // Blocks of 4 rows
     for (CeedInt j = 0; j < (J / JJ) * JJ; j += JJ) {
       for (CeedInt c = 0; c < (C / CC) * CC; c += CC) {
-        __m256d vv[JJ][CC / 4];  // Output tile to be held in registers
+        rtype vv[JJ][CC / 4];  // Output tile to be held in registers
```
jeremylt, line 55: Should this magic 4 also be updated to be 4 doubles or 8 floats? (I should have made this change ages ago but forgot about it)

backends/avx/ceed-avx-tensor.c
```diff
@@ -97,26 +114,26 @@ static inline int CeedTensorContract_Avx_Remainder(CeedTensorContract contract,
...
+        for (CeedInt jj = 0; jj < JJ; jj++) vv[jj] = loadu(&v[(a * J + j + jj) * C + c]);
 
         for (CeedInt b = 0; b < B; b++) {
-          __m256d tqu;
-          if (C - c == 1) tqu = _mm256_set_pd(0.0, 0.0, 0.0, u[(a * B + b) * C + c + 0]);
-          else if (C - c == 2) tqu = _mm256_set_pd(0.0, 0.0, u[(a * B + b) * C + c + 1], u[(a * B + b) * C + c + 0]);
-          else if (C - c == 3) tqu = _mm256_set_pd(0.0, u[(a * B + b) * C + c + 2], u[(a * B + b) * C + c + 1], u[(a * B + b) * C + c + 0]);
-          else tqu = _mm256_loadu_pd(&u[(a * B + b) * C + c]);
+          rtype tqu;
+          if (C - c == 1) tqu = set(0.0, 0.0, 0.0, u[(a * B + b) * C + c + 0]);
```
jeremylt, line 122: ah, I think this is why I didn't use 8 floats originally

jeremylt: There's probably a way around this, but I haven't invested time since floats aren't a hotspot for performance in any of our current applications.

sebastiangrimberg: Yeah the block size of 4 is pretty hardcoded throughout, I would shy away from changing it if it's not an issue.

jeremylt (approved):
Lgtm

PR #1256, opened 2023-07-13
fluids: Set KSP max iterations to 1000

jrwrigh:


jeremylt:
Seems fine to me. Do we want a lower default for CI?

examples/fluids/src/setupts.c
```diff
@@ -475,6 +475,13 @@ PetscErrorCode TSSolve_NS(DM dm, User user, AppCtx app_ctx, Physics phys, Vec *Q
   }
   PetscCall(TSGetAdapt(*ts, &adapt));
   PetscCall(TSAdaptSetStepLimits(adapt, 1.e-12 * user->units->second, 1.e2 * user->units->second));
+  {
+    SNES snes;
+    KSP  ksp;
+    PetscCall(TSGetSNES(*ts, &snes));
+    PetscCall(SNESGetKSP(snes, &ksp));
+    PetscCall(KSPSetTolerances(ksp, PETSC_DEFAULT, PETSC_DEFAULT, PETSC_DEFAULT, 1000));
```
jedbrown, line 483: What is the rationale here? To waste less time in a "bad" configuration? Should we also use `PCSetType(pc, PCVPBJACOBI)` or keep the default bjacobi/ilu, which is better for some problems on CPU?

jrwrigh: > To waste less time in a "bad" configuration?

Yes, exactly.

> Should we also use `PCSetType(pc, PCVPBJACOBI)` or keep the default bjacobi/ilu, which is better for some problems on CPU?

I'd say keep the same default settings for right now, with default being the """CPU optimal""". If we do want to switch to have `VPBJACOBI`, I feel like we should switch over to the `-amat_type shell -pmat_pbdiagonal` as defaults to go along with it.

jedbrown: Okay, and I take it needing more than 1000 iterations should continue to be interpreted as failure, not recovered by continuing Newton (or shortening a time step)? I think the standard blasius setup with vpbjacobi needed more than 1000 with `gmres` (but not `lgmres`).

jedbrown: That would be user-friendly, but also makes debugging by switching back to the host harder. If we're going to change defaults, I'd rather make them the same for all backends.

jrwrigh: >  I take it needing more than 1000 iterations should continue to be interpreted as failure, not recovered by continuing Newton (or shortening a time step)? 

Yes, though perhaps we should wait for @KennethEJansen to confirm.

> I think the standard blasius setup with vpbjacobi needed more than 1000 with `gmres` (but not `lgmres`).

Yeah, you're right. I'm getting around 2000 iterations for that configuration on CPU. But for production runs, we wouldn't be running that kind of problem. IIRC, PHASTA maxes out at 200 iterations by default.

jedbrown: We might want to modify the blasius setup so it's not so stiff.

> PHASTA maxes out at 200 iterations by default.

... and then it aborts the run or it just continues with the next Newton step? The desired semantics for "this problem is tough" are different from "maybe GPU-aware MPI is broken and we should stop wasting time on it".

jrwrigh: It continues to the next Newton step. I was using it more in the "if we're ok with moving forward at 200, then at 1000 something must be wrong."

PR #1260, opened 2023-07-18
fluids: Don't show partition stats for serial jobs

jrwrigh:
- It results in nans (from divide by zeros) which may confuse users and doesn't actually show anything useful.
 
Originally brought up by @KennethEJansen


jrwrigh:
We could also just set the `max/median` value to zero for serial jobs, but I think it also makes sense to not show the interface values at all when they don't exist. I'm up for either solution.

jeremylt (approved):

PR #1264, opened 2023-07-25
PetscFE determines basis for examples/fluids/navierstokes

KennethEJansen:
As we prepare to bring simplex meshes into the examples/fluids/navierstokes application within libCEED, it is important to let PETSc define the basis functions. This code revision followed similar work done in Ratel to address this issue.

This has been completed at this point for the primary operators but not for the Data/Learning extensions but hopefully those will be resolved within this PR.

As all changes are within examples/fluids and off of a current main, all tests of libCEED should still pass. 

As the work mimicked  Ratel development (and work from a branch of Rezgar Shekeri) to acquire the basis information from PETSc,  credit should go those efforts while responsibility for miss-interpretations of that effort belong to Kenneth Jansen with co-author help from James Wright. 

At this time the fluids example has been verified correct  without loss of computational efficiency on Polaris in both serial and parallel including STG with Q1, Q2, and Q3 elements. Further tests with simplex elements are ongoing and updates will be provided as they become available.


jrwrigh:
TODO:
 - [ ] Remove `Q_dim` from `GetRestrictionForDomain` arguments (just use `Q` for quadrature points per element, aligned with CeedBasis nomenclature)
 - [ ] Style
 - [ ] Fix all calls to `GetRestrictionForDomain`
 - [ ] Fix test errors.

KennethEJansen:
In the failing tests I see `-q_extra 2`  I guess it is obvious that PestscFE does not understand/honor this request.

@jedbrown how do you propose we resolve this?  Is there a way to communicate this request to PETSc?

jrwrigh:
> In the failing tests I see `-q_extra 2` I guess it is obvious that PestscFE does not understand/honor this request.
> 
> @jedbrown how do you propose we resolve this? Is there a way to communicate this request to PETSc?

I'm guessing we probably just need to change: https://github.com/CEED/libCEED/blob/51f8a934fc545fa09816d615483b8962eeaa9abd/examples/fluids/src/setupdm.c#L53
from `PETSC_DECIDE` to the actual quadrature order we want. See [PETSc docs](https://petsc.org/release/manualpages/FE/PetscFECreateLagrange/#petscfecreatelagrange)

examples/fluids/src/grid_anisotropy_tensor.c
```diff
@@ -145,17 +142,18 @@ PetscErrorCode GridAnisotropyTensorCalculateCollocatedVector(Ceed ceed, User use
...
-  CeedInt Q_dim = CeedIntPow(num_qpts_1d, dim);
   CeedElemRestrictionGetNumElements(ceed_data->elem_restr_q, &loc_num_elem);
-  CeedElemRestrictionCreateStrided(ceed, loc_num_elem, Q_dim, *num_comp_aniso, *num_comp_aniso * loc_num_elem * Q_dim, CEED_STRIDES_BACKEND,
+  CeedElemRestrictionCreateStrided(ceed, loc_num_elem, num_qpts, *num_comp_aniso, *num_comp_aniso * loc_num_elem * num_qpts, CEED_STRIDES_BACKEND,
                                    elem_restr_grid_aniso);
 
+  // TODO Use CreateBasisFromPlex, but need set custom number of components
+  CeedInt num_qpts_1d;
+  CeedBasisGetNumQuadraturePoints1D(ceed_data->basis_q, &num_qpts_1d);
   CeedBasisCreateTensorH1Lagrange(ceed, dim, *num_comp_aniso, num_nodes_1d, num_qpts_1d, CEED_GAUSS, &basis_grid_aniso);
```
jrwrigh, line 157: Not sure how we'd get around this, but I don't want it to hold up this PR.

jeremylt: If you want to use `CreateBasisFromPlex`, then the DMPlex FE should have the number of components you need. We create auxiliary DMs for extra stuff like this in Ratel

jrwrigh: Right, I create auxillary DMs for most other situations, except for this one, as there isn't a need for a intermediary DM otherwise. It seems odd to create an entire DM just to get the basis, but I don't see a problem with it either.

jrwrigh: ...or I can actually read the code and realize A) we do have a DM built for this and B) that `basis_grad_aniso` isn't even used in this context, so I don't even need to build a basis.

KennethEJansen: Agreed this should not hold up the PR.  Who marks this resolved?

examples/fluids/src/setuplibceed.c
```diff
@@ -213,33 +210,230 @@ PetscErrorCode SetupBCQFunctions(Ceed ceed, PetscInt dim_sur, PetscInt num_comp_
...
+    PetscCall(PetscFEGetNumComponents(fe, &num_comp));
+    PetscCall(PetscFEGetDualSpace(fe, &dual_space));
+    PetscCall(PetscDualSpaceGetDimension(dual_space, &num_dual_basis_vectors));
+    P = num_dual_basis_vectors / num_comp;
+    PetscCall(PetscQuadratureGetData(quadrature, NULL, NULL, &Q, NULL, NULL));
+
+    CeedInt P_1d = (CeedInt)round(pow(P, 1.0 / dim));
+    CeedInt Q_1d = (CeedInt)round(pow(Q, 1.0 / dim));
+
+    CeedBasisCreateTensorH1Lagrange(ceed, dim, num_comp, P_1d, Q_1d, quadMode, basis);
```
jeremylt, line 400: We cut out the `quadMode` (convention would be `quad_mode`, right?) and this creation of our own `TensorH1Lagrange` basis if we could get the 1D basis matrices from the FE. Did we talk about how easy it would be to get that out of the FE @jedbrown?

examples/fluids/src/differential_filter.c
```diff
@@ -64,9 +62,8 @@ PetscErrorCode DifferentialFilterCreateOperators(Ceed ceed, User user, CeedData
       char                field_name[PETSC_MAX_PATH_LEN];
       CeedElemRestriction elem_restr_filter;
       CeedBasis           basis_filter;
-
-      PetscCall(GetRestrictionForDomain(ceed, dm_filter, 0, 0, 0, i, num_qpts_1d, 0, &elem_restr_filter, NULL, NULL));
-      CeedBasisCreateTensorH1Lagrange(ceed, dim, diff_filter->num_field_components[i], num_nodes_1d, num_qpts_1d, CEED_GAUSS, &basis_filter);
+      PetscCall(GetRestrictionForDomain(ceed, dm_filter, 0, 0, 0, i, -1, 0, &elem_restr_filter, NULL, NULL));
```
jeremylt, line 65: Silly question, but why `-1` in the args?

examples/fluids/src/grid_anisotropy_tensor.c
```diff
@@ -124,6 +121,7 @@ PetscErrorCode GridAnisotropyTensorProjectionSetupApply(Ceed ceed, User user, Ce
   // -- Cleanup
   PetscCall(NodalProjectionDataDestroy(grid_aniso_proj));
   PetscCall(OperatorApplyContextDestroy(l2_rhs_ctx));
+  PetscCall(OperatorApplyContextDestroy(mass_matop_ctx));
```
jeremylt, line 124: I'm scanning for differences - this line was added but doesn't seem to be part of the main thrust of this PR?

examples/fluids/src/grid_anisotropy_tensor.c
```diff
@@ -166,11 +151,13 @@ PetscErrorCode GridAnisotropyTensorCalculateCollocatedVector(Ceed ceed, User use
...
   CeedOperatorSetField(op_colloc, "v", *elem_restr_grid_aniso, CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
-  CeedOperatorSetNumQuadraturePoints(op_colloc, CeedIntPow(num_qpts_1d, dim));
+  CeedOperatorSetNumQuadraturePoints(op_colloc, num_qpts);
 
   CeedElemRestrictionCreateVector(*elem_restr_grid_aniso, aniso_colloc_ceed, NULL);
 
   CeedOperatorApply(op_colloc, CEED_VECTOR_NONE, *aniso_colloc_ceed, CEED_REQUEST_IMMEDIATE);
 
+  CeedQFunctionDestroy(&qf_colloc);
+  CeedOperatorDestroy(&op_colloc);
```
jeremylt, line 161: This is also a change - should we do some scrubbing of the fluids app for memory leaks? I thought we had those checks in CI?

examples/fluids/src/setupdm.c
```diff
@@ -123,7 +156,7 @@ PetscErrorCode VizRefineDM(DM dm, User user, ProblemData *problem, SimpleBC bc,
     if (i + 1 == user->app_ctx->viz_refine) d = 1;
     PetscCall(DMGetVecType(dm, &vec_type));
     PetscCall(DMSetVecType(dm_hierarchy[i + 1], vec_type));
-    PetscCall(SetUpDM(dm_hierarchy[i + 1], problem, d, bc, phys));
+    PetscCall(SetUpDM(dm_hierarchy[i + 1], problem, d, 0, bc, phys));
```
jeremylt, line 159: We should probably drop a comment here explaining why its ok to have a dissimilar quadrature space on this DM

examples/fluids/src/differential_filter.c
jrwrigh: Mostly because I can't put a `NULL` there. That argument isn't used unless it's generating a collocated element restriction. Conversely the (often) four `0` zero arguments before it _are_ always used so I put -1 there to make it look different.

Should probably just create a separate "GetCollocatedRestriction" function, but I was trying to tackle one thing at a time.

examples/fluids/src/grid_anisotropy_tensor.c
jrwrigh: Correct. It came about because I found several leaks/unused variables in the other function in the file. I've tried with and without the changes, but the test results are still the same.

examples/fluids/src/grid_anisotropy_tensor.c
jrwrigh: We definitely should do some scrubbing. I'm not sure if we got memory leak checks into CI.

examples/fluids/src/setupdm.c
jrwrigh: I could also make it have the same quadrature space? I'm up for either, this PR was kinda caught mid-hack to get working code before going back and cleaning up.

examples/fluids/src/grid_anisotropy_tensor.c
jrwrigh: Sounds good to me. #1266

examples/fluids/src/setuplibceed.c
```diff
@@ -213,33 +210,230 @@ PetscErrorCode SetupBCQFunctions(Ceed ceed, PetscInt dim_sur, PetscInt num_comp_
...
+      PetscCall(DMGetLocalSection(dm, &section));
+      PetscCall(PetscSectionGetClosurePermutation(section, (PetscObject)dm, dim, num_comp * P, &permutation));
+      PetscCall(ISGetIndices(permutation, &permutation_indices));
+    }
+
+    // -- Copy interp, grad matrices
+    PetscCall(PetscCalloc(P * Q * sizeof(CeedScalar), &interp));
+    PetscCall(PetscCalloc(P * Q * dim * sizeof(CeedScalar), &grad));
+    const CeedInt c = 0;
+    for (CeedInt q = 0; q < Q; q++) {
```
nbeams, line 330: What is the point of `c` here if it's always 0?

examples/fluids/src/turb_spanstats.c
```diff
@@ -178,24 +178,27 @@ PetscErrorCode GetQuadratureCoords(Ceed ceed, DM dm, CeedElemRestriction elem_re
...
   CeedElemRestrictionCreateVector((*stats_data)->elem_restr_parent_qd, &(*stats_data)->q_data, NULL);
 
-  CeedBasisCreateTensorH1Lagrange(ceed, dim, num_comp_x, 2, Q, CEED_GAUSS, &(*stats_data)->basis_x);
-  CeedBasisCreateTensorH1Lagrange(ceed, dim, num_comp_stats, P, Q, CEED_GAUSS, &(*stats_data)->basis_stats);
+  {
+    DM dm_coord;
+    PetscCall(DMGetCoordinateDM(dm, &dm_coord));
+    PetscCall(CreateBasisFromPlex(ceed, dm_coord, 0, 0, 0, 0, CEED_GAUSS, &(*stats_data)->basis_x));
+    PetscCall(CreateBasisFromPlex(ceed, dm, 0, 0, 0, 0, CEED_GAUSS, &(*stats_data)->basis_stats));
+  }
```
jeremylt, line 201: This is going to be a completely different basis than your one with P and Q from `CeedBasisGetNum[Nodes,QuadraturePoints]1D`

examples/fluids/src/grid_anisotropy_tensor.c
jrwrigh: (fixed in #1266)

examples/fluids/src/setuplibceed.c
jrwrigh: That is a good point. I have no idea. That code was taken from Ratel: https://gitlab.com/micromorph/ratel/-/blob/main/src/internal/ratel-dm.c?ref_type=heads#L292-302

@jeremylt any ideas?

jeremylt: See the source of the tabulation here:
https://petsc.org/release//manualpages/FE/PetscFEGetCellTabulation/

PETSc's tabulation contains separate rows/cols for each component, but libCEED handles that with tensor contractions, so we only need one (WLOG the first) component's rows/columns.

jeremylt: This may need to be revisited if we start using different FE spaces

examples/fluids/src/turb_spanstats.c
```diff
@@ -178,24 +178,27 @@ PetscErrorCode GetQuadratureCoords(Ceed ceed, DM dm, CeedElemRestriction elem_re
...
   PetscCall(PetscNew(stats_data));
 
   PetscCall(DMGetDimension(dm, &dim));
-  CeedBasisGetNumQuadraturePoints1D(ceed_data->basis_q, &Q);
-  CeedBasisGetNumNodes1D(ceed_data->basis_q, &P);
-
-  PetscCall(GetRestrictionForDomain(ceed, dm, 0, 0, 0, 0, Q, problem->q_data_size_sur, &(*stats_data)->elem_restr_parent_stats,
+  CeedBasisGetNumQuadraturePoints1D(ceed_data->basis_q, &num_qpts_child);
+  num_qpts_child *= dim;
+  PetscCall(GetRestrictionForDomain(ceed, dm, 0, 0, 0, 0, num_qpts_child, problem->q_data_size_sur, &(*stats_data)->elem_restr_parent_stats,
```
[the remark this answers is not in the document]
jrwrigh: > This change would make it the same as previous, though I have no idea why we are using the 1D number of quadrature points? That feels funky

It actually wouldn't be the same as previous since `GetRestrictionForDomain` changed to request the _total_ number of quadrature points per element rather than just the number of 1D quadrature points.

The reason I'm using 1D here is that the "parent" basis functions are for 2D elements, but need to match the quadrature rule of the 3D "child" basis functions. It feels funky because it is; we're explicitly taking advantage of the tensor product nature of the quadrature points to do an $L^2$ projection on the 2D mesh based on the spanwise integral.

examples/fluids/src/differential_filter.c
jeremylt: Note - places where you aren't setting up a strided restriction you should probably be calling `CreateRestrictionFromPlex` directly. Then you wouldn't have to worry about this `-1` business

examples/fluids/src/setupdm.c
```diff
@@ -38,17 +39,43 @@ PetscErrorCode CreateDM(MPI_Comm comm, ProblemData *problem, MatType mat_type, V
...
+      PetscCall(DMGetCoordinateDM(dm, &dm_coord));
+      PetscCall(DMGetCoordinateDim(dm, &num_comp_coord));
+      PetscCall(DMGetRegionDS(dm_coord, NULL, NULL, &ds_coord, NULL));
+      PetscCall(PetscDSGetDiscretization(ds_coord, 0, (PetscObject *)&fe_coord_current));
+      PetscCall(PetscFEGetDualSpace(fe_coord_current, &fe_coord_dual_space));
+      PetscCall(PetscDualSpaceGetOrder(fe_coord_dual_space, &fe_coord_order));
+
+      // Create FE for coordinates
+      fe_coord_order = 1;
+      PetscCall(PetscFECreateLagrange(PETSC_COMM_SELF, problem->dim, num_comp_coord, is_simplex, fe_coord_order, q_order, &fe_coord_new));
```
jeremylt, line 74: Why hard-code a lower order coordinates than the mesh might actually have?
Also, will you want to use faces in the future?
```suggestion
      PetscCall(PetscFECreateLagrange(PETSC_COMM_SELF, problem->dim, num_comp_coord, is_simplex, fe_coord_order, q_order, &fe_coord_new));
      PetscCall(PetscFEGetHeightSubspace(fe_coord_new, 1, &fe_coord_face_new));
```

examples/fluids/src/setupdm.c
```diff
@@ -77,6 +104,12 @@ PetscErrorCode SetUpDM(DM dm, ProblemData *problem, PetscInt degree, SimpleBC bc
     }
 
     PetscCall(DMPlexSetClosurePermutationTensor(dm, PETSC_DETERMINE, NULL));
+    if (!is_simplex) {
+      DM dm_coord;
+      PetscCall(DMGetCoordinateDM(dm, &dm_coord));
+      PetscCall(DMPlexSetClosurePermutationTensor(dm_coord, PETSC_DETERMINE, NULL));
+    }
```
jeremylt, line 111:
```suggestion
    if (!is_simplex) {
      DM dm_coord;
      PetscCall(DMGetCoordinateDM(dm, &dm_coord));
      PetscCall(DMPlexSetClosurePermutationTensor(dm, PETSC_DETERMINE, NULL));
      PetscCall(DMPlexSetClosurePermutationTensor(dm_coord, PETSC_DETERMINE, NULL));
    }
```

examples/fluids/src/setupdm.c
jrwrigh: > Why hard-code a lower order coordinates than the mesh might actually have?

Because we haven't tested with higher-order geometry yet I guess? There are other things that will need to change before we get there. Probably better to warn/error out than hard code though.

> Also, will you want to use faces in the future?

Not sure what you mean by that, but we do use faces for boundary integrals. Also not sure what `PetscFEGetHeightSubspace` does without actually doing anything with `fe_coord_face_new`. Does that need an additional call to `DMProjectCoordinates(dm, fe_coord_face_new)`?

jeremylt: That call sets up the FE on the faces here so it ready when you need it.

examples/fluids/src/differential_filter.c
jrwrigh: TBD off of #1267

examples/fluids/src/setuplibceed.c
jrwrigh: I've updated the `quad_mode` name. I'll close this for now.

examples/fluids/src/setuplibceed.c
```diff
@@ -213,33 +210,230 @@ PetscErrorCode SetupBCQFunctions(Ceed ceed, PetscInt dim_sur, PetscInt num_comp_
...
+  PetscCall(PetscFree(grad));
+
+  PetscFunctionReturn(PETSC_SUCCESS);
+}
+
+// -----------------------------------------------------------------------------
+// Get CEED Basis from DMPlex
+// -----------------------------------------------------------------------------
+PetscErrorCode CreateBasisFromPlex(Ceed ceed, DM dm, DMLabel domain_label, CeedInt label_value, CeedInt height, CeedInt dm_field,
+                                   CeedQuadMode quad_mode, CeedBasis *basis) {
```
jeremylt, line 359: This `CeedQuadMode` input won't work as you want when you start using simplex meshes. I don't think I would add this argument, as its saying to explicitly ignore the quadrature space of the FE and use something else

jrwrigh: Yeah, I thought it was used somewhere for `CEED_GAUSS_LOBATTO`, but forgot that all that went away with `CeedBasisCreateProjection`.

jrwrigh: Fixed.

PR #1265, opened 2023-07-25
CeedElemRestriction for H(curl)

sebastiangrimberg:
Reopening https://github.com/CEED/libCEED/pull/1168 from within CEED/libCEED for GitLab CI tests.

From #1168:

High-order H(curl) spaces (p > 1) in 3D involve pairs of element face dofs which must be treated specially in the L-vector to E-vector transformation. These transformations are more complicated than the simple sign flip accounted for in https://github.com/CEED/libCEED/pull/889. In general, the pairs of dofs are transfomed by 2x2 blocks with entries {-1, 0, 1} to write the transformed E-vector dofs as linear combinations of the L-vector dofs. Thus, if each pair of face dofs are arranged to be consecutive in the element dof ordering, we can represent the transformation required as a tridiagonal matrix with entries {-1, 0, 1} applied to the E-vector. Note this is a (more expensive) generalization of the simple orientation flip case where the transformation matrix is just diagonal with entries {-1, 1}. Note that an alternative approach is to apply some global ordering of the nodes of the mesh as a preprocessing step after which a simple signed restriction suffices (like in https://epubs.siam.org/doi/10.1137/08073901X). However this strategy has its own shortcomings in particular when doing things like AMR.

More information can be found in this paper: https://dl.acm.org/doi/pdf/10.1145/3524456 (section 4.1.4, for example). See also https://github.com/mfem/mfem/pull/1046 and https://github.com/mfem/mfem/blob/master/fem/doftrans.hpp for examples of how this transformation is handled in MFEM.

This PR adds CeedElemRestrictionCreateCurlOriented, which is similar to the CeedElemRestrictionCreateOriented constructor introduced in https://github.com/CEED/libCEED/pull/889. Currently only implemented for CPU backends, with the main change coming in CeedElemRestrictionApply_Ref_Core.

Also adds some missing methods which should make blocked and opt backends work for oriented restrictions.


python/ceed_basis.py
```diff
@@ -444,7 +444,7 @@ def apply(self, nelem, emode, u, v):
            Args:
              nelem: the number of elements to apply the basis evaluation to;
                       the backend will specify the ordering in a
-                      Blocked ElemRestriction
+                      BlockedElemRestriction
```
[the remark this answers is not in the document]
sebastiangrimberg: Nice catch! Fixed.

jeremylt (approved):
Overall looks good to me :tada: big feature update!

jedbrown (approved):
Pipeline is clean now and coverage is better. Thanks.

julia/LibCEED.jl/src/generated/libceed_bindings.jl
```diff
@@ -1366,6 +1430,10 @@ end
 
 const CeedInt_FMT = "d"
 
+const CeedSize_FMT = "td"
+
+const CeedInt8_FMT = "d"
```
jeremylt, line 1435:
```suggestion
const CeedInt_FMT = "d"
const CeedSize_FMT = "td"
const CeedInt8_FMT = "d"
```

sebastiangrimberg: This file is automatically generated with `LibCEED.jl/gen/generator.jl`, do you actually want to make manual adjustments?

jeremylt: huh, odd that it vomited those on separate lines. Nevermind

sebastiangrimberg:
> Pipeline is clean now and coverage is better. Thanks.

@jeremylt, and last comments or is this ready to merge from your POV?

jeremylt (approved):

sebastiangrimberg:
Thank you @jedbrown @jeremylt @nbeams @rezgarshakeri for all your help with this feature and PR! Now onto GPU backends for H(div) and H(curl)... 🚀

rezgarshakeri:
> Thank you @jedbrown @jeremylt @nbeams @rezgarshakeri for all your help with this feature and PR! Now onto GPU backends for H(div) and H(curl)... rocket

Thank you for this nice feature.

PR #1266, opened 2023-07-26
fluids: Fix mem leaks in grid_anisotropy_tensor.c

jrwrigh:


jeremylt (approved):
I'm happy if Noether is happy

PR #1274, opened 2023-08-01
Use fallback parent for QFunction assembly if possible

jeremylt:
This should offer a slight performance improvement during CPU assembly.

Issue IDed by @sebastiangrimberg

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -132,7 +132,10 @@ static int CeedOperatorCreateFallback(CeedOperator op) {
   }
   CeedCall(CeedOperatorSetName(op_fallback, op->name));
   CeedCall(CeedOperatorCheckReady(op_fallback));
-  op->op_fallback = op_fallback;
+  // Note: No ref-counting here so we don't get caught in a reference loop.
+  //       The op holds the only reference to op_fallback and is responsible for deleting itself and op_fallback.
+  op->op_fallback                 = op_fallback;
+  op_fallback->op_fallback_parent = op;
 
   return CEED_ERROR_SUCCESS;
 }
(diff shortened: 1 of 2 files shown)


interface/ceed-preconditioning.c
```diff
@@ -1506,6 +1509,9 @@ int CeedOperatorLinearAssembleQFunction(CeedOperator op, CeedVector *assembled,
   if (op->LinearAssembleQFunction) {
     // Backend version
     CeedCall(op->LinearAssembleQFunction(op, assembled, rstr, request));
+  } else if (op->op_fallback_parent && op->op_fallback_parent->LinearAssembleQFunction) {
+    // Backend version for fallback parent
+    CeedCall(CeedOperatorLinearAssembleQFunction(op->op_fallback_parent, assembled, rstr, request));
```
jrwrigh, line 1514: I don't understand this tbh. From `CeedOperatorCreateFallback`, wouldn't `op->op_fallback_parent` be the same as `op`, thus `op->op_fallback_parent->LinearAssembleQFunction == op->LinearAssembleQFunction`?

jeremylt: Different `op`. Here, `op_fallback_parent` would fallback to `op` for diagonal or full assembly, then diagonal or full assembly calls the function to assemble the QFunction, and here we're checking if the `op_fallback_parent` has a faster way to assembly the QFunction. (which should really come before checking if the `op` can assemble

jeremylt: CodeCov showed me that I made the change in the wrong place. Hopefully this latest change helps clarify

jrwrigh: Ah, I was misreading before anyways. Now this makes sense.

jrwrigh (approved):

PR #1275, opened 2023-08-01
fluids: Add PetscCallCeed to capture ceed errors with Petsc

jrwrigh:
Thus when Ceed throws an error, it will be caught by Petsc, which can then provide a nice stacktrace in addition to the error message. 

Naming taken from `PetscCallMPI` (and other associates). Open to discussion on the name though.

To be done after #1264 .

### DIFF
--- a/examples/fluids/navierstokes.c
+++ b/examples/fluids/navierstokes.c
@@ -104,22 +104,24 @@ int main(int argc, char **argv) {
   // ---------------------------------------------------------------------------
   // -- Initialize backend
   Ceed ceed;
-  CeedInit(app_ctx->ceed_resource, &ceed);
+  PetscCheck(CeedInit(app_ctx->ceed_resource, &ceed) == CEED_ERROR_SUCCESS, comm, PETSC_ERR_LIB, "Ceed initialization failed");
   user->ceed = ceed;
+
+  PetscCheck(CeedSetErrorHandler(ceed, CeedErrorStore) == CEED_ERROR_SUCCESS, comm, PETSC_ERR_LIB, "Setting libCEED error handler failed");
 
   // -- Check preferred MemType
   CeedMemType mem_type_backend;
-  CeedGetPreferredMemType(ceed, &mem_type_backend);
+  PetscCallCeed(ceed, CeedGetPreferredMemType(ceed, &mem_type_backend));
 
   {
     const char *resource;
-    CeedGetResource(ceed, &resource);
+    PetscCallCeed(ceed, CeedGetResource(ceed, &resource));
     if (strstr(resource, "/gpu/sycl")) {
       PetscDeviceContext dctx;
       PetscCall(PetscDeviceContextGetCurrentContext(&dctx));
       void *stream_handle;
       PetscCall(PetscDeviceContextGetStreamHandle(dctx, &stream_handle));
-      CeedSetStream(ceed, stream_handle);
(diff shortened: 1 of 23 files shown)


examples/fluids/navierstokes.c
```diff
@@ -107,19 +107,25 @@ int main(int argc, char **argv) {
   CeedInit(app_ctx->ceed_resource, &ceed);
   user->ceed = ceed;
 
+  {  // Set error handler
+    PetscErrorCode ierr;
+    ierr = CeedSetErrorHandler(ceed, CeedErrorStore);
+    if (ierr != CEED_ERROR_SUCCESS) SETERRQ(comm, PETSC_ERR_LIB, "Setting libCEED error handler failed");
+  }
```
jeremylt, line 114: I updated this in Ratel to this approach so we don't need to have an `ierr`
```suggestion
  PetscCheck(CeedSetErrorHandler(ceed, CeedErrorStore) == CEED_ERROR_SUCCESS, comm, PETSC_ERR_LIB, "Setting libCEED error handler failed");
```

examples/fluids/navierstokes.c
```diff
@@ -256,65 +262,65 @@ int main(int argc, char **argv) {
...
+  PetscCallCeed(ceed, CeedOperatorDestroy(&user->op_ifunction_vol));
   PetscCall(OperatorApplyContextDestroy(user->op_rhs_ctx));
-  CeedOperatorDestroy(&user->op_ifunction);
-  CeedOperatorDestroy(&user->op_ijacobian);
+  PetscCallCeed(ceed, CeedOperatorDestroy(&user->op_ifunction));
+  PetscCallCeed(ceed, CeedOperatorDestroy(&user->op_ijacobian));
 
   // -- Ceed
-  CeedDestroy(&ceed);
+  PetscCall(CeedDestroy(&ceed));
```
jeremylt, line 323:
```suggestion
  PetscCheck(CeedDestroy(&ceed) == CEED_ERROR_SUCCESS, comm, PETSC_ERROR_LIB, "Destroying Ceed object failed");
```

examples/fluids/navierstokes.c
jeremylt: Side question - PETSc uses `PETSC_SUCCESS` instead of `PETSC_ERROR_SUCCESS`. I've wondered if we should do the same (totally not an important question)

examples/fluids/navierstokes.c
jeremylt: Probably CeedInit needs this too

examples/fluids/navierstokes.c
jrwrigh: I see both sides. `CEED_ERROR_SUCCESS` keeps the enum prefix consistent for all libCEED return codes. But also having "error" and "success" in the same name is a bit jarring.

I'd lean towards `CEED_SUCCESS`, but don't have a strong preference.

examples/fluids/navierstokes.c
```diff
@@ -107,19 +107,21 @@ int main(int argc, char **argv) {
   CeedInit(app_ctx->ceed_resource, &ceed);
```
jeremylt, line 107: Need to check for a clean return here, then this looks good to merge to me

jrwrigh: Done.

jeremylt (approved):
Looks good. I forget if there's an easy way to set a complier flag to check for unused return values to make sure you got every function call.

PR #1278, opened 2023-08-02
fluids: Fix STG and Blasius for non-box meshes

jrwrigh:
- Allow for blasius problem to not modify the mesh at all via `-platemesh_modify_mesh`
 - Allow streamwise element length to be set explicitly via `-stg_dx`

### DIFF
--- a/examples/fluids/README.md
+++ b/examples/fluids/README.md
@@ -160,6 +160,10 @@ The following options are common among all problem types:
   - Viewer for the force on each no-slip wall, e.g., `ascii:force.csv:ascii_csv` to write a CSV file.
   -
 
+* - `-mesh_transform`
+  - Transform the mesh, usually for an initial box mesh.
+  - `none`
+
 * - `-snes_view`
   - View PETSc `SNES` nonlinear solver configuration
   -
--- a/examples/fluids/README.md
+++ b/examples/fluids/README.md
@@ -963,6 +963,11 @@ The Blasius problem has the following command-line options in addition to the Ne
   - `1.01E5`
   - `Pa`
 
+* - `-platemesh_modify_mesh`
+  - Whether to modify the mesh using the given options below.
+  - `false`
+  -
+
 * - `-platemesh_refine_height`
   - Height at which `-platemesh_Ndelta` number of elements should refined into
   - `5.9E-4`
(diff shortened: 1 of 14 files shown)


KennethEJansen:
I checked out, built, and ran a P1 case with no change to the inputs.  The code did not crash (which I take to mean  that my lack of a request  for '-platemesh_modify_mesh` successfully bypassed the mesh modification).  It ran for 200 steps with the usual variation of the Krylov vectors and residuals.  Finally `-stg_dx` "appeared" as a used parameter so, short of walking through a bugger to be sure it was actually used in the right way, this is as much verification as I can do of this development.  It seems simple enough that this should be enough of a test.

jrwrigh:
Sweet. Thanks!

jedbrown (approved):
- [ ] This reads the other `platemesh_*` options when when they won't be used, thus the user doesn't get unused option warnings. 
- [ ] If we add other transforms, we'd have to handle (and probably error) if people ask for multiple transforms. Perhaps this should be `mesh_transform: platemesh`. When we finish a continuous variant is platemesh, we can move it to petsc, where it can become another transform avaliable for any mesh.

jrwrigh:
> * [ ]  This reads the other `platemesh_*` options when when they won't be used, thus the user doesn't get unused option warnings.

Is achieving that as simple as putting an if statement around the `PetscOptions*` lines for the `platemesh_*` options?
 
>     * [ ]  If we add other transforms, we'd have to handle (and probably error) if people ask for multiple transforms. Perhaps this should be `mesh_transform: platemesh`. When we finish a continuous variant is platemesh, we can move it to petsc, where it can become another transform avaliable for any mesh.

For sure, but I think we can cross that bridge when we get there?

jedbrown:
1. Yeah.
2. It would be better to fix the design now (e.g., using `PetscOptionsEnum`) so we don't have to break/deprecate the code later. Perhaps that's not really important since it'll eventually be a `dm_plex_transform_type`, but changing user input schema is error-prone as people update their scripts or try to run the same configuration on old/new branch (as we've encountered with sycl branches).

jrwrigh:
Not sure what's happening with Noether, but are you ok if I merge this @jedbrown with the modifications?

jedbrown:
CI failed here because this branch hasn't rebased or merged from main to get the CI updates.

jrwrigh:
Well, now RTD is failing for some reason. I can't figure out what's going wrong. There are no errors in the logs, so I'm assuming something broke on RTD's side of things?

Regardless, I can't merge the PR with that failure, so feel free to merge this in.

jeremylt:
https://gitlab.com/micromorph/ratel/-/merge_requests/585


Is this needed here ?

jrwrigh:
I don't think so. Looks like other RTD builds have started passing again, so I've pushed up a new commit to see if it now works.

jeremylt:
Cool. I should check if Ratel still needs that when I get back

Issue #1163, opened 2023-02-21
Parallelism for CPU backends


sebastiangrimberg:
Please forgive me if this is a silly question, but I am wondering how to best use CPU backends in a multicore setting.

Is the recommended approach to use MPI parallelism over cores with a vectorized (AVX or LIBXSMM) backend? Unless I'm missing something, it seems like there is an opportunity here for shared-memory threads + AVX or LIBXSMM so you don't need so many independent MPI processes. It looks like OCCA provides this with it's OpenMP kernels though I'm not sure if each thread is explicitly vectorized like some of the specific vectorized kernels. I'm not super familiar with OCCA and have never used it but is this the way to go if you don't wan to create many MPI processes per node?

jeremylt:
libCEED is designed to work inside of a single MPI process; the parent code is responsible for MPI parallelism. The CPU backends handle parallelism inside of a single element or across batches of elements. The AVX and libXSMM backends provide this functionality.

See here for the logical model for decomposition: https://libceed.org/en/latest/libCEEDapi/#finite-element-operator-decomposition

Note: The OCCA backend is experimental and has very poor performance. I recommend against using it at this time.

jedbrown:
libCEED avoids globals so you can use it from threads, where each thread has its own `Ceed`. We haven't made an OpenMP backend (lack of demand), but I think it would not be hard to make one for CPU by sprinkling pragmas around the critical loops. The most subtle choice will be accumulating from the transpose restriction into the L-vector (where there can be contention between threads, so you'd need to use atomics or other options for some points).

sebastiangrimberg:
Thanks for the fast responses and the note about OCCA, this all makes sense and I definitely understand that libCEED operates in a single MPI process.

Maybe to better explain my motivation, I am thinking there are cases where users of libCEED want the library to exploit parallelism across cores of a CPU and wondering how to make that happen. For example, maybe someone has an application which doesn't support MPI and they want to maximize single-node performance. Alternatively they might have some serial bottlenecks across MPI processes and thus minimizing the number of independent processes is important, so we want to exploit shared-memory parallelism wherever possible.

My thought was that generally most of the work exploiting the parallelism for GPUs could be useful for CPUs on the same node for users who fall into the above situations. I see that the application could just be left to manage threads and create multiple `Ceed`s, but I wondered if the blocking and management of threads in GPU backends could be reused somehow to make this easier on users.

jedbrown:
[...]

PR #1279, opened 2023-08-03
Support CPU shared-memory parallelism with OpenMP

sebastiangrimberg:
This PR adds two (relatively small) code changes in order to support application codes which want to parallelize `Ceed` objects across threads with OpenMP (each thread has its own `Ceed`):
- Add `#pragma omp threadprivate` statements around some file-scope static variables which by default are shared across threads even if declared/initialized in parallel regions. This resolves issues in particular with calling `CeedInit` and `CeedDestroy` across threads in parallel. I added the minimum amount of statements to silence all warnings appearing with `-fsanitize=thread`.
- Even if each thread applies its own thread-local operator to a thread-local vector, there are cases where the memory associated with the vector is shared (if that makes sense, for example `CeedVectorSetArray` is used to point to a shared memory location). Thus, the transpose restriction from E-vector to L-vector needs to be atomic, so I added the appropriate `#pragma omp atomic update` for the CPU backends.

Support for builds with OpenMP build is added to the `Makefile` and activated with the `OPENMP` variable (off by default). This will set up the correct compiler and linker flags as well as define `CEED_USE_OPENMP` for the build. When `OPENMP` is not specified, there should be zero change from the existing code behavior by the changes in this PR.

See #1163 for some further discussion of the motivation behind this feature.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,13 +107,18 @@ MARCHFLAG.oneAPI        := $(MARCHFLAG.clang)
 OMP_SIMD_FLAG.gcc       := -fopenmp-simd
 OMP_SIMD_FLAG.clang     := $(OMP_SIMD_FLAG.gcc)
 OMP_SIMD_FLAG.icc       := -qopenmp-simd
-OMP_SIMD_FLAG.oneAPI    := $(OMP_SIMD_FLAG.clang)
+OMP_SIMD_FLAG.oneAPI    := $(OMP_SIMD_FLAG.icc)
+OMP_FLAG.gcc            := -fopenmp
+OMP_FLAG.clang          := $(OMP_FLAG.gcc)
+OMP_FLAG.icc            := -qopenmp
+OMP_FLAG.oneAPI         := $(OMP_FLAG.icc)
 SYCL_FLAG.gcc           :=
 SYCL_FLAG.clang         := -fsycl
 SYCL_FLAG.icc           :=
 SYCL_FLAG.oneAPI        := -fsycl -fno-sycl-id-queries-fit-in-int
 OPT.gcc                 := -g -ffp-contract=fast
 OPT.clang               := $(OPT.gcc)
+OPT.icc                 := $(OPT.gcc)
 OPT.oneAPI              := $(OPT.clang)
 OPT.emcc                :=
 CFLAGS.gcc              := $(if $(STATIC),,-fPIC) -std=c99 -Wall -Wextra -Wno-unused-parameter -MMD -MP
@@ -151,6 +156,7 @@ PEDANTICFLAGS ?= -Werror -pedantic
 OPT    ?= -O $(MARCHFLAG) $(OPT.$(CC_VENDOR)) $(OMP_SIMD_FLAG)
 CFLAGS ?= $(OPT) $(CFLAGS.$(CC_VENDOR)) $(if $(PEDANTIC),$(PEDANTICFLAGS))
 CXXFLAGS ?= $(OPT) $(CXXFLAGS.$(CC_VENDOR)) $(if $(PEDANTIC),$(PEDANTICFLAGS))
+FFLAGS ?= $(OPT) $(FFLAGS.$(FC_VENDOR))
 LIBCXX ?= -lstdc++
 NVCCFLAGS ?= -ccbin $(CXX) -Xcompiler "$(OPT)" -Xcompiler -fPIC
 ifneq ($(CUDA_ARCH),)
@@ -160,10 +166,16 @@ HIPCCFLAGS ?= $(filter-out $(OMP_SIMD_FLAG),$(OPT)) -fPIC -munsafe-fp-atomics
 ifneq ($(HIP_ARCH),)
   HIPCCFLAGS += --amdgpu-target=$(HIP_ARCH)
 endif
-
 SYCL_FLAG := $(SYCL_FLAG.$(CC_VENDOR))
 SYCLFLAGS ?= $(SYCL_FLAG) -fPIC -std=c++17 $(filter-out -std=c++11,$(CXXFLAGS)) $(filter-out $(OMP_SIMD_FLAG),$(OPT))
-FFLAGS ?= $(OPT) $(FFLAGS.$(FC_VENDOR))
+
+OPENMP ?=
+ifneq ($(OPENMP),)
+	OMP_FLAG := $(OMP_FLAG.$(CC_VENDOR))
+	OMP_FLAG := $(if $(call cc_check_flag,$(OMP_FLAG)),$(OMP_FLAG))
+  CFLAGS += $(OMP_FLAG) -DCEED_USE_OPENMP
+  CEED_LDFLAGS += $(OMP_FLAG)
+endif
 
 ifeq ($(COVERAGE), 1)
   CFLAGS += --coverage
--- a/Makefile
+++ b/Makefile
@@ -171,8 +171,8 @@ SYCLFLAGS ?= $(SYCL_FLAG) -fPIC -std=c++17 $(filter-out -std=c++11,$(CXXFLAGS)) 
 
 OPENMP ?=
 ifneq ($(OPENMP),)
-	OMP_FLAG := $(OMP_FLAG.$(CC_VENDOR))
-	OMP_FLAG := $(if $(call cc_check_flag,$(OMP_FLAG)),$(OMP_FLAG))
+  OMP_FLAG := $(OMP_FLAG.$(CC_VENDOR))
+  OMP_FLAG := $(if $(call cc_check_flag,$(OMP_FLAG)),$(OMP_FLAG))
   CFLAGS += $(OMP_FLAG) -DCEED_USE_OPENMP
   CEED_LDFLAGS += $(OMP_FLAG)
 endif
@@ -430,8 +430,8 @@ endif
 
 # CUDA Backends
 ifneq ($(CUDA_DIR),)
-	CUDA_LIB_DIR := $(wildcard $(foreach d,lib lib64 lib/x86_64-linux-gnu,$(CUDA_DIR)/$d/libcudart.${SO_EXT}))
-	CUDA_LIB_DIR := $(patsubst %/,%,$(dir $(firstword $(CUDA_LIB_DIR))))
+  CUDA_LIB_DIR := $(wildcard $(foreach d,lib lib64 lib/x86_64-linux-gnu,$(CUDA_DIR)/$d/libcudart.${SO_EXT}))
+  CUDA_LIB_DIR := $(patsubst %/,%,$(dir $(firstword $(CUDA_LIB_DIR))))
 endif
 CUDA_LIB_DIR_STUBS := $(CUDA_LIB_DIR)/stubs
 CUDA_BACKENDS = /gpu/cuda/ref /gpu/cuda/shared /gpu/cuda/gen
@@ -452,7 +452,7 @@ HIP_LIB_DIR := $(wildcard $(foreach d,lib lib64,$(ROCM_DIR)/$d/libamdhip64.${SO_
 HIP_LIB_DIR := $(patsubst %/,%,$(dir $(firstword $(HIP_LIB_DIR))))
 HIP_BACKENDS = /gpu/hip/ref /gpu/hip/shared /gpu/hip/gen
 ifneq ($(HIP_LIB_DIR),)
-	HIPCONFIG_CPPFLAGS := $(subst =,,$(shell $(ROCM_DIR)/bin/hipconfig -C))
+  HIPCONFIG_CPPFLAGS := $(subst =,,$(shell $(ROCM_DIR)/bin/hipconfig -C))
   $(hip-all.c:%.c=$(OBJDIR)/%.o) $(hip-all.c:%=%.tidy): CPPFLAGS += $(HIPCONFIG_CPPFLAGS)
   ifneq ($(CXX), $(HIPCC))
     $(hip-all.cpp:%.cpp=$(OBJDIR)/%.o) $(hip-all.cpp:%=%.tidy): CPPFLAGS += $(HIPCONFIG_CPPFLAGS)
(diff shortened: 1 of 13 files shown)


jeremylt:
Looks good to me, though I don't have OpenMP experience.

Did you do a quick check that there was no obvious performance change with the minor loop restructuring?

sebastiangrimberg:
Just tested on `main` and I see no change. This should be expected, I can't imagine a compiler not being able to optimize out the temporary `CeedScalar` in the `ElemRestrictionApplyTranspose` loop.

@jedbrown does this OpenMP usage look OK to you?

jeremylt:
Yeah, I suspected but thanks for double checking

include/ceed/backend.h
```diff
@@ -52,12 +52,27 @@
...
 #else
 #define CeedPragmaOptimizeOn
 #endif
 #endif
 
+/// This macro provides the appropriate OpenMP Pragmas for the compilation environment.
+/// @ingroup Ceed
+#ifdef CEED_USE_OPENMP
+#ifndef CeedPragmaAtomic
+#define CeedPragmaAtomic _Pragma("omp atomic update")
```
jedbrown, line 65: I think it would be nice to stringize and have everything OMP under `CeedPragmaOMP`. I think the logic below can be simplified by making `CeedPragmaOMP(...)` evaluate to nothing when not compiled with `_OPENMP`.
```suggestion
#define CeedPragmaAtomic CeedPragmaOMP(atomic update)
```

Makefile
```diff
@@ -160,10 +166,16 @@ HIPCCFLAGS ?= $(filter-out $(OMP_SIMD_FLAG),$(OPT)) -fPIC -munsafe-fp-atomics
...
-
 SYCL_FLAG := $(SYCL_FLAG.$(CC_VENDOR))
 SYCLFLAGS ?= $(SYCL_FLAG) -fPIC -std=c++17 $(filter-out -std=c++11,$(CXXFLAGS)) $(filter-out $(OMP_SIMD_FLAG),$(OPT))
-FFLAGS ?= $(OPT) $(FFLAGS.$(FC_VENDOR))
+
+OPENMP ?=
+ifneq ($(OPENMP),)
+  OMP_FLAG := $(OMP_FLAG.$(CC_VENDOR))
+  OMP_FLAG := $(if $(call cc_check_flag,$(OMP_FLAG)),$(OMP_FLAG))
+  CFLAGS += $(OMP_FLAG) -DCEED_USE_OPENMP
```
jedbrown, line 176: This is a preprocessor flag and thus belongs in `CPPFLAGS`, but I'm not sure it's needed. Why not use `_OPENMP` as specified by the standard? Is the concern that the compiler may activate it by default, but we may not want to use atomics unless the user really wants to?

Makefile
```diff
@@ -107,13 +107,18 @@ MARCHFLAG.oneAPI        := $(MARCHFLAG.clang)
 OMP_SIMD_FLAG.gcc       := -fopenmp-simd
 OMP_SIMD_FLAG.clang     := $(OMP_SIMD_FLAG.gcc)
 OMP_SIMD_FLAG.icc       := -qopenmp-simd
-OMP_SIMD_FLAG.oneAPI    := $(OMP_SIMD_FLAG.clang)
+OMP_SIMD_FLAG.oneAPI    := $(OMP_SIMD_FLAG.icc)
```
jedbrown, line 110: `icx` supports both.
```
  -fopenmp                Parse OpenMP pragmas and generate parallel code. Advanced users who prefer to use OpenMP* as it is implemented by the LLVM community can get most of that functionality by using -fopenmp.
```
and
```
  -qopenmp                Parse OpenMP pragmas and generate parallel code.
```

 Is there a reason to prefer one over the other?

jedbrown (commented):
Do we need `PREDEFINED` entries for new macros in `Doxyfile`?

Makefile
sebastiangrimberg: Whoops, of course it's better to just use `_OPENMP`. I wasn't aware this was actually in the standard and consistent across implementations, thanks!

include/ceed/backend.h
sebastiangrimberg: Nice, done in [d148e0c](https://github.com/CEED/libCEED/pull/1279/commits/d148e0c9a35736f3a15bcfb5399ab207a92b12e4).

sebastiangrimberg:
> Do we need `PREDEFINED` entries for new macros in `Doxyfile`?

Fixed in [d148e0c](https://github.com/CEED/libCEED/pull/1279/commits/d148e0c9a35736f3a15bcfb5399ab207a92b12e4).

I think I've addressed all the comments. Let me know if anything else still stands out as needing changes!

Makefile
jedbrown: Fair enough, let's keep as is.

jedbrown (approved):
LGTM, thanks!

jedbrown (changes requested):
Should we enable OpenMP in one of the test environments so it at least gets built with those compilation flags?

We should also document what this feature means (which I think amounts to being allowed to apply different operators from different threads inside an `omp parallel`).

sebastiangrimberg:
> Should we enable OpenMP in one of the test environments so it at least gets built with those compilation flags?

Added to the Intel build in [6877d6f](https://github.com/CEED/libCEED/pull/1279/commits/6877d6f31d80bd80a9ace343a42b18ae0d1774d8).

> We should also document what this feature means (which I think amounts to being allowed to apply different operators from different threads inside an `omp parallel`).

Sure, where do you think this should go in the documentation? Perhaps at the end of https://libceed.org/en/latest/gettingstarted/#building I can add a one-sentence note, just before "WebAssembly"? And yes that's exactly what it does, with the added option that the active vector can be shared across threads.

jedbrown:
Yeah, that sounds good, and mention it in release notes.

sebastiangrimberg:
Done! [3802174](https://github.com/CEED/libCEED/pull/1279/commits/380217418e02fdfca4d994a4a0d25b990ae76ebf)

interface/ceed.c
```diff
@@ -26,7 +26,9 @@ static struct {
   int (*init)(const char *resource, Ceed f);
   unsigned int priority;
 } backends[32];
+CeedPragmaThreadPrivate(backends)
 static size_t num_backends;
+CeedPragmaThreadPrivate(num_backends)
```
jedbrown, line 31: Hmm, does this mean every thread needs to call `CeedInit`? That feels inflexible compared to being able to create multiple `Ceed` objects and access them from threads later. What if we used a mutex/critical section for registration so any threads can call it, the run-time or app might choose to join those threads and create new ones, and each thread would still be able to safely use any number of `Ceed`s?

Among other things, one thread could call `CeedInit` a few times for different resources and deal them out to OpenMP threads dynamically.

sebastiangrimberg: Yeah that was the idea: Each thread operates on its own `Ceed` object for its full lifetime. This is basically the idea laid out here as I understood it: https://github.com/CEED/libCEED/issues/1163#issuecomment-1438955698.

The reason for ending up with the current one thread--one `Ceed` design after kind of attempting to support the idea you state above was because it seemed to get very messy very quickly to ensure thread safety with multiple threads accessing a single `Ceed`. It seemed like `omp critical`s would start to pop up everywhere or else we would just rely on the user to not create race conditions (reference counting on a `Ceed` object, for example). So, enforcing that each thread owns its own `Ceed` seems safest to me, with a minimal amount of code changes and a fully clean `-fsanitize=thread` report. I don't think it hurts the user too badly either, since this is meant to be an alternative to MPI parallelism where you have a hard wall between `Ceed`s owned by different MPI processes. This PR just allows those `Ceed`s to all live in the same address space.

The use case I have in mind and have implemented is for the app to manage a `threadprivate` `Ceed` as:

```
Ceed ceed;
#pragma omp threadprivate(ceed)

...

void f() {
  #pragma omp parallel
  {
    CeedInit(ceed_spec, &ceed);
  }
}
```

In the example you describe, where one thread calls `CeedInit` and doles them out later on to each thread, I think that one thread will still need to call `CeedInit` once for each thread which may be spawned later on because of the issues associated with threads sharing a `Ceed` object above. In this case I don't believe the above code proposes much of a limitation.

Perhaps it's easier to discuss this over a call to work out the best solution, I am happy to adjust the implementation but as is I'm not sure it would be so easy to change and ensure thread safety.

jedbrown: I can see how it feels like it's getting messy. I'm not sure it's that bad.

Per the OpenMP-5 standard, it's very delicate whether both `omp parallel` blocks would have the same thread private variables.

```c
Ceed ceed;
#pragma omp threadprivate(ceed)

#pragma omp parallel
{
  CeedInit(ceed_spec, &ceed);
}
// do some serial stuff
#pragma omp parallel
{
  CeedVectorCreate(ceed, ...);
}
```

image

This includes implementation-defined behavior if the user has not exported `OMP_DYNAMIC=false`, which just feels too fragile. I think there are a couple options:

1. We could do registration into a local variable an `CeedInit` and free that list. This should eliminate global variables entirely.
2. We can have a mutex around registration (like a `once_cell`)

In either case, I think the rule is that only one thread can call methods on any given `Ceed` or object at the same time. You can "give" a `Ceed` (or other object) to a different thread, but the caller is responsible for the memory ordering necessary to ensure the last task on the donor thread completes before the recipient thread uses it.

sebastiangrimberg: To make sure I understand, is the vision that the application code might configure an array of `Ceed` objects with no private data in any of them such that any one can be accessed by any worker thread? Then a parallel region could look like:

```
#pragma omp parallel
{
  int i = omp_get_thread_num();
  CeedVectorCreate(ceed[i], ...);
}
```

jedbrown: Yeah, or even a higher level library (above libCEED) wants to write an interface `TakeOneStep(...)` that is meant to be called outside a parallel region. So it'll have an `omp parallel` region inside and that should be reliable and not dependent on globals.

sebastiangrimberg: OK, cool that makes sense. Just in the interest of saving time, do you have a vision of how you'd make backend and Q-function registration (the two places `#pragma omp threadprivate` is currently required) into local varaibles only? These global variables (the `backends` and `gallery_qfunctions` arrays, along with `register_all_called` variables for each), are shared across various initialization functions. Where would they be kept if local?

This seems like a big change in the handling of initialization and the registration "state". Alternatively the approach of keeping the globals but adding a mutex would just involve adding a critical around use of `register_all_called` in `ceed-register.c` and `ceed-qfunction-register.c`, in addition to `critial`s or a mutex around the global `backends` and `gallery_qfunctions` variables when they are being modified. Maybe the second approach actually does sound simpler, as you were saying?

sebastiangrimberg:
Alright! I took a stab at revamping the approach to be more flexible in bb28596. No more `omp threadprivate`, using instead `omp critical` around the global variable access points when there is potential for race conditions during registration. This results in a simpler set of changes, so I think this is nice. I tested it out and it seems to work, though the implementation I'm testing from is still creating one `Ceed` per thread and so I haven't tested the more general cases which now should be supported.

Let me know what you think, and if this is a better approach in your view. Thanks a lot for the help!

interface/ceed-register.c
```diff
@@ -28,11 +28,15 @@ static bool register_all_called;
...
+    if (!register_all_called) {
+      CeedDebugEnv256(1, "\n---------- Registering Backends ----------\n");
+#define CEED_BACKEND(name, ...) ierr += name();
 #include "../backends/ceed-backend-list.h"
 #undef CEED_BACKEND
-  return CEED_ERROR_SUCCESS;
+      register_all_called = true;
+    }
+  }
+  return ierr;
```
jedbrown, line 41: Should we raise an error with a useful message here, not just return a count as though it were an enum value. Or are we not able to do so because error handling hasn't been defined yet?

jedbrown (approved):
Looks good to me now, assuming all tests pass and docs look good (I haven't checked that yet).

interface/ceed-register.c
sebastiangrimberg: I just kept the behavior as before which used `CeedChk` on the return value (which has the behavior to just return `ierr` if nonzero and not to raise any error). It looks like the `ierr` value is handled by the caller in `CeedInit`, but there too it just uses `CeedCall` so the nonzero error code is returned to user and no errors are raised. I assume this behavior was an intentional choice at some point so would probably keep it for the sake of this PR unless you would rather address that change.

interface/ceed-operator.c
```diff
@@ -1493,8 +1493,8 @@ int CeedOperatorApply(CeedOperator op, CeedVector in, CeedVector out, CeedReques
         }
       }
       // Apply
-      for (CeedInt i = 0; i < op->num_suboperators; i++) {
-        CeedCall(CeedOperatorApplyAdd(op->sub_operators[i], in, out, request));
```
jeremylt, line 1497: side note - this probably could have been stripped out as a separate fix. I like merging fast fix PRs like that

interface/ceed-register.c
jedbrown: So we probably should have made the error code strongly typed all along (so each function would return `CeedErrorType` rather than `int`), but the numbers have semantic meaning (`CeedErrorType`). We don't want that loop to do something like calculate `CEED_ERROR_MAJOR + CEED_ERROR_BACKEND == CEED_ERROR_UNSUPPORTED`. It's perhaps a side-show in this scenario, but it's confusing to have this pattern in the code. My preference would be to break early from the critical section so we return after the first error with the correct error code rather than charging ahead (possibly creating many errors) and returning what (the first or last or some mashup).

sebastiangrimberg: How is [5230104](https://github.com/CEED/libCEED/pull/1279/commits/523010420027eb90f18075bec7b03c33137a94b3)? This should do exactly that rather than accumulating into `ierr` and changing the result. As soon as one registration fails, the rest are skipped in the critical section.

interface/ceed-operator.c
sebastiangrimberg: Yeah whoops I'm sorry about this one: See https://github.com/CEED/libCEED/pull/1313. Thanks!

backends/ceed-backend-weak.c
```diff
@@ -19,11 +19,12 @@ static int CeedInit_Weak(const char *resource, Ceed ceed) {
...
 static int CeedRegister_Weak(const char *name, int num_prefixes, ...) {
   va_list prefixes;
+  int     ierr;
+
   va_start(prefixes, num_prefixes);
-  int ierr;
   for (int i = 0; i < num_prefixes; i++) {
     const char *prefix = va_arg(prefixes, const char *);
-    CeedDebugEnv("** Weak Register: %s", prefix);
+    CeedDebugEnv("Weak Register: %s", prefix);
```
jeremylt, line 27: This is here so when you run with `CEED_DEBUG=1` environment variable the list of registered backends is aligned and the weakly registered backends (the ones not actually built) stand out.

```suggestion
    CeedDebugEnv("** Weak Register: %s", prefix);
```

Could also do this I suppose
```suggestion
    CeedDebugEnv("Weak Register   : %s", prefix);
```

Or something else to add 3 characters before the name of the backend

interface/ceed.c
```diff
@@ -301,7 +309,6 @@ int CeedFree(void *p) {
   @ref Backend
 **/
 int CeedRegister(const char *prefix, int (*init)(const char *, Ceed), unsigned int priority) {
-  CeedDebugEnv("Backend Register: %s", prefix);
```
jeremylt, line 304: Why was this dropped?

sebastiangrimberg: Just moved it into `CeedRegisterImpl` so that the debugging statement would accompany the actual implementation. But I see now that the idea was to differentiate "Backend Register" and "Weak Register" print statements now, so I will revert it if that is OK with you?

jeremylt: Thanks!

backends/ceed-backend-weak.c
sebastiangrimberg: Fixed as well in [695c241](https://github.com/CEED/libCEED/pull/1279/commits/695c241f98c558443702e1226c36e4fb82ee64da).

jeremylt:
Ready to merge?

sebastiangrimberg:
> Ready to merge?

Ah! Sorry didn't realize you were waiting on me for this. Yes ready to go from my point of view, as long as all of @jedbrown's final comments are resolved (which looking at them I think they are). So feel free to merge!

PR #1280, opened 2023-08-03
Update CeedCall macro

jeremylt:
Updating the `CeedCall` macro to simpler logic revealed a few `make tidy` issues, which are also included in this PR.


jrwrigh (approved):
Looks good to me. I'm guessing it wouldn't be a bad idea to also merge `PetscCeedChk` into `PetscCallCeed`?

jeremylt:
Yeah, I ended up removing the equivalent of PetscCeedChk in Ratel as unneeded as part of a similar update.

PR #1281, opened 2023-08-03
Clarify op fallback parent interface

jeremylt:
This is a small update that makes the Operator fallback a bit clearer

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -201,7 +201,7 @@ int CeedOperatorGetFallbackParent(CeedOperator op, CeedOperator *parent) {
   @ref Developer
 **/
 int CeedOperatorGetFallbackParentCeed(CeedOperator op, Ceed *parent) {
-  *parent = op->op_fallback_parent ? op->op_fallback_parent->ceed : NULL;
+  *parent = op->op_fallback_parent ? op->op_fallback_parent->ceed : op->ceed;
   return CEED_ERROR_SUCCESS;
 }
 
@@ -2333,7 +2333,6 @@ int CeedOperatorCreateFDMElementInverse(CeedOperator op, CeedOperator *fdm_inv, 
   // Default interface implementation
   CeedCall(CeedOperatorGetCeed(op, &ceed));
   CeedCall(CeedOperatorGetFallbackParentCeed(op, &ceed_parent));
-  ceed_parent = ceed_parent ? ceed_parent : ceed;
   CeedCall(CeedOperatorGetQFunction(op, &qf));
 
   // Determine active input basis
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -176,7 +176,7 @@ int CeedOperatorGetFallback(CeedOperator op, CeedOperator *op_fallback) {
 }
 
 /**
-  @brief Get the parent Ceed context associated with a fallback Ceed context for a CeedOperator
+  @brief Get the parent CeedOperator for a fallback CeedOperator
 
   @param[in]  op     CeedOperator context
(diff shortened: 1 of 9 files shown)


interface/ceed-preconditioning.c
```diff
@@ -175,6 +175,36 @@ int CeedOperatorGetFallback(CeedOperator op, CeedOperator *op_fallback) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Get the parent Ceed context associated with a fallback Ceed context for a CeedOperator
```
jrwrigh, line 179:
```suggestion
  @brief Get the parent CeedOperator associated with a fallback Ceed context for a CeedOperator
```

Not sure if this is the exact correct wording, but it does return an operator rather than the context.

jrwrigh: Also, `interface/ceed-preconditioning.c` seems like an odd location for these, but I guess they're predominantly used here, so it's probably fine?

jrwrigh (approved):

interface/ceed-preconditioning.c
jeremylt: Good catch on the doc issue.

The fallback mechanism is only used for partial inheritance of preconditioning functionality for operators

jeremylt: The whole description was bad, so hopefully this edit is better (pushed)

jrwrigh: Yep, those makes sense to me!

sebastiangrimberg:
Hi @jeremylt, sorry for the annoying question here but I'm wondering, what's the need for the `  CeedCallBackend(CeedOperatorGetFallbackParentCeed(op, &ceed_parent));` for the `cuda`, `hip`, and `sycl` backends? It seems like if we've made it into `CeedOperatorLinearAssembleQFunctionCore` for the given GPU backend, we don't need any access to the parent anymore. But I'm probably wrong, so hopefully there's an explanation. Thanks!

jeremylt:
It allows uneven development of new features. One of them I want to work on eventually for CPUs is a per-element assembly of the QFunction instead of all at oncee

sebastiangrimberg:
Hm, I'm not sure I see the use case still so maybe you can help me understand what you have in mind. Right now I see the use case as when a resource falls back to some base backend's implementation of `LinearAssembleQFunction` or `LinearAssembleQFunctionUpdate`, but needs the actual parent backend to for `CeedVectorCreate` and `CeedElemRestrictionCreateStrided`. Currently I think this is a no-op for all backends. Are you suggesting that in the future, we may have backends where we still want to use a base implementation of `LinearAssembleQFunction` but the element restriction or vector construction is specialized by the parent?

jeremylt:
In this new case, the fallback operator's assembly would be calling this more efficient single element QFunction assembly, which may only be defined on its parent

sebastiangrimberg:
Totally, but doesn't that promotion to the parent QFunction assembly call happen inside of `ceed-preconditioning.c`? That is, `CeedOperatorLinearAssembleQFunctionBuildOrUpdate` will promote the operator to the highest parent which defines the QFunction assembly which is going to be the most specialized. But the code inside of each backend shouldn't need to worry about promoting up to other parents, I would think.

jeremylt:
Backends used to need this functionality. I'd rather not remove an internal capability just because it isn't currently being used.

sebastiangrimberg:
Ah, OK I suppose that's fair. I just thought it would be an opportunity for clarification as I ran into this while working on something related earlier and it confused me.

jeremylt:
I definitely think the interface and internals can be clarified. I am just reticent to remove functionality while doing so.

PR #1282, opened 2023-08-07
fluids: Reset dim before reference_velocity option

jrwrigh:
If `-g` is not given, then
```
  PetscCall(PetscOptionsRealArray("-g", "Gravitational acceleration", NULL, g, &dim, NULL));
```
will set `dim` to zero. This then causes
```
  PetscCall(PetscOptionsScalarArray("-reference_velocity", "Reference/initial velocity", NULL, reference.velocity, &dim, NULL));
```
to not read any values.

Solution is just to reset `dim` to 3 (since `reference.velocity` is always of size 3 rather than size `problem->dim`).


jeremylt (approved):

jedbrown (commented):
Hmm, seems to me we should require that `-g` if given is a vector of full length. I also think it should be called `-gravity`, and I would not object to turning it off by default since we've been doing engineering flows lately and those conventionally don't turn on gravity.

PR #1284, opened 2023-08-09
DMSwarm Example with Interpolation and Projection

jeremylt:
This basic example will have 2 main features

- [x] `DMSwarm[Get,Restore]ReferenceCoordinates` prototype
- [x] Comparing PETSc and libCEED interp at swarm points
- [x] Use new `ElemRestrictionApplyAtPointsInElement` 
- [x] Projection from swarm to mesh
- [ ] New CI build for updated PETSc


jeremylt:
This will probably take a little longer than anticipated - this is slow going.

I think the basic logic is now in here though. Its ugly code that needs to be tidied though.

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,410 @@
...
+    PetscReal *coords_cell_points_true, *coords_cell_points_ref;
+    PetscInt  *cell_points;
+    PetscInt   num_points_in_cell;
+
+    PetscCall(DMSwarmSortGetPointsPerCell(dm_swarm, cell, &num_points_in_cell, &cell_points));
+    PetscCall(DMGetWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_cell_points_true));
+    PetscCall(DMGetWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_cell_points_ref));
+    // -- Reference coordinates for swarm points in background mesh element
+    for (PetscInt p = 0; p < num_points_in_cell; p++) {
+      point_cell_numbers[cell_points[p]] = cell;
```
jeremylt, line 92: Maybe this? I'm still deciding exactly what to put in this index set.
```suggestion
      point_cell_numbers[cell_points[p]] = cell - cell_start;
```

Related - we should decide what sort of restriction operation we need to create for this and how to update it. Of note for this restriction is the fact that a) we have different numbers of swarm points per cell 2) we need to update the mapping or create/destroy this restriction once per time step

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,410 @@
...
+  PetscScalar result = 1, center = 0.1;
+  for (PetscInt d = 0; d < dim; d++) {
+    if (d == direction) result *= 1.0 / cosh(x[d] - center) / cosh(x[d] - center);
+    else result *= tanh(x[d] - center);
+    center += 0.1;
+  }
+  return result;
+}
+
+PetscErrorCode DMSwarmCreateReferenceCoordinates(DM dm_swarm, DM dm_mesh, IS *is_points, Vec *ref_coords) {
```
jeremylt, line 55: Note - I put `dm_mesh` as an argument here because there is no `DMSwarmGetCellDM`. This is fine since this will be an internal only method, but I think we should *also* add `DMSwarmGetCellDM` since that will be useful for us.

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,410 @@
...
+    PetscCall(PetscFree(cell_points));
+  }
+
+  // Cleanup
+  PetscCall(DMSwarmRestoreField(dm_swarm, "DMSwarmPIC_coor", NULL, NULL, (void **)&coords_points_ref));
+  PetscCall(VecRestoreArrayRead(*ref_coords, &coords_points_true));
+  PetscCall(DMSwarmSortRestoreAccess(dm_swarm));
+
+  // Create index set
+  PetscCall(ISCreateGeneral(PETSC_COMM_SELF, num_points_local, point_cell_numbers, PETSC_OWN_POINTER, is_points));
```
jeremylt, line 112: It is my understanding that using `PETSC_COMM_SELF` here means this will be local to the process?

examples/petsc/dmswarm.c
jeremylt, line 55: ToDo - add PETSc style documentation header

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,410 @@
...
+    PetscCall(PetscFEDestroy(&fe_coord));
+  }
+
+  // -- Final background mesh
+  PetscCall(PetscObjectSetName((PetscObject)dm_mesh, "Background Mesh"));
+  PetscCall(DMViewFromOptions(dm_mesh, NULL, "-dm_mesh_view"));
+
+  // -- libCEED objects from background mesh
+  PetscCall(CreateBasisFromPlex(ceed, dm_mesh, NULL, 0, 0, 0, /* Don't want BP data here */ bp_options[CEED_BP1], &basis_mesh_u));
+  PetscCall(CreateRestrictionFromPlex(ceed, dm_mesh, 0, NULL, 0, &restriction_mesh_u));
```
jeremylt, line 162: These functions aren't linking correctly. Also, I don't really want to have to include the BP data just for the purpose of including it here. `CreateBasisFromPlex` should be updated to accept a `NULL` BP data and use standard quadrature in that case.

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,410 @@
...
+      CeedVectorTakeArray(u_lvec, CEED_MEM_HOST, (CeedScalar **)&u_array);
+      PetscCall(VecRestoreArrayRead(U_loc, &u_array));
+      CeedVectorDestroy(&u_lvec);
+    }
+    {
+      CeedInt P;
+
+      CeedBasisGetNumNodes(basis_mesh_u, &P);
+      CeedVectorCreate(ceed, P, &u_cell);
+    }
```
jeremylt, line 318: We do this internally in the CPU backends a lot. It might be nice to have a utility to make a single element evec or a 8 element batch evec?

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,410 @@
...
+        PetscInt    p = 0;
+        CeedScalar *x;
+
+        CeedVectorGetArrayWrite(x_points, CEED_MEM_HOST, &x);
+        for (PetscInt i = 0; i < num_points_local; i++) {
+          if (all_points[i] == cell) {
+            for (PetscInt d = 0; d < dim; d++) x[p * dim + d] = coords_points_ref[i * dim + d];
+            p++;
+          }
+        }
```
jeremylt, line 350: Need to think more here - I use this pattern since I'm trying to only use the data available from the call to `DMSwarmCreateReferenceCoordinates` (the new function we're creating in this MR)

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,412 @@
...
+  // Interpolate from mesh to points via libCEED
+  {
+    IS                 is_points;
+    Vec                U_loc, ref_coords;
+    PetscInt           num_points_local;
+    const PetscInt    *all_points;
+    PetscInt           cell_start, cell_end;
+    PetscScalar       *u_all_points_array;
+    const PetscScalar *coords_points_ref;
+    CeedVector         u_evec, u_cell;
```
jeremylt, line 296: Variable naming is a bit of a thick soup right now. I'm iterating as I go to balance clarity, accuracy, and brevity.

examples/petsc/dmswarm.c
jedbrown: For the particle restriction, two options come to mind:

1. Sort the particles by cell and provide the offsets `CeedInt particle_offset[num_cells + 1]`. These are similar to row starts in a CSR matrix format. The particles for each cell are contiguous in memory.
2. Create an integer permutation array `particle_index[num_particles]` so that the particles themselves don't need to be sorted. This requires less up-front work, but then a less regular memory access pattern.

I think either is fine, though I don't know if `DMSwarm` supports that intrusive sort (which has to also reorder every field). The second option is more flexible, especially if different operators need to work over different parts of the domain (so that the particles they need to work cannot be contiguous for each operator). That makes me lean toward 2.

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,412 @@
...
+    PetscCall(DMAddField(dm_mesh, NULL, (PetscObject)fe));
+    PetscCall(PetscFEDestroy(&fe));
+  }
+  PetscCall(DMCreateDS(dm_mesh));
+
+  // -- Coordinate FE space
+  {
+    PetscFE fe_coord;
+
+    PetscCall(PetscFECreateLagrange(comm, dim, num_comp, PETSC_FALSE, geometry_order, q_order, &fe_coord));
```
zatkins-dev, line 150: Small note, I think this should be:
```suggestion
    PetscCall(PetscFECreateLagrange(comm, dim, dim, PETSC_FALSE, geometry_order, q_order, &fe_coord));
```

zatkins-dev:
I think I found a bug in `CeedBasisApplyAtPoints` while trying to get this to run: #1287

zatkins-dev:
I think I found a bug in `CeedBasisApplyAtPoints` while trying to get this to run: #1287

jeremylt:
Example links and runs! I need to clean up the code, but it seems that the initial usage of DMSwarm is working as intended.

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,462 @@
...
+  PetscCall(DMSwarmRegisterPetscDatatypeField(dm_swarm, "u", num_comp, PETSC_SCALAR));
+  PetscCall(DMSwarmFinalizeFieldRegister(dm_swarm));
+  PetscCall(DMSwarmSetLocalSizes(dm_swarm, num_points, 0));
+  PetscCall(DMSetFromOptions(dm_swarm));
+
+  // -- Set swarm point locations
+  {
+    PetscScalar *point_coords;
+
+    PetscCall(DMSwarmGetField(dm_swarm, "DMSwarmPIC_coor", NULL, NULL, (void **)&point_coords));
```
zatkins-dev, line 219: We should use `DMSwarmPICField_coor` instead of a string constant here.

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,462 @@
...
+
+    PetscCall(DMGetLocalVector(dm_mesh, &U_loc));
+    PetscCall(VecZeroEntries(U_loc));
+    PetscCall(DMGlobalToLocal(dm_mesh, U_mesh, INSERT_VALUES, U_loc));
+
+    PetscCall(DMGetDS(dm_mesh, &ds));
+    PetscCall(PetscDSGetDiscretization(ds, 0, (PetscObject *)&fe));
+    PetscCall(DMSwarmSortGetAccess(dm_swarm));
+    PetscCall(DMPlexGetHeightStratum(dm_mesh, 0, &cell_start, &cell_end));
+    PetscCall(DMSwarmGetField(dm_swarm, "DMSwarmPIC_coor", NULL, NULL, (void **)&coord_points));
```
zatkins-dev, line 278: Use `DMSwarmPICField_coor`

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,462 @@
...
+      PetscCall(PetscFEDestroyCellGeometry(fe, &fe_geometry));
+      PetscCall(DMPlexVecRestoreClosure(dm_mesh, NULL, U_loc, cell, NULL, &u_cell));
+      PetscCall(DMRestoreWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_points_true));
+      PetscCall(DMRestoreWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_points_ref));
+      PetscCall(PetscTabulationDestroy(&tabulation));
+      PetscCall(PetscFree(points));
+    }
+
+    // -- Cleanup
+    PetscCall(DMSwarmRestoreField(dm_swarm, "DMSwarmPIC_coor", NULL, NULL, (void **)&coord_points));
```
zatkins-dev, line 325: Use `DMSwarmPICField_coor`

examples/petsc/dmswarm.c
zatkins-dev:
```suggestion
    PetscCall(DMSwarmGetField(dm_swarm, DMSwarmPICField_coor, NULL, NULL, (void **)&point_coords));
```

examples/petsc/dmswarm.c
zatkins-dev:
```suggestion
    PetscCall(DMSwarmRestoreField(dm_swarm, DMSwarmPICField_coor, NULL, NULL, (void **)&coord_points));
```

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,462 @@
...
+  {
+    PetscScalar *point_coords;
+
+    PetscCall(DMSwarmGetField(dm_swarm, "DMSwarmPIC_coor", NULL, NULL, (void **)&point_coords));
+    for (PetscInt p = 0; p < num_points; p++) {
+      point_coords[p * dim + 0] = -PetscCosReal((PetscReal)(p + 1) / (PetscReal)(num_points + 1) * PETSC_PI);
+      if (dim > 1) point_coords[p * dim + 1] = -PetscSinReal((PetscReal)(p + 1) / (PetscReal)(num_points + 1) * PETSC_PI);
+      if (dim > 2) point_coords[p * dim + 2] = PetscSinReal((PetscReal)(p + 1) / (PetscReal)(num_points + 1) * PETSC_PI);
+    }
+    PetscCall(DMSwarmRestoreField(dm_swarm, "DMSwarmPIC_coor", NULL, NULL, (void **)&point_coords));
```
zatkins-dev:
```suggestion
    PetscCall(DMSwarmRestoreField(dm_swarm, DMSwarmPICField_coor, NULL, NULL, (void **)&point_coords));
```

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,462 @@
...
+  // Allocate index set array
+  PetscCall(VecGetLocalSize(*ref_coords, &num_points_local));
+  PetscCall(DMGetDimension(dm_mesh, &dim));
+  num_points_local /= dim;
+  PetscCall(PetscMalloc1(num_points_local, &point_cell_numbers));
+
+  // Get reference coordinates for each swarm point wrt the elements in the background mesh
+  PetscCall(DMSwarmSortGetAccess(dm_swarm));
+  PetscCall(DMPlexGetHeightStratum(dm_mesh, 0, &cell_start, &cell_end));
+  PetscCall(DMSwarmGetField(dm_swarm, "DMSwarmPIC_coor", NULL, NULL, (void **)&coords_points_true));
```
zatkins-dev, line 81:
```suggestion
  PetscCall(DMSwarmGetField(dm_swarm, DMSwarmPICField_coor, NULL, NULL, (void **)&coords_points_true));
```

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,462 @@
...
+
+  PetscFunctionBeginUser;
+
+  // Create vector to hold reference coordinates
+  {
+    Vec true_coords;
+
+    PetscCall(DMSwarmCreateLocalVectorFromField(dm_swarm, "DMSwarmPIC_coor", &true_coords));
+    PetscCall(VecDuplicate(true_coords, ref_coords));
+    PetscCall(DMSwarmDestroyLocalVectorFromField(dm_swarm, "DMSwarmPIC_coor", &true_coords));
```
zatkins-dev, line 69:
```suggestion
    PetscCall(DMSwarmCreateLocalVectorFromField(dm_swarm, DMSwarmPICField_coor, &true_coords));
    PetscCall(VecDuplicate(true_coords, ref_coords));
    PetscCall(DMSwarmDestroyLocalVectorFromField(dm_swarm, DMSwarmPICField_coor, &true_coords));
```

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    // -- Cleanup
+    PetscCall(DMRestoreWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_cell_points_true));
+    PetscCall(DMRestoreWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_cell_points_ref));
+    PetscCall(PetscFree(cell_points));
+  }
+
+  // Cleanup
+  PetscCall(DMSwarmRestoreField(dm_swarm, "DMSwarmPIC_coor", NULL, NULL, (void **)&coords_points_ref));
```
zatkins-dev, line 108:
```suggestion
  PetscCall(DMSwarmRestoreField(dm_swarm, DMSwarmPICField_coor, NULL, NULL, (void **)&coords_points_ref));
```

examples/petsc/dmswarm.c
zatkins-dev:
```suggestion
    PetscCall(DMSwarmGetField(dm_swarm, DMSwarmPICField_coor, NULL, NULL, (void **)&coord_points));
```

zatkins-dev (changes requested):
We should use the built-in constants for built-in fields

examples/petsc/dmswarm.c
zatkins-dev: Isn't this `DMSwarmGetCellDM`? https://gitlab.com/petsc/petsc/-/blob/zach/bug-dm-swarm-tensor-permutation/src/dm/impls/swarm/swarm.c#L872

jeremylt: dunno how I missed that

zatkins-dev (commented):
empty comment

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,465 @@
...
+
+      for (PetscInt d = 0; d < dim; d++) x[d] = x_array[d * num_mesh_points_loc + i];
+      u_array[i] = EvalU(dim, x);
+    }
+    PetscCall(VecRestoreArray(U_loc, &u_array));
+    PetscCall(VecRestoreArrayRead(X_loc, &x_array));
+    PetscCall(VecZeroEntries(U_mesh));
+    PetscCall(DMLocalToGlobal(dm_mesh, U_loc, INSERT_VALUES, U_mesh));
+    PetscCall(DMRestoreLocalVector(dm_mesh, &U_loc));
+  }
```
zatkins-dev, line 260: The sizes of the coordinate and local vectors aren't related like this in general, especially for mesh order > 1

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,465 @@
...
+      PetscInt        num_points_in_cell;
+
+      PetscCall(DMSwarmSortGetPointsPerCell(dm_swarm, cell, &num_points_in_cell, &points));
+      PetscCall(DMGetWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_points_true));
+      PetscCall(DMGetWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_points_ref));
+      // ---- Reference coordinates for swarm points in background mesh element
+      for (PetscInt p = 0; p < num_points_in_cell; p++) {
+        for (PetscInt d = 0; d < dim; d++) coords_points_true[p * dim + d] = coord_points[points[p] * dim + d];
+      }
+      PetscCall(DMPlexCoordinatesToReference(dm_mesh, cell, num_points_in_cell, coords_points_true, coords_points_ref));
```
zatkins-dev, line 299: This function seems to blow up (`coords_points_ref[0]` and `coords_points_ref[1]` in particular)

examples/petsc/dmswarm.c
jeremylt: oh, we should project the coordinate space to match the solution space, for the purposes of this example

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,465 @@
...
+
+      // ---- Interpolate values from current element in background mesh to swarm points
+      PetscCall(PetscFECreateTabulation(fe, 1, num_points_in_cell, coords_points_ref, 1, &tabulation));
+      PetscCall(DMPlexComputeCellGeometryFEM(dm_mesh, cell, NULL, v, J, invJ, &detJ));
```
zatkins-dev, line 303: we aren't using the outputs of this call, that seems fishy

examples/petsc/dmswarm.c
jeremylt: i.e. remove `geometry_order` and use `solution_order` in its place (or whatever I called those variables)

examples/petsc/dmswarm.c
jeremylt: That's a placeholder for the grad checking code

examples/petsc/dmswarm.c
zatkins-dev: replacing with this fixes the blow up
```suggestion
     PetscCall(DMPlexComputeCellGeometryFEM(dm_mesh, cell, NULL, v, J, invJ, &detJ));
      for (PetscInt p = 0; p < num_points_in_cell; ++p) {
        CoordinatesRealToRef(dim, dim, v0ref, v, invJ, &coord_points_true[p * dim], &coords_points_ref[p * dim]);
      }
```

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,465 @@
...
+
+    PetscCall(DMGetCoordinatesLocal(dm_mesh, &X_loc));
+    PetscCall(DMGetLocalVector(dm_mesh, &U_loc));
+    PetscCall(VecGetLocalSize(U_loc, &num_mesh_points_loc));
+    PetscCall(VecGetArray(U_loc, &u_array));
+    PetscCall(VecGetArrayRead(X_loc, &x_array));
+    for (PetscInt i = 0; i < num_mesh_points_loc; i++) {
+      PetscScalar x[dim];
+
+      for (PetscInt d = 0; d < dim; d++) x[d] = x_array[d * num_mesh_points_loc + i];
```
zatkins-dev, line 252: Our coordinate ordering is wrong here, it should be:
```suggestion
      for (PetscInt d = 0; d < dim; d++) x[d] = x_array[i * dim + d];
```

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,522 @@
...
+
+      PetscCall(DMSwarmSortGetPointsPerCell(dm_swarm, cell, &num_points_in_cell, &points));
+      PetscCall(DMGetWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_points_true));
+      PetscCall(DMGetWorkArray(dm_mesh, num_points_in_cell * dim, MPIU_REAL, &coords_points_ref));
+      // ---- Reference coordinates for swarm points in background mesh element
+      for (PetscInt p = 0; p < num_points_in_cell; p++) {
+        for (PetscInt d = 0; d < dim; d++) coords_points_true[p * dim + d] = coords_points[points[p] * dim + d];
+      }
+      PetscCall(DMPlexComputeCellGeometryFEM(dm_mesh, cell, NULL, v, J, invJ, &detJ));
+      for (PetscInt p = 0; p < num_points_in_cell; ++p) {
```
jeremylt, line 325: Stray ++i instead of i++

examples/petsc/dmswarm.c
jeremylt: As we dig more into this - I think what libCEED really needs is all of the indices of the points in the first local cell, followed by all the indices of the points in the second local cell, etc
something like
4
6
84
102
-1
23
35
76
-1
-1
46
78
...

where -1 indicates a break between cells

right now we have a nested loop with num_elem * num_local_points iterations all over the place, but this would bring us down to num_elem * ~num_points_per_elem, which is much, much fewer iterations

jeremylt: resolved per conversation in Zulip - see changes in latest commit

jeremylt:
Ready for review - merge blocked by https://gitlab.com/petsc/petsc/-/merge_requests/6843

zatkins-dev:
> Ready for review - merge blocked by https://gitlab.com/petsc/petsc/-/merge_requests/6843

No longer blocked!

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,606 @@
...
+  DMSwarmCreateReferenceCoordinates - Compute the cell reference coordinates for local DMSwarm points.
+
+  Collective
+
+  Input Parameters:
++ dm_swarm  - the `DMSwarm`
+
+  Output Parameter:
++ is_points    - The IS object for indexing into points per cell
++ X_points_ref - Vec holding the cell reference coordinates for local DMSwarm points
```
zatkins-dev, line 272:
```suggestion
. dm_swarm  - the `DMSwarm`

  Output Parameter:
+ is_points    - The IS object for indexing into points per cell
- X_points_ref - Vec holding the cell reference coordinates for local DMSwarm points
```
Minor: fix PETSc docstring parameter prefixes

jeremylt:
Good point - we just need to update the PETSc builds on Noether. I got the Ratel ones, but I don't think I "own" the folders for the libCEED CI builds.

Really... we should get around to building containers, at least for our CPU builds. Not sure how to best also do containers for GPU builds (though I'd love for that to be an option for Ratel too)

examples/petsc/dmswarm.c
jeremylt: I meant to ask about that - what's the magic decoder ring for what the . vs + vs - mean?

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,606 @@
...
+
+/*@C
+  DMSwarmCreateReferenceCoordinates - Compute the cell reference coordinates for local DMSwarm points.
+
+  Collective
+
+  Input Parameters:
+. dm_swarm  - the `DMSwarm`
+
+  Output Parameter:
```
zatkins-dev, line 270:
```suggestion
  Input Parameter:
. dm_swarm  - the `DMSwarm`

  Output Parameters:
```
also plural/singular

examples/petsc/dmswarm.c
jedbrown: In PETSc? That's a sowing idiosyncrasy.
https://wgropp.cs.illinois.edu/projects/software/sowing/doctext/node12.htm#Node12

zatkins-dev:
Are we planning on implementing the other direction, i.e. points to cells, in this example?

jeremylt:
Eventually. I think we can incrementally add functionality as our libCEED infrastructure matures this month.

jeremylt:
I plan to update this with #1333 

If we add this to the v0.12 release, we can use it to help paint the picture that we are moving to MPM. Then we can refine this example when we add the CeedOperator with *AtPoints interface (hopefully by the end of the month, or at least a prototype version)

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,990 @@
...
+//
+// Build with:
+//
+//     make dmswarm [PETSC_DIR=</path/to/petsc>] [CEED_DIR=</path/to/libceed>]
+//
+// Sample runs:
+//
+//     ./dmswarm
+///
+//TESTARGS -ceed {ceed_resource} -test -dm_plex_dim 3 -dm_plex_box_faces 3,3,3 -dm_plex_box_lower -1.0,-1.0,-1.0 -dm_plex_simplex 0 -dm_plex_hash_location true -num_comp 2 -uniform_swarm
```
zatkins-dev, line 21: Can we add either a second test or just an example invocation for the lumped mass case? And maybe label this one as KSP?

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,990 @@
...
+    PetscCall(PetscOptionsInt("-dm_plex_dim", "Background mesh dimension", NULL, dim, &dim, NULL));
+    PetscCall(PetscOptionsIntArray("-dm_plex_box_faces", "Number of cells", NULL, num_cells, &dim, NULL));
+    PetscInt total_num_cells = num_cells[0] * num_cells[1] * num_cells[2];
+    PetscInt points_per_cell = 1;
+
+    for (PetscInt i = 0; i < dim; i++) points_per_cell *= solution_order + 2;
+    num_points = 0;
+    for (PetscInt i = 0; i < total_num_cells; i++) num_points += points_per_cell;
+  }
+  PetscCall(PetscOptionsInt("-points", "Number of swarm points", NULL, num_points, &num_points, NULL));
```
zatkins-dev, line 119: The `-points` option (and its default) are somewhat unclear. Can we make this a `-points_per_cell` option (or otherwise support that)?

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,990 @@
...
+  PetscCall(PetscOptionsScalar("-tolerance", "Absolute tolerance for swarm point values", NULL, tolerance, &tolerance, NULL));
```
zatkins-dev, line 120:
```suggestion
  PetscCall(PetscOptionsScalar("-tolerance", "Tolerance for swarm point values and projection relative L2 error", NULL, tolerance, &tolerance, NULL));
```
Should make it clear that we are using this differently for point values and projection error

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,990 @@
...
+
+  PetscCall(PetscInitialize(&argc, &argv, NULL, help));
+  comm = PETSC_COMM_WORLD;
+
+  // Read command line options
+  PetscOptionsBegin(comm, NULL, "libCEED example using PETSc with DMSwarm", NULL);
+
+  PetscCall(PetscOptionsBool("-test", "Testing mode (do not print unless error is large)", NULL, test_mode, &test_mode, NULL));
+  PetscCall(PetscOptionsBool("-gauss_swarm", "Use gauss points for coordinates in swarm", NULL, set_gauss_swarm, &set_gauss_swarm, NULL));
+  PetscCall(PetscOptionsBool("-uniform_swarm", "Use uniform coordinates in swarm", NULL, set_uniform_swarm, &set_uniform_swarm, NULL));
```
zatkins-dev, line 96: In an ideal world this should be an enum, but I don't think it really matter for this example.

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,990 @@
...
+
+typedef PetscErrorCode (*DMFunc)(PetscInt dim, PetscReal time, const PetscReal x[], PetscInt Nc, PetscScalar *u, void *ctx);
+
+typedef struct DMSwarmCeedContext_ *DMSwarmCeedContext;
+struct DMSwarmCeedContext_ {
+  Ceed                ceed;
+  CeedVector          u_mesh_loc, v_mesh_loc, u_mesh_elem, u_points_loc, u_points_elem, x_ref_points_loc, x_ref_points_elem;
+  CeedElemRestriction restriction_u_mesh, restriction_x_points, restriction_u_points;
+  CeedBasis           basis_u;
+};
```
zatkins-dev, line 46: I think we should revisit this struct when we integrate with Ratel -- parts of this should be stored on the whole mesh, but other parts are field-specific. Maybe something like this for handling multiple fields? Not sure how we will want to minimize redundant work computing particles in cells. 
```c
struct DMSwarmCeedFieldContext_ {
  CeedElemRestriction restriction_u_mesh, restriction_u_points;
  CeedBasis           basis_u;
  const char          fieldname[MAX_NAME_LENGTH];
}
```

zatkins-dev (approved):
Mostly small comments and thoughts about moving forward. I would be fine with merging in its current state.

examples/petsc/dmswarm.c
jeremylt: Added in the lastest - good idea

examples/petsc/dmswarm.c
jeremylt: Definitely - I think we'll ultimately store this into an operator field so we don't have to create a special struct

examples/petsc/dmswarm.c
jeremylt: Yeah. My main hangup is that I couldn't figure out what to call the other version of the points distribution? (Though we don't really use it much currently it is nice having the option to have a different number of points in each cell to see what that does to accuracy)

examples/petsc/dmswarm.c
jeremylt: Added a `-points_per_cell` option

examples/petsc/dmswarm.c
zatkins-dev: That's fair, maybe like "path" or "semicircle"?

zatkins-dev:
Will we want to update the `junit.py` to ensure this is only run with CPU backends?

jeremylt:
`ApplyBlock` should give a "Not supported" message, so it should error out ok

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,998 @@
...
+
+// ------------------------------------------------------------------------------------------------
+// Swarm point location utility
+// ------------------------------------------------------------------------------------------------
+PetscErrorCode DMSwarmInitalizePointLocations(DM dm_swarm, PetscInt num_points, PetscBool set_gauss_swarm, PetscBool set_uniform_swarm) {
+  PetscFunctionBeginUser;
+
+  if (set_gauss_swarm || set_uniform_swarm) {
+    // -- Set gauss quadrature point locations in each cell
+    PetscSwarm user_set_points_per_cell = PETSC_FALSE;
```
zatkins-dev, line 528:
```suggestion
    PetscBool user_set_points_per_cell = PETSC_FALSE;
```

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,998 @@
...
+    PetscInt   dim = 3, points_per_cell = num_points;
+    PetscInt   num_cells[] = {1, 1, 1};
+    DM         dm_mesh;
```
zatkins-dev, line 531:
```suggestion
    PetscBool user_set_points_per_cell = PETSC_FALSE;
    PetscInt  dim = 3, points_per_cell = num_points;
    PetscInt  num_cells[] = {1, 1, 1};
    DM        dm_mesh;
```

examples/petsc/dmswarm.c
```diff
@@ -0,0 +1,998 @@
...
+        for (PetscInt k = 0; k < points_per_cell_1d; k++) {
+          PetscInt p = (i * points_per_cell_1d + j) * points_per_cell_1d + k;
+
+          point_coords[p * dim + 0] = points_1d[i];
+          point_coords[p * dim + 1] = points_1d[j];
+          point_coords[p * dim + 2] = points_1d[k];
+        }
+      }
+    }
+    PetscCall(DMSwarmSetPointCoordinatesCellwise(dm_swarm, points_per_cell_1d * points_per_cell_1d * points_per_cell_1d, point_coords));
```
zatkins-dev, line 569: `points_per_cell_1d` is scoped wrong here

jedbrown (approved):
LGTM. We've run and visualized results here, and it seems correct.

jeremylt:
@jedbrown can you rebuild PETSc for CI? I can't
```
git@gitlab.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

jedbrown:
Done.

jeremylt:
CI is running. I'll squash+merge when it finishes.

Issue #1283, opened 2023-08-09
Python Updates


jeremylt:
Two issues have come up in Ratel as points to discuss/fix here

- [x] Update the libCEED `junit.py` to reflect updates in Ratel's

- [ ] Discuss if we want to change Python linter [Black, ruff]

zatkins-dev:
I think that it would be good to separate out the parts of `junit.py` which should be exactly the same between Ratel and libCEED to simplify maintaining parity in the future

jeremylt:
True. A lot of it should carry over, except for the specific skip rules and paths to the source files

PR #1285, opened 2023-08-10
refactor junit.py for parity with Ratel

zatkins-dev:
This PR refactors `junit.py` into a CEED-specific `junit.py` and a common driver `junit_common.py`, the latter of which will have an exact copy in the Ratel repository. This should prevent future parity issues between the libCEED and Ratel test suites.

Addresses part of #1283.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -642,6 +642,7 @@ allexamples = $(examples) $(external_examples)
 search ?= t ex
 realsearch = $(search:%=%%)
 matched = $(foreach pattern,$(realsearch),$(filter $(OBJDIR)/$(pattern),$(tests) $(allexamples)))
+JUNIT_BATCH ?= ''
 
 # Test core libCEED
 test : $(matched:$(OBJDIR)/%=run-%)
@@ -661,7 +662,7 @@ prove-all :
 	+$(MAKE) prove realsearch=%
 
 junit-% : $(OBJDIR)/%
-	@printf "  %10s %s\n" TEST $(<:$(OBJDIR)/%=%); $(PYTHON) tests/junit.py --ceed-backends $(BACKENDS) --nproc $(NPROC_TEST) $(<:$(OBJDIR)/%=%)
+	@printf "  %10s %s\n" TEST $(<:$(OBJDIR)/%=%); $(PYTHON) tests/junit.py --ceed-backends $(BACKENDS) --nproc $(NPROC_TEST) --junit-batch $(JUNIT_BATCH) $(<:$(OBJDIR)/%=%)
 
 junit : $(matched:$(OBJDIR)/%=junit-%)
 
--- a/Makefile
+++ b/Makefile
@@ -616,8 +616,12 @@ $(examples) : $(libceed)
 $(tests) : $(libceed)
 $(tests) $(examples) : override LDFLAGS += $(if $(STATIC),,-Wl,-rpath,$(abspath $(LIBDIR))) -L$(LIBDIR)
 
+# Set number processes for testing
+NPROC_TEST ?= 1
+export NPROC_TEST
+
 run-% : $(OBJDIR)/%
-	@$(PYTHON) tests/junit.py --mode tap $(<:$(OBJDIR)/%=%)
+	@$(PYTHON) tests/junit.py --mode tap --ceed-backends $(BACKENDS) --nproc $(NPROC_TEST) $(<:$(OBJDIR)/%=%)
 
 external_examples := \
 	$(if $(MFEM_DIR),$(mfemexamples)) \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -616,8 +616,21 @@ $(examples) : $(libceed)
...
+export NPROC_TEST
+
+# Make sure the backend list is separated with spaces and not newlines
+EMPTY :=
+SPACE := $(EMPTY) $(EMPTY)
+define \n
+
+
+endef
+BACKENDS_LIST := $(subst ${\n},$(SPACE),$(BACKENDS))
```
zatkins-dev, line 630: I don't particularly like this -- it is only necessary because the GitLab CI passes a `\n` separated list of backends instead of a space-separated one. @jedbrown  or @jeremylt, I'm not sure how to edit those pipelines (if that's possible)

jeremylt: `BACKENDS_CPU=$(make info-backends-all | grep -o '/cpu[^ ]*')`

This is the magic evocation that's creating the hassle, I think

zatkins-dev: Yep, it is

zatkins-dev: we could definitely just pipe that into sed to replace the newlines with spaces

zatkins-dev:
```
BACKENDS_CPU=$(make info-backends-all | grep -o '/cpu[^ ]*' | sed -e 's/\n/ /g')
```
I think this would work

jeremylt: The pattern is here: https://github.com/CEED/libCEED/blob/main/.gitlab-ci.yml

tests/junit.py
```diff
@@ -1,264 +1,128 @@
...
+            return f'occa mode not supported {test} {resource}'
+        elif 'Backend does not implement' in stderr:
+            return f'not implemented {test} {resource}'
+        elif 'Can only provide HOST memory for this backend' in stderr:
+            return f'device memory not supported {test} {resource}'
+        elif 'Test not implemented in single precision' in stderr:
+            return f'not implemented {test} {resource}'
+        elif 'No SYCL devices of the requested type are available' in stderr:
+            return f'sycl device type not available {test} {resource}'
+        return None
```
jeremylt, line 76: We can take the opportunity no make this more human readable

tests/junit.py
```diff
@@ -1,264 +1,128 @@
...
+        if 'OCCA backend failed to use' in stderr:
+            return f'occa mode not supported {test} {resource}'
+        elif 'Backend does not implement' in stderr:
+            return f'not implemented {test} {resource}'
+        elif 'Can only provide HOST memory for this backend' in stderr:
+            return f'device memory not supported {test} {resource}'
+        elif 'Test not implemented in single precision' in stderr:
+            return f'not implemented {test} {resource}'
+        elif 'No SYCL devices of the requested type are available' in stderr:
+            return f'sycl device type not available {test} {resource}'
```
jeremylt, line 75:
```suggestion
            return f'SYCL device type not available {test} {resource}'
```

tests/junit.py
```diff
@@ -1,264 +1,128 @@
...
+            test.startswith('petsc-') and contains_any(resource, ['occa']),
+            test.startswith('fluids-') and contains_any(resource, ['occa']),
+            test.startswith('solids-') and contains_any(resource, ['occa']),
+            test.startswith('t318') and contains_any(resource, ['/gpu/cuda/ref']),
+            test.startswith('t506') and contains_any(resource, ['/gpu/cuda/shared']),
+        )) else None
+
+    def check_post_skip(self, test: str, spec: TestSpec, resource: str, stderr: str) -> Optional[str]:
+        if 'OCCA backend failed to use' in stderr:
+            return f'occa mode not supported {test} {resource}'
```
jeremylt, line 67:
```suggestion
            return f'OCCA mode not supported {test} {resource}'
```

tests/junit.py
```diff
@@ -1,264 +1,128 @@
...
+if __name__ == '__main__':
+    args = create_argparser().parse_args()
+
+    # run tests
+    mode: RunMode = RunMode(args.mode.lower())
+    result: TestSuite = run_tests(args.test, args.ceed_backends, mode, args.nproc, CeedSuiteSpec())
+
+    # build output
+    if mode is RunMode.JUNIT:
+        junit_batch: str = f'-{os.environ["JUNIT_BATCH"]}' if 'JUNIT_BATCH' in os.environ else ''
```
jeremylt, line 119: This should also probably be an explicit argument? I forgot about this one when making the changes in Ratel

jeremylt:
I'd run `autopep8 --in-place --aggressive --max-line-length=150 this-file.py` (pretty sure that's the options list) on these files

zatkins-dev:
> I'd run `autopep8 --in-place --aggressive --max-line-length=150 this-file.py` (pretty sure that's the options list) on these files

Ah, wasn't sure on your command line args. I think I used `--max-line-length 170`.

jeremylt:
> Ah, wasn't sure on your command line args. I think I used `--max-line-length 170`.

That works too. Anything reasonably bigger than the default (tiny) line length.

tests/junit.py
zatkins-dev: updated them with more readable output, as well as for the pre-skip reasons

jeremylt (approved):
This looks great to me. Can you open the equivalent MR in Ratel as well so we ensure the common file does everything needed? Then we can approve and merge the two at the same time.

jeremylt:
Thanks, both look ready to merge to me.

PR #1286, opened 2023-08-14
fluids: Refactor gravity option

jrwrigh:
- Use `-gravity` instead of `-g`
 - Enforce that the input should be a length 3 vector
 - Update documentation

Misc: Also update documented default value for `mu`.

Changes stems from this PR comment: https://github.com/CEED/libCEED/pull/1282#pullrequestreview-1574293147


jedbrown (approved):
LGTM, thanks!

PR #1290, opened 2023-08-15
Vector - error-free path for GetArray[Write] when size=0

jeremylt:
Follow-up to https://github.com/CEED/libCEED/pull/897

I think this will fix https://gitlab.com/micromorph/ratel/-/merge_requests/579 more directly @zatkins-dev, but I haven't verified yet. If it doesn't, we'll need similar logic added somewhere else too.


zatkins-dev:
This doesn't fully fix the issue. Since in `CeedVectorGetArrayWrite`, a vector with length `0` returns `NULL` without doing anything to the `CeedVector`, calling `CeedVectorSetValue` on a length `0` array doesn't mark the array has having valid data.

I think a "better" fix would be to treat a `CeedVector` of length `0` as if it always has valid data, although I worry that will break some things. Unfortunately, the return value of `malloc(0)` is implementation-defined, so there's no good way to handle length `0` arrays.

zatkins-dev:
I have a PR #1291 which fixes the issue by allowing for empty `CeedVector` arrays. Let me know what you think about that option.

PR #1291, opened 2023-08-15
Vector - error-free usage when length = 0

zatkins-dev:
This PR addresses the same issue as #1290. 

It differs by changing the behavior of `CeedVectorHasValidArray` to always output `true` for length `0` vectors.
Additional changes are made to short-circuit calls to most functions in the `CeedVector` interface when a vector has length `0`.
This ensures that calls to `CeedVector*` with an empty `CeedVector` will not cause crashes, regardless of the backend.

This PR also adds test cases for most `t1*-vector` tests for length `0` vectors.

### DIFF
--- a/tests/t101-vector.c
+++ b/tests/t101-vector.c
@@ -1,8 +1,12 @@
 /// @file
 /// Test CeedVectorSetValue
 /// \test Test CeedVectorSetValue
+
+//TESTARGS(name="length 10") {ceed_resource} 10
+//TESTARGS(name="length 0") {ceed_resource} 0
 #include <ceed.h>
 #include <stdio.h>
+#include <stdlib.h>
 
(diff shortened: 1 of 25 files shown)


tests/t101-vector.c
```diff
@@ -1,8 +1,12 @@
 /// @file
 /// Test CeedVectorSetValue
 /// \test Test CeedVectorSetValue
+
+//TESTARGS(name="length 10") {ceed_resource} 10
```
jeremylt, line 5: Can you double check the minimum number of tests we need to add a length 0 version of so we don't add more test time than we strictly need?

zatkins-dev: sure, I figured they weren't all necessary

zatkins-dev: okay, I think that I've slimmed it down

jeremylt (approved):
Thanks!

Issue #1288, opened 2023-08-14
`magma` backend `CeedElemRestriction`


sebastiangrimberg:
It looks to me like the MAGMA backend implementation of `CeedElemRestriction` is very similar (if not basically the same?) as the fallback `cuda-ref` or `hip-ref` implementations. Can someone perhaps provide context for why it has a separate implementation or, as a separate question, would it make sense to remove the MAGMA implementations? This would be nicer for long term library maintainability as it would allow us to reduce some code duplication, and especially would make GPU backends for #1265 require fewer added lines of code. Of course if there is actually a reason for having the different implementations of `CeedElemRestriction` then I am all for keeping both.

nbeams:
Starting with the short answer first: 
> would it make sense to remove the MAGMA implementations? 

I have been wondering this myself, lately.  If we did so, I think it would make the most sense to get rid of the `magma-det` backend and turn the `magma` backend into, essentially, the `magma-det` backend (but with simpler setup and shorter name). I'm not sure if that would be considered a major/breaking change or not?

The longer answer: 

The restrictions aren't exactly the same, as the MAGMA restrictions use atomic adds in the transpose restrictions, and the ref kernels do not.  Hence the distinction between the `magma` backend (not deterministic due to atomic adds) and `magma-det` (deterministic, uses MAGMA basis but ref restrictions).  I haven't studied it rigorously, but most of my tests over the past 2-ish years have indicated no real performance benefit to the version with atomics, so that's why I've been wondering if we should simplify and cut it, as you suggested.

As for why they exist, that's mostly just a remnant of libCEED development history.  At the time we added these restrictions to the `magma` backend, the other GPU backends used the same E-vector ordering as the CPU backends.  For the non-tensor basis calculations in MAGMA, we wanted to swap the E-vector ordering to have the layout we needed for the GEMMs.  So, we added our own restrictions to get the ordering we wanted.  Some time after this, we ended up switching the E-vector ordering of the other GPU backends to be the same as what MAGMA was using, which then allowed MAGMA to use the ref restrictions again, and added the `magma-det` backend which did just that.

jeremylt:
I am fine with trimming out the MAGMA backend restrictions, though we should do a performance study first to make sure we aren't making a bad decision here. I'd be especially interested in seeing the performance study on non-tensor elements since that's the area where we really look to the MAGMA backend for performance.

sebastiangrimberg:
[...]

PR #1292, opened 2023-08-15
Unify `magma` backend restriction with `cuda-ref` and `hip-ref`

sebastiangrimberg:
This keeps the previous behavior of `cuda-ref`, `hip-ref`, `magma`, and `magma/ref` backends where the first, second, and last are deterministic but `magma` is not. Since the element restriction code is compiled at runtime, we can just pull the deterministic kernel vs. non-deterministic (with `atomicAdd`) based on what the parent backend is. This allows for us to limit code duplication while also avoiding compile time configuration of a deterministic element restriction transpose (see https://github.com/CEED/libCEED/issues/1288).

This also opens the opportunity to make `[cuda/hip]-ref` or `*-shared` non-deterministic by default and add `[cuda/hip]-ref/det` variants like for `magma`. I can add this to this PR, but this changes default behavior and would require updating the documentation, etc. So for now this is just a starting point.


backends/magma/ceed-magma-det.c
```diff
@@ -19,40 +19,17 @@ static int CeedInit_Magma_Det(const char *resource, Ceed ceed) {
...
-
-  int currentDeviceID;
-  magma_getdevice(&currentDeviceID);
-  if (deviceID >= 0 && currentDeviceID != deviceID) {
-    magma_setdevice(deviceID);
-    currentDeviceID = deviceID;
-  }
-  // create a queue that uses the null stream
-  data->device = currentDeviceID;
-
```
nbeams, line 37: Since you mentioned this in your TODO:  I do think we still need this here, for the case where we want to set a device ID other than default (0).  Not only do we want to set the device in the backend data (line 36), but we want to make sure the `magma` backend `ceedref` objects uses the correct device.  If we've already set the current device to something other than 0, then the `magma` delegate object will use the current value (after calling `magma_getdevice` in its own initialization), even though we don't explicitly pass the device ID in the resource string when creating `ceedref` (line 42 or 44).  It doesn't have to be this exact code, but we do need to make sure we can handle non-zero device ID specifications.

sebastiangrimberg: Ah, thanks @nbeams! I've added this back in, via a common `ceed-magma-common.h/c` similar to what is done for `ceed-cuda/hip-common.h/c`. This should now be resolved!

sebastiangrimberg:
Some benchmarking results from the PETSc BPs on a single V100 (averaged over 10 calls to `KSPApply`):

`./bps -ceed /gpu/cuda/ref,/gpu/cuda/shared,/gpu/cuda/magma,/gpu/cuda/magma/det -problem bp1,bp3 -degree 2 -q_extra 1 -local_nodes 100000 -ksp_max_it 1000 -dm_vec_type cuda -use_gpu_aware_mpi 0`

`main`:

- BP1
    - `/gpu/cuda/ref`: 382.185 (382.185) million
    - `/gpu/cuda/shared`: 430.387 (430.387) million
    - `/gpu/cuda/magma`: 430.342 (430.342) million
    - `/gpu/cuda/magma/det`: 470.795 (470.795) million
- BP3
    - `/gpu/cuda/ref`: 185.279 (185.279) million
    - `/gpu/cuda/shared`: 267.995 (267.995) million
    - `/gpu/cuda/magma`: 259.761 (259.761) million
    - `/gpu/cuda/magma/det`: 275.36 (275.36) million

This PR:

- BP1
    - `/gpu/cuda/ref`: 384.122 (384.122) million
    - `/gpu/cuda/shared`: 431.78 (431.78) million
    - `/gpu/cuda/magma`: 439.73 (439.73) million
    - `/gpu/cuda/magma/det`: 472.557 (472.557) million
- BP3
    - `/gpu/cuda/ref`: 184.382 (184.382) million
    - `/gpu/cuda/shared`: 267.246 (267.246) million
    - `/gpu/cuda/magma`: 264.148 (264.148) million
    - `/gpu/cuda/magma/det`: 274.607 (274.607) million

Adding `-simplex` to test non-tensor:

- BP1
    - `/gpu/cuda/ref`: 213.481 (213.481) million
    - `/gpu/cuda/shared`: 227.984 (227.984) million
    - `/gpu/cuda/magma`: 195.89 (195.89) million
    - `/gpu/cuda/magma/det`: 316.437 (316.437) million
- BP3
    - `/gpu/cuda/ref`: 130.388 (130.388) million
    - `/gpu/cuda/shared`: 129.475 (129.475) million
    - `/gpu/cuda/magma`: 122.026 (122.026) million
    - `/gpu/cuda/magma/det`: 155.724 (155.724) million

This PR:

- BP1
    - `/gpu/cuda/ref`: 210.664 (210.664) million
    - `/gpu/cuda/shared`: 233.607 (233.607) million
    - `/gpu/cuda/magma`: 282.889 (282.889) million
    - `/gpu/cuda/magma/det`: 322.007 (322.007) million
- BP3
    - `/gpu/cuda/ref`: 130.213 (130.213) million
    - `/gpu/cuda/shared`: 129.264 (129.264) million
    - `/gpu/cuda/magma`: 150.29 (150.29) million
    - `/gpu/cuda/magma/det`: 155.395 (155.395) million

The only notable difference is for the non-tensor basis case, where the non-deterministic `/gpu/cuda/magma` backend with `atomicAdd` is much higher performing than before. In both cases, however, it is slower than the deterministic `*/ref` restriction which I thought was interesting.

jeremylt:
Looks good to me, thanks! I'm happy if @nbeams is happy

jedbrown (approved):
I defer to @nbeams, but this looks reasonable to me.

nbeams:
I don't know if I'll have time to review in detail until next week, but I'll try to do it Monday...

sebastiangrimberg:
> I don't know if I'll have time to review in detail until next week, but I'll try to do it Monday...

No rush @nbeams, thank you for your help!

nbeams (approved):
LGTM! Confirmed device ID !=0 works with both magma backends on a local machine that has multiple NVIDIA GPUs per node.  I think most of the deletions from `ceed-magma.h` are indeed just vestiges of the backend development history and were mostly for internal testing/use in the first place.

jeremylt:
Are we good to merge?

sebastiangrimberg:
Yeah! I was waiting for you but I can rebase + merge if you want? Let me know.

PR #1295, opened 2023-08-17
Silence compiler warnings for PETSc examples

sebastiangrimberg:
Remove some trailing semicolons from PETSc examples function definitions to silence compiler warnings.


jedbrown (approved):
Thanks!

PR #1298, opened 2023-08-20
Fluids: Add Taylor-Green vortex problem

jrwrigh:
Adds the Taylor-Green vortex problem to the fluids example. This is just a new initial condition, with intention to be run tri-periodic.

Given it's just an initial condition, do we want/need a test for that?


jeremylt:
I'm generally a fan of 'test everything because if it's untested then it's certainly going to break without us knowing'

examples/fluids/navierstokes.c
```diff
@@ -18,6 +18,7 @@
 //     ./navierstokes -ceed /cpu/self -problem density_current -degree 1
 //     ./navierstokes -ceed /gpu/cuda -problem advection -degree 1
 //
+//TESTARGS(name="Taylor Green Vortex IC") -ceed {ceed_resource} -problem taylor_green -test_type solver -dm_plex_dim 3 -dm_plex_box_faces 6,6,6 -ts_max_steps 0 -compare_final_state_atol 1e-12 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-taylor-green-IC.bin
 //TESTARGS(name="blasius_SGS_DataDriven") -ceed {ceed_resource} -options_file examples/fluids/tests-output/blasius_stgtest.yaml -sgs_model_type data_driven -sgs_model_dd_leakyrelu_alpha 0.3 -sgs_model_dd_parameter_dir examples/fluids/dd_sgs_data -ts_dt 2e-9 -state_var primitive -ksp_rtol 1e-12 -snes_rtol 1e-12 -stg_mean_only -stg_fluctuating_IC -test_type solver -compare_final_state_atol 1e-10 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-blasius-sgs-data-driven.bin
```
jeremylt, line 22: We should (here or in another PR) change all these test names to the "human readable" convention

jrwrigh: Sure, might as well do it now.

jrwrigh: Done.

jeremylt:
I think switching to running some of these tests in parallel might help. Something for a future PR

jeremylt (approved):
Looks like Noether is happy too

jeremylt:
> Looks like that currently goes through prove to do parallel tests though?

Right. The tests themselves are still executed serially (though we could update that too) but each test can be executed in parallel.

PR #1299, opened 2023-08-21
Style fixes for `cuda` and `hip` backend consistency

sebastiangrimberg:
These changes come from `diff`-ing the identically named files in each of the `backend/hip*` vs. `backend/cuda*` and `include/jit-source/hip` vs. `include/jit-source/cuda` directories. Also includes some corresponding small changes to the `sycl*` backends (see first commit).

Style changes:
- Minor variable renaming and whitespace changes to make the backends consistent and ensure maintainability in the future when making changes which should affect all device backends similarly.
- Some IWYU fixes between CUDA and HIP backends.
- Include guards on `jit-source` code (not necessary in most cases, but it's there for `*-shared` backends as well as `sycl*` and is needed in certain places, so to be consistent I think it's better to have everywhere.
- Removed unused `jit-source/hip/hip-shared-basis.h` and `jit-source/cuda/cuda-shared-basis.h`, as well as `sycl-shared-basis.hpp` file.
- Similar minor changes to `sycl` backend for consistency against the matching `hip` or `cuda` backends, no change to actual functionality (all style).

One functional change:
- `hip-ref` tensor backend `CEED_EVAL_WEIGHT` implementation in `jit-source/hip/hip-ref-basis-tensor.h` differed from the `cuda-ref` implementation, for no real apparent reason, which adds some confusion, since these are reference implementations especially. I feel they should be the same looking at the code (HIP-specific tuning could go into `hip/shared` but actually the weight implementations there are identical to `cuda/shared`). If I am wrong here, happy to revert. EDIT: The reason was that https://github.com/CEED/libCEED/pull/803 introduced a fix for `cuda/ref` but a similar adjustment was not made for `hip/ref`. So this PR brings the `hip/ref` kernels into agreement.


jeremylt:
I really appreciate these consistency passes you've done. Also on my to-do eventually is to clean up inconsistent locations for variable declarations (in many files but not all, variable declarations are all at the top of a block).

nbeams:
> hip-ref tensor backend CEED_EVAL_WEIGHT implementation in jit-source/hip/hip-ref-basis-tensor.h differed from the cuda-ref implementation, for no real apparent reason

In case you are curious, the reason/divergence point between the two backends was here: https://github.com/CEED/libCEED/issues/802 / https://github.com/CEED/libCEED/pull/803

Since it looks like you've brought the current CUDA kernels into HIP instead of vice versa, this issue shouldn't pop up again (it was a very weird/niche issue anyway).

sebastiangrimberg:
> In case you are curious, the reason/divergence point between the two backends was here: #802 / #803

Ah! Thanks @nbeams. Indeed this PR then just ports the same fix for `cuda/ref` from #803 into `hip/ref`. I've updated the PR comments accordingly.

sebastiangrimberg:
> I really appreciate these consistency passes you've done. Also on my to-do eventually is to clean up inconsistent locations for variable declarations (in many files but not all, variable declarations are all at the top of a block).

Thanks! I'm glad, and it keeps me sane when trying to introduce changes across the different backends to have the code be the same when it's supposed to be the same. If curious, most of these changes were found the assistance of a short script:

```
#!/bin/bash

CUDA_DIR=$1
ROOT_DIR=$(dirname $CUDA_DIR)
BASE_DIR=$(basename $CUDA_DIR)

CUDA_FILES="$CUDA_DIR/*.[ch]*"
for f in $CUDA_FILES; do
    echo "Processing $f..."
    HIP_NAME=$(echo $f | sed "s|cuda|hip|g")
    # echo "diff $f $HIP_NAME"
    diff $f $HIP_NAME
done
```

Then I can just run `./diff.sh backends/cuda-ref` and the differences stick out pretty clearly from the expected `hip` -> `cuda` name changes.

> Also on my to-do eventually is to clean up inconsistent locations for variable declarations (in many files but not all, variable declarations are all at the top of a block).

This is definitely nice and would further help track down bugs from (near) duplicate code. But obviously involves a lot of work and touches a lot of code!

jeremylt:
I'm happy to merge these changes if they are ready

sebastiangrimberg:
> I'm happy to merge these changes if they are ready

Nice! Marked as ready, I've been testing it plenty and everything is working as expected (in addition to unit tests passing).

PR #1300, opened 2023-08-21
H(div) and H(curl) basis support for `cuda-ref` and `hip-ref` backends

sebastiangrimberg:
Add support for `BasisCreateHdiv` and `BasisCreateHcurl`, as well as `CEED_EVAL_INTERP`, `CEED_EVAL_DIV`, and `CEED_EVAL_CURL` for non-tensor `cuda-ref` and `hip-ref` backends.


sebastiangrimberg:
This PR is up to date with `main` after #1315 and should be ready for review. Thanks!

nbeams:
Sorry for the delayed review. Overall, LGTM. I ran the MFEM benchmarks, mostly out of an abundance of caution because I knew from our previous experience that issues can pop up with 3D non-tensor elements that may not be captured by the unit tests.  The first set of results looked really weird so I ran again -- the weirdest (2nd order elements) appears to have been some sort of transient blip (thankfully), but the drop to ~60% of `main` for 4th order elements is reproducible. 

[AMD MI210 BP3 hip-ref comparison - run 1 with weird 2nd order results](https://github.com/CEED/libCEED/files/13493156/hip_ref_compare_latest_mi210_bp3_speedup.pdf)
[AMD MI210 BP3 hip-ref comparison - run 2 with reproduced 4th order drop](https://github.com/CEED/libCEED/files/13493157/hip_ref_compare_latest_mi210_bp3_run2_speedup.pdf)

This is the reference backend, so performance isn't really a major concern, and I don't necessarily think this should hold up the merging of this PR, especially since we've already merged the MAGMA backend H(curl) and H(div) PR. It's just potentially interesting, and not what I expected. (I didn't check any NVIDIA hardware, but I suspect we might see the same trend as before, where performance was less affected than for AMD.)

nbeams (approved):

sebastiangrimberg:
Thanks for the review @nbeams. I don't love seeing this performance regression but I agree that:

1. This is the `ref` backend and thus isn't designed to be tuned for performance.
2. Given how specific the "blip" is in terms of problem size and solution order, this could be an AMD compiler issue which for the given sizes leads to suboptimal occupancy/register spillage. So, we could spend time tuning it in libCEED or we could just hope that future compiler/runtime versions will be "smarter" and resolve this issue automatically as it appears to be with `nvcc`, for example.

jeremylt:
This looks like its ready to merge?

backends/cuda-ref/ceed-cuda-ref-basis.c
```diff
@@ -38,7 +36,7 @@ int CeedBasisApply_Cuda(CeedBasis basis, const CeedInt num_elem, CeedTransposeMo
   CeedCallBackend(CeedVectorGetArrayWrite(v, CEED_MEM_DEVICE, &d_v));
 
   // Clear v for transpose operation
-  if (t_mode == CEED_TRANSPOSE) {
+  if (transpose) {
```
jeremylt, line 39: I haven't noticed this before, but our style would usually be `is_transpose` or something like that for bools. At some point I should check the GPU backends for variable naming conventions, but its not important here.

sebastiangrimberg: Sounds good to leave it to a later PR where it can be addressed everywhere (sycl, occa backends both have `if (transpose)` statements in my quick search).

PR #1301, opened 2023-08-21
H(div) and H(curl) basis support for `magma` backend

### LINKED ISSUE #1252: Performance issues with `/gpu/cuda/magma` for non-tensor basis
JIT compilation of non-tensor basis code for `/gpu/cuda/magma` takes an excessive amount of time, especially for large values of `P`.
I believe this is due to the use of `#pragma unroll` in [`magma_common_nontensor.h`](https://github.com/CEED/libCEED/blob/main/include/ceed/jit-source/magma/magma_common_nontensor.h) for loops of length `P`, `Q`, or even `P*Q`. 

The results below, from a NVIDIA RTX 2070 Super with CUDA version 12.2, support this conclusion. I reproduced this slowdown on an NVIDIA A30 with CUDA version 12.1, although I do not have explicit timing results for that.

This issue was discussed with @nbeams on Zulip.

Current implementation (`CeedCompileMagma[in]` refers to the compilation for the `in`th non-tensor kernel instance):
```
P=8, Q=16
CeedCompileMagma[0] took 0 seconds 171 milliseconds
CeedCompileMagma[1] took 0 seconds 442 milliseconds
CeedCompileMagma[2] took 0 seconds 318 milliseconds
CeedCompileMagma[3] took 0 seconds 304 milliseconds
CeedCompileMagma[4] took 0 seconds 241 milliseconds
P=8, Q=16
CeedCompileMagma[0] took 0 seconds 175 milliseconds
CeedCompileMagma[1] took 0 seconds 437 milliseconds
CeedCompileMagma[2] took 0 seconds 318 milliseconds
CeedCompileMagma[3] took 0 seconds 298 milliseconds
CeedCompileMagma[4] took 0 seconds 242 milliseconds
P=64, Q=16
CeedCompileMagma[0] took 1 seconds 280 milliseconds
CeedCompileMagma[1] took 0 seconds 750 milliseconds
CeedCompileMagma[2] took 1 seconds 76 milliseconds
CeedCompileMagma[3] took 2 seconds 676 milliseconds
CeedCompileMagma[4] took 2 seconds 681 milliseconds
P=27, Q=16
CeedCompileMagma[0] took 0 seconds 472 milliseconds
CeedCompileMagma[1] took 0 seconds 666 milliseconds
CeedCompileMagma[2] took 0 seconds 687 milliseconds
CeedCompileMagma[3] took 0 seconds 595 milliseconds
CeedCompileMagma[4] took 0 seconds 977 milliseconds
P=8, Q=16
CeedCompileMagma[0] took 0 seconds 174 milliseconds
CeedCompileMagma[1] took 0 seconds 442 milliseconds
CeedCompileMagma[2] took 0 seconds 322 milliseconds
CeedCompileMagma[3] took 0 seconds 303 milliseconds
CeedCompileMagma[4] took 0 seconds 257 milliseconds
```

Without `#pragma unroll` in [`magma_common_nontensor.h`](https://github.com/CEED/libCEED/blob/main/include/ceed/jit-source/magma/magma_common_nontensor.h):
```
P=8, Q=16
CeedCompileMagma[0] took 0 seconds 169 milliseconds
CeedCompileMagma[1] took 0 seconds 560 milliseconds
CeedCompileMagma[2] took 0 seconds 408 milliseconds
CeedCompileMagma[3] took 0 seconds 364 milliseconds
CeedCompileMagma[4] took 0 seconds 242 milliseconds
P=8, Q=16
CeedCompileMagma[0] took 0 seconds 170 milliseconds
CeedCompileMagma[1] took 0 seconds 386 milliseconds
CeedCompileMagma[2] took 0 seconds 277 milliseconds
CeedCompileMagma[3] took 0 seconds 240 milliseconds
CeedCompileMagma[4] took 0 seconds 240 milliseconds
P=64, Q=16
CeedCompileMagma[0] took 0 seconds 426 milliseconds
CeedCompileMagma[1] took 0 seconds 437 milliseconds
CeedCompileMagma[2] took 0 seconds 408 milliseconds
CeedCompileMagma[3] took 0 seconds 509 milliseconds
CeedCompileMagma[4] took 0 seconds 501 milliseconds
P=27, Q=16
CeedCompileMagma[0] took 0 seconds 384 milliseconds
CeedCompileMagma[1] took 0 seconds 512 milliseconds
CeedCompileMagma[2] took 0 seconds 594 milliseconds
CeedCompileMagma[3] took 0 seconds 490 milliseconds
CeedCompileMagma[4] took 0 seconds 511 milliseconds
P=8, Q=16
CeedCompileMagma[0] took 0 seconds 170 milliseconds
CeedCompileMagma[1] took 0 seconds 382 milliseconds
CeedCompileMagma[2] took 0 seconds 271 milliseconds
CeedCompileMagma[3] took 0 seconds 239 milliseconds
CeedCompileMagma[4] took 0 seconds 238 milliseconds
```

sebastiangrimberg:
Add support for BasisCreateHdiv and BasisCreateHcurl, as well as CEED_EVAL_INTERP, CEED_EVAL_DIV, and CEED_EVAL_CURL for non-tensor elements in the `magma` backend. Follow-up to #1300.

~~This also resolves #1252 by compiling only the custom non-tensor MAGMA kernels when needed.\ (during first call to `CeedBasisApply`).~~ (EDIT: Merged as part of #1382)

~~Other included improvements to the `magma` backend~~ (EDIT: Merged as part of #1382):
- Renames files in the `magma` backend to match the other GPU backends in the library, along with other style formatting for consistency with other backends and the [style guide](https://libceed.org/en/latest/libCEEDdev/#style-guide) (variable casing, for example).
- Simplify some of the macro definitions in `ceed-magma.h` to eliminate duplication between tensor and non-tensor basis code.
- Add include guards to `jit-source/magma` files as done in #1299 for `cuda` and `hip` backends.
- Use JiT for the non-tensor weight kernels for the `CEED_EVAL_WEIGHT` in the `magma` backend (as for `cuda/ref` and `hip/ref`).


sebastiangrimberg:
Coverage comparisons (to ensure the tests are actually running):

From #1299:
Screenshot 2023-08-21 at 4 28 51 PM

From this PR:
Screenshot 2023-08-21 at 4 29 31 PM

backends/magma/ceed-magma-basis.c
```diff
@@ -257,401 +254,441 @@ CEED_INTERN "C"
...
-        CeedScalar    alpha = 1.0, beta = 0.0;
-
-        void *args[] = {&transA, &transB, &N, &alpha, &impl->dinterp, &P, &du, &K, &beta, &dv, &M};
-        CeedCallBackend(CeedRunKernelDimSharedMagma(ceed, *interp, grid, M, ntcol, 1, shmem, args));
-      } else {
-        if (tmode == CEED_TRANSPOSE)
-          magma_gemm_nontensor(MagmaNoTrans, MagmaNoTrans, P, nelem * ncomp, Q, 1.0, impl->dinterp, P, du, Q, 0.0, dv, P, data->queue);
-        else magma_gemm_nontensor(MagmaTrans, MagmaNoTrans, Q, nelem * ncomp, P, 1.0, impl->dinterp, P, du, P, 0.0, dv, Q, data->queue);
+  // Compile kernels for NB as needed (blocking size for N and NB from first call is used for all future calls)
+  if (!impl->module) {
```
nbeams, line 323: Hi @sebastiangrimberg, thanks for all your work on this.  I haven't looked through in detail yet, but I wanted to go ahead and comment on this part.  I don't think this behavior is what we want.  I agree that it is likely to not matter in a lot of practical use cases.  But technically, according to the libCEED API, a user could create one basis object to use with two very different operators -- e.g. what if one of them is only tied to a specific part of the local mesh, a region of interest where some quantity is evaluated, and the two operators therefore have different values of $N$ -- different enough that we would want to choose different tuning parameters for them?  Again, I'm not sure how common such a use case would be.  But I think we have to respect that the API allows for it.  "Locking in" a basis object's tuning with its first application could cause some weird/unexpected results for users (why is the performance of a similar operator so much better in this one application vs when I use it over here in this application?).  I suppose we could add documentation specifying that if you have such a use case in mind, you should create two separate basis objects to use, even if they are identical in basis type/size/dimension.  But that feels rather clunky to me, and like something we shouldn't force the user to take care of for us.

If we're concerned about compile time during basis creation, we could only compile each version of the kernel as needed by checking during the basis apply.  I know that's more complicated, but I think it's potentially important.  The QFunction already delays compilation until its first application (though of course it only ever needs to be compiled once, and wouldn't need to store multiple kernels).

sebastiangrimberg: Yeah I was worried if this might be a use case we want to support. In that case we can still do the lazy compilation and we will just allow for caching the different compiled kernels for each value of `N` instead of just one.

sebastiangrimberg: Hi @nbeams, the latest commit https://github.com/CEED/libCEED/pull/1301/commits/60805e1fa49636f8be12a4d108cd17c389a19df5 should resolve this. Basically we now allow caching of JiT kernels for each possible value of `N` and check if one exists during `CeedBasisApply`, else we RTC it. Let me know what you think and if this resolves the concern.

Note that the weight kernels are independent of `N` so we can just compile them once for all during basis construction, which is nice. But this requires storing a separate `CUmodule/hipModule_t`. I think that's fine though.

nbeams: Thanks! Have you tested this locally? I don't know if we have any tests that would capture the full behavior (not just the delay in compilation, but if we end up needing to compile 2 or more different instances throughout the lifetime of the basis object).

include/ceed/jit-source/magma/magma-basis-interp-nontensor.h
```diff
@@ -0,0 +1,199 @@
...
+
+#include "magma-common-nontensor.h"
+
+////////////////////////////////////////////////////////////////////////////////
+extern "C" __launch_bounds__(MAGMA_BASIS_BOUNDS(BASIS_Q, MAGMA_MAXTHREADS_1D)) __global__
+    void magma_interp_nontensor_n(int n, CeedScalar const* dA, int ldda, CeedScalar const* dB, int lddb, CeedScalar* dC, int lddc) {
+  MAGMA_DEVICE_SHARED(CeedScalar, shared_data);
+
+  const int tx      = threadIdx.x;
+  const int ty      = threadIdx.y;
```
nbeams, line 21: I find it confusing that the `interp` and `deriv` kernels are both in a file named `magma-basis-interp-nontensor.h`.  But perhaps you were already planning to change that?

At the moment, it looks like `interp` and `deriv` are identical except for the fact that they need to be compiled with different values for `QCOMP` and `NB`, no?  Were you thinking to refactor into a common templated function like you did for the `ref` backends, then have `interp` and `deriv` just instantiate the kernel with different values?

I know I said previously that I thought it would be good to have "scalar" and "vector" versions (i.e., a kernel where `QCOMP` is 1 and a kernel where it is > 1) so that the "scalar" one could avoid the `QCOMP` loop altogether.  That also allows us to use `=` instead of `+=` in the transpose mode application for the scalar kernel, when we know we only have one component.  Of course I don't know how much difference that would really make in the end.  Hopefully very little, if the compiler is smart...  

Also, I realize looking through the kernels again that that wouldn't affect the MAGMA code, anyway, so I guess there's no benefit at all to not rearranging into a common core function like you did for `ref`.  (And *if* there is any penalty in `ref`, well...it's `ref`.  So probably not a big deal.)

nbeams (commented):
So far this and #1300 are looking pretty good I think...do you still have any other major plans for them, or are you ready to move from draft to ready for review?

include/ceed/jit-source/magma/magma-basis-interp-nontensor.h
sebastiangrimberg: I'll attempt to respond in order:
1. Yeah I can rename that file. It seems like a better name would be `magma-basis-interp-deriv-nontensor.h`, right?
2. That's right `interp` and `deriv` are identical aside from the compile time constants. I can play around with combining them to reduce the code duplication.
3. I agree the performance difference of the loop with size = 1 shouldn't change much, since it's known at compile time. For the `+=` vs. `=` in both the `ref` and `magma` cases, let me see what I can do it would be nice to fix this.

sebastiangrimberg:
> So far this and #1300 are looking pretty good I think...do you still have any other major plans for them, or are you ready to move from draft to ready for review?

Thanks! No major plans, no. I'll address your comments here on the MAGMA basis backends but otherwise no changes are planned. I have only left them as draft since I have tested some amount locally (via MFEM) but not super extensively. But I'm happy to mark as ready for review if you are ready.

backends/magma/ceed-magma-basis.c
sebastiangrimberg: I have not tested this locally, as I'm in the same case of not having any tests for this type of behavior (hence my removing it on the first go around 😄 ).

sebastiangrimberg:
OK, marked as ready for review along with #1300 after addressing the comments [here](https://github.com/CEED/libCEED/pull/1301#discussion_r1313040104). To address them I:
- Refactored the MAGMA basis `interp` and `deriv` methods to share the same template function, similar to `ref`.
- Optimized the MAGMA transpose basis apply for `QCOMP = 1`.

No rush to review here, I'm plenty busy working on top of these changes and testing things further. But no planned changes on my part so happy to proceed with review.

sebastiangrimberg:
Rebased to resolve conflicts on `main` and this PR is ready for review. I'm unsure of the reason for the failing codecov test, but everything else is finalized from my end. @nbeams, I believe I've addressed your comments, but let me know if there are other changes you'd like to see. Thanks!

nbeams:
Overall, I think this looks good.  I modified t510 locally to confirm the lazy compilation works as intended when using the same basis on sizes belonging to different "N groups". 

I have run into something odd, though. I wanted to make sure the changes didn't significantly affect performance of the current basis modes (interp and grad), though I didn't expect they would. I ran some MFEM benchmarks for BP1 (mass matrix/interpolation) and BP3 (Poisson/gradient) on A100 and MI250X and compared `main` to this branch. Mostly, things looked fairly even, or with an overall advantage to the new code. But there was this odd outlier for the Poisson/gradient case with 4th order tetrahedrons on MI250X: 

[hip_magma_compare_new_mi250x_rocm55_bp3_speedup.pdf](https://github.com/CEED/libCEED/files/12677810/hip_magma_compare_new_mi250x_rocm55_bp3_speedup.pdf)

This is speedup of new vs old. (I included one non-RTC order (5th) just for comparison).

This drop off the cliff for the 4th order basis functions appears to be reproducible. From rocprof output, the culprit is the non-transpose basis kernel, which suddenly starts dominating the overall GPU time as we move from the last "good" point to the first "bad" one. I'm surprised by this, though, because based on the tuning, both problem sizes should use the same NB, despite belonging to different groups, meaning the compiled kernel should be the same. However, the first "bad" point is the first one to be in the biggest N group, so maybe it's as simple as needing to choose a different NB for this group now.

sebastiangrimberg:
Thanks for the thorough check @nbeams. Let me dig into it and see what I can find.

backends/magma/ceed-magma-gemm-selector.cpp
```diff
@@ -92,57 +80,50 @@ void gemm_selector(int gpu_arch, char precision, char transA, int m, int n, int
...
 }
+#endif
 
 ////////////////////////////////////////////////////////////////////////////////
-CeedInt nontensor_rtc_get_nb(int gpu_arch, char precision, CeedEvalMode emode, CeedTransposeMode tmode, int P_, int N, int Q_) {
-  CeedInt P  = (tmode == CEED_TRANSPOSE) ? P_ : Q_;
-  CeedInt Q  = (tmode == CEED_TRANSPOSE) ? Q_ : P_;
+CeedInt nontensor_rtc_get_nb(int gpu_arch, char transA, int qcomp, int P_, int N, int Q_) {
+  CeedInt P  = (transA == 'n') ? P_ : Q_;
+  CeedInt Q  = (transA == 'n') ? Q_ : P_;
```
sebastiangrimberg, line 116: Welp, this is certainly a difference from the previous behavior. Let me fix that.

sebastiangrimberg: Fixed this one in [511d170](https://github.com/CEED/libCEED/pull/1301/commits/511d170bb98bfe27abfe974249cd74d63808d730).

nbeams:
An important note for everything below is that I don't think we have any 3D non-tensor tests in the test suite (topology always TRIANGLE and not TET), which could explain why the CI still passes. We may want to close this gap in our tests... 

So unfortunately both https://github.com/CEED/libCEED/pull/1301/commits/25324f38915d5b68bc6ed49d13158dae51ebb3b0 and https://github.com/CEED/libCEED/pull/1301/commits/511d170bb98bfe27abfe974249cd74d63808d730 seem to cause correctness issues in the 3D MFEM benchmark. With the latest version of this branch, I was often getting very obvious errors from MFEM, with the solver balking because the operator wasn't positive definite. When I rolled back just the latest commit (changes to `__syncthreads`), the obvious errors went away, but if I looked more closely at the output, I saw that the tests still weren't converging properly -- actually diverging, with the residual growing.

The spot you noted above (in the code for getting `NB` for the RTC kernels) is indeed the culprit of the performance regression, it's just that something in 511d170bb98bfe27abfe974249cd74d63808d730 is causing some other issue.  

I tried rolling back before 511d170 and simply changing the `n` to `t`, as in the original check:
```
CeedInt nontensor_rtc_get_nb(int gpu_arch, char transA, int qcomp, int P_, int N, int Q_) {
  CeedInt P  = (transA == 't') ? P_ : Q_;
  CeedInt Q  = (transA == 't') ? Q_ : P_;
  CeedInt NB = 1;
```

Which fixes the convergence issues, and also helps a lot with the performance regression:
[hip_magma_compare_new_mi250x_rocm55_bp3_speedup.pdf](https://github.com/CEED/libCEED/files/12690161/hip_magma_compare_new_mi250x_rocm55_bp3_speedup.pdf)

sebastiangrimberg:
OK great news! Sorry I was really just trying anything to come up with reasons for the performance regression so understand these hacks might have broken correctness. I will revert them and also figure out what else in 511d170 is breaking correctness. Sorry and thanks so much for your help @nbeams!

nbeams (commented):
I know this PR isn't the source of this issue, but I just noticed it, and I thought maybe it'd be a good place to fix it? 

We end up trying to add `(ROCM_DIR)/include/include`...

Hmm, my attempt to comment on the specific spot in the code didn't work for some reason.  I'm talking about the part of the Makefile where we set `HIP_INC_DIR_MAGMA=$(ROCM_DIR)/include` and then try to include `-I$(HIP_INC_DIR_MAGMA)/include`

sebastiangrimberg:
> I know this PR isn't the source of this issue, but I just noticed it, and I thought maybe it'd be a good place to fix it?
> 
> We end up trying to add `(ROCM_DIR)/include/include`...
> 
> Edit: Hmm, my attempt to comment on the specific spot in the code didn't work for some reason. I'm talking about the part of the Makefile where we set `HIP_INC_DIR_MAGMA=$(ROCM_DIR)/include` and then try to include `-I$(HIP_INC_DIR_MAGMA)/include`

Along with this, is there ever a case where `HIP_LIB_DIR_MAGMA` is not the same as `HIP_LIB_DIR`?

nbeams:
Probably not...I'd imagine it's a remnant from earlier days of libCEED HIP development or maybe when we made the switch to `ROCM_DIR` instead of `HIP_DIR`?  Good call to streamline it, I think.

nbeams:
Unfortunately I'm still seeing incorrect results with the latest version. I tried reverting just [bc23d45](https://github.com/CEED/libCEED/pull/1301/commits/bc23d45523e45723255f272849f100e467f1e2a9) and still got an error, just a different one (nan instead of non-positive-definite error).

sebastiangrimberg:
Does reverting both [bc23d45](https://github.com/CEED/libCEED/pull/1301/commits/bc23d45523e45723255f272849f100e467f1e2a9) and [25324f3](https://github.com/CEED/libCEED/pull/1301/commits/25324f38915d5b68bc6ed49d13158dae51ebb3b0) work? I'm still playing around with the `__syncthreads` calls, this needs more testing on my end.

sebastiangrimberg:
Hi @nbeams, [0c6524f](https://github.com/CEED/libCEED/pull/1301/commits/0c6524f3e23713d2b943bd6cd4d299d03681f845) should fix the correctness issue you were seeing. Once OK, I will rebase this branch and squash out all the intermediate commits for final testing/review.

nbeams:
The correctness issue is indeed fixed, and A100 performance looks fine compared to `main`. Unfortunately, I'm still seeing MI250X performance regression.  I'm wondering if the NB is still somehow not being set correctly, as by my quick calculations, most of the drops happen at an "N group" boundary, and the third order outlier on the mass matrix test (bp1) is the only one belonging to its group, for example...

[BP1 comparison MI250X](https://github.com/CEED/libCEED/files/12704826/hip_magma_compare_new_mi250x_rocm55_bp1_speedup.pdf)
[BP3 comparison MI250X](https://github.com/CEED/libCEED/files/12704827/hip_magma_compare_new_mi250x_rocm55_bp3_speedup.pdf)

sebastiangrimberg:
Humph! Ok I'll see if my "fixes" are actually breaking prior fixes.

sebastiangrimberg:
Hi @nbeams, a couple of questions here as I continue to try to debug (without access to any reproducing example/MI250X nodes, this is a bit of a challenge):

- Looking at the bp3 data, I think the fall off around 1e6 dofs for orders 2-4 is still worrisome. But I'm curious the parameters you're using for the quadrature order. For the PETSc BP3 in `examples/petsc`, for example, degree = 4 with q_extra = 0 yields P = 35 (as expected for a tetrahedron) and Q = 46, thus it shouldn't even be using the RTC kernels. Are you running with something different in your MFEM example? It looks like the MFEM BPs in the libCEED repo at `examples/mfem` only support tensor product elements, so some detail on your benchmark would be useful to try to reproduce (or if you could reproduce the same slowdown in the PETSc benchmark, that would also be helpful).
- Maybe for the further debugging, could you give the exact parameters for the bad p = 3 BP1 example? For example, how many elements, what quadrature order? That way I can try to exactly compare the code in this PR vs. main to see if something is changing unexpectedly.

nbeams:
The quadrature is just the MFEM standard choice for tetrahedrons. MFEM has special "hard-coded" quadrature rules for lower-order tets in `fem/intrules.cpp`. At the top of that file, in a comment, it lists the reference for the rules: https://nines.cs.kuleuven.be/research/ecf/ecf.html.  I looked into it a few years ago and according to my notes, Q is less than P for up to and including 5th order tets in MFEM, then it switches to Grundmann-Moller quadrature and Q > P.

I am using a test from my fork of the CEED benchmarks repo: https://github.com/nbeams/benchmarks/blob/master/tests/mfem_ex1_hip/ex1.cpp
If you already have MFEM built with libCEED, it should be simple to build/run this test as a standalone without worrying about the rest of the benchmarks setup. (There is also a version for MFEM built without MPI: https://github.com/nbeams/benchmarks/blob/master/tests/mfem_ex1_serial_hip/ex1.cpp.)

For the test, switch to non-tensor elements with the argument `-e 1`. It just uses MFEM's built-in function to split hexes into 6 tets (starting from the hex mesh that is used in the tensor benchmarks).  You can toggle between mass and Poisson with the `-p` flag: `-p 0` is mass/BP1, `-p 1` is Poisson/BP3.

The "bad" point for BP1 was from this call: `./ex1 -p 0 -o 3 -l 16 -d ceed-hip:/gpu/hip/magma -e 1` 
It has 393,216 elements in the mesh.

jedbrown:
> Q is less than P for up to and including 5th order tets in MFEM

This is interesting since it implies the mass matrix for a single element is singular. Or is this only for differential forms that can be integrated exactly with that quadrature?

nbeams:
Oh yeah, I went back through my notes and those sizes are for the Poisson operator specifically. For mass, e.g., first order tets have 4 basis nodes and 4 quadrature nodes, second order have 10 and 11, ...

sebastiangrimberg:
OK, I was able to build your MFEM-based benchmarks. I thought I might be able to target the AMD bug by forcing use of the MI250X data for NB (even on my V100 setup) and checking if any of the optimized block sizes were different. But for both BP1 and BP3 for your MFEM  benchmark, everything appears the same and I see similar performance between `main` and this PR. That is, for BP1, I have Q = 24, N = 393216 (n_array[iN] = 512000), P = 20 -> NB = 6 for the non-transpose operation and NB = 10 for the transpose one.

Is there a chance your most recent benchmark is not reproducible? This is very hard to debug on my end not having access to the hardware that seems to show the issue, so I'm a relying on you a bit much for help I realize. But hoping that maybe this could have been a one time issue?

nbeams:
Every time I've tried to reproduce one of the noticeable "blips", I've been able to. So I do think there is something going on. If the NB is definitely correct now, then it might be that we should change the NB for some sizes. From tests in the past, I've noticed that seemingly small changes in kernel code can result in big changes in assembly output from the HIP compiler (as represented in terms of reported theoretical occupancy, register spilling, etc.). It's possible that something like this has happened with the kernels, and what used to be a good choice for NB is now one of the worse choices...I can do some digging.

sebastiangrimberg:
Yeah, I'm a bit wary of requiring changes to the tuning though that could be it since I know what you are saying about small kernel changes leading to very different assembly. But any changes to the kernels to generalize for H(curl) and H(div) were really just syntactical I believe. Especially in the case of the Poisson tests, the kernel for the gradient is really unchanged I believe. That said, if rerunning the tuning on the new code in this PR does allow us to get better performance against `main`, then I'd be fine with that and I guess it would just confirm your suspicions.

nbeams:
I can reproduce the performance regression on a different machine with MI210 GPUs. `rocprof` shows that it's the transpose mode basis kernel that's the problem for the `-o 3 -l 16 -p 0` case. I tried making a standalone test with just the basis kernels, but haven't gotten anything useful from it yet...

Unrelated, but another thing I was wondering was about the `interp` kernel's use of `magma_basis_nontensor_device_[n/t]`. If we have `q_comp` > 1, like in a vector-valued basis functions, won't we also need to advance the `B` (input vector) pointers inside the loop? Since input and output (B and C) would have the same total size, right?

sebastiangrimberg:
I've just pushed a commit reverting the loop unrolling which may have been causing the AMD performance issue (fingers crossed?).

I have validated two test of your MFEM test cases that all blockings are identical for both the offline A100 and MI250X tunings using my V100 setup:

```
./ex1 -o 3 -l 16 -p 0 -e 1 --device ceed-cuda:/gpu/cuda/magma
./ex1 -o 4 -l 16 -p 1 -e 1 --device ceed-cuda:/gpu/cuda/magma
```

Both of these examples had bad performance drops previously in your tests, but show no difference in `NB` or in the resulting performance on V100 to me.

nbeams:
With latest changes, still seeing performance regression on the MI210. I looked at the compiler output (assembly) for the transpose kernel with `-p 0 -o 3 -l 16` and `-p 0 -o 3 -l 17` and two things stood out: the `-l16` case shows register spilling, and also it looks like the `addmul_rAsBrC_1D_nosync` function is not getting inlined for some reason. The `-l17` case has no spilling, and the `addmul` is inlined from what I can tell.

Here are the definitions used in compilation:

`-l 16`:
```
#define BASIS_Q_COMP_INTERP 1
#define BASIS_Q_COMP_DERIV 3
#define BASIS_P 20
#define BASIS_Q 24
#define BASIS_NB_INTERP_N 6
#define BASIS_NB_INTERP_T 10
```

`-l 17`:
```
#define BASIS_Q_COMP_INTERP 1
#define BASIS_Q_COMP_DERIV 3
#define BASIS_P 20
#define BASIS_Q 24
#define BASIS_NB_INTERP_N 5
#define BASIS_NB_INTERP_T 16
```

Not sure what exactly about the changes causes the compiler to do this...but it would explain the sudden drop in performance.

> Likewise, is there a reason why the tuning neglects the smallest two N data points (1024 and 5120)?

I believe it was mostly due to concerns about compile time, with smaller sizes seen as less important in terms of what users were likely to want (since the GPU backends generally need more elements per GPU to make efficient use of the hardware). However, the change to lazy compilation removes this conern, so we could consider using all the points again (or different ones). 

We don't currently have a standard tuning script ready, but it would be a good idea to have one. I have some benchmarks I have been using, but they should be streamlined to make it easier to test changes in the kernels (short version: they don't currently use RTC because I was mostly trying to get a comparison between CUDA/HIP and SYCL versions, and there are some issues with RTC and SYCL).

jeremylt:
Today is probably a good cutoff for merges before the release. Which of #1300, #1301, #1302 will be ready? Its ok if we wait on these to avoid performance regressions.

sebastiangrimberg:
I would be in favor of postponing all three, since they should realistically all go together and I don't think it makes sense to force them in at this point. Anyone who is going to be looking for this functionality probably will be fine building from `main` instead of waiting for v0.13 once they are merged.

@nbeams, this is again very helpful info. I can confirm we are moving between the following tuned parameters when switching from `-l 16` to `-l 17` on AMD for the transpose interpolation:

```
...
{21, 512000,  16, 10},
{21, 1024000, 16, 16},
...
```

Maybe the initial kernel was just very sensitive the way it was written and we were just getting lucky that `hipcc` was inlining `addmul_rAsBrC_1D_nosync` and we were avoiding register spilling.

I'm going to try a fix here: `addmul_rAsBrC_1D_nosync` does a prefetching of the shared memory `sB` to registers `rB` which leads to an allocation of another `Q` scalars on the stack. If we just do the multiplication from shared memory, as is done for the shared memory kernels, for example, we are going to decrease register pressure I believe. I'm going to try this now on my V100 setup to ensure it doesn't negatively affect performance and then I'll push it.

It's strange to my relatively unexperienced brain the relationship between increasing `NB` and somehow using fewer registers. That is `BASIS_NB_INTERP_T` goes from 10 -> 16 which is stored by each thread in `rC`. But maybe I'm thinking about it wrong.

jedbrown:
There's also `__attribute__((flatten))` as in `CEED_QFUNCTION_ATTR`.

sebastiangrimberg:
> There's also `__attribute__((flatten))` as in `CEED_QFUNCTION_ATTR`.

It looks like `cuda-jit.h` overrides `CEED_QFUNCTION_HELPER` as just `inline __device__`, so we never actually use `__attribute__((flatten))` for CUDA backends. Is that correct?

jedbrown:
Sure looks that way. I think the compiler is generally more aggressive about inlining in kernels, but perhaps we should have more of these attributes as we needed for CPUs. (For CPUs, failing to inline also implies failure to vectorize.)

sebastiangrimberg:
This might be reading into my own V100 debugging too much, but doing the local `rC = rA * rB` multiplication vs. using `sB` (shared memory), as in the last commit, does make the function "easier" to inline for `nvrtc` at least. I'm adding `"--optimization-info=inline"` to the `nvrtc` options and I see a cost of `-15895` for `addmul_rAsBrC_1D_nosync` previously versus `-15945` after the change. This is marginal relative to the threshold of 1625, but maybe indicates this change might fix the issue for `hipcc`?

sebastiangrimberg:
Hi @nbeams! OK, kind of a big update here. I added an automatic tuning framework in https://github.com/CEED/libCEED/pull/1366 and I'm hoping we can use that to regenerate the NB data from the changes in this PR. Following that we should be able to profile against `main` and, if all works out, avoid the performance regressions and hopefully even see some improvements.

This issue is that I currently only have access to a V100 setup. It shouldn't be too difficult for anyone with access to the various targets to rerun the autotuning and push the updated data to https://github.com/CEED/libCEED/pull/1366. I'm hoping you might be able to do this at least for your MI250X and MI100 machines. Do you think this sounds reasonable? Alternatively maybe @jeremylt can help with the machines he has access to or can delegate accordingly. Let me know what you both think and if I can help in any way. I'm hopeful that retuning the NB for the slight changes to the kernels will fix things, but we will see.

jeremylt:
I'm not sure which all machines our group has access to, but I wouldn't be surprised if we have access to newer AMD GPUs. I mainly use our development server, which has only a Radeon VII.

jedbrown:
We have access to MI250X (at OLCF and LLNL) and MI100 via RMACC.

sebastiangrimberg:
> We have access to MI250X (at OLCF and LLNL) and MI100 via RMACC.

OK! The tuning is very easy to run, so I would just need a volunteer with access to go through the following (for example):

```
git clone git@github.com:CEED/libCEED.git
git checkout sjg/magma-rtc-autotuning
cd libCEED/backends/magma/tuning
python generate_tuning.py -ceed /gpu/hip/magma -arch mi250x -max-nb 32 -build-cmd "make MAGMA_DIR="
```

The `-build-cmd` is just a command that will be called to rebuild the library for various test `NB` values. If you just do a `make config <ARGS>` in the `libCEED/` source directory, I think this just needs to be `-build-cmd "make"`, but in my testing I just have a bash script `build.sh` which calls `make <ARGS>` and pass `-build-cmd build.sh`.

This generates a file at `backends/magma/tuning/mi250x_rtc.h` which could just be be committed and pushed.

nbeams:
The framework in #1366 looks like it will be an excellent addition -- thanks @sebastiangrimberg!!  I can probably run these on MI250X and A100 early next week (unless someone beats me to it before then).  One thing I wanted to clarify to make sure we run the benchmarks the same way you did the V100 ones: in the README on #1366, you have this note: 

> Note that in order for the benchmarks to make sense for `magma` backends, the `ceed_magma_queue_sync` in `ceed-magma.h` should be set to `cudaDeviceSynchronize()` or `hipDeviceSynchronize()`.

Is that what you did when you ran your tests locally? Just to be clear, since you didn't mention it in the instructions above. Another option would be to add one device synchronization call prior to stopping the clock in the tester itself (e.g. after the loop at line 85 in `tuning.cpp`). Since we are just "self-comparing" for the best time among `NB` values I doubt it makes much difference, but why not be as consistent as possible, I figure?

sebastiangrimberg:
Thanks @nbeams!

To answer your Q, the Python driver will add `cudaDeviceSynchronize()`/`hipDeviceSynchronize()` automatically, so you don't have to do it yourself. But I made that remark in the README in case people are running the `tuning` benchmark aside from the Python driver in which case they need to make the addition themselves and rebuild the library.

The reason for not adding the device synchronization in `tuning.cpp` directly was I didn't want to have to add an explicit CUDA or HIP dependency to this C++ file. If we had a wrapper for `cuda/hipDeviceSynchronize()` exposed by libCEED then we could just call that, but I wasn't sure if that's something which would be desirable to add.

nbeams:
Ok, here are the updated benchmarks for MI250X using the tuning data that I added to the `sjg/magma-rtc-autotuning` branch. For the BP1/mass plot, I've added numbers indicating which "N group" an x-axis location belongs to, so we see that the lowest one is again the only point in its group. However, its performance is now around 85% of main -- still lower than I was expecting, but not like the drop seen in previous versions. Do you think it would be worth trying the autotuning for the version before 239b46e? 

[BP1 speedup vs main (with N groups marked)](https://github.com/CEED/libCEED/files/12860743/hip_magma_compare_new_mi250x_rocm55_bp1_speedup_with_groups.pdf)
[BP3 speedup vs main](https://github.com/CEED/libCEED/files/12860746/hip_magma_compare_new_mi250x_rocm55_bp3_speedup.pdf)

sebastiangrimberg:
That seems worth a try! See f66de507. I did not rebase `sjg/magma-rtc-autotuning` onto this branch so it might be easiest for you to just cherry-pick it into your local copy of `sjg/magma-rtc-autotuning` to regenerate the tuning and rerun the benchmarks.

Thanks a lot for this help! Seems like we are making progress.

sebastiangrimberg:
Hi @nbeams, something else I noticed that is contributing here and we might be able to fix. MFEM's default integration rule for a `MassIntegrator` (BP 1) on a linear tet mesh uses order `2*p`. The MFEM integration rules thus use 4, 11, 24, and 43 quadrature points for orders 1-3. This is not what's in the training set from https://github.com/CEED/libCEED/pull/1366 (which was just copied from what's on `main`). For `DiffusionIntegrator` (BP 3), the quadrature order is `2*p-2` and thus uses 1, 4, 11, and 24 points, which is what's in the training set.

Since we have the automated tuning framework now. I've just added these training points to the NB calculation in https://github.com/CEED/libCEED/pull/1366, and regenerated the tuning data for v100. You can probably do the same for mi250x and hopefully we see some improvement. Also this branch now includes the original multiplication kernel where we cache shared memory to registers before doing the multiplication, so that is no longer a difference from main.

Thanks as always!

nbeams:
Updated MI250X benchmarks with the latest `sjg/magma-rtc-autotuning`, with the tuning that I added in 6138f1682a213d2e1e9db165a70a7b239e7f97bf:

[BP1, with N groups](https://github.com/CEED/libCEED/files/12899704/hip_magma_compare_new_mi250x_rocm55_bp1_speedup_with_groups.pdf)
[BP3](https://github.com/CEED/libCEED/files/12899705/hip_magma_compare_new_mi250x_rocm55_bp3_speedup.pdf)

BP1 is all sorted out, I'd say. Not quite sure what's with the consistent slight downward trend for larger sizes of BP3, but they are all > 90% of `main` now, so that's good. I just re-ran the BP3 test on `main` to make sure there wasn't anything weird with my original data collection for that, but it looks like the two runs match.

sebastiangrimberg:
OK, that's good at least. Still some drop in performance on BP 1 for the 3, 4, 5 blocks for p = 3 which is odd but the better tuning seems to just have shifted the whole curve up.

For BP3, I was indeed worried about that trend too. At this point my only next step would be to try to separate this PR into a few separate ones (maybe first tuning, then address kernel RTC issue, followed by H(curl)/H(div) support?) so we could test each individually. I don't see what could be hurting BP3 performance at large N. Should I do these separation so we can investigate further or should we just leave it? @jeremylt happy for your input here as well.

nbeams:
I would also tag in @abdelfattah83 for his opinion

abdelfattah83:
The new results for BP3 are within 90% of the original ones, so that shouldn't stand against merging the PR, IMHO. I will leave it to you to decide whether to break it down to smaller PR's for further investigation.

sebastiangrimberg:
I have opened https://github.com/CEED/libCEED/pull/1382 to be reviewed/merged first in an effort to reduce the size of this PR.

sebastiangrimberg:
Now that https://github.com/CEED/libCEED/pull/1382 is merged, this PR is much smaller and more digestable. Should be good for a final review and I propose we address the tuning separately in the follow-up https://github.com/CEED/libCEED/pull/1366. This can be reviewed alongside https://github.com/CEED/libCEED/pull/1300.

sebastiangrimberg:
In [cb71a19](https://github.com/CEED/libCEED/pull/1301/commits/cb71a19ce54253cf0620c86158d90b87d0e5cfc0), I added one performance change instead of breaking it out into another PR. The transpose kernel can be aided by using shared-memory to load A from global memory into registers, similar to the non-transpose kernel. This has big effect on the transpose gradient application for "large" P, Q sizes.

No shared-memory (after NB tuning, listed as `P`, `Q`, `N`, `q_comp`, `transpose`, `MFLOPS`) :

Screenshot 2023-10-18 at 6 00 16 PM

With shared-memory (after NB tuning):

Screenshot 2023-10-18 at 5 59 52 PM

With the change, the performance gap between the transpose and non-transpose application is much smaller. For large `N`, for example, I see 1.34 TFLOPS -> 1.79 TFLOPS (non-transpose is 2.05 and 2.11 in each run).

jedbrown:
That's great! Is the remaining discrepancy related to atomic writes? Would there potentially be benefit to explicitly transposing `A` so the transpose basis look like a non-transpose?

sebastiangrimberg:
I'm not really sure. I don't think it's atomic writes, the basis application writes to global memory the same way in both the transpose and non-transpose cases (for `q_comp = 1`, at least). The reads from global to shared memory of A are identical in the non-transpose and transpose cases, the only difference is shared memory to registers. Storing A^T explicitly would only help to avoid potential bank conflicts in the transpose reads from shared memory to registers.

I think more likely the performance gap is just given by the occupancy (this is said without any actual detailed profiling data). For the transpose case we are launching thread blocks of dimension 35 x 3, but for the non-transpose we are launching 24 x 5 (for the data above, for example). This is bound to cause differences in performance is my guess, regardless of the actual kernel code. The case for interpolation (`q_comp` = 1) is a good reference here since the kernel code is almost entirely identical aside from the read from shared A -> register A vs. A^T and the launched block sizes.

jedbrown:
That's reassuring to see the transpose is faster for some cases. I think we can hypothesize that it's attributable to packing/occupancy for now and proceed with this PR. I'll leave it to @abdelfattah83 or @nbeams for final approval.

sebastiangrimberg:
Hey @jedbrown, I couldn't help myself so dug into this a bit further. The transpose kernels zero out the vector `V` in a call to `magmablas_[d/s]laset` before applying the basis while the non-transpose don't:

```
  // Clear v for transpose operation
  if (t_mode == CEED_TRANSPOSE) {
    CeedSize length;

    CeedCallBackend(CeedVectorGetLength(v, &length));
    if (CEED_SCALAR_TYPE == CEED_SCALAR_FP32) {
      magmablas_slaset(MagmaFull, length, 1, 0.0, 0.0, (float *)d_v, length, data->queue);
    } else {
      magmablas_dlaset(MagmaFull, length, 1, 0.0, 0.0, (double *)d_v, length, data->queue);
    }
    ceed_magma_queue_sync(data->queue);
  }
```

I'm not sure this is really required by the non-tensor kernel since the writes to V are not accumulated for the transpose operation. But correctness aside for now, here's the performance with the zeroing out:

image

Here it is without:

image

So, I feel pretty confident in saying that was the culprit for the performance difference between the transpose and non-transpose basis application. I'll look some more into if we can just avoid setting V = 0 here for correctness.

jedbrown:
That's great!

sebastiangrimberg:
@jeremylt @nbeams

I’m not sure the timeline for tagging v0.12, but this PR as well as #1300 and #1366 are in a state that is ready (pending final approval from @nbeams and a final run for #1366). If you’d like to include these in the release I think that can be doable.

jedbrown:
I'm not opposed to making a new release in the near future with this and new particle operators.

nbeams:
Ack, I hate to be the bearer of bad news, but it seems that we have encountered accuracy problems again (which do not show up in the unit tests). When running the 3D mass and diffusion benchmarks with this branch on MI250X, I see `nan`s or `"The operator is not positive definite"` errors. For diffusion, the trouble seems to start with order 2, while for mass/interp, it's right away at order 1.
@sebastiangrimberg have you tried my `ex1.cpp` linked above (from earlier comment) on the latest version of this branch? Can you confirm whether you see this same behavior on your test setup?

sebastiangrimberg:
Shoot, definitely no need to apologize. I haven't been running those tests so will go test it now, probably a simple bug maybe associated with the transpose or #1385.

sebastiangrimberg:
I got it: 1a0eda08. Fat finger from commit to use shared-memory for transpose operation.

From ex1 (`./ex1 -o 3 -l 16 -p 0 -e 1`):
- Prev:
```
Number of finite element unknowns: 1815937
   Iteration :   0  (B r, r) = 1.15752e-06 ...
   Iteration :  53  (B r, r) = 8.25646e-31
Average reduction factor = 0.591835

Total CG time:    0.110275 (0.110275) sec.
Time per CG step: 0.00208066 (0.00208066) sec.

"DOFs/sec" in CG: 872.77 (872.77) million.
```
- New:
```
Number of finite element unknowns: 1815937
   Iteration :   0  (B r, r) = 1.15752e-06 ...
   Iteration :  53  (B r, r) = 8.25646e-31
Average reduction factor = 0.591835

Total CG time:    0.104001 (0.104001) sec.
Time per CG step: 0.00196228 (0.00196228) sec.

"DOFs/sec" in CG: 925.422 (925.422) million.
```

include/ceed/jit-source/magma/magma-basis-interp-deriv-nontensor.h
```diff
@@ -0,0 +1,185 @@
...
+    CeedScalar rA[Q];
+
+    // read A using all threads
+    read_A_notrans_g2r_1D_nosync<CeedScalar, P, Q, MAGMA_BASIS_NTCOL(P, MAGMA_MAXTHREADS_1D)>(tx, ty, dA, sA, rA);
+    __syncthreads();
+
+    // read B
+    if (id < nblocks) {
+      read_B_g2s_1D_nosync<CeedScalar, P, Q, NB>(tx, myn, dB, sB);
+    }
```
nbeams, line 128: Same comment as above.

nbeams (commented):
I can confirm the correctness issue is fixed. The performance on MI250X was as I anticipated: a little better overall than the initial tests from much earlier in this PR history, but with somewhat severe performance regressions for some sizes, which we expect will be mostly or entirely fixed by #1366 when it is finished with the latest kernels.

I just have some minor comments, and then I think this will be ready. Due to the known performance regressions, I guess we should make it a priority to merge #1366 as soon as possible after this gets merged to `main`. Or, of course, we could merge #1366 into this branch and take care of the new tuning here too, but this PR is already pretty big, and I think it makes sense to have #1366 as its own thing. So I would be fine with either plan.

include/ceed/jit-source/magma/magma-basis-interp-deriv-nontensor.h
```diff
@@ -0,0 +1,185 @@
...
+  const int id      = blockIdx.x * blockDim.y + ty;
+  const int nblocks = (n + NB - 1) / NB;
+  const int myn     = min(NB, n - id * NB);
+
+  dB += id * P * NB;
+  dC += id * Q * NB;
+
+  // A is P x Q
+  CeedScalar *sB = shared_data + ty * P * NB;
+  CeedScalar *sA = shared_data + blockDim.y * P * NB;
```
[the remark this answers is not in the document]
sebastiangrimberg: `sB` is the first `P * NB * blockdim.Y` entries, where depending on the thread's y-index it points to a chunk of `P * NB` in that array.
`sA` points on all threads to the last `P * Q` block of shared memory.

It could just as easily be the opposite order, where the first `P * Q` entries are for `sA` and then after comes `sB`. Not sure if this has performance implications but as written it's in order of access.

include/ceed/jit-source/magma/magma-basis-interp-deriv-nontensor.h
```diff
@@ -0,0 +1,185 @@
...
+
+  // read B once for all C's
+  if (id < nblocks) {
+    read_B_g2s_1D_nosync<CeedScalar, Q, P, NB>(tx, myn, dB, sB);
+  }
+
```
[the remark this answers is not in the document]
sebastiangrimberg: I don't think we can return early. The `A` reads in both cases use all `blockDim.x * blockDim.y` threads, not just the ones which are handling valid parts of `B`. This is the reason for the non-transpose kernels using if checks instead of early returns and now is needed for both cases since shared memory is also being used for the `A` reads for the transpose kernels.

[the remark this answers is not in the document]
nbeams: Ah right, I was forgetting that `read_A_notrans` was changed and has a new thread config. Never mind. I think we can wait to try another specialized version until we see if it actually affects mass operator performance in a noticeable way after updating the tuning.

include/ceed/jit-source/magma/magma-basis-interp-deriv-nontensor.h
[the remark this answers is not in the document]
nbeams: Ah, makes sense.

nbeams (approved):

PR #1302, opened 2023-08-21
GPU element restriction for `CEED_RESTRICTION_ORIENTED` and `CEED_RESTRICTION_CURL_ORIENTED`

sebastiangrimberg:
NOTE: Based on #1316

Follows up #1300 and #1301 to complete H(div) and H(curl) functionality on GPU backends. Implements diagonal assembly and full assembly for the `cuda/ref` and `hip/ref` backends when using H(div) or H(curl) finite element spaces, while also following up #1316 with support for rectangular operator full assembly.

~~This currently does not include deterministic element restriction transpose operations for the GPU backends when the restrictions are created by `ElemRestrictionCreateOriented` or `CreateCurlOriented`. The non-determistic kernels using `atomicAdd` are easier to write. If this is a sticking point I guess we can go ahead and add deterministic variants for the transpose operations for these restrictions. I believe all that's necessary is correctly permuting the `orients` and `curl_orients` arrays for access similar to `offsets` during construction, but I am not fully certain on this and so would welcome any input here on how to go about it/if this should be added or not.~~ UPDATE 11/9: This is now added and `/gpu/[cuda,hip]/magma/det` is implemented for all restriction types.

### DIFF
--- a/backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
+++ b/backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
@@ -1073,28 +1073,14 @@ static int CeedSingleOperatorAssembleSetup_Sycl(CeedOperator op) {
   // Kernel setup
   int elems_per_block   = 1;
   asmb->elems_per_block = elems_per_block;
-  CeedInt block_size    = elem_size * elem_size * elems_per_block;
-
-  /* CeedInt maxThreadsPerBlock = sycl_data->sycl_device.get_info<sycl::info::device::max_work_group_size>();
-  bool    fallback           = block_size > maxThreadsPerBlock;
-  asmb->fallback             = fallback;
-  if (fallback) {
-    // Use fallback kernel with 1D threadblock
-    block_size         = elem_size * elems_per_block;
-    asmb->block_size_x = elem_size;
-    asmb->block_size_y = 1;
-  } else {  // Use kernel with 2D threadblock
-    asmb->block_size_x = elem_size;
-    asmb->block_size_y = elem_size;
-  }*/
-  asmb->block_size_x   = elem_size;
-  asmb->block_size_y   = elem_size;
-  asmb->num_e_mode_in  = num_e_mode_in;
-  asmb->num_e_mode_out = num_e_mode_out;
-  asmb->num_qpts       = num_qpts;
-  asmb->num_nodes      = elem_size;
-  asmb->block_size     = block_size;
-  asmb->num_comp       = num_comp;
+  asmb->block_size_x    = elem_size;
+  asmb->block_size_y    = elem_size;
+  asmb->num_e_mode_in   = num_e_mode_in;
+  asmb->num_e_mode_out  = num_e_mode_out;
+  asmb->num_qpts        = num_qpts;
+  asmb->num_nodes       = elem_size;
+  asmb->block_size      = elem_size * elem_size * elems_per_block;
+  asmb->num_comp        = num_comp;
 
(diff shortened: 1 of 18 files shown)


sebastiangrimberg:
Coverage comparisons (to ensure the tests are actually running):

From #1301:
Screenshot 2023-08-21 at 4 29 31 PM

From this PR:
Screenshot 2023-08-21 at 4 29 38 PM

jeremylt:
As a heads up, we anticipating releasing libCEED v0.12 in Sept so we can make a new release of Ratel by the end of Sept.

sebastiangrimberg:
> As a heads up, we anticipating releasing libCEED v0.12 in Sept so we can make a new release of Ratel by the end of Sept.

Very cool! I have been testing this PR and the prior 2 for H(curl) and H(div) basis support in `cuda/ref`, `hip/ref`, and `magma` and everything is working surprisingly well. I haven't taken them out of draft since it's only been a few days of working on this alongside some other stuff but if you'd like to start review for v0.12 that would be fine with me.

jeremylt:
It looks like this PR needs to be rebased for changes in main but is otherwise ready to merge?

sebastiangrimberg:
> It looks like this PR needs to be rebased for changes in main but is otherwise ready to merge?

Yes! This is ready for review/merge from my point of view, I have been using it extensively on GPU backends for the past while.

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
```diff
@@ -1071,16 +1071,16 @@ static int CeedSingleOperatorAssembleSetup_Sycl(CeedOperator op) {
...
   /* CeedInt maxThreadsPerBlock = sycl_data->sycl_device.get_info<sycl::info::device::max_work_group_size>();
   bool    fallback           = block_size > maxThreadsPerBlock;
   asmb->fallback             = fallback;
   if (fallback) {
     // Use fallback kernel with 1D threadblock
-    block_size         = elem_size * elem_per_block;
+    block_size         = elem_size * elems_per_block;
     asmb->block_size_x = elem_size;
     asmb->block_size_y = 1;
   } else {  // Use kernel with 2D threadblock
```
jeremylt, line 1086: Since we're here - we should probably ditch the dead code; its in the commit history if we want it back

jeremylt:
This looks good (conflicts with main currently) but any thoughts on why the code coverage is so low?

sebastiangrimberg:
> This looks good (conflicts with main currently) but any thoughts on why the code coverage is so low?

Not really sure. I will take a deeper look but on first pass I would have thought the unit tests should have covered this (especially for the element restriction, I believe).

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
sebastiangrimberg: Done in [4826ef4](https://github.com/CEED/libCEED/pull/1302/commits/4826ef44fa2b959953a68071d4331d5eb8192ac4).

jeremylt:
Ah, its all the memory space specific copying logic. Maybe we should do a follow-up after this PR to do some helper utilities for this common logic that is spread over restrictions, bases, and vectors for some of this memory space copying?

jeremylt (approved):
Approved with the caveat that we should follow-up with a plan to reduce common memory space copying code.

sebastiangrimberg:
> Approved with the caveat that we should follow-up with a plan to reduce common memory space copying code.

See https://github.com/CEED/libCEED/issues/1441 for the issue tracker.

PR #1303, opened 2023-08-23
fluids: Refactor setupgeo*, calculate dXdx directly for strong BC

jrwrigh:
This PR does two different, but related things:
- Refactor the `setupgeo.h` functions into helpers
- Refactor the strong boundary condition setup to use those helpers

There is not change to functionality here, just cleaning things up a bit.


jeremylt (approved):
LGTM

PR #1308, opened 2023-08-25
Fluids: Add DMSetupByOrder helper functions 

jrwrigh:
These functions help to make the setting up of auxillary DMs and PetscFE objects more consistent.

Also created `src/dm_utils.c` as a central location for these DM functions that are used in several different places.


examples/fluids/src/dm_utils.c
```diff
@@ -0,0 +1,352 @@
...
+  // Set tensor permutation if needed
+  if (!is_simplex) {
+    PetscCall(DMPlexSetClosurePermutationTensor(dm, PETSC_DETERMINE, NULL));
+    if (setup_coords) {
+      DM dm_coord;
+
+      PetscCall(DMGetCoordinateDM(dm, &dm_coord));
+      PetscCall(DMPlexSetClosurePermutationTensor(dm_coord, PETSC_DETERMINE, NULL));
+    }
+  }
```
jeremylt, line 349: I worry that this is really easy to forget and is a potential hard-to-find-bug bomb. But I'm not seeing an obvious workaround if you aren't setting any Dirichlet BCs inside `DMSetupByOrder_FEM` like we do in Ratel.

examples/fluids/src/grid_anisotropy_tensor.c
```diff
@@ -29,18 +28,13 @@ PetscErrorCode GridAnisotropyTensorProjectionSetupApply(Ceed ceed, User user, Ce
...
     PetscSection section;
-    PetscInt     q_order = user->app_ctx->degree + user->app_ctx->q_extra;
-    PetscCall(PetscFECreateLagrange(PETSC_COMM_SELF, dim, grid_aniso_proj->num_comp, PETSC_FALSE, user->app_ctx->degree, q_order, &fe));
-    PetscCall(PetscObjectSetName((PetscObject)fe, "Grid Anisotropy Tensor Projection"));
-    PetscCall(DMAddField(grid_aniso_proj->dm, NULL, (PetscObject)fe));
-    PetscCall(DMCreateDS(grid_aniso_proj->dm));
-    PetscCall(DMPlexSetClosurePermutationTensor(grid_aniso_proj->dm, PETSC_DETERMINE, NULL));
+    PetscCall(DMSetupByOrder_FEM(PETSC_TRUE, PETSC_TRUE, user->app_ctx->degree, 1, user->app_ctx->q_extra, 1, &grid_aniso_proj->num_comp,
+                                 grid_aniso_proj->dm));
+    PetscCall(DMSetupByOrder_Finalize(PETSC_TRUE, grid_aniso_proj->dm));
```
jeremylt, line 37: Maybe a helper for when these two are back-to-back?

examples/fluids/src/dm_utils.c
```diff
@@ -0,0 +1,352 @@
...
+    PetscCall(PetscFEGetNumComponents(fe, &num_comp));
+    PetscCall(PetscFEGetDualSpace(fe, &dual_space));
+    PetscCall(PetscDualSpaceGetDimension(dual_space, &num_dual_basis_vectors));
+    P = num_dual_basis_vectors / num_comp;
+    PetscCall(PetscQuadratureGetData(quadrature, NULL, NULL, &Q, NULL, NULL));
+
+    CeedInt P_1d = (CeedInt)round(pow(P, 1.0 / dim));
+    CeedInt Q_1d = (CeedInt)round(pow(Q, 1.0 / dim));
+
+    PetscCallCeed(ceed, CeedBasisCreateTensorH1Lagrange(ceed, dim, num_comp, P_1d, Q_1d, CEED_GAUSS, basis));
```
jeremylt, line 255: See this MR - we have an update to this logic that we may want to migrate here, probably in a separate PR

https://gitlab.com/micromorph/ratel/-/merge_requests/584

examples/fluids/src/dm_utils.c
jrwrigh: Yeah, I will want to move to calling `DMAddBoundary` within `DMSetupByOrder_FEM`, but once I go through with the boundary condition refactor. 

The main issue is that the `DMAddBoundary` calls requires other structs that might be be available everywhere `DMSetupByOrder_FEM` needs to be called. Otherwise I'd just add them in and call it good enough for now.

examples/fluids/src/grid_anisotropy_tensor.c
jrwrigh: Yeah, I'm fine with that. Suggestions on name?

jrwrigh: Maybe `DMSetupByOrder_FEM_NoBC`?

examples/fluids/src/dm_utils.c
jrwrigh: Yeah, I noticed that before. I'll leave that for another PR.

examples/fluids/src/grid_anisotropy_tensor.c
jeremylt: Depends upon if we intend this to stick around long term. 

Perhaps
`DMSetupByOrderBegin_FEM`
`DMSetupByOrderEnd_FEM`
`DMSetupByOrder_FEM` (the two together)

or something along those lines

jrwrigh: I don't *intend* this to stick around long term. But intent is different than reality sometimes.

I'm fine going that route and then changing it in the future. I don't think the migration between the two will be too difficult.

jrwrigh: Done.

examples/fluids/src/dm_utils.c
jrwrigh: Addressed in this thread: https://github.com/CEED/libCEED/pull/1308#discussion_r1306219024

PR #1309, opened 2023-08-26
fluids: Add DMPlexCeedElemRestriction*Create

jrwrigh:
Replaces CreateRestrictionFromPlex and GetRestrictionForDomain

Dependent on #1308, mostly because I wanted to put these functions in `src/dm_utils.c`

Todos:
- [x] Scan through for manually created strided element restrictions to replace with these new functions
- [x] Clean out `num_elem` and `num_qpnts` from the code using the afore mentioned manual strided restriction creation.

### DIFF
--- a/examples/fluids/src/dm_utils.c
+++ b/examples/fluids/src/dm_utils.c
@@ -93,16 +93,13 @@ PetscErrorCode DMPlexCeedElemRestrictionCreate(Ceed ceed, DM dm, DMLabel domain_
 **/
 PetscErrorCode DMPlexCeedElemRestrictionCoordinateCreate(Ceed ceed, DM dm, DMLabel domain_label, PetscInt label_value, PetscInt height,
                                                          CeedElemRestriction *restriction) {
-  DM        dm_coord;
-  PetscBool is_simplex = PETSC_FALSE;
+  DM dm_coord;
 
   PetscFunctionBeginUser;
   PetscCall(DMGetCellCoordinateDM(dm, &dm_coord));
   if (!dm_coord) {
     PetscCall(DMGetCoordinateDM(dm, &dm_coord));
   }
-  PetscCall(DMPlexIsSimplex(dm_coord, &is_simplex));
-  if (!is_simplex) PetscCall(DMPlexSetClosurePermutationTensor(dm_coord, PETSC_DETERMINE, NULL));
   PetscCall(DMPlexCeedElemRestrictionCreate(ceed, dm_coord, domain_label, label_value, height, 0, restriction));
 
   PetscFunctionReturn(PETSC_SUCCESS);
@@ -456,11 +453,12 @@ PetscErrorCode DMSetupByOrderBegin_FEM(PetscBool setup_faces, PetscBool setup_co
     PetscCall(PetscDualSpaceGetOrder(fe_coord_dual_space, &fe_coord_order));
 
     // Create FE for coordinates
-    PetscCheck(fe_coord_order == 1 || coord_order == 1, PetscObjectComm((PetscObject)dm), PETSC_ERR_USER_INPUT,
-               "Only linear mesh geometry supported. Recieved %d order", fe_coord_order);
+    PetscCheck(fe_coord_order == 1 || coord_order == 1, comm, PETSC_ERR_USER_INPUT, "Only linear mesh geometry supported. Recieved %d order",
+               fe_coord_order);
     PetscCall(PetscFECreateLagrange(comm, dim, num_comp_coord, is_simplex, fe_coord_order, q_order, &fe_coord_new));
     if (setup_faces) PetscCall(PetscFEGetHeightSubspace(fe_coord_new, 1, &fe_coord_face_new));
     PetscCall(DMProjectCoordinates(dm, fe_coord_new));
+    PetscCall(DMLocalizeCoordinates(dm));  // Update CellCoordinateDM with projected coordinates
     PetscCall(PetscFEDestroy(&fe_coord_new));
   }
   PetscFunctionReturn(PETSC_SUCCESS);
@@ -489,6 +487,8 @@ PetscErrorCode DMSetupByOrderEnd_FEM(PetscBool setup_coords, DM dm) {
 
       PetscCall(DMGetCoordinateDM(dm, &dm_coord));
       PetscCall(DMPlexSetClosurePermutationTensor(dm_coord, PETSC_DETERMINE, NULL));
+      PetscCall(DMGetCellCoordinateDM(dm, &dm_coord));
+      if (dm_coord) PetscCall(DMPlexSetClosurePermutationTensor(dm_coord, PETSC_DETERMINE, NULL));
     }
   }
   PetscFunctionReturn(PETSC_SUCCESS);
--- a/examples/fluids/src/dm_utils.c
+++ b/examples/fluids/src/dm_utils.c
@@ -14,54 +14,6 @@
 
 #include "../navierstokes.h"
 
-// Utility function to create local CEED restriction
-PetscErrorCode CreateRestrictionFromPlex(Ceed ceed, DM dm, CeedInt height, DMLabel domain_label, CeedInt label_value, PetscInt dm_field,
-                                         CeedElemRestriction *elem_restr) {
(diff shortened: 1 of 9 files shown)


examples/fluids/src/dm_utils.c
```diff
@@ -96,6 +47,202 @@ PetscErrorCode DMFieldToDSField(DM dm, DMLabel domain_label, PetscInt dm_field,
...
+  @param[in]   height         Height of `DMPlex` topology
+  @param[in]   q_data_size    Number of components for `QFunction` data
+  @param[in]   is_collocated  Boolean flag indicating if the data is collocated on the nodes (`PETSC_TRUE`) on on quadrature points (`PETSC_FALSE`)
+  @param[out]  restriction    Strided `CeedElemRestriction` for `QFunction` data
+
+  @return An error code: 0 - success, otherwise - failure
+**/
+static PetscErrorCode DMPlexCeedElemRestrictionStridedCreate(Ceed ceed, DM dm, DMLabel domain_label, PetscInt label_value, PetscInt height,
+                                                             PetscInt q_data_size, PetscBool is_collocated, CeedElemRestriction *restriction) {
+  PetscInt num_elem, num_qpts, dm_field = 0;
```
jrwrigh, line 130: I'm not sure how I feel about always setting `dm_field` equal to 0. On one hand, I don't think I have an example of needing strided restrictions for DM fields other than the default one, but I _do_ create restrictions for a single field of a DM. 

I don't see any reason to not ask for `dm_field` here. @jeremylt was there any specific thought given to this, or was it just that Ratel doesn't use anything other than `dm_field = 0` and so reducing the surface area of the API?

jeremylt: The quadrature space has to have the same number of points across all fields, even if each field has different numbers of nodes on each element

jrwrigh: > The quadrature space has to have the same number of points across all fields, even if each field has different numbers of nodes on each element

That's true only if they're used in the same qfunctions though, right? I can't think of a reason to have two fields in the same DM, used in different QFunctions with different quadrature rules, but I've had a field be used in one QFunction, then used in a wholly unrelated QFunction before (in the SmartSim PR, I use a field to perform differential filtering, then use that field to gather the ML training data together). 

All that said, there might be reason to, but that's very complicated and if I run into it, running the plumbing for adding a `dm_field` shouldn't be too hard.

jeremylt: The convention I've been using is 1 dm == 1 quadrature space, especially since the coordinates can only have one quadrature space. I agree we should see if we have a need to break this convention in the future and generalize further then.

jrwrigh:
@jeremylt Good for review whenever

examples/fluids/src/dm_utils.c
```diff
@@ -96,6 +47,202 @@ PetscErrorCode DMFieldToDSField(DM dm, DMLabel domain_label, PetscInt dm_field,
...
+  DM        dm_coord;
+  PetscBool is_simplex = PETSC_FALSE;
+
+  PetscFunctionBeginUser;
+  PetscCall(DMGetCellCoordinateDM(dm, &dm_coord));
+  if (!dm_coord) {
+    PetscCall(DMGetCoordinateDM(dm, &dm_coord));
+  }
+  PetscCall(DMPlexIsSimplex(dm_coord, &is_simplex));
+  if (!is_simplex) PetscCall(DMPlexSetClosurePermutationTensor(dm_coord, PETSC_DETERMINE, NULL));
```
jeremylt, line 106: Why not call this only once when we create the DM instead of calling every time we make a restriction?

jeremylt: We only get the coordinate restriction a couple of times, but this still feels like DM - creating a restriction shouldn't mutate the DM

jrwrigh: Because the CellCoordinateDM isn't necessarily setup in the `DMSetupByOrderEnd_FEM`. It's only an issue when periodic faces are specified. Maybe we should setup both CoordinateDM and CellCoordinateDM during setup? I'm not familiar with why both exist or why you'd want to choose one over the other (excluding the CoordinateDM with periodic faces is wrong).

jeremylt: This still  feels like the wrong place to do mutation. Can we push up the setting up of the CellCoordinateDM to some time before this?

jrwrigh: You're right, it's definitely not the best place for it. Trying to figure out the best way to integrate CellCoordinateDM into the `DMSetupByOrder{Begin,End}_FEM` right now. Can't tell if it needs to go through the same coordinate projection as the normal CoordinateDM.

examples/fluids/src/dm_utils.c
```diff
@@ -313,11 +457,12 @@ PetscErrorCode DMSetupByOrderBegin_FEM(PetscBool setup_faces, PetscBool setup_co
     PetscCall(PetscDualSpaceGetOrder(fe_coord_dual_space, &fe_coord_order));
 
     // Create FE for coordinates
-    PetscCheck(fe_coord_order == 1 || coord_order == 1, PetscObjectComm((PetscObject)dm), PETSC_ERR_USER_INPUT,
+    PetscCheck(fe_coord_order == 1 || coord_order == 1, comm, PETSC_ERR_USER_INPUT,
                "Only linear mesh geometry supported. Recieved %d order", fe_coord_order);
     PetscCall(PetscFECreateLagrange(comm, dim, num_comp_coord, is_simplex, fe_coord_order, q_order, &fe_coord_new));
     if (setup_faces) PetscCall(PetscFEGetHeightSubspace(fe_coord_new, 1, &fe_coord_face_new));
     PetscCall(DMProjectCoordinates(dm, fe_coord_new));
+    PetscCall(DMLocalizeCoordinates(dm));  // Update CellCoordinateDM with projected coordinates
```
jrwrigh, line 465: I'm not sure how necessary this is, primarily because we don't have any tests that actually change the FE coordinate space. But I think this is required to get the cell geometry correct. And I believe that this should still respect `-dm_sparse_localize` as well.

examples/fluids/src/dm_utils.c
jrwrigh: Alright, I've moved the `DMPlexSetClosurePermutationTensor` for the cell coordinate DM to `DMSetupByOrderEnd_FEM`.

examples/fluids/src/dm_utils.c
```diff
@@ -96,6 +47,199 @@ PetscErrorCode DMFieldToDSField(DM dm, DMLabel domain_label, PetscInt dm_field,
...
+  @param[in]   height        Height of `DMPlex` topology
+  @param[out]  restriction   `CeedElemRestriction` for mesh
+
+  @return An error code: 0 - success, otherwise - failure
+**/
+PetscErrorCode DMPlexCeedElemRestrictionCoordinateCreate(Ceed ceed, DM dm, DMLabel domain_label, PetscInt label_value, PetscInt height,
+                                                         CeedElemRestriction *restriction) {
+  DM dm_coord;
+
+  PetscFunctionBeginUser;
```
jeremylt, line 99: What is the standard PETSc convention here? I always have a blank line after `PetscFunctionBeginUser`

jrwrigh: From https://petsc.org/release/developers/style/#c-formatting
> 4. From All PETSc subroutine code blocks must start with a single blank line between the local variable declarations followed by [PetscFunctionBegin](https://petsc.org/release/manualpages/Sys/PetscFunctionBegin/).

From https://petsc.org/release/developers/style/#usage-of-petsc-functions-and-macros
> 7. Do not put a blank line immediately after [`PetscFunctionBegin`](https://petsc.org/release/manualpages/Sys/PetscFunctionBegin/); or a blank line immediately before `PetscFunctionReturn(PETSC_SUCCESS);`.

So technically it's for `PetscFunctionBegin` rather `PetscFunctionBeginUser`, but I'm applying it to both.

jeremylt (approved):

examples/fluids/src/dm_utils.c
jrwrigh: I'll go ahead and call this good since we can't really test it right now. Either including it is wrong and we'll find out when we get there, or not including it is right... and we'll still find that out later anyways.

PR #1310, opened 2023-08-29
Remove CeedOperatorSetNumQuadraturePoints

jeremylt:
This should be merged right before we cut the new release.

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -757,10 +757,10 @@ found:
     op->has_restriction = true;  // Restriction set, but num_elem may be 0
   }
   CeedCall(CeedBasisReferenceCopy(b, &(*op_field)->basis));
-  if (op->num_qpts == 0) CeedCall(CeedOperatorSetNumQuadraturePoints(op, num_qpts));
-
+  op->num_qpts = num_qpts;
   op->num_fields += 1;
   CeedCall(CeedStringAllocCopy(field_name, (char **)&(*op_field)->field_name));
+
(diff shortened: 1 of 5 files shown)


interface/ceed-operator.c
```diff
@@ -757,10 +757,10 @@ int CeedOperatorSetField(CeedOperator op, const char *field_name, CeedElemRestri
     op->has_restriction = true;  // Restriction set, but num_elem may be 0
   }
   CeedCall(CeedBasisReferenceCopy(b, &(*op_field)->basis));
-  if (op->num_qpts == 0) CeedCall(CeedOperatorSetNumQuadraturePoints(op, num_qpts));
-
+  op->num_qpts = num_qpts;
```
jrwrigh, line 760: Maybe we should move this up to line 712, where `num_qpts` is checked against it's previous value? I feel like this makes more sense up there than here. (plus I spent about 15 minutes trying to figure out whether removing the `if (op->num_qpts == 0)` had any consequence on the code. I think this on line 712 would've prevented that, as it puts the setting of the value right by where it gets checked).

jrwrigh (approved):
Minor comment on code placement

interface/ceed-operator.c
jrwrigh: Maybe even keeping the if statement to show intent would be a good idea. ie. we're not going to reset the value every single call. It should only ever be one value.

jeremylt: I put back the `if (op->num_qpts == 0)` since we cannot move this logic

PR #1311, opened 2023-08-29
fluids: Fix preloading vector

jrwrigh:
- Preloading should use Q_preload, not Q. Otherwise the initial condition for the "actual" solve is changed by the preload step


jedbrown (approved):
Thanks. I imagine you can confirm easily enough that a verification test fails when you add `-preload` with the old code, and passes with this PR.

jrwrigh:
Yep, `-preload` no longer causes tests to fail.

PR #1316, opened 2023-08-31
Non-square operator full assembly

### LINKED ISSUE #1277: `CeedSingleOperatorAssemble` with rectangular operators
Right now, the assertion at https://github.com/CEED/libCEED/blob/main/interface/ceed-preconditioning.c#L590 fails when trying to fully assembly a rectangular operator with different input and output bases. Looking at the code, I think it should still work (so long as there is only one single active input basis and one single active output basis). However, I'm not very familiar with this multiple-active-bases concept so wanted to ask first. Does this look like an easy thing to fix?

sebastiangrimberg:
Adds test `t569-operator.c` which fails on `main`, and subsequently resolves it by adding support for rectangular operators in `CeedSingleOperatorAssemble`.

Resolves #1277.

TODO
- [ ] ~~GPU backends~~ EDIT: GPU backend full assembly for rectangular operators will be completed as part of #1302.


sebastiangrimberg:
Hi @jeremylt, this is ready for review if you're interested and OK with waiting on a separate PR for GPU backends (https://github.com/CEED/libCEED/pull/1302). I decided to split it up into two to avoid getting too large. For now full assembly with `/gpu/cuda` or `/gpu/ref` and rectangular operators will continue to skip tests with `"Backend does not implement..."` messages (same behavior as previously).

jeremylt (approved):
Yeah, I'm fine splitting off the GPU fix separately

sebastiangrimberg:
> Yeah, I'm fine splitting off the GPU fix separately

Nice! This is now working as part of https://github.com/CEED/libCEED/pull/1302, which I will update and mark as ready for review shortly. For testing, beyond the unit tests, I'm able to fully assemble mixed operators from MFEM, as well as linear interpolators (like prolongation matrices or discrete gradient) either for passing to external libraries or to compare the partial assembly and sparse matrix performance for various orders. Very cool!

jeremylt:
Feel free to merge if this is ready. Since this adds a failing test and then fixes it, I think we should squash+merge

sebastiangrimberg:
Sounds good, will do! I'll squash + merge once #1279 is in. Thanks!

PR #1317, opened 2023-08-31
Update clang-format version to v16

sebastiangrimberg:
Version 16 catches one minor formatting change in `interface/ceed.c` which is somewhat tedious to have to keep restoring after every `make format`. Homebrew's default `clang` version is 16 since Spring of 2023. If other developers are running into the same issue, then this PR would provide a minor fix if desired.

Alternatively if this change is undesired I could look into what fixes are needed in `.clang-format` to keep the current formatting.

### DIFF
--- a/.github/workflows/c-fortran-test-style.yml
+++ b/.github/workflows/c-fortran-test-style.yml
@@ -19,7 +19,10 @@ jobs:
     - name: Environment setup
       uses: actions/checkout@v3
     - name: Install clang-format
-      run: sudo apt update && sudo apt install clang-format-16
+      run: |
+          wget -O- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+          sudo add-apt-repository 'deb http://apt.llvm.org/bullseye/ llvm-toolchain-bullseye-16 main'
+          sudo apt update && sudo apt install clang-format-16
     - name: C style
       env:
         CC: ${{ matrix.compiler }}
--- a/.github/workflows/c-fortran-test-style.yml
+++ b/.github/workflows/c-fortran-test-style.yml
@@ -19,11 +19,11 @@ jobs:
     - name: Environment setup
       uses: actions/checkout@v3
     - name: Install clang-format
-      run: sudo apt update && sudo apt install clang-format-15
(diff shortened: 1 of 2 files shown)


jrwrigh:
>  If other developers are running into the same issue

I've also started running into this issue. It's slightly annoying, but not too bad since it's a single file

jeremylt:
> E: Unable to locate package clang-format-16

not available for CI yet

sebastiangrimberg:
> not available for CI yet

Whoops, forgot to add that. Fixed now but I understand if its desired to use the CI OS default clang version.

.github/workflows/c-fortran-test-style.yml
```diff
@@ -19,11 +19,14 @@ jobs:
     - name: Environment setup
       uses: actions/checkout@v3
     - name: Install clang-format
-      run: sudo apt update && sudo apt install clang-format-15
+      run: |
+          wget -O- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+          sudo add-apt-repository 'deb http://apt.llvm.org/bullseye/ llvm-toolchain-bullseye-16 main'
+          sudo apt update && sudo apt install clang-format-16
```
jeremylt, line 25: ah, this wouldn't be needed (I think) if GitHub Actions was using Ubuntu 23.04, but that doesn't seem like it will happen any time soon, and I don't think 20.04 will get `clang-format-16` any time soon. Long story short - I think this is probably our best bet for now.

It looks like `clang-format-16` will be the default on Ubuntu 23.10, and this is a small change, so I figure we can merge now?

jeremylt (approved):

.github/workflows/c-fortran-test-style.yml
jeremylt: Oh, but `clang-16` isn't conveniently available on Arch linux yet

jeremylt (approved):

.github/workflows/c-fortran-test-style.yml
sebastiangrimberg: > It looks like clang-format-16 will be the default on Ubuntu 23.10, and this is a small change, so I figure we can merge now?

Nice, I agree with this

Issue #1319, opened 2023-08-31
Ratel HIP Regression


jeremylt:
```console
jeth8984@noether:/projects/ratel/libCEED$ git bisect good
b2165e7a2e371018feedcb47974a027ed47e0487 is the first bad commit
commit b2165e7a2e371018feedcb47974a027ed47e0487
Author: Sebastian Grimberg <sjg@amazon.com>
Date:   Fri Aug 11 10:17:04 2023 -0700

    Whitespace, style, and formatting updates for consistency between CUDA and HIP backends
    
    Adds include guards in JiT header files, even if not strictly necessary, to match the precedent set in cuda-shared and hip-shared as well as sycl.

 backends/cuda-gen/ceed-cuda-gen-operator-build.cpp | 16 ++--
 backends/cuda-gen/ceed-cuda-gen-operator.c         |  5 +-
 backends/cuda-gen/ceed-cuda-gen.c                  |  7 +-
 backends/cuda-ref/ceed-cuda-ref-basis.c            | 35 +++++----
 backends/cuda-ref/ceed-cuda-ref-operator.c         | 12 ++-
 backends/cuda-ref/ceed-cuda-ref-vector.c           |  4 +-
 backends/cuda-ref/ceed-cuda-ref.c                  |  1 -
 backends/cuda-ref/ceed-cuda-ref.h                  |  6 +-
 backends/cuda-shared/ceed-cuda-shared-basis.c      | 21 ++++--
 backends/cuda-shared/ceed-cuda-shared.c            |  3 +-
 backends/cuda/ceed-cuda-common.c                   |  1 -
 backends/cuda/ceed-cuda-common.h                   |  2 +-
 backends/cuda/ceed-cuda-compile.cpp                |  8 +-
 backends/cuda/ceed-cuda-compile.h                  |  1 -
 backends/hip-gen/ceed-hip-gen-operator-build.cpp   | 11 +++
 backends/hip-gen/ceed-hip-gen-operator.c           |  1 +
 backends/hip-gen/ceed-hip-gen.c                    |  7 +-
 backends/hip-ref/ceed-hip-ref-basis.c              | 36 ++++-----
 backends/hip-ref/ceed-hip-ref-operator.c           |  6 +-
 backends/hip-ref/ceed-hip-ref-qfunction.c          |  4 +-
 backends/hip-ref/ceed-hip-ref-qfunctioncontext.c   | 11 ++-
[...]

PR #1321, opened 2023-09-01
gpu - fix SetArray on device logic bug

jeremylt:
And the second half of #1319 

We copied a bug from CUDA to HIP


sebastiangrimberg (approved):
Yep, looks good, thank you! Sorry for not catching this one in https://github.com/CEED/libCEED/pull/1299.

PR #1322, opened 2023-09-01
Fix a few header includes following PR #1320

sebastiangrimberg:
Some added ones for MAGMA to follow in #1301.


jeremylt (approved):
woops, thanks for catching these I missed!

PR #1324, opened 2023-09-05
Fix order of tensor contractions for `CeedBasisApplyAtPoints`

zatkins-dev:
Changes the ordering of tensor contractions in `CeedBasisApplyAtPoints` such that the inner dimension is `x` rather than `z` in order to be consistent with other `CeedBasisApply*` for tensor bases. 

This also requires changing the order of the coordinate vectors in `t35*`.


jeremylt (approved):
Looks great to me, thanks!

PR #1325, opened 2023-09-05
Change coordinate ordering to be consistent with tensor basis application

zatkins-dev:
Changes the order of coordinates in basis tests to be consistent with changes in #1324. Due to test setup, this required little or no additional changes to most of the tests. The main exception is `t319`, which also required a change in the indexing of the output `du_to` array from `du_array[p_to_dim * (dim - 1 - d) + i]` to `du_array[p_to_dim * d + i]` (the fact this permutation was ever necessary seems like an indicator that the old ordering was incorrect).

### DIFF
--- a/tests/t313-basis-f.f90
+++ b/tests/t313-basis-f.f90
@@ -54,7 +54,7 @@
 
         do d=0,dimn-1
           do i=1,xdim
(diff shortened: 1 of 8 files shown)


jeremylt (approved):
Thanks for the fast update

PR #1329, opened 2023-09-08
fluids: Add qdata packing/unpacking helpers

jrwrigh:
Adopts the [`RatelStoredValuesPack` and `RatelStoredValuesUnpack` functions](https://gitlab.com/micromorph/ratel/-/blob/main/include/ratel/qfunctions/utils.h?ref_type=heads#L99-131).

This cleans up the code a bit while also helping prevent any odd unpacking errors.


examples/fluids/qfunctions/newtonian.h
```diff
@@ -126,20 +126,13 @@ CEED_QFUNCTION(RHSFunction_Newtonian)(void *ctx, CeedInt Q, const CeedScalar *co
 
   // Quadrature Point Loop
   CeedPragmaSIMD for (CeedInt i = 0; i < Q; i++) {
-    CeedScalar U[5];
+    CeedScalar U[5], wdetJ, dXdx[3][3];
     for (int j = 0; j < 5; j++) U[j] = q[j][i];
+    StoredValuesUnpack(Q, i, 0, 1, q_data, &wdetJ);
+    StoredValuesUnpack(Q, i, 1, 9, q_data, (CeedScalar *)dXdx);
```
jeremylt, line 132: I haven't found a good way around the potential bug which is these raw indices yet. Any thoughts that aren't messier?

jeremylt (approved):
Looks good to me (naturally)

examples/fluids/qfunctions/newtonian.h
jrwrigh: Nah, I can't really think of anything obviously better. That's why I created the `Qdata*Unpack` functions to get away from that. 

Someone has to know what data is where and how big it is. Enums/Global consts seem like the only way around that. Or creating a struct for every unique q_data type and then "casting" the data into the struct, but that's a bit bulky seems a bit fragile.

jrwrigh:
Not sure how/why codecov isn't happy, but feel free to merge whenever; I'm done with changes here.

PR #1330, opened 2023-09-11
fluids: Add StatePhysicalGradientFromReference util

jrwrigh:
Add helper function to calculate physical gradients of solution `State` from reference-element gradients.

### DIFF
--- a/examples/fluids/qfunctions/newtonian.h
+++ b/examples/fluids/qfunctions/newtonian.h
@@ -110,10 +110,10 @@ CEED_QFUNCTION(ICsNewtonianIG_Conserv)(void *ctx, CeedInt Q, const CeedScalar *c
 // *****************************************************************************
 CEED_QFUNCTION(RHSFunction_Newtonian)(void *ctx, CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out) {
   // Inputs
-  const CeedScalar(*q)[CEED_Q_VLA]         = (const CeedScalar(*)[CEED_Q_VLA])in[0];
-  const CeedScalar(*Grad_q)[5][CEED_Q_VLA] = (const CeedScalar(*)[5][CEED_Q_VLA])in[1];
-  const CeedScalar(*q_data)                = in[2];
-  const CeedScalar(*x)[CEED_Q_VLA]         = (const CeedScalar(*)[CEED_Q_VLA])in[3];
+  const CeedScalar(*q)[CEED_Q_VLA] = (const CeedScalar(*)[CEED_Q_VLA])in[0];
+  const CeedScalar(*Grad_q)        = in[1];
+  const CeedScalar(*q_data)        = in[2];
+  const CeedScalar(*x)[CEED_Q_VLA] = (const CeedScalar(*)[CEED_Q_VLA])in[3];
 
   // Outputs
   CeedScalar(*v)[CEED_Q_VLA]         = (CeedScalar(*)[CEED_Q_VLA])out[0];
@@ -134,12 +134,7 @@ CEED_QFUNCTION(RHSFunction_Newtonian)(void *ctx, CeedInt Q, const CeedScalar *co
     State            s      = StateFromU(context, U, x_i);
 
     State grad_s[3];
-    for (CeedInt k = 0; k < 3; k++) {
-      CeedScalar dx_i[3] = {0}, dU[5];
-      for (CeedInt j = 0; j < 5; j++) dU[j] = Grad_q[0][j][i] * dXdx[0][k] + Grad_q[1][j][i] * dXdx[1][k] + Grad_q[2][j][i] * dXdx[2][k];
-      dx_i[k]   = 1.;
-      grad_s[k] = StateFromU_fwd(context, s, dU, x_i, dx_i);
-    }
+    StatePhysicalGradientFromReference(Q, i, context, s, x_i, STATEVAR_CONSERVATIVE, Grad_q, dXdx, false, grad_s);
 
     CeedScalar strain_rate[6], kmstress[6], stress[3][3], Fe[3];
     KMStrainRate_State(grad_s, strain_rate);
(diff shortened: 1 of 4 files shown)


examples/fluids/qfunctions/newtonian.h
```diff
@@ -134,12 +134,7 @@ CEED_QFUNCTION(RHSFunction_Newtonian)(void *ctx, CeedInt Q, const CeedScalar *co
     State            s      = StateFromU(context, U, x_i);
 
     State grad_s[3];
-    for (CeedInt k = 0; k < 3; k++) {
-      CeedScalar dx_i[3] = {0}, dU[5];
-      for (CeedInt j = 0; j < 5; j++) dU[j] = Grad_q[0][j][i] * dXdx[0][k] + Grad_q[1][j][i] * dXdx[1][k] + Grad_q[2][j][i] * dXdx[2][k];
-      dx_i[k]   = 1.;
-      grad_s[k] = StateFromU_fwd(context, s, dU, x_i, dx_i);
-    }
+    StatePhysicalGradientFromReference(Q, i, context, s, x_i, STATEVAR_CONSERVATIVE, Grad_q, dXdx, false, grad_s);
```
jrwrigh, line 137: Is it possible/desirable to use `CEED_Q_VLA` here instead of just `Q`? With the current implementation, the compiler knows the exact offset via `CEED_Q_VLA` being set. I'd imagine that for GPUs it could possibly compile better if it knew the exact offset that would be required to get to each component of `Grad_q`.

jeremylt: That value of `Q` is a compile time constant set during the code generation:
```
  code << "  const CeedInt Q_1d = " << Q_1d << ";\n";
```

jrwrigh: So it's basically equivalent to `CEED_Q_VLA`?

jeremylt: After all the compilation optimization passes, I think its functionally the same. `CEED_Q_VLA`  drops in an integer literal for the GPU JiT to get around the lack of support for VLAs in C++ blocking our recasting for readable indexing. But the JiT compiler will know that `Q` isn't changing because its based on a const value and should be able to optimize to the same place.

examples/fluids/qfunctions/newtonian_state.h
```diff
@@ -404,4 +404,33 @@ CEED_QFUNCTION_HELPER void Vorticity(const State grad_s[3], CeedScalar vorticity
...
+                                                                      const CeedScalar dXdx[2][3], bool zero_dx, State grad_s[3]) {
+  for (CeedInt k = 0; k < 3; k++) {
+    CeedScalar dx_i[3] = {0}, dqi[5];
+    for (CeedInt j = 0; j < 5; j++) {
+      dqi[j] = grad_q[(Q * 5) * 0 + Q * j + i] * dXdx[0][k] + grad_q[(Q * 5) * 1 + Q * j + i] * dXdx[1][k];
+    }
+    dx_i[k]   = zero_dx ? 0. : 1.;
+    grad_s[k] = StateFromQ_fwd(gas, s, dqi, x_i, dx_i, state_var);
+  }
+  return CEED_ERROR_SUCCESS;
```
jedbrown, line 433: I think these can return `void` since there is no plausible error path.

examples/fluids/qfunctions/newtonian.h
jedbrown: I suppose we could use a macro that uses `CEED_Q_VLA` so it doesn't have to be written and the implementation could use `CEED_Q_VLA` to avoid the more clumsy flattened indexing. The issue is that C++ doesn't have VLA (VLA pointers) so we can't dimension based on the "variable" `Q` even when the compiler will specialize it based on known compile-time values. Similarly, this is not valid C++
```cpp
const int n = 3;
int x[n];
```

jedbrown (approved):

examples/fluids/qfunctions/newtonian_state.h
jrwrigh: Yeah, I just did that since a lot of Ratel QFunction helpers I've ~~stolen~~ copied have return values. Not sure if there's any benefit either way.

jeremylt: We don't actually use any of the error codes in QFunction helpers currently (if I recall correctly, its because CUDA/HIP backends don't actually check the error codes) but we should probably add some flavor of `CeedCall` for the CPU side of things that expands to nothing for JiT

examples/fluids/qfunctions/newtonian.h
jrwrigh: > I suppose we could use a macro that uses `CEED_Q_VLA` so it doesn't have to be written and the implementation could use `CEED_Q_VLA` to avoid the more clumsy flattened indexing.

I'm not sure I follow. Do you mean converting `StatePhysicalGradientFromReference*` to be a macro rather than a helper?

examples/fluids/qfunctions/newtonian_state.h
jedbrown: I think of returning an error code like returning `Result` in Rust. If the operation is not supposed to have a "failure" path, then returning `void` is fine. Higher level PETSc operations pretty much all have failure paths so we use the Call macros for uniformity. In purely numerical code, many things should not be able to fail.

examples/fluids/qfunctions/newtonian.h
jedbrown: If you did that, then it could use `CEED_Q_VLA`. I don't have a strong opinion either way, and expect that what you did here has no impact on performance.

examples/fluids/qfunctions/newtonian_state.h
jrwrigh: Ok, I'll remove them for now, just for consistency. And probably remove them from the QFunction helpers.

examples/fluids/qfunctions/newtonian.h
jrwrigh: I'll double check the performance real quick. I highly doubt it'll change either.

jrwrigh: Everything looks good to me. If anything, this branch is slightly faster than `main` right now for some reason (after rebaseing off of main of course).

PR #1332, opened 2023-09-12
junit - remove walrus for old systems

### LINKED ISSUE #1331: Testing fails due to walrus operator (`:=`)
Looks like there's a walrus operator `:=` in the `junit_common.py` file, which makes it fail for Python 3.6 (it was introduced in 3.8). 

We could probably add a test of the python version to make sure it's capable of handling that. Or just not use the walrus operator.

Thoughts? @zatkins-dev @jeremylt

jeremylt:
Fixes #1331 

Its mildly uglier code, but not by much.


jrwrigh (approved):
Thanks!

PR #1333, opened 2023-09-12
Add ElemRestrictionAtPoints

jeremylt:
In this PR we start adding the interface and CPU implementation for `CeedElemRestrictionApplyAtPoints`


interface/ceed-elemrestriction.c
```diff
@@ -638,11 +680,73 @@ int CeedElemRestrictionCreateStrided(Ceed ceed, CeedInt num_elem, CeedInt elem_s
...
+  The offsets array is arranged as
+
+  element_0_start_index
+  element_1_start_index
+  ...
+  element_n_start_index
+  element_n_stop_index
+  element_0_point_0
+  element_0_point_1
+  ...
```
jeremylt, line 696: Of note here - I'm using a slightly different convention than in the PETSc example - the CeedElemRestriction knows the num_elem, so I removed that from the offsets array here.

backends/ref/ceed-ref-restriction.c
```diff
@@ -392,6 +416,9 @@ static inline int CeedElemRestrictionApply_Ref_Core(CeedElemRestriction r, const
                                                              vv);
         }
         break;
+      case CEED_RESTRICTION_POINTS:
```
jeremylt, line 419: My plan here is to set this up so the ref impl makes a cell-ordered e-vec of all points, but in practice I think we'll restrict one cell at a time.

zatkins-dev:
This looks good to me!

jeremylt:
Ok, with the latest all the new logic should be tested (letting CodeCov run to double check) so if we like the design we can merge and move on to operators (the hard part)

backends/ref/ceed-ref-restriction.c
```diff
@@ -431,6 +458,9 @@ static inline int CeedElemRestrictionApply_Ref_Core(CeedElemRestriction rstr, co
                                                                                elem_size, v_offset, uu, vv));
         }
         break;
+      case CEED_RESTRICTION_POINTS:
+        CeedCallBackend(CeedElemRestrictionApplyAtPointsInElement_Ref(rstr, num_comp, start, stop, t_mode, uu, vv));
```
jrwrigh, line 462: Why does this call `CeedElemRestrictionApplyAtPointsInElement_Ref` and not `CeedElemRestrictionApplyAtPoints_Ref`?

backends/ref/ceed-ref-restriction.c
```diff
@@ -738,12 +781,16 @@ int CeedElemRestrictionCreate_Ref(CeedMemType mem_type, CeedCopyMode copy_mode,
...
-  CeedCallBackend(CeedSetBackendFunction(ceed, "ElemRestriction", rstr, "ApplyBlock", CeedElemRestrictionApplyBlock_Ref));
+  if (rstr_type == CEED_RESTRICTION_POINTS) {
+    CeedCallBackend(CeedSetBackendFunction(ceed, "ElemRestriction", rstr, "ApplyAtPointsInElement", CeedElemRestrictionApplyAtPoints_Ref));
+  } else {
+    CeedCallBackend(CeedSetBackendFunction(ceed, "ElemRestriction", rstr, "ApplyUnsigned", CeedElemRestrictionApplyUnsigned_Ref));
+    CeedCallBackend(CeedSetBackendFunction(ceed, "ElemRestriction", rstr, "ApplyUnoriented", CeedElemRestrictionApplyUnoriented_Ref));
+    CeedCallBackend(CeedSetBackendFunction(ceed, "ElemRestriction", rstr, "ApplyBlock", CeedElemRestrictionApplyBlock_Ref));
+    CeedCallBackend(CeedSetBackendFunction(ceed, "ElemRestriction", rstr, "GetOrientations", CeedElemRestrictionGetOrientations_Ref));
+    CeedCallBackend(CeedSetBackendFunction(ceed, "ElemRestriction", rstr, "GetCurlOrientations", CeedElemRestrictionGetCurlOrientations_Ref));
+  }
```
jrwrigh, line 792: I'm not sure I get this, as other "special" element restrictions (like the specially oriented ones) don't get different treatment.

jrwrigh: I guess it's this way because the restriction should only be used via `CeedElemRestrictionApplyAtPointsInElement` and not by anything else?

jrwrigh (approved):
If I understand correctly, this kind of restriction isn't supposed to be used inside a CeedOperator at all, right?

backends/ref/ceed-ref-restriction.c
jeremylt: Yeah, the use of the indices is fundamentally different in the case of the `AtPoints` restriction, while for all the other restrictions we are including or excluding orientation information but everything else means the same thing. I think it would be harmless to drop the `else` there.

backends/ref/ceed-ref-restriction.c
jeremylt: I can adjust the name for clarity, but `CeedElemRestrictionApplyAtPointsInElement_Ref` accepts a range of elements to restrict to.

backends/blocked/ceed-blocked-operator.c
```diff
@@ -97,6 +97,9 @@ static int CeedOperatorSetupFields_Blocked(CeedQFunction qf, CeedOperator op, bo
           CeedCallBackend(CeedElemRestrictionCreateBlockedStrided(ceed_rstr, num_elem, elem_size, block_size, num_comp, l_size, strides,
                                                                   &block_rstr[i + start_e]));
         } break;
+        case CEED_RESTRICTION_POINTS:
+          // Empty case - won't occur
+          break;
```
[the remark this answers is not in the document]
jeremylt: We have a check at the interface level. We don't usually repeat the same check deeper once its already been made.

backends/ref/ceed-ref-restriction.c
jeremylt: I pushed some renaming that hopefully makes things a bit clearer

backends/blocked/ceed-blocked-operator.c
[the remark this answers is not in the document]
jrwrigh: Fair. Just feels off to have an empty case statement.

backends/ref/ceed-ref-restriction.c
jrwrigh: Yep, those look good to me!

backends/ref/ceed-ref-restriction.c
jrwrigh: Gotcha. I'm good either way.

jrwrigh: I'll leave this unresolved as a reminder in case you did want to remove the `else`. Otherwise, feel free to resolve and merge.

jeremylt: I went ahead and dropped the `else`

PR #1315, opened 2023-08-31
Backend Consistency Fixes

jeremylt:
This is a follow-up to https://github.com/CEED/libCEED/pull/1312 applying the same changes to the backends

Generally I 

* Use `snake_case` consistently, like in the interface level

* Pull almost all variable declarations up to the top of their scope

* Use `const` where able

* Limit variable's scopes where practical (could do more of this in the future)

PR #1335, opened 2023-09-14
Fix some missing backend consistency from #1315

sebastiangrimberg:
#1315 made some changes to the CUDA backends without addressing similar spots in the HIP ones. This PR catches some of those (surely not all, but a step in the right direction).

This PR also introduces pointer alignment formatting via clang-format (previously this was a source of inconsistency across backends).

### DIFF
--- a/include/ceed/backend.h
+++ b/include/ceed/backend.h
@@ -219,7 +219,7 @@ CEED_EXTERN int CeedGetDelegate(Ceed ceed, Ceed *delegate);
 CEED_EXTERN int CeedSetDelegate(Ceed ceed, Ceed delegate);
 CEED_EXTERN int CeedGetObjectDelegate(Ceed ceed, Ceed *delegate, const char *obj_name);
 CEED_EXTERN int CeedSetObjectDelegate(Ceed ceed, Ceed delegate, const char *obj_name);
-CEED_EXTERN int CeedGetOperatorfallback_resource(Ceed ceed, const char **resource);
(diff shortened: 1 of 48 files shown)


jeremylt (commented):
LGTM, thanks!

include/ceed/backend.h
```diff
@@ -219,7 +219,7 @@ CEED_EXTERN int CeedGetDelegate(Ceed ceed, Ceed *delegate);
 CEED_EXTERN int CeedSetDelegate(Ceed ceed, Ceed delegate);
 CEED_EXTERN int CeedGetObjectDelegate(Ceed ceed, Ceed *delegate, const char *obj_name);
 CEED_EXTERN int CeedSetObjectDelegate(Ceed ceed, Ceed delegate, const char *obj_name);
-CEED_EXTERN int CeedGetOperatorfallback_resource(Ceed ceed, const char **resource);
+CEED_EXTERN int CeedGetOperatorFallbackResource(Ceed ceed, const char **resource);
```
sebastiangrimberg, line 222: Let me know if this change was intentional, but it looks to me like a find and replace bug so I reverted it.

jeremylt: Crap, yeah. That was a goof on my part

jeremylt (approved):

sebastiangrimberg:
Thanks for the fast review @jeremylt. I caught a few more misses, see the final three commits. If everything looks good to you, go ahead and merge as desired. Thank you!

jeremylt (approved):

PR #1338, opened 2023-09-18
fluids: Use explicit array size for `new_bounds`

jrwrigh:
- `new_bounds` was previously changed to remove the explicit array size declaration because Sycl's online compiler didn't like it.
 - Trying to do a direct cast results in: 
 
 ``` 
/home/jrwrigh/software/libCEED/examples/fluids/problems/../qfunctions/sgs_dd_model.h:67:39: error: used type 'const CeedScalar[6][2]' (aka 'const double[6][2]') where arithmetic or pointer type is required
  const CeedScalar new_bounds[6][2] = (const CeedScalar[6][2]) & sgsdd_ctx->data[sgsdd_ctx->offsets.out_scaling];
                                      ^                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
- This represents the best compromise; can't do a direct pointer cast  (requiring more compiler work to be efficient and obfuscating what's  happening), but can at least keep the explicit array sizes for  `new_bounds`


jeremylt (approved):
I am starting to dislike the SYCL compiler more and more

jrwrigh:
In fairness, it's really the online OpenCL compiler. SYCL might not run into these issues, but it doesn't have an online compiler to my understanding.

PR #1339, opened 2023-09-19
Spelling is hard

jeremylt:
We should really use spell-check on every PR.


jrwrigh (approved):
Never used it, but maybe [Codespell](https://github.com/codespell-project/codespell) would be useful?

PR #1340, opened 2023-09-20
Fix CEED_COPY_VALUES for ref rstr at points

jeremylt:
Previously, because `elem_size == 0`, using `CEED_COPY_VALUES` for an "AtPoints' restriction would result in attempting to read from unallocated memory.

### DIFF
--- a/interface/ceed-elemrestriction.c
+++ b/interface/ceed-elemrestriction.c
@@ -1272,7 +1272,7 @@ int CeedElemRestrictionGetElementSize(CeedElemRestriction rstr, CeedInt *elem_si
 
   @return An error code: 0 - success, otherwise - failure
 
-  @ref Backend
+  @ref User
 **/
 int CeedElemRestrictionGetNumPoints(CeedElemRestriction rstr, CeedInt *num_points) {
   Ceed ceed;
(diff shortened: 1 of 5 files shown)


interface/ceed-elemrestriction.c
```diff
@@ -1289,6 +1263,54 @@ int CeedElemRestrictionGetElementSize(CeedElemRestriction rstr, CeedInt *elem_si
...
+/**
+
+  @brief Get the number of points in the l-vector for a points CeedElemRestriction
+
+  @param[in]  rstr       CeedElemRestriction
+  @param[out] num_points The number of points in the l-vector
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
```
jrwrigh, line 1275:
```suggestion
  @ref User
```

interface/ceed-elemrestriction.c
```diff
@@ -1289,6 +1263,54 @@ int CeedElemRestrictionGetElementSize(CeedElemRestriction rstr, CeedInt *elem_si
...
+
+  @brief Get the number of points in an element of a points CeedElemRestriction
+
+  @param[in]  rstr       CeedElemRestriction
+  @param[in]  elem       Index number of element to retrieve the number of points for
+  @param[out] num_points The number of points in the element at index elem
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
```
jrwrigh, line 1298:
```suggestion
  @ref User
```

jrwrigh (approved):

PR #1341, opened 2023-09-20
Avoid VLA for assembled element matrices in `CeedSingleOperatorAssemble`

sebastiangrimberg:
This avoids a potentially very large (10^4 or 10^6 nnz) stack allocation with VLA for CPU-based full assembly when testing assembly of high-order operators (expected to be slow). This was frustrating to debug when I ran into this running with many threads, and the OpenMP default stack size per thread is often much smaller than the system setting.


jeremylt (approved):
LGTM

Issue #1342, opened 2023-09-21
Mixed Diagonal Assembly


jeremylt:
Unfortunately #1316 breaks critical functionality for Ratel - we need to be able to assemble diagonals with different numbers of active bases.

See #1150 

We either need to revert or fix this before we release.

jeremylt:
It looks non-trivial to fix this, so for now I'm making a PR to revert the regression.

sebastiangrimberg:
The changes to diagonal assembly might be able to be reverted without reverting the entire PR. Let me see if I can take a look at this, I think I see the case which I had not anticipated. Prior to #1316 the diagonal code looped over `num_active_bases` which was the input and output bases together. So now that these are separate into `active_bases_in` and `active_bases_out`, I think we just need to modify that loop in `CeedSingleOperatorAssembleAddDiagonal_Core` without any changes to the new assembly data structures.

I haven't used Ratel but seems like it might be a good prerequisite to test that when developing, changes in libCEED don't break Ratel. Is there a way we can tes this in the CI?

jedbrown:
I think this is a viable way to run a Ratel pipeline from within a libCEED pipeline (probably after the others). https://docs.gitlab.com/ee/ci/pipelines/downstream_pipelines.html

@sebastiangrimberg If you can try the change you have in mind, we can hold off on merging the revert PR #1343.

sebastiangrimberg:
OK thanks for the chance to fix it. Question here for my understanding based on the original PR #1150:

If there is an active basis for the operator which is associated with the input but not output (or output but not input), then from my reading `num_eval_modes_out/in` will be zero for this basis as constructed in `CeedAssemblyDataCreate`. This means there's no contribution to the diagonal in the loop in `CeedSingleOperatorAssembleAddDiagonal_Core`. Is this correct?

So the current change from #1343 is "correct" in the sense that if the `active_bases_in` and `active_bases_out` arrays are sorted so that the first `min(num_active_bases_in, num_active_bases_out)` elements point to matching basis objects, we can just loop for `b = 0, b < min(num_active_bases_in, num_active_bases_out), b++` iterations.

Can you confirm my understanding of the use case of the multiple active bases with where the input and output active bases counts differ?

jeremylt:
This is the offending operator:
```
CeedOperator - platen jacobian - face 5 orientation 4
  16 elements with 9 quadrature points each
  6 fields
  5 input fields:
    Input field 0:
      Name: "q data"
[...]

PR #1343, opened 2023-09-21
Revert regression for Ratel

jeremylt:
See #1342 

We cannot break Ratel before the release.


jeremylt:
Close in favor of #1344

PR #1344, opened 2023-09-22
Fix regression in multiple active bases diagonal assembly (alternative to #1343)

### LINKED ISSUE #1342: Mixed Diagonal Assembly

sebastiangrimberg:
Resolves #1342.


jedbrown (approved):
Ratel tests pass for me with this, thanks.

PR #1345, opened 2023-09-22
fluids: Add ASAN, Correct memory out-of-bounds error

jrwrigh:
- Add address sanitizer stuff to the `Makefile` for fluids
- Fix up an out-of-bounds memory error that I found once doing that above.


jeremylt (approved):
Thanks for adding that. Probably it should be our standard nowadays

If Noether's happy, I'm happy

PR #1347, opened 2023-09-25
ctx - only update in CeedQFunctionContextSet* when data is different

### LINKED ISSUE #1276: CeedOperatorSetContext* shoud check whether value needs updating
ie. Do what `UpdateContextLabel` does in the fluids example: https://github.com/CEED/libCEED/blob/54202681a900bdc61f3ce53842046e235d25108b/examples/fluids/src/setupts.c#L70-L91

jeremylt:
Fixes #1276 

This should trim out a few host->device updates


jrwrigh (approved):
Thanks for this!

PR #1351, opened 2023-09-28
Makefile fixes

sebastiangrimberg:
- Fix missing install for JiT headers for `magma` and `sycl` JiT headers
- Fix MAGMA HIP include flags


sebastiangrimberg:
Looks like an unrelated CI error with builds using PETSc? Indeed this is failing on `main` too after https://github.com/CEED/libCEED/pull/1284.

jeremylt:
Just merged a fix to main

jeremylt (approved):
@jedbrown's Make-foo is stronger than mine, but this looks good to me. CI doesn't install libCEED and test the install - I'm guessing you found this via some manual testing involving installing libCEED and this fixed the issues you had?

sebastiangrimberg:
Yep, if you try to run using MAGMA backend, for example, on an installed version of libCEED you'll get missing JiT source files errors during the RTC (from `CeedGetJitAbsolutePath`).

PR #1352, opened 2023-09-28
Set interface only tests to only='cpu'

jeremylt:
I set the interface level tests to only run on the CPU. This should marginally speed up CI and validate the infrastructure to only run a subset of the fluids test on GPU to significantly speed up CI.


tests/junit_common.py
```diff
@@ -213,11 +213,12 @@ def parse_test_line(line: str) -> TestSpec:
     raw_test_args: str = args[0][args[0].index('TESTARGS(') + 9:args[0].rindex(')')]
     # transform 'name="myname",only="serial,int32"' into {'name': 'myname', 'only': 'serial,int32'}
     test_args: dict = dict([''.join(t).split('=') for t in re.findall(r"""([^,=]+)(=)"([^"]*)\"""", raw_test_args)])
+    name: str = test_args.get('name', '')
```
jrwrigh, line 216: Do we want to allow for anonymous tests? I guess for unit tests that makes sense.

jrwrigh (approved):

tests/junit_common.py
jeremylt: Yeah, most of our tests have no name. I added this option because I found myself repeating very similar names over and over again for the vector state value and access lock tests and those names weren't adding much value.

jrwrigh: Cool. Makes sense.

PR #1353, opened 2023-09-28
CodeCov Fixes

jeremylt:
Here are some minor CodeCov fixes. It sucks that CodeCov produces non-deterministic results so we dig into the discrepancies less than we really should.

### DIFF
--- a/interface/ceed-jit-tools.c
+++ b/interface/ceed-jit-tools.c
@@ -243,7 +243,7 @@ int CeedPathConcatenate(Ceed ceed, const char *base_file_path, const char *relat
   @brief Find the relative filepath to an installed JiT file
 
   @param[in]  absolute_file_path Absolute path to installed JiT file
-  @param[out] relative_file_path Relative path to installed JiT file
+  @param[out] relative_file_path Relative path to installed JiT file, a substring of the absolute path
 
   @return An error code: 0 - success, otherwise - failure
 
@@ -260,7 +260,7 @@ int CeedGetJitRelativePath(const char *absolute_file_path, const char **relative
 
   @param[in]  ceed               Ceed object for error handling
   @param[in]  relative_file_path Relative path to installed JiT file
-  @param[out] absolute_file_path String buffer for absolute path to target file
+  @param[out] absolute_file_path String buffer for absolute path to target file, to be freed by caller
 
   @return An error code: 0 - success, otherwise - failure
 
@@ -287,7 +287,9 @@ int CeedGetJitAbsolutePath(Ceed ceed, const char *relative_file_path, char **abs
     CeedCall(CeedCheckFilePath(ceed, *absolute_file_path, &is_valid));
 
     if (is_valid) return CEED_ERROR_SUCCESS;
+    // LCOV_EXCL_START
     else CeedCall(CeedFree(absolute_file_path));
+    // LCOV_EXCL_STOP
   }
   // LCOV_EXCL_START
   return CeedError(ceed, CEED_ERROR_MAJOR, "Couldn't find matching JiT source file: %s", relative_file_path);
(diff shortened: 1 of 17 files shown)


interface/ceed-jit-tools.c
```diff
@@ -287,7 +287,9 @@ int CeedGetJitAbsolutePath(Ceed ceed, const char *relative_file_path, char **abs
     CeedCall(CeedCheckFilePath(ceed, *absolute_file_path, &is_valid));
 
     if (is_valid) return CEED_ERROR_SUCCESS;
+    // LCOV_EXCL_START
     else CeedCall(CeedFree(absolute_file_path));
```
jrwrigh, line 291: Not sure I understand why we'd want to free the array, but only when there's an error.

interface/ceed-qfunction.c
```diff
@@ -743,7 +743,9 @@ int CeedQFunctionAddInput(CeedQFunction qf, const char *field_name, CeedInt size
     CeedCheck(strcmp(field_name, qf->input_fields[i]->field_name), qf->ceed, CEED_ERROR_MINOR, "QFunction field names must be unique");
   }
   for (CeedInt i = 0; i < qf->num_output_fields; i++) {
+    // LCOV_EXCL_START
     CeedCheck(strcmp(field_name, qf->output_fields[i]->field_name), qf->ceed, CEED_ERROR_MINOR, "QFunction field names must be unique");
+    // LCOV_EXCL_STOP
   }
```
jrwrigh, line 749: Is this never called because we always set inputs before outputs? (thus `num_output_fields` is always zero and the loop never iterates)

If that's true, would it be better to flip the input/output calling order of a test rather than exclude this section of code?

tests/t408-qfunction.c
```diff
@@ -26,10 +26,12 @@ int main(int argc, char **argv) {
...
   CeedQFunctionContextGetDataRead(ctx, CEED_MEM_HOST, &ctx_data_copy);
   CeedQFunctionContextGetData(ctx, CEED_MEM_HOST, &ctx_data_copy);
+  // LCOV_EXCL_START
   CeedQFunctionContextRestoreData(ctx, &ctx_data_copy);
   CeedQFunctionContextRestoreDataRead(ctx, &ctx_data_copy);
 
   CeedQFunctionContextDestroy(&ctx);
   CeedDestroy(&ceed);
   return 0;
+  // LCOV_EXCL_STOP
```
jrwrigh, line 36: Codecov doesn't acknowledge these functions being called?

jrwrigh (approved):
Minor questions/suggestions.

tests/t408-qfunction.c
jeremylt: No, because line 28 generates an error

interface/ceed-jit-tools.c
jeremylt: It isn't freeing when there is an error exactly. `CeedCheckFilePath` is seeing if we generated a real filepath from the given relative path and the current root dir we are checking it with. If the generated absolute path isn't valid, then we need to free the generate (bad) absolute path and make a new one with the next root dir

interface/ceed-qfunction.c
jeremylt: Yeah.

6 of one, half dozen of the other. We could move the loop, but why always do more work?

tests/t408-qfunction.c
jrwrigh: Ahhhhh. Gotcha.

interface/ceed-qfunction.c
jrwrigh: I don't follow what you mean by moving the loop. I'm not suggesting changing the function, just a test. Like for t523, change:
https://github.com/CEED/libCEED/blob/ed5267d9dde71d634ebc200fed11e1feb22e46ab/tests/t523-operator.c#L87-L90

to

```c
  CeedQFunctionCreateInterior(ceed, 1, mass, mass_loc, &qf_mass_tet);
  CeedQFunctionAddInput(qf_mass_tet, "rho", 1, CEED_EVAL_NONE);
  CeedQFunctionAddOutput(qf_mass_tet, "v", 1, CEED_EVAL_INTERP);
  CeedQFunctionAddInput(qf_mass_tet, "u", 1, CEED_EVAL_INTERP);
```

PR #1357, opened 2023-10-02
docs: Clarify QFunction source compatability

jrwrigh:
Documentation for the source code compatibility for QFunctions is not general enough. It's only specifies c99, c++11, and CUDA, but HIP and OpenCL/SYCL compatibility are (possibly) needed there too. 

Currently changed it to just offer examples of what's needed (as the list of possible backends may expand in the future). Maybe C99 should be explicitly mentioned? Open to whatever suggestions (or just throwing out what I wrote and doing something else).

### DIFF
--- a/interface/ceed-qfunction.c
+++ b/interface/ceed-qfunction.c
@@ -573,7 +573,10 @@ int CeedQFunctionGetFlopsEstimate(CeedQFunction qf, CeedSize *flops) {
   @param[in]  f          Function pointer to evaluate action at quadrature points.
                            See \ref CeedQFunctionUser.
   @param[in]  source     Absolute path to source of QFunction, "\abs_path\file.h:function_name".
-                           For support across all backends, this source must only contain constructs supported by C99, C++11, and CUDA.
+                           The entire source file must only contain constructs supported by all targeted backends (i.e. CUDA for `/gpu/cuda`,
+                           OpenCL/SYCL for `/gpu/sycl`, etc.).
+                           The entire contents of this file and all locally included files are used during JiT compilation for GPU backends.
+                           All source files must be at the provided filepath at runtime for JiT to function.
   @param[out] qf         Address of the variable where the newly created CeedQFunction will be stored
 
   @return An error code: 0 - success, otherwise - failure


interface/ceed-qfunction.c
```diff
@@ -573,7 +573,7 @@ int CeedQFunctionGetFlopsEstimate(CeedQFunction qf, CeedSize *flops) {
   @param[in]  f          Function pointer to evaluate action at quadrature points.
                            See \ref CeedQFunctionUser.
   @param[in]  source     Absolute path to source of QFunction, "\abs_path\file.h:function_name".
-                           For support across all backends, this source must only contain constructs supported by C99, C++11, and CUDA.
+                           This source must only contain constructs supported by the backends desired to run with (i.e. CUDA for `/gpu/cuda`, OpenCL/SYCL for `/gpu/sycl`).
```
jeremylt, line 576:
```suggestion
                           This source must only contain constructs supported by all targeted backends (i.e. CUDA for `/gpu/cuda`, OpenCL/SYCL for `/gpu/sycl`).
```

Maybe something like this? "desired to run with" is a bit awk to my eye

jrwrigh: Ah, that works better. It felt awkward when I wrote it, but couldn't figure out better phrasing.

interface/ceed-qfunction.c
```diff
@@ -573,7 +573,7 @@ int CeedQFunctionGetFlopsEstimate(CeedQFunction qf, CeedSize *flops) {
   @param[in]  f          Function pointer to evaluate action at quadrature points.
                            See \ref CeedQFunctionUser.
   @param[in]  source     Absolute path to source of QFunction, "\abs_path\file.h:function_name".
-                           For support across all backends, this source must only contain constructs supported by C99, C++11, and CUDA.
+                           This source must only contain constructs supported by all targeted backends (i.e. CUDA for `/gpu/cuda`, OpenCL/SYCL for `/gpu/sycl`, etc.).
```
jeremylt, line 576: Should update to clarify that the entire contents of the file will be JiTed and must stand alone with all local includes

jrwrigh: Done.

jrwrigh:
I've fixed the reported style issue (turns out my Makefile is also not including `interface/*` files in `format-c` target for some reason). It's requiring breaking that line into two, which I know we try to avoid for documentation things. Should we change (somehow?) the clang format settings to allow arbitrary length comments?

interface/ceed-qfunction.c
```diff
@@ -573,7 +573,8 @@ int CeedQFunctionGetFlopsEstimate(CeedQFunction qf, CeedSize *flops) {
   @param[in]  f          Function pointer to evaluate action at quadrature points.
                            See \ref CeedQFunctionUser.
   @param[in]  source     Absolute path to source of QFunction, "\abs_path\file.h:function_name".
-                           For support across all backends, this source must only contain constructs supported by C99, C++11, and CUDA.
+                           The entire header file must only contain constructs supported by all targeted backends (i.e. CUDA for `/gpu/cuda`,
+                           OpenCL/SYCL for `/gpu/sycl`, etc.).
```
jeremylt, line 577:
```suggestion
                           OpenCL/SYCL for `/gpu/sycl`, etc.). The entire contents of this file and all locally included files are used during JiT compilation for GPU backends.
```

Something like this

interface/ceed-qfunction.c
```diff
@@ -573,7 +573,8 @@ int CeedQFunctionGetFlopsEstimate(CeedQFunction qf, CeedSize *flops) {
   @param[in]  f          Function pointer to evaluate action at quadrature points.
                            See \ref CeedQFunctionUser.
   @param[in]  source     Absolute path to source of QFunction, "\abs_path\file.h:function_name".
-                           For support across all backends, this source must only contain constructs supported by C99, C++11, and CUDA.
+                           The entire header file must only contain constructs supported by all targeted backends (i.e. CUDA for `/gpu/cuda`,
```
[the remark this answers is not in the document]
jrwrigh: Sure. I guess the important part is that the source is available at runtime. Should we explicitly mention that?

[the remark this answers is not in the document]
jeremylt: Yeah, maybe adding that to the last sentence?

interface/ceed-qfunction.c
```diff
@@ -573,7 +573,8 @@ int CeedQFunctionGetFlopsEstimate(CeedQFunction qf, CeedSize *flops) {
...
+                           OpenCL/SYCL for `/gpu/sycl`, etc.). The entire contents of this file and all locally included files are used during JiT compilation for GPU backends.
```
jrwrigh, line 577:
```suggestion
                           OpenCL/SYCL for `/gpu/sycl`, etc.). The entire contents of this file and all locally included files are used during JiT compilation for GPU backends and must be available at runtime.
```

Maybe "readable" instead of "available"?

jeremylt: Maybe 'All source files must be at the provided firepath at runtime for JiT to function.'

jrwrigh: Done.

jeremylt (approved):
ship it!

PR #1359, opened 2023-10-03
Improve CPU-based operator full assembly by using optimized backend tensor contraction

sebastiangrimberg:
This PR was initially motivated by noticing that the code in `CeedSingleOperatorAssemble` was much slower than I might have expected (even though it's expected to be pretty slow, especially at high order). It turns out that this was almost entirely spent in the naiive matrix-matrix product implementation from `CeedMatrixMatrixMultiply`. Instead, we can use the `CeedTensorContract` object from basis objects associated with the operator to get a big speedup. In my testing, going from `main` to this branch using the `xsmm` backend I decrease the time spent in a number of calls to the full assembly for prism elements up to p=4 from 25.6 s to 3.2 s.

Two other related improvements are included:
- ~~I tried to further clarify the fallback parent code from https://github.com/CEED/libCEED/pull/1281, where now multiple levels of fallback between backends should be supported.~~
- ~~I am not sure if this was intentional, so correct me if I'm wrong, but the `/cpu/self/opt/blocked` backend used to fall back to `ref/serial`. In this case, the QFunction assembly was using the `ref/serial` code instead of correctly falling back to the `opt/serial` code. I've updated the `opt/blocked` backend to fix this. This improves the assembly for all `*/blocked` backends like `avx` and `xsmm` because they fell back to `opt/blocked` and thus to `ref/serial` for QFunction assembly previously.~~

UPDATE: Both of the above points have been taken care of separately.

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -623,9 +623,11 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
 
   // Loop over elements and put in data structure
   // We store B_mat_in, B_mat_out, BTD, elem_mat in row-major order
-  CeedSize    count = 0;
-  CeedScalar *vals, *BTD_mat = NULL, *elem_mat = NULL, *elem_mat_b = NULL;
+  CeedSize           count = 0;
+  CeedTensorContract contract;
+  CeedScalar        *vals, *BTD_mat = NULL, *elem_mat = NULL, *elem_mat_b = NULL;
 
+  CeedCall(CeedTensorContractCreate(ceed, &contract));
   CeedCall(CeedCalloc(elem_size_out * num_qpts_in * num_eval_modes_in[0], &BTD_mat));
   CeedCall(CeedCalloc(elem_size_out * elem_size_in, &elem_mat));
   if (elem_rstr_curl_orients_in || elem_rstr_curl_orients_out) CeedCall(CeedCalloc(elem_size_out * elem_size_in, &elem_mat_b));
@@ -654,7 +656,8 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
         }
 
         // Form element matrix itself (for each block component)
-        CeedCall(CeedMatrixMatrixMultiply(ceed, BTD_mat, B_mat_in, elem_mat, elem_size_out, elem_size_in, num_qpts_in * num_eval_modes_in[0]));
+        CeedCall(CeedTensorContractApply(contract, 1, num_qpts_in * num_eval_modes_in[0], elem_size_in, elem_size_out, BTD_mat, CEED_NOTRANSPOSE,
+                                         false, B_mat_in, elem_mat));
 
         // Transform the element matrix if required
         if (elem_rstr_orients_out) {
@@ -716,6 +719,7 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
   CeedCall(CeedVectorRestoreArray(values, &vals));
 
   // Cleanup
+  CeedCall(CeedTensorContractDestroy(&contract));
   CeedCall(CeedFree(&BTD_mat));
   CeedCall(CeedFree(&elem_mat));
   CeedCall(CeedFree(&elem_mat_b));
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -623,8 +623,8 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
 
   // Loop over elements and put in data structure
   // We store B_mat_in, B_mat_out, BTD, elem_mat in row-major order
-  CeedSize           count = 0;
   CeedTensorContract contract;
+  CeedSize           count = 0;
   CeedScalar        *vals, *BTD_mat = NULL, *elem_mat = NULL, *elem_mat_b = NULL;
 
   CeedCall(CeedBasisGetTensorContract(basis_in, &contract));
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -627,7 +627,7 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
   CeedTensorContract contract;
   CeedScalar        *vals, *BTD_mat = NULL, *elem_mat = NULL, *elem_mat_b = NULL;
 
-  CeedCall(CeedTensorContractCreate(ceed, &contract));
+  CeedCall(CeedBasisGetTensorContract(basis_in, &contract));
   CeedCall(CeedCalloc(elem_size_out * num_qpts_in * num_eval_modes_in[0], &BTD_mat));
   CeedCall(CeedCalloc(elem_size_out * elem_size_in, &elem_mat));
   if (elem_rstr_curl_orients_in || elem_rstr_curl_orients_out) CeedCall(CeedCalloc(elem_size_out * elem_size_in, &elem_mat_b));
@@ -719,7 +719,6 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
   CeedCall(CeedVectorRestoreArray(values, &vals));
 
   // Cleanup
-  CeedCall(CeedTensorContractDestroy(&contract));
   CeedCall(CeedFree(&BTD_mat));
   CeedCall(CeedFree(&elem_mat));
   CeedCall(CeedFree(&elem_mat_b));
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -656,8 +656,12 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
         }
 
         // Form element matrix itself (for each block component)
-        CeedCall(CeedTensorContractApply(contract, 1, num_qpts_in * num_eval_modes_in[0], elem_size_in, elem_size_out, BTD_mat, CEED_NOTRANSPOSE,
-                                         false, B_mat_in, elem_mat));
+        if (contract) {
+          CeedCall(CeedTensorContractApply(contract, 1, num_qpts_in * num_eval_modes_in[0], elem_size_in, elem_size_out, BTD_mat, CEED_NOTRANSPOSE,
+                                           false, B_mat_in, elem_mat));
+        } else {
+          CeedCall(CeedMatrixMatrixMultiply(ceed, BTD_mat, B_mat_in, elem_mat, elem_size_out, elem_size_in, num_qpts_in * num_eval_modes_in[0]));
+        }
 
         // Transform the element matrix if required
         if (elem_rstr_orients_out) {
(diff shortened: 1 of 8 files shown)


interface/ceed-preconditioning.c
```diff
@@ -606,13 +614,15 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
...
+  // Assemble QFunction (backend version for op fallback parent is faster, if it exists)
+  CeedOperator        op_parent;
   CeedInt             layout_qf[3];
   const CeedScalar   *assembled_qf_array;
   CeedVector          assembled_qf        = NULL;
   CeedElemRestriction assembled_elem_rstr = NULL;
 
-  CeedCall(CeedOperatorLinearAssembleQFunctionBuildOrUpdate(op, &assembled_qf, &assembled_elem_rstr, CEED_REQUEST_IMMEDIATE));
+  CeedCall(CeedOperatorGetFallbackParent(op, &op_parent));
+  CeedCall(CeedOperatorLinearAssembleQFunctionBuildOrUpdate(op_parent, &assembled_qf, &assembled_elem_rstr, CEED_REQUEST_IMMEDIATE));
```
jeremylt, line 625: I would double check - I don't think we need this here? Isn't getting the parent part of this function call internally?

interface/ceed-preconditioning.c
```diff
@@ -688,9 +698,11 @@ static int CeedSingleOperatorAssemble(CeedOperator op, CeedInt offset, CeedVecto
 
   // Loop over elements and put in data structure
   // We store B_mat_in, B_mat_out, BTD, elem_mat in row-major order
-  CeedSize    count = 0;
-  CeedScalar *vals, *BTD_mat = NULL, *elem_mat = NULL, *elem_mat_b = NULL;
+  CeedSize           count = 0;
+  CeedTensorContract contract;
+  CeedScalar        *vals, *BTD_mat = NULL, *elem_mat = NULL, *elem_mat_b = NULL;
 
+  CeedCall(CeedBasisGetTensorContract(basis_in, &contract));
```
jeremylt, line 705: This is not guaranteed to work in general. We need some logic to handle when the basis doesn't have a TensorContract object. Anything in the interface level files must work for any backend.

interface/ceed-preconditioning.c
```diff
@@ -1748,27 +1761,10 @@ int CeedOperatorLinearAssembleQFunction(CeedOperator op, CeedVector *assembled,
...
 **/
 int CeedOperatorLinearAssembleQFunctionBuildOrUpdate(CeedOperator op, CeedVector *assembled, CeedElemRestriction *rstr, CeedRequest *request) {
-  int (*LinearAssembleQFunctionUpdate)(CeedOperator, CeedVector, CeedElemRestriction, CeedRequest *) = NULL;
-  CeedOperator op_assemble                                                                           = NULL;
-  CeedOperator op_fallback_parent                                                                    = NULL;
-
   CeedCall(CeedOperatorCheckReady(op));
 
-  // Determine if fallback parent or operator has implementation
-  CeedCall(CeedOperatorGetFallbackParent(op, &op_fallback_parent));
```
jeremylt, line 1758: Why remove this here?

interface/ceed-preconditioning.c
```diff
@@ -2579,7 +2576,8 @@ int CeedOperatorCreateFDMElementInverse(CeedOperator op, CeedOperator *fdm_inv,
 
   // Default interface implementation
   CeedCall(CeedOperatorGetCeed(op, &ceed));
-  CeedCall(CeedOperatorGetFallbackParentCeed(op, &ceed_parent));
+  CeedCall(CeedOperatorGetFallbackParent(op, &op_parent));
```
jeremylt, line 2579: This wouldn't be needed if the QF assembly logic wasn't changed. The intention was to use the fastest assembly process available any time the BuildOrAssemble function is called

interface/ceed-tensor.c
```diff
@@ -23,26 +23,25 @@
...
     CeedCall(CeedGetObjectDelegate(ceed, &delegate, "TensorContract"));
     CeedCheck(delegate, ceed, CEED_ERROR_UNSUPPORTED, "Backend does not support TensorContractCreate");
-    CeedCall(CeedTensorContractCreate(delegate, basis, contract));
+    CeedCall(CeedTensorContractCreate(delegate, contract));
     return CEED_ERROR_SUCCESS;
   }
 
   CeedCall(CeedCalloc(1, contract));
   CeedCall(CeedReferenceCopy(ceed, &(*contract)->ceed));
-  CeedCall(ceed->TensorContractCreate(basis, *contract));
```
jeremylt, line 45: Removing this argument probably belongs as a separate PR

interface/ceed-preconditioning.c
sebastiangrimberg: OK, is it sufficient then to move `CeedMatrixMatrixMultiply` back to a non-static function and just call that implementation when `contract = NULL`?

jeremylt: Yeah, that would probably be the simplest fix?

interface/ceed-preconditioning.c
sebastiangrimberg: Addressing this comment along with https://github.com/CEED/libCEED/pull/1359#discussion_r1344715992 and https://github.com/CEED/libCEED/pull/1359#discussion_r1344714266 here:

The logic in `CeedOperatorLinearAssembleQFunctionBuildOrUpdate` used to go up one parent level and check for `LinearAssembleQFunctionUpdate`. Now, `CeedOperatorGetFallbackParent` goes all the way to the top (bottom? haha) of the inheritance tree to allow for multiple levels of overrides. In doing so this means `CeedOperatorLinearAssembleQFunctionBuildOrUpdate` is now written in the same way as `CeedOperatorLinearAssembleQFunction`: Only fallbacks are searched for definitions, no parent promotion. In addition to enabling this multi-level inheritance, it also helps reduce some confusion between what the two methods are doing. I think it can be very confusing to a user if calling `CeedOperatorLinearAssembleQFunctionBuildOrUpdate` instead of `CeedOperatorLinearAssembleQFunction` directly leads to a different backend being called.

interface/ceed-tensor.c
sebastiangrimberg: Yeah sorry, this was needed in order to implement the approach in the first commit, but that ended up being reverted for the approach using `CeedBasisGetTensorContract`. I'll pull it out.

interface/ceed-preconditioning.c
jeremylt: The goal with BuildOrUpdate was to build the QData as fast as possible without the caller having the write the same parent chasing all over the place though. All the backends share the same assembly data because of how the fallback is set up.

sebastiangrimberg: I will refactor this to revert the need to perform the fallback parent search externally. That should go a long way towards improving your concern I think. The use case I have in mind, that breaks with the previous implementation, is that if we call `CeedOperatorLinearAssembleQFunctionBuildOrUpdate` having fallen back to `/cpu/self/ref/serial`, then only one backend level up was searched for having an `LinearAssembleQFunctionUpdate` definition. This would be incorrect if a backend which was two levels up in the fallback tree had a different definition. The code now handles this with `CeedOperatorGetFallbackParent`, but if no corresponding change was made to `CeedOperatorLinearAssembleQFunctionBuildOrUpdate` then this wouldn't work either since you can end up in infinite loops.

I will push a change shortly to have another go at this while addressing your concerns, give me a minute. Hopefully that will be an improvement.

interface/ceed-preconditioning.c
sebastiangrimberg: Fixed, see below.

interface/ceed-preconditioning.c
sebastiangrimberg: Fixed now

interface/ceed-preconditioning.c
sebastiangrimberg: Fixed, reverted

interface/ceed-tensor.c
sebastiangrimberg: Fixed, see https://github.com/CEED/libCEED/pull/1360

interface/ceed-preconditioning.c
jeremylt: Originally it was intended that there would only be two levels, the fallback and its parent. The fallback would only be created and be responsible for any preconditioning implementation the parent is missing.

sebastiangrimberg: Ah, this I did not realize. This isn't exactly the case anymore though, right? For CPU backends like `xsmm`, the first call to `CeedOperatorLinearAssemble` will go all the way back to `ref`, and then the parent points to `opt`, not `xsmm`, if I've traced this out correctly.

jeremylt: There are two different parents. The inheritance of ceed functionality it separate from the fallback

sebastiangrimberg: Right, I guess this is all not super important since the original reason why I was wanting to change this was to get access to the top level `Ceed` resource for creating a new `CeedTensorContract`. But that was a bad idea anyway and reverted. So I'll revert these changes here.

jeremylt: This confusion is part of why I don't love the fallback mechanism.

---

`CeedSetDelegate` is our way to manage inheritance.

`/cpu/self/*/xsmm` inherits from `/cpu/self/*/opt` for all objects this backend doesn't implement (Vector, ElemRestriction, Basis, QFunction, and Operator)

`/cpu/self/*/opt` inherits from `/cpu/self/*/ref` for all objects this backend doesn't implement (Vector, ElemRestriction, Basis, and QFunction)

then `/cpu/self/*/ref` implements everything.

This does have the result that the `/xsmm` backend will create an `/opt` CeedOperator and use Vectors, Bases, and QFunction from `/ref` and a `TensorContract` from `/xsmm`

This is all-or-nothing for each object though. Either the backend fully implements the object or it  doesn't and we track back up the inheritance chain.

---

The idea with the fallback for CeedOperators only was to add partial inheritance just for CeedOperator functionality to preconditioning.

This way the `CeedOperatorApply` would take the faster path of the implementation at the bottom of the inheritance chain, but then a clone of the CeedOperator would be created by the fallback Ceed for the missing preconditioning related implementations. If there is no fallback Ceed given, then the interface level implementation is used instead for missing preconditioning related CeedOperator implimentations.

Probably, the code is not best organized to communicate that intent.

sebastiangrimberg: Thanks a lot for the detailed description, it definitely does make sense. I've reverted all of the changes I had made with my incorrect interpretations, and even if the code isn't perfectly organized, it works so I have no complaints. Thanks a lot for your help!

jeremylt: That description probably should live... somewhere? I'll try to find a place in the dev notes for it

sebastiangrimberg: There's a brief few lines at the end of libCEEDdev.md which describes in less detail and without the usage context what you've said. I would say that copy and pasting you're few sentences into that `## Backend Inheritance` would be a great change.

jeremylt (approved):
LGTM!

How much faster were you seeing in practice?

sebastiangrimberg:
> How much faster were you seeing in practice?

It's a pretty substantial difference, especially for high orders. I noted this above, but going from `main` to this branch using the `xsmm` backend I decrease the time spent in a number of calls to the full assembly for prism elements up to p=4 (this is the time from several different COO assemblies) from 25.6 s to 3.2 s. I noticed this because I was comparing to MFEM's legacy assembly on CPU, and was noticing that the libCEED COO assembly + conversion to CSR was much slower in almost all cases. After this change, it's faster for every problem type, typically about 2x.

jeremylt:
> It's a pretty substantial difference, especially for high orders. I noted this above, but going from `main` to this branch using the `xsmm` backend I decrease the time spent in a number of calls to the full assembly for prism elements up to p=4 (this is the time from several different COO assemblies) from 25.6 s to 3.2 s. I noticed this because I was comparing to MFEM's legacy assembly on CPU, and was noticing that the libCEED COO assembly + conversion to CSR was much slower in almost all cases. After this change, it's faster for every problem type, typically about 2x.

Awesome! More speed => more betterer

jedbrown:
Great work @sebastiangrimberg !

PR #1362, opened 2023-10-04
Update  Delegation/Fallback Docs + CPU Usage

jeremylt:
This PR does two things

1) added an example of Delegation and Fallback in the developer docs

2) cleaned up (and fixed a small bug) with the use of Delegation/Fallback in the CPU backends

### DIFF
--- a/backends/blocked/ceed-blocked.c
+++ b/backends/blocked/ceed-blocked.c
@@ -16,8 +16,7 @@
 // Backend Init
 //------------------------------------------------------------------------------
 static int CeedInit_Blocked(const char *resource, Ceed ceed) {
-  Ceed       ceed_ref;
-  const char fallback_resource[] = "/cpu/self/ref/serial";
+  Ceed ceed_ref;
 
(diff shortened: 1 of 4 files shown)


backends/blocked/ceed-blocked.c
```diff
@@ -27,9 +26,6 @@ static int CeedInit_Blocked(const char *resource, Ceed ceed) {
   CeedCallBackend(CeedInit("/cpu/self/ref/serial", &ceed_ref));
   CeedCallBackend(CeedSetDelegate(ceed, ceed_ref));
 
-  // Set fallback Ceed resource for advanced operator functionality
-  CeedCallBackend(CeedSetOperatorFallbackResource(ceed, fallback_resource));
```
jeremylt, line 31: Note - in none of these 3 backends does the `/cpu/self/ref/serial` backend impliment any preconditioning related functionality that these backends don't implement themselves, so falling back and creating the cloned operator is just a waste of memory.

sebastiangrimberg: Isn't this needed for `CeedOperatorLinearAssembleQFunction_Blocked` and `CeedOperatorLinearAssembleQFunction_Opt`?

PR #1364, opened 2023-10-04
Update CeedContextFieldLabelGetDescription with offset

jeremylt:
In order to clone a `CeedQFunctionContext` fully, we need to clone the labels. To clone the labels, we need the offsets. Since this is only used by us, this interface update in minor.


rezgarshakeri (approved):

interface/ceed-qfunctioncontext.c
```diff
@@ -741,19 +741,21 @@ int CeedQFunctionContextGetAllFieldLabels(CeedQFunctionContext ctx, const CeedCo
 
   @param[in]  label             CeedContextFieldLabel
   @param[out] field_name        Name of labeled field
-  @param[out] field_description Description of field, or NULL for none
+  @param[out] field_offset      Offset of field register
```
jeremylt, line 744: Typo

PR #1367, opened 2023-10-05
Flag assembly data as stale if ctx data changed

jeremylt:
If context data changes, then the `QFunctionAssemblyData` for an operator will be stale.

Its easy for the user to forget this, so this is a bit of belt-and-suspenders.


jrwrigh (approved):
For my own understanding, I'm assuming it's just the `vec` member of `CeedQFunctionAssemblyData` that is invalid on context data changes?
https://github.com/CEED/libCEED/blob/533adc1e28eea3d12734d65713e13d9fda16729c/include/ceed-impl.h#L309-L317

jeremylt:
Yeah, this function call indicates that the data is the Vec is stale

jeremylt:
Yeah, this function call indicates that the data is the Vec is stale

PR #1368, opened 2023-10-06
Fix compiler vendor detection for `cc`

sebastiangrimberg:
On my RHEL machine, the output of `cc` is:

```
$ cc --version
cc (GCC) 7.3.1 20180712 (Red Hat 7.3.1-17)
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

This doesn't get picked up by the `CC_VENDOR` definition in the `Makefile`. This PR fixes this.


jeremylt (approved):
looks reasonable to me

PR #1371, opened 2023-10-11
Drop Potential Energy from Etot conservation variable

KennethEJansen:
Potential energy removed from all state helpers so no more x (or dx) dependence for them. Interior integrals still need x for IDL but boundary elements don't so removed.  This touches lots of qfunctions that make state calls and a few problems as well.

Boundary integrals also kept coordinate dependence (restored) since a few still need them.  They are setup for all as it is in a macro but those that don't need them don't extract them from `in`.

Body force added to energy equation.

As this changes the meaning of the 5th conservative state variable, a number of tests had to have their .bin files updated to the output of this branch (which of course means prior branches won't match the new bin files once this merges to main).


jedbrown (commented):
Can we update the documentation at this time (just moving potential energy into a body force)?

jedbrown:
There are some segfaults in the CI and significant changes compared to reference solutions (maybe because the body force isn't handled properly).

examples/fluids/qfunctions/freestream_bc.h
```diff
@@ -395,8 +393,7 @@ CEED_QFUNCTION_HELPER int Freestream_Jacobian(void *ctx, CeedInt Q, const CeedSc
                                               RiemannFluxType flux_type) {
   const CeedScalar(*dq)[CEED_Q_VLA] = (const CeedScalar(*)[CEED_Q_VLA])in[0];
   const CeedScalar(*q_data_sur)     = in[2];
-  const CeedScalar(*x)[CEED_Q_VLA]  = (const CeedScalar(*)[CEED_Q_VLA])in[3];
-  const CeedScalar(*jac_data_sur)   = in[4];
+  const CeedScalar(*jac_data_sur)   = in[3];
```
jrwrigh, line 396: I'm guessing this is probably one of the sources of a segfault. The inputs to the qfunctions haven't been changed, so `jac_data_sur` is probably still assigned to `in[4]`.

examples/fluids/qfunctions/freestream_bc.h
```diff
@@ -548,8 +542,7 @@ CEED_QFUNCTION_HELPER int RiemannOutflow_Jacobian(void *ctx, CeedInt Q, const Ce
   const CeedScalar(*dq)[CEED_Q_VLA] = (const CeedScalar(*)[CEED_Q_VLA])in[0];
   const CeedScalar(*Grad_dq)        = in[1];
   const CeedScalar(*q_data_sur)     = in[2];
-  const CeedScalar(*x)[CEED_Q_VLA]  = (const CeedScalar(*)[CEED_Q_VLA])in[3];
-  const CeedScalar(*jac_data_sur)   = in[4];
+  const CeedScalar(*jac_data_sur)   = in[3];
```
jrwrigh, line 545: Same here

```suggestion
  const CeedScalar(*jac_data_sur)   = in[4];
```

examples/fluids/qfunctions/freestream_bc.h
```diff
@@ -694,8 +683,7 @@ CEED_QFUNCTION_HELPER int PressureOutflow_Jacobian(void *ctx, CeedInt Q, const C
...
```
jrwrigh, line 686:
```suggestion
  const CeedScalar(*jac_data_sur)   = in[4];
```

examples/fluids/qfunctions/newtonian.h
```diff
@@ -419,8 +416,7 @@ CEED_QFUNCTION_HELPER int BoundaryIntegral_Jacobian(void *ctx, CeedInt Q, const
   const CeedScalar(*dq)[CEED_Q_VLA] = (const CeedScalar(*)[CEED_Q_VLA])in[0];
   const CeedScalar(*Grad_dq)        = in[1];
   const CeedScalar(*q_data_sur)     = in[2];
-  const CeedScalar(*x)[CEED_Q_VLA]  = (const CeedScalar(*)[CEED_Q_VLA])in[3];
-  const CeedScalar(*jac_data_sur)   = in[4];
+  const CeedScalar(*jac_data_sur)   = in[3];
```
jrwrigh, line 419:
```suggestion
  const CeedScalar(*jac_data_sur)   = in[4];
```

examples/fluids/qfunctions/velocity_gradient_projection.h
```diff
@@ -19,26 +19,23 @@ CEED_QFUNCTION_HELPER int VelocityGradientProjectionRHS(void *ctx, CeedInt Q, co
...
     for (CeedInt k = 0; k < 3; k++) {
-      CeedScalar dx_i[3] = {0}, dqi[5];
+      CeedScalar dqi[5];
       for (CeedInt j = 0; j < 5; j++) {
         dqi[j] = Grad_q[0][j][i] * dXdx[0][k] + Grad_q[1][j][i] * dXdx[1][k] + Grad_q[2][j][i] * dXdx[2][k];
       }
-      dx_i[k]   = 1.;
-      grad_s[k] = StateFromQ_fwd(context, s, dqi, x_i, dx_i, state_var);
+      grad_s[k] = StateFromQ_fwd(context, s, dqi, state_var);
     }
```
jrwrigh, line 39: Reminder to myself: This probably should use `StatePhysicalGradient*` if possible.

examples/fluids/src/setuplibceed.c
```diff
@@ -52,7 +52,6 @@ PetscErrorCode AddBCSubOperator(Ceed ceed, DM dm, CeedData ceed_data, DMLabel do
   PetscCallCeed(ceed, CeedOperatorSetField(op_apply_bc, "q", elem_restr_q_sur, ceed_data->basis_q_sur, CEED_VECTOR_ACTIVE));
   PetscCallCeed(ceed, CeedOperatorSetField(op_apply_bc, "Grad_q", elem_restr_q_sur, ceed_data->basis_q_sur, CEED_VECTOR_ACTIVE));
   PetscCallCeed(ceed, CeedOperatorSetField(op_apply_bc, "surface qdata", elem_restr_qd_i_sur, CEED_BASIS_NONE, q_data_sur));
-  PetscCallCeed(ceed, CeedOperatorSetField(op_apply_bc, "x", elem_restr_x_sur, ceed_data->basis_x_sur, ceed_data->x_coord));
```
jrwrigh, line 55: I thought we agreed to not drop the coordinates until the follow-up PR?

examples/fluids/qfunctions/freestream_bc.h
KennethEJansen: I did drop coordinates from the boundary integrals because their only need was for STATE.

KennethEJansen: The STG case does run properly on both Polaris and on Sunspot.

examples/fluids/src/setuplibceed.c
KennethEJansen: Not dropped from Volume integrals because they have the IDL need.

KennethEJansen:
Seg faults are gone but channel and densityCurrent are still not happy.  The latter does not surprise me but
```
Test: fluids-navierstokes Channel
  $ build/fluids-navierstokes -ceed /cpu/self/ref/serial -test_type solver -options_file examples/fluids/channel.yaml -compare_final_state_atol 2e-11 -compare_final_state_filename examples/fluids/tests-output/fluids-navierstokes-channel.bin -dm_plex_box_faces 5,5,1 -ts_max_steps 5
FAIL: stdout
Output: 
Test failed with error norm 0.00134024```
does.  As I look at the yaml and the inputs, it is periodic.  Usually periodic requires a body force and I don't see that set on the command line nor in the yaml.  Am I looking at the right channel.yaml in the examples dir?

**Comment by KennethEJansen** [comment]
I minimal pass on the docs change has been completed.

**Comment by jrwrigh** [comment]
>  Usually periodic requires a body force and I don't see that set on the command line nor in the yaml. Am I looking at the right channel.yaml in the examples dir?

You may have already figured it out, but `channel.c` calculates the required gravity vector for the given problem parameters.

**Comment by KennethEJansen** [comment]
I was really expecting that the failure of densityCurrent and channel cases was due to my failure to add the body force term to the energy equation but, they are still failing. The term is pretty simple (Dot3( s.U.momentum, g) ) for the 5th component of body_force in QFunction and IFunction and s->ds for IJacobian with the sign being the same as it is for momentum so I don't see how I could screw that up.  Further, there should not be any affect on boundary integrals.  I guess I will have to get into the debugger to see where the differences are. 

I assume the convergence test is measuring something "primitive" and does not involve the total energy which we have of course re-defined with this formulation?

**Comment by jrwrigh** [comment]
The files used for regression testing save the solution variables, whatever they may be. So for "Density Current" and "Channel", those errors are almost assuredly stemming from the change in the total energy definition.

For "Channel Primitive" though, that should be comparing apples-to-apples.

**Comment by KennethEJansen** [comment]
I rolled back to main and found that the prim var cases for my build are also failing so I decided to try to roll back on hash and try to just do a test and got this:
```
(base) kjansen@viz003: /projects/tools/libCEED/libCEED ((851646d1...))$ make test
make: 'lib' with optional backends: /cpu/self/memcheck/serial /cpu/self/memcheck/blocked /cpu/self/avx/serial /cpu/self/avx/blocked
          CC build/backends/avx/ceed-avx-blocked.o
          CC build/backends/avx/ceed-avx-serial.o
          CC build/backends/avx/ceed-avx-tensor.o
        LINK lib/libceed.so
      LINK.c build/t000-ceed
Traceback (most recent call last):
  File "tests/junit.py", line 2, in <module>
    from junit_common import *
  File "/projects/tools/libCEED/libCEED/tests/junit_common.py", line 58, in <module>
    class SuiteSpec(ABC):
  File "/projects/tools/libCEED/libCEED/tests/junit_common.py", line 134, in SuiteSpec
    def check_required_failure(self, test: str, spec: TestSpec, resource: str, stderr: str) -> tuple[str, bool]:
TypeError: 'type' object is not subscriptable
Makefile:634: recipe for target 'run-t000-ceed' failed
make: *** [run-t000-ceed] Error 1
```

@jrwrigh is this a conda python issue?

**Comment by KennethEJansen** [comment]
I was reproduce a passing channel case with main and "failing" with this branch.

I use quotes because I am not convinced it is an actual failure.

channel.c passes a zeroed x_1 coordinate to get a periodic initial condition but this is not really consistent with the state.U.Etot that are produced from the helper functions that then get thrown into the energy inviscid flux at the quadrature points.  Its not so easy to track the difference because that gets added to the Grad_v slot while the new energy equation body_force[4]=Dot3(s.momentum, g) goes into v.

The residual and the convergence are better with the new code.

I will next try to dump a solution to CGNS and pull them both up in paraview to see the real solution difference.

**Comment by KennethEJansen** [comment]
If I have not made a major miss-step, the following data suggests that the periodic channel with potential energy in its E_tot has a rather violent reaction to our attempts to make the solution periodic while the helper functions still had potential energy.
Top is the new code that has dropped potential energy out of all quadrature point work.  There is no change to the initial condition and zooms on the CGNS 0 step solution show the IC differs by 2e-12.  However as we step from 1, 2, 3, 4, 5 in various pictures we see this new branch converge to a periodic flow while main gets worse and worse.

Probably a good idea for someone else to repeat the test but I have repeated it (correctly or not) a few times and this is what I see.

Screen Shot 2023-10-12 at 9 22 31 PM
Screen Shot 2023-10-12 at 9 42 27 PM
Screen Shot 2023-10-12 at 9 43 06 PM
Screen Shot 2023-10-12 at 9 43 24 PM
Screen Shot 2023-10-12 at 9 43 55 PM
Screen Shot 2023-10-12 at 9 46 05 PM
Screen Shot 2023-10-12 at 9 46 20 PM


I suppose it is obvious that this is primitive variable channel.   I will try to also use paraview to assess whether density current differences are due to anything more than potential energy (e.g. look at Temperature and momentum and density maybe).

**Comment by KennethEJansen** [comment]
Here is one more picture with line plots 
Screen Shot 2023-10-12 at 10 20 16 PM
closer to the inflow.

**Comment by KennethEJansen** [comment]
I don't think I am am going to be able to "fix" density current as quickly as I had hoped.

I have committed one change as it dropped the reported error by 3 orders of magnitude but, as I mention in the commit statement, even if we get the IC right and the evolution right, if the error norm is measuring a different Etot it cannot pass.

**Comment by jedbrown** [comment]
I'm okay with re-baselining so long as we're confident the new version is as good or better. Sounds like above, you're talking about a conservative variable test, which we expect to be different even if the physical meaning of the solution is identical.

**Comment by KennethEJansen** [comment]
I think someone else should look at the periodic channel to be sure I have not crossed things up somehow but, if I have not goofed up something, that solution is better.  

I did not spend as much time on DC as I am falling behind on some other stuff but I think it is much closer now and yes, that is a conservative variable test.

**Comment by KennethEJansen** [comment]
For DC, the latest push has not gone through CI testing yet (is it down or just paused)?

**Comment by jrwrigh** [comment]
I'll double check everything in a bit.

I think the CI behavior you were seeing is that the Northern CI action doesn't run (sometimes?) until the others tests have run. Or at least there can be a delay between commit push and the GitLab CI starting.

**Comment by KennethEJansen** [comment]
As you can see from the last commit, I miss read what the DC-IC code was doing but can see now that it is setting primitive and thus does not need any change (reverted the change).

**Comment by KennethEJansen** [comment]
@jrwrigh lets divide and conquer.  I will get the style and compiler warnings cleaned up as I know how to do that and perhaps you can fix the testing (which I should learn at some point but perhaps not today).

**Comment by KennethEJansen** [comment]
I pushed a warning-free code. If it generates any style issues I will resolve them.  Just to be sure I understand the plan, are we going to use this branch to generate bin files for the DC and channel cases, copy them to the right spot, and then of course this will pass the tests?

I think @jrwrigh  5000 step run seems convincing that DC is doing the right thing  but I spent some time this morning trying understand some REALLY small differences in the 10 step "test" run with paraview profiles.  I see that @jrwrigh compared temperature so he must have also come across the issue that paraview z coordinate had to be scaled by 100 x (internal CEED units change I guess) to get internal+kinetic energy to match at step 0.

**Comment by jrwrigh** [comment]
> I pushed a warning-free code. If it generates any style issues I will resolve them. Just to be sure I understand the plan, are we going to use this branch to generate bin files for the DC and channel cases, copy them to the right spot, and then of course this will pass the tests?

Yep.

> I think @jrwrigh 5000 step run seems convincing that DC is doing the right thing but I spent some time this morning trying understand some REALLY small differences in the 10 step "test" run with paraview profiles. I see that @jrwrigh compared temperature so he must have also come across the issue that paraview z coordinate had to be scaled by 100 x (internal CEED units change I guess) to get internal+kinetic energy to match at step 0.

Yep, figured that out too.

File: examples/fluids/index.md [hunk 7]
```diff
@@ -13,22 +13,22 @@ Moreover, the Navier-Stokes example has been developed using PETSc, so that the
...
-In equations {eq}`eq-ns`, $\rho$ represents the volume mass density, $U$ the momentum density (defined as $\bm{U}=\rho \bm{u}$, where $\bm{u}$ is the vector velocity field), $E$ the total energy density (defined as $E = \rho e$, where $e$ is the total energy), $\bm{I}_3$ represents the $3 \times 3$ identity matrix, $g$ the gravitational acceleration constant, $\bm{\hat{k}}$ the unit vector in the $z$ direction, $k$ the thermal conductivity constant, $T$ represents the temperature, and $P$ the pressure, given by the following equation of state
+In equations {eq}`eq-ns`, $\rho$ represents the volume mass density, $U$ the momentum density (defined as $\bm{U}=\rho \bm{u}$, where $\bm{u}$ is the vector velocity field), $E$ the total energy density (defined as $E = \rho e$, where $e$ is the total energy including thermial and kinetic but not potential energy), $\bm{I}_3$ represents the $3 \times 3$ identity matrix, $\bm{b}$ is a body force vector (e.g., gravity vector),  $k$ the thermal conductivity constant, $T$ represents the temperature, and $P$ the pressure, given by the following equation of state
```
jrwrigh, line 28:
```suggestion
In equations {eq}`eq-ns`, $\rho$ represents the volume mass density, $U$ the momentum density (defined as $\bm{U}=\rho \bm{u}$, where $\bm{u}$ is the vector velocity field), $E$ the total energy density (defined as $E = \rho e$, where $e$ is the total energy including thermial and kinetic but not potential energy), $\bm{I}_3$ represents the $3 \times 3$ identity matrix, $\bm{b}$ is a body force vector (e.g., gravity vector, $\rho \bm{g} z$),  $k$ the thermal conductivity constant, $T$ represents the temperature, and $P$ the pressure, given by the following equation of state
```

tests/junit.py
```diff
@@ -26,6 +26,7 @@ def create_argparser() -> argparse.ArgumentParser:
     parser.add_argument('-n', '--nproc', type=int, default=1, help='number of MPI processes')
     parser.add_argument('-o', '--output', type=Optional[Path], default=None, help='Output file to write test')
     parser.add_argument('-b', '--junit-batch', type=str, default='', help='Name of JUnit batch for output file')
+    parser.add_argument('-np', '--pool-size', type=int, default=1, help='Number of test cases to run in parallel')
```
jrwrigh, line 29: I'm not sure what these change to `junit.py` and `junit_common.py` are, but they shouldn't be here. These lines of code are currently in `main`, so I'm not sure why they're showing up here. 

I think this just needs to be rebased onto `main`?

KennethEJansen:
I have updated the PR description to reflect the above conversations and changes.

jrwrigh (approved):
Looks good to me, just a small suggestion on the documentation.

examples/fluids/index.md
KennethEJansen: First, I see I miss spelled thermal (no i).  As for the suggested change,   there is no z there.
\bm{b} = \bm{g}.   It is fairly traditional (in aero and atmospheric flows)  to choose a coordinate system where the earths surface is x, y and thus g=0 0 -9.81 (as we apply in the DC inputs)  and then the ma=F of NS has a  -9.81*rho "forcing", F on m a_3. Shakib puts script F on the RHS of the equation and has it = rho{0, b_i, b_i u_i} and here b_3=-9.81.

Where this gets tricky is that in a residual, we pull it to the same side of equation and call it - script F which means momentum terms get -b_i, energy -b_i u_i.   As I was reviewing code I was seeing lots of negations of g.   I think the existing docs has -S instead of -F but I will take a look at that after I send this comment (learned the hard way a few times over the last couple of days that when I leave this page my comment disappears)

KennethEJansen: After reviewing the docs, I am fine with dropping the i from thermial to make thermal and just dropping " z" from your suggestion.  Care with the sign of components of g_i is always part or setting up problems properly and, as Evans accurately points out, verbosity in the attempt to be "utterly clear" is sometimes working against its objective and simple is often better.

PR #1374, opened 2023-10-13
fix(test): Use explicit `typing` objects for containers

jrwrigh:
- Using `list[str]` for type hints is incompatible with Python 3.7.9 (at least, possibly with newer versions as well)

Error reported by @KennethEJansen . TBH, I didn't even know that Python allowed to use `tuple/list` directly in type hints like that. I guess we're paying the burden of them not doing the obvious thing to begin with.

### DIFF
--- a/tests/junit.py
+++ b/tests/junit.py
@@ -134,7 +134,7 @@ class CeedSuiteSpec(SuiteSpec):
             return f'SYCL device type not available'
         return None
 
-    def check_required_failure(self, test: str, spec: TestSpec, resource: str, stderr: str) -> tuple[str, bool]:
+    def check_required_failure(self, test: str, spec: TestSpec, resource: str, stderr: str) -> Tuple[str, bool]:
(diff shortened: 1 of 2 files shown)


zatkins-dev:
I guess that form of type annotation was added in 3.9, so we can use the old `List` version for backward compatibility. 

Still, I would like to set a hard limit on what version of Python we support.

zatkins-dev:
Specifically, `dataclasses` was added in 3.7, so I would like to not support anything older than that.

zatkins-dev (approved):
Looks good to me!

jedbrown:
PETSc uses [vermin](https://pypi.org/project/vermin/) to check that Python syntax is supported by all the versions we target.

jedbrown (approved):
Thanks for the quick fix.

jeremylt:
Can you make the same PR in Ratel?

jrwrigh:
Sure.

PR #1379, opened 2023-10-13
fluids: Don't set ts_max_steps if not TESTTYPE_NONE

jrwrigh:
Arguments passed to `TESTARGS` can be used verbatim to create the data used for regression testing. Previously, the number of timesteps could (would) be different whether `-test_type solver` was given or not.


jeremylt (approved):

jeremylt:
Ultimately, I think it would be good to have a folder of test case ymls

jedbrown (approved):

PR #1380, opened 2023-10-13
Drop Coordinate Basis from IJacobian (without breaking IDL)

KennethEJansen:
On some platforms, there is a large performance drop for having coordinate basis in IJacobian.   This branch will compute IDL coefficient in IFunction that has access to coordinates, append this to jac_data so that IJacobian can avoid pulling coordinates through basis.

### DIFF
--- a/examples/fluids/qfunctions/newtonian.h
+++ b/examples/fluids/qfunctions/newtonian.h
@@ -20,9 +20,8 @@
 #include "stabilization.h"
 #include "utils.h"
 
-CEED_QFUNCTION_HELPER void InternalDampingLayer(const NewtonianIdealGasContext context, const State s, const CeedScalar x_i[3], CeedScalar damp_Y[5],
+CEED_QFUNCTION_HELPER void InternalDampingLayer(const NewtonianIdealGasContext context, const State s, const CeedScalar sigma, CeedScalar damp_Y[5],
                                                 CeedScalar damp_residual[5]) {
-  const CeedScalar sigma = LinearRampCoefficient(context->idl_amplitude, context->idl_length, context->idl_start, x_i[0]);
   ScaleN(damp_Y, sigma, 5);
   State damp_s = StateFromY_fwd(context, s, damp_Y);
 
@@ -231,8 +230,10 @@ CEED_QFUNCTION_HELPER int IFunction_Newtonian(void *ctx, CeedInt Q, const CeedSc
 
     for (CeedInt j = 0; j < 5; j++) v[j][i] = wdetJ * (U_dot[j] - body_force[j]);
     if (context->idl_enable) {
+      const CeedScalar sigma = LinearRampCoefficient(context->idl_amplitude, context->idl_length, context->idl_start, x_i[0]);
+      jac_data[14 * Q + i] = sigma;
       CeedScalar damp_state[5] = {s.Y.pressure - P0, 0, 0, 0, 0}, idl_residual[5] = {0.};
-      InternalDampingLayer(context, s, x_i, damp_state, idl_residual);
+      InternalDampingLayer(context, s,  sigma, damp_state, idl_residual);
       for (int j = 0; j < 5; j++) v[j][i] += wdetJ * idl_residual[j];
     }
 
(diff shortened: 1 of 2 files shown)


KennethEJansen:
This is 2 files and only about 10ish lines of code.  The place I could use some help is that currently have bumped jac_data by 1 whether we have IDL or not.  I assume we don't want that performance penalty but it is currently a declared constant so I guess we would need to make it not that so that as soon as we can use context->idl_enable to decide to bump it up?

KennethEJansen:
It looks like I finally got all the spaces in the right places for the style requirements.  

This is following exactly @jedbrown suggested mod. I  don't yet see a way to only increase the size of jac_data when IDL is needed so it is just on all the time.

KennethEJansen:
I think I figured out how to do the bump by 1 on jac_data only when idl_enable is true.

KennethEJansen:
FWIW, we are setting up restriction operators for jac_data without checking that it is implicit.  I am sure its not a big deal but just an observation.

KennethEJansen:
Passing all tests before the last style spacing change so I think this is now complete (assuming there is not a better way than pulling the context out to determine if idl_enabled.   I suppose we could also wrap the jac_data restriction in the same conditional to save a trivial amount of setup work for explicit problems (had to do it here as the context I am checking does not exist for explicit problems).

examples/fluids/src/setuplibceed.c
```diff
@@ -163,7 +163,16 @@ PetscErrorCode SetupLibceed(Ceed ceed, CeedData ceed_data, DM dm, User user, App
...
   // *****************************************************************************
   const PetscInt num_comp_q = 5;
-  const CeedInt  dim = problem->dim, num_comp_x = problem->dim, q_data_size_vol = problem->q_data_size_vol, jac_data_size_vol = num_comp_q + 6 + 3;
+  const CeedInt  dim = problem->dim, num_comp_x = problem->dim, q_data_size_vol = problem->q_data_size_vol;
+  CeedInt        jac_data_size_vol = num_comp_q + 6 + 3;
+
+  if (problem->apply_vol_ifunction.qfunction) {
+    NewtonianIdealGasContext gas;
+    PetscCallCeed(ceed, CeedQFunctionContextGetDataRead(problem->apply_vol_ifunction.qfunction_context, CEED_MEM_HOST, &gas));
+    jac_data_size_vol += (gas->idl_enable ? 1 : 0);
```
jrwrigh, line 172:
```suggestion
  CeedInt        jac_data_size_vol = num_comp_q + 6 + 3 + (gas->idl_enable ? 1 : 0);

  if (problem->apply_vol_ifunction.qfunction) {
    NewtonianIdealGasContext gas;
    PetscCallCeed(ceed, CeedQFunctionContextGetDataRead(problem->apply_vol_ifunction.qfunction_context, CEED_MEM_HOST, &gas));
```

examples/fluids/qfunctions/newtonian.h
```diff
@@ -231,8 +230,10 @@ CEED_QFUNCTION_HELPER int IFunction_Newtonian(void *ctx, CeedInt Q, const CeedSc
 
     for (CeedInt j = 0; j < 5; j++) v[j][i] = wdetJ * (U_dot[j] - body_force[j]);
     if (context->idl_enable) {
+      const CeedScalar sigma   = LinearRampCoefficient(context->idl_amplitude, context->idl_length, context->idl_start, x_i[0]);
+      jac_data[14 * Q + i]     = sigma;
```
jrwrigh, line 234:
```suggestion
       StoredValuesPack(Q, i, 14, 1, sigma, jac_data);
```

A bit overkill, but I'd like to storage and access it consistent.

examples/fluids/src/setuplibceed.c
```diff
@@ -408,7 +415,7 @@ PetscErrorCode SetupLibceed(Ceed ceed, CeedData ceed_data, DM dm, User user, App
   if (app_ctx->turb_spanstats_enable) PetscCall(TurbulenceStatisticsSetup(ceed, user, ceed_data, problem));
   if (app_ctx->diff_filter_monitor) PetscCall(DifferentialFilterSetup(ceed, user, ceed_data, problem));
 
-  PetscCallCeed(ceed, CeedElemRestrictionDestroy(&elem_restr_jd_i));
+  PetscCallCeed(ceed, CeedElemRestrictionDestroy(&elem_restr_jd_i));  // not used by explicit applications
```
jrwrigh, line 418: Not sure how I feel about adding the comment here. But here's a clarification.

```suggestion
  PetscCallCeed(ceed, CeedElemRestrictionDestroy(&elem_restr_jd_i));  // not used by explicit time integration
```

jrwrigh (approved):
A few minor style comments. Otherwise good to me!

examples/fluids/src/setuplibceed.c
KennethEJansen: hmmmm   I thought I had removed all of these after using a commit to simply surface the question about whether we wanted to wrap these in a conditional or not.  I will make sure they are fully removed  from this branch as they are not part of the primary objective.

KennethEJansen: Indeed I missed removing this one.  It has been removed now in the latest commit that reverted the StorePackedValues change so I am resolving this.

examples/fluids/src/setuplibceed.c
jrwrigh: Ope, you're right. Didn't noticed that `idl_enable` was coming from the QF context.

jrwrigh:
Yep, you're right. The line needs to be:
```c
StoredValuesPack(Q, i, 14, 1, &sigma, jac_data);
```

Issue #1373, opened 2023-10-11
Slowdown in PETSc multigrid example with tetrahedral elements and `/gpu/*/magma` backends


zatkins-dev:
The `petsc-multigrid` example performs substantially slower on magma backends when compared to other GPU/CPU backends for tetrahedral elements (~70 sec vs ~15 sec).

See this pipeline as an example: https://gitlab.com/libceed/libCEED/-/pipelines/1033906336/test_report
![image](https://github.com/CEED/libCEED/assets/42555448/f9a56e1b-8a1c-410a-b09b-67c86acd9fcc)

The test suite observes this issue with `/gpu/hip/magma*`, but I have also reproduced locally with `/gpu/cuda/magma*`.

zatkins-dev:
Possibly related to #1252 and may be fixed by #1301

PR #1382, opened 2023-10-17
Lazy RTC compilation of MAGMA basis kernels for small P, Q

sebastiangrimberg:
Refactored preliminary PR for #1301. Compiles custom non-tensor MAGMA kernels only when needed (during first call to CeedBasisApply).

Other included improvements to the `magma` backend:
- Renames files in the `magma` backend to match the other GPU backends in the library, along with other style formatting for consistency with other backends and the [style guide](https://libceed.org/en/latest/libCEEDdev/#style-guide) (variable casing, for example).
- Simplify some of the macro definitions in `ceed-magma.h` to eliminate duplication between tensor and non-tensor basis code.
- Add include guards to `jit-source/magma` files as done in #1299 for `cuda` and `hip` backends.
- Use JiT for the non-tensor weight kernels for the `CEED_EVAL_WEIGHT` in the `magma` backend (as for `cuda/ref` and `hip/ref`). 

TODO
- [x] Test if resolves https://github.com/CEED/libCEED/issues/1373


sebastiangrimberg:
Indeed this PR resolves #1373:

Before:

Screenshot 2023-10-17 at 2 45 22 PM

After:

Screenshot 2023-10-17 at 2 46 28 PM

backends/magma/ceed-magma-gemm-nontensor.cpp
```diff
@@ -5,7 +5,9 @@
 //
 // This file is part of CEED:  http://github.com/ceed
 
-#include "ceed-magma.h"
+#include "ceed-magma-gemm-nontensor.h"
+
+#include "ceed-magma-gemm-selector.h"
```
jeremylt, line 10: minor - did clang-format put that extra line break in there?

jeremylt (approved):
thanks for these changes!

backends/magma/ceed-magma-gemm-nontensor.cpp
sebastiangrimberg: No that was me following the style for header/definiton file pairs to include first the associated header with delcarations, followed by newline, then IWYU. The same approach is in `ceed-magma-gemm-selector.cpp`. Should I change it to group all local includes together in the same block?

jeremylt: I wasn't aware of such a convention - it doesn't look like we uniformly use it. I have no strong feelings

sebastiangrimberg:
Looks like the only CI issue is a -0.05% coverage change (noise). So I think this can be merged when ready.

PR #1384, opened 2023-10-20
Add Vermin in CI

### LINKED ISSUE #1375: Minimum Python version for testing
Stemming from #1374, #1332, and #1331.

I guess the main discussion points are:
0. Do we want to specify a minimum python version for testing? (I think we all agree yes, but stating it explicitly is probably important)

1. What minimum Python version should we set?
2. Do we want to implement [vermin](https://pypi.org/project/vermin/) in testing/CI to verify that version is set?

For 1, it looks like vermin now has the minimum required Python version at 3.6. Note that dataclasses can be backported to 3.6 via a [PyPI package](https://pypi.org/project/dataclasses/), so that's not a hard limitation there. (Plus somehow #1332 shows it working on 3.6 even though I didn't explicitly install `dataclasses`).

For 2, I don't see a reason why not. 

Thoughts?

jeremylt:
This adds `make vermin` and checks for Python 3.8+ compatibility. I used Python 3.8 since 3.7 and older has reached the EOL. We pass for 3.7 but not for 3.6, but I'm not that worried about either.

Closes #1375

### DIFF
--- a/.github/workflows/python-test-with-style.yml
+++ b/.github/workflows/python-test-with-style.yml
@@ -45,3 +45,9 @@ jobs:
         FC: gfortran-11
       run: |
         make format-py && git diff --exit-code
+    - name: Python version
+      env:
+        CC: ${{ matrix.compiler }}
+        FC: gfortran-11
+      run: |
+        make vermin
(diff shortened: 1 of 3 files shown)


jrwrigh (approved):
Thanks!

PR #1366, opened 2023-10-05
`magma` backend RTC autotuning

sebastiangrimberg:
NOTE: Based on https://github.com/CEED/libCEED/pull/1301

This PR adds functionality for automatically generating the `NB` selection data for the non-tensor MAGMA backend kernels. See the README.md in `backends/magma/tuning` for information on how to automatically generate the data in a header file for a new target device. In particular, there is a C++ program in `tuning.cpp` which runs the benchmarks for a predefined set of `P`, `Q`, and `N` values (currently some basic quadrature rules on H1 tet and tri elements as was the case before, but can be expanded), for a certain `NB`. This is called from a Python driver `generate_tuning.py` which computes the best value of `NB` over all possibilities and writes the header file.

TODO:
- [x] Generate tuning data for A100, MI250x, MI100 (currently only V100)
- [x] Check if this resolves AMD performance regression issues observed in https://github.com/CEED/libCEED/pull/1301

PR #1385, opened 2023-10-20
Improve transpose basis performance for `magma` backend

sebastiangrimberg:
In profiling it turns out that the cost of the `magma_dlaset` call to zero out `V` for the basis transpose application was quite expensive and the cause of the transpose kernels being 10-20% slower than the non-transpose ones. This PR removes these calls and modifies the relevant `magma` backend kernels to write to `V` instead of accumulating into it after it was set to zero.

NOTE: This PR currently targets #1301, and can be either merged in there or we can wait until #1301 is approved and merged to `main` before rebasing and merging this one.

### DIFF
--- a/backends/magma/ceed-magma-basis.c
+++ b/backends/magma/ceed-magma-basis.c
@@ -54,19 +54,6 @@ static int CeedBasisApply_Magma(CeedBasis basis, CeedInt num_elem, CeedTranspose
   else CeedCheck(e_mode == CEED_EVAL_WEIGHT, ceed, CEED_ERROR_BACKEND, "An input vector is required for this CeedEvalMode");
   CeedCallBackend(CeedVectorGetArrayWrite(v, CEED_MEM_DEVICE, &d_v));
 
-  // Clear v for transpose operation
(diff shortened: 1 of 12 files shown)


sebastiangrimberg:
Here is an example performance comparison for the non-tensor basis application with P = 20, Q = 24 (p = 3 tet) on a V100. This PR should also help improve performance for the `magma` tensor backends as well.

Before:

image

After:

image

nbeams:
Hi @sebastiangrimberg, I'm trying to get caught up on the activity with these PRs. I'd vote for going ahead and merging this into #1301 as you are currently proposing, then we can merge #1301 into main, then rebase #1366 with the updated kernels so that I can create/add the updated MI250X and A100 data to that PR. Is that what you had in mind? Or do you prefer to do #1366 before #1301?

sebastiangrimberg:
Sounds great. I will merge this into #1301 and we can proceed with final review there. Thanks @nbeams! Note the failing Python CI test is unrelated, should be fixed by https://github.com/CEED/libCEED/pull/1387.

PR #1390, opened 2023-11-02
OperatorAtPoints

jeremylt:
Super, duper WIP

- [x] Add interface function signatures
- [x] Update field update logic for AtPoints
- [x] Add basic CPU impl (probably easiest to do 1 element at a time like in opt)
- [x] Add basic t-5* tests
- ~~Convert PETSc DMSwarm example~~

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -618,8 +618,12 @@ static int CeedOperatorSetupFieldsAtPoints_Ref(CeedQFunction qf, CeedOperator op
         q_size = (CeedSize)max_num_points * size;
         CeedCallBackend(CeedVectorCreate(ceed, q_size, &q_vecs[i]));
         break;
-      case CEED_EVAL_WEIGHT:
-        // Not used at this time
+      case CEED_EVAL_WEIGHT:  // Only on input fields
+        CeedCallBackend(CeedOperatorFieldGetBasis(op_fields[i], &basis));
+        q_size = (CeedSize)max_num_points;
+        CeedCallBackend(CeedVectorCreate(ceed, q_size, &q_vecs[i]));
+        CeedCallBackend(
+            CeedBasisApplyAtPoints(basis, max_num_points, CEED_NOTRANSPOSE, CEED_EVAL_WEIGHT, CEED_VECTOR_NONE, CEED_VECTOR_NONE, q_vecs[i]));
         break;
     }
     if (is_input && e_vecs[i]) {
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -548,6 +548,409 @@ static int CeedOperatorLinearAssembleQFunctionUpdate_Ref(CeedOperator op, CeedVe
 }
 
 //------------------------------------------------------------------------------
+// Setup Input/Output Fields
+//------------------------------------------------------------------------------
+static int CeedOperatorSetupFieldsAtPoints_Ref(CeedQFunction qf, CeedOperator op, bool is_input, CeedVector *e_vecs_full, CeedVector *e_vecs,
+                                               CeedVector *q_vecs, CeedInt start_e, CeedInt num_fields, CeedInt Q) {
+  Ceed                ceed;
+  CeedSize            e_size, q_size;
+  CeedInt             max_num_points, num_comp, size, P;
+  CeedQFunctionField *qf_fields;
+  CeedOperatorField  *op_fields;
+
+  CeedCallBackend(CeedOperatorGetCeed(op, &ceed));
+  if (is_input) {
+    CeedCallBackend(CeedOperatorGetFields(op, NULL, &op_fields, NULL, NULL));
+    CeedCallBackend(CeedQFunctionGetFields(qf, NULL, &qf_fields, NULL, NULL));
+  } else {
+    CeedCallBackend(CeedOperatorGetFields(op, NULL, NULL, NULL, &op_fields));
+    CeedCallBackend(CeedQFunctionGetFields(qf, NULL, NULL, NULL, &qf_fields));
+  }
+
+  // Get max number of points
+  {
+    CeedInt             dim;
+    CeedElemRestriction rstr_points = NULL;
+    CeedOperator_Ref   *impl;
+
+    CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, NULL));
+    CeedCallBackend(CeedElemRestrictionGetMaxPointsInElement(rstr_points, &max_num_points));
+    CeedCallBackend(CeedElemRestrictionGetNumComponents(rstr_points, &dim));
+    CeedCallBackend(CeedElemRestrictionDestroy(&rstr_points));
+    CeedCallBackend(CeedOperatorGetData(op, &impl));
+    if (is_input) CeedCallBackend(CeedVectorCreate(ceed, dim * max_num_points, &impl->point_coords_elem));
+  }
+
+  // Loop over fields
+  for (CeedInt i = 0; i < num_fields; i++) {
+    CeedEvalMode eval_mode;
+    CeedBasis    basis;
+
+    CeedCallBackend(CeedQFunctionFieldGetEvalMode(qf_fields[i], &eval_mode));
+    if (eval_mode != CEED_EVAL_WEIGHT) {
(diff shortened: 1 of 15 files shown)


interface/ceed-operator.c
```diff
@@ -791,6 +834,50 @@ int CeedOperatorGetFields(CeedOperator op, CeedInt *num_input_fields, CeedOperat
...
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Advanced
+**/
+int CeedOperatorAtPointsSetPoints(CeedOperator op, CeedElemRestriction rstr_points, CeedVector point_coords) {
+  CeedCheck(op->is_at_points, op->ceed, CEED_ERROR_MINOR, "Only defined for operator at points");
+  CeedCheck(!op->is_immutable, op->ceed, CEED_ERROR_MAJOR, "Operator cannot be changed after set as immutable");
+
+  CeedCall(CeedElemRestrictionReferenceCopy(rstr_points, &op->rstr_points));
+  CeedCall(CeedVectorReferenceCopy(point_coords, &op->point_coords));
```
jeremylt, line 855: Note to self - I need to update the destroy for these!

interface/ceed-operator.c
```diff
@@ -791,6 +834,50 @@ int CeedOperatorGetFields(CeedOperator op, CeedInt *num_input_fields, CeedOperat
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Set the arbitrary points in each element for a CeedOperatorAtPoints.
+
+  Note: Calling this function asserts that setup is complete and sets the CeedOperator as immutable.
+
+  @param[in,out] op           CeedOperatorAtPoints
+  @param[in]     rstr_points  CeedElemRestriction for points in each element
```
jedbrown, line 843: Looks good overall. I think we'll want a more explicit statement of what this restriction is.

jeremylt: Agreed - I'll make this PR a priority tomorrow

jeremylt:
Ok, I have hacked in the rough internals. Untested, undebugged, etc

interface/ceed-operator.c
```diff
@@ -758,14 +802,14 @@ int CeedOperatorSetField(CeedOperator op, const char *field_name, CeedElemRestri
     op->has_restriction = true;  // Restriction set, but num_elem may be 0
   }
   CeedCall(CeedBasisReferenceCopy(b, &(*op_field)->basis));
-  if (op->num_qpts == 0) op->num_qpts = num_qpts;
+  if (op->num_qpts == 0 && !op->is_at_points) op->num_qpts = num_qpts;  // no consistent number of qpts for OperatorAtPoints
```
jeremylt, line 805: Note (for somewhere around here) - for an AtPoints operator, only a AtPoints restriction can be CEED_BASIS_NONE + CEED_EVAL_NONE

interface/ceed-operator.c
```diff
@@ -791,6 +839,50 @@ int CeedOperatorGetFields(CeedOperator op, CeedInt *num_input_fields, CeedOperat
...
+
+  @ref Advanced
+**/
+int CeedOperatorAtPointsSetPoints(CeedOperator op, CeedElemRestriction rstr_points, CeedVector point_coords) {
+  CeedCheck(op->is_at_points, op->ceed, CEED_ERROR_MINOR, "Only defined for operator at points");
+  CeedCheck(!op->is_immutable, op->ceed, CEED_ERROR_MAJOR, "Operator cannot be changed after set as immutable");
+
+  CeedCall(CeedElemRestrictionReferenceCopy(rstr_points, &op->rstr_points));
+  CeedCall(CeedVectorReferenceCopy(point_coords, &op->point_coords));
+  return CEED_ERROR_SUCCESS;
```
zatkins-dev, line 861: I feel like there needs to be some kind of check to make sure that the offsets arrays used in `CeedElemRestrictionCreateAtPoints` for the points input/output fields matches the ones provided here. Otherwise, I could imagine a scenario where we used an old version of the restriction for a field which would cause an opaque seg fault in the backend code. At the same time, that's kind of expensive, so maybe just check the first `num_elem + 1` entries (describing the number of points in each element).

Maybe we could implement this as a comparison function for `CeedElemRestriction`, then call it here?

zatkins-dev (commented):
Overall I think this looks great, it definitely clarifies the overall interface design.

backends/ref/ceed-ref-operator.c
```diff
@@ -547,6 +547,412 @@ static int CeedOperatorLinearAssembleQFunctionUpdate_Ref(CeedOperator op, CeedVe
...
+
+    switch (eval_mode) {
+      case CEED_EVAL_NONE:
+        CeedCallBackend(CeedQFunctionFieldGetSize(qf_fields[i], &size));
+        e_size = (CeedSize)max_num_points * size;
+        CeedCallBackend(CeedVectorCreate(ceed, e_size, &e_vecs[i]));
+        q_size = (CeedSize)max_num_points * size;
+        CeedCallBackend(CeedVectorCreate(ceed, q_size, &q_vecs[i]));
+        break;
+      case CEED_EVAL_INTERP:
```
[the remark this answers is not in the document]
jeremylt: Yeah

interface/ceed-operator.c
jeremylt: Yeah, I've been noodling on this and I agree. I think I have an approach (see future commit later this morning)

jeremylt: alrighty, its the future now and I have added that check

tests/t590-operator.c
```diff
@@ -0,0 +1,107 @@
...
+
+  {
+    CeedScalar        sum = 0.0;
+    const CeedScalar *v_array;
+
+    CeedVectorGetArrayRead(v, CEED_MEM_HOST, &v_array);
+    for (CeedInt i = 0; i < num_nodes; i++) sum += v_array[i];
+    CeedVectorRestoreArrayRead(v, &v_array);
+    // Summing 9 reference elements, each 2x2 => 36 sq units area
+    if (fabs(sum - 4.0 * num_elem) > CEED_EPSILON * 5e3) printf("Incorrect area computed, %f != %f\n", sum, 4.0 * num_elem);
```
jeremylt, line 94: I think two other tests would be helpful here 
1) compute element detJ on the mesh and pass it into the points operator
2) an input or output at points

tests/t591-operator.h
```diff
@@ -0,0 +1,26 @@
...
+// This file is part of CEED:  http://github.com/ceed
+
+#include <ceed.h>
+
+CEED_QFUNCTION(setup)(void *ctx, const CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out) {
+  const CeedScalar(*J)[2][2] = (const CeedScalar(*)[2][2])in[0];
+  CeedScalar *q_data         = out[0];
+
+  // Quadrature point loop
+  CeedPragmaSIMD for (CeedInt i = 0; i < Q; i++) { q_data[i] = J[i][0][0] * J[i][1][0] - J[i][0][1] * J[i][1][1]; }
```
jeremylt, line 15: This is an odd Jacobian ordering!

jeremylt: fixed

jeremylt:
CI blocked by #1391

tests/t591-operator.h
```diff
@@ -0,0 +1,26 @@
...
+}
+
+CEED_QFUNCTION(mass)(void *ctx, const CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out) {
+  const CeedScalar *u = in[0], *rho = in[1];
+  CeedScalar       *v = out[0];
+
+  // Quadrature point loop
+  CeedPragmaSIMD for (CeedInt i = 0; i < Q; i++) { v[i] = rho[i] * u[i]; }
+  return 0;
+}
```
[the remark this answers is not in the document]
jeremylt: See #1397 for 1)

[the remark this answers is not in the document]
jeremylt: See #1405 for 2)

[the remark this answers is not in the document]
jeremylt: I made the updates as separate commits in this branch, but my plan is to squash+merge because these commits aren't independent

zatkins-dev (approved):
I think that with tests passing, I'm good with merging this now to unblock parallel development. I anticipate smaller fixes will be needed along the way, but this should get us most of the way there.

PR #1392, opened 2023-11-04
[julia] Add compat bounds for stdlibs

giordano:
It's now [necessary to add compat bounds for stdlibs](https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958) too, this is currently blocking https://github.com/JuliaRegistries/General/pull/94712#issuecomment-1792859750.  After this PR is merged, you can register the same version number again, and that PR will be automatically updated.


jeremylt (approved):
Looks good to me - Thanks!

I did the same to LFAToolkit and forgot to make the change here, woops!

jedbrown:
Thanks!

PR #1391, opened 2023-11-03
Plex project coord renamed

KennethEJansen:
Copy past of @jeremylt fix for Ratel though possibly not in the best place.  It works. 

This also includes an actual code fix as a commit that I backed out in favor of Jeremy's fix that will work across PETSc versions.

PR #1396, opened 2023-11-15
Add Boolan context field register/set/get/restore

jeremylt:
This PR adds the option to register and use boolean context fields.

### DIFF
--- a/interface/ceed-qfunctioncontext.c
+++ b/interface/ceed-qfunctioncontext.c
@@ -48,7 +48,7 @@ int CeedQFunctionContextGetFieldIndex(CeedQFunctionContext ctx, const char *fiel
   @param[in]     field_name        Name of field to register
   @param[in]     field_offset      Offset of field to register
   @param[in]     field_description Description of field, or NULL for none
-  @param[in]     field_type        Field data type, such as double or int32
+  @param[in]     field_type        Field data type, such as double, int32, or bool
   @param[in]     num_values        Number of values to register, must be contiguous in memory
 
   @return An error code: 0 - success, otherwise - failure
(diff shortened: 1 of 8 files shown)


jeremylt:
Merging blocked by #1391

interface/ceed-qfunctioncontext.c
```diff
@@ -74,6 +74,19 @@ int CeedQFunctionContextRegisterGeneric(CeedQFunctionContext ctx, const char *fi
...
   CeedCall(CeedCalloc(1, &ctx->field_labels[ctx->num_fields]));
 
+  // Compute field size
+  switch (field_type) {
+    case CEED_CONTEXT_FIELD_DOUBLE:
+      field_size = sizeof(double);
+      break;
+    case CEED_CONTEXT_FIELD_INT32:
+      field_size = sizeof(int);
+      break;
```
jrwrigh, line 84: Seems odd to me to specify int32 in the enum when we're technically relying on `int` to be 32 bit. Granted, 99% of the time that is true, but maybe we could explicitly specify `int32_t` or just use `CeedInt`?

Side note, any reason why the context fields don't use `CeedScalar`/`CeedInt` instead of compiler-generic values? I think most of the time we assume that `double` == `CeedScalar` and `int` == `CeedInt` in the QF context, but that might not always be true I think.

jrwrigh (approved):

interface/ceed-qfunctioncontext.c
jeremylt: I use `int` and `double` because I want to be explicit about how many bytes are being copied. A `PetscScalar` and `CeedScalar` or a `PetscInt` and a `CeedInt` can be completely different types.

PR #1397, opened 2023-11-16
Enable CEED_EVAL_WEIGHT for BasisApplyAtPoints

jeremylt:
This enables `CEED_EVAL_WEIGHT` for `CeedBasisApplyAtPoints` for better compatibility between FEM and PIC `CeedQFunction`s.

### DIFF
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -1472,17 +1472,22 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
   CeedCall(CeedBasisGetNumComponents(basis, &num_comp));
   CeedCall(CeedBasisGetNumQuadratureComponents(basis, eval_mode, &num_q_comp));
   CeedCall(CeedBasisGetNumNodes(basis, &num_nodes));
-  CeedCall(CeedVectorGetLength(x_ref, &x_length));
   CeedCall(CeedVectorGetLength(v, &v_length));
-  CeedCall(CeedVectorGetLength(u, &u_length));
+  if (x_ref != CEED_VECTOR_NONE) CeedCall(CeedVectorGetLength(x_ref, &x_length));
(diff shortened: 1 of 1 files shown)


jeremylt:
CI blocked by #1391

interface/ceed-basis.c
```diff
@@ -1495,8 +1500,10 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
       good_dims = ((t_mode == CEED_TRANSPOSE && (u_length >= num_points * num_q_comp * dim || v_length >= num_nodes * num_comp)) ||
                    (t_mode == CEED_NOTRANSPOSE && (v_length >= num_points * num_q_comp * dim || u_length >= num_nodes * num_comp)));
       break;
-    case CEED_EVAL_NONE:
     case CEED_EVAL_WEIGHT:
+      good_dims = t_mode == CEED_NOTRANSPOSE && (v_length >= num_points);
```
zatkins-dev, line 1504: Strictly speaking, I don't think we need the `t_mode == CEED_NOTRANSPOSE` here since we check that above. Not a big deal either way though.

zatkins-dev (approved):
LGTM!

zatkins-dev:
I think that it's reasonable to ignore the code coverage issues here, but if others disagree we should add a tiny test that uses `CEED_EVAL_WEIGHT` with `CeedBasisApplyAtPoints`.

interface/ceed-basis.c
jeremylt: Yeah, I mostly included it so we don't have to read and think about why it looks different than the other cases.

jeremylt:
> I think that it's reasonable to ignore the code coverage issues here, but if others disagree we should add a tiny test that uses `CEED_EVAL_WEIGHT` with `CeedBasisApplyAtPoints`.

Agreed - we're immediately using it in the follow-up PR

PR #1398, opened 2023-11-16
Remove legacy `CeedChk` and `CeedChkBackend`

sebastiangrimberg:
Address this comment in `backend.h`:

```
/* Note - these are legacy macros that should be removed eventually */
#define CeedChk(...) CeedCall(__VA_ARGS__)
#define CeedChkBackend(...) CeedCallBackend(__VA_ARGS__)
```

This isn't super important but I figured was worth addressing after the previous release.


jeremylt (approved):
LGTM, thanks!

PR #1399, opened 2023-11-17
fluids: Cleanup advection problems

jrwrigh:
- Remove misc struct members that are no longer used
- Use enums rather than integers in contexts


jeremylt (approved):
All seems reasonable to me - as long as Noether's happy, I'm happy

jeremylt:
Probably needs a rebase on main?

PR #1405, opened 2023-11-27
Change order of ElemRestriction/Basis AtPoints

jeremylt:
This PR swaps to "contiguous by qpt" ordering for the output of AtPoints operations so we can reuse our QFunctions.

~~Something is wrong with my TensorContract in 2d/3d~~

### DIFF
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -1609,7 +1609,7 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
 
             for (CeedInt d = 0; d < dim; d++) {
               // ------ Tensor contract with current Chebyshev polynomial values
-              CeedCall(CeedChebyshevPolynomialsAtPoint(x_array_read[p * dim + d], Q_1d, chebyshev_x));
+              CeedCall(CeedChebyshevPolynomialsAtPoint(x_array_read[d * num_points + p], Q_1d, chebyshev_x));
               CeedCall(CeedTensorContractApply(basis->contract, pre, Q_1d, post, 1, chebyshev_x, t_mode, false,
                                                d == 0 ? chebyshev_coeffs : tmp[d % 2], tmp[(d + 1) % 2]));
               pre /= Q_1d;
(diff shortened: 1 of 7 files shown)


interface/ceed-basis.c
```diff
@@ -1599,22 +1599,25 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
...
+
+            // ------ Build interp matrix
+            for (CeedInt p = 0; p < num_points; p++) {
+              CeedCall(CeedChebyshevPolynomialsAtPoint(x_array_read[p * dim + d], Q_1d, &interp_1d[p * Q_1d]));
             }
+            // ------ Tensor contract with current Chebyshev polynomial values
+            CeedCall(CeedTensorContractApply(basis->contract, pre, Q_1d, post, num_points, interp_1d, t_mode, false,
+                                             d == 0 ? chebyshev_coeffs : tmp[d % 2], d == (dim - 1) ? v_array : tmp[(d + 1) % 2]));
+            pre /= Q_1d;
+            post *= 1;
```
jeremylt, line 1620: I think something has to be screwy here.

The 1d contraction works. 2D is wrong.

 I'm trying to contract from Q_1d * Q_1d (for example) down to 1 * num_points

jeremylt: `CeedTensorContractStridedApply` might be what I need here?

jeremylt: No that won't work either.

I need to go Q_1d * Q_1d -> Q_1d * num_points -> 1 * num_points

or in 3D
Q_1d * Q_1d * Q_1d-> Q_1d * Q_1d * num_points -> Q_1d * num_points -> 1 * num_points

Our current tensor contractions don't do that.

jeremylt: Maybe setting post = Q_1d^dim-1 and decreasing it?

interface/ceed-basis.c
```diff
@@ -1609,12 +1609,13 @@ int CeedBasisApplyAtPoints(CeedBasis basis, CeedInt num_points, CeedTransposeMod
...
               // ------ Tensor contract with current Chebyshev polynomial values
-              CeedCall(CeedChebyshevPolynomialsAtPoint(x_array_read[p * dim + d], Q_1d, chebyshev_x));
+              CeedCall(CeedChebyshevPolynomialsAtPoint(x_array_read[d * num_points + p], Q_1d, chebyshev_x));
               CeedCall(CeedTensorContractApply(basis->contract, pre, Q_1d, post, 1, chebyshev_x, t_mode, false,
-                                               d == 0 ? chebyshev_coeffs : tmp[d % 2], d == (dim - 1) ? &v_array[p * num_comp] : tmp[(d + 1) % 2]));
+                                               d == 0 ? chebyshev_coeffs : tmp[d % 2], tmp[(d + 1) % 2]));
               pre /= Q_1d;
               post *= 1;
             }
+            for (CeedInt c = 0; c < num_comp; c++) v_array[c * num_points + p] = tmp[dim % 2][c];
```
jeremylt, line 1618: This meets the target ordering. I think we should chase performance in a separate PR.

zatkins-dev (approved):
Looks good to me. I don't think the extra copy is particularly impactful performance-wise (and more importantly, it seems unavoidable at the moment).

PR #1409, opened 2023-11-30
Docstring consistency

### LINKED ISSUE #1393: Docstring Inconsistency
Currently, our use of `\ref CeedOperatorSetField` vs `CeedOperatorSetField()` is inconsistent in the docs. (and our use of `@ref` vs `\ref` confuses me).

We should fix this.

Also, we have CeedOperator and similar as plain text rather than wrapping them in code blocks like Ratel and PETSc do.

jeremylt:
Fixes #1393 -

- [x] consistently use `@ref` vs `\ref`
- [x] swap away from `@a` to ref args since Sphinx doesn't use it
- [x] minor wording fixes
- [x] consistently use code formatting for arguments referenced in sentences
- [x] Consistent capitalization and spacing in arg lists

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -8,6 +8,7 @@ AlignEscapedNewlines:         true
 AlignOperands:                Align
 AllowShortIfStatementsOnASingleLine: AllIfsAndElse
 ColumnLimit:                  150
+ReflowComments:               false
 CommentPragmas:               'TESTARGS'
 DerivePointerAlignment:       false
 IncludeBlocks:                Preserve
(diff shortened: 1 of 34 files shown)


jrwrigh (approved):
Looks good. Only major question is the logic behind putting some "named" Ceed objects inside ticks, while others get the full `@ref` treatment? Is `@ref` only valid for a small subset (ie; not for `CeedBasis` and `CeedElemRestriction`)?

jeremylt:
Yeah, @ref doesn't work right for the `Ceed*` objects, for some reason.

jrwrigh:
Except for `CeedQFunction`?

jeremylt:
Nope, that one doesn't link right either. Did I miss it somewhere?

jrwrigh:
You added them in for the gallery functions

jeremylt:
Ah, fixed. Thanks

PR #1411, opened 2023-12-01
Update petsc-dmswarm example to use CeedOperatorAtPoints

zatkins-dev:
Still WIP, but tests are passing. A few design questions: 
1. We mentioned wanting to get rid of/reduce the `DMSwarmCeedContext` struct, which I cleaned up a little so far. Are we happy with its current state? Or should we try to refactor more of it? My main thought is that it saves us having to build the bases, restrictions, and points every time we need them, which definitely makes the code cleaner.
2. It turns out that we can do everything we want for this example using only identity QFunctions. While this is nice, it feels...odd. Am I missing something here that should require a non-trivial QFunction?


jeremylt:
I was thinking that we could swap over to a real mass matrix with qdata, like in #1390

zatkins-dev:
> I was thinking that we could swap over to a real mass matrix with qdata, like in #1390 

I thought that might be the case. I tried that first, but ran into some issues where the solution was off by a scalar multiple. I can try to diagnose that more tomorrow. 

Would the RHS need to change as well? That could explain the issue.

zatkins-dev:
> Yeah, the RHS would also need to be scaled by the geometric factors. Off by what, a factor of 8?

It was a factor of 26 (exactly) which seems very random.

jeremylt:
Hmm, how many elements are there?

zatkins-dev:
I think 27 (3x3x3)

zatkins-dev:
The issue was just scaling the RHS!

jeremylt:
This is looking great. Two thoughts with the goal of adding Poisson in view

1) Can we totally drop most of the struct of Ceed objects in favor of a CeedOperator if we swap to a single setup function for the CeedOperator and RHS?

2) Can we generalize those functions that call the CeedOperator (probably will be best in a future PR now that I think of  it)

zatkins-dev:
>     1. Can we totally drop most of the struct of Ceed objects in favor of a CeedOperator if we swap to a single setup function for the CeedOperator and RHS?

Done, good call. We're down to basically a operator apply context, just with 3 operators.

zatkins-dev:
> 2\. Can we generalize those functions that call the CeedOperator (probably will be best in a future PR now that I think of  it)

I'm thinking that we can refactor out a lot of the functions in `dmswarm.c` into a `swarmutils.c`/`swarmutils.h` to get ready for additional swarm problems. If I understand correctly, we're always going to have to use the mass operator to go from points to the mesh, so maybe we add an additional `op_apply` in the `DMSwarmCeedContext` struct to actually do the problem (e.g. the laplacian for BP3)?

examples/petsc/dmswarm.h
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and other CEED contributors.
```
jeremylt, line 1: Note - probably should be in the `qfunctions` folder.

zatkins-dev: Moved in refactor

examples/petsc/dmswarm.c
```diff
@@ -65,30 +54,11 @@ PetscErrorCode EvalU_Tanh_proj(PetscInt dim, PetscReal t, const PetscReal x[], P
...
-PetscErrorCode VecC2P(CeedVector x_ceed, PetscMemType mem_type, Vec X_petsc);
-PetscErrorCode VecReadP2C(Vec X_petsc, PetscMemType *mem_type, CeedVector x_ceed);
-PetscErrorCode VecReadC2P(CeedVector x_ceed, PetscMemType mem_type, Vec X_petsc);
-PetscErrorCode DMSwarmPICFieldP2C(DM dm_swarm, const char *field, CeedVector x_ceed);
-PetscErrorCode DMSwarmPICFieldC2P(DM dm_swarm, const char *field, CeedVector x_ceed);
-
 // Swarm to mesh and mesh to swarm
-PetscErrorCode DMSwarmCreateReferenceCoordinates(DM dm_swarm, IS *is_points, Vec *ref_coords);
 PetscErrorCode DMSwarmInterpolateFromCellToSwarm_Petsc(DM dm_swarm, const char *field, Vec U_mesh);
 PetscErrorCode DMSwarmInterpolateFromCellToSwarm_Ceed(DM dm_swarm, const char *field, Vec U_mesh);
```
zatkins-dev, line 59: Unsure on if this function should be moved to `swarmutils` or not. Ultimately, the action of this function is done automatically when a more complex problem is involved, so I think that it probably can stay where it is.

jeremylt: I wouldn't overthink it too much since this is a testbed. You'll see more of what should be moved when adding BP3

zatkins-dev: At the same time, this is the only place where the `op_mesh_to_points` in the `DMSwarmCeedContext` is used, so maybe just include it for completeness? Or reconsider whether `op_mesh_to_points` needs to live in that struct? Not sure

examples/petsc/include/swarmutils.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
```
jeremylt, line 1: All files in the libCEED repo have this boilerplate at the top

```
// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and other CEED contributors.
// All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
//
// SPDX-License-Identifier: BSD-2-Clause
//

/// @file
/// libCEED QFunctions for mass operator example using PETSc
```

examples/petsc/src/swarmutils.c
```diff
@@ -0,0 +1,589 @@
...
+  PetscOptionsBegin(PETSC_COMM_WORLD, NULL, "Swarm-to-Mesh Projection Options", NULL);
+  PetscCall(PetscOptionsBool("-test", "Testing mode (do not print unless error is large)", NULL, test_mode, &test_mode, NULL));
+  PetscOptionsEnd();
+
+  comm = PetscObjectComm((PetscObject)dm_swarm);
+  PetscCall(DMSwarmGetCellDM(dm_swarm, &dm_mesh));
+  PetscCall(DMGetApplicationContext(dm_mesh, (void *)&swarm_ceed_context));
+  PetscCall(VecDuplicate(U_mesh, &B_mesh));
+
+  // -- Setup "mass matrix"
```
jeremylt, line 523: Can probably ditch the `--` in the comments here

examples/petsc/include/swarmutils.h
zatkins-dev: Gotcha -- it seemed a little inconsistent in the other source/headers in the petsc examples

examples/petsc/dmswarm.c
jeremylt: Probably worth keeping around in there for now for symmetry, I'd say

examples/petsc/include/petscutils.h
```diff
@@ -10,6 +10,10 @@
 #endif
 
 CeedMemType      MemTypeP2C(PetscMemType mtype);
+PetscErrorCode   VecP2C(Vec X_petsc, PetscMemType *mem_type, CeedVector x_ceed);
+PetscErrorCode   VecC2P(CeedVector x_ceed, PetscMemType mem_type, Vec X_petsc);
+PetscErrorCode   VecReadP2C(Vec X_petsc, PetscMemType *mem_type, CeedVector x_ceed);
+PetscErrorCode   VecReadC2P(CeedVector x_ceed, PetscMemType mem_type, Vec X_petsc);
```
jeremylt, line 16: if you want a quick task at some point we should swap over to using these utils in the other PETSc example code (`matops.c` is probably the one?).

jeremylt (approved):
I think this leaves the code well set-up for the BP3

examples/petsc/include/petscutils.h
zatkins-dev: Can do, I'll make a new branch

PR #1414, opened 2023-12-03
Add support for running BPs with integration at particle points

zatkins-dev:
~~Extremely WIP, tons of bad, unused, and messy code. But~~ the BPs work and converge under grid refinement using the uniform, Gauss, and cellwise-random integration points.

Sample run:
```
./build/petsc-bpsswarm -problem bp3 -dm_plex_dim 3 -dm_plex_box_faces 3,3,3 -dm_plex_simplex 0 -swarm uniform -points_per_cell 64 -solution_view vtk:solution_swarm.vtu -degree 2 -rhs_view vtk:rhs.vtu
```

To do:
 - [x] Change error operator to be computed at mesh quadrature (using points for that is cheating in my opinion)
 - [x] Remove all unused functions and commented code
 - [x] Re-evaluate whether `DMSwarmCeedContext` is needed for BPs (leaning toward no)
 - [x] Add tests

### DIFF
--- a/examples/petsc/bpsswarm.c
+++ b/examples/petsc/bpsswarm.c
@@ -22,7 +22,9 @@
 //     bpssphere -problem bp2 -degree 3
 //     bpssphere -problem bp3 -degree 3
 //
-//TESTARGS -ceed {ceed_resource} -test -problem bp3 -dm_plex_dim 3 -dm_plex_box_faces 10,10,10 -dm_plex_simplex 0 -swarm uniform -points_per_cell 500
+//TESTARGS(name="BP2") -ceed {ceed_resource} -test -problem bp2 -dm_plex_dim 3 -dm_plex_box_faces 3,3,3 -dm_plex_simplex 0 -swarm uniform -points_per_cell 64
+//TESTARGS(name="BP3") -ceed {ceed_resource} -test -problem bp3 -dm_plex_dim 3 -dm_plex_box_faces 4,4,4 -dm_plex_simplex 0 -swarm uniform -points_per_cell 64 -tolerance 3e-2
+//TESTARGS(name="BP5") -ceed {ceed_resource} -test -problem bp5 -dm_plex_dim 3 -dm_plex_box_faces 3,3,3 -dm_plex_simplex 0 -swarm uniform -points_per_cell 64
 
 /// @file
 /// CEED BPs example using PETSc with DMPlex
@@ -49,7 +51,8 @@ int main(int argc, char **argv) {
   MPI_Comm             comm;
   char                 ceed_resource[PETSC_MAX_PATH_LEN] = "/cpu/self", filename[PETSC_MAX_PATH_LEN];
   double               my_rt_start, my_rt, rt_min, rt_max;
-  PetscInt             comm_size, degree = 3, q_extra, l_size, g_size, dim = 3, num_comp_u = 1, xl_size, num_points = 1728, num_points_per_cell = 64;
+  PetscScalar          tolerance;
+  PetscInt             comm_size, degree, q_extra, l_size, g_size, dim = 3, num_comp_u = 1, xl_size, num_points = 1728, num_points_per_cell = 64;
   PetscBool            test_mode, benchmark_mode, read_mesh, write_solution, write_true_solution_swarm;
   PetscLogStage        solve_stage;
   Vec                  X, X_loc, rhs;
@@ -94,7 +97,7 @@ int main(int argc, char **argv) {
   write_true_solution_swarm = PETSC_FALSE;
   PetscCall(PetscOptionsBool("-write_true_solution_swarm", "Write true solution at swarm points for visualization", NULL, write_true_solution_swarm,
                              &write_true_solution_swarm, NULL));
-  degree = test_mode ? 3 : 2;
+  degree = 2;
   PetscCall(PetscOptionsInt("-degree", "Polynomial degree of tensor product basis", NULL, degree, &degree, NULL));
   q_extra = bp_options[bp_choice].q_extra;
   PetscCall(PetscOptionsInt("-q_extra", "Number of extra quadrature points", NULL, q_extra, &q_extra, NULL));
@@ -103,6 +106,8 @@ int main(int argc, char **argv) {
   PetscCall(PetscOptionsString("-hostname", "Hostname for output", NULL, hostname, hostname, sizeof(hostname), NULL));
   read_mesh = PETSC_FALSE;
   PetscCall(PetscOptionsString("-mesh", "Read mesh from file", NULL, filename, filename, sizeof(filename), &read_mesh));
+  tolerance = 1e-2;
+  PetscCall(PetscOptionsScalar("-tolerance", "Tolerance for L2 error", NULL, tolerance, &tolerance, NULL));
   PetscCall(PetscOptionsEnum("-swarm", "Swarm points distribution", NULL, point_swarm_types, (PetscEnum)point_swarm_type,
                              (PetscEnum *)&point_swarm_type, NULL));
   {
@@ -173,7 +178,12 @@ int main(int argc, char **argv) {
   // -- Swarm field
   PetscCall(DMSwarmRegisterPetscDatatypeField(dm_swarm, DMSwarmPICField_u, num_comp_u, PETSC_SCALAR));
   PetscCall(DMSwarmFinalizeFieldRegister(dm_swarm));
-  PetscCall(DMSwarmSetLocalSizes(dm_swarm, num_points, 0));
+  {
+    PetscInt c_start, c_end, num_cells_local;
+    PetscCall(DMPlexGetHeightStratum(dm_mesh, 0, &c_start, &c_end));
+    num_cells_local = c_end - c_start;
+    PetscCall(DMSwarmSetLocalSizes(dm_swarm, num_cells_local * num_points_per_cell, 0));
+  }
   PetscCall(DMSetFromOptions(dm_swarm));
 
   // -- Set swarm point locations
@@ -240,8 +250,9 @@ int main(int argc, char **argv) {
     PetscMPIInt comm_size;
     PetscCall(MPI_Comm_size(comm, &comm_size));
 
-    PetscInt num_points_local;
+    PetscInt num_points_local, num_points_global;
     PetscCall(DMSwarmGetLocalSize(dm_swarm, &num_points_local));
+    PetscCall(DMSwarmGetSize(dm_swarm, &num_points_global));
 
     PetscCall(PetscPrintf(comm,
                           "\n-- CEED Benchmark Problem %" CeedInt_FMT " -- libCEED + PETSc --\n"
@@ -268,7 +279,7 @@ int main(int argc, char **argv) {
                           "    Avg points per cell                     : %" PetscInt_FMT "\n"
                           "    Point distribution                      : %s\n",
                           bp_choice + 1, hostname, comm_size, ranks_per_node, vec_type, used_resource, CeedMemTypes[mem_type_backend], P, Q, q_extra,
-                          g_size / num_comp_u, num_cells_local, l_size / num_comp_u, num_comp_u, num_points, num_points_local,
+                          g_size / num_comp_u, num_cells_local, l_size / num_comp_u, num_comp_u, num_points_global, num_points_local,
                           num_cells_local > 0 ? num_points_local / num_cells_local : 0, point_swarm_types[point_swarm_type]));
   }
 
@@ -391,8 +402,7 @@ int main(int argc, char **argv) {
       PetscScalar l2_error;
       PetscCall(ComputeL2Error(X, &l2_error, op_error_ctx));
 
-      PetscReal tol = 5e-4;
-      if (!test_mode || l2_error > tol) {
+      if (!test_mode || l2_error > tolerance) {
         PetscCall(MPI_Allreduce(&my_rt, &rt_min, 1, MPI_DOUBLE, MPI_MIN, comm));
         PetscCall(MPI_Allreduce(&my_rt, &rt_max, 1, MPI_DOUBLE, MPI_MAX, comm));
         PetscCall(PetscPrintf(comm,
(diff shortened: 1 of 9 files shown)


zatkins-dev:
Note to self: Something is wrong with the error calculations... running with `-error_view` shows much lower error than the printed error.

examples/petsc/bpsswarm.c
```diff
@@ -0,0 +1,439 @@
...
+//
+//     make bpsswarm [PETSC_DIR=</path/to/petsc>] [CEED_DIR=</path/to/libceed>]
+//
+// Sample runs:
+//
+//     bpssphere -problem bp1 -degree 3
+//     bpssphere -problem bp2 -degree 3
+//     bpssphere -problem bp3 -degree 3
+//
+//TESTARGS -ceed {ceed_resource} -test -problem bp3 -dm_plex_dim 3 -dm_plex_box_faces 10,10,10 -dm_plex_simplex 0 -swarm uniform -points_per_cell 500
```
jedbrown, line 25: Can maybe decrease the problem size a bit so it's not expensive. (Coarser mesh and 64 points per element seems like enough.)

jedbrown (commented):
This is very cool. I think `-swarm cell_random` still has a bug since `Local points` is way smaller than it should be. 

```console
$ build/petsc-bpsswarm -problem bp3 -dm_plex_dim 3 -dm_plex_box_faces 10,10,10 -dm_plex_simplex 0 -swarm cell_random -points_per_cell 100
[...]
  Swarm:
    Global points                           : 125000
    Local points                            : 18347
    Avg points per cell                     : 18
    Point distribution                      : cell_random
  KSP:
    KSP Type                                : cg
    KSP Convergence                         : DIVERGED_INDEFINITE_PC
    Total KSP Iterations                    : 142
    Final rnorm                             : 4.202751e-10
  Performance:
    L2 Error                                : 8.765200e-02
    CG Solve Time                           : 3.18343 (3.18343) sec
```

zatkins-dev:
> Note to self: Something is wrong with the error calculations... running with `-error_view` shows much lower error than the printed error.

It seems like all of the error is occurring at higher-order nodes. I'm not entirely sure why that would be the case.
image

zatkins-dev:
> This is very cool. I think `-swarm cell_random` still has a bug since `Local points` is way smaller than it should be.

You're right -- there were some issues with the upper/lower bounds and the scaling of the random number

examples/petsc/src/swarmutils.c
```diff
@@ -593,3 +608,213 @@ PetscErrorCode DMSwarmProjectFromSwarmToCells(DM dm_swarm, const char *field, Ve
...
+        const PetscScalar *x;
+
+        PetscCall(DMGetCoordinatesLocal(dm_mesh, &X_loc));
+        PetscCall(VecGetLocalSize(X_loc, &len));
+        CeedVectorCreate(ceed, len, &x_coord);
+
+        PetscCall(VecGetArrayRead(X_loc, &x));
+        CeedVectorSetArray(x_coord, CEED_MEM_HOST, CEED_COPY_VALUES, (CeedScalar *)x);
+        PetscCall(VecRestoreArrayRead(X_loc, &x));
+      }
```
jeremylt, line 697: Note to investigate for here/Ratel - really we shouldn't need to copy since this CeedVector is destroyed at the end of this function. Could P2C and C2P it back. Marginal memory savings reusing the same array instead of copying

examples/petsc/src/swarmutils.c
```diff
@@ -282,14 +283,16 @@ PetscErrorCode DMSwarmInitalizePointLocations(DM dm_swarm, PointSwarmType point_
     } break;
     case SWARM_CELL_RANDOM: {
       // -- Set points randomly in each cell
-      PetscInt     dim = 3, num_cells_total = 1, num_cells[] = {1, 1, 1};
-      PetscScalar *point_coords;
+      PetscInt     dim = 3, num_cells_total = 1, num_cells[] = {1, 1, 1}, num_lower, num_upper;
```
jedbrown:
```suggestion
      PetscInt     dim = 3, num_cells_total = 1, num_cells[] = {1, 1, 1}, num_lower = 3, num_upper = 3;
```
to prevent non-deterministic
```
[0]PETSC ERROR: Argument out of range
[0]PETSC ERROR: n (-1439124880) cannot be negative
[0]PETSC ERROR: WARNING! There are unused option(s) set! Could be the program crashed before usage or a spelling mistake, etc!
[0]PETSC ERROR:   Option left: name:-ksp_monitor (no value) source: command line
[0]PETSC ERROR:   Option left: name:-ksp_view_solution value: cgns:s.cgns source: command line
[0]PETSC ERROR: See https://petsc.org/release/faq/ for trouble shooting.
[0]PETSC ERROR: Petsc Development GIT revision: v3.20.1-154-g1024d75c3cc  GIT Date: 2023-11-06 20:45:07 +0000
[0]PETSC ERROR: build/petsc-bpsswarm on a ompi named kichatna by jed Mon Dec  4 11:21:14 2023
[0]PETSC ERROR: Configure options --download-cgns --download-chaco --download-ctetgen --download-hpddm --download-ml --download-mumps --download-ptscotch=/home/jed/dl/scotch_7.0.4beta3.tar.gz --download-scalapack --download-superlu --download-triangle --with-cuda --with-debugging=0 --with-fortran-bindings=0 --with-hdf5 --with-hip=0 --with-metis --with-mpi-dir=/home/jed/usr/ccache/ompi --with-opengl --with-openmp --with-parmetis --with-strict-petscerrorcode --with-suitesparse --with-zlib COPTFLAGS="-O2 -march=native -ffp-contract=fast -g" PETSC_ARCH=ompi
[0]PETSC ERROR: #1 PetscOptionsScalarArray_Private() at /home/jed/petsc/src/sys/objects/aoptions.c:1669
[0]PETSC ERROR: #2 DMSwarmInitalizePointLocations() at /home/jed/src/libCEED/examples/petsc/src/swarmutils.c:294
[0]PETSC ERROR: #3 main() at /home/jed/src/libCEED/examples/petsc/bpsswarm.c:180
```

zatkins-dev: Apparently there's a built-in cell-wise random coordinate function, which avoids some nasty issues with writing to the field in parallel.

examples/petsc/src/swarmutils.c
```diff
@@ -306,8 +309,10 @@ PetscErrorCode DMSwarmInitalizePointLocations(DM dm_swarm, PointSwarmType point_
 
           for (PetscInt i = 0; i < dim; i++) {
             PetscCall(PetscRandomGetValue(rng, &random_value));
-            point_coords[point_index * dim + i] = -1.0 + cell_index[i] * 2.0 / (num_cells[i] + 1.0) + random_value;
+            point_coords[point_index * dim + i] = lower[i] + (cell_index[i] * (upper[i] - lower[i]) + random_value) / num_cells[i];
           }
+          PetscCall(PetscPrintf(PETSC_COMM_WORLD, "Point %d: (%g, %g, %g)\n", point_index, point_coords[point_index * dim + 0],
```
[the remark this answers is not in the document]
zatkins-dev: Yes, left in by mistake

examples/petsc/bpsswarm.c
zatkins-dev: Okay, I had to tweak error tolerances, but I got the problem sizes much smaller. I also added other BPs to make sure they also work fine.

examples/petsc/src/swarmutils.c
zatkins-dev: That's certainly true here, though there are some places in Ratel which hold onto the `CeedVector` as a passive input (e.g. platens). Although we have discussed removing that in favor of a more precise subset of that vector in the past.

jeremylt: I think its worth not copying here and anywhere else in the PETSc examples and Ratel that we have this pattern inside the same function call of
1) create vector
2) copy coords over
3) use Operator
4) destroy Operator + vector

jeremylt: Oh, I already changed this in Ratel

examples/petsc/src/petscutils.c
```diff
@@ -154,6 +154,7 @@ PetscErrorCode SetupDMByDegree(DM dm, PetscInt p_degree, PetscInt q_extra, Petsc
     PetscCall(DMGetCoordinateDim(dm, &num_comp_coord));
     PetscCall(PetscFECreateLagrange(comm, dim, num_comp_coord, is_simplex, 1, q_degree, &fe_coords));
     PetscCall(DMSetCoordinateDisc(dm, fe_coords, PETSC_TRUE));
+    PetscCall(DMGetCoordinatesLocalSetUp(dm));
```
jeremylt, line 157: Hmm, why was this needed? I used this for a bit in Ratel to support different logic on different processes previously, but it lead to a massive amount of headache.

examples/petsc/src/libceedsetup.c
```diff
@@ -211,4 +212,125 @@ PetscErrorCode CeedLevelTransferSetup(DM dm, Ceed ceed, CeedInt level, CeedInt n
...
+  CeedOperatorSetField(op_setup_rhs, "x", elem_restr_x, basis_x, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(op_setup_rhs, "qdata", elem_restr_qd_i, CEED_BASIS_NONE, q_data);
+  CeedOperatorSetField(op_setup_rhs, "true solution", elem_restr_u_i, CEED_BASIS_NONE, target);
+  CeedOperatorSetField(op_setup_rhs, "rhs", elem_restr_u, basis_u, CEED_VECTOR_ACTIVE);
+
+  // Set up the libCEED context
+  CeedQFunctionContext ctx_rhs_setup;
+  CeedQFunctionContextCreate(ceed, &ctx_rhs_setup);
+  CeedScalar rhs_setup_data[2] = {R, l};
+  CeedQFunctionContextSetData(ctx_rhs_setup, CEED_MEM_HOST, CEED_COPY_VALUES, sizeof rhs_setup_data, &rhs_setup_data);
```
jeremylt, line 296: What's the convention we've been using in this repo? I standardized Ratel to `sizeof(thing)` over `sizeof thing` a while ago since we had a mix.

examples/petsc/src/petscutils.c
zatkins-dev: It's required for the call to `DMSwarmSetPointCoordinatesRandom` for some reason.

jeremylt: Ah, eww.

knepley: Here is what `DMGetCoordinatesLocalSetUp()` does. We would like to call `DMGetCoordinatesLocal()` to extract the local coordinates without being collective. However, if only the global coordinates exist, we try to make them using GlobalToLocal, which is collective. Thus, to assure non-collective access, we need to first call `DMGetCoordinatesLocalSetUp()`.

examples/petsc/src/libceedsetup.c
zatkins-dev: I think I copied directly from another function in this file, but I'm not sure on the general convention.

examples/petsc/src/petscutils.c
jeremylt: Right, I've used it before. We've had problems with mismatched non-collective and collective logic before since its easy to screw up.

jeremylt: > It's required for the call to `DMSwarmSetPointCoordinatesRandom` for some reason.

Can we put it only in the two swarm driver functions right before we set the mesh DMPlex for  the DMSwarm? That way its not changing the behavior of the other BP examples?

examples/petsc/src/libceedsetup.c
jeremylt: No worries if it isn't standardized here. That can be on the list of "nice to tidy some day".

examples/petsc/bpsswarm.h
```diff
@@ -0,0 +1,17 @@
...
+// This file is part of CEED:  http://github.com/ceed
+
+#ifndef LIBCEED_PETSC_EXAMPLES_BPSSWARM_H
+#define LIBCEED_PETSC_EXAMPLES_BPSSWARM_H
+
+// -----------------------------------------------------------------------------
+// Command Line Options
+// -----------------------------------------------------------------------------
+
+static const char *const bp_types[] = {"bp1", "bp2", "bp3", "bp4", "bp5", "bp6", "BPType", "CEED_BP", 0};
```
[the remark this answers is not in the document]
zatkins-dev: yeah, honestly this header can probably be removed and we can just use `bps.h`. I originally added it to limit which BPs were allowed.

examples/petsc/src/petscutils.c
zatkins-dev: Sure, though it shouldn't change any behavior -- if I understand correctly, it just pre-caches something that will be created either way. Maybe I'm wrong there.

jeremylt: Yeah, it will behave the same. I'm more thinking on the development side. Since we don't really set up any of our function to be non-collective, I'd rather keep getting an error early if we violate that assumption that all our functions are collective.

zatkins-dev: We could also put it right before the call to the `DMSwarmSetPointCoordinatesRandom`

jeremylt: Sure, that works for me since that's a collective operation.

zatkins-dev:
I just went through and cleaned up the commit history so we can avoid a squash-merge.

examples/petsc/src/swarmutils.c
```diff
@@ -281,37 +282,11 @@ PetscErrorCode DMSwarmInitalizePointLocations(DM dm_swarm, PointSwarmType point_
...
-          for (PetscInt i = 0; i < dim; i++) {
-            PetscCall(PetscRandomGetValue(rng, &random_value));
-            point_coords[point_index * dim + i] = -1.0 + cell_index[i] * 2.0 / (num_cells[i] + 1.0) + random_value;
-          }
-        }
-      }
-      PetscCall(DMSwarmRestoreField(dm_swarm, DMSwarmPICField_coor, NULL, NULL, (void **)&point_coords));
-      PetscCall(PetscRandomDestroy(&rng));
+      PetscCall(DMSwarmGetCellDM(dm_swarm, &dm_mesh));
+      PetscCall(DMGetCoordinatesLocalSetUp(dm_mesh));
```
jeremylt, line 288: Can we add a note to our future selves why this call is needed?

jeremylt (approved):
looks awesome, thanks!

jedbrown (approved):

PR #1415, opened 2023-12-04
fluids: Add ratio dx for stg

jrwrigh:
Define the streamwise length based on ratios of the spanwise via `-stg_dx_ratio`


jrwrigh:
Closed in favor of #1606

PR #1419, opened 2023-12-07
junit - handle tests without names

jeremylt:
The extra , in TAP mode when the test doesn't have a name was driving me a little batty

### DIFF
--- a/tests/junit_common.py
+++ b/tests/junit_common.py
@@ -314,15 +314,16 @@ def diff_cgns(test_cgns: Path, true_cgns: Path, tolerance: float = 1e-12) -> str
 
 def test_case_output_string(test_case: TestCase, spec: TestSpec, mode: RunMode,
                             backend: str, test: str, index: int) -> str:
-    output_str = ''
+    output_str: str = ''
     if mode is RunMode.TAP:
         # print incremental output if TAP mode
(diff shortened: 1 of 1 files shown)


tests/junit_common.py
```diff
@@ -384,8 +385,9 @@ def run_test(index: int, test: str, spec: TestSpec, backend: str,
 
     # run test
     skip_reason: str = suite_spec.check_pre_skip(test, spec, backend, nproc)
+    test_name_str: str = f'"{spec.name}", ' if spec.name else ''
```
jedbrown, line 388: Should we require that all tests be named (many testing frameworks do), or perhaps automatically assign them names?

jeremylt: Adding names would be a lot of changes to almost every test file. I'm not opposed, but its also another thing to manually change every time we make a new test. I could go either way

jedbrown: It looks like every file with more than one test names them. We could assert that (if you have multiple tests they all need names), but I feel like the sed operation to add `name="default",` to all the unnamed ones in `t*.c` files is also fine. One thing I've considered is using libtest-mimic to support using `cargo test` to run these tests, and if I do that, I'll need to name them all (either explicitly or programmatically).

zatkins-dev: I'm good with either approach

PR #1421, opened 2023-12-08
Fix op swarm input access

jeremylt:
@zatkins-dev this fixes the error registered with `petsc-dmswarm`, though I don't see why it this is needed given line `728-730` :thinking: 

Ohh, strike that, I do know why. Once we swap to the memcheck Vector, it exposes that those lines violate the memory access assumptions. I'm still not sure why this isn't an issue with regular operators, but for regular operators we actually call a basis operation, so this isn't an issue there.

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -598,13 +598,21 @@ static int CeedOperatorSetupFieldsAtPoints_Ref(CeedQFunction qf, CeedOperator op
     }
 
     switch (eval_mode) {
-      case CEED_EVAL_NONE:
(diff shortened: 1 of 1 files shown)


zatkins-dev (approved):
Looks good to me, with the caveat that we revisit the memory access assumptions later

jeremylt:
I think I want to rework this to reference copy the evector to work as the qvector

jeremylt:
@zatkins-dev I think this is a cleaner fix that avoids the dirty memory access hack

zatkins-dev (approved):
This is a much cleaner solution, thanks Jeremy!

PR #1423, opened 2023-12-11
fluids: Support higher-order geometry

jrwrigh:
Really, verify support for higher-order geometry that was already present.

To verify, I took the `vortexshedding.yaml` problem, made the mesh higher order in gmsh, then ran with that higher order geometry. Result can be seen below:
image

Upper image has 2nd order geometry, lower has 1st order geometry. Both have 2nd order solutions on them. Below is the same, but just showing the mesh:
image


jedbrown:
Great! Cc: @AdelekeBankole

jedbrown (approved):

PR #1424, opened 2023-12-12
Minor function signature fix caught with ASan

sebastiangrimberg:
Silences the following warning:

```
libCEED/interface/ceed-vector.c:445:5: runtime error: call to function CeedVectorGetArrayWrite_Ref through pointer to incorrect function type 'int (*)(struct CeedVector_private *, CeedMemType, double **)'
ceed-ref-vector.c:138: note: CeedVectorGetArrayWrite_Ref defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior libCEED/interface/ceed-vector.c:445:5
```


jeremylt (approved):
Thanks for catching this

PR #1425, opened 2023-12-13
Consistency fix for a few missed CeedElemRestriction and CeedBasis variable names in the public API

sebastiangrimberg:
Fixes `CeedElemRestriction r, CeedBasis b, CeedVector v` -> `CeedElemRestriction rstr, CeedBasis basis, CeedVector vec` to match the rest of the library.

Also it looks like https://github.com/CEED/libCEED/pull/1409 introduced a lot of trailing whitespace changes in `interface/ceed-operator.cpp`. I'm not sure if that was intentional but I typically have trim-whitespace-on-save enabled and so they are stripped as part of this PR. Let me know if they were intentional and I can revert that chagne.

### DIFF
--- a/include/ceed/ceed.h
+++ b/include/ceed/ceed.h
@@ -398,7 +398,7 @@ CEED_EXTERN int CeedOperatorCreate(Ceed ceed, CeedQFunction qf, CeedQFunction dq
 CEED_EXTERN int CeedOperatorCreateAtPoints(Ceed ceed, CeedQFunction qf, CeedQFunction dqf, CeedQFunction dqfT, CeedOperator *op);
 CEED_EXTERN int CeedCompositeOperatorCreate(Ceed ceed, CeedOperator *op);
 CEED_EXTERN int CeedOperatorReferenceCopy(CeedOperator op, CeedOperator *op_copy);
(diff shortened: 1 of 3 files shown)


jedbrown:
Hmm, I had thought clang-format does this by default, but I now see there is some semantic where it tries to leave whitespace alone except in the context where something is changed. I have no problem with the whitespace fixes alongside the variable naming. Can we make clang-format stricter or do we need a different tool?

sebastiangrimberg:
Yeah, you are right that `clang-format` will only do it if other style changes are made in the file. Two options come to mind to automate this:
- Add something like `find $EXEC_PATH -type f \( -iname "*.hpp" -or -iname "*.h" -or -iname "*.c" -or -iname "*.cpp" \) -not -path "*/import/*" | xargs sed -e 's/[[:blank:]]\+$//' -i` to the `clean` rule in the Makefile.
- Perturb all files before running `clang-format` (not sure how, but the idea would be the perturbation is something which is reverted by clang-format).

Unsure if either of these make sense in this case and if it's worth addressing here, open to input.

jeremylt (approved):

PR #1426, opened 2023-12-14
Fix trailing spaces in non-linted files

jeremylt:
These files never get linted, so this wasn't caught before. No big deal, but it does mean we can search for ` \n` as a pattern and only `tests/README.md` has trailing whitespace on a line (on purpose).


jrwrigh (approved):
I've had success with [findent](https://sourceforge.net/projects/findent/) as a fortran formatter if we ever care to do that.

jedbrown (commented):
Can we add a `sed` to strip trailing whitespace to `make format` so it also gets enforced?

jeremylt:
Yeah, we can add that sed. I just need a moment to think about the right chain to exclude only the file `tests/README.md`

jedbrown:
`git ls-files | rg -v '^tests/README.md'` would do that, but I'm not following why that file must have trailing whitespace.

jeremylt:
That file is only viewed on GitHub or in local editors. In GitHub, the list is mangled to a single line for some reason unless each line has a trailing space

jedbrown:
Oh, that's because it's invalid markdown. This would fix it.
<details>

```diff
diff --git i/tests/README.md w/tests/README.md
index f3ce55fe0..f724bb342 100644
--- i/tests/README.md
+++ w/tests/README.md
@@ -4,30 +4,30 @@ This page provides a brief description of the tests for the libCEED library.
 
 The tests are organized by API object, and some tests are further organized, as required.
 
-0. Ceed Tests  
-1. CeedVector Tests  
-    1.0. CeedVector general tests  
-    1.1. CeedVector error tests  
-2. CeedElemRestriction Tests  
-3. CeedBasis Tests  
-    3.0. CeedBasis utility tests  
-    3.1. CeedBasis tensor basis tests  
-    3.2. CeedBasis simplex basis tests  
-    3.3. CeedBasis non-tensor H(div) basis tests  
-    3.4. CeedBasis non-tensor H(curl) basis tests  
-    3.5. CeedBasis evaluation at arbitrary points tests  
-4. CeedQFunction Tests  
-    4.0. CeedQFunction user code tests  
-    4.1. CeedQFunction gallery code tests  
-5. CeedOperator Tests  
-    5.0. CeedOperator with tensor bases tests  
-    5.1. CeedOperator with simplex bases tests  
-    5.2. CeedOperator with operator composition tests  
-    5.3. CeedOperator diagonal and CeedQFunction assembly tests  
-    5.4. CeedOperator element inverse tests  
-    5.5. CeedOperator multigrid level tests  
-    5.6. CeedOperator full assembly tests  
-    5.7. CeedOperator full assembly tests (extended)  
-    5.8. CeedOperator H(div) and H(curl) tests
-    5.9. CeedOperatorAtPoints tests
+0. Ceed Tests
+1. CeedVector Tests
+    0. CeedVector general tests
+    1. CeedVector error tests
+2. CeedElemRestriction Tests
+3. CeedBasis Tests
+    0. CeedBasis utility tests
+    1. CeedBasis tensor basis tests
+    2. CeedBasis simplex basis tests
+    3. CeedBasis non-tensor H(div) basis tests
+    4. CeedBasis non-tensor H(curl) basis tests
+    5. CeedBasis evaluation at arbitrary points tests
+4. CeedQFunction Tests
+    0. CeedQFunction user code tests
+    1. CeedQFunction gallery code tests
+5. CeedOperator Tests
+    0. CeedOperator with tensor bases tests
+    1. CeedOperator with simplex bases tests
+    2. CeedOperator with operator composition tests
+    3. CeedOperator diagonal and CeedQFunction assembly tests
+    4. CeedOperator element inverse tests
+    5. CeedOperator multigrid level tests
+    6. CeedOperator full assembly tests
+    7. CeedOperator full assembly tests (extended)
+    8. CeedOperator H(div) and H(curl) tests
+    9. CeedOperatorAtPoints tests
 
```
</details>

Makefile
```diff
@@ -760,14 +760,18 @@ AUTOPEP8_OPTS     += --in-place --aggressive --max-line-length 120
...
+format.ot := $(filter-out rust/libceed-sys/c-src/% doc/sphinx/source/CODE_OF_CONDUCT.md doc/sphinx/source/CONTRIBUTING.md, $(shell git ls-files '*.[ch]pp' '*.[ch]' '*.py' '*.jl' '*.md' '*.f90'))
 
 format-c  :
 	$(CLANG_FORMAT) $(CLANG_FORMAT_OPTS) $(format.ch)
 
 format-py :
 	$(AUTOPEP8) $(AUTOPEP8_OPTS) $(format.py)
 
-format    : format-c format-py
+format-ot:
```
jeremylt, line 771: How about this for a whitespace trimming rule? Do we want to change the list of targeted files in any way?

jedbrown: I think this can race with `format-c`. Maybe stripping whitespace for C sources can go into `format-c`.

jeremylt: I think only comment blocks are at risk of trailing whitespace right now in C files.

For C, the funkiness comes  from
```
ReflowComments:               false
```
I turned this on to prevent this sort of silliness
```C
/**
  A long comment with one more word than Clang
likes.
**/
```
where we have a line with just one word on it for no particular reason. For some reason that option means that trailing spaces are allowed in comment blocks?

jeremylt: Trimmed to .f90 and .md, which aren't covered by other linting

PR #1431, opened 2023-12-20
Silence compiler warning for `-Wstrict-prototypes`

sebastiangrimberg:
Building with Clang 17 and `PEDANTIC=1`, I get:

```
./include/ceed/backend.h:239:114: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
  239 | CEED_EXTERN int CeedSetBackendFunction(Ceed ceed, const char *type, void *object, const char *func_name, int (*f)());
```

This PR resolves these errors.

### DIFF
--- a/include/ceed/backend.h
+++ b/include/ceed/backend.h
@@ -222,7 +222,7 @@ CEED_INTERN int CeedFree(void *p);
 #define CeedCalloc(n, p) CeedCallocArray((n), sizeof(**(p)), p)
 #define CeedRealloc(n, p) CeedReallocArray((n), sizeof(**(p)), p)
 
-/* Allow users to call CeedSetBackendFunctionImpl using incompatible pointer types */
+/* Allows calling CeedSetBackendFunctionImpl using incompatible pointer types */
 #define CeedSetBackendFunction(ceed, type, object, func_name, f) CeedSetBackendFunctionImpl(ceed, type, object, func_name, (void (*)(void))f)
 
 CEED_EXTERN int CeedRegister(const char *prefix, int (*init)(const char *, Ceed), unsigned int priority);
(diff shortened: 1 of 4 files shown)


sebastiangrimberg:
As an aside, the other annoying Clang warning with pedantic is:

```
./include/ceed/backend.h:110:54: error: token pasting of ',' and __VA_ARGS__ is a GNU extension [-Werror,-Wgnu-zero-variadic-macro-arguments]
  110 |     if (CeedDebugFlag(ceed)) CeedDebugImpl256(color, ##__VA_ARGS__); \
```

I'm not sure of a good fix for this, it seems like the token pasting (https://gcc.gnu.org/onlinedocs/gcc/Variadic-Macros.html) is actually genuinely useful here and I'm not sure if it's worth changing. Probably it's fine for users to just provide `-Wgnu-zero-variadic-macro-arguments` to silence this?

include/ceed/backend.h
```diff
@@ -222,6 +222,9 @@ CEED_INTERN int CeedFree(void *p);
 #define CeedCalloc(n, p) CeedCallocArray((n), sizeof(**(p)), p)
 #define CeedRealloc(n, p) CeedReallocArray((n), sizeof(**(p)), p)
 
+/* Allow users to call CeedSetBackendFunctionImpl using incompatible pointer types */
```
jeremylt, line 225: This isn't something "users" should be using, only library backend developers.

interface/ceed.c
```diff
@@ -570,7 +570,7 @@ int CeedSetDeterministic(Ceed ceed, bool is_deterministic) {
 
   @ref Backend
 **/
-int CeedSetBackendFunction(Ceed ceed, const char *type, void *object, const char *func_name, int (*f)()) {
+int CeedSetBackendFunctionImpl(Ceed ceed, const char *type, void *object, const char *func_name, void (*f)(void)) {
```
jeremylt, line 573: I'm a bit confused as to why this is a `void (*f)(void)` instead of `int (*f)(void)` - shouldn't all our internal backend functions return an error code?

sebastiangrimberg: GCC warns when casting function pointers to/from anything other than `void (*)(void)`. Inside the actual definition you can see that the `void (*)(void)` generic type is cast to `int (*)(void)` for `fpointer`.

jeremylt (approved):

PR #1432, opened 2023-12-20
Define CeedPragmaSIMD for Clang

sebastiangrimberg:
Clang has support for `#pragma omp simd` and I am pretty sure this should be enabled even when not building with `-fopenmp`, if I understand correctly.

### DIFF
--- a/include/ceed/types.h
+++ b/include/ceed/types.h
@@ -87,7 +87,9 @@ values for CPU backends.
 /// Miscompilation with pragma ivdep observed with icc (ICC) 19.0.5.281 20190815 at -O2 and above.
 #elif defined(__GNUC__) && __GNUC__ >= 5
 #define CeedPragmaSIMD _Pragma("GCC ivdep")
-#elif defined(__clang__) || (defined(_OPENMP) && _OPENMP >= 201307)  // OpenMP-4.0 (July, 2013)
+#elif defined(__clang__)
+#define CeedPragmaSIMD _Pragma("clang loop vectorize(enable)")
+#elif defined(_OPENMP) && _OPENMP >= 201307  // OpenMP-4.0 (July, 2013)
 #define CeedPragmaSIMD _Pragma("omp simd")
 #else
 #define CeedPragmaSIMD
--- a/include/ceed/types.h
+++ b/include/ceed/types.h
@@ -87,7 +87,7 @@ values for CPU backends.
 /// Miscompilation with pragma ivdep observed with icc (ICC) 19.0.5.281 20190815 at -O2 and above.
 #elif defined(__GNUC__) && __GNUC__ >= 5
 #define CeedPragmaSIMD _Pragma("GCC ivdep")
-#elif defined(_OPENMP) && _OPENMP >= 201307  // OpenMP-4.0 (July, 2013)
+#elif defined(__clang__) || (defined(_OPENMP) && _OPENMP >= 201307)  // OpenMP-4.0 (July, 2013)
 #define CeedPragmaSIMD _Pragma("omp simd")
 #else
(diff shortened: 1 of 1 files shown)


jedbrown:
So clang accepts the pragma without warnings, but it still needs `-fopenmp-simd` to vectorize properly. Try removing that flag here.
https://godbolt.org/z/8f4rssWxP
Is there a better pragma/attribute that always works with clang?

sebastiangrimberg:
Ah, interesting. I guess a close equivalent of `#pragma GCC ivdep` for GCC seems like it would be `#pragma clang loop vectorize(assume_safety)`? (From my 5 minutes of Googling, I didn't actually try it but can tomorrow.)

sebastiangrimberg:
OK, what I'm learning is auto-vectorization is very testy 😄 

But it appears for Clang we should use `clang loop vectorize(enable)`. This vectorizes with `-O2` and no `-fopenmp-simd`: https://godbolt.org/z/4nM7ozT3r.

However, I could not get anything to vectorize with GCC 13 until I got to `-Ofast` from what I can tell: https://godbolt.org/z/ead6EWod1.

I guess I'm not sure what the best sequence of priorities is for defining `CeedPragmaSIMD`. From what I read online it seems like `omp simd` is the most aggressive and since the Makefile adds `-fopenmp-simd` or the appropriate flag automatically maybe this should be first? That said there's probably a reason the legacy Intel compiler and GCC are prioritized currently, which was why I was experimenting with GCC above and confused about it not doing a great job at vectorizing.

jedbrown:
> However, I could not get anything to vectorize with GCC 13 until I got to `-Ofast` from what I can tell

Here you can see it vectorize (but not unroll) with `-O2 -fopenmp-simd`, though you need to use proper omp reduction syntax to allow reassociation of `sum`.

https://godbolt.org/z/v3TY8M54s

I can't think of any important libCEED operations that rely on reduction semantics like this so while this example is good for learning, I don't think it's representative. Note that `restrict` has no effect because we aren't writing to an array that could (without the qualifier) alias.

sebastiangrimberg:
Oh of course, I didn’t know that the OpenMP SIMD variant supported reduction clauses as well, very nice. I guess a related question I have to the changes in this PR is should we just make the OpenMP SIMD the highest priority pragma? The default Makefile flags are already adding `-openmp-simd`, so is there a way to prefer this pragma by default reliably? From what I am reading it seems like it typically results in the most aggressive vectorization, but also maybe you already feel that it doesn’t matter much so prioritizing the specialized Intel/GCC/(newly added) clang pragmas first would be fine. I’m curious what you think would be the best approach here.

jedbrown:
So `-fopenmp-simd` is not testable (it does not define `_OPENMP` or any other macros). I believe `GCC ivdep` was at least as good for vectorized math (our actual use case) in my experiments some years ago, and it's trivial to test that it's available.

I'm not tied to any specific choice, but I'd rather not make it too easy to either (a) get warnings due to using an unsupported pragma or (b) silently not use the pragma and have terrible performance. It's not uncommon for people to `make OPT='-O3 -march=native'`, and that would currently lose the flags. This also applies to compilation of qfunctions in user code, where we have even less visibility as to what flags are used. So the vendor-specific pragmas seemed like the safest. What do you think?

sebastiangrimberg:
I agree that all makes sense. Especially a good point on user written QFunctions not being necessarily subject to the same compiler flags. In that case, I think this PR should be merged so that users get (close to) the same auto-vectorization experience for Clang as GCC without needing to add `-fopenmp` to trigger the `omp simd` pragma. Do i understand correctly?

jedbrown (approved):
Sounds good. Approved for squash-merge.

PR #1435, opened 2023-12-22
GPU - Avoid calling diagonal assembly kernels if input arrays are null pointers

### LINKED ISSUE #1434: `CeedOperatorAssembleDiagonalCore_*` fails on GPU for active fields with `CEED_BASIS_NONE`
Attempting to call `CeedOperatorLinearAssembleDiagonal` on an operator with active fields defined with `CEED_BASIS_NONE` leads to a crash, see [this thread on Ratel](https://gitlab.com/micromorph/ratel/-/merge_requests/394#note_1705073430). 

The issue appears to be due to the fact that `CEED_BASIS_NONE` causes `diag->num_nodes` to be zero, and thus all of the calls to `cuMalloc` and `hipMalloc` return null pointers. This causes an error when attempting to run the kernels on the GPU. 

The fix is simple, we should not call the `CeedRunKernel_*` functions if `diag->num_nodes == 0` (since the result of that operation would not change the output vector anyway).

zatkins-dev:
Resolves issue described in #1434 by only applying diagonal assembly kernels when the input arrays are set to valid memory addresses, i.e., when the number of basis nodes is non-zero.

Closes #1434.

We may want a similar fix for the full assembly kernels as well.

### DIFF
--- a/backends/cuda-ref/ceed-cuda-ref-operator.c
+++ b/backends/cuda-ref/ceed-cuda-ref-operator.c
@@ -809,25 +809,28 @@ static inline int CeedOperatorAssembleDiagonalCore_Cuda(CeedOperator op, CeedVec
   }
   CeedCallBackend(CeedVectorSetValue(elem_diag, 0.0));
 
-  // Assemble element operator diagonals
-  CeedCallBackend(CeedVectorGetArray(elem_diag, CEED_MEM_DEVICE, &elem_diag_array));
-  CeedCallBackend(CeedVectorGetArrayRead(assembled_qf, CEED_MEM_DEVICE, &assembled_qf_array));
-  CeedCallBackend(CeedElemRestrictionGetNumElements(diag_rstr, &num_elem));
+  // Only assemble diagonal if the basis has nodes, otherwise inputs are null pointers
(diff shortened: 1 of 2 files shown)


zatkins-dev:
We may want a test for linear assemble/linear assemble diagonal with `CEED_BASIS_NONE` as well.

jeremylt:
Sycl probably has the same code + could be separate pr

zatkins-dev:
> Sycl probably has the same code + could be separate pr

I added it to this MR -- I'm already doing CUDA and HIP

jeremylt:
But we cannot merge this PR today with that commit since we have to wait for the sycl commit to be tested

zatkins-dev:
Oh, I didn't realize that wasn't automated. I'll roll back

zatkins-dev:
Also, I think the same issue will pop up in full assembly, and the fix should be about the same. Should I add that here?

jeremylt:
Lets get this clean win in so your Ratel branch passes ci and then follow up improvements in separate PRs. We can merge this on once CI completes

jeremylt (approved):

PR #1436, opened 2023-12-23
Add deal.II example

peterrum:
This PR adds a libCEED example using deal.II. It solves BP1-BP6 and shows how one would implement an operator with `libCEED` and with the native matrix-free infrastructure of deal.II.

The code was written jointly with @kronbichler.

I discussed with @jedbrown that such an example would be useful. I am not sure what the next steps are. Where can I add tests? For deal.II, there are docker images that could be used: https://hub.docker.com/layers/dealii/dealii/latest/images/sha256-cb75bc036f4309e9d920036b1e4d22f1eba09e11072874c6a297a227126c82da?context=explore.


jeremylt:
I can jump on this after New Years, but I think hooking this into CI with that container should be pretty easy.

examples/deal.II/bp.h
```diff
@@ -0,0 +1,1165 @@
...
+        AssertThrow(n_components == 1, ExcInternalError());
+      }
+    else
+      {
+        AssertThrow(n_components == dim, ExcInternalError());
+      }
+
+    // 1) create CEED instance -> "MatrixFree"
+    const char *ceed_spec = "/cpu/self/avx/blocked";
+    CeedInit(ceed_spec, &ceed);
```
peterrum, line 491: > How do we set the libCEED backend when we run the test?

This is hard coded for now. But this can be changed.

examples/deal.II/README.MD
```diff
@@ -0,0 +1,24 @@
...
+Run the executable to create an input file:
+
+```
+./bp > input.json
+```
+
+Modify the input file and run the executable:
+```
+mpirun -np 8 ./bp input.json
+```
```
peterrum, line 23: > What are the command line options to change the test parameters such as number of elements and FE order? (I'll read the docs and the code to help get me oriented here)

I went with a json file. Currently, number of global refinements, fe degree, and type of BP can be changed.

examples/deal.II/bp.h
```diff
@@ -0,0 +1,1165 @@
...
+{
+  CeedInt dim, space_dim;
+};
+
+
+
+/**
+ * libCEED Q-function for building quadrature data for a mass operator
+ */
+CEED_QFUNCTION(f_build_mass)
```
peterrum, line 102: > I think we'll need the QFunction source in a separate file to support the GPU backends (or use the gallery, which I can help add [or do both like ex1/2 in the examples/ceed folder])

Sure. I was not able to test the code on a GPU since I currently, don't have access to one.

examples/deal.II/bp.h
```diff
@@ -0,0 +1,1165 @@
...
+
+    // 1) create CEED instance -> "MatrixFree"
+    const char *ceed_spec = "/cpu/self/avx/blocked";
+    CeedInit(ceed_spec, &ceed);
+
+    // 2) create shape functions -> "ShapeInfo"
+    const unsigned int fe_degree  = fe.tensor_degree();
+    const unsigned int n_q_points = quadrature.get_tensor_basis()[0].size();
+    CeedBasisCreateTensorH1Lagrange(
+      ceed, dim, n_components, fe_degree + 1, n_q_points, CEED_GAUSS, &sol_basis);
```
peterrum, line 497: > It would be cool if we could build the CeedBasis from information from the deal.II FE space

I like this suggestion. I didn't feel too happy with the approach I have taken here.

peterrum (commented):
> I can jump on this after New Years, but I think hooking this into CI with that container should be pretty easy.

@jeremylt  Great! Thank you! It is the first time I used libCEED so I probably did a couple of naive mistakes...

Below the answers to your questions!

examples/deal.II/bp.h
jeremylt: What do you think about us focusing on some small tweaks to get this new code running in our CI pipeline in this PR and then making this update in a follow-up? I don't think it should be too much extra work, but I don't want to block merging with a bunch of little changes

examples/deal.II/bp.h
jeremylt: I still haven't tested yet, but I think we should be good on GPUs now.

examples/deal.II/bp.h
jeremylt: This would go in the `Parameters` struct to make it runtime selectable if I'm understanding correctly? Selecting the libCEED backend resource at runtime is the big change needed to get this working in our CI pipeline.

examples/deal.II/README.MD
jeremylt: Is this JSON the only way to modify the runtime arguments? I think we want a default version of the JSON in the repo for testing and then we can copy this default JSON for each test run, append the ceed backend string, and then delete the copy of the JSON once the test finishes.

jedbrown: Files (and necessary care to avoid race conditions) could be avoided if the json can be read from stdin or an argument, as in `./bp "$(sed s,RESOURCE,/gpu/cuda, input.json)"`?

jeremylt: Yeah, I'd prefer to avoid writing files is there is a way to use only the command line.

peterrum: I have implemented two options:
 - https://github.com/CEED/libCEED/pull/1436/commits/a09fefd3353fec5b52e44ad9514015d0372486a6 -> json can be passed via command line as suggested by @jedbrown 
 - https://github.com/CEED/libCEED/pull/1436/commits/a538308c7fe1c18af133aa14ba22f88fcfcd8662 -> all options are passed via command line; I think this would be your preference, @jeremylt?

peterrum: Any preferences? At the moment, I tend to the second option. Once agreed, I would update the readme file and squash the commits.

jrwrigh: Not speak for everyone, but I believe the latter (options via command line) will be the overall preferred. That's what the test suite handles most easily. 

I think even passing json via command line would require some modifications to the test suite to handle.

jedbrown: Yeah, the command line options aren't that much more code/maintenance effort, and I think they'll be more familiar to libCEED users experimenting with deal.II or comparing the different implementations.

peterrum: OK. I went with the command-line arguments. What is the next step?

peterrum:
> I need to play around a bit and see how best to link the docker container into our CI.

@jeremylt Is there anything I can do to help?

examples/deal.II/README.MD
```diff
@@ -0,0 +1,20 @@
...
+make
+```
+
+To run the executable, write:
+
+```
+./bp
+```
+
+Optional command-line arguments are shown by adding the command-line argument "--help".
```
jeremylt, line 20: I think my biggest frustration right now is trying to figure out the best way to use this all from our top level makefile during our CI. I can't seem to even get cmake to actually correctly recognize the root libCEED dir

peterrum: >  top level makefile during our CI.

Could you point me to the relevant file.?

jedbrown: In the top-level makefile, see the logic for `solidsexamples` or `mfemexamples`. So if `DEAL_II_DIR` is set, we need a rule to make an executable (or symlink to executable) that can be run (it gets run by `junit.py`).

jeremylt: Yeah, here is the relevant parts of the toplevel makefile:

https://github.com/CEED/libCEED/blob/b698cfb9c4f7c41a310553a1468bd8185bfedcf5/Makefile#L243

https://github.com/CEED/libCEED/blob/b698cfb9c4f7c41a310553a1468bd8185bfedcf5/Makefile#L581

Ultimately, the top level makefile needs to copy the executable to `/build/dealii-bps` and then test suite will read the contents of `examples/deal.II/bps.cc` to look for a comment block like this to get the test cases

https://github.com/CEED/libCEED/blob/b698cfb9c4f7c41a310553a1468bd8185bfedcf5/examples/petsc/bps.c#L27https://github.com/CEED/libCEED/blob/b698cfb9c4f7c41a310553a1468bd8185bfedcf5/examples/petsc/bps.c#L27

I can set up the comment block, I'm mostly having trouble seeing the best way to get the top level makefile to build the deal.II example.

Our only example that uses a something other than a makefile to build currently is Nek5000, but that is a very fragile setup. The MFEM example is probably the simplest example building we have to look at?

examples/deal.II/bp.cc
```diff
@@ -0,0 +1,235 @@
+// ---------------------------------------------------------------------
```
jeremylt, line 1: minor note - for MFEM, NEK5000, and PETSc we call this file `bps.*` since it can do all 6 BPs

jeremylt:
hmm, I'm getting this error when I try to use my own build of deal.II with either the main branch or the last release. I have to be missing something?

```console
$ make
[ 50%] Building CXX object CMakeFiles/bp.dir/bp.cc.o
/home/jeremy/Dev/libCEED/examples/deal.II/bp.cc: In function ‘int main(int, char**)’:
/home/jeremy/Dev/libCEED/examples/deal.II/bp.cc:163:64: error: use of deleted function ‘dealii::parallel::distributed::Triangulation<dim, spacedim>::Triangulation(MPI_Comm, typename dealii::Triangulation<dim, spacedim>::MeshSmoothing, Settings) [with int dim = 2; int spacedim = 2; MPI_Comm = int; typename dealii::Triangulation<dim, spacedim>::MeshSmoothing = dealii::Triangulation<2, 2>::MeshSmoothing]’
  163 |   parallel::distributed::Triangulation<dim> tria(MPI_COMM_WORLD);
      |                                                                ^
In file included from /home/jeremy/Dev/libCEED/examples/deal.II/bp.cc:24:
/home/jeremy/Dev/dealii/install/include/deal.II/distributed/tria.h:1024:16: note: declared here
 1024 |       explicit Triangulation(
      |                ^~~~~~~~~~~~~
In file included from /home/jeremy/Dev/libCEED/examples/deal.II/bp.cc:48:
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h: In instantiation of ‘void OperatorCeed<dim, Number>::reinit() [with int dim = 2; Number = double]’:
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:155:5:   required from ‘OperatorCeed<dim, Number>::OperatorCeed(const dealii::Mapping<dim>&, const dealii::DoFHandler<dim>&, const dealii::AffineConstraints<number>&, const dealii::Quadrature<dim>&, const BPType&, const std::string&) [with int dim = 2; Number = double; std::string = std::__cxx11::basic_string<char>]’
/home/jeremy/Dev/libCEED/examples/deal.II/bp.cc:229:78:   required from here
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:250:13: warning: missing braces around initializer for ‘std::__array_traits<int, 3>::_Type’ {aka ‘int [3]’} [-Wmissing-braces]
  250 |     strides = {1, quadrature.size(), quadrature.size() * n_components_metric};
      |     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:250:13: warning: missing braces around initializer for ‘std::__array_traits<int, 3>::_Type’ {aka ‘int [3]’} [-Wmissing-braces]
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:250:13: warning: missing braces around initializer for ‘std::__array_traits<int, 3>::_Type’ {aka ‘int [3]’} [-Wmissing-braces]
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:250:34: warning: narrowing conversion of ‘(&((OperatorCeed<2, double>*)this)->OperatorCeed<2, double>::quadrature)->dealii::Quadrature<2>::size()’ from ‘unsigned int’ to ‘int’ [-Wnarrowing]
  250 |     strides = {1, quadrature.size(), quadrature.size() * n_components_metric};
      |                   ~~~~~~~~~~~~~~~^~
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:250:56: warning: narrowing conversion of ‘((&((OperatorCeed<2, double>*)this)->OperatorCeed<2, double>::quadrature)->dealii::Quadrature<2>::size() * ((unsigned int)n_components_metric))’ from ‘unsigned int’ to ‘int’ [-Wnarrowing]
  250 |     strides = {1, quadrature.size(), quadrature.size() * n_components_metric};
      |                                      ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h: In instantiation of ‘static std::vector<double> OperatorCeed<dim, Number>::compute_metric_data(Ceed_private* const&, const dealii::Mapping<dim>&, const dealii::Triangulation<dim, dim>&, const dealii::Quadrature<dim>&, BPType) [with int dim = 2; Number = double; Ceed = Ceed_private*]’:
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:248:40:   required from ‘void OperatorCeed<dim, Number>::reinit() [with int dim = 2; Number = double]’
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:155:5:   required from ‘OperatorCeed<dim, Number>::OperatorCeed(const dealii::Mapping<dim>&, const dealii::DoFHandler<dim>&, const dealii::AffineConstraints<number>&, const dealii::Quadrature<dim>&, const BPType&, const std::string&) [with int dim = 2; Number = double; std::string = std::__cxx11::basic_string<char>]’
/home/jeremy/Dev/libCEED/examples/deal.II/bp.cc:229:78:   required from here
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:573:45: warning: narrowing conversion of ‘(& quadrature)->dealii::Quadrature<2>::size()’ from ‘unsigned int’ to ‘CeedInt’ {aka ‘int’} [-Wnarrowing]
  573 |     CeedInt strides[3] = {1, quadrature.size(), quadrature.size() * n_components};
      |                              ~~~~~~~~~~~~~~~^~
/home/jeremy/Dev/libCEED/examples/deal.II/bp.h:573:67: warning: narrowing conversion of ‘((& quadrature)->dealii::Quadrature<2>::size() * ((unsigned int)n_components))’ from ‘unsigned int’ to ‘CeedInt’ {aka ‘int’} [-Wnarrowing]
  573 |     CeedInt strides[3] = {1, quadrature.size(), quadrature.size() * n_components};
      |                                                 ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/bp.dir/build.make:76: CMakeFiles/bp.dir/bp.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/bp.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
```

kronbichler:
The error message `use of deleted function` indicates that deal.II was not built with the p4est library, which is required by this particular test case. I think we could also make it work with just METIS as partitioning tool, but either variant will have an external dependency. We should make sure that this is reflected in our `CMakeLists.txt` scripts.

jeremylt:
Haha! Thank you, that's what I was missing!

jeremylt:
Great success!

```console
$ ./bp --resource /cpu/self --bp BP4 --fe_degree 3
ceed: 6 11.6387 0.000544714
dealii: 6 11.6387 0.000640734
```

We have a working install on our CI machine!

examples/deal.II/bp.cc
```diff
@@ -0,0 +1,235 @@
...
+// it, and/or modify it under the terms of the GNU Lesser General
+// Public License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// The full text of the license can be found in the file LICENSE.md at
+// the top level directory of deal.II.
+//
+//  Authors: Peter Munch, Martin Kronbichler
+//
+// ---------------------------------------------------------------------
+
```
jeremylt, line 17:
```suggestion

//TESTARGS(name="BP1") --resource {ceed_resource} --bp BP1 --fe_degree 2
//TESTARGS(name="BP4") --resource {ceed_resource} --bp BP4 --fe_degree 3

```

I think this should be a reasonable set of test cases? Our test harness would then run every libCEED backend against these two test cases once the top level Makefile is set up to correctly build this example

examples/deal.II/bp.cc
```diff
@@ -0,0 +1,235 @@
...
+    const auto time =
+      std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now() - now)
+        .count() /
+      1e9;
+
+
+    pout << label << ": " << reduction_control.last_step() << " " << v.l2_norm() << " "
+         << (params.print_timings ? time : 0.0) << std::endl;
+  };
```
jeremylt, line 225:
```suggestion
    if (params.print_timings || SOME SOLUTION ERROR TOLERANCE)
      {
          pout << label << ": " << reduction_control.last_step() << " " << v.l2_norm() << " "
             << time << std::endl;
       }
  };
```
I think something like this would give us no output during testing. Our testing script treats all output as an indication of an error.

examples/deal.II/bp.cc
```diff
@@ -0,0 +1,235 @@
...
+    try
+      {
+        // solve problem
+        SolverCG<VectorType> solver(reduction_control);
+        now = std::chrono::system_clock::now();
+        solver.solve(op, v, u, diagonal_matrix);
+      }
+    catch (const SolverControl::NoConvergence &)
+      {
+        // nothing to do
```
jeremylt, line 213:
```suggestion
        pout << "Error: solver failed to converge" << std::endl;
```
How about this? For us, this would likely indicate that the libCEED backend has some bug that is spoiling convergence.

examples/deal.II/bp.cc
peterrum: Looks good to me!

examples/deal.II/bp.cc
peterrum: OK. I have made the change. I am surprised that you don't check the iteration numbers in CI.

examples/deal.II/bp.cc
peterrum: Good idea!

peterrum:
> Great success!

Excellent! Thank you, @jeremylt!!!

> Ahha! Thank you, that's what I was missing!

@kronbichler Is right.

I have pushed two versions. Before the last commit, `p4est` is required. In the last commit, `p4est` is made optional and a different mesh class is used. This mesh class optionally uses `METIS` for partitioning, however, falls back to some more basic partitioning schemes if that cannot be detected. I would go with the latter option. What do you think?

I have also fixed the warnings that are shown in the output above.

jedbrown:
Thanks, I think being able to run with minimal dependencies is desirable.

examples/deal.II/bp.cc
jeremylt: We usually use the L2 error between the target and computed solution to determine correctness in situations like this where there is an analytic solution

jeremylt:
Yeah, everything looks great for me locally

If we go the Nek5000 style route for building, then we'd add

```diff
$ git diff
diff --git a/Makefile b/Makefile
index f5de43e3..d5e79861 100644
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,12 @@ endif
 export NEK5K_DIR
 MPI ?= 1
 
+# DEAL_II_DIR env variable should point to sibling directory
+ifneq ($(wildcard ../dealii/install/lib/libdeal_II.*),)
+  DEAL_II_DIR ?= ../dealii/install
+endif
+export DEAL_II_DIR
+
 # CEED_DIR env for NEK5K testing
 export CEED_DIR = $(abspath .)
 
@@ -248,6 +254,8 @@ nekexamples  := $(OBJDIR)/nek-bps
 # PETSc Examples
 petscexamples.c := $(wildcard examples/petsc/*.c)
 petscexamples   := $(petscexamples.c:examples/petsc/%.c=$(OBJDIR)/petsc-%)
+# deal.II Examples
+dealiiexamples  := $(OBJDIR)/dealii-bps
 # Fluid Dynamics Examples
 fluidsexamples.c  := $(sort $(wildcard examples/fluids/*.c))
 fluidsexamples.py := examples/fluids/smartsim_regression_framework.py
@@ -344,6 +352,7 @@ info:
        $(info MFEM_DIR      = $(MFEM_DIR))
        $(info NEK5K_DIR     = $(NEK5K_DIR))
        $(info PETSC_DIR     = $(PETSC_DIR))
+       $(info DEAL_II_DIR    = $(DEAL_II_DIR))
        $(info ------------------------------------)
        $(info prefix        = $(prefix))
        $(info includedir    = $(value includedir))
@@ -605,6 +614,13 @@ $(OBJDIR)/petsc-% : examples/petsc/%.c examples/petsc/libutils.a.PHONY $(libceed
          PETSC_DIR="$(abspath $(PETSC_DIR))" OPT="$(OPT)" $*
        cp examples/petsc/$* $@
 
+# Note: Invoking deal.II's CMAKE build system here
+$(OBJDIR)/dealii-bps : $(libceed) | $$(@D)/.DIR
+       mkdir -p examples/deal.II/build
+       cmake -B examples/deal.II/build -S examples/deal.II -DDEAL_II_DIR=$DEAL_II_DIR -DCEED_DIR=$(PWD)
+       make -C examples/deal.II/build
+       cp examples/deal.II/build/bps $(OBJDIR)/dealii-bps
+
 $(OBJDIR)/fluids-% : examples/fluids/%.c examples/fluids/src/*.c examples/fluids/*.h examples/fluids/problems/*.c examples/fluids/qfunctions/*.h $(libceed) $(ceed.pc) | $$(@D)/.DIR
        +$(call quiet,MAKE) -C examples/fluids CEED_DIR=`pwd` \
          PETSC_DIR="$(abspath $(PETSC_DIR))" OPT="$(OPT)" $*
@@ -640,6 +656,7 @@ external_examples := \
        $(if $(MFEM_DIR),$(mfemexamples)) \
        $(if $(PETSC_DIR),$(petscexamples)) \
        $(if $(NEK5K_DIR),$(nekexamples)) \
+       $(if $(DEAL_II_DIR),$(dealiiexamples)) \
        $(if $(PETSC_DIR),$(fluidsexamples)) \
        $(if $(PETSC_DIR),$(solidsexamples))
 
diff --git a/examples/deal.II/bps.cc b/examples/deal.II/bps.cc
index de492d0f..eece53a6 100644
--- a/examples/deal.II/bps.cc
+++ b/examples/deal.II/bps.cc
@@ -48,6 +48,9 @@
 // include operators
 #include "bps.h"
 
+// Test cases
+//TESTARGS(name="BP1") --resource {ceed_resource} --bp BP1 --fe_degree 2 --print_timings 0
+//TESTARGS(name="BP4") --resource {ceed_resource} --bp BP4 --fe_degree 3 --print_timings 0
 
 /**
  * Relevant parameters.
diff --git a/tests/junit.py b/tests/junit.py
index ec4af59e..c79a4598 100755
--- a/tests/junit.py
+++ b/tests/junit.py
@@ -50,6 +50,8 @@ class CeedSuiteSpec(SuiteSpec):
             return (Path('examples') / 'mfem' / rest).with_suffix('.cpp')
         elif prefix == 'nek':
             return (Path('examples') / 'nek' / 'bps' / rest).with_suffix('.usr')
+        elif prefix == 'dealii':
+            return (Path('examples') / 'deal.II' / rest).with_suffix('.cc')
         elif prefix == 'fluids':
             return (Path('examples') / 'fluids' / rest).with_suffix('.c')
         elif prefix == 'solids':
diff --git a/tests/junit_common.py b/tests/junit_common.py
index c1733def..29d4d372 100644
--- a/tests/junit_common.py
+++ b/tests/junit_common.py
@@ -236,7 +236,7 @@ def get_test_args(source_file: Path) -> List[TestSpec]:
         List[TestSpec]: List of parsed `TestSpec` objects, or a list containing a single, default `TestSpec` if none were found
     """
     comment_str: str = ''
-    if source_file.suffix in ['.c', '.cpp']:
+    if source_file.suffix in ['.c', '.cc', '.cpp']:
         comment_str = '//'
     elif source_file.suffix in ['.py']:
         comment_str = '#'
```

Unfortunately, BP4 doesn't converge with the memcheck backend. I'm investigating.

examples/deal.II/bps.h
```diff
@@ -0,0 +1,878 @@
...
+    /**
+     * Constructor.
+     */
+    VectorTypeCeed(const Ceed &ceed, const VectorType &vec)
+    {
+      const unsigned int n_dofs =
+        vec.get_partitioner()->locally_owned_size() + vec.get_partitioner()->n_ghost_indices();
+
+      CeedVectorCreate(ceed, n_dofs, &vec_ceed);
+      CeedVectorSetArray(vec_ceed, CEED_MEM_HOST, CEED_USE_POINTER, vec.get_values());
```
jeremylt, line 402: I think this is be the problem I founding using the memcheck backend, which would also be an issue for all GPU backends. In our GPU backends (which the memcheck backend emulates), each CeedVector has two memory spaces, `CEED_MEM_HOST` and `CEED_MEM_DEVICE`. The CeedVector manages synchronization between these two spaces, but only if the caller uses `CeedVectorSetArray` and `CeedVectorTakeArray` in pairs.

Here's how we do that in the PETSc examples every time `CeedOperatorApply` is called
```C
  // Setup CEED vectors
  PetscCall(VecGetArrayReadAndMemType(user->X_loc, (const PetscScalar **)&x, &x_mem_type));
  PetscCall(VecGetArrayAndMemType(user->Y_loc, &y, &y_mem_type));
  CeedVectorSetArray(user->x_ceed, MemTypeP2C(x_mem_type), CEED_USE_POINTER, x);
  CeedVectorSetArray(user->y_ceed, MemTypeP2C(y_mem_type), CEED_USE_POINTER, y);

  // Apply CEED operator
  CeedOperatorApply(user->op, user->x_ceed, user->y_ceed, CEED_REQUEST_IMMEDIATE);

  // Restore PETSc vectors
  CeedVectorTakeArray(user->x_ceed, MemTypeP2C(x_mem_type), NULL);
  CeedVectorTakeArray(user->y_ceed, MemTypeP2C(y_mem_type), NULL);
  PetscCall(VecRestoreArrayReadAndMemType(user->X_loc, (const PetscScalar **)&x));
  PetscCall(VecRestoreArrayAndMemType(user->Y_loc, &y));
```

I'm not sure where best to put that logic though. Just adding
```C
    /**
     * Destructor: destroy vector view.
     */
    ~VectorTypeCeed()
    {
      CeedScalar *ptr;
      CeedVectorTakeArray(vec_ceed, CEED_MEM_HOST, &ptr);
      CeedVectorDestroy(&vec_ceed);
    }
```
isn't enough

examples/deal.II/bps.h
```diff
@@ -0,0 +1,878 @@
...
+        VectorTypeCeed src_ceed(ceed, src_tmp);
+        VectorTypeCeed dst_ceed(ceed, dst_tmp);
+
+        copy_to_block_vector(src_tmp, src); // copy to block vector
+
+        // apply operator
+        CeedOperatorApply(op_apply, src_ceed(), dst_ceed(), CEED_REQUEST_IMMEDIATE);
+
+        copy_from_block_vector(dst, dst_tmp); // copy from block vector
+      }
```
jeremylt, line 338: Ah, I think the issue is here. This seems to fix it, though I'm not sure if there is a more idiomatic approach

```diff
$ git diff
diff --git a/examples/deal.II/bps.h b/examples/deal.II/bps.h
index cfde7f7f..7a76f985 100644
--- a/examples/deal.II/bps.h
+++ b/examples/deal.II/bps.h
@@ -325,14 +325,17 @@ public:
         src_tmp.reinit(this->extended_local_size(), true);
         dst_tmp.reinit(this->extended_local_size(), true);
 
-        // create libCEED view on deal.II vectors
-        VectorTypeCeed src_ceed(ceed, src_tmp);
-        VectorTypeCeed dst_ceed(ceed, dst_tmp);
+        // copy to block vector
+        copy_to_block_vector(src_tmp, src);
 
-        copy_to_block_vector(src_tmp, src); // copy to block vector
+        {
+          // create libCEED view on deal.II vectors
+          VectorTypeCeed src_ceed(ceed, src_tmp);
+          VectorTypeCeed dst_ceed(ceed, dst_tmp);
 
-        // apply operator
-        CeedOperatorApply(op_apply, src_ceed(), dst_ceed(), CEED_REQUEST_IMMEDIATE);
+          // apply operator
+          CeedOperatorApply(op_apply, src_ceed(), dst_ceed(), CEED_REQUEST_IMMEDIATE);
+        }
 
         copy_from_block_vector(dst, dst_tmp); // copy from block vector
       }
@@ -416,6 +419,8 @@ private:
      */
     ~VectorTypeCeed()
     {
+      CeedScalar *ptr;
+      CeedVectorTakeArray(vec_ceed, CEED_MEM_HOST, &ptr);
       CeedVectorDestroy(&vec_ceed);
     }
```

jeremylt:
Ok, with these two diffs, I can run locally with passing tests for all CPU backends:

Update top level makefile and test script:
https://github.com/CEED/libCEED/pull/1436#issuecomment-1912657741

Update deal.II Vector <-> CeedVector memory handoffs
https://github.com/CEED/libCEED/pull/1436#discussion_r1468144149

jeremylt:
Ok, last update for today, I promise. I fixed convergence for the HIP backends:

```diff
$ git diff
diff --git a/examples/deal.II/bps.h b/examples/deal.II/bps.h
index cfde7f7f..0d187e91 100644
--- a/examples/deal.II/bps.h
+++ b/examples/deal.II/bps.h
@@ -635,7 +640,11 @@ private:
     CeedElemRestrictionDestroy(&geo_restriction);
     CeedVectorDestroy(&node_coords);
     CeedElemRestrictionDestroy(&q_data_restriction);
-    CeedVectorDestroy(&q_data);
+    {
+      CeedScalar *ptr;
+      CeedVectorTakeArray(q_data, CEED_MEM_HOST, &ptr);
+      CeedVectorDestroy(&q_data);
+    }
     CeedBasisDestroy(&geo_basis);
 
     return weights;
```

We need this effect, but please tweak as needed to get more idiomatic code for deal.II

peterrum:
@jeremylt Thank you for all the debugging work! I thought I adopted the view part from the MFEM example. But maybe I just skipped some steps, since it was working for me locally.

I have pushed the suggestions. Could check if everything is in? I could not apply the diffs as patches so that I needed to make the modifications  manually.

jeremylt:
I've got fixes to get CI running here: https://github.com/CEED/libCEED/pull/1453

Feel free to cherry-pick over

peterrum:
@jeremylt Thank you! Shall I squash the commits? What is the policy?

jeremylt:
Depending upon which you prefer, we can squash+merge or you can squash/rebase as needed to a set of tidy atomic commits.

jeremylt (approved):

peterrum:
@jeremylt I have rebased and squashed and added! Thank you for all the help!

jeremylt:
Thanks for the work getting this together! If you are interested in follow-ups, I think two things I'd like to help add are

1) using deal.II FE data to build the libCEED basis object

2) handing off compatible device memory pointers to libCEED when deal.II vector uses device memory (I think I saw deal.II can compile with CUDA, KOKKOS, and maybe some others?)

peterrum:
Sure!

> using deal.II FE data to build the libCEED basis object

Can you point to the right function I could use?

> handing off compatible device memory pointers to libCEED when deal.II vector uses device memory (I think I saw deal.II can compile with CUDA, KOKKOS, and maybe some others?)

That is a good idea! We have removed CUDA and replaced the relevant code parts with Kokkos. I will need some time to extend the code, since our CPU and Kokkos matrix-free operators are not the same and have a slightly different interface.

jrwrigh:
> > using deal.II FE data to build the libCEED basis object
> 
> Can you point to the right function I could use?

You'd use `CeedBasisCreate{H1,TensorH1,Hdiv,Hcurl}` to construct it, depending on the FE space definition. [These](https://github.com/CEED/libCEED/blob/8ae08b988ca140d07893cbfe633ed64adfe89353/examples/fluids/src/dm_utils.c#L263) [two](https://github.com/CEED/libCEED/blob/8ae08b988ca140d07893cbfe633ed64adfe89353/examples/fluids/src/dm_utils.c#L356) functions in the fluids example do a similar job, but going from PETSc definitions of the FE space to CeedBasis objects rather than from deal.II

jeremylt:
A note here - with the way the CI is set up for the deal.II example, CI will only run on the example for branches created in the libCEED repo, not any forks.

PR #1438, opened 2024-01-04
Add __restrict__ to resolve vectorization warnings with Clang 17

sebastiangrimberg:
This resolves a few failed vectorization warnings for the offset, oriented, and curl-oriented element restriction operation: `loop not vectorized: cannot identify array bounds [-Rpass-analysis]`.

Also renames `CeedElemRestrictionApplyStandard` to `CeedElemRestrictionApplyOffset` in the `ref` backend to match the GPU backends.


jedbrown (approved):
Thanks!

PR #1440, opened 2024-01-04
Fix CeedVectorAXPBY Implementations

zatkins-dev:
Right now, `CeedVectorAXPBY` is implemented incorrectly in host/ref code as:
$$\mathbf{y} = \mathbf{y} + a\mathbf{x} + b\mathbf{y}$$

This PR corrects the implementation to:
$$\mathbf{y} = a\mathbf{x} + b\mathbf{y}$$

### DIFF
--- a/backends/cuda-ref/ceed-cuda-ref-vector.c
+++ b/backends/cuda-ref/ceed-cuda-ref-vector.c
@@ -691,7 +691,7 @@ static int CeedVectorAXPY_Cuda(CeedVector y, CeedScalar alpha, CeedVector x) {
 // Compute y = alpha x + beta y on the host
 //------------------------------------------------------------------------------
 static int CeedHostAXPBY_Cuda(CeedScalar *y_array, CeedScalar alpha, CeedScalar beta, CeedScalar *x_array, CeedSize length) {
-  for (CeedSize i = 0; i < length; i++) y_array[i] += alpha * x_array[i] + beta * y_array[i];
+  for (CeedSize i = 0; i < length; i++) y_array[i] = alpha * x_array[i] + beta * y_array[i];
   return CEED_ERROR_SUCCESS;
 }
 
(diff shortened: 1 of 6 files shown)


backends/cuda-ref/ceed-cuda-ref-vector.c
```diff
@@ -509,7 +509,7 @@ static int CeedVectorNorm_Cuda(CeedVector vec, CeedNormType type, CeedScalar *no
           CeedCallCublas(ceed, cublasSnrm2(handle, (CeedInt)sub_length, (float *)d_array_start, 1, &sub_norm));
           norm_sum += sub_norm * sub_norm;
         }
-        *norm            = sqrt(norm_sum);
```
jeremylt, line 512: looks like the formatter really wants this here

tests/t125-vector.c
```diff
@@ -39,9 +39,9 @@ int main(int argc, char **argv) {
 
     CeedVectorGetArrayRead(y, CEED_MEM_HOST, &read_array);
     for (CeedInt i = 0; i < len; i++) {
-      if (fabs(read_array[i] - (10.0 + i) * 1.5) > 1e-14) {
```
jeremylt, line 42: Rust and Python need similar fixes

backends/cuda-ref/ceed-cuda-ref-vector.c
zatkins-dev: my (`clang-format-17`) formatter strongly disagrees

jeremylt: I personally disagree, but CI uses 16

zatkins-dev: I'll force it to cooperate

zatkins-dev: that's so weird, there's similar code nearby that formats correctly...

jeremylt: Yeah, every so often clang-format hits an edge case.

jeremylt (approved):

PR #1443, opened 2024-01-11
fluids: Change slip boundary condition formulation

jrwrigh:
Two major changes:
- `bc_slip_i` $\longrightarrow$ `bc_symmetry_i` for `i` $\in$ `{x,y,z}`
- Implement new weak slip boundary conditions as `bc_slip`


examples/fluids/blasius.yaml
```diff
@@ -30,17 +30,17 @@ dm_plex_box_lower: 0,0,0
...
 gravity: 0,0,0
 
-stg:
-  use: false
-  inflow_path: "./STGInflow_blasius.dat"
-  mean_only: true
+# stg:
+#   use: false
+#   inflow_path: "./STGInflow_blasius.dat"
+#   mean_only: true
```
jeremylt, line 43: accidental dead code?

jrwrigh: It's required to pass tests. I have the code set to warn if there are unused options in test mode (from being bit by that before). Since `stg_use == false`, then the other two options aren't read, so it causes the test to fail on non-empty stdout.

jeremylt: Ah. Dead code is usually a 'no-no' in the repo

jrwrigh: We comment out example options in a few yaml files, mostly for code as a "demonstration" of what can be done, but commented out because it imposes extra library requirements (CGNS for the `ts_monitor_solution` viewer) or aren't really minimum working examples.

Those options don't get tested though, so I'm ok to remove them. I'm curious what @jedbrown thought it, as I think he was the first to put them in.

KennethEJansen:
I ran the bump case with Q2 elements with this merged into my code base. Results after 10,000 times steps look quite good so I think we can be confident that the bc-slip  is working properly. Before finalizing this as ` approved` should there be any discussion of whether it was necessary to rename the BC functions. I assume there was a good reason but we do need to consider the fact that this does make merges far far more difficult when functions change names. @jedbrown do you have thoughts on this?

jrwrigh:
Did CUDA get updated on Noether since the last time this branch ran tests? I'm not sure where these errors are coming from:
```
[0]PETSC ERROR: Error in external library
[0]PETSC ERROR: /var/lib/gitlab-runner/builds/N8GCsqus/2/libceed/libCEED/backends/cuda/ceed-cuda-compile.cpp:97 in CeedCompile_Cuda(): NVRTC_ERROR_COMPILATION
default_program(3139): error: expected an expression

default_program(3140): error: expected an expression
...
```

Looks like there are some errors with MFEM and ROCM too.

KennethEJansen:
> It should still work to use `bc_slip_x` (as a deprecated name for `bc_symmetry_x`) so I don't anticipate problems merging or porting job specifications. They have different implications for linear algebra so I think we want to support both as distinct BCs (though you're welcome to use `bc_slip` everywhere).

I think I was unclear. I am fully on board with naming the bases as they are named. My concern is that prior to this PR we had files named freestream_bc and those files have been re-factored to bc_freestream for example. In terms of git this looks like a file was removed in another file was added and we don't have linkage between them. When source files change names like this it does introduce some challenge to follow the changes that are present in the files.

jedbrown:
Git doesn't care about the file name (quite literally actually) when tracking changes. This is what `git log --stat` shows on this branch and `git blame` also reports the changes prior to renaming.

```
 examples/fluids/README.md                                                 |   4 ++
 examples/fluids/navierstokes.h                                            |   9 +--
 examples/fluids/problems/{freestream_bc.c => bc_freestream.c}             |   2 +-
 examples/fluids/problems/bc_slip.c                                        |  42 +++++++++++++
 examples/fluids/problems/gaussianwave.c                                   |   2 +-
 examples/fluids/problems/newtonian.c                                      |   7 ++-
 examples/fluids/qfunctions/{freestream_bc.h => bc_freestream.h}           | 325 +---------------------------------------------------------------------------------------------------
 examples/fluids/qfunctions/{freestream_bc_type.h => bc_freestream_type.h} |   0
 examples/fluids/qfunctions/bc_slip.h                                      | 107 +++++++++++++++++++++++++++++++++
 examples/fluids/qfunctions/riemann_solver.h                               | 353 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 examples/fluids/src/cloptions.c                                           |   3 +
 examples/fluids/src/setuplibceed.c                                        |   7 +++
 12 files changed, 530 insertions(+), 331 deletions(-)
```

KennethEJansen:
It sounds like the challenges are on my side and my choice of using `meld` as a way to see the changes and evolution of a file. My git difftool invokes meld and meld treats these as a completely separate files (shows the other as an empty file)and  doesn't understand that there was a name change and thus creates the challenge I speak of. Now of course I can check out both branches and use meld directed at the two files to see the change but that's just more tedious. I am probably a little bit more cranky about this since everything I'm doing coding wise is one-handed pecking.

jedbrown:
Ah, possibly James or I can help with the merge you need.

jedbrown:
@jrwrigh We haven't changed the CUDA install and other pipelines are not encountering this issue. Can you reproduce on Noether?

KennethEJansen:
On the Aurora SDK strangeness I will try to rebuild from scratch a 1215 SDK to see if that performs better what I had used which was an existing arch and I just rebuilt the source I did not rebuild all of its dependencies. It seems unlikely to me that the dependencies would cause a divergence in the first step but I suppose that's possible still as you can imagine my reluctance to rebuild from scratch stems from the fact that that's still a multi hour build for PETSc.

jrwrigh:
@KennethEJansen Have your recent runs (and successes) used this branch?

jeremylt:
The issue on Cuda persists. This usually indicates that there is something in the QFunction source that isn't able to be parsed by Cuda.

jrwrigh:
Gotcha. Any tips for seeing what code it's compiling? Otherwise I'm pretty much just down to scanning the new code and seeing if I missed anything.

jeremylt:
I like to use `CEED_DEBUG=1 ./build/fluids-navierstokes ...... > barf.out`. That will dump the concatenated QF source that is being JiTed to the `barf.out`

jrwrigh:
TIL that CUDA doesn't allow for structs to be initialized with their member names. ie
```c
StateConservative star = {.density = 0, .momentum = {1,1,1}, .E_total = 1};
```
isn't allowed. Annoying.

---
@jeremylt Other thing I did which was quite useful was to add a `printf` statement to the error checking if statement in `CeedCompile_Cuda` to print out the code buffer when it errors. IDK if that be a more useful permanent addition, with some kind of flag with it obviously (maybe included in `CEED_DEBUG`, but with a special header message to distinguish from the normal JIT messages?).

PR #1446, opened 2024-01-18
fluids: Misc refactor/updates for advection and stabilization

jrwrigh:
Various small changes to the fluids advection example and stabilization
- Refactor diagonal $\tau$ calculation to use the `Mat` and `Dot` utility functions
  - This was primarily to make sure my own understanding and future implementation of stabilization parameters was correct
- Correct documentation on $\tau$ formulation
- Increase the threshold for the skew advection problem initial condition
  - Had problems with 3rd order not quite getting there
- Convert the Advection problem to use the newtonian State structs
  - Not a perfect match, but helps utilize some of the gradient utilities already in use.
  - Also updated to use `Mat` and `Dot` utilities when appropriate.

### DIFF
--- a/examples/fluids/index.md
+++ b/examples/fluids/index.md
@@ -271,11 +271,11 @@ $$
 
 $$
 \mathcal{F} = \sqrt{ \rho^2 \left [ \left(\frac{2C_t}{\Delta t}\right)^2
-+ \bm u \cdot (\bm u \cdot  \bm g)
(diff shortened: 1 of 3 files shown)


jrwrigh:
Not sure what the Rust test failure is about, but everything else passes.

jeremylt (approved):
This all seems reasonable to me

PR #1447, opened 2024-01-20
fluids: Unify 2D and 3D Advection problems

jrwrigh:
Make the 2D and 3D advection problems share the same general code and functionality.

Depends on #1446


jeremylt (approved):
seems reasonable to me

jrwrigh:
Thanks!

PR #1449, opened 2024-01-23
Improve vectorization for CPU curl-oriented element restriction transpose

sebastiangrimberg:
This is a minor change which produces some modest speedups for low-order operators in particular.


jedbrown (approved):
LGTM

PR #1450, opened 2024-01-23
Refactoring to speed up GPU kernel JIT compilation

sebastiangrimberg:
This PR reduces the overhead of JIT during construction of several operators in an application code by ~40%. In profiling with NVIDIA Nsight I noticed that a lot of time was spent in `libnvrtc` and so tried to speed things up by avoiding the compilation of unused kernels. This includes:

- In element restriction: Only compile the kernels used for the particular element restriction type. Also, lazily compile them so we do not do any RTC when constructing a strided element restriction which will just be destroyed without being used (happens sometimes in QFunction assembly, I think).
- In diagonal assembly, compile only the diagonal or pointwise-diagonal assembly kernel, only as needed.
- In the MAGMA backend, compile the non-tensor element weight kernel as part of the interpolation/derivature module to a void needing to store a separate module.
- For all `include/ceed/jit-source` files, I removed the unneeded `#include <ceed.h>` directive. This is taken care of by the addition of the `cuda/hip/sycl-jit.h` header added to the source before compilation.


sebastiangrimberg:
Resolved the issues with HIP (I don't have a way to test this locally), so now marking as ready for review.

include/ceed/jit-source/cuda/cuda-ref-basis-nontensor-templates.h
```diff
@@ -10,8 +10,6 @@
 #ifndef CEED_CUDA_REF_BASIS_NONTENSOR_TEMPLATES_H
 #define CEED_CUDA_REF_BASIS_NONTENSOR_TEMPLATES_H
 
-#include <ceed.h>
```
jeremylt, line 13: We have historically included `#include <ceed.h>` in these files so IDEs can more easily understand types in the file.

include/ceed/jit-source/cuda/cuda-gen-templates.h
```diff
@@ -10,8 +10,6 @@
 #ifndef CEED_CUDA_GEN_TEMPLATES_H
 #define CEED_CUDA_GEN_TEMPLATES_H
 
-#include <ceed/types.h>
```
jeremylt, line 13: This should have been `#include <ceed.h>`

include/ceed/jit-source/cuda/cuda-ref-basis-nontensor-templates.h
jeremylt: Hmm, let me double check but it should be not actually part of the JiTed string

sebastiangrimberg: Oh, maybe I did not look closely enough. From [ceed-jit-tools.c](https://github.com/CEED/libCEED/blob/main/interface/ceed-jit-tools.c#L131), maybe system headers which are incuded with `<*>` are ignored, unless they match one of the four given specific ones?

jeremylt:
```

      bool  is_ceed_header    = is_hash_include && next_left_chevron && (next_new_line - next_left_chevron > 0) &&
                            (!strncmp(next_left_chevron, "<ceed/jit-source/", 17) || !strncmp(next_left_chevron, "<ceed/types.h>", 14) ||
                             !strncmp(next_left_chevron, "<ceed/ceed-f32.h>", 17) || !strncmp(next_left_chevron, "<ceed/ceed-f64.h>", 17));
```

Yeah, it should only be these 4 headers that are dropping in for JiT, so including `#include <ceed.h>` should result in the line being omitted from the source string for JiT.

sebastiangrimberg: Cool, reverted in [d69fde3](https://github.com/CEED/libCEED/pull/1450/commits/d69fde354f92f0f87cc1cc99880249b8250190ee) and addressed [this comment](https://github.com/CEED/libCEED/pull/1450#discussion_r1463925390) in [0854542](https://github.com/CEED/libCEED/pull/1450/commits/085454240ef0787e2bf0778613c6c2e31bdf9a43).

jeremylt:
That's for tidying this. I'm curious to see how long the GPU CI runs take now

jeremylt (approved):

jedbrown:
Indeed, this is a nice improvement.

nother-rocm: 27:46 -> 21:30
noether-cuda: 9:17 -> 7:20
nother-float: 7:19 -> 6:02

jedbrown (approved):

jedbrown:
@sebastiangrimberg Do you prefer squash-merge or rebasing to clean up your branch prior to a regular merge?

sebastiangrimberg:
> @sebastiangrimberg Do you prefer squash-merge or rebasing to clean up your branch prior to a regular merge?

I can clean up the history and force push!

jeremylt:
Ratel GPU CI down from 18:30 to 13:30!

PR #1451, opened 2024-01-23
rust - drop failing tests

jeremylt:
This drops the failing tests while we figure out how to fix this functionality


jedbrown (approved):
Could just comment (or change the conditional to false) to save time making it a real test in the future.

jeremylt:
There's a certain way you have to comment out  otherwise it becomes code that's hidden in documentation but still part of the test. I don't recall the details

PR #1453, opened 2024-01-29
deal.II Updates

jeremylt:
Some small deal.II fixes and turning on CI


jeremylt:
Ok, this should be passing CI now.

Of note, commit https://github.com/CEED/libCEED/pull/1453/commits/19dbfea1eb80e8e3f6567fe16ac54b08d40d4c7a fixes the memory synchronization for GPUs, but https://github.com/CEED/libCEED/pull/1453/commits/0209b97c42c0f14aef8964975dd858c1e2ed4fc5 adds an alternative approach that's more explicit and might be preferred.

jeremylt:
Hmm, digging into this error on Noether:

```
build/dealii-bps: symbol lookup error: /lib/x86_64-linux-gnu/libmkl_gf_lp64.so: undefined symbol: mkl_lapack_zhesvxx
```

I at no point in the build process requested deal.II use MKL, so I have no idea why its looking for it.

PR #1457, opened 2024-01-30
Replace calls to deprecated `local_accessor::get_pointer()`

uumesh:
`local_accessor::get_pointer()` has been deprecated in the SYCL standard. Replacing it with `get_multi_ptr()`


jedbrown:
It looks like `get_pointer` was deprecated in SYCL 2020 so there is no compatibility issue (we don't support pre-2020).
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html

jedbrown (approved):

jrwrigh:
@uumesh Does this pass tests for you on Aurora? If so, what compiler environment?

uumesh:
Just ran all the `t3` unit tests on Aurora with `/gpu/sycl/ref` backend which would be the ones affected by these changes and they all pass. 

I used the latest oneapi/eng-compiler module.

Issue #1460, opened 2024-02-02
memcheck and CEED_STRIDES_BACKEND


jedbrown:
From @knepley 

>  If you call `CeedElemRestrictionCreateStrided()` with `CEED_STRIDES_BACKEND`, and you stick in values on the CPU, you can get code that works fine with the CPU device and fails with the GPU device, and it is very hard to see why that is happening.

A couple ideas to catch this sooner:
* We could make `memcheck` use a weird permutation for its backend so you can at least detect this on the host.
* We could forbid access to vectors associated with `CEED_STRIDES_BACKEND`. I think this would require extra state. It should still be possible to compute permutation-invariant things like norms.

jeremylt:
Really, the issue is that `CEED_STRIDES_BACKEND` was specifically intended to only be valid when you keep using the same backend. Everything works fine if you use the exact same backend throughout. I thought the documentation called this out.

```
                          @ref CEED_STRIDES_BACKEND may be used for `CeedVector` ordered by the same `Ceed` backend.
```

That isn't the best explanation though. We should make it clearer (though people don't always read the documentation carefully to catch all implications of what's in there).

Technically, the L-vector ordering is supposed to be consistent across backends when using `CEED_STRIDES_BACKEND`, so it just matters when you try to get at the raw E-vector ordering, which isn't something users are supposed to be doing.

Making the `memcheck` backend use a bizarre permutation would also mean we need a brand new `memcheck` `CeedOperator` implementation, so I'd recommend against that.

knepley:
```
Technically, the L-vector ordering is supposed to be consistent across backends when using CEED_STRIDES_BACKEND, so it just matters when you try to get at the raw E-vector ordering, which isn't something users are supposed to be doing.
```
What I did is pull out the array as `CEED_MEMORY_HOST` and fill it up (with geometric data), and then use that as a kernel input. On the CPU it worked fine, but on the GPU it was reordered. When I used explicit strides, everything worked normally.

jeremylt:
Woops, misremembered and mispoke above. The strides specify the L-vector ordering, so `CEED_STRIDES_BACKEND` says you are using the specific ordering only designed to work for the backend that created the `CeedElemRestriction`. So the only valid usage of `CEED_STRIDES_BACKEND` is for a `CeedOperator` filling an output vector which is an input vector for a different `CeedOperator` from the exact same backend.

[...]

PR #1463, opened 2024-02-02
CEED_STRIDES_BACKEND related updates

jeremylt:
I have ~~two~~ four changes here 

1) Expanded note in the documentation about `CEED_STRIDES_BACKEND`

2) Set `CEED_STRIDES_BACKEND` to be different for memcheck than the default layouts for both CPU and GPU

3) Add `CeedElemRestrictionGetLLayout` for strided restrictions

4) Fix `CeedElemRestrictionGet[L,E]Layout` signature, previously there ways an unneeded pointer

This won't provide a clear warning message, but it will at least demonstrate bugs if a user mixes and matches memcheck `CEED_STRIDES_BACKEND` with other backends.

I don't think we can be really any clearer with the name that `CEED_STRIDES_BACKEND` is specific to the backend?

Offering `CeedElemRestrictionGetLLayout` probably fixes the root cause of why someone would see this problem though.

See #1460 

Note: if we like this, a666b53816e11fb415b90bd360ae18427fea068c, 0e6bd4c965abc248c0e8f58a997dfe99af63f364, and 32058e1b79154fab37205790e9d03bfd80da3adb should be squashed together

Also note: if we like `CeedElemRestrictionGetLLayout`, I'll go add some quick unit test

### DIFF
--- a/backends/memcheck/ceed-memcheck-restriction.c
+++ b/backends/memcheck/ceed-memcheck-restriction.c
@@ -655,16 +655,29 @@ int CeedElemRestrictionCreate_Memcheck(CeedMemType mem_type, CeedCopyMode copy_m
   CeedCallBackend(CeedElemRestrictionGetBlockSize(rstr, &block_size));
   CeedCallBackend(CeedElemRestrictionGetNumComponents(rstr, &num_comp));
   CeedCallBackend(CeedElemRestrictionGetCompStride(rstr, &comp_stride));
-  CeedInt layout[3] = {1, elem_size, elem_size * num_comp};
+  CeedCallBackend(CeedElemRestrictionGetType(rstr, &rstr_type));
 
   CeedCheck(mem_type == CEED_MEM_HOST, ceed, CEED_ERROR_BACKEND, "Only MemType = HOST supported");
 
   CeedCallBackend(CeedCalloc(1, &impl));
   CeedCallBackend(CeedElemRestrictionSetData(rstr, impl));
-  CeedCallBackend(CeedElemRestrictionSetELayout(rstr, layout));
+
+  // Set layouts
+  {
+    bool    has_backend_strides;
+    CeedInt e_layout[3] = {1, elem_size, elem_size * num_comp}, l_layout[3] = {0};
+
+    CeedCallBackend(CeedElemRestrictionSetELayout(rstr, e_layout));
+    if (rstr_type == CEED_RESTRICTION_STRIDED) {
+      CeedCallBackend(CeedElemRestrictionHasBackendStrides(rstr, &has_backend_strides));
+      if (has_backend_strides) {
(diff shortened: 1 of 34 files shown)


backends/memcheck/ceed-memcheck-restriction.c
```diff
@@ -0,0 +1,801 @@
...
+      // CPU default
+      strides[0] = 1;
+      strides[1] = num_elem * elem_size;
+      strides[2] = elem_size;
+      // GPU default
+      strides[0] = 1;
+      strides[1] = elem_size;
+      strides[2] = elem_size * num_comp;
+     **/
+  }
```
jeremylt, line 56: 4 different orderings is probably overkill, but I was trying to avoid some accidental correspondence between vectors that masks a bug. This here should be the 6 possible dense orderings, so I avoided the CPU and GPU default layouts for l-vectors.

backends/memcheck/ceed-memcheck-serial.c
```diff
@@ -11,22 +11,42 @@
...
+  CeedCallBackend(CeedSetData(ceed, impl));
+
   // Create reference Ceed that implementation will be dispatched through unless overridden
   CeedCallBackend(CeedInit("/cpu/self/ref/serial", &ceed_ref));
   CeedCallBackend(CeedSetDelegate(ceed, ceed_ref));
 
   CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "VectorCreate", CeedVectorCreate_Memcheck));
+  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "ElemRestrictionCreate", CeedElemRestrictionCreate_Memcheck));
+  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "ElemRestrictionCreateBlocked", CeedElemRestrictionCreate_Memcheck));
+  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "ElemRestrictionCreateAtPoints", CeedElemRestrictionCreate_Memcheck));
```
jeremylt, line 46: Unfortunately, the way we split restrictions means we have only these 3 backend constructor functions as options. Maybe we should be splitting strided, standard, oriented, and curl oriented constructors. Keep the same signature for each though so the backends can recycle the same core function as they currently do?

jedbrown (approved):
LGTM at a high level. Thanks for solving this problem. I take it we aren't enforcing anything new outside the memcheck backend?

jeremylt:
> I take it we aren't enforcing anything new outside the memcheck backend?

The only thing that is disruptive is https://github.com/CEED/libCEED/pull/1463/commits/f447f4b392b4a1c8e4aedfe1f2d1777cef0ee24a - really this function signature should have never had a pointer to an array

PR #1464, opened 2024-02-06
xsmm - update for function name change

### LINKED ISSUE #1462: LIBXSMM interface broken with `main`
After https://github.com/libxsmm/libxsmm/pull/850, the `xsmm` backend fails to compile. This is probably indicates LIBXSMM is about to release v2, after which we should make the change in libCEED for the minimum required version.

jeremylt:
This branch fixes #1462 

It turns out it was just a name change. We can merge once libXSMM releases their new version, or sooner if needed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -72,8 +72,8 @@ noether-cpu:
     - echo "-------------- FC ------------------" && $FC --version
     - echo "-------------- GCOV ----------------" && gcov --version
 # Libraries for backends
-# -- LIBXSMM 2c145a109b5a8ad4e15f60ea42a86b9056bdc8b8
-    - cd .. && export XSMM_HASH=2c145a109b5a8ad4e15f60ea42a86b9056bdc8b8 && { [[ -d libxsmm-$XSMM_HASH ]] || { curl -L https://github.com/libxsmm/libxsmm/archive/$XSMM_HASH.tar.gz -o xsmm.tar.gz && tar zvxf xsmm.tar.gz && rm xsmm.tar.gz && make -C libxsmm-$XSMM_HASH -j$(nproc); }; } && export XSMM_DIR=$PWD/libxsmm-$XSMM_HASH && cd libCEED
+# -- LIBXSMM 7 April 2024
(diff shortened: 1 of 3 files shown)


jedbrown (commented):
Is there a way to identify the version? It seems like only the patch version has changed, like libxsmm `main` is now at `1.17-3708`, and I assume `1.18` is next? Is there a better way?

jeremylt:
I think I contributed a macro to them like ours. I think we need to ask them if a release is coming soon and then we can add a macro to that tensor file that uses their version marco

jeremylt:
libXSMM should be making a new release in roughly 2-3 months time

backends/xsmm/ceed-xsmm.h
```diff
@@ -11,6 +11,13 @@
 #include <ceed.h>
 #include <ceed/backend.h>
 
+#if ((LIBXSMM_VERSION_MAJOR > 1 || \
+      (LIBXSMM_VERSION_MAJOR == 1 && (LIBXSMM_VERSION_MINOR > 17 || (LIBXSMM_VERSION_MINOR == 17 && LIBXSMM_VERSION_PATCH >= 3710)))))
+#define libxsmm_dispatch_gemm(a, b, c) libxsmm_dispatch_gemm(a, b, c)
+#else
+#define libxsmm_dispatch_gemm(a, b, c) libxsmm_dispatch_gemm_v2(a, b, c)
+#endif
+
```
jeremylt, line 20: Fun fact, the LIBXSMM main goes from patch 3710 to 3708 to 3710 to 3708 to 3715 to 3716 to 3710 to 3711 to 3719 to 3715 to 3718... so we can't really rely upon the patch level to be sequential

sebastiangrimberg: 🙄

jeremylt:
The LIBXSMM release is taking a while. What about merging this fix now since we're already requiring a min version that isn't the latest release. Then we can switch the min version to the next LIBXSMM release when it happens

sebastiangrimberg (approved):

jedbrown (approved):
Okay, but we should document the min version somewhere more than CI, as this bleeding edge non-release version may be surprising to users.

PR #1466, opened 2024-02-08
Don't error CeedOperatorGetFieldByName when field not found

jeremylt:
This brings us in line with the behavior for `CeedOperatorGetContextFieldLabel`

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -897,7 +897,9 @@ int CeedOperatorAtPointsGetPoints(CeedOperator op, CeedElemRestriction *rstr_poi
 }
 
 /**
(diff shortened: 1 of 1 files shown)


interface/ceed-operator.c
```diff
@@ -929,12 +932,6 @@ int CeedOperatorGetFieldByName(CeedOperator op, const char *field_name, CeedOper
       return CEED_ERROR_SUCCESS;
     }
   }
-  // LCOV_EXCL_START
```
zatkins-dev, line 932: We still need a `return CEED_ERROR_SUCCESS;` at the end here, right?

jeremylt: oh, right

zatkins-dev (approved):
Looks good to me!

PR #1471, opened 2024-02-13
Skip tet tests without tetget installed

jeremylt:
This gives less confusing output for `make prove-all` when PETSc is present but wasn't build with tetgen or ctetgen


zatkins-dev (approved):
Looks good to me!

jrwrigh (approved):
Thanks!

PR #1472, opened 2024-02-13
junit - allow looser CGNS tol

jeremylt:
@zatkins-dev feel free to adjust as desired to better fit, but we needed this in Ratel

### DIFF
--- a/tests/junit_common.py
+++ b/tests/junit_common.py
@@ -97,13 +97,14 @@ class SuiteSpec(ABC):
         """
         raise NotImplementedError
 
-    def get_cgns_tol(self) -> float:
-        """retrieve CGNS test tolerance.
+    @property
+    def cgns_tol(self):
+        """Absolute tolerance for CGNS diff"""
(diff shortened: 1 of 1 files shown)


tests/junit_common.py
```diff
@@ -97,6 +97,14 @@ def get_output_path(self, test: str, output_file: str) -> Path:
         """
         raise NotImplementedError
 
+    def get_cgns_tol(self) -> float:
+        """retrieve CGNS test tolerance.
+
+        Returns:
+            tolerance (float): Test tolerance
+        """
+        return self.cgns_tol if hasattr(self, 'cgns_tol') else 1.0e-12
```
zatkins-dev, line 106: This pattern is a code smell -- we should either add the attribute as a property to the base class or make this function return a constant and override this function in the Ratel subclass to return the attribute. 

I am good with either, whatever you think is more reasonable.

zatkins-dev: Using a property is probably the more pythonic approach

zatkins-dev (approved):
Looks good to me, thanks!

PR #1473, opened 2024-02-14
fluids: Shakib tau for advection

jrwrigh:
Add in the Shakib definition of tau to the advection example:

```math
\tau = \left [ \left(\frac{2 C_t}{\Delta t}\right)^2
+ \frac{\mathbf{ u} \cdot (\mathbf u \cdot  \mathbf g)}{C_a}
+ \frac{\kappa^2 \Vert \mathbf g \Vert_F ^2}{C_d} \right]^{-1/2}
```

for $\mathbf g$ the metric tensor. This is very similar in form to the Shakib $\tau$ formulation for the Navier-Stokes equations.
Note that diffusion isn't yet implemented in advection, but I've left the documentation there as I will get to that soon.

$C_a = P^2$ by default to roughly account for the order of the element (whether or not that works well is apart of my research, but it seems like a sane default).

Depends on #1447


jrwrigh:
Tests pass moving to `-ksp_type gmres` and a more physically realistic `-Ctau_a`.

jedbrown (approved):
This looks good at a high level. What do we have in the way of test problems to assess which is better when?

jrwrigh:
Periodic sine wave advection is probably the best way. I currently have a separate branch for that right now.

PR #1476, opened 2024-02-20
Fallback from GPU gen to shared for non-tensor

jeremylt:
Minor PR to automatically fall back to shared backends when you find an operator with non-tensor bases for GPU

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator.c
+++ b/backends/cuda-gen/ceed-cuda-gen-operator.c
@@ -117,11 +117,11 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec, 
     bool has_tensor_bases;
 
     CeedCallBackend(CeedOperatorHasTensorBases(op, &has_tensor_bases));
-    // -- Fallback to shared if not all bases are tensor-product
+    // -- Fallback to ref if not all bases are tensor-product
     if (!has_tensor_bases) {
       CeedOperator op_fallback;
 
-      CeedDebug256(ceed, CEED_DEBUG_COLOR_SUCCESS, "Falling back to /gpu/cuda/shared due to non-tensor bases in CeedOperator");
+      CeedDebug256(ceed, CEED_DEBUG_COLOR_SUCCESS, "Falling back to /gpu/cuda/ref CeedOperator due to non-tensor bases");
       CeedCallBackend(CeedOperatorGetFallback(op, &op_fallback));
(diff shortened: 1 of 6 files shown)


jeremylt:
When this is approved, I'll make a matching PR to do the same to SYCL

nbeams:
`*-shared` doesn't have any non-tensor kernels, only `-*ref` does (and magma). So if you create a `shared` fallback, it will actually be calling `ref` for everything through delegation. I think the fallback (and message) might as well be for `ref`, to avoid confusion.

jeremylt:
Good point, I'll update that since its really the same core operator handling between the two

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -112,6 +112,22 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
   CeedCallBackend(CeedOperatorGetFields(op, &num_input_fields, &op_input_fields, &num_output_fields, &op_output_fields));
   CeedCallBackend(CeedQFunctionGetFields(qf, NULL, &qf_input_fields, NULL, &qf_output_fields));
 
+  // Check for tensor-product bases
+  {
+    bool has_tensor_bases;
+
+    CeedCallBackend(CeedOperatorHasTensorBases(op, &has_tensor_bases));
+    // -- Fallback to shared if not all bases are tensor-product
```
nbeams, line 120: Should probably update the comment too, to avoid confusing future selves :)

backends/hip-gen/ceed-hip-gen-operator.c
```diff
@@ -48,6 +48,22 @@ static int CeedOperatorApplyAdd_Hip_gen(CeedOperator op, CeedVector input_vec, C
   CeedCallBackend(CeedOperatorGetFields(op, &num_input_fields, &op_input_fields, &num_output_fields, &op_output_fields));
   CeedCallBackend(CeedQFunctionGetFields(qf, NULL, &qf_input_fields, NULL, &qf_output_fields));
 
+  // Check for tensor-product bases
+  {
+    bool has_tensor_bases;
+
+    CeedCallBackend(CeedOperatorHasTensorBases(op, &has_tensor_bases));
+    // -- Fallback to shared if not all bases are tensor-product
```
nbeams, line 56: same

nbeams (approved):

backends/cuda-gen/ceed-cuda-gen-operator.c
jeremylt: thanks

PR #1477, opened 2024-02-20
Fallback to sycl/ref from gen for non-tensor bases

jeremylt:
Mirror #1476 for SYCL. Copy-pasted from the GPU code that works in CI, so should be fine for Sycl.


jeremylt:
SYCL CI build passes

jrwrigh (approved):
LGTM here too. I don't think we've tested with non-tensor product elements with SYCL anyways.

PR #1479, opened 2024-02-21
fluids: Report both amat and pmat types

jrwrigh:
When giving `-amat_type shell`, the `DM MatType` would not report `shell` but whatever the default DM MatType was. This now reports the A MatType and P MatType separately.


jedbrown (approved):
Thanks!

PR #1484, opened 2024-02-22
Should error when trying to use QF flops, not retrieve

jeremylt:
This should make more sense


zatkins-dev (approved):
This makes much more sense

PR #1487, opened 2024-02-23
Add const to GetName functions

### LINKED ISSUE #1486: QFunction/OperatorFieldGetName
Why is `QFunctionFieldGetName` and `OperatorFieldGetName` using `char **name` instead of `const char **name`? Compare with `CeedGetResource`. I think we should swap these signatures over (and then fix the wrappers)

jeremylt:
Fixes #1486 

These functions shouldn't have allowed this mutation at any point.

### DIFF
--- a/backends/memcheck/ceed-memcheck-qfunction.c
+++ b/backends/memcheck/ceed-memcheck-qfunction.c
@@ -57,7 +57,8 @@ static int CeedQFunctionApply_Memcheck(CeedQFunction qf, CeedInt Q, CeedVector *
   }
   // Check for unset output values
   {
-    char *kernel_path, *kernel_name;
+    const char *kernel_name;
+    char       *kernel_path;
 
     CeedCallBackend(CeedQFunctionGetSourcePath(qf, &kernel_path));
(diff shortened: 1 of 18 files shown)


jrwrigh (approved):
TIL a reason to have a `const` in a struct.

LGTM!

backends/memcheck/ceed-memcheck-qfunction.c
```diff
@@ -57,7 +57,8 @@ static int CeedQFunctionApply_Memcheck(CeedQFunction qf, CeedInt Q, CeedVector *
   }
   // Check for unset output values
   {
-    char *kernel_path, *kernel_name;
+    const char *kernel_name;
+    char       *kernel_path;
 
     CeedCallBackend(CeedQFunctionGetSourcePath(qf, &kernel_path));
```
jedbrown, line 63: Should this be mutable?

jeremylt: Aww crap, I missed one. Thanks. I'll spin up a follow-up PR

PR #1493, opened 2024-02-29
Add `CeedOperatorLinearAssembleQFunction` to At Points operators with Ref backend

zatkins-dev:
Adds QFunction assembly for points operators.

The major difference is the the output restriction is an AtPoints restriction and the output assembled matrix is defined at points. Because of this difference, we will also need to add `CeedOperatorLinearAssembleSingleAtPoints`, either in interface code or in the Ref backend, in order to have functional full operator assembly.

However, given the complexities of that task, I think we should split it into a different PR.

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -884,6 +884,229 @@ static int CeedOperatorApplyAddAtPoints_Ref(CeedOperator op, CeedVector in_vec, 
 }
 
 //------------------------------------------------------------------------------
+// Core code for assembling linear QFunction
+//------------------------------------------------------------------------------
+static inline int CeedOperatorLinearAssembleQFunctionAtPointsCore_Ref(CeedOperator op, bool build_objects, CeedVector *assembled,
+                                                                      CeedElemRestriction *rstr, CeedRequest *request) {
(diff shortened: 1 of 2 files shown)


backends/ref/ceed-ref-operator.c
```diff
@@ -883,6 +883,229 @@ static int CeedOperatorApplyAddAtPoints_Ref(CeedOperator op, CeedVector in_vec,
   return CEED_ERROR_SUCCESS;
 }
 
+//------------------------------------------------------------------------------
+// Core code for assembling linear QFunction
+//------------------------------------------------------------------------------
+static inline int CeedOperatorLinearAssembleQFunctionAtPointsCore_Ref(CeedOperator op, bool build_objects, CeedVector *assembled,
```
jeremylt, line 889: We may end up always having to `build_objects`, but that's something we can change in the future if needed

jeremylt (approved):

backends/ref/ceed-ref-operator.c
zatkins-dev: I say leave the option for now, I think that we will be able to reuse the assembly objects at least between SNES iterations, since we only move the points after we converge (unless I'm forgetting something).

PR #1495, opened 2024-03-01
Diagonal Assembly Clarification for AtPoints

jeremylt:
I changed this MR to clean up some internal logic and then we can do the diagonal assembly later if it becomes important

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -187,7 +187,8 @@ static int CeedOperatorCreateFallback(CeedOperator op) {
 
   @ref Developer
 **/
-static inline int CeedSingleOperatorAssembleAddDiagonal_Core(CeedOperator op, CeedRequest *request, const bool is_point_block, CeedVector assembled) {
+static inline int CeedSingleOperatorLinearAssembleAddDiagonal(CeedOperator op, CeedRequest *request, const bool is_point_block,
+                                                              CeedVector assembled) {
   Ceed ceed;
   bool is_composite;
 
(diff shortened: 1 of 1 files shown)


jeremylt:
Note - CI failure is unrelated, this is fine to merge now

interface/ceed-preconditioning.c
```diff
@@ -187,7 +187,8 @@ static int CeedOperatorCreateFallback(CeedOperator op) {
 
   @ref Developer
 **/
-static inline int CeedSingleOperatorAssembleAddDiagonal_Core(CeedOperator op, CeedRequest *request, const bool is_point_block, CeedVector assembled) {
+static inline int CeedSingleOperatorLinearAssembleAddDiagonal_Mesh(CeedOperator op, CeedRequest *request, const bool is_point_block,
+                                                                   CeedVector assembled) {
```
jrwrigh, line 191: I don't like `Mesh` as a name for "not AtPoints" for some reason, but I also can't think of a better name, nor am I familiar with the relationship between AtPoints and the "normal" operators.

interface/ceed-preconditioning.c
```diff
@@ -361,6 +362,30 @@ static inline int CeedSingleOperatorAssembleAddDiagonal_Core(CeedOperator op, Ce
...
+  @ref Developer
+**/
+static inline int CeedSingleOperatorLinearAssembleAddDiagonal(CeedOperator op, CeedRequest *request, const bool is_point_block,
+                                                              CeedVector assembled) {
+  Ceed ceed;
+  bool is_at_points;
+
+  CeedCall(CeedOperatorGetCeed(op, &ceed));
+  CeedCall(CeedOperatorIsAtPoints(op, &is_at_points));
+  CeedCheck(!is_at_points, ceed, CEED_ERROR_UNSUPPORTED, "Composite operator not supported");
```
jrwrigh, line 384: Should this read:
```suggestion
  CeedCheck(!is_at_points, ceed, CEED_ERROR_UNSUPPORTED, "AtPoints operator not supported");
```
?

jrwrigh (approved):

jrwrigh:
I think you need to rebase onto main to get the DMPlexFilter corrections.

interface/ceed-preconditioning.c
jeremylt: Yeah, I don't love it, but I don't have a better name either

jrwrigh: We can always change it later if we think of something better.

PR #1496, opened 2024-03-03
Fluids: Fix mass operator for explicit timestepping, move to KSP for mass operator

jrwrigh:
There are two major changes in this PR:
1) Moves the mass operator for the explicit timestepping to a `KSP` object rather than a single vector. This allows for using the consistent mass matrix if desired or sticking with rowsum (which is the default, using `-ksp_type preonly -pc_jacobi_type rowsum`).

2) Corrects the previous mass matrix lumping technique by setting the local vector to zero and doing a `DMGlobalToLocal` of a (global) vector of 1s onto that local vector. The reason for this is detailed in the document below:
[Mass matrix with dirichlet BCs.pdf](https://github.com/CEED/libCEED/files/14471204/Mass.matrix.with.dirichlet.BCs.pdf)
In words, the essential nodes (strong BCs) should _always_ be zero for the matrix-vector product of the mass matrix. Instead, the effect of those essential boundary conditions should only be present on the residual vector (which for this timestepping use case is the output from the RHS residual operator). However, since we normally assume that $\dot U = 0$ for strongly-constrained nodes, no modification of the residual vector is needed. If at some point we want to break that assumption (ie. for STG), then we'd need to factor that into the residual vector formulation.

---
A few verifications on the performance aspect of moving to KSP, the rowsum vector stored indefnitely by the preconditioner is held until it is detected that the matrix has changed. For the shell mat that is used, the "detection" only occurs if `MatAssembly{Begin,End}` is run on that matrix. 

Convergence is not checked for `KSPPREONLY`. The residual of the KSP is only checked if `-mass_ksp_monitor` is enabled ([and thus `ksp->numbermonitors > 0`](https://gitlab.com/petsc/petsc/-/blob/main/src/ksp/ksp/impls/preonly/preonly.c?ref_type=heads#L37-51)). So in the default configuration, the only computationally taxing thing is the point-wise multiplication, just as it was before.


examples/fluids/navierstokes.c
```diff
@@ -197,14 +197,7 @@ int main(int argc, char **argv) {
...
 
-  // ---------------------------------------------------------------------------
-  // Set up lumped mass matrix
-  // ---------------------------------------------------------------------------
-  // -- Set up global mass vector
-  PetscCall(VecDuplicate(Q, &user->M_inv));
-
-  // -- Compute lumped mass matrix
-  PetscCall(ComputeLumpedMassMatrix(ceed, dm, ceed_data, user->M_inv));
+  PetscCall(CreateKspMassOperator(user, ceed_data));
```
jeremylt, line 200:
```suggestion
  PetscCall(CreateKSPMassOperator(user, ceed_data));
```
minor - PETSc capitalizes `KSP` in function names

jrwrigh: Ah, yep.

jeremylt (approved):
LGTM

jrwrigh:
Here's a more clear version of the above document. Slight change in organization and nomenclature. [Mass matrix with dirichlet BCs.pdf](https://github.com/CEED/libCEED/files/14487212/Mass.matrix.with.dirichlet.BCs.1.pdf)

PR #1500, opened 2024-03-05
fix(fluids): RHS stabilization term should be negative

jrwrigh:
Accidentally changed in #1447 here: https://github.com/CEED/libCEED/pull/1447/files?diff=unified&w=0#diff-bab1280d31cf5e98c99acf2a7c3e708571414a19ea4c67ccdc48eef6fd2334deL285-L286

This part of the code was not under a test, hence how this got through testing. I've now added a test for this case, which also tests the consistent mass matrix solve introduced in #1496 .


jrwrigh:
_Converting to draft as there is not a test for the explicit stabilized methods. I'll probably make that example also use the consistent mass matrix_

PR #1504, opened 2024-03-06
Write source to debug when CUDA/HIP fails to compile

jeremylt:
This change, suggested by @jrwrigh, dumps the last JiT source to stdout when CUDA/HIP fails to compile.

### DIFF
--- a/backends/cuda/ceed-cuda-compile.cpp
+++ b/backends/cuda/ceed-cuda-compile.cpp
@@ -73,8 +73,6 @@ int CeedCompile_Cuda(Ceed ceed, const char *source, CUmodule *module, const Ceed
   }
   code << jit_defs_source;
   code << "\n\n";
-  CeedCallBackend(CeedFree(&jit_defs_path));
-  CeedCallBackend(CeedFree(&jit_defs_source));
 
   // Non-macro options
   opts[0] = "-default-device";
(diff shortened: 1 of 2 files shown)


backends/cuda/ceed-cuda-compile.cpp
```diff
@@ -97,11 +95,19 @@ int CeedCompile_Cuda(Ceed ceed, const char *source, CUmodule *module, const Ceed
     char  *log;
     size_t log_size;
 
+    CeedDebug256(ceed, CEED_DEBUG_COLOR_ERROR, "---------- CEED JIT SOURCE FAILED TO COMPILE ----------\n");
+    CeedDebug(ceed, "File: %s\n", jit_defs_path);
+    CeedDebug(ceed, "Source:\n%s\n", jit_defs_source);
```
jrwrigh, line 100:
```suggestion
    CeedDebug(ceed, "Source:\n%s\n", code);
```

I actually printed out `code`, not `jit_defs_source`. Not sure how much of a difference that makes, but might as well have the full buffer just in case?

jeremylt: Ah, duh. Silly mistake on my part

jrwrigh (approved):
Thanks!

PR #1505, opened 2024-03-08
Add Ceed*ReturnCeed

jeremylt:
This trims out a fair amount of code and seems like a reasonable add

### DIFF
--- a/backends/cuda-ref/ceed-cuda-ref-vector.c
+++ b/backends/cuda-ref/ceed-cuda-ref-vector.c
@@ -74,7 +74,7 @@ static inline int CeedVectorSyncD2H_Cuda(const CeedVector vec) {
   CeedCallBackend(CeedVectorGetCeed(vec, &ceed));
   CeedCallBackend(CeedVectorGetData(vec, &impl));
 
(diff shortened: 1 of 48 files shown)


backends/cuda-ref/ceed-cuda-ref-vector.c
```diff
@@ -67,14 +67,14 @@ static inline int CeedVectorSyncH2D_Cuda(const CeedVector vec) {
...
-  Ceed             ceed;
   CeedSize         length;
+  Ceed             ceed;
   CeedVector_Cuda *impl;
 
   CeedCallBackend(CeedVectorGetCeed(vec, &ceed));
   CeedCallBackend(CeedVectorGetData(vec, &impl));
 
-  CeedCheck(impl->d_array, ceed, CEED_ERROR_BACKEND, "No valid device data to sync to host");
+  CeedCheck(impl->d_array, CeedVectorReturnCeed(vec), CEED_ERROR_BACKEND, "No valid device data to sync to host");
```
jrwrigh, line 77:
```suggestion
  CeedCheck(impl->d_array, ceed, CEED_ERROR_BACKEND, "No valid device data to sync to host");
```

IDC either way, but why change if we already need `ceed` for the rest of the

jrwrigh (approved):
LGTM

PR #1508, opened 2024-03-11
Initalize scratch vectors in AtPoints Operator impl

jeremylt:
@zatkins-dev haven't tested, but this might fix it.

closes #1507


zatkins-dev:
P.S. we should probably test with memcheck backends in one of the fast test suites

zatkins-dev:
I don't think that will fix it... the issue is here, where `array` is something like `&e_data[i][num_points_offset * size]`.
```c
  memcpy(impl->array_allocated, impl->array, length * sizeof(array[0]));
```
In any instance where the last cell has fewer than `length/num_comp` points, this `memcpy` will read past the edge of the array.

jeremylt:
Ohh, I see. My fix here is still probably important though

zatkins-dev:
You had the right fix in the wrong spot -- should be fixed now

jeremylt:
Cool, if it fixes your issue, feel free to merge when CI is done

PR #1510, opened 2024-03-11
fluids: Add MatCeed

jrwrigh:
~~Stealing~~ Borrowing Ratel's implementation of a `Mat` based on libCEED. Useful for handling COO assembly of the operator matrices.


examples/fluids/src/petsc_ops.c
```diff
@@ -499,4 +507,63 @@ PetscErrorCode CreateMatShell_Ceed(OperatorApplyContext ctx, Mat *mat) {
...
+ *
+ * If `Amat` uses `mat_ceed`, then `Pmat` is either assembled or uses `mat_ceed` based on the preconditioner choice in `ksp`.
+ *
+ * @param[in]  ksp      `KSP` object for used for solving
+ * @param[in]  mat_ceed `MATCEED` for the linear operator
+ * @param[out] Amat     `Mat` to be used for the solver `Amat`
+ * @param[out] Pmat     `Mat` to be used for the solver `Pmat`
+ * @param[in]  assemble Whether to assemble `Amat` and `Pmat` if they are not `mat_ceed`
+ */
+PetscErrorCode MatCeedReturnSolveOperators(KSP ksp, Mat mat_ceed, Mat *Amat, Mat *Pmat, PetscBool assemble) {
```
jeremylt, line 524: The word `Return` in the code would imply that the return value is these 'solve operators'. I think you want `Get` or `Create`. Also, usually all in parameters come before all out parameters.

I don't think this is core `MatCEED` functionality, so the name is a bit confusing. A `KSP` is higher in the PETSc object hierarchy, so a `MatCEED` function shouldn't rely upon a `KSP`. This would be more of an internal application function, to my eye.

examples/fluids/include/mat-ceed.h
```diff
@@ -0,0 +1,42 @@
...
+MATCEED_INTERN PetscErrorCode MatCeedGetInnerMatType(Mat mat, MatType *type);
+MATCEED_INTERN PetscErrorCode MatCeedSetOperation(Mat mat, MatOperation op, void (*g)(void));
+MATCEED_INTERN PetscErrorCode MatCeedSetLocalVectors(Mat mat, Vec X_loc, Vec Y_loc_transpose);
+MATCEED_INTERN PetscErrorCode MatCeedGetLocalVectors(Mat mat, Vec *X_loc, Vec *Y_loc_transpose);
+MATCEED_INTERN PetscErrorCode MatCeedRestoreLocalVectors(Mat mat, Vec *X_loc, Vec *Y_loc_transpose);
+MATCEED_INTERN PetscErrorCode MatCeedGetCeedOperators(Mat mat, CeedOperator *op_mult, CeedOperator *op_mult_transpose);
+MATCEED_INTERN PetscErrorCode MatCeedRestoreCeedOperators(Mat mat, CeedOperator *op_mult, CeedOperator *op_mult_transpose);
+MATCEED_INTERN PetscErrorCode MatCeedSetLogEvents(Mat mat, PetscLogEvent log_event_mult, PetscLogEvent log_event_mult_transpose);
+MATCEED_INTERN PetscErrorCode MatCeedGetLogEvents(Mat mat, PetscLogEvent *log_event_mult, PetscLogEvent *log_event_mult_transpose);
+MATCEED_INTERN PetscErrorCode MatCeedCreateMatForAssembly(Mat mat_ceed, Mat *mat_assemble);
```
jeremylt, line 40: I think this might actually be a duplication of the functionality for `MatConvert`? Not sure, I'm still learning what existing pieces of `Mat` infrastructure we can leverage

https://petsc.org/release/manualpages/Mat/MatConvert/

examples/fluids/src/petsc_ops.c
jrwrigh: Oh, agreed. I didn't think intend this to be a core `MatCEED` function. Though I agree, the name isn't very good.

examples/fluids/include/mat-ceed.h
jrwrigh: Maybe. What I'm really wanting to do is to get a `DMCreateMat` with the MatCEED's DM and using it's inner mat type. I guess I can get the same effect via `MatConvert` with `MatCeedGetInnerMatType`. I can try that out.

jeremylt: I think an odd piece of compatibility is that naturally a `MatCeed` has 2 `DM`s (for PMG transfers) while a `Mat` will only have 1 `DM`

jeremylt:
I merged some small changes to Ratel main

examples/fluids/src/mat-ceed.c
```diff
@@ -0,0 +1,1429 @@
...
+  @brief Setup `Mat` for assembly
+
+  Collective across MPI processes.
+
+  @param[in]   mat_ceed   `MATCEED` to setup
+  @param[out]  mat_assemble `Mat` for assembly
+
+  @return An error code: 0 - success, otherwise - failure
+**/
+PetscErrorCode MatCeedCreateMatForAssembly(Mat mat_ceed, Mat *mat_assemble) {
```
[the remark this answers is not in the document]
jrwrigh: Removed in favor of `MatCeedGetInnerMatType` + `MatConvert`

jrwrigh:
No idea what the "C/Fortran / test (ubuntu-22.04, clang)" test failure is about; It's segfaulting on library tests that I don't use.

jrwrigh:
Ok, I think this should be ready to review now.

examples/fluids/include/mat-ceed-impl.h
```diff
@@ -0,0 +1,69 @@
...
+#define MATCEED_INTERN
+#else
+#define MATCEED_INTERN MATCEED_EXTERN __attribute__((visibility("hidden")))
+#endif
+
+/**
+  @brief Calls a libCEED function and then checks the resulting error code.
+  If the error code is non-zero, then a PETSc error is set with the libCEED error message.
+**/
+#define PetscCeedCall(ceed_, ...)                                   \
```
jeremylt, line 26: Note - renamed `PetscCallCeed` in Ratel to match PETSc conventions

jrwrigh: Yep, I saw that. Figured I'd do that and the other changes from https://gitlab.com/micromorph/ratel/-/merge_requests/794 in a later PR. But changing this one isn't a big deal.

jeremylt: Cool. I figure there will be some minor effort keeping these two in sync, but nothing too bad

jrwrigh: Yep, that was my plan.

examples/fluids/Makefile
```diff
@@ -41,6 +41,7 @@ AFLAGS ?= -fsanitize=address #-fsanitize=undefined -fno-omit-frame-pointer
 CFLAGS += $(if $(ASAN),$(AFLAGS))
 FFLAGS += $(if $(ASAN),$(AFLAGS))
 LDFLAGS += $(if $(ASAN),$(AFLAGS))
+CPPFLAGS += -I./include
```
jeremylt, line 44: I'm pretty meh at Make, but I'm confused as to why this would be only for `CPPFLAGS` and not `CFLAGS`?

jrwrigh: That... a great question. TBH, I actually just [took the line from Ratel](https://gitlab.com/micromorph/ratel/-/blob/main/Makefile?ref_type=heads#L80).

jrwrigh: I guess the preprocessor is responsible for handling the include file substitutions, so it's the only one that needs to know about them.

examples/fluids/src/petsc_ops.c
```diff
@@ -425,78 +425,83 @@ PetscErrorCode ApplyAddCeedOperatorLocalToLocal(Vec X_loc, Vec Y_loc, OperatorAp
...
+    PC     pc;
+    PCType pc_type;
 
-  // -- Local-to-Global
-  PetscCall(VecC2P(ctx->y_ceed, mem_type, Y_loc));
-  PetscCall(VecZeroEntries(D));
-  PetscCall(DMLocalToGlobal(ctx->dm_y, Y_loc, ADD_VALUES, D));
+    PetscCall(KSPGetPC(ksp, &pc));
+    PetscCall(PCGetType(pc, &pc_type));
+    PetscCall(PetscStrcmpAny(pc_type, &use_matceed_pmat, PCJACOBI, PCVPBJACOBI, PCPBJACOBI, ""));
```
jeremylt, line 475: Are there other common PCs that you use here that we could try to get MatCEED to support? I think AMG would be hard to support, but other MATOPS are easier to support

examples/fluids/Makefile
jeremylt: Hmm, its in the libCEED main Makefile too so there has to be something new to learn about Make or an old benign bug. Not sure which

examples/fluids/src/petsc_ops.c
jrwrigh: I honestly not sure what "common" PCs are used. VPBJacobi handles pretty much all of our use cases. Otherwise ILU is nice, but I think we've already established supporting that would require changes to how that is implemented in PETSc.

Other random musings from looking around haphazardly:

- [`PCKACZMARZ`](https://petsc.org/release/manualpages/PC/PCKACZMARZ/) requires `MatGetRow`. Shouldn't be too difficult to implement, but we'd need to be careful about when to assemble (currently the assembly is agressive, on every call to `MatGetDiagonal` and friends)
- [`PCASM`](https://petsc.org/release/manualpages/PC/PCASM/) Requires `MatCreateSubMatrices` (not sure if we have that), `MatIncreaseOverlap`

examples/fluids/Makefile
jrwrigh: I think it's probably what I said above; The include files are only needed for preprocessing. They're not used during the compilation or linking phase, so no need to put them there too.

examples/fluids/src/petsc_ops.c
jeremylt: PC Jacobi knows to keep the diagonal until the mat changes, so we're fine there with costs.

MatCreateSubMatrices wouldn't be easy to enable, but `MatGetRow` wouldn't be hard. I'd assemble once into the inner full AIJ and then only reassemble if the mat state changes between calls to `MatGetRow`

jeremylt (approved):
Looks good to me with the plan to keep Ratel and this code in sync

examples/fluids/Makefile
jedbrown: Note that `CFLAGS` is never used without `CPPFLAGS`, but there are times we need only `CPPFLAGS`.
```makefile
$(OBJDIR)/%.o : $(CURDIR)/%.c | $$(@D)/.DIR
	$(call quiet,CC) $(CPPFLAGS) $(CFLAGS) -c -o $@ $(abspath $<)
```
I suppose if we stored the result of preprocessing, it would technically be possible to compile without `CPPFLAGS`.

examples/fluids/src/petsc_ops.c
jrwrigh: Yeah, I think we'd need internally keep track of the the `MATCEED`'s object state. Then whenever `MatGet*` is called, [check if the state has changed ](https://petsc.org/release/manualpages/Sys/PetscObjectStateGet/#petscobjectstateget) since the last internal assembly. Then we'd indicate that internal reassembly needs to be redone via `MatAssemblyBegin/End`.

jeremylt: Yeah, I had similar logic floating around somewhere but I'm not sure where it went or why

PR #1513, opened 2024-03-13
Tidy AtPoints Scratch Vector Initalization

jeremylt:
I'm trying to track down that intermittent segfault we're seeing in CI.

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -19,7 +19,6 @@
 static int CeedOperatorSetupFields_Ref(CeedQFunction qf, CeedOperator op, bool is_input, CeedVector *e_vecs_full, CeedVector *e_vecs,
                                        CeedVector *q_vecs, CeedInt start_e, CeedInt num_fields, CeedInt Q) {
   Ceed                ceed;
-  bool                is_at_points;
   CeedSize            e_size, q_size;
   CeedInt             num_comp, size, P;
   CeedQFunctionField *qf_fields;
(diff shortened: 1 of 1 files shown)


jeremylt:
Hmm, the segfault seems to be hitting all sorts of t5* tests

backends/ref/ceed-ref-operator.c
```diff
@@ -662,9 +658,8 @@ static int CeedOperatorSetupFieldsAtPoints_Ref(CeedQFunction qf, CeedOperator op
             CeedBasisApplyAtPoints(basis, max_num_points, CEED_NOTRANSPOSE, CEED_EVAL_WEIGHT, CEED_VECTOR_NONE, CEED_VECTOR_NONE, q_vecs[i]));
         break;
     }
-    if (is_input && e_vecs[i]) {
-      CeedCallBackend(CeedVectorSetArray(e_vecs[i], CEED_MEM_HOST, CEED_COPY_VALUES, NULL));
-    }
+    if (e_vecs[i]) CeedCallBackend(CeedVectorSetValue(e_vecs[i], 0.0));
+    if (eval_mode != CEED_EVAL_WEIGHT) CeedCallBackend(CeedVectorSetValue(q_vecs[i], 0.0));
   }
```
jrwrigh, line 663: More learning moment for me, why the shift of these operations downwards?

jrwrigh (approved):
Looks all reasonable to me

backends/ref/ceed-ref-operator.c
jrwrigh: Ok,, I see why `q_vecs` got moved. But not sure about the change to `e_vecs`

jeremylt: I realized we had `CeedVectorSetValue` separately in all but 1 case, so I wanted the initialization logic together.

`CeedVectorSetArray` with a `NULL` array and `CeedVectorSetValue` with `0.0` both initialize the memory for the vector. We were using both in this function so I went with `CeedVectorSetValue` as it felt clearer what was happening than `CeedVectorSetArray` with `NULL`.

jeremylt: I updated the usages to match - interestingly only the CPU CeedOperator impls seemed to do this.

PR #1515, opened 2024-03-14
Avoid integer conversion warnings for fields in SgsDDData

lindsayad:
And also edit sample command-lines for the navierstokes example

PR #1516, opened 2024-03-14
fluids: Fix issues with 64 bit PetscInt builds

jrwrigh:
Follow on from #1514, pairs with #1515


jeremylt:
Note - will need to rebase on main for CI pipeline to run correctly.

PR #1519, opened 2024-03-14
Update to DMPlexFilter in PETSc 3.21

lindsayad:
Ran across this after updating my PETSc today

### DIFF
--- a/examples/fluids/navierstokes.h
+++ b/examples/fluids/navierstokes.h
@@ -23,6 +23,7 @@
 
 #if PETSC_VERSION_LT(3, 21, 0)
 #define DMSetCoordinateDisc(a, b, c) DMProjectCoordinates(a, b)
+#define DMPlexFilter(a, b, c, d, e, f, g) DMPlexFilter(a, b, c, g)
 #endif
 
 #define PetscCallCeed(ceed, ...)                                    \
--- a/examples/fluids/src/turb_spanstats.c
+++ b/examples/fluids/src/turb_spanstats.c
@@ -67,7 +67,7 @@ PetscErrorCode CreateStatsDM(User user, ProblemData *problem, PetscInt degree) {
     }
 
     PetscCall(DMPlexLabelComplete(user->dm, label));
-    PetscCall(DMPlexFilter(user->dm, label, 1, &user->spanstats.dm));
+    PetscCall(DMPlexFilter(user->dm, label, 1, PETSC_FALSE, PETSC_FALSE, NULL, &user->spanstats.dm));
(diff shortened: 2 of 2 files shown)


lindsayad:
I'm sure you don't want to give just anybody push access to this organization's repository, but I'll be happy to push directly here instead of my fork if that access is ever given

jrwrigh:
Yeah, I saw that change move through PETSc this morning. I'd prefer to keep this kind of fix more similar to our other ["DM API-breaking change" patches](https://github.com/CEED/libCEED/blob/main/examples%2Ffluids%2Fnavierstokes.h#L26).

lindsayad:
isn't that a little bit of a different scenario since in this case the API name is the same? Were you thinking to define a macro with a different name?

jrwrigh:
It's a little bit different, but it's the same basic idea; we want to main code to be the "up-to-date" version, but then have it fall back to an old version for compatibility. Also, in this case, I have no plans to use the added features of the API anyways. So something like:

```diff
diff --git i/examples/fluids/navierstokes.h w/examples/fluids/navierstokes.h
index 4346ecd4..914b1e6e 100644
--- i/examples/fluids/navierstokes.h
+++ w/examples/fluids/navierstokes.h
@@ -24,6 +24,7 @@

 #if PETSC_VERSION_LT(3, 21, 0)
 #define DMSetCoordinateDisc(a, b, c) DMProjectCoordinates(a, b)
+#define DMFilter(a, b, c, d, e, f, g) DMFilter(a, b, d, g)
 #endif

 // -----------------------------------------------------------------------------
 ```
 
 Plus the relevant change to `turb_spanstats.c` works well.

**Comment by jeremylt** [comment]
DMPlex tends to change at least a couple of the function signatures that we use a few times between PETSc releases. We don't tend to put a lot work into maintaining backwards compatibility against previous DMPlex function signatures since that's not needed for the primary users of the fluids code, so our typical approach is to adapt the code to the new function signature, with a macro that converts to the old function signature if someone is running the last released version of PETSc (v3.20 in this case). But that does require us updating the build on our CI machine first. So in this case, the macro would be
```C
#if PETSC_VERSION_LT(3, 21, 0)
#define DMPlexFilter(a, b, c, d, e, f, g) DMPlexFilter(a, b, c, g)
#endif
```

jrwrigh:
_For documentation's sake, [relevant PETSc MR](https://gitlab.com/petsc/petsc/-/merge_requests/7362)_

jeremylt:
I'll fix our CI build in the morning so we can merge this. Thanks!

jeremylt (approved):
CI updated, we're good to merge this

lindsayad:
thanks guys!

PR #1234, opened 2023-06-19
fluids: Compatability for PETSc w/ 64bit ints

jrwrigh:
Makes fluids example compatible with PETSc builds using 64bit integers (ie. `--with-64-bit-indices=1`). 

Note: depends on [this PETSc MR](https://gitlab.com/petsc/petsc/-/merge_requests/6618) for `PetscInt32` support.

PR #1520, opened 2024-03-16
fluids: enable reading checkpoint files as i32 or i64

jedbrown:
Depends on https://gitlab.com/petsc/petsc/-/merge_requests/7377

I think this will enable testing the i64 code path in CI, apropos https://github.com/CEED/libCEED/discussions/1514#discussioncomment-8787956

### DIFF
--- a/examples/fluids/src/misc.c
+++ b/examples/fluids/src/misc.c
@@ -86,23 +86,42 @@ PetscErrorCode DMPlexInsertBoundaryValues_FromICs(DM dm, PetscBool insert_essent
   PetscFunctionReturn(PETSC_SUCCESS);
 }
 
+static PetscErrorCode BinaryReadIntoInt(PetscViewer viewer, PetscInt *out, PetscDataType file_type) {
+  PetscFunctionBeginUser;
+  if (file_type == PETSC_INT32) {
+    PetscInt32 val;
+    PetscCall(PetscViewerBinaryRead(viewer, &val, 1, NULL, PETSC_INT32));
+    *out = val;
+  } else if (file_type == FLUIDS_FILE_TOKEN_64) {
+    PetscInt64 val;
+    PetscCall(PetscViewerBinaryRead(viewer, &val, 1, NULL, PETSC_INT64));
+    *out = val;
+  } else {
+    PetscCall(PetscViewerBinaryRead(viewer, &out, 1, NULL, PETSC_INT));
+  }
+  PetscFunctionReturn(PETSC_SUCCESS);
+}
+
 // @brief Load vector from binary file, possibly with embedded solution time and step number
 PetscErrorCode LoadFluidsBinaryVec(MPI_Comm comm, PetscViewer viewer, Vec Q, PetscReal *time, PetscInt *step_number) {
-  PetscInt   file_step_number;
-  PetscInt32 token;
-  PetscReal  file_time;
+  PetscInt      file_step_number;
+  PetscInt32    token;
+  PetscReal     file_time;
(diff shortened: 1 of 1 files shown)


examples/fluids/src/misc.c
```diff
@@ -86,23 +86,42 @@ PetscErrorCode DMPlexInsertBoundaryValues_FromICs(DM dm, PetscBool insert_essent
...
+
 // @brief Load vector from binary file, possibly with embedded solution time and step number
 PetscErrorCode LoadFluidsBinaryVec(MPI_Comm comm, PetscViewer viewer, Vec Q, PetscReal *time, PetscInt *step_number) {
-  PetscInt   file_step_number;
-  PetscInt32 token;
-  PetscReal  file_time;
+  PetscInt      file_step_number;
+  PetscInt32    token;
+  PetscReal     file_time;
+  PetscDataType file_type = PETSC_INT;
```
jrwrigh, line 110:
```suggestion
  PetscDataType file_type = PETSC_INT32;
```

The default needs to be 32bit for binary vectors that were written before the `FLUIDS_FILE_TOKEN` was used. 

I think we _could_ make this more robust by testing the token with 64bit integers like [the PETSc MR does](https://gitlab.com/petsc/petsc/-/merge_requests/7377), but I think that's not necessary; that case is only if we want to read binary vecs written with 64bit ints _and_ weren't written by the fluids code.

jrwrigh (approved):
Minor fix, but looks good to me. Thanks!

(Obviously this shouldn't be merged until [petsc/petsc!7377](https://gitlab.com/petsc/petsc/-/merge_requests/7377) is merged and the CI is updated).

examples/fluids/src/misc.c
jedbrown: Oh, I was thinking of `FLUIDS_FILE_TOKEN` before it distinguished 32- vs 64-bit. Were there no old files written with 64-bit indices? I'll defer to your experience with this evolution.

jrwrigh: At least in the fluids tests, there aren't any built with 64bit ints. I'm also pretty sure that the fluids example was never used with 64bit PETSc before #1234 (mostly because I don't think it was possible before that PR), and fluids already had the `FLUIDS_FILE_TOKEN` by that time.

examples/fluids/src/misc.c
```diff
@@ -86,23 +86,42 @@ PetscErrorCode DMPlexInsertBoundaryValues_FromICs(DM dm, PetscBool insert_essent
...
+  if (file_type == PETSC_INT32) {
+    PetscInt32 val;
+    PetscCall(PetscViewerBinaryRead(viewer, &val, 1, NULL, PETSC_INT32));
+    *out = val;
+  } else if (file_type == FLUIDS_FILE_TOKEN_64) {
+    PetscInt64 val;
+    PetscCall(PetscViewerBinaryRead(viewer, &val, 1, NULL, PETSC_INT64));
+    *out = val;
+  } else {
+    PetscCall(PetscViewerBinaryRead(viewer, &out, 1, NULL, PETSC_INT));
```
jrwrigh, line 100:
```suggestion
    PetscCall(PetscViewerBinaryRead(viewer, out, 1, NULL, PETSC_INT));
```

jrwrigh:
[petsc/petsc!7377](https://gitlab.com/petsc/petsc/-/merge_requests/7377) has been merged into PETSc main.

With the above fixes, I've confirmed that the PR should work with 32bit PETSc builds, so this should be good to merge now. 

@jeremylt then we can look at a 64bit PETSc test in CI.

PR #1528, opened 2024-03-20
Restriction CeedInt vs CeedSize

jeremylt:
I promoted the `CeedInt` to `CeedSize` in the ref and memcheck backends while indexing into the e-vecs.

@sebastiangrimberg I'm not sure if this is sufficient for your use case?

The first two commits do pure promotion, while the 3rd commit swaps the loop iterator variable to `CeedSize`. Which do we prefer? I think the 3rd commit is easier to read

### DIFF
--- a/backends/memcheck/ceed-memcheck-restriction.c
+++ b/backends/memcheck/ceed-memcheck-restriction.c
@@ -56,12 +56,12 @@ static inline int CeedElemRestrictionApplyStridedNoTranspose_Memcheck_Core(CeedE
   else CeedCallBackend(CeedElemRestrictionGetStrides(rstr, strides));
 
   // Apply restriction
-  for (CeedInt e = start * block_size; e < stop * block_size; e += block_size) {
-    CeedPragmaSIMD for (CeedInt k = 0; k < num_comp; k++) {
-      CeedPragmaSIMD for (CeedInt n = 0; n < elem_size; n++) {
-        CeedPragmaSIMD for (CeedInt j = 0; j < block_size; j++) {
-          vv[e * elem_size * (CeedSize)num_comp + (k * (CeedSize)elem_size + n) * block_size + j - v_offset] =
-              uu[n * strides[0] + k * strides[1] + CeedIntMin(e + j, num_elem - 1) * strides[2]];
+  for (CeedSize e = start * block_size; e < stop * block_size; e += block_size) {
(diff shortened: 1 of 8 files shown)


sebastiangrimberg:
I agree I think the 3rd is better and like it as is. I've added a few more fixes in [8167034](https://github.com/CEED/libCEED/pull/1528/commits/81670346db790fd69581584ebefdf7924e18afee). With these, things works (and I've checked that all the errors get resolved when building with `-ftrapv` and running). Feel free to adjust that commit as desired or let me know if something doesn't look right.

jeremylt:
Looks good to me. I think we can merge this then

sebastiangrimberg (approved):

backends/memcheck/ceed-memcheck-restriction.c
```diff
@@ -60,7 +60,7 @@ static inline int CeedElemRestrictionApplyStridedNoTranspose_Memcheck_Core(CeedE
     CeedPragmaSIMD for (CeedInt k = 0; k < num_comp; k++) {
       CeedPragmaSIMD for (CeedInt n = 0; n < elem_size; n++) {
         CeedPragmaSIMD for (CeedInt j = 0; j < block_size; j++) {
-          vv[e * elem_size * num_comp + (k * elem_size + n) * block_size + j - v_offset] =
+          vv[e * elem_size * (CeedSize)num_comp + (k * (CeedSize)elem_size + n) * block_size + j - v_offset] =
```
jedbrown: Note that the product is evaluated from left to right, with promotion only when required. So for example, this is UB (signed integer overflow), producing a negative result.
```c
    int a = 1024*1024, b = 2048, c = 3;
    long prod = a * b * (long)c;
```
https://godbolt.org/z/h8MveE9b3

I think we should either check that there cannot be overflow or promote the loop indices.

jeremylt: I swapped the indices to be `CeedSize` 33e3c8898fbd5307119640c2a4eb4dede2c8b0eb

sebastiangrimberg: I think this was addressed in [33e3c88](https://github.com/CEED/libCEED/pull/1528/commits/33e3c8898fbd5307119640c2a4eb4dede2c8b0eb), but @jeremylt can correct me if I'm wrong.

jedbrown: Oh, good. I was looking at the wrong diff.

PR #1529, opened 2024-03-22
fluids: Fix SmartSim training multiplicity, add GetInverseMultiplicity

jrwrigh:
First, found a bug int the SmartSim training, where the multiplicity correction did not factor in global multiplicity (ie. multiplicity across partitions and periodicity). This hasn't been a problem before, because all usages of inverse multiplicity have been local only. The training routine goes to the global context to remove shared data. This is fixed in the first commit.

Second, I added a `GetInverseMultiplicity` function to get the inverse of the multiplicity, be it global or local. This is implemented in a few contexts.

I've also changed the smartsim test to now use zbox with periodicity to verify both the fix and `GetInverseMultiplicity` (specifically the "global" mode of it).

Bug reported by @rickybalin


examples/fluids/smartsim_regression_framework.py
```diff
@@ -79,6 +79,9 @@ def test(self, ceed_resource) -> Tuple[bool, Exception, str]:
                 '-degree', '1',
                 '-sgs_train_enable',
                 '-sgs_train_put_tensor_interval', '2',
+                '-bc_symmetry_z', '',
```
jedbrown, line 82: What is the empty argument for?

examples/fluids/tests-output/y0_output.npy
jedbrown, line 1: Let's add `*.npy` to `.gitignore` to hopefully prevent this accident in the future.

jedbrown (approved):

examples/fluids/smartsim_regression_framework.py
jrwrigh: The `blasius.yaml` file specifies the bc_symmetry faces, but this is to unset them. The empty argument is necessary, but is there to clarify what it's doing (setting those faces to nothing). I can remove it if you want.

examples/fluids/tests-output/y0_output.npy
jrwrigh: The issue wasn't with the .npy file, but with the implementation. The npy file here is just updated for the new test that verifies the fix. The file size is cut in half because, in the move to periodicity in the spanwise direction, the global vector size is halved.

All that said, I'm fine to add `*.npy` to the `.gitignore` if you want. Though it's not much different than a CGNS file or other deliberately generated output files.

jedbrown: Yeah, okay. We do have `*.cgns` listed in `.gitignore` and I think the same logic applies here. Adding such files should feel exceptional and one should double-check the size and think about whether a smaller test could provide similar substance.

examples/fluids/smartsim_regression_framework.py
jedbrown: I suppose that's okay. It's unusual to make an zero-length argument on the command line, which is why it looked odd to my eye.

jedbrown (approved):

examples/fluids/tests-output/y0_output.npy
jrwrigh: Ope, yep. Didn't realize that cgns files were in the `.gitignore` file. Updated with `.npy`.

I've also reduced the size of the problem so the output file is 4KB instead of 470KB. My bad for not catching that the first time; I think I was focused more on test execution time when making the test originally, where the smartredis database spin-up is by far the more dominant cost than the solve. So size of the comparison file didn't come to mind.

examples/fluids/smartsim_regression_framework.py
jrwrigh: I've just gone ahead and removed the extra argument.

PR #1534, opened 2024-03-25
Only include each JiT source once

### LINKED ISSUE #1532: Don't duplicate JiT sources
Loading source to a string for JiT currently chases all local includes. It would help for debugging if we make sure we only include each file once so the contents are easier to read.

jeremylt:
Fixes #1532 

We only need to include each file once when JiTing. This is especially relevant for Ratel with all its repeated includes of the helper files.

### DIFF
--- a/interface/ceed-jit-tools.c
+++ b/interface/ceed-jit-tools.c
@@ -63,15 +63,19 @@ int CeedCheckFilePath(Ceed ceed, const char *source_file_path, bool *is_valid) {
 /**
   @brief Load source file into initialized string buffer, including full text of local files in place of `#include "local.h"`
 
-  @param[in]  ceed             `Ceed` object for error handling
-  @param[in]  source_file_path Absolute path to source file
-  @param[out] buffer           String buffer for source file contents
+  @param[in]     ceed             `Ceed` object for error handling
+  @param[in]     source_file_path Absolute path to source file
+  @param[in]     depth            Depth of recursion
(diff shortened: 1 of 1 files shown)


interface/ceed-jit-tools.c
```diff
@@ -160,8 +165,14 @@ int CeedLoadSourceToInitializedBuffer(Ceed ceed, const char *source_file_path, c
           CeedCall(CeedFree(&ceed_relative_path));
         }
         // ---- Recursive call to load source to buffer
-        CeedDebug256(ceed, CEED_DEBUG_COLOR_SUCCESS, "JiT Including: %s\n", include_source_path);
-        CeedCall(CeedLoadSourceToInitializedBuffer(ceed, include_source_path, buffer));
+        for (CeedInt i = 0; i < *num_files; i++) is_included |= !strcmp(include_source_path, (*filepaths)[i]);
```
jrwrigh, line 168: TIL that `|=` is a thing.

jrwrigh (approved):

interface/ceed-jit-tools.c
jedbrown: Just note that it's bitwise or (not logical or), and that C implicit conversion rules are permissive and sometimes surprising.
```
bool b = false;
b |= 2; // b now has value 1
b = 2; // b now has value 1
b &= 2; // b now has value 0
```

PR #1535, opened 2024-03-25
Fixed SSDB issue with SmartSim Regression Test

rickybalin:
Initializing the Python client in `examples/fluids/smartsim_regression_framework.py` would sometimes fail when the SSDB environment variable with the database address was not correctly exported. This PR fixes the issue by pasisng the database address when initializing the client.


jrwrigh (approved):
Thanks!

PR #1537, opened 2024-03-26
Use #pragma once for non-JiT headers

### LINKED ISSUE #1323: Investigate `#pragma once`
Currently, we use header guards (though they are probably redundant in most cases) to protect our JiT files from multiple definitions during creation of the JiTTed code.

We could use `#pragma once`, if it works in all our JiT environments (SYCL and OCCA are probably the two biggest concerns). See discussion here:
https://gitlab.com/petsc/petsc/-/merge_requests/6822

jeremylt:
With the way our JiT works, I think its still a good idea to use header guards for some files that will be JiTed because even after #1534 there is still ways for a user to write their QF source file such that there are duplicate definitions when combined with the default source added to all CUDA/HIP/SYCL backend source strings prior to JiT. But for our headers such as `<ceed.h>` and `<ceed/backend.h>` and similar, I think `#pragma once` should be supported by all compilers nowadays.

Still needs guards after #1534
- `ceed/types.h`
- `ceed/ceed-f32.h`
- `ceed/ceed-f64.h`
- `ceed/jit-source/cuda/*.h`
- `ceed/jit-source/cuda/*.h`
- `ceed/jit-source/hip/*.h`
- `ceed/jit-source/magma/*.h`
- `ceed/jit-source/sycl/*.h`

Undecided but probably fine to drop guards
- `ceed/jit-source/gallery/*.h`
- QF source for examples

Closes #1323

### DIFF
--- a/examples/fluids/include/ceed-utils.h
+++ b/examples/fluids/include/ceed-utils.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2023, Lawrence Livermore National Security, LLC and other CEED contributors.
+// Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and other CEED contributors.
 // All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
 //
 // SPDX-License-Identifier: BSD-2-Clause
--- a/examples/fluids/include/ceed-utils.h
+++ b/examples/fluids/include/ceed-utils.h
@@ -1,5 +1,10 @@
-#ifndef CEED_UTILS_H
-#define CEED_UTILS_H
+// Copyright (c) 2017-2023, Lawrence Livermore National Security, LLC and other CEED contributors.
+// All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
+//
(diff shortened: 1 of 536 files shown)


jrwrigh:
> Undecided but probably fine to drop guards
>   ...
>     * QF source for examples

Pretty sure they're still needed for all CPU backends, right? Those aren't run through the changes in #1534.

examples/fluids/include/ceed-utils.h
```diff
@@ -1,5 +1,10 @@
-#ifndef CEED_UTILS_H
-#define CEED_UTILS_H
+// Copyright (c) 2017-2023, Lawrence Livermore National Security, LLC and other CEED contributors.
```
jrwrigh, line 1:
```suggestion
// Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and other CEED contributors.
```

Similar to the other license headers as well.

jeremylt:
It depends upon the QF source. Probably we only need to guard QF helper files.

jrwrigh (approved):
LGTM, just the fixes to the license boiler plate.

PR #1539, opened 2024-03-26
fluids: Make Newtonian BCs compatible with explicit timestepping, add test

jrwrigh:
This makes most Newtonian BCs compatible with explicit timestepping. Also adds a test for this case. Obviously the test doesn't cover every BC, but this at least tests out the infrastructure changes necessary for the compatibility.


examples/fluids/qfunctions/bc_freestream.h
```diff
@@ -20,14 +20,12 @@ CEED_QFUNCTION_HELPER int Freestream(void *ctx, CeedInt Q, const CeedScalar *con
...
-
-  CeedScalar(*v)[CEED_Q_VLA] = (CeedScalar(*)[CEED_Q_VLA])out[0];
-  CeedScalar(*jac_data_sur)  = out[1];
+  CeedScalar(*v)[CEED_Q_VLA]       = (CeedScalar(*)[CEED_Q_VLA])out[0];
 
   const FreestreamContext        context     = (FreestreamContext)ctx;
   const NewtonianIdealGasContext newt_ctx    = &context->newtonian_ctx;
   const bool                     is_implicit = newt_ctx->is_implicit;
-  const CeedScalar               zeros[6]    = {0.};
+  CeedScalar(*jac_data_sur)                  = is_implicit ? out[1] : NULL;
```
jrwrigh, line 28: This is good to go, but I'm curious on thoughts about the implementation. I like having the inputs and outputs specified together, but the `jac_data_sur` output depends on `is_implicit` from the context. I could move the context casting with something like:

```c
CEED_QFUNCTION_HELPER int Freestream(void *ctx, CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out, StateVariable state_var,
                                     RiemannFluxType flux_type) {
  const FreestreamContext context  = (FreestreamContext)ctx;
  const CeedScalar(*q)[CEED_Q_VLA] = (const CeedScalar(*)[CEED_Q_VLA])in[0];
  const CeedScalar(*q_data_sur)    = in[2];
  CeedScalar(*v)[CEED_Q_VLA]       = (CeedScalar(*)[CEED_Q_VLA])out[0];
  CeedScalar(*jac_data_sur)        = context->newtonian_ctx.is_implicit ? out[1] : NULL;

  const NewtonianIdealGasContext newt_ctx    = &context->newtonian_ctx;
  const bool                     is_implicit = newt_ctx->is_implicit;
```

but I'm not sure if that's better. Thoughts? (@jeremylt)

jeremylt: I kinda like moving up the context a little more?

jrwrigh: Yeah, I'm leaning that way too after looking back at it.

examples/fluids/qfunctions/bc_freestream.h
```diff
@@ -154,16 +155,13 @@ CEED_QFUNCTION_HELPER CeedScalar Softplus_fwd(CeedScalar x, CeedScalar dx, CeedS
 // little or no benefit in the tests we've run thus far, thus we recommend
 // skipping this feature and just allowing recirculation.
 CEED_QFUNCTION_HELPER int RiemannOutflow(void *ctx, CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out, StateVariable state_var) {
-  // Inputs
```
jeremylt, line 157: Q - does fluids consistently not use these comments? Should pick one way or the other

examples/fluids/qfunctions/channel.h
```diff
@@ -153,27 +148,21 @@ CEED_QFUNCTION(Channel_Inflow)(void *ctx, CeedInt Q, const CeedScalar *const *in
 
     // -- Total Energy Density
     v[4][i] -= wdetJb * u_normal * (E + P);
-
-  }  // End Quadrature Point Loop
```
jeremylt, line 157: Same comment about this comment

examples/fluids/qfunctions/bc_freestream.h
jrwrigh: It's not consistent, but I (intend) not to for every code I touch.

jrwrigh: Quick grepping, definitely inconsistent, but the desire is to not have them.

examples/fluids/qfunctions/channel.h
jrwrigh: Discussion [above](https://github.com/CEED/libCEED/pull/1539#discussion_r1548310072) applies to this as well.

examples/fluids/qfunctions/bc_freestream.h
jeremylt: Cool. Consistent is better. This will just be a point of divergence between fluids and Ratel

jeremylt (approved):

PR #1540, opened 2024-03-27
Drop JiT Guards in Most QF Source

jeremylt:
Follow-up on #1534 and #1537, dropping QF source header guards where able

### DIFF
--- a/examples/fluids/qfunctions/riemann_solver.h
+++ b/examples/fluids/qfunctions/riemann_solver.h
@@ -19,7 +19,6 @@
 //
 // The right state is exterior to the domain and the left state is the interior to the domain.
 // Much of the work references Eleuterio F. Toro's "Riemann Solvers and Numerical Methods for Fluid Dynamics", 2009
-
 #include "newtonian_state.h"
 #include "newtonian_types.h"
 
--- a/examples/fluids/qfunctions/riemann_solver.h
+++ b/examples/fluids/qfunctions/riemann_solver.h
@@ -20,9 +20,6 @@
 // The right state is exterior to the domain and the left state is the interior to the domain.
 // Much of the work references Eleuterio F. Toro's "Riemann Solvers and Numerical Methods for Fluid Dynamics", 2009
 
-#ifndef riemann_solver_h
-#define riemann_solver_h
-
 #include "newtonian_state.h"
 #include "newtonian_types.h"
 
@@ -346,5 +343,3 @@ CEED_QFUNCTION_HELPER StateConservative RiemannFlux_HLLC_fwd(NewtonianIdealGasCo
     return dflux_right;
   }
 }
-
-#endif  // riemann_solver_h
--- a/examples/fluids/qfunctions/stabilization.h
+++ b/examples/fluids/qfunctions/stabilization.h
@@ -8,9 +8,6 @@
 /// @file
 /// Helper functions for computing stabilization terms of a newtonian simulation
 
-#ifndef stabilization_h
-#define stabilization_h
-
 #include <ceed.h>
 
 #include "newtonian_state.h"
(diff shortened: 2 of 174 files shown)


jeremylt:
Fluids is the only one left - need to check for multiply included files and still guard those

Working on parallel changes in Ratel

jrwrigh:
Is there any harm in adding `#pragma once` to all of them? It should work for CPUs and the GPU backends should handle the case fine, right?

jeremylt:
`#pragma once` won't do anything for the GPU backends as all the included files are put into one long string buffer, but I think I have handled that with #1532. My plan is to first drop all guards that I can without using `#pragma once` and then see how many files that leaves. Then I'll see if `#pragma once` works with what's left.

jeremylt:
Do you know if SYCL JiT handles `#pragma once` in a source file fine?

jeremylt:
Turns out, ROCm doesn't like JiTing `#pragma once` so we can't do that.

jrwrigh:
RIP, not sure about SYCL JIT, but I guess if ROCM doesn't then we probably shouldn't for any of the others.

Or (if we _really_ wanted to) we could have `CeedLoadSourceToInitializedBuffer_Private` remove any `#pragma once` and rely on it to only allow one header for the JITed backends.

jeremylt:
I'm hesitant to hide too much from the users. I think the way to go for now is to ask users to use `#ifndef` style guards for now under the umbrella of "mutually supported C99 + ROCm + CUDA" syntax. This PR drops 767 lines we don't need anymore, so I think we take the win

jedbrown:
Is there any `#pragma once` in code that could be concatenated into a string, or is that always staying in headers that we just `#include`?

jeremylt:
In this branch `pragma once` is not in any file that we turn into a string.

jrwrigh:
> I'm hesitant to hide too much from the users. 

That's valid, but also we're already implementing what the `#pragma once` directive does. So it should be transparent to the user at all times, regardless. We basically just wrote a really simple preprocessor for your JIT string processing.

But it's 6-half-dozen-or-the-other to me.

jeremylt:
The easy part is enabling the pragma for QFunctions. Backend JiT source is harder, but I have an idea

jeremylt:
Ok, "turned on" `#pragma once` for QFunction source by stripping it out.

The way that Basis source is built means this won't work without another backend only interface, but I know what I have to expose for that to work.

examples/fluids/qfunctions/riemann_solver.h
jrwrigh, line 1: This isn't necessarily recursively included, but it is included into other QFunction headers. Not sure if we want to be proactively adding `#pragma once` to headers included in other headers, or truly leave it as a "only if needed".

examples/fluids/qfunctions/stabilization.h
jrwrigh, line 1: Similar here, this file is included in other QFunction headers.

tests/t406-qfunction.h
```diff
@@ -6,11 +6,15 @@
 // This file is part of CEED:  http://github.com/ceed
 
 // Note: intentionally testing strange spacing in '#include's
```
jrwrigh, line 8: Could probably also do similar testing for `#pragma once` as well, since you're trying to allow spaces between `#` and `pragma`.

jrwrigh (approved):
Few comments, overall looks good.

examples/fluids/qfunctions/riemann_solver.h
jeremylt: My intent was to use it only where required

tests/t406-qfunction.h
jeremylt: Updated

PR #1541, opened 2024-03-27
Drop SmartSim test on ROCm

jeremylt:
@jrwrigh, the ROCm job is our longest. Do we need to test SmartSim on ROCm when we're already testing it on CUDA and doing the rest of the fluids tests on ROCm?


jrwrigh:
I'm fine dropping it.

jrwrigh (approved):

PR #1542, opened 2024-03-28
Fix JiT path cleaning logic

jeremylt:
Fixes JiT path checking for paths like `/../../foo.h`


jeremylt:
Confirmed fix works for Ratel too
https://gitlab.com/micromorph/ratel/-/jobs/6505900569

doc/sphinx/source/releasenotes.md
```diff
@@ -17,6 +17,7 @@ On this page we provide a summary of the main API changes, new features and exam
 - Add `CeedOperatorCreateAtPoints` which evaluates the `CeedQFunction` at arbitrary locations in each element, for use in Particle in Cell, Material Point Method, and similar methods.
 - Add `CeedElemRestrictionGetLLayout` to provide L-vector layout for strided `CeedElemRestriction` created with `CEED_BACKEND_STRIDES`.
 - Add `CeedVectorReturnCeed` and similar when parent `Ceed` context for a libCEED object is only needed once in a calling scope.
+- Enable `#pragma once` for JiT source; remove duplicate includes in JiT source string before compilation.
```
jrwrigh, line 20:
```suggestion
- Enable `#pragma once` for all JiT source; remove duplicate includes in JiT source string before compilation.
```

IDK why anyone *wouldn't* want a header only included once, but IIUC it's being done for all JIT files, regardless of whether they have `#pragma once`.

jrwrigh: Not sure if we want to put that information somewhere else as well.

jrwrigh (approved):

PR #1543, opened 2024-03-28
Fluids: Add stabilized mass operator for advection problem

jrwrigh:
The "normal" mass operator looks like:

```math
\int_{\Omega} \dot u v \ \mathrm{d}\Omega
```

where as the "stabilized" mass operator includes the time derivative term of the SUPG term:

```math
\int_{\Omega} \dot u v + \frac{\partial v}{\partial x} \tau a \dot u \ \mathrm{d}\Omega
```
where $a$ is the advective velocity.

Since using the "normal" mass operator is really just SU stabilization, the "stabilized" mass term is used automatically when choosing an explicit time integrator and SUPG stabilization for the advection problem.


jrwrigh:
Note this also features a correction to the RHS operator to include the `stab_tau_type: advdiff_shakib`. This should be good for review now.

jeremylt (approved):

PR #1544, opened 2024-04-01
fix(fluids): Update to new isoperiodic API

jrwrigh:
[This (my) PETSc MR](https://gitlab.com/petsc/petsc/-/merge_requests/7305) changed the interface for the isoperiodic functions. This is not a "same, but new arguments" breakage as most previous Plex breakages have been, but actually requires a small change to the code itself. Thus, a simple macro overload isn't an option.

**How do we want to handle that transitions?**

I see a two main options:
1. Just say "must use PETSc version >= 3.21.0". We've done similar things in the past
2. Add in a macro code block. I'd love to keep these things out of the code, but beggars and choosers and such.

I prefer route 1 (especially since the PETSc version macros aren't effective for builds off of PETSc `main`), but am open to either. Thoughts?

Regardless of which way we go, this will require updating the PETSc version on Noether.

### DIFF
--- a/examples/fluids/src/turb_spanstats.c
+++ b/examples/fluids/src/turb_spanstats.c
@@ -43,19 +43,22 @@ PetscErrorCode CreateStatsDM(User user, ProblemData *problem, PetscInt degree) {
   user->spanstats.span_width = domain_max[2] - domain_min[1];
 
   {  // Get DM from surface
-    DM          parent_distributed_dm;
-    PetscSF     isoperiodicface;
-    DMLabel     label;
-    PetscMPIInt size;
+    DM             parent_distributed_dm;
+    const PetscSF *isoperiodicface;
+    PetscInt       num_isoperiodicface;
+    DMLabel        label;
+    PetscMPIInt    size;
 
-    PetscCall(DMPlexGetIsoperiodicFaceSF(user->dm, &isoperiodicface));
+    PetscCall(DMPlexGetIsoperiodicFaceSF(user->dm, &num_isoperiodicface, &isoperiodicface));
 
     if (isoperiodicface) {
       PetscSF         inv_isoperiodicface;
-      PetscInt        nleaves;
+      PetscInt        nleaves, isoperiodicface_index = -1;
       const PetscInt *ilocal;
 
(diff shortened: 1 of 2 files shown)


jeremylt:
I'd like 1) and pushing all our min PETSc to 3.21

examples/fluids/src/turb_spanstats.c
```diff
@@ -43,19 +43,20 @@ PetscErrorCode CreateStatsDM(User user, ProblemData *problem, PetscInt degree) {
...
-    PetscCall(DMPlexGetIsoperiodicFaceSF(user->dm, &isoperiodicface));
+    PetscCall(DMPlexGetIsoperiodicFaceSF(user->dm, &num_isoperiodicface, &isoperiodicface));
 
     if (isoperiodicface) {
       PetscSF         inv_isoperiodicface;
       PetscInt        nleaves;
       const PetscInt *ilocal;
 
-      PetscCall(PetscSFCreateInverseSF(isoperiodicface, &inv_isoperiodicface));
+      PetscCall(PetscSFCreateInverseSF(isoperiodicface[num_isoperiodicface - 1], &inv_isoperiodicface));
```
jedbrown, line 59: This sounds like an environment assumption that the last isoperiodic dimension is the one on which to do spanwise averaging. Can we make it explicit?

jedbrown (commented):
We need to update PETSc for Noether CI. I think it's fair to require petsc >= 3.21. There may be a minor headache for Ken with branch rebasing, but we can't support ad-hoc versions of PETSc main.

jeremylt:
Updated Noether PETSc builds

jeremylt:
Can you swap all the
```
#if PETSC_VERSION_LT(3, 20, 0)
#error "PETSc v3.20 or later is required"
#endif
```
to
```
#if PETSC_VERSION_LT(3, 21, 0)
#error "PETSc v3.21 or later is required"
#endif
```
and drop the unneeded compatibility macros in the PETSc examples as part of this PR?

jrwrigh:
ROCM failure is unrelated to this branch, so I think this should be good to go.

jeremylt:
We should update the min version of the other examples, here or in a follow-up

jeremylt (approved):

Issue #1522, opened 2024-03-19
CI Use --with-strict-petscerrorcode


jeremylt:
We should check that all our examples build fine with `--with-strict-petscerrorcode` then add that to the configurations for the CI builds of PETSc

jrwrigh:
Confirmed that they all pass locally for me.

jeremylt:
Updated on Noether, CI passing on re-run

jrwrigh:
Annoyingly, the config flag apparently doesn't automatically catch this: https://github.com/CEED/libCEED/blob/47823346d50250045ce3eaaaa305de645045a703/examples/fluids/src/strong_boundary_conditions.c#L87

Though somehow `clangd` does recognize the issue:
> Ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]

Maybe we need to add an extra warning flag to our Makefile?

PR #1548, opened 2024-04-02
fix(fluids): Catch SetupStrongStg_QF

jrwrigh:
Wrap function in `PetscCall`. 

See https://github.com/CEED/libCEED/issues/1522#issuecomment-2032336521


jeremylt (approved):

jeremylt:
Maybe its the PETSc build + pedantic that catches this? Do we build the PETSc examples with the pedantic flags ever? Probably not because we throw away the libCEED return codes

jrwrigh:
I thought that too, but running (old) `main` locally with `PEDANTIC=1` didn't return any errors.

PR #1550, opened 2024-04-02
refactor(fluids): Make QF comments consistent

jrwrigh:
See discussion: https://github.com/CEED/libCEED/pull/1539#discussion_r1548312418


jeremylt (approved):
Cool, looks good to me. We need to remember to put them back in when any of these files move to Ratel

PR #1554, opened 2024-04-05
fluids: Update topological periodicity localization

### LINKED ISSUE #1470: PETSc MR 6964 causes Channel fluids examples to fail
*This is an already known failure (and the GitLab CI PETSc build has reverted that commit to allow tests to pass), but I've realized it never made it's way to a full issue and/or discussion on what the issue is or how to possibly address it.*

Starting with [PETSc MR 6964](https://gitlab.com/petsc/petsc/-/merge_requests/6964/), the Channel tests of the fluids example fail. 

My guess is that this has something to do with DMPlex's isoperiodicity, as that is the odd-feature-out between all the fluids tests and the Channel tests.

## What's (probably) happening
- the MR changed `PetscSectionFieldGetTensorDegree_Private` to get the degree of a tensor via it's `PetscDualSpace`.
- The only place `PetscSectionFieldGetTensorDegree_Private` is used is in `DMPlexSetClosurePermutationTensor`, which changes the ordering of the PetscSection DoF ordering to a tensor-product ordering

So as I see it, there are two possible ways the code is changing behavior:
- `PetscSectionFieldGetTensorDegree_Private` incorrectly reports the field as discontinuous
  - This is taken from `PetscDualSpaceLagrangeGetContinuity`
  - That function in turn gets it from a struct member, which is set in `PetscDualSpaceSetUp_Lagrange`
  - That function sets continuity based on the reported order of the dual space and whether it's `formDegree` matches the physical dimension. 
- `PetscSectionFieldGetTensorDegree_Private` does not calculate the tensor degree correctly
  - This is based on `PetscDualSpaceGetDimension`, which is in turn based on the size of the section associated with the PetscDualSpace
  
Unfortunately, I don't see an obvious place where isoperiodicity changes the DualSpace side of things. @jedbrown Any insight/ideas as to where to look?

Paging @rezgarshakeri since you made the PETSc MR.

Edit: The issue is not isoperiodicity per-say, but that periodicity requires using the Cell Coordinate DM, which then registers as discontinuous.

jrwrigh:
See https://github.com/CEED/libCEED/issues/1470 for details. Relies on PETSc MR in https://gitlab.com/petsc/petsc/-/merge_requests/7438

tl;dr is that the localization of the continuous coordinate DM to the discontinuous coordinate DM (i.e. the cell DM) must be done *after* the coordinate DM is set to the tensor product ordering.

Fixes #1470

### DIFF
--- a/examples/fluids/src/dm_utils.c
+++ b/examples/fluids/src/dm_utils.c
@@ -457,7 +457,6 @@ PetscErrorCode DMSetupByOrderBegin_FEM(PetscBool setup_faces, PetscBool setup_co
     PetscCall(PetscFECreateLagrange(comm, dim, num_comp_coord, is_simplex, fe_coord_order, q_order, &fe_coord_new));
     if (setup_faces) PetscCall(PetscFEGetHeightSubspace(fe_coord_new, 1, &fe_coord_face_new));
     PetscCall(DMSetCoordinateDisc(dm, fe_coord_new, PETSC_TRUE));
-    PetscCall(DMLocalizeCoordinates(dm));  // Update CellCoordinateDM with projected coordinates
     PetscCall(PetscFEDestroy(&fe_coord_new));
   }
   PetscFunctionReturn(PETSC_SUCCESS);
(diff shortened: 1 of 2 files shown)


jeremylt:
@ me when its ready to go on the PETSc side and I'll rebuild CI on Noether

jeremylt (approved):

jrwrigh:
I'll wait for Noether CI to pass (just to make sure), then I think we're all good. Unless you're comfortable going ahead and merging this.

PR #1555, opened 2024-04-08
Fixed small bug in sgs_dd_train Q function

rickybalin:
Fixed small bug in sgs_dd_training q function that was setting the training data to values from 0 to 11.


jrwrigh:
Bug introduced in #1529. Whoops.

jrwrigh:
Note there was also a bug in the testing framework that was actually preventing all SmartSim tests from running. This is now fixed.

jrwrigh (approved):

PR #1557, opened 2024-04-10
fluids: Add diffusion to advection problem

jrwrigh:
Adds diffusion to the advection problem (now making it a "normal" linear advection-diffusion problem). I've tested locally on an over-resolved advecting sine wave and get the analytical solution within 5 digits of accuracy over 2,000 timesteps.


jrwrigh:
Not sure what's happening with the ROCM error. It's for a completely different fluids tests, and a BPS test failed, so it's probably unreleated.

jeremylt (approved):
Just for grins, do we have a cool GIF we can make with this to show off?

jeremylt:
There's a failure on re-run, so we need to sort that out before merging

jrwrigh:
Here's a gif for ya. IDK about cool, but it shows the diffusion dampening the higher frequency mode much faster the the lower frequency mode.
out

jrwrigh:
> There's a failure on re-run, so we need to sort that out before merging

It's non-deterministic both in whether it fails and by the error metric and what backend. Regardless, a third run through passed fine, so this should be fine now.

PR #1558, opened 2024-04-12
fluids: Add option for multiple filter widths during in-situ SGS training

jrwrigh:
Setting `-sgs_train_filter_widths` to a list of real scalars, a training dataset will be created for each filter width.

This is a breaking change from previous filter width setting for SGS training; instead of using `-diff_filter_width_scaling` to define one filter, `-sgs_train_filter_widths` is used to set multiple filter widths. Moving to `-sgs_train_filter_widths` does not allow for non-isotropic scaling of the filter width (beyond the grid-based filter width); this may be revisited in the future if we want.

This also introduces the `num_filter_widths` array that is placed in the training database. It has one value, stating the number of filter widths used.


examples/fluids/src/smartsim/sgs_dd_training.c
```diff
@@ -244,71 +267,84 @@ PetscErrorCode TSMonitor_SGS_DD_Training(TS ts, PetscInt step_num, PetscReal sol
...
 
-    {
-      CeedOperatorField op_field;
-      PetscCallCeed(ceed, CeedOperatorGetFieldByName(sgs_dd_train->op_training_data_calc_ctx->op, "q", &op_field));
-      PetscCallCeed(ceed, CeedOperatorFieldGetVector(op_field, &filtered_fields));
-    }
-    PetscCall(VecPetscToCeed(FilteredFields_loc, &filtered_fields_mem_type, filtered_fields));  // filtered_fields is an implicit input
-
-    PetscCall(ApplyCeedOperatorGlobalToGlobal(FilteredVelocityGradient, TrainingData, sgs_dd_train->op_training_data_calc_ctx));
+      PetscCall(DifferentialFilterApply(user, solution_time, Q, FilteredFields));
```
jrwrigh, line 291: Note we are calling `DifferentialFilterApply` multiple times per filter width. This is slightly inefficient, as the RHS gets recalculated each time, even though the value is the same.

One way around this would be to save the RHS result and track the `PetscObjectState` of the solution vector used to create the RHS vector; if the solution vector state hasn't changed, then the previous existing RHS doesn't need to be updated. However, [`PetscObjectStateGet`](https://petsc.org/release/manualpages/Sys/PetscObjectStateGet/#petscobjectstateget) is kept behind `include/petsc/private/petscimpl.h`. 

@jedbrown @jeremylt IDK what y'all's thoughts are behind using functions in `include/petsc/private/petscimpl.h`, but I'm guessing I probably shouldn't be doing that?

jeremylt: I'm using it inside of `ratel-pmg.c`. I wonder if maybe `VecGetState` and `MatGetState` should be public PETSc functions?

jrwrigh: Ok, I've implemented the `PetscObjectState` solution in https://github.com/CEED/libCEED/pull/1558/commits/56f5e8119536c76269613d7f3474b8135eed8bb0



> I wonder if maybe `VecGetState` and `MatGetState` should be public PETSc functions?

Not sure. Probably though? I don't see any reason for them not to be.

jeremylt: https://gitlab.com/petsc/petsc/-/merge_requests/7470

examples/fluids/include/smartsim.h
```diff
@@ -13,16 +13,17 @@
...
-    PetscBool disable_calls = PETSC_FALSE;                                                                                            \
-    PetscStackUpdateLine;                                                                                                             \
-    PetscCall(PetscOptionsGetBool(NULL, NULL, "-smartsim_disable_calls", &disable_calls, NULL));                                      \
-    if (disable_calls == PETSC_TRUE) break;                                                                                           \
-    ierr_smartredis_call_q_ = __VA_ARGS__;                                                                                            \
-    if (PetscUnlikely(ierr_smartredis_call_q_ != SRNoError))                                                                          \
-      SETERRQ(PETSC_COMM_SELF, ierr_smartredis_call_q_, "SmartRedis Error (Code %d): %s", ierr_smartredis_call_q_, SRGetLastError()); \
+#define PetscSmartRedisCall(...)                                                                                                   \
```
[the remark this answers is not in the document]
jrwrigh: Thanks!

rickybalin:
Testing with full workflow revealed that the time step number was being sent to the DB computing training data for each of the filter widths. This was causing issues with the online ML workflow. Now the time step number is sent after all the training data, with all filter widths, is computed.

jeremylt (approved):
LGTM

Issue #1090, opened 2022-11-12
Improve Fluids boundary condition prescribing options


jrwrigh:
Stemming from [this comment](https://github.com/CEED/libCEED/pull/1088#pullrequestreview-1177915045):
> At some point, we may want to reflect on making our options orthogonal, and maybe -bc_inflow 1 -bc_inflow_stg 2 to apply fluctuations on one surface, but laminar on the other.

I'd like to move to more generic boundary condition definitions, though it'd be a significant change, so I'd want to get some feedback/thumbs up before really going at it. I've honestly been thinking about changing it for awhile now. 

## General Idea
Effectively, rather than have dedicated flags for each bc type,  (ie. the `bc_{inflow,outflow,freestream,wall,slip}` flags), group faces into generic BC definition structures. This is more similar to how PHASTA defines it's boundary conditions (though I'm not the most knowledgeable when it comes to the details for how that's done).

The end result is something that's more composable than what we currently have. I've divided my idea into a UI and Backend implementation, which can be implemented independently. The UI might be difficult to implement, as I don't know how flexible PETSc's Options Database is. I think the Backend would be a fairly straight forward and simplify the code a decent bit, particularly when it comes to trying out different BC combinations.

## User Interface 
So the user interface would be something like:
```yaml
bcs:
  - faces: 1,2
    type: problem_inflow

  - faces: 3,4
    type: custom
    strong: stg #(or could be constant)
    strong_components: 1, 2, 3
    weak: constant
    weak_components: 4
    weak_temperature: 288

  - faces: 5
    type: custom
    strong: constant
    strong_components: 1, 2, 3, 4
    strong_state: 0, 0, 0, 288 # No-slip wall + specified temperature
    weak: consistency_integral

```
I'm know this is valid YAML, but I'm not sure how PETSc would interpret the sequence of sets of key-value pairs into it's options database (if it even can do that, as I'm not sure how it'd translate onto command line). At a minimum, it conveys the general idea:
- Have a list of BC definitions
- `type` can be either a problem defined BC, periodic, or custom
- `type: custom` can define out specific implementations of strong and weak enforcement

## Backend 
In the code itself, we'd have enums for the different options and a `BCDefinition` struct that would contain all the data needed for the boundary condition setup:
```c
// Already implemented in code, here for clarity
typedef struct {
  CeedQFunctionUser    qfunction;
  const char           *qfunction_loc;
  CeedQFunctionContext qfunction_context;
} ProblemQFunctionSpec;

[...]

PR #1559, opened 2024-04-16
fix(fluids): Fix primary PETSc memory leak, others

### LINKED ISSUE #1490: fluids: Fix memory leak
When running with `-malloc_dump`, PETSc reports quite a few memory leaks. Judging from the earliest report:
```
[ 0] 4992 bytes
      [0]  DMCreate() at /home/jrwrigh/software/petsc/src/dm/interface/dm.c:63
      [0]  CreateDM() at /home/jrwrigh/software/libCEED/examples/fluids/src/setupdm.c:22
```
I'm guessing the `DM` is being duplicated somewhere, but not being freed. Or, more likely, it's being duplicated, attached to another struct, and that struct isn't running `DMDestroy` on it.

Note that I can go all the way back to `7fb7f898` (#1143) and the problem is still present. So it's something very fundamental that is wrong.

jrwrigh:
All the memory leaks reported in regression tests via `-malloc_dump` are now related to the boundary conditions and their handling of QFunctionContexts, which is a much more difficult problem and will be addressed later by implementing #1090.

Fixes #1490

### DIFF
--- a/examples/fluids/src/misc.c
+++ b/examples/fluids/src/misc.c
@@ -50,23 +50,30 @@ PetscErrorCode ICs_FixMultiplicity(DM dm, CeedData ceed_data, User user, Vec Q_l
 
 // Record boundary values from initial condition
 PetscErrorCode SetBCsFromICs(DM dm, Vec Q, Vec Q_loc) {
-  Vec Qbc, boundary_mask;
-
   PetscFunctionBeginUser;
-  PetscCall(DMGetNamedLocalVector(dm, "Qbc", &Qbc));
-  PetscCall(VecCopy(Q_loc, Qbc));
-  PetscCall(VecZeroEntries(Q_loc));
-  PetscCall(DMGlobalToLocal(dm, Q, INSERT_VALUES, Q_loc));
-  PetscCall(VecAXPY(Qbc, -1., Q_loc));
-  PetscCall(DMRestoreNamedLocalVector(dm, "Qbc", &Qbc));
+  {  // Capture initial condition values in Qbc
+    Vec Qbc;
+
+    PetscCall(DMGetNamedLocalVector(dm, "Qbc", &Qbc));
+    PetscCall(VecCopy(Q_loc, Qbc));
+    PetscCall(VecZeroEntries(Q_loc));
+    PetscCall(DMGlobalToLocal(dm, Q, INSERT_VALUES, Q_loc));
+    PetscCall(VecAXPY(Qbc, -1., Q_loc));
+    PetscCall(DMRestoreNamedLocalVector(dm, "Qbc", &Qbc));
+  }
   PetscCall(PetscObjectComposeFunction((PetscObject)dm, "DMPlexInsertBoundaryValues_C", DMPlexInsertBoundaryValues_FromICs));
 
-  PetscCall(DMGetNamedLocalVector(dm, "boundary mask", &boundary_mask));
-  PetscCall(DMGetGlobalVector(dm, &Q));
-  PetscCall(VecZeroEntries(boundary_mask));
-  PetscCall(VecSet(Q, 1.0));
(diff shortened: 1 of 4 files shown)


examples/fluids/src/misc.c
```diff
@@ -62,11 +62,12 @@ PetscErrorCode SetBCsFromICs(DM dm, Vec Q, Vec Q_loc) {
...
   PetscCall(DMGetNamedLocalVector(dm, "boundary mask", &boundary_mask));
-  PetscCall(DMGetGlobalVector(dm, &Q));
+  PetscCall(DMGetGlobalVector(dm, &ones));
   PetscCall(VecZeroEntries(boundary_mask));
-  PetscCall(VecSet(Q, 1.0));
-  PetscCall(DMGlobalToLocal(dm, Q, INSERT_VALUES, boundary_mask));
+  PetscCall(VecSet(ones, 1.0));
+  PetscCall(DMGlobalToLocal(dm, ones, INSERT_VALUES, boundary_mask));
   PetscCall(DMRestoreNamedLocalVector(dm, "boundary mask", &boundary_mask));
+  PetscCall(DMRestoreGlobalVector(dm, &ones));
```
jedbrown, line 70: Purely stylistic, but you could declare it closer.
```c
{
  Vec ones;
  PetscCall(DMGetGlobalVector(dm, &ones));
  PetscCall(VecZeroEntries(boundary_mask));
  PetscCall(VecSet(ones, 1.0));
  PetscCall(DMGlobalToLocal(dm, ones, INSERT_VALUES, boundary_mask));
  PetscCall(DMRestoreNamedLocalVector(dm, "boundary mask", &boundary_mask));
  PetscCall(DMRestoreGlobalVector(dm, &ones));
}
```

jedbrown (approved):
Thanks for tracking these down.

examples/fluids/src/misc.c
jrwrigh: Yep, I thought about doing that too, but didn't execute for whatever reason.

examples/fluids/problems/newtonian.c
```diff
@@ -363,8 +363,10 @@ PetscErrorCode NS_NEWTONIAN_IG(ProblemData *problem, DM dm, void *ctx, SimpleBC
   problem->apply_vol_rhs.qfunction_context = newtonian_ig_context;
   PetscCallCeed(ceed, CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_vol_ifunction.qfunction_context));
   PetscCallCeed(ceed, CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_vol_ijacobian.qfunction_context));
-  PetscCallCeed(ceed, CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_inflow.qfunction_context));
-  PetscCallCeed(ceed, CeedQFunctionContextReferenceCopy(newtonian_ig_context, &problem->apply_inflow_jacobian.qfunction_context));
+  if (bc->num_inflow > 0) {
```
jeremylt, line 366: Its safe to Destroy NULL, so is there a way to get around needing so many > 0 checks?

jeremylt: This sort of reduction in special casing logic would probably be the sort of thing that could spread all over the place and would probably be worth being its own branch

jrwrigh: Yeah, I think so. I realized that the context destroying code in `navierstokes.c` repeated a few values, so I fixed that in an update. I'll double check if that absolves the need for the if statement here.

As for the other `> 0` checks, that necessary for the boundary condition setup in general. Something I want to get rid of with a re-write of the BC handling code. But that's for another day.

jeremylt: fair, just wanted to put that on the radar since its bit me a lot in these projects

jeremylt (approved):

PR #1560, opened 2024-04-17
fluids: Remove ceed backend from advection.yaml

jrwrigh:
Due to ordering, this overides the regression tests. So all the new advection tests that used `advection.yaml` were actually running with the wrong ceed backend the entire time. Let's hope everything still passes...


jedbrown (approved):
Good catch

PR #1561, opened 2024-04-22
fluids - use VecGetState

jeremylt:
`VecGetState` is available in PETSc now


jrwrigh (approved):
Also need to remove the `#include <petsc/private/petscimpl.h>` line at the top of the file for the private PETSc header.

Thanks!

jeremylt:
Note - the intermittent failures for MFEM and fluids in CI are still there. Odd

PR #1562, opened 2024-04-24
fluids: Make ProblemData a pointer to struct

jrwrigh:
We basically use `problem` as a pointer-to-struct type everywhere, so might as well make it consistent with the other structs.


jrwrigh:
Less of a catch and more of a "yeah... might as well fix that now". I've known about the discrepancy for awhile, just didn't change it because "too busy" (even though it only took 4 minutes to fix). :sweat_smile:

jrwrigh:
Minor addition of a missing `PetscCall`

PR #1565, opened 2024-04-25
Add Diagonal Assembly for AtPoints

jeremylt:
Adding diagonal assembly for AtPoints operators by probing each element with unit vectors

- [x] CPU impl
- [x] Tests

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -1109,8 +1109,212 @@ static int CeedOperatorLinearAssembleQFunctionAtPointsUpdate_Ref(CeedOperator op
 }
 
 //------------------------------------------------------------------------------
-// Assemble Operator
+// Assemble Operator Diagonal AtPoints
 //------------------------------------------------------------------------------
+static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Ref(CeedOperator op, CeedVector assembled, CeedRequest *request) {
+  bool                is_active_at_points = true;
+  CeedInt             num_points_offset = 0, num_input_fields, num_output_fields, num_elem, elem_size_active = 1, num_comp_active;
+  CeedScalar         *e_data[2 * CEED_FIELD_MAX] = {0};
+  Ceed                ceed;
+  CeedVector          point_coords = NULL, in_vec, out_vec;
+  CeedElemRestriction rstr_points  = NULL;
+  CeedQFunctionField *qf_input_fields, *qf_output_fields;
+  CeedQFunction       qf;
+  CeedOperatorField  *op_input_fields, *op_output_fields;
+  CeedOperator_Ref   *impl;
+
+  CeedCallBackend(CeedOperatorGetData(op, &impl));
+  CeedCallBackend(CeedOperatorGetNumElements(op, &num_elem));
+  CeedCallBackend(CeedOperatorGetQFunction(op, &qf));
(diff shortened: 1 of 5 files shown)


examples/petsc/bpsswarm.c
```diff
@@ -302,14 +303,9 @@ int main(int argc, char **argv) {
...
       PetscCall(PCSetType(pc, PCJACOBI));
-      PetscCall(PCJacobiSetType(pc, PC_JACOBI_ROWSUM));
+      PetscCall(PCJacobiSetType(pc, PC_JACOBI_DIAGONAL));
     } else {
       PetscCall(PCSetType(pc, PCNONE));
-      MatNullSpace nullspace;
-
-      PetscCall(MatNullSpaceCreate(PETSC_COMM_WORLD, PETSC_TRUE, 0, 0, &nullspace));
-      PetscCall(MatSetNullSpace(mat_O, nullspace));
-      PetscCall(MatNullSpaceDestroy(&nullspace));
```
jeremylt, line 312: I made the two match. Alternatively, we could add this to `bps.c`?

zatkins-dev: Ah, I think I copied from `bpssphere.c` originally -- if this isn't needed here I'm good with removing.

backends/ref/ceed-ref-operator.c
```diff
@@ -1109,8 +1109,212 @@ static int CeedOperatorLinearAssembleQFunctionAtPointsUpdate_Ref(CeedOperator op
...
+    CeedCallBackend(CeedElemRestrictionApplyAtPointsInElement(rstr_points, e, CEED_NOTRANSPOSE, point_coords, impl->point_coords_elem, request));
+    CeedCallBackend(CeedElemRestrictionGetNumPointsInElement(rstr_points, e, &num_points));
+
+    // Input basis apply for non-active bases
+    CeedCallBackend(CeedOperatorInputBasisAtPoints_Ref(e, num_points_offset, num_points, qf_input_fields, op_input_fields, num_input_fields, in_vec,
+                                                       impl->point_coords_elem, true, e_data, impl, request));
+
+    // Loop over points on element
+    e_vec_size = (is_active_at_points ? num_points : elem_size_active) * num_comp_active;
+    for (CeedInt s = 0; s < e_vec_size; s++) {
```
zatkins-dev, line 1192: Just to clarify, for our typical Jacobian operator at points, this is a loop over the nodes of each element that computes the diagonal contributions for that node from the points within that element, right?

I want to make sure we aren't doing something with like `O(num_points^2)`  complexity.

zatkins-dev (approved):
Looks good to me, thanks! Have you looked at any performance comparison between the old implementation and this one?

backends/ref/ceed-ref-operator.c
jeremylt: For our typical approach, we assemble the QF by repeated application of D [1, 0, 0...] and then compute the matrix triple product B D B^T and extract the element operator matrix diagonal. (we might be doing something clever there so we don't compute the full matrix triple product)

Here we are instead repeatedly applying B D B^T [1, 0, 0...] to build the element operator matrix diagonal.

This PR should be more expensive than our typical approach when we have consistent quadrature points in each element but less expensive than building B for every element

zatkins-dev: Gotcha, sounds good to me.

PR #1566, opened 2024-04-25
fluids: Use State in blasius context

jrwrigh:
Instead of storing velocity, temperature, and pressure separately, just store a `State` struct.

Also clarifies usage of `P0` in the `NewtonianIdealGasContext` struct.


jeremylt (approved):
All seems reasonable to me

PR #1571, opened 2024-04-29
fluids: Add CreateQData and CreateQDataBoundary functions

jrwrigh:
Create function for creating Qdata (ie. Separate the setup functions and other info from everything else)
So we can remove `ceed_data->qf_setup_sur`, which is used by the spanstats setup. This will also remove `spec_setup_{vol,sur}` as well.

Depends on https://github.com/CEED/libCEED/pull/1567

### DIFF
--- a/examples/fluids/src/qdata.c
+++ b/examples/fluids/src/qdata.c
@@ -1,0 +1,199 @@
+// Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and other CEED contributors.
+// All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+//
+// This file is part of CEED:  http://github.com/ceed
+
+#include "../navierstokes.h"
+
+#include <petscsection.h>
+#include "../qfunctions/setupgeo.h"
+#include "../qfunctions/setupgeo2d.h"
+
+/**
+ * @brief Get number of components of quadrature data for domain
+ *
+ * @param[in]  dm          DM where quadrature data would be used
+ * @param[out] q_data_size Number of components of quadrature data
+ */
+PetscErrorCode QDataGetNumComponents(DM dm, CeedInt *q_data_size) {
+  PetscInt num_comp_x, dim;
+
+  PetscFunctionBeginUser;
+  PetscCall(DMGetDimension(dm, &dim));
+  {  // Get number of coordinate components
+    DM           dm_coord;
+    PetscSection section_coord;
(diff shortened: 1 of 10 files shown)


examples/fluids/src/qdata.c
```diff
@@ -0,0 +1,199 @@
...
+ * @param[in]  domain_label  DMLabel that quadrature data would be used one
+ * @param[in]  label_value   Value of label
+ * @param[in]  elem_restr_x  CeedElemRestriction of the coordinates (must match `domain_label` and `label_value` selections)
+ * @param[in]  basis_x       CeedBasis of the coordinates
+ * @param[in]  x_coord       CeedVector of the coordinates
+ * @param[out] elem_restr_qd CeedElemRestriction of the quadrature data
+ * @param[out] q_data        CeedVector of the quadrature data
+ * @param[out] q_data_size   number of components of quadrature data
+ */
+PetscErrorCode QDataGet(Ceed ceed, DM dm, DMLabel domain_label, PetscInt label_value, CeedElemRestriction elem_restr_x, CeedBasis basis_x,
```
jeremylt, line 73: Is there ever a point in which you will call `QDataGet` for the same volumetric region multiple times and might  want to re-use the q-data?

jrwrigh: Yep. That would help remove `ceed_data->elem_restr_qd_i` and `ceed_data->q_data`. I originally had this named `QDataCreate`, but realized I'd probably want to do that in the future, so changed to name to `Get`.

Primary roadblock there is how to ID a `{elem_restr_qdata, q_data, q_data_size}` tuple. I don't want to ID off of the DM, as many DMClones can (and currently do) use the same qdata. The other thought would be to always create a `elem_restr_qd`, then ID off of that, but idk if we have a `CeedElemRestrictionCompare` or a way to get equivalent functionality.

jrwrigh: Or, always create the `q_data` vector, then compare values (with some tolerance). A bit expensive, but pretty much fool proof.

jeremylt: We have an object corresponding to material region in Ratel and that object keeps track of the QData but I'm not entirely satisfied with it as a general purpose approach. I do think ultimately we'll want to attach to the DM as part of the eventual upstreaming to PETSc. Its not a blocker here but definitely something to think about as we are maturing the integration between PETSc and libCEED

jeremylt: For Ratel the material region keeps track of its own boundary since some of our boundary operators rely upon the material model selected for the region

jeremylt (approved):
There we go, Noether is happy now

PR #1573, opened 2024-05-01
fluids: Correct printing of MatTypes

jrwrigh:
This does a few things (beyond the title):

- Makes `MatCeedSetUpInnerMat` public
    - Previous usage of `MatConvert` did not actually work in my testing
    - Additionally, `MatConvert` does manual assembly of the matrix during the conversion, when all we want is a matrix ready for COO assembly.
    - @jeremylt I'm not sure if this is the "best" solution, but it was the simplest.
- Moves the run info printing to be done right before `TSSolve` is called
- Gets it's Amat and Pmat info from the `TS` directly.
    - This printing was incorrect previously
    
I'm not 100% sure on the format of the info being printed right now. A few questions for feedback:

Do we even want to print out `A MatType` or `P MatType` when running explicit? They don't really mean anything:
```
$ build/fluids-navierstokes -ceed /cpu/self/memcheck/blocked -options_file examples/fluids/gaussianwave.yaml -dm_plex_box_faces 2,2,1 -ts_max_steps 5 -degree 3 -implicit false -ts_type rk -stab supg -state_var conservative -mass_ksp_type gmres -mass_pc_jacobi_type diagonal

[...]
  PETSc:
    Box Faces                          : 2,2,1
    A MatType                          : seqaij
    P MatType                          : seqaij
```

Currently, instead of `shell`, it will print `ceed`. This obviously conflicts with the `-amat_type` flag. We could go with deprecating the `-amat_type` flag like I had proposed awhile back (and I think we all oked the idea? Relevant [Zulip link](https://phypid.zulipchat.com/#narrow/stream/232375-ceed-phasta/topic/Change.20default.20Amat.20type.20behavior/near/426357458). But not sure if we want to do that now.

### DIFF
--- a/examples/fluids/src/misc.c
+++ b/examples/fluids/src/misc.c
@@ -397,8 +397,10 @@ PetscErrorCode RegisterLogEvents() {
 }
 
 // Print information about the given simulation run
-PetscErrorCode PrintRunInfo(User user, Physics phys_ctx, ProblemData problem, MPI_Comm comm) {
-  Ceed ceed = user->ceed;
+PetscErrorCode PrintRunInfo(User user, Physics phys_ctx, ProblemData problem, TS ts) {
+  Ceed     ceed = user->ceed;
+  MPI_Comm comm = PetscObjectComm((PetscObject)ts);
+
   PetscFunctionBeginUser;
   // Header and rank
   char        host_name[PETSC_MAX_PATH_LEN];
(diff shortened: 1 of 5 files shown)


examples/fluids/src/petsc_ops.c
```diff
@@ -323,7 +321,7 @@ PetscErrorCode CreateSolveOperatorsFromMatCeed(KSP ksp, Mat mat_ceed, PetscBool
   }
 
   if (assemble_amat) {
-    PetscCall(MatConvert(mat_ceed, mat_ceed_inner_type, MAT_INITIAL_MATRIX, Amat));
+    PetscCall(MatCeedSetupInnerMat(mat_ceed, Amat));
```
jeremylt, line 324: We definitely don't want to `MatConvert` here for sure. We don't want to lose the `mat_ceed`.

This not what `MatCeedSetupInnerMat` is for per the description so I'd be really careful making a function public that was designed to only be working on internal data. The language around "InnerMat" would be very confusing.

How about

- Convert `MatCeedSetupInnerMat` to `MatCeedCreateMatCOO`
- Rename option `ceed_inner_mat_type` to `ceed_coo_mat_type`
- Split the COO setup out of `MatCeedAssembleCOO` into `MatCeedSetPreallocationCOO`
- Call `MatCeedSetPreallocationCOO` in both `MatCeedAssembleCOO` and `MatCeedCreateMatCOO`

Thoughts?

jrwrigh: > We definitely don't want to `MatConvert` here for sure. We don't want to lose the `mat_ceed`.

Well, `MatConvert` doesn't lose `mat_ceed`; it generates an entirely new matrix. We just can't use that matrix for COO assembly. Or at least that's my interpretation of the errors that came from trying to do that.

> This not what `MatCeedSetupInnerMat` is for per the description so I'd be really careful making a function public that was designed to only be working on internal data. The language around "InnerMat" would be very confusing.

Yeah, I initially tried to make a new function, but couldn't figure out what the structure should be (or the naming) and just went with publicizing `MatCeedSetupInnerMat` as the simplest proof-of-concept.

> 
> How about
>  * Convert `MatCeedSetupInnerMat` to `MatCeedCreateMatCOO`

:+1: 

>  * Rename option `ceed_inner_mat_type` to `ceed_coo_mat_type`

:+1: 

>  * Split the COO setup out of `MatCeedAssembleCOO` into `MatCeedSetPreallocationCOO`
>  * Call `MatCeedSetPreallocationCOO` in both `MatCeedAssembleCOO` and `MatCeedCreateMatCOO`

I think I prefer keeping assembly out of `MatCeedCreateMatCOO`. Based on other errors I was encountering, the assembly of our Jacobian matrix can't happen until the IFunction is run (and `jac_data` is populated) and the matrix creation needs to happen before TSSolve. I think also having `MatCeedCreateMatCOO` sounds more analogous to `DMCreateMat`, which also just creates an "empty" `Mat` object.

jeremylt: Right, I'm proposing that assembly doesn't happen during `MatCeedCreateMatCOO`, only setting the COO pattern

jrwrigh: Ope, misread. I guess I'm not sure why we'd want to have the COO preallocation in both `MatCeedCreateMatCOO` and `MatCeedAssembleCOO`? Leaving it only in `MatCeedAssembleCOO` is the simpler change; we can always migrate to COO-setup in both later.

jeremylt: My thinking was that you could View the Mat with the sparsity pattern already set out of the box. I can show you what I'm thinking in a quick Ratel MR - its the same complexity either way.

jeremylt: https://gitlab.com/micromorph/ratel/-/merge_requests/831

jrwrigh: https://github.com/CEED/libCEED/pull/1574

jrwrigh:
Alright, running the full gamit of (common) configurations, we have:

- Implicit with fully assembled preconditioner:
```
$ build/fluids-navierstokes -ceed /cpu/self/memcheck/blocked -options_file examples/fluids/gaussianwave.yaml -dm_plex_box_faces 2,2,1 -ts_max_steps 5 -degree 3 -amat_type shell -pc_type ilu -ts_alpha_radius 0.5

[...]
  PETSc:
    Box Faces                          : 2,2,1
    IJacobian A MatType                : ceed
    IJacobian P MatType                : seqaij
  Mesh:
```

- Implicit with MatCeed preconditioner (will display the inner COO MatType):
```
$ build/fluids-navierstokes -ceed /cpu/self/memcheck/blocked -options_file examples/fluids/gaussianwave.yaml -dm_plex_box_faces 2,2,1 -ts_max_steps 5 -degree 3 -amat_type shell -pc_type vpbjacobi -ts_alpha_radius 0.5

[...]
  PETSc:
    Box Faces                          : 2,2,1
    IJacobian A MatType                : ceed
    IJacobian P MatType                : ceed (COO MatType: aij)
  Mesh:
```

- Explicit (doesn't show the `IJacobian * MatType` at all)
```
$ build/fluids-navierstokes -ceed /cpu/self/memcheck/blocked -options_file examples/fluids/gaussianwave.yaml -dm_plex_box_faces 2,2,1 -ts_max_steps 5 -degree 3 -implicit false -ts_type rk -stab supg -state_var conservative -mass_ksp_type gmres -mass_pc_jacobi_type diagonal

[...]
  PETSc:
    Box Faces                          : 2,2,1
    Time Stepping Scheme               : explicit
  Mesh:
```

I'm pretty happy with this now.

examples/fluids/src/misc.c
```diff
@@ -397,7 +397,7 @@ PetscErrorCode RegisterLogEvents() {
 }
 
 // Print information about the given simulation run
-PetscErrorCode PrintRunInfo(User user, Physics phys_ctx, ProblemData problem, MPI_Comm comm) {
+PetscErrorCode PrintRunInfo(User user, Physics phys_ctx, ProblemData problem, TS ts, MPI_Comm comm) {
```
jedbrown, line 400: We don't need `comm` if we already pass a `TS`. (There is also `user->comm`, which should match.)

jedbrown (approved):

PR #1574, opened 2024-05-02
fluids: MatCeed and Ratel Updates

jrwrigh:
Sync with Ratel, mostly for MatCeed. Namely:
- `MatCeedCreateCOOMat`: https://gitlab.com/micromorph/ratel/-/merge_requests/831
- `MatCeedSetContextDouble`

*EDIT: moved to #1575*: Note @jeremylt I also added a `MatCeed{Get,Set}ContextReal` instead of the `MatCeedSet{Time,Dt,Shifts}`. I think that makes more sense than the application-specific `MatCeedSet*`for upstreaming to PETSc. I could see them as `RatelJacobianSet{Time,Dt,Shifts}` though.

### DIFF
--- a/examples/fluids/include/mat-ceed-impl.h
+++ b/examples/fluids/include/mat-ceed-impl.h
@@ -7,45 +7,16 @@
 #pragma once
 
 #include <ceed.h>
+#include <petsc-ceed.h>
 #include <petscdm.h>
 #include <petscmat.h>
 #include <petsc/private/petscimpl.h>
-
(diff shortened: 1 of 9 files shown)


jeremylt:
Do you have specific examples of context values you want to set that are not dt, time, or the shifts? I'd split this PR in half - one half syncing with Ratel and one half proposing an interface change.

jrwrigh:
PR split now.

jeremylt (approved):

PR #1575, opened 2024-05-02
fluids: Add `MatCeed{Get,Set}ContextReal`

jrwrigh:
*Split from #1574, copied relevant description from there:*

> Note @jeremylt I also added a `MatCeed{Get,Set}ContextReal` instead of the `MatCeedSet{Time,Dt,Shifts}`. I think that makes more sense than the application-specific `MatCeedSet*` for upstreaming to PETSc. I could see them as `RatelJacobianSet{Time,Dt,Shifts}` though.

### DIFF
--- a/examples/fluids/include/mat-ceed.h
+++ b/examples/fluids/include/mat-ceed.h
@@ -26,6 +26,8 @@ PETSC_CEED_INTERN PetscErrorCode MatCeedGetContextDouble(Mat mat, const char *na
 // Advanced functionality
 PETSC_CEED_EXTERN PetscErrorCode MatCeedSetContext(Mat mat, PetscErrorCode (*f)(void *), void *ctx);
 PETSC_CEED_EXTERN PetscErrorCode MatCeedGetContext(Mat mat, void *ctx);
+PETSC_CEED_EXTERN PetscErrorCode MatCeedSetContextReal(Mat mat, const char *name, PetscReal value);
+PETSC_CEED_EXTERN PetscErrorCode MatCeedGetContextReal(Mat mat, const char *name, PetscReal *value);
 
 PETSC_CEED_EXTERN PetscErrorCode MatCeedSetOperation(Mat mat, MatOperation op, void (*g)(void));
 PETSC_CEED_EXTERN PetscErrorCode MatCeedSetCOOMatType(Mat mat, MatType type);
--- a/examples/fluids/src/mat-ceed.c
+++ b/examples/fluids/src/mat-ceed.c
@@ -710,6 +710,25 @@ PetscErrorCode MatCeedSetContextDouble(Mat mat, const char *name, double value) 
 }
 
 /**
+  @brief Set the current `PetscReal` value of a context field for a `MatCEED`.
+
+  Not collective across MPI processes.
+
+  @param[in,out]  mat    `MatCEED`
+  @param[in]      name   Name of the context field
+  @param[in]      value  New context field value
+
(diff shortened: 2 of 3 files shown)


jrwrigh:
*From [comment](https://github.com/CEED/libCEED/pull/1574#issuecomment-2091782647) in #1574*:
> Do you have specific examples of context values you want to set that are not dt, time, or the shifts?

Yep, I use a label for the differential filter width, which is changed for different training datasets:
https://github.com/CEED/libCEED/blob/f7488153ba60dd9aeacf17f1494bdb6a59881ae6/examples/fluids/src/differential_filter.c#L152

jrwrigh:
Actually, in that case, I'm actually setting an array of `double`s, not just a scalar, which won't work with the current `MatCeedSetContextDouble` interface.

jeremylt:
I'd prefer to keep `{Dt, Time, Shifts}` so the user doesn't have to know what we called those fields internally, just what PETSc calls them.

How about we add `MatCeedSetContextDouble` and `MatCeedGetContextDoubles`? I used `Double` in the name to be clear that a cast was required to update these values

jrwrigh:
> I'd prefer to keep `{Dt, Time, Shifts}` so the user doesn't have to know what we called those fields internally, just what PETSc calls them.

Maybe that's my misunderstanding. Do we want to enforce to the PETSc user what those label names should be? I think it makes sense for Ratel and HONEE to enforce consistent label names, but I'm not sure if that should be enforced to the PETSc user. I also might not be seeing the bigger picture of what that integration looks like. Are we wanting a PETSc user to hand over a `MatCeed` as a jacobian and (somehow?) have the `TSIJacobianFn` (as in `DMTSSetIJacobian`) be automatically set to a `FormIJacobian_CEED`?
 
> How about we add `MatCeedSetContextDouble` and `MatCeedGetContextDoubles`? I used `Double` in the name to be clear that a cast was required to update these values

We already have `MatCeedSetContextDouble`. I'm fine having another function to handle the array case. The `Real` in this PRs function names is referring to it handling the `PetscReal -> double` casting implicitly, as I don't think any PETSc user would/should be working with raw `double` instead of `PetscReal/PetscScalar`.

jeremylt:
Ideally PETSc would be generating the CeedOperator and the user would never see it themselves, which is why I want to use the PETSc Jacobian parameter names.

A PetscReal can be a float, double, __float128, or half precision, which is why I want to communicate that the user exactly gets 64 bits and not 16, 32, or 128.

jrwrigh:
> Ideally PETSc would be generating the CeedOperator and the user would never see it themselves, which is why I want to use the PETSc Jacobian parameter names.

Ok, that makes sense. Though I'm still not sure how I feel about it being in the `MatCeed` namespace, since it's specific to Jacobians rather than a generic linear operator.

> A PetscReal can be a float, double, __float128, or half precision, which is why I want to communicate that the user exactly gets 64 bits and not 16, 32, or 128.

Right... which is why the cast is handled inside the function:
https://github.com/CEED/libCEED/pull/1575/files#diff-7acfd90d719666cfce7338f11bb54ed0efd1e436ccf34478881ee5a95f02d10eR724

Maybe our difference in perspective comes with whether the user will care about loss of precision for casting __float128 to double? Or are we talking past each other? I'm not saying replace `MatCeed{Get,Set}ContextDouble` with a `PetscReal` version, just another more convenient, generic interface.

jeremylt:
Ah, I guess we could keep all of these different options

jeremylt (approved):

PR #1577, opened 2024-05-05
fluids: Change Amat assemble flag name

jrwrigh:
Differentiate between whether we're telling the code to actively assemble the matrix or just that we want Amat to be a mat type that can be assembled. Flag is new enough that nobody has used it before, so no need to deprecate, just replace.

### DIFF
--- a/examples/fluids/src/petsc_ops.c
+++ b/examples/fluids/src/petsc_ops.c
@@ -308,7 +308,7 @@ PetscErrorCode ApplyAddCeedOperatorLocalToLocal(Vec X_loc, Vec Y_loc, OperatorAp
  * @param[out] Pmat     `Mat` to be used for the solver `Pmat`
  */
 PetscErrorCode CreateSolveOperatorsFromMatCeed(KSP ksp, Mat mat_ceed, PetscBool assemble, Mat *Amat, Mat *Pmat) {
-  PetscBool use_matceed_pmat, assemble_amat = PETSC_FALSE;
+  PetscBool use_matceed_pmat, use_assembled_amat = PETSC_FALSE;
 
   PetscFunctionBeginUser;
   {  // Determine if Amat should be MATCEED or assembled
@@ -316,11 +316,11 @@ PetscErrorCode CreateSolveOperatorsFromMatCeed(KSP ksp, Mat mat_ceed, PetscBool 
 
     PetscCall(KSPGetOptionsPrefix(ksp, &ksp_prefix));
     PetscOptionsBegin(PetscObjectComm((PetscObject)mat_ceed), ksp_prefix, "", NULL);
-    PetscCall(PetscOptionsBool("-matceed_assemble_amat", "Assemble the A matrix for KSP solve", NULL, assemble_amat, &assemble_amat, NULL));
+    PetscCall(PetscOptionsBool("-use_assembled_amat", "Assemble the A matrix for KSP solve", NULL, use_assembled_amat, &use_assembled_amat, NULL));
     PetscOptionsEnd();
   }
 
-  if (assemble_amat) {
+  if (use_assembled_amat) {
     PetscCall(MatCeedCreateMatCOO(mat_ceed, Amat));
     if (assemble) PetscCall(MatCeedAssembleCOO(mat_ceed, *Amat));
 


examples/fluids/src/petsc_ops.c
```diff
@@ -308,19 +308,19 @@ PetscErrorCode ApplyAddCeedOperatorLocalToLocal(Vec X_loc, Vec Y_loc, OperatorAp
...
+  PetscBool use_matceed_pmat, use_assembled_amat = PETSC_FALSE;
 
   PetscFunctionBeginUser;
   {  // Determine if Amat should be MATCEED or assembled
     const char *ksp_prefix = NULL;
 
     PetscCall(KSPGetOptionsPrefix(ksp, &ksp_prefix));
     PetscOptionsBegin(PetscObjectComm((PetscObject)mat_ceed), ksp_prefix, "", NULL);
-    PetscCall(PetscOptionsBool("-matceed_assemble_amat", "Assemble the A matrix for KSP solve", NULL, assemble_amat, &assemble_amat, NULL));
+    PetscCall(PetscOptionsBool("-use_assembled_amat", "Assemble the A matrix for KSP solve", NULL, use_assembled_amat, &use_assembled_amat, NULL));
```
jedbrown, line 319: This is stylistic, but I dislike `-use_` as a prefix. I think `-mat_ceed_amat_assemble` would be better, though still I dislike MATCEED knowing what an `amat` is. Indeed, it is not uniquely defined in the context of nested solves. For example, if you use `-pc_type ksp` then you'll have one matrix that is the pmat for the outer KSP and the amat for the inner KSP, unless you `-pc_use_amat`.

What we're doing here is injecting an extra layer of indirection because we want the decision of matrix representation to bind later than the choice of PC. That pattern doesn't exist in PETSc and the price we pay for it is making assumptions about the environment that are not true of what the object model can represent (and what a user may have good reason to do). This is not merely hypothetical as there is reason to believe that nesting KSP as above may be our fastest production solve configuration.

I think I've mentioned before that I have mixed feelings about the magical format-PC matching. Yes, it's better user experience for a user that doesn't want to think about these things, but it has caused breakage and implementation complexity (and made some options silently discarded, which has been a pain for debugging).

I think it would be a good exercise to explain this situation concisely and make a thread on PETSc discord so we can think creatively about how PETSc could better support such things without tangling the object graph.

jrwrigh: Appreciate the feedback!

> This is stylistic, but I dislike `-use_` as a prefix. I think `-mat_ceed_amat_assemble` would be better, though still I dislike MATCEED knowing what an `amat` is. Indeed, it is not uniquely defined in the context of nested solves. For example, if you use `-pc_type ksp` then you'll have one matrix that is the pmat for the outer KSP and the amat for the inner KSP, unless you `-pc_use_amat`.

The reason I wanted to go away from `-mat_ceed*` prefix is that, as you say, MATCEED has no concept of an `amat` vs `pmat`; it's just a `Mat`.

The reason I wanted to go to `use_assembled_amat` vs `amat_assemble` is to distinguish between performing the assembly and just using an assemble-able `Mat`. Though that's more an issue when reading the source code than on the command-line interface side of things.
 
> What we're doing here is injecting an extra layer of indirection because we want the decision of matrix representation to bind later than the choice of PC. That pattern doesn't exist in PETSc and the price we pay for it is making assumptions about the environment that are not true of what the object model can represent (and what a user may have good reason to do). This is not merely hypothetical as there is reason to believe that nesting KSP as above may be our fastest production solve configuration.
> 
> I think I've mentioned before that I have mixed feelings about the magical format-PC matching. Yes, it's better user experience for a user that doesn't want to think about these things, but it has caused breakage and implementation complexity (and made some options silently discarded, which has been a pain for debugging).

Correct me if I'm misunderstanding, but I think the root of the issue is that the function is conflating two different behavior modifications that the user wants to do; what preconditioner to use and the computational implementation used to execute it.

Unfortunately, these are tied together since not all preconditioners are compatible with MATCEED. If MATCEED could be used for all preconditioners (i.e. handling lazy assembly by itself), then MatType and PC selection would be independent choices, and the choice of MatType would be solely for computational implementation purposes. 

[This thread in the PETSc discord](https://discord.com/channels/1119324534303109172/1162157330624942173/1162157330624942173) kinda covers the wall that @jeremylt ran into awhile back with MATCEED PC compatibility. In a previous conversation, @jeremylt mentioned that there might be an added interface for `Mat` which would make this feasible, but I don't know the status or details there.

> I think it would be a good exercise to explain this situation concisely and make a thread on PETSc discord so we can think creatively about how PETSc could better support such things without tangling the object graph.

Assuming you agree with the above, should that discord thread focus on PC compatibility with generic MATSHELL? Perhaps @jeremylt has a better idea of what the extra interface would look like, but I'd imagine a `MATOP_GET_AIJ` would work, while then embedding that into the PC logic?

jeremylt: I don't have a good solution myself but I share Jed's concerns and have been trying to weaken this induced coupling in Ratel as well.

jeremylt: I think the biggest issue really is that GAMG will only work on AIJ matrices. Any other PC (I think) should probably work with a MatCEED fine, though it may require adding an extra MatOp that passes through to the inner AIJ. If we could expose a way to give access to our inner AIJ in a MatCEED to GAMG that would be huge for improving ergonomics and reducing our code complexity I think

jrwrigh: I thought there were issues with or PCs as well, like ILU? At least that was my take away from the PETSc discord thread I linked above. Specifically:

> Using a factor matrix seems to assert that it is permissible for the PC to use the raw internal data of the Mat without using any accessors,

jeremylt: Ugh, that's right. This is why I get grumpy when the API is not respected!

jrwrigh: Right. Is the GAMG limitation similar, in that it just needs to respect the MAT API? Or is there something inherent to GAMG that means it needs to use AIJ matrices?

jrwrigh: If the latter is the case (GAMG needs to use AIJ matrices), then do we need to add a `MATOP_SHELL_ASSEMBLE_AIJ` to the API?

jeremylt: GAMG only works on AIJ matrices, so it checks if it has an AIJ matrix so it can grab private Mat data

jeremylt: I don't think we need a special matop to assemble AIJ? I'm not sure what the best API would be exactly, but somehow granting access to the MatCEED's inner AIJ for full PC support would be lovely

jrwrigh: I meant use the assemble AIJ matop as a way to pass the inner AIJ for others to use.

jedbrown: "the Mat interface" is pretty squishy. `Mat` is a vessel for both the operator used by KSP (`MatMult`, possibly transpose) and for preconditioning data, as well as a cache for assembled data. To understand the "cache" if we only gave the KSP a `MatFactory` that held the mathematical essence for the operator and was able to provide matrix-free and assembled representations, you might end up concretizing the same matrix multiple times in the context of a nested solve. Never mind code, I don't have a good box-and-arrows concept of a good solution here, but I prefer explicit with low-assumptions over implicit with high-assumptions (which means they'll often be wrong).

jrwrigh: One other thing we need to go back to the "explicit with low-assumptions" for point-block diagonal assembly is a public interface for MatCEED to assemble the point-block diagonal representation. I know that Jeremy had that implemented at some point, but then I thought it wasn't necessary so we removed it.

jeremylt: I'm not sure why we need that public? We can wrap all the MatOps related to a PB PC I thought?

jrwrigh: Yes, that's why I thought it wasn't necessary at the time. But determining whether the we can get away with MatCeed vs a fully assembled Mat requires querying the PC for what type it is, which (if I understand right) is where Jed's discomfort comes from. 

That said, we could explicitly declare whether to pass the MatCeed representation or the fully-assembled representation independent of the PC choice, which accomplishes the same thing.

He's also mentioned previously wanting to assemble the PB "manually" and feed that into, for example, an LU PC to verify the PB diagonal inversion in PETSc, which he had run into issues with previously. Though that's more of a debugging idea than a proper feature.

PR #1361, opened 2023-10-03
fluids: Add external SGS DD inference capability

jrwrigh:
This adds the capability to use external libraries to do the inference of the SGS DD model. Currently, this is just a framework with the native libCEED inference method enabled. External libraries will be added in later.

More verbosely, this introduces to different ways to evaluate the data-driven SGS model: fused and sequential.

Fused is what was implemented before, where the input creation, model inference, and output handling were all done in a single CeedOperator. This requires a native libCEED inference implementation.

Sequential has separate function calls/CeedOperators for input creation, model inference, and output handling. This allows for functions calling external libraries to be used for the model inference step.

Closes #1326

PR #1581, opened 2024-05-09
fluids: Add PyTorch external DD SGS evaluation

jrwrigh:
Add PyTorch as a external DD SGS evaluation. This is a "follow up" to #1361 where we use PyTorch to run the data-driven model instead of a native implementation.

ToDos:
- [x] Add in `PetscLogEvents` (particularly for inference and the data transfer steps)
- [x] Add documentation
- [x] Cleanup the Makefile additions (remove the automatic `USE_LIBTORCH` testing)
- [x] Add weak symbols so that this can be compiled without pytorch
- [x] Add command line switch to enable pytorch vs internal sequential
- [x] Move createPyTorchModel into the testing directory (it's only useful for documenting the creation of the testing model (instead of it just being a binary blob).
- [x] Rename "sequential_internal" to "sequential_ceed"


examples/fluids/pytorch_pkgconfig.py
jrwrigh, line 1: To explain the reasoning for this:

CMake is the only supported build system for PyTorch/libTorch. I initially tried to write a small CMake script to just get the necessary compiler flags, but this ended up not working because normal pip-installed versions of PyTorch have libraries built with CUDA and CMake _requires_ that CUDA be available for those libraries even though it's completely unnecessary if you're not using the CUDA backends of PyTorch.

To circumvent this, the PyTorch python package gives some useful information to build against the underlying libTorch (include and library directories mostly). 

However, it still doesn't give required compiler flags, namely `GLIBCXX_USE_CXXX11_ABI`. To get around that, I parse the `TorchConfig.cmake` file which has that information embedded in it. Rather than break out my grep-fu, I decided it'd be easiest to do it all in Python since I'm using Python for the include and library paths (and for where `TorchConfig.cmake` is located).

The easiest way I could think to transfer all this information to the `Makefile` *while* keeping the number of Python startups (which can take about 5 seconds per startup depending on the system in my experience) to a minimum is to write it to a pkgconfig file from Python that the make file then uses. Yeah, this relies on some POSIX compliant file system behavior, but I figured that's the case for most build environments.

Any who, thanks for attending my TED talk/rant.

jrwrigh:
This is good for review now. Tests pass on Noether for CPU and CUDA. I'm having some difficulties getting PyTorch+ROCM to build with Spack, but that's not a huge priority since we don't have hardware to run that on right now.

tests/junit.py
```diff
@@ -28,13 +28,17 @@ def create_argparser() -> argparse.ArgumentParser:
...
+    parser.add_argument('--use_torch', type=bool, default=False, help='Whether to build with torch')
     parser.add_argument('test', help='Test executable', nargs='?')
 
     return parser
 
 
 # Necessary functions for running tests
 class CeedSuiteSpec(SuiteSpec):
+    def __init__(self, use_torch: bool):
+        self.use_torch: bool = use_torch
```
jeremylt, line 40: Minor consistency - in Ratel we'd say `has_torch` to indicate the env has Torch available

jeremylt (approved):
This all seems reasonable to me

PR #1583, opened 2024-05-14
Fix docs for `CeedElemRestrictionGetNumPoints`

zatkins-dev:
Updates the wording to clarify that the output `num_points` is the number of points in the offsets array

### DIFF
--- a/interface/ceed-elemrestriction.c
+++ b/interface/ceed-elemrestriction.c
@@ -1490,10 +1490,10 @@ int CeedElemRestrictionGetElementSize(CeedElemRestriction rstr, CeedInt *elem_si
 
 /**
 
-  @brief Get the number of points in the l-vector for a points `CeedElemRestriction`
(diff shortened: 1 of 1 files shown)


interface/ceed-elemrestriction.c
```diff
@@ -1490,10 +1490,10 @@ int CeedElemRestrictionGetElementSize(CeedElemRestriction rstr, CeedInt *elem_si
 
 /**
 
-  @brief Get the number of points in the l-vector for a points `CeedElemRestriction`
+  @brief Get the number of points in the offsets array for a points `CeedElemRestriction`
```
jeremylt, line 1493: Maybe 'number of points being restricted to' or 'number of points is the E-vector' ?

Does the constructor need similar updated language?

zatkins-dev: the constructor uses the "offsets array" language

jeremylt: fair enough

PR #1584, opened 2024-05-14
Fix gpu AtPoints conversion to standard indexing

jeremylt:
@zatkins-dev herein lies my confusion and mistake


zatkins-dev:
Ah, that would do it!

PR #1588, opened 2024-05-22
backends/cuda: NVRTC compile to CUBIN when supported (resolve #1587)

### LINKED ISSUE #1587: NVRTC compile directly to SASS/CUBIN
When using a CUDA toolkit that is newer than the driver, one will get `CUDA_ERROR_UNSUPPORTED_PTX_VERSION` when trying to load the PTX produced by NVRTC. Starting in CUDA-11.1, NVRTC can compile directly to SASS, in which case you replace calls to `nvrtcGetPTXSize` and `nvrtcGetPTX` with `nvrtcGetCUBINSize` and `nvrtcGetCUBIN` respectively. This should reduce pain, especially on machines like Polaris that are extremely slow about updating their drivers.

https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#dynamic-code-generation

jedbrown:
This allows using a newer CUDA runtime with an older driver, and seems to have no downsides.

>  NVRTC can generate cubins directly starting with CUDA 11.1. [...]
  NVRTC used to support only virtual architectures through the option
  -arch, since it was only emitting PTX. It will now support actual
  architectures as well to emit SASS. The interface is augmented to
  retrieve either the PTX or cubin if an actual architecture is
  specified.

https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#dynamic-code-generation

### DIFF
--- a/backends/cuda/ceed-cuda-compile.cpp
+++ b/backends/cuda/ceed-cuda-compile.cpp
@@ -80,9 +80,19 @@ int CeedCompile_Cuda(Ceed ceed, const char *source, CUmodule *module, const Ceed
   opts[0] = "-default-device";
   CeedCallBackend(CeedGetData(ceed, &ceed_data));
   CeedCallCuda(ceed, cudaGetDeviceProperties(&prop, ceed_data->device_id));
-  std::string arch_arg = "-arch=compute_" + std::to_string(prop.major) + std::to_string(prop.minor);
-  opts[1]              = arch_arg.c_str();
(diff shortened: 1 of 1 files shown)


jrwrigh (approved):
Nice! Have you had a chance to verify this on Polaris?

jedbrown:
I haven't, but I did test locally with an upgraded CUDA 12.5 and the current Linux driver 550.78, which only supports CUDA 12.4 (and thus fails the old way, works this new way).

PR #1590, opened 2024-05-23
Update t531

### LINKED ISSUE #1589: Incorrect indexing for qd in Test 531
The symmetric components of the qw/det(J) Adj(J) * Ajd(J)^T operation are not being stored in the correct indices as described in the storage format comment of line 23 of [t531-operator.h](https://github.com/CEED/libCEED/blob/ef9a992f4cf09f2be4ec72f649495c67ec03f813/tests/t531-operator.h). Currently also affects the values being calculated in dv = qd * du. Indexing of qd of lines 31 and 32 need to be switched.  

@jrwrigh, as per our discussion, not sure if this also effects output checks in [t531-operator.c](https://github.com/CEED/libCEED/blob/ef9a992f4cf09f2be4ec72f649495c67ec03f813/tests/t531-operator.c) .

jeremylt:
Fixes #1589 

The operator is now a true Poisson operator. Also, the grid is no longer rectangular and the vector u is non-constant.

### DIFF
--- a/tests/t531-operator-f.f90
+++ b/tests/t531-operator-f.f90
@@ -15,7 +15,7 @@
       integer bx,bu
       integer qf_setup,qf_diff,qf_diff_lin
       integer op_setup,op_diff,op_diff_lin
-      integer qdata,x,a,u,v
+      integer qdata,x,a,u,v,v_lin
       integer nelem,p,q,d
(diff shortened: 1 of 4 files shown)


jrwrigh (approved):
Thanks!

PR #1602, opened 2024-06-10
solids - reduce to minimal example set, see Ratel for full

jeremylt:
Reduce solids example now that Ratel is mature


rezgarshakeri:
We used those NH qfunctions that are deleted here in this [paper](https://asmedigitalcollection.asme.org/IMECE/proceedings-abstract/IMECE2021/V012T12A027/1133236) to see which one is a better strategy for saving variable in residual. I don't know if we are going to reproduce the results in the future.

rezgarshakeri:
BTW, I am okay with deleting them.

jeremylt:
Yeah, I thought about that. We can always resurrect these by going back to the v0.12 release

rezgarshakeri:
Cool, so I guess since we don't have any current configuration we can delete [ this section](https://libceed.org/en/latest/examples/solids/#hyperelasticity-in-current-configuration) from doc.

jeremylt:
Good point, thanks

rezgarshakeri:
and this part too https://libceed.org/en/latest/examples/solids/#jacobian-representation

rezgarshakeri:
Do we need Mooney-Rivlin model?

jeremylt:
Need probably not, but its one of the three current test cases, so I kept it since its not much code and its a cool pic in the docs. Should probably also ditch small strain since its not tested

rezgarshakeri:
Sounds good. I can add hyperelastic small strain in Ratel if we need it.

jeremylt:
I don't know that we have a use case past helping fill out the diagram in the docs :shrug:

rezgarshakeri (approved):
Thanks!

PR #1605, opened 2024-06-11
fluids: Add IDL to RHSFunction_Newtonian

jrwrigh:
This adds the ability to use the internal damping layer for explicit formulations.

Below is a validation run of the implementation of the gaussian wave problem:
image

The IDL is setup for the left quarter of the domain. We see the wave is damped out in the implicit and explicit versions to a similar (if not identical) way. The right most figure is implicit without IDL to serve as reference for the damping that IDL is doing.

- [x] Modify existing explicit Gaussian wave test to test IDL as well.


KennethEJansen (approved):
Looks like what I tested and it seems from the Gaussian Wave test it is working.

PR #1606, opened 2024-06-11
fluids: Add dyScale and dz to STG options

jrwrigh:
This adds two options to the STG inflow:
- `-stg_dy_scale` to scale the element size in the y direction
- `-stg_dz` to explicitly set the nodal spacing in the z direction

Note @KennethEJansen original implementation used `-stg_dyScale`, so I've kept this available, but as a deprecated option.

- [x] Document new options

### DIFF
--- a/examples/fluids/problems/stg_shur14.c
+++ b/examples/fluids/problems/stg_shur14.c
@@ -221,7 +221,7 @@ PetscErrorCode SetupStg(const MPI_Comm comm, const DM dm, ProblemData problem, U
   char                     stg_inflow_path[PETSC_MAX_PATH_LEN] = "./STGInflow.dat";
   char                     stg_rand_path[PETSC_MAX_PATH_LEN]   = "./STGRand.dat";
   PetscBool                mean_only = PETSC_FALSE, use_stgstrong = PETSC_FALSE, use_fluctuating_IC = PETSC_FALSE, given_stg_dx = PETSC_FALSE;
-  CeedScalar               u0 = 0.0, alpha = 1.01, stg_dx = 1.0e-3;
+  CeedScalar               u0 = 0.0, alpha = 1.01, stg_dx = -1, stg_h_scale_factor = 1 / user->app_ctx->degree;
   CeedQFunctionContext     stg_context;
   NewtonianIdealGasContext newtonian_ig_ctx;
 
@@ -235,7 +235,11 @@ PetscErrorCode SetupStg(const MPI_Comm comm, const DM dm, ProblemData problem, U
   PetscCall(PetscOptionsBool("-stg_strong", "Enforce STG inflow strongly", NULL, use_stgstrong, &use_stgstrong, NULL));
   PetscCall(PetscOptionsBool("-stg_fluctuating_IC", "\"Extrude\" the fluctuations through the domain as an initial condition", NULL,
                              use_fluctuating_IC, &use_fluctuating_IC, NULL));
-  PetscCall(PetscOptionsReal("-stg_dx", "Element size in streamwise direction at inflow", NULL, stg_dx, &stg_dx, &given_stg_dx));
+  PetscCall(PetscOptionsReal("-stg_dx", "Element length in x direction at inflow", NULL, stg_dx, &stg_dx, &given_stg_dx));
+  PetscCall(PetscOptionsReal("-stg_h_scale_factor", "Scale element size for cutoff frequency calculation", NULL, stg_h_scale_factor,
+                             &stg_h_scale_factor, NULL));
+  PetscCall(PetscOptionsDeprecated("-stg_dyScale", NULL, "libCEED 0.12.0", "Use -stg_h_scale_factor to scale all the element dimensions"));
+  PetscCall(PetscOptionsDeprecated("-stg_dz", NULL, "libCEED 0.12.0", NULL));
   PetscOptionsEnd();
 
   PetscCall(PetscCalloc1(1, &global_stg_ctx));
@@ -247,6 +251,7 @@ PetscErrorCode SetupStg(const MPI_Comm comm, const DM dm, ProblemData problem, U
   global_stg_ctx->use_fluctuating_IC = use_fluctuating_IC;
   global_stg_ctx->theta0             = theta0;
   global_stg_ctx->P0                 = P0;
+  global_stg_ctx->h_scale_factor     = stg_h_scale_factor;
 
   {  // Calculate dx assuming constant spacing
     PetscReal domain_min[3], domain_max[3], domain_size[3];
@@ -256,6 +261,8 @@ PetscErrorCode SetupStg(const MPI_Comm comm, const DM dm, ProblemData problem, U
     PetscInt nmax = 3, faces[3];
     PetscCall(PetscOptionsGetIntArray(NULL, NULL, "-dm_plex_box_faces", faces, &nmax, NULL));
     global_stg_ctx->dx = given_stg_dx ? stg_dx : domain_size[0] / faces[0];
+    PetscCheck((global_stg_ctx->dx > 0) && PetscIsNormalReal((PetscReal)global_stg_ctx->dx), comm, PETSC_ERR_LIB,
+               "STG dx must be positive normal number, got %g", global_stg_ctx->dx);
   }
 
   PetscCallCeed(ceed, CeedQFunctionContextGetData(problem->apply_vol_rhs.qfunction_context, CEED_MEM_HOST, &newtonian_ig_ctx));
--- a/examples/fluids/qfunctions/stg_shur14.h
+++ b/examples/fluids/qfunctions/stg_shur14.h
@@ -102,12 +102,12 @@ CEED_QFUNCTION_HELPER CeedScalar Calc_qn(const CeedScalar kappa, const CeedScala
 }
 
 // Calculate hmax, ke, keta, and kcut
-CEED_QFUNCTION_HELPER void SpectrumConstants(const CeedScalar wall_dist, const CeedScalar eps, const CeedScalar lt, const CeedScalar h[3],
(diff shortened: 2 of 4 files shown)


examples/fluids/problems/stg_shur14.c
```diff
@@ -255,7 +260,9 @@ PetscErrorCode SetupStg(const MPI_Comm comm, const DM dm, ProblemData problem, U
 
     PetscInt nmax = 3, faces[3];
     PetscCall(PetscOptionsGetIntArray(NULL, NULL, "-dm_plex_box_faces", faces, &nmax, NULL));
-    global_stg_ctx->dx = given_stg_dx ? stg_dx : domain_size[0] / faces[0];
+    global_stg_ctx->dx      = given_stg_dx ? stg_dx : domain_size[0] / faces[0];
+    global_stg_ctx->dyScale = given_stg_dyScale ? stg_dyScale : 1;
+    global_stg_ctx->dz      = given_stg_dz ? stg_dz : domain_size[2] / faces[2];
```
jrwrigh, line 265: @KennethEJansen Any qualms with the default being to setting the dz spacing based on the element size rather than `domain_size[2] / faces[2]`, which assumes a PETSc box mesh?

KennethEJansen: The issue is that the size SHOULD NOT be determined solely by the mesh for a higher order method.  I switched to this form  so that I had the ability to scale the element "measure" for polynomial order.

We had discussed a default scaling of 1/p

jrwrigh: Right, but using the domain_size and faces is worse than just going based on the element size (while also being the same).

Can we:
- Remove `-stg_dy_scale`
- Add `-stg_element_scale` to affect all the directions equally
- Have `-stg_dx` and `-stg_dz` represent actual element size (rather than nodal distance)

> the element "measure"

I was thinking about what we want this quantity to be and it's really just the inverse of the Nyquist frequency for the element, which is exactly equal to $h/p$. So a $1/p$ scale factor default really does make the most sense I think.

KennethEJansen: I don't know that I believe STG  should always be h/p.  While I am the vertex spacing can't be right for all p. I am further not sure that it should the same for all directions but certainly that is an improvement and possibly the best that can be done so......

I agree to the first 2 of your bullets but don't follow the third.  What is the actual element size to you?
I don't follow the fourth bullet either

The point after "measure" quote seems to argue against letting the user  scale, forcing it to be 1/p which I don't agree with.

Can we say,
Currently the code can and does compute element lengths in the wall normal and span direction from the in-plane metric tensor. 
We will refer to these as stg_d_wn and stg_d_sw  These are element sizes (akin to corner vertex separations or bounding box dimensions. 
At some later date we expect to be able to compute the plane-normal element size stg_d_pn.
Until that time, the user MUST provide stg_d_pn as a physical length of elements near the inflow.
User is free to provide a scaling  for those THREE element lengths.  If no scaling is provided 1/p will be a default.


I am ok leaving _wn=y,  _sw=z, and _pn=x for now  if it touches too much code to change quickly, but we really should not have assumed a coordinate system for the inflow.

jrwrigh: Your suggestion is what I was trying to explain. I can get on that.

> I am ok leaving _wn=y, _sw=z, and _pn=x for now if it touches too much code to change quickly, but we really should not have assumed a coordinate system for the inflow.

We have those kind of bad-but-ok coordinate system assumptions all over the place right now (IDL, spanwise statistics). A wall-oriented coordinate system is better and would be pretty easily implemented, but we need distance-to-wall to do that generically. I'm not sure if you had something else in mind.

jrwrigh: Updated now.

examples/fluids/qfunctions/stg_shur14.h
```diff
@@ -118,18 +118,18 @@ CEED_QFUNCTION_HELPER void SpectrumConstants(const CeedScalar wall_dist, const C
  * @param[in]  wall_dist Distance to the nearest wall
  * @param[in]  eps       Turbulent dissipation w/rt wall_dist
  * @param[in]  lt        Turbulent length scale w/rt wall_dist
- * @param[in]  h         Element lengths in coordinate directions
+ * @param[in]  hNodSep   Element lengths in coordinate directions
```
jeremylt, line 121: Convention usually is words with underscores instead of camel case

examples/fluids/qfunctions/stg_shur14.h
```diff
@@ -186,23 +186,24 @@ CEED_QFUNCTION_HELPER void StgShur14Calc(const CeedScalar X[3], const CeedScalar
  * @param[in]  ubar      Mean velocity at X
  * @param[in]  cij       Cholesky decomposition at X
  * @param[in]  Ektot     Total spectrum energy at this location
- * @param[in]  h         Element size in 3 directions
+ * @param[in]  hNodSep   Element size in 3 directions
```
jeremylt, line 189: and so on here

KennethEJansen:
The code changes look right (at least from a read through).  I am not sure how to read/review if/how the tests have been updated. Can you comment on that @jrwrigh  ? If that has been, we could try to test one of the flat plates that I had been using those developments on Sunspot for a more convincing test that these changes reproduce the capabilities I had in my branch because that is really more convincing than "the code looks right".

examples/fluids/problems/stg_shur14.c
jedbrown: If we consider the metric tensor on the STG inflow plane, its eigenvalues are a coordinate-system-independent length scale for the element. Should we use that by default?

examples/fluids/problems/stg_shur14.c
```diff
@@ -221,7 +221,7 @@ PetscErrorCode SetupStg(const MPI_Comm comm, const DM dm, ProblemData problem, U
   char                     stg_inflow_path[PETSC_MAX_PATH_LEN] = "./STGInflow.dat";
   char                     stg_rand_path[PETSC_MAX_PATH_LEN]   = "./STGRand.dat";
   PetscBool                mean_only = PETSC_FALSE, use_stgstrong = PETSC_FALSE, use_fluctuating_IC = PETSC_FALSE, given_stg_dx = PETSC_FALSE;
-  CeedScalar               u0 = 0.0, alpha = 1.01, stg_dx = 1.0e-3;
+  CeedScalar               u0 = 0.0, alpha = 1.01, stg_dx = 1.0e-3, stg_h_scale_factor = 1 / user->app_ctx->degree;
```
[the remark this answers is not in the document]
jrwrigh: So we query whether the `-stg_dx` flag as been used and replace the value with `domain_size[0] / face[0]` if it's not. So that `stg_dx` magic value is never used.

That being said, I do see the danger, but I'm not sure how to address it. I can set the value to something else (`0`, `-1`), but it's still a silent error; the code will still run, just incorrectly.

examples/fluids/problems/stg_shur14.c
jrwrigh: That is the current default for the wall-parallel directions, but we can't get that for the boundary-normal direction, which requires wall-normal gradients at the boundary for the metric tensor. I'm currently working on copying that over from Ratel, so hopefully that'll be done soon :tm:. But for right now the element length in that direction needs to be given explicitly.

examples/fluids/problems/stg_shur14.c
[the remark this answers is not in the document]
jedbrown: Can we just not initialize `stg_dx` so we get a compiler warning if it isn't set? Or use `-1` as a placeholder and check somewhere that it has indeed been set? Just looking for something less error-prone.

examples/fluids/problems/stg_shur14.c
jedbrown: I think you mean "requires (STG) boundary-normal gradients", but some time ago, we discussed using the spanwise (longest length scale in STG boundary plane) and multiplying by a factor (2.5 could be default, which matches the 6-15 or 12-30) to get the STG boundary-normal length scale.

examples/fluids/problems/stg_shur14.c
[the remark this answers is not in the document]
jrwrigh: Sure, I've added a `PetscCheck` now to verify the size of `stg_dx`.

jedbrown (approved):

examples/fluids/problems/stg_shur14.c
jrwrigh: Yeah, I meant boundary normal.

I had a PR for the dx scaling factor in #1415, but Ken went with the more direct way for his branch. Regardless of which method (directly specifying $\Delta x$ or setting it as a scale of $\Delta z$), they should be superseded once the boundary-normal gradient in merged in.

examples/fluids/problems/stg_shur14.c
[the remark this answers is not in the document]
KennethEJansen: Agreed and that is my bad.  I guess the two options I can think of (until we get 3D metric computation to directly compute) are:
1. error out if user does not set
2. use 1.5 times larger of the two in-plane metrics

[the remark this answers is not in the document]
jrwrigh: You're fine; pretty sure I'm the one who put that there.

The issue with 1 is that the default behavior has been to just go off of the box-mesh related quantities.
The issue with 2 is that it's an assumption (usually a good one, but an assumption none-the-less).

Note that this will error out if STG is not run on a PETSc-generated box mesh and `-stg_dx` isn't set. (the `-dm_plex_box_faces` default value is all zeros, so the calculated `stg_dx` will be infinitty and trigger the `PetscIsNormalReal` part of the `PetscCheck`.

PR #1607, opened 2024-06-12
GPU consistency

jeremylt:
Just a couple of consistency fixes I noted while working on AtPoints

### DIFF
--- a/include/ceed/jit-source/cuda/cuda-ref-basis-tensor.h
+++ b/include/ceed/jit-source/cuda/cuda-ref-basis-tensor.h
@@ -17,7 +17,7 @@
 //------------------------------------------------------------------------------
 // Interp
 //------------------------------------------------------------------------------
-extern "C" __global__ void Interp(const CeedInt num_elem, const CeedInt transpose, const CeedScalar *__restrict__ interp_1d,
+extern "C" __global__ void Interp(const CeedInt num_elem, const CeedInt is_transpose, const CeedScalar *__restrict__ interp_1d,
                                   const CeedScalar *__restrict__ u, CeedScalar *__restrict__ v) {
   const CeedInt i = threadIdx.x;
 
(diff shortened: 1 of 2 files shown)


include/ceed/jit-source/cuda/cuda-ref-basis-tensor.h
```diff
@@ -47,14 +47,11 @@ extern "C" __global__ void Interp(const CeedInt num_elem, const CeedInt transpos
...
 
-      for (CeedInt k = i; k < u_size; k += blockDim.x) {
-        s_buffer_1[k] = cur_u[k];
-      }
       for (CeedInt d = 0; d < BASIS_DIM; d++) {
         __syncthreads();
         // Update buffers used
         pre /= P;
-        const CeedScalar *in       = d % 2 ? s_buffer_2 : s_buffer_1;
+        const CeedScalar *in       = d == 0 ? cur_u : (d % 2 ? s_buffer_2 : s_buffer_1);
```
jrwrigh, line 54: This looks like a change in behavior? Mostly in that I'm not sure what `s_buffer_1` does now since you removed the for loop earlier that sets it to a value.

include/ceed/jit-source/hip/hip-ref-basis-tensor.h
```diff
@@ -47,14 +47,11 @@ extern "C" __global__ void Interp(const CeedInt num_elem, const CeedInt transpos
...
 
-      for (CeedInt k = i; k < u_size; k += blockDim.x) {
-        s_buffer_1[k] = cur_u[k];
-      }
       for (CeedInt d = 0; d < BASIS_DIM; d++) {
         __syncthreads();
         // Update buffers used
         pre /= P;
-        const CeedScalar *in       = d % 2 ? s_buffer_2 : s_buffer_1;
+        const CeedScalar *in       = d == 0 ? cur_u : (d % 2 ? s_buffer_2 : s_buffer_1);
```
jrwrigh, line 54: Same here with regard to the apparent change in behavior.

jrwrigh (approved):

include/ceed/jit-source/cuda/cuda-ref-basis-tensor.h
jeremylt: s_buffer_1 and s_buffer_2 are the intermediate places to store the calculations during the tensor contractions. On main, Interp copies part of the input array into s_buffer_1 before starting. But if you look lower in this file, for Grad, that kernel reads the input directly instead of copying into a buffer. I don't see a reason for Interp not to do the same.

jrwrigh: Ok, I see that Grad has the same structure. Is `s_buffer_1` set to something outside of this function? I feel iffy about having it in the expression without knowing if it has anything at all.

jeremylt: The loop writes from 'in' to 'out' and those two are always the opposite buffer or u/v, so valid data is always read

PR #1597, opened 2024-05-31
GPU BasisApplyAtPoints

jeremylt:
~~Very WIP, blocked by #1595~~

I'm puzzling my way through exactly how the parallelism works.

Edit: Ready to rock

PR #1610, opened 2024-06-20
AtPoints CeedOperator for GPU

jeremylt:
~~Added for CUDA, HIP next~~

Working for CUDA locally, HIP running in CI with the same changes


jeremylt:
Good news: The code in this branch is fine

Bad news: I need to use way less memory for #1597

PR #1611, opened 2024-06-21
basis: Allow CreateProjection for mixed-tensor bases

jrwrigh:
Also added an additional case to `t319` to test the case of mixed-tensor bases.

### DIFF
--- a/tests/t319-basis.c
+++ b/tests/t319-basis.c
@@ -150,9 +150,9 @@ int main(int argc, char **argv) {
 
     VerifyProjectedBasis(basis_project, dim, p_to_dim, p_from_dim, x_to, x_from, u_to, u_from, du_to);
 
-    // Test projection on non-tensor bases
+    // Create non-tensor bases
+    CeedBasis basis_from_nontensor, basis_to_nontensor;
     {
-      CeedBasis         basis_from_nontensor, basis_to_nontensor;
       CeedElemTopology  topo;
       CeedInt           num_comp, num_nodes, nqpts;
       const CeedScalar *interp, *grad;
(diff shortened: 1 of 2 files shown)


tests/t319-basis.c
```diff
@@ -82,66 +148,30 @@ int main(int argc, char **argv) {
...
+      const CeedScalar *interp, *grad, *q_ref, *q_weights;
+
+      CeedBasisGetTopology(basis_from, &topo);
+      CeedBasisGetNumComponents(basis_from, &num_comp);
+      CeedBasisGetNumNodes(basis_from, &num_nodes);
+      CeedBasisGetNumQuadraturePoints(basis_from, &nqpts);
+      CeedBasisGetInterp(basis_from, &interp);
+      CeedBasisGetGrad(basis_from, &grad);
+      CeedBasisGetQRef(basis_from, &q_ref);
+      CeedBasisGetQWeights(basis_from, &q_weights);
```
jrwrigh, line 167: So `CeedBasisInterp` has a tensor-product 1D version for it, but `CeedBasisQref` and `CeedBasisGetQWeights`. Is there any reason why we don't have full vs 1D versions of the quadrature reference data?

jeremylt: Is there a reason? No.

If I'm remembering correctly those are BS for the new projection Basis and we should be using [1...] for the weights and [0...] for the pts

jrwrigh: I was about to go ahead and implement a `CeedBasisQRef1D` and `CeedBasisQWeights1D`, but just realized that would actually break the current interface for `CeedBasisQRef` and `CeedBasisQWeights`. Like, if you call those for tensor basis now, you'll get the 1D representation. But the new version would change it's output. I don't like breaking code, but this seems like it was almost broken to begin with (or at least misleading).

jeremylt: Yeah, I'd leave this alone for now, at least until we can discuss and see what we want to do

But I think updating the new q_ref and q_weight to not be lies would be great

jrwrigh: Yeah, I can correct the projection basis. 

I'll made something in Zulip to discuss. It just makes the test case more annoying to setup. haha

jrwrigh: Oh, it's q_ref and q_weights are optional inputs to `CeedBasisCreateH1`. So not a pain at all.

jrwrigh: Looks like `q_ref` and `q_weight` is already being set to zero (via calloc). Do we want to change the behavior there?

I feel like setting them be `NULL` would also make sense.

jeremylt: Yeah, since they are optional inputs for `CeedBasisCreateH1` and `CeedBasisCreateTensorH1` we really should use `NULL` there since those aren't valid for this Basis

jrwrigh: Done.

tests/t319-basis.c
```diff
@@ -4,6 +4,7 @@
 #include <ceed.h>
 #include <math.h>
 #include <stdio.h>
+#include <stdlib.h>
```
jeremylt, line 7: What's coming in from stdlib?

jrwrigh: Oh, a malloc that I got rid of.

jeremylt:
Ope, you'll have to first add a commit to HIP and CUDA backends (probably also SYCL and MAGMA) checking that the q_ref and q_weight arrays are valid before trying to allocate and copy them over to the device. (I'd error if they are unset and CEED_EVAL_WEIGHTS is used).

At this point, this might be 2 PRs 1) fix backends to accommodate q_weight possibly being un-set 2) update to allow non-tensor + tensor projection basis

jrwrigh:
Ok, I'll just revert the last commit on this branch, then this should be good to go independently.

jeremylt:
You'd think, but your fix to t319 also is swept up in that minor backend bug

jeremylt:
Yea
```
Test: t319-basis 
  $ build/t319-basis /gpu/cuda/ref
ERROR: returncode = 254
Output: 
NO MESSAGE
FAIL: stderr
Output: 
/var/lib/gitlab-runner/builds/N8GCsqus/0/libceed/libCEED/backends/cuda-ref/ceed-cuda-ref-basis.c:277 in CeedBasisCreateH1_Cuda(): CUDA_ERROR_INVALID_VALUE
```
CI is unhappy

jrwrigh:
Yeah, saw that. working on the fix.

jrwrigh:
Alright, #1613 is merged, so this should be working now, :crossed_fingers:

jeremylt (approved):
As long as CI is happy, this should be good

Thanks for chasing the lose ends

PR #1612, opened 2024-06-21
op - add terse view

jeremylt:


interface/ceed-operator.c
```diff
@@ -1469,16 +1469,16 @@ int CeedOperatorSetName(CeedOperator op, const char *name) {
...
-  @param[in] op     `CeedOperator` to view
+  @param[in] op     `CeedOperator` to view brief summary
   @param[in] stream Stream to write; typically `stdout` or a file
 
   @return Error code: 0 - success, otherwise - failure
 
   @ref User
 **/
-int CeedOperatorView(CeedOperator op, FILE *stream) {
+int CeedOperatorView_Core(CeedOperator op, FILE *stream, bool is_full) {
```
jrwrigh, line 1481: I feel like reversing the boolean to `is_terse` makes more sense since the associated function is `CeedOperatorTerseView`.

interface/ceed-operator.c
```diff
@@ -1492,16 +1492,46 @@ int CeedOperatorView(CeedOperator op, FILE *stream) {
...
+  @brief View a brief summary `CeedOperator`
+
+  @param[in] op     `CeedOperator` to view brief summary
+  @param[in] stream Stream to write; typically `stdout` or a file
+
+  @return Error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedOperatorTerseView(CeedOperator op, FILE *stream) {
```
jrwrigh, line 1530: Maybe `CeedOperatorViewTerse` is more inline with naming conventions? Since this is a special case of the `View` function?

jrwrigh (approved):
Minor suggestions. Thanks!

interface/ceed-operator.c
jrwrigh: Eh, I just saw that `CeedOperatorSingleView` exists, so maybe what you have is more inline with naming conventions.

interface/ceed-operator.c
jeremylt: I did not do that on purpose so the conditionals below say `if (true) doThing` instead of `if (!true) doThing`

interface/ceed-operator.c
jeremylt: Sure

interface/ceed-operator.c
jrwrigh: Ah, fair enough.

PR #1613, opened 2024-06-21
basis: Correct backend handling of NULL q_weight

jrwrigh:
Also have `CeedBasisCreateProjection` set q_ref, q_weight to NULL.

Spun off from #1611


jeremylt (approved):
Looks good to me, modulo `make style`'s unhappiness

jrwrigh:
Tests pass on Sunspot, so the SYCL implementation seems good to go. I'll wait on the CUDA and HIP test though just to make sure everything's good though.

PR #1614, opened 2024-06-24
Strided Vec Utils

jeremylt:
Adding `CeedVectorSetValueStrided` and `CeedVectorCopyStrided` to help with device AtPoints diagonal assembly. Threw in some small fixes I noticed at the same time.


interface/ceed-vector.c
```diff
@@ -238,6 +242,50 @@ int CeedVectorCopy(CeedVector vec, CeedVector vec_copy) {
...
+  @param[in]     vec      `CeedVector` to copy
+  @param[in]     start    First index to copy
+  @param[in]     step     Stride between indices to copy
+  @param[in,out] vec_copy `CeedVector` to copy values to
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedVectorCopyStrided(CeedVector vec, CeedSize start, CeedInt step, CeedVector vec_copy) {
```
jrwrigh, line 257: Do we want to check that the length of `vec` is evenly divisible by `step`? Might be too constrictive, but seems a useful debugging feature to me at least.

interface/ceed-vector.c
```diff
@@ -288,6 +336,7 @@ int CeedVectorSetValue(CeedVector vec, CeedScalar value) {
 
   if (vec->SetValue) {
     CeedCall(vec->SetValue(vec, value));
+    vec->state += 2;
```
jrwrigh, line 339: Out of curiosity, why is `state` incremented by 2? Is the first bit reserved for something? (read/write access to the array)?

jrwrigh: Nevermind, I see that's exactly what it's used for in `CeedVectorSetValueStrided`. haha

interface/ceed-vector.c
jrwrigh: Looks like this check would break the test you added (length of 10, but step of 3).

Are there times when we would expect the length to not be evenly divisible by `step` when using these strided routines?

backends/hip-ref/kernels/hip-ref-vector.hip.cpp
```diff
@@ -21,7 +21,7 @@ __global__ static void setValueK(CeedScalar *__restrict__ vec, CeedSize size, Ce
 //------------------------------------------------------------------------------
 // Set value on device memory
 //------------------------------------------------------------------------------
-extern "C" int CeedDeviceSetValue_Hip(CeedScalar *d_array, CeedSize length, CeedScalar val) {
+extern "C" int CeedDeviceSetValueStrided_Hip(CeedScalar *d_array, CeedSize length, CeedScalar val) {
```
jrwrigh, line 24:
```suggestion
extern "C" int CeedDeviceSetValue_Hip(CeedScalar *d_array, CeedSize length, CeedScalar val) {
```

backends/hip-ref/ceed-hip-ref-vector.c
```diff
@@ -270,6 +270,41 @@ static int CeedVectorSetValue_Hip(CeedVector vec, CeedScalar val) {
...
+//------------------------------------------------------------------------------
+// Set host array to value strided
+//------------------------------------------------------------------------------
+static int CeedHostSetValueStrided_Hip(CeedScalar *h_array, CeedSize start, CeedSize step, CeedSize length, CeedScalar val) {
+  for (CeedSize i = start; i < length; i += step) h_array[i] = val;
+  return CEED_ERROR_SUCCESS;
+}
+
+//------------------------------------------------------------------------------
+// Set device array to value strided (impl in .cu file)
```
jrwrigh, line 282:
```suggestion
// Set device array to value strided (impl in .hip.cpp file)
```

jrwrigh (approved):
Minor questions/fixes

interface/ceed-vector.c
jeremylt: I don't think we need to care that the length of the vec is a multiple of the step?

interface/ceed-vector.c
jeremylt: An odd state indicates write access has been granted. Jumping by 2 means write access was granted and restored so array contents must be assumed to be different.

PR #1616, opened 2024-06-24
tidy - add check tidy needs

jeremylt:
These weren't caught in CI because that job has stopped failing on `make tidy` errors for some unknown reason


jeremylt:
I think `make tidy` CI job didn't fail because tidy didn't have a proposed fix

jrwrigh (approved):

Issue #1618, opened 2024-06-25
Memory efficiency when assembling diagonal of a linear operator


sebastiangrimberg:
If I understand the code correctly, if I assemble the diagonal using `CeedOperatorLinearAssembleAddDiagonal`, the assembled `QFunction` is kept in memory as part of the `CeedQFunctionAssemblyData` and not destroyed until the operator is destroyed. If I don't plan on assembling the QFunction for the operator again, it might be nice to be able to release this memory. I think I have observed this being a problem when assembling diagonals for several different operators and keeping the operators around as well.

First, is my understanding correct, and second, is there a recommended way to free this memory? Should I just call `CeedQFunctionAssemblyDataDestroy` of `op->qf_assembled`? I suppose I could also call `CeedOperatorAssemblyDataDestroy` on `op->op_assembled`, though I think this data structure is not as large in memory. Thanks for your help!

jedbrown:
I think it would make sense to have a `CeedOperator` method to free any data that can be rebuilt. Verb could be Compress or Strip?

This could be a flag to eagerly destroy temporary data structures (rather than persist them, which is the default), though if you assemble the diagonal and then assemble a point-block diagonal (or do full assembly), the inner assembly would be done twice.

sebastiangrimberg:
That sounds like a good idea. I think a good place to start is just adding the method for the user to call and having a flag can be a later step. In general I would expect the user should be aware of the implications of destroying the temporary data structures causing duplicated work if they need them to be recreated later on. One place this might not be so clear is `CeedOperatorMultigridLevelCreate`, where if you were to free the assembled `QFunction` data before the call you would assemble the `QFunction` twice when getting diagonals for each multigrid level, while you could avoid this if waiting until later to free the temporary data. But again, I think that's fine to put on the user to discern and the default behavior will be to keep everything around in memory as it is now.

sebastiangrimberg:
See #1619. Open to renaming or any other adjustments, not sure if there is other temporary data I missed.

PR #1619, opened 2024-06-26
Allow user to free temporary memory associated with QFunction and operator assembly

sebastiangrimberg:
Add `CeedOperatorGetQFunctionAssemblyData` to support `CeedOperatorAssemblyDataStrip` which calls `CeedQFunctionAssemblyDataDestroy` and `CeedOperatorAssemblyDataDestroy` to free up temporary operator assembly memory.

Resolves https://github.com/CEED/libCEED/issues/1618

### DIFF
--- a/include/ceed/backend.h
+++ b/include/ceed/backend.h
@@ -393,6 +393,8 @@ CEED_EXTERN int  CeedQFunctionContextReference(CeedQFunctionContext ctx);
 
 CEED_EXTERN int CeedOperatorGetBasisPointer(CeedBasis basis, CeedEvalMode eval_mode, const CeedScalar *identity, const CeedScalar **basis_ptr);
 CEED_EXTERN int CeedOperatorCreateActivePointBlockRestriction(CeedElemRestriction rstr, CeedElemRestriction *pointblock_rstr);
+
+CEED_EXTERN int CeedOperatorGetQFunctionAssemblyData(CeedOperator op, CeedQFunctionAssemblyData *data);
 CEED_EXTERN int CeedQFunctionAssemblyDataCreate(Ceed ceed, CeedQFunctionAssemblyData *data);
 CEED_EXTERN int CeedQFunctionAssemblyDataReference(CeedQFunctionAssemblyData data);
 CEED_EXTERN int CeedQFunctionAssemblyDataSetReuse(CeedQFunctionAssemblyData data, bool reuse_assembly_data);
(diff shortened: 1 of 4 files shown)


include/ceed/backend.h
```diff
@@ -393,6 +393,8 @@ CEED_EXTERN int  CeedQFunctionContextReference(CeedQFunctionContext ctx);
 
 CEED_EXTERN int CeedOperatorGetBasisPointer(CeedBasis basis, CeedEvalMode eval_mode, const CeedScalar *identity, const CeedScalar **basis_ptr);
 CEED_EXTERN int CeedOperatorCreateActivePointBlockRestriction(CeedElemRestriction rstr, CeedElemRestriction *pointblock_rstr);
+
+CEED_EXTERN int CeedOperatorGetQFunctionAssemblyData(CeedOperator op, CeedQFunctionAssemblyData *data);
```
jeremylt, line 397: Note - I try to keep the source file and header file in the same order, so this would prompt moving some functions around

interface/ceed-operator.c
```diff
@@ -2106,6 +2110,35 @@ int CeedOperatorApplyAdd(CeedOperator op, CeedVector in, CeedVector out, CeedReq
...
+/**
+  @brief Destroy temporary assembly data associated with a `CeedOperator`
+
+  @param[in,out] op `CeedOperator` whose assembly data to destroy
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedOperatorAssemblyDataStrip(CeedOperator op) {
```
jeremylt, line 2122: Can we call this in `CeedOperatorDestroy` to make sure it does what we think it does?

include/ceed/backend.h
sebastiangrimberg: I think actually this is consistent with the definition order in `ceed-preconditioning.c`. But if I am wrong let me know!

interface/ceed-operator.c
sebastiangrimberg: Done in [f883f0a](https://github.com/CEED/libCEED/pull/1619/commits/f883f0a5423dccfdefcc45bb5c497717e44013fc). One nuance here is that in `CeedOperatorAssemblyDataStrip` I am destroying assembly for all sub-operators of a composite operator which is useful for external users. This just means we need to call it before destroying the sub-operators.

include/ceed/backend.h
```diff
@@ -404,6 +406,7 @@ CEED_EXTERN int CeedQFunctionAssemblyDataSetObjects(CeedQFunctionAssemblyData da
 CEED_EXTERN int CeedQFunctionAssemblyDataGetObjects(CeedQFunctionAssemblyData data, CeedVector *vec, CeedElemRestriction *rstr);
 CEED_EXTERN int CeedQFunctionAssemblyDataDestroy(CeedQFunctionAssemblyData *data);
 
+CEED_EXTERN int CeedOperatorGetOperatorAssemblyData(CeedOperator op, CeedOperatorAssemblyData *data);
```
jeremylt, line 409: I meant this one, sorry. You moved the function definition

sebastiangrimberg: Yeah this is now consistent with `ceed-preconditioning.c` where we define `GetOperatorAssemblyData` right before `OperatorAssemblyDataCreate`.

interface/ceed-operator.c
```diff
@@ -2184,16 +2186,21 @@ int CeedOperatorDestroy(CeedOperator *op) {
...
   }
-  // AtPoints data
+  CeedCall(CeedFree(&(*op)->input_fields));
+  CeedCall(CeedFree(&(*op)->output_fields));
+  // Destroy AtPoints data
   CeedCall(CeedVectorDestroy(&(*op)->point_coords));
   CeedCall(CeedElemRestrictionDestroy(&(*op)->rstr_points));
   CeedCall(CeedElemRestrictionDestroy(&(*op)->first_points_rstr));
+  // Destroy assembly data
+  CeedCall(CeedOperatorAssemblyDataStrip(*op));
```
jeremylt, line 2196: can we note here that this has to come before destroying sub ops?

jeremylt (approved):

interface/ceed-operator.c
sebastiangrimberg: Done in [c6b536a](https://github.com/CEED/libCEED/pull/1619/commits/c6b536a8adc1c8ea763ec6cf1b342657e84d2699)

PR #1623, opened 2024-06-27
fluids - drop SGS, SmartSim, Torch (moved to HONEE)

jeremylt:
Now that this is up and running in HONEE, we can trim some time out of the libCEED testing pipeline


jeremylt:
@jrwrigh feel free to push to this branch to fixup any mistakes I made

jrwrigh:
I've kinda changed it to drop all SGS modeling. That's something you might've done accidentally, but is probably a good idea as we'd almost assuredly be cutting that out anyways. Plus is makes it easy to check with a simple `rg sgs **/*.{c,h}` command. haha

jrwrigh (approved):
Assuming Noether is happy, this looks fine by me.

PR #1637, opened 2024-07-15
GPU Gen Reorganize

jeremylt:
Starting to reorganize GPU Gen JiT string creation so we can more easily enable non-tensor and AtPoints.

This isn't the final form that will enable non-tensor and AtPoints, but this reorganization gives me a good starting point to move forward from.

Plan is squash+merge.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -64,26 +64,26 @@ static int CeedOperatorBuildKernelFieldData_Cuda_gen(std::ostringstream &code, C
     case CEED_EVAL_INTERP:
       data->B.inputs[i] = basis_data->d_interp_1d;
       code << "  __shared__ CeedScalar s_B" << var_suffix << "[" << P_1d * Q_1d << "];\n";
-      code << "  loadMatrix<" << P_name << "," << Q_name << ">(data, B.inputs[" << i << "], s_B" << var_suffix << ");\n";
+      code << "  loadMatrix<" << P_name << ", " << Q_name << ">(data, B.inputs[" << i << "], s_B" << var_suffix << ");\n";
       break;
     case CEED_EVAL_GRAD:
       data->B.inputs[i] = basis_data->d_interp_1d;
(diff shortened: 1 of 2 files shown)


jrwrigh:
It'd probably be nice to migrate these changes to SYCL, but given lack of easy testing access that could be a separate PR.

jeremylt:
Yeah, these changes are really easy to get wrong so I'd want to add them to SYCL separately

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -23,358 +23,227 @@
...
-
-  // Get QFunction source and name
-  string qfunction_source(qf_data->qfunction_source);
-  string qfunction_name(qf_data->qfunction_name);
-  string operator_name;
-  operator_name = "CeedKernelCudaGenOperator_" + qfunction_name;
-
+static int CeedOperatorBuildKernelData_Cuda_gen(Ceed ceed, CeedInt num_input_fields, CeedOperatorField *op_input_fields,
+                                                CeedQFunctionField *qf_input_fields, CeedInt num_output_fields, CeedOperatorField *op_output_fields,
+                                                CeedQFunctionField *qf_output_fields, CeedInt *max_P_1d, CeedInt *Q_1d, CeedInt *dim, bool *is_tensor,
```
jrwrigh, line 30: Maybe a bit nitpicky (and something I'm inconsistent on, but trying to get better at):

It'd be nice to have the parameters that are meant to be arrays as `type_t paramter[]` instead of `type_t *paramter` to help distinguish the output scalars from input/output arrays. This could also be solved with a docstring, but I don't think we want to be docstringing every single internal function.

jrwrigh (approved):
Really not familiar with the original code, but nothing looks egregious.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
jeremylt: This isn't the current style, so I'll leave it for this PR and open an issue to see if we want to do a pass to standardize to this

jeremylt:
My plan is to do SYCL in a follow-up since that will require more work

PR #1638, opened 2024-07-16
gpu - minimum input/output array size of 1

jeremylt:
Minor bug for QFunctions with no inputs/outputs, like `SetPointFields_MPM_NeoHookean`


jrwrigh (approved):
Is there not an equivalent in the SYCL backend?

jeremylt:
Ugh, I always forget about SYCL

jrwrigh:
Easy to do when there's no way for us to test it easily.

PR #1643, opened 2024-08-05
Minor performance improvement for op with repeat input rstr

jeremylt:
This PR skips the duplicate restriction if two fields share the same input vector and restriction. I'd like to do this for output as well, but that's more involved. This should offer a very mild performance improvement for the linear damage model in Ratel.

### DIFF
--- a/backends/blocked/ceed-blocked-operator.c
+++ b/backends/blocked/ceed-blocked-operator.c
@@ -16,7 +16,7 @@
 //------------------------------------------------------------------------------
 // Setup Input/Output Fields
 //------------------------------------------------------------------------------
-static int CeedOperatorSetupFields_Blocked(CeedQFunction qf, CeedOperator op, bool is_input, const CeedInt block_size,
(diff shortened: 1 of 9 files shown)


jrwrigh (approved):

jrwrigh:
I has a similar thought the other day, but for composite operators. Do they share input E-Vecs that have the same restriction and vector, or are they recalculated as well?

jeremylt:
I think its much less common that composite operators will share e-vecs in many cases. At least in our typical usage of composite operators I've seen, we then to use them for different regions of the mesh, like faces vs volumes

jrwrigh:
I was thinking about possibly including one of the SGS models as a sub-operator rather than a separate function call like it is right now, in which case there'd be a duplication of vectors there. But fair, it's pretty rare otherwise.

jeremylt:
I'm planning on waiting unit Nother is back up to merge, but then I want to do the same thing on the output side once I create `CeedBasisApplyAdd`

PR #1644, opened 2024-08-06
Add CeedBasisApplyAdd

jeremylt:
WIP, GPU impl and CeedBasisApplyAddAtPoints to come

### DIFF
--- a/backends/magma/ceed-magma-basis.c
+++ b/backends/magma/ceed-magma-basis.c
@@ -26,7 +26,8 @@
 //------------------------------------------------------------------------------
 // Basis apply - tensor
 //------------------------------------------------------------------------------
-static int CeedBasisApply_Magma(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode, CeedEvalMode e_mode, CeedVector u, CeedVector v) {
+static int CeedBasisApplyCore_Magma(CeedBasis basis, bool apply_add, CeedInt num_elem, CeedTransposeMode t_mode, CeedEvalMode e_mode, CeedVector u,
+                                    CeedVector v) {
   Ceed              ceed;
   Ceed_Magma       *data;
   CeedInt           dim, num_comp, num_nodes, P_1d, Q_1d, P, Q;
@@ -52,7 +53,8 @@ static int CeedBasisApply_Magma(CeedBasis basis, CeedInt num_elem, CeedTranspose
   // Read vectors
   if (u != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArrayRead(u, CEED_MEM_DEVICE, &d_u));
   else CeedCheck(e_mode == CEED_EVAL_WEIGHT, ceed, CEED_ERROR_BACKEND, "An input vector is required for this CeedEvalMode");
-  CeedCallBackend(CeedVectorGetArrayWrite(v, CEED_MEM_DEVICE, &d_v));
+  if (apply_add) CeedCallBackend(CeedVectorGetArray(v, CEED_MEM_DEVICE, &d_v));
+  else CeedCallBackend(CeedVectorGetArrayWrite(v, CEED_MEM_DEVICE, &d_v));
 
   // Apply basis operation
   switch (e_mode) {
@@ -115,7 +117,8 @@ static int CeedBasisApply_Magma(CeedBasis basis, CeedInt num_elem, CeedTranspose
       void   *args[] = {&impl->d_interp_1d, &d_u, &u_elem_stride, &u_comp_stride, &d_v, &v_elem_stride, &v_comp_stride, &num_elem};
 
       if (t_mode == CEED_TRANSPOSE) {
-        CeedCallBackend(CeedRunKernelDimSharedMagma(ceed, impl->InterpTranspose, grid, num_threads, num_t_col, 1, shared_mem, args));
+        CeedCallBackend(CeedRunKernelDimSharedMagma(ceed, apply_add ? impl->InterpTransposeAdd : impl->InterpTranspose, grid, num_threads, num_t_col,
+                                                    1, shared_mem, args));
       } else {
         CeedCallBackend(CeedRunKernelDimSharedMagma(ceed, impl->Interp, grid, num_threads, num_t_col, 1, shared_mem, args));
       }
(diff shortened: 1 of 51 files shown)


jeremylt:
Ok, design decision point - the usage really only makes sense for `CEED_TRANSPOSE` especially for why I need this, so I made that an explicit requirement.

I think for GPU I think I'll want separate Apply vs ApplyTranspose kernels in some cases?

jrwrigh:
> the usage really only makes sense for `CEED_TRANSPOSE` especially for why I need this, so I made that an explicit requirement.

Agreed that having the transpose as an input, but verifying it to always be transpose works well. Might also include that in the docstring for it, ie:

```
  @param[in]  t_mode    @ref CEED_TRANSPOSE to apply the transpose, mapping from quadrature points to nodes. CEED_NOTRANSPOSE is not valid for CeedBasisApplyAdd.
  ```

**Comment by jeremylt** [comment]
Update - CUDA/HIP ref and shared are good, need to tackle MAGMA

**Comment by nbeams** [comment]
I'm not able to follow libCEED activity as closely these days, but please ping me if you run into any issues with the MAGMA backend.

**Comment by jeremylt** [comment]
Thanks! The TLDR here is that I want to sum into the target vector with CeedBasisApplyAdd for CEED_TRANSPOSE for CEED_EVAL_INTERP and CEED_EVAL_GRAD. It looks not too tricky for MAGMA at a glance but I haven't dug in on it

**Comment by jeremylt** [comment]
Only tensor product grad is correctly summing into the target vector with my changes for MAGMA, unclear why but I'll dig more Monday

File: backends/magma/ceed-magma-basis.c [hunk 1]
```diff
@@ -248,6 +252,16 @@ static int CeedBasisApply_Magma(CeedBasis basis, CeedInt num_elem, CeedTranspose
   return CEED_ERROR_SUCCESS;
 }
 
+static int CeedBasisApply_Magma(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode, CeedEvalMode e_mode, CeedVector u, CeedVector v) {
+  CeedCallBackend(CeedBasisApplyCore_Magma(basis, false, num_elem, t_mode, e_mode, u, v));
+  return CEED_ERROR_SUCCESS;
+}
+
+static int CeedBasisApplyAdd_Magma(CeedBasis basis, CeedInt num_elem, CeedTransposeMode t_mode, CeedEvalMode e_mode, CeedVector u, CeedVector v) {
+  CeedCallBackend(CeedBasisApplyCore_Magma(basis, false, num_elem, t_mode, e_mode, u, v));
```
nbeams, line 261: copy-paste error, I'd guess -- second parameter should be true?

backends/magma/ceed-magma-basis.c
```diff
@@ -388,15 +405,15 @@ static int CeedBasisApplyNonTensor_Magma(CeedBasis basis, CeedInt num_elem, Ceed
...
+          Kernel = apply_add ? impl->InterpTransposeAdd[iN] : impl->InterpTranspose[iN];
           NB     = impl->NB_interp_t[iN];
         } else {
           Kernel = impl->Interp[iN];
           NB     = impl->NB_interp[iN];
         }
       } else {
         if (t_mode == CEED_TRANSPOSE) {
-          Kernel = impl->DerivTranspose[iN];
+          Kernel = apply_add ? impl->DerivTranspose[iN] : impl->DerivTranspose[iN];
```
nbeams, line 416: Needs `DerivTransposeAdd` for first option

backends/magma/ceed-magma-basis.c
```diff
@@ -414,7 +431,7 @@ static int CeedBasisApplyNonTensor_Magma(CeedBasis basis, CeedInt num_elem, Ceed
     } else {
       for (CeedInt d = 0; d < q_comp; d++) {
         if (t_mode == CEED_TRANSPOSE) {
-          const CeedScalar beta = (d > 0) ? 1.0 : 0.0;
+          const CeedScalar beta = apply_add || ((d > 0) ? 1.0 : 0.0);
```
nbeams, line 434: Wouldn't this be cleaner if we did `(apply_add || (d > 0)) ? 1.0 : 0.0` ? To avoid having a scalar value on one side of the `||`?

nbeams (commented):
Did a quick review, found some copy-paste bugs

include/ceed/jit-source/magma/magma-common-tensor.h
```diff
@@ -178,6 +207,23 @@ static __device__ __inline__ void write_V_3d(T *dV, const int compstride, T rV[D
...
+// dV is assumed to point directly to the element (i.e. already offset by elem-stride)
+// register is assumed to be rV[DIM_V][NUM_COMP][rV_SIZE]
+// i_DIM specifies which dimension is being written to in dV
+// rV_SIZE can be different from P (e.g. max(P, Q))
+template <typename T, int Q, int DIM_V, int NUM_COMP, int rV_SIZE, int i_DIM>
+static __device__ __inline__ void sum_V_3d(T *dV, const int compstride, T rV[DIM_V][NUM_COMP][rV_SIZE], const int tx) {
+  if (tx < (Q * Q)) {
+    for (int comp = 0; comp < NUM_COMP; comp++) {
+      for (int j = 0; j < Q; j++) {
+        dV[comp * compstride + j * (Q * Q) + tx] = rV[i_DIM][comp][j];
```
nbeams, line 221: Somehow missed getting this one in my review. Need `+=`.

backends/magma/ceed-magma-basis.c
jeremylt: ugh, that's an embarrassing oops

backends/magma/ceed-magma-basis.c
jeremylt: That one too

include/ceed/jit-source/magma/magma-common-tensor.h
jeremylt: Ok, these fixes fixed it all. You are l33t, as the kids say (I think... do they still say that?). Thank you for catching those fixes!

nbeams: Ha, I think copy-paste bugs have this weird thing where they are much easier to spot when you are not the one who did the copy-pasting in the first place. Maybe a quantum phenomenon, where they exist in a state that is both camouflaged and not camouflaged at the same time until observed ;)

nbeams:
As a side note, looking at the new tests in this PR reminded me that we might want to check on the test coverage for the non-tensor basis in MAGMA. Now that we have two options -- the RTC kernels for lower orders and standard library calls for the rest -- I'm not sure how often, if at all, the "standard" route is being tested. (For t363 I manually lowered `MAGMA_NONTENSOR_CUSTOM_KERNEL_MAX_P`/`MAGMA_NONTENSOR_CUSTOM_KERNEL_MAX_Q` to make sure it also passed with the non-custom-kernel path -- and it did -- but of course that won't happen in the usual testing).

zatkins-dev (approved):
Overall I like this refactor. I noticed a few places where it is probably more clear to index an array instead of doing pointer arithmetic, but if you disagree that's fine.

PR #1646, opened 2024-08-15
Must Destroy OperatorField Objects

### LINKED ISSUE #1639: Create `CeedOperatorFieldGet*Read` routines with `const` parameters
I think on the existing `CeedOperatorFieldGet*` probably aren't appropriate for "normal" uses by end-users; Those objects shouldn't be editable (or at least Destroyable), but the lack of a `const` qualifier on the parameters means you can. See [HONEE MR](https://gitlab.com/phypid/honee/-/merge_requests/27#note_2014079011) for example of accidentally Destroying something gotten from `CeedOperatorFieldGet*`.

Adding `const` qualifiers would break existing code unfortunately, so assuming that's a non-starter, maybe we could implement "read" versions of the Getters?
i.e.:
```c
CEED_EXTERN int CeedOperatorFieldGetElemRestrictionRead(CeedOperatorField op_field, const CeedElemRestriction *rstr);
CEED_EXTERN int CeedOperatorFieldGetBasisRead(CeedOperatorField op_field, const CeedBasis *basis);
CEED_EXTERN int CeedOperatorFieldGetVectorRead(CeedOperatorField op_field, const CeedVector *vec);
CEED_EXTERN int CeedOperatorFieldGetDataRead(CeedOperatorField op_field, const char **field_name, const CeedElemRestriction *rstr, const CeedBasis *basis,
                                         const CeedVector *vec);
```

jeremylt:
Fixes #1639 

What do we think @jrwrigh. You can see how this does get a bit invasive.

---

Ok, now the Vector/ElemRestriction/Basis retrieved from the OperatorField needs to be Destroyed per the corresponding interface.

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -50,6 +50,7 @@ static int CeedOperatorSetupFields_Ref(CeedQFunction qf, CeedOperator op, bool i
     if (eval_mode != CEED_EVAL_WEIGHT) {
       CeedCallBackend(CeedOperatorFieldGetElemRestriction(op_fields[i], &elem_rstr));
       CeedCallBackend(CeedElemRestrictionCreateVector(elem_rstr, NULL, &e_vecs_full[i + start_e]));
+      CeedCallBackend(CeedElemRestrictionDestroy(&elem_rstr));
     }
 
     switch (eval_mode) {
@@ -70,12 +71,14 @@ static int CeedOperatorSetupFields_Ref(CeedQFunction qf, CeedOperator op, bool i
         CeedCallBackend(CeedVectorCreate(ceed, e_size, &e_vecs[i]));
         q_size = (CeedSize)Q * size;
         CeedCallBackend(CeedVectorCreate(ceed, q_size, &q_vecs[i]));
+        CeedCallBackend(CeedBasisDestroy(&basis));
         break;
       case CEED_EVAL_WEIGHT:  // Only on input fields
         CeedCallBackend(CeedOperatorFieldGetBasis(op_fields[i], &basis));
         q_size = (CeedSize)Q;
         CeedCallBackend(CeedVectorCreate(ceed, q_size, &q_vecs[i]));
         CeedCallBackend(CeedBasisApply(basis, 1, CEED_NOTRANSPOSE, CEED_EVAL_WEIGHT, CEED_VECTOR_NONE, q_vecs[i]));
+        CeedCallBackend(CeedBasisDestroy(&basis));
         break;
     }
   }
@@ -98,7 +101,11 @@ static int CeedOperatorSetupFields_Ref(CeedQFunction qf, CeedOperator op, bool i
           CeedCallBackend(CeedVectorReferenceCopy(e_vecs_full[i + start_e], &e_vecs_full[j + start_e]));
           skip_rstr[j] = true;
         }
+        CeedCallBackend(CeedVectorDestroy(&vec_j));
+        CeedCallBackend(CeedElemRestrictionDestroy(&rstr_j));
       }
+      CeedCallBackend(CeedVectorDestroy(&vec_i));
+      CeedCallBackend(CeedElemRestrictionDestroy(&rstr_i));
     }
   } else {
     for (CeedInt i = num_fields - 1; i >= 0; i--) {
@@ -120,7 +127,11 @@ static int CeedOperatorSetupFields_Ref(CeedQFunction qf, CeedOperator op, bool i
           apply_add_basis[i]    = true;
           e_data_out_indices[j] = i;
         }
+        CeedCallBackend(CeedVectorDestroy(&vec_j));
+        CeedCallBackend(CeedElemRestrictionDestroy(&rstr_j));
       }
+      CeedCallBackend(CeedVectorDestroy(&vec_i));
+      CeedCallBackend(CeedElemRestrictionDestroy(&rstr_i));
     }
   }
   return CEED_ERROR_SUCCESS;
@@ -198,14 +209,15 @@ static inline int CeedOperatorSetupInputs_Ref(CeedInt num_input_fields, CeedQFun
                                               CeedVector in_vec, const bool skip_active, CeedScalar *e_data_full[2 * CEED_FIELD_MAX],
                                               CeedOperator_Ref *impl, CeedRequest *request) {
   for (CeedInt i = 0; i < num_input_fields; i++) {
-    uint64_t            state;
-    CeedEvalMode        eval_mode;
-    CeedVector          vec;
-    CeedElemRestriction elem_rstr;
+    bool         is_active;
+    uint64_t     state;
+    CeedEvalMode eval_mode;
+    CeedVector   vec;
 
     // Get input vector
     CeedCallBackend(CeedOperatorFieldGetVector(op_input_fields[i], &vec));
-    if (vec == CEED_VECTOR_ACTIVE) {
+    is_active = vec == CEED_VECTOR_ACTIVE;
+    if (is_active) {
       if (skip_active) continue;
       else vec = in_vec;
     }
@@ -218,13 +230,17 @@ static inline int CeedOperatorSetupInputs_Ref(CeedInt num_input_fields, CeedQFun
       CeedCallBackend(CeedVectorGetState(vec, &state));
       // Skip restriction if input is unchanged
       if ((state != impl->input_states[i] || vec == in_vec) && !impl->skip_rstr_in[i]) {
+        CeedElemRestriction elem_rstr;
+
         CeedCallBackend(CeedOperatorFieldGetElemRestriction(op_input_fields[i], &elem_rstr));
         CeedCallBackend(CeedElemRestrictionApply(elem_rstr, CEED_NOTRANSPOSE, vec, impl->e_vecs_full[i], request));
+        CeedCallBackend(CeedElemRestrictionDestroy(&elem_rstr));
       }
       impl->input_states[i] = state;
       // Get evec
       CeedCallBackend(CeedVectorGetArrayRead(impl->e_vecs_full[i], CEED_MEM_HOST, (const CeedScalar **)&e_data_full[i]));
     }
+    if (!is_active) CeedCallBackend(CeedVectorDestroy(&vec));
   }
   return CEED_ERROR_SUCCESS;
 }
@@ -243,14 +259,18 @@ static inline int CeedOperatorInputBasis_Ref(CeedInt e, CeedInt Q, CeedQFunction
 
     // Skip active input
     if (skip_active) {
+      bool       is_active;
       CeedVector vec;
 
       CeedCallBackend(CeedOperatorFieldGetVector(op_input_fields[i], &vec));
-      if (vec == CEED_VECTOR_ACTIVE) continue;
+      is_active = vec == CEED_VECTOR_ACTIVE;
+      CeedCallBackend(CeedVectorDestroy(&vec));
+      if (is_active) continue;
     }
     // Get elem_size, eval_mode, size
     CeedCallBackend(CeedOperatorFieldGetElemRestriction(op_input_fields[i], &elem_rstr));
     CeedCallBackend(CeedElemRestrictionGetElementSize(elem_rstr, &elem_size));
+    CeedCallBackend(CeedElemRestrictionDestroy(&elem_rstr));
     CeedCallBackend(CeedQFunctionFieldGetEvalMode(qf_input_fields[i], &eval_mode));
     CeedCallBackend(CeedQFunctionFieldGetSize(qf_input_fields[i], &size));
     // Basis action
@@ -266,6 +286,7 @@ static inline int CeedOperatorInputBasis_Ref(CeedInt e, CeedInt Q, CeedQFunction
         CeedCallBackend(CeedBasisGetNumComponents(basis, &num_comp));
         CeedCallBackend(CeedVectorSetArray(impl->e_vecs_in[i], CEED_MEM_HOST, CEED_USE_POINTER, &e_data_full[i][(CeedSize)e * elem_size * num_comp]));
         CeedCallBackend(CeedBasisApply(basis, 1, CEED_NOTRANSPOSE, eval_mode, impl->e_vecs_in[i], impl->q_vecs_in[i]));
+        CeedCallBackend(CeedBasisDestroy(&basis));
         break;
       case CEED_EVAL_WEIGHT:
         break;  // No action
@@ -289,6 +310,7 @@ static inline int CeedOperatorOutputBasis_Ref(CeedInt e, CeedInt Q, CeedQFunctio
     // Get elem_size, eval_mode
     CeedCallBackend(CeedOperatorFieldGetElemRestriction(op_output_fields[i], &elem_rstr));
     CeedCallBackend(CeedElemRestrictionGetElementSize(elem_rstr, &elem_size));
+    CeedCallBackend(CeedElemRestrictionDestroy(&elem_rstr));
     CeedCallBackend(CeedQFunctionFieldGetEvalMode(qf_output_fields[i], &eval_mode));
     // Basis action
     switch (eval_mode) {
@@ -307,6 +329,7 @@ static inline int CeedOperatorOutputBasis_Ref(CeedInt e, CeedInt Q, CeedQFunctio
         } else {
           CeedCallBackend(CeedBasisApply(basis, 1, CEED_TRANSPOSE, eval_mode, impl->q_vecs_out[i], impl->e_vecs_out[i]));
         }
+        CeedCallBackend(CeedBasisDestroy(&basis));
         break;
       // LCOV_EXCL_START
       case CEED_EVAL_WEIGHT: {
@@ -328,10 +351,13 @@ static inline int CeedOperatorRestoreInputs_Ref(CeedInt num_input_fields, CeedQF
 
     // Skip active inputs
     if (skip_active) {
+      bool       is_active;
       CeedVector vec;
 
       CeedCallBackend(CeedOperatorFieldGetVector(op_input_fields[i], &vec));
-      if (vec == CEED_VECTOR_ACTIVE) continue;
+      is_active = vec == CEED_VECTOR_ACTIVE;
+      CeedCallBackend(CeedVectorDestroy(&vec));
+      if (is_active) continue;
     }
     // Restore input
     CeedCallBackend(CeedQFunctionFieldGetEvalMode(qf_input_fields[i], &eval_mode));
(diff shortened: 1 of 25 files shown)


jeremylt:
ToDo: Track down all the CPU memory leaks, add GPU restores, language wrapers

backends/ref/ceed-ref-operator.c
```diff
@@ -172,16 +180,21 @@ static inline int CeedOperatorSetupInputs_Ref(CeedInt num_input_fields, CeedQFun
...
-    if (vec == CEED_VECTOR_ACTIVE) {
-      if (skip_active) continue;
-      else vec = in_vec;
+    is_active = vec == CEED_VECTOR_ACTIVE;
+    if (is_active) {
+      if (skip_active) {
+        CeedCallBackend(CeedOperatorFieldRestoreVector(op_input_fields[i], &vec));
+        continue;
+      } else {
+        vec = in_vec;
```
jrwrigh, line 196: Continuing this function with `vec = CEED_VECTOR_ACTIVE` seems odd... Like, there are calls to `CeedVectorGetState()`, which is probably valid. But I'm surprised the call to `CeedElemRestrictionApply()` works for `CEED_VECTOR_ACTIVE`

backends/ref/ceed-ref-operator.c
```diff
@@ -1296,18 +1366,19 @@ static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Ref(CeedOperator op, Ce
 
         // -- Grab diagonal value
         for (CeedInt j = 0; j < num_output_fields; j++) {
-          bool                is_active_output = false;
-          CeedInt             elem_size        = 0;
+          bool                is_active = false;
```
jrwrigh, line 1369: Any reason to initialize this `is_active` and not the others?

examples/fluids/src/setuplibceed.c
```diff
@@ -47,6 +47,21 @@ static PetscErrorCode CreateKSPMassOperator_Unstabilized(User user, CeedOperator
...
+
+    PetscCallCeed(ceed, CeedCompositeOperatorGetSubList(user->op_rhs_ctx->op, &sub_ops));
+    PetscCallCeed(ceed, CeedOperatorGetFieldByName(sub_ops[sub_op_index], "q", &field));
+    PetscCallCeed(ceed, CeedOperatorFieldRestoreElemRestriction(field, &elem_restr_q));
+    PetscCallCeed(ceed, CeedOperatorFieldRestoreBasis(field, &basis_q));
+
+    PetscCallCeed(ceed, CeedOperatorGetFieldByName(sub_ops[sub_op_index], "qdata", &field));
+    PetscCallCeed(ceed, CeedOperatorFieldRestoreElemRestriction(field, &elem_restr_qd_i));
+    PetscCallCeed(ceed, CeedOperatorFieldRestoreVector(field, &q_data));
+  }
```
jrwrigh, line 63: Maybe I'm being shortsighted, but is there reason to not just require that the given restrictions, basis, and vec just be destroyed rather than restored?

In most of the backend cases it's not a problem, but seeing this code here, it does make me wonder if just forcing the user to destroy the resulting objects would be cleaner.

interface/ceed-preconditioning.c
```diff
@@ -860,20 +862,32 @@ static int CeedSingleOperatorMultigridLevel(CeedOperator op_fine, CeedVector p_m
...
+    } else {
+      CeedElemRestriction rstr_field;
+      CeedBasis           basis_field;
+
+      CeedCall(CeedOperatorFieldGetElemRestriction(input_fields[i], &rstr_field));
+      CeedCall(CeedElemRestrictionReferenceCopy(rstr_field, &rstr));
+      CeedCall(CeedOperatorFieldRestoreElemRestriction(input_fields[i], &rstr_field));
+      CeedCall(CeedOperatorFieldGetBasis(input_fields[i], &basis_field));
+      CeedCall(CeedBasisReferenceCopy(basis_field, &basis));
+      CeedCall(CeedOperatorFieldRestoreBasis(input_fields[i], &basis_field));
```
jrwrigh, line 885: Use `CeedOperatorFieldGetData()`? Similar for usage further down the file.

jrwrigh (commented):
General question is whether the extra check during the restore is worth the new API, or just force the users to call `*Destroy` on the objects obtained from `CeedOperatorFieldGet*`.

I'm not sure what bugs the extra `CeedCheck(*vec == op_field->vec, ...)` would catch.

See [later comment](https://github.com/CEED/libCEED/pull/1646#discussion_r1720109745) for the fluids code that drove me to the above question.

examples/fluids/src/setuplibceed.c
jeremylt: I think we'd need to call it something other than Get in that case since it isn't the Get/Restore pattern.

I think Get/Restore makes the Rust interface work easier, but I haven't finished exploring that yet

jedbrown: In the `let y = x.get_y()` pattern, it's easier if `Drop` for `y` does not need a reference to `x`, though we have achieved that when necessary using a closure.

jrwrigh: We can have the restore not require the field argument (ie. have it just be a wrapper around destroy).  That'd keep the Get/Restore pattern without having to include the `field` parameter. But that also feels like a lot of needless/confusing indirection just to match a function naming pattern.

jrwrigh: To clarify, I agree that keeping a Get/Restore naming pattern is useful and important. I'm maybe questioning if the function just needs a different name.

jrwrigh: Maybe `CeedOperatorFieldGetReference*`/`CeedOperatorFieldGetRef*``? I think that distinguishes it enough from normal Get/Restore patterns.

jeremylt: Yeah, something like that is probably the way to go. I'll rebuild the branch along  those lines once the two PRs preceding this one merge

jeremylt:
Ok, rebased to work off the branch with the lingering Operator memory fix. Impl for CPU ref is in, rest in progress

jeremylt:
CPU blocked and ref in, GPU non-gen next. (would need to rebase in another open PR to do gen)

OCCA will be annoying so I'm saving it for last

jeremylt:
Ok, all backends should be covered, including SYCL (not tested)

The ToDo is handling this for Rust

jeremylt:
Working on Rust, need to fix error handing though

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
```diff
@@ -292,35 +296,30 @@ static inline int CeedOperatorInputBasis_Sycl(CeedInt num_elem, CeedQFunctionFie
...
         CeedCallBackend(CeedVectorSetArray(impl->q_vecs_in[i], CEED_MEM_DEVICE, CEED_USE_POINTER, e_data[i]));
         break;
       case CEED_EVAL_INTERP:
-        CeedCallBackend(CeedOperatorFieldGetBasis(op_input_fields[i], &basis));
-        CeedCallBackend(CeedBasisApply(basis, num_elem, CEED_NOTRANSPOSE, CEED_EVAL_INTERP, impl->e_vecs[i], impl->q_vecs_in[i]));
-        break;
       case CEED_EVAL_GRAD:
         CeedCallBackend(CeedOperatorFieldGetBasis(op_input_fields[i], &basis));
-        CeedCallBackend(CeedBasisApply(basis, num_elem, CEED_NOTRANSPOSE, CEED_EVAL_GRAD, impl->e_vecs[i], impl->q_vecs_in[i]));
+        CeedCallBackend(CeedBasisApply(basis, num_elem, CEED_NOTRANSPOSE, eval_mode, impl->e_vecs[i], impl->q_vecs_in[i]));
```
jrwrigh, line 321: Could maybe limit scope of `basis` to just this switch case.

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
```diff
@@ -407,22 +408,21 @@ static int CeedOperatorApplyAdd_Sycl(CeedOperator op, CeedVector in_vec, CeedVec
...
       case CEED_EVAL_NONE:
         break;
       case CEED_EVAL_INTERP:
-        CeedCallBackend(CeedOperatorFieldGetBasis(op_output_fields[i], &basis));
-        CeedCallBackend(CeedBasisApply(basis, num_elem, CEED_TRANSPOSE, CEED_EVAL_INTERP, impl->q_vecs_out[i], impl->e_vecs[i + impl->num_e_in]));
-        break;
       case CEED_EVAL_GRAD:
         CeedCallBackend(CeedOperatorFieldGetBasis(op_output_fields[i], &basis));
-        CeedCallBackend(CeedBasisApply(basis, num_elem, CEED_TRANSPOSE, CEED_EVAL_GRAD, impl->q_vecs_out[i], impl->e_vecs[i + impl->num_e_in]));
+        CeedCallBackend(CeedBasisApply(basis, num_elem, CEED_TRANSPOSE, eval_mode, impl->q_vecs_out[i], impl->e_vecs[i + impl->num_e_in]));
```
jrwrigh, line 424: Similarly here for limiting `basis` scope.

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
```diff
@@ -992,63 +1000,81 @@ static int CeedSingleOperatorAssembleSetup_Sycl(CeedOperator op) {
...
   for (CeedInt i = 0; i < num_output_fields; i++) {
-    CeedEvalMode eval_mode;
-    CeedVector   vec;
+    CeedVector vec;
 
     CeedCallBackend(CeedOperatorFieldGetVector(output_fields[i], &vec));
     if (vec == CEED_VECTOR_ACTIVE) {
-      CeedCallBackend(CeedOperatorFieldGetBasis(output_fields[i], &basis_out));
-      CeedCallBackend(CeedOperatorFieldGetElemRestriction(output_fields[i], &rstr_out));
-      CeedCheck(!rstr_out || rstr_out == rstr_in, ceed, CEED_ERROR_BACKEND, "Backend does not implement multi-field non-composite operator assembly");
```
jrwrigh, line 1035: Is the change in check here deliberate?

jrwrigh: Obviously it aligns more with the CUDA and HIP backends, but just making sure we're at least mostly certain that the check isn't necessary anymore.

jrwrigh (approved):
I'll run on Sunspot real quick to verify the SYCL changes.

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
```diff
@@ -453,14 +455,14 @@ static int CeedOperatorApplyAdd_Sycl(CeedOperator op, CeedVector in_vec, CeedVec
...
     // Restrict
+    CeedCallBackend(CeedOperatorFieldGetVector(op_output_fields[i], &vec));
+    is_active = vec == CEED_VECTOR_ACTIVE;
     CeedCallBackend(CeedOperatorFieldGetElemRestriction(op_output_fields[i], &rstr));
-    // Active
-    if (vec == CEED_VECTOR_ACTIVE) vec = out_vec;
-
-    CeedCallBackend(CeedElemRestrictionApply(rstr, CEED_TRANSPOSE, impl->e_vecs[i + impl->num_e_in], vec, request));
+    if (is_active) vec = out_vec;
+    CeedCallBackend(CeedElemRestrictionApply(rstr, CEED_TRANSPOSE, impl->e_vecs[i + impl->num_inputs], vec, request));
```
jrwrigh, line 463: `impl->num_inputs` and `impl->num_outputs` is never set, so this is causing a segfault. Working to see if there are any other issues.

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
jeremylt: The SYCL backend is a huge problem IMHO. During development, changes were made from the CUDA/HIP source that didn't have well captured reasons, and since then the CUDA/HIP source has changed, so its really tricky to make good changes here, especially with the limited access to a good development environment. I don't see a good fix in the near term

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
jeremylt: See the most recent commit for a batch of fixes

jrwrigh:
I've pushed up more fixes to `jrwrigh/get-restore`, but there are still a lot of failures to do with the assembly. The failing tests are:
```
Test: t537-operator
Test: t569-operator
Test: t537-operator
Test: t569-operator
Test: t536-operator-f
Test: t533-operator
Test: t506-operator
Test: t534-operator-f
Test: t536-operator
Test: t570-operator
Test: t534-operator
Test: t533-operator-f
Test: t535-operator-f
Test: t535-operator
Test: t538-operator
Test: t506-operator-f
```

`t570-operator` fails with:
```
/home/jrwrigh/software/libCEED/backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp:1086 in CeedSingleOperatorAssembleSetup_Sycl(): Cannot assemble operator without inputs/outputs
```
where it's `num_eval_mode_out = 0` that is triggering the failure. 
`t504-operator` fails with:
```
Large Problem, Component 1: Computed Area 1.254524 != True Area 1.0
```

The rest are correctness failures in assembly. 

I'd try limiting the SYCL changes to the absolute minimum required to get this PR merged.

jeremylt:
Those changes in that branch look good, so feel free to push them to this branch

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
jrwrigh: @jedbrown Thoughts on getting Jeremy access to Sunspot? CI would be nice too, but these kind of dev changes would be easier in a shorter debug-loop.

jeremylt: Giving me a way to build and test would be very helpful - I'd like to make a number of code quality fixes and transfer some CUDA and HIP improvements over

jedbrown: I don't have authority to grant Sunspot access (cc: @KennethEJansen ). I would consider buying a (modestly priced) Intel GPU to put in Noether for dev/testing. Cursory web search hasn't been good at finding actual vendors and I don't know what pricing on the Max 1100 looks like.

jeremylt: Yeah, one way or another it would be hugely helpful to the health of those backends to get me access to Sunspot or to get an Intel card somewhere ourselves, otherwise they run the risk of becoming like the OCCA backend

jeremylt:
@jrwrigh I think I put back the pieces that were accidentally lost

jrwrigh:
Needed to apply the following diff to compile:
```diff
diff --git i/backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp w/backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
index 9d2ecec6..a784e1e5 100644
--- i/backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
+++ w/backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
@@ -662,7 +662,7 @@ static inline int CeedOperatorAssembleDiagonalSetup_Sycl(CeedOperator op) {
       CeedBasis           basis;

       CeedCallBackend(CeedOperatorFieldGetElemRestriction(op_fields[i], &elem_rstr));
-      CeedCheck(!rstr_in || rstr_in == rstr, ceed, CEED_ERROR_BACKEND,
+      CeedCheck(!rstr_in || rstr_in == elem_rstr, ceed, CEED_ERROR_BACKEND,
                 "Backend does not implement multi-field non-composite operator diagonal assembly");
       if (!rstr_in) CeedCallBackend(CeedElemRestrictionReferenceCopy(elem_rstr, &rstr_in));
       CeedCallBackend(CeedOperatorFieldGetBasis(op_fields[i], &basis));
```

But I'm still seeing correctness failures for:

```
t535-operator
t536-operator
t537-operator
t533-operator
t533-operator-f
t506-operator
t536-operator-f
t535-operator-f
t506-operator-f
t538-operator
t534-operator-f
t537-operator
t534-operator
```

All assembly errors though, so you managed to squash some of the issues.

jeremylt:
I'll keep plunking away, thanks for the cross-check!

jeremylt:
`t506` is the odd one, as that's using the same QF for 2 different CeedOperators

jeremylt:
Note
```
make prove -j BACKENDS="/gpu/sycl/ref /gpu/sycl/gen" PROVE_OPTS=-v
```
will give the details on the failures for just those

jrwrigh:
Fixed t506, but the others are still failing: _sorted this time_
```
t533-operator
t533-operator-f
t534-operator
t534-operator-f
t535-operator
t535-operator-f
t536-operator
t536-operator-f
t537-operator
t537-operator
t537-operator
t538-operator
```

They fail with all the SYCL backends (ref, shared, and gen). Not too surprising since the assembly always goes through the ref backend.

jeremylt:
`t506` better is good. That narrows my search.

Its the same routine for all of them, so now you only need to look at `/gpu/sycl/ref`

jeremylt:
Found another difference. Any useful error messages in the test suite output with `PROVE_OPTS=-v` or is it just "entry x is different" sort of stuff?

jrwrigh:
I've been running with `make junit` to get the error messages. But yeah, they're just:
```
[0] Error in assembly: 0.000000 != 0.002963
[1] Error in assembly: 0.000000 != 0.011852
```

or similar.

jeremylt:
Interesting, I wouldn't have expected getting 0s

jrwrigh:
Yeah, all the errors are in the form `Error in assembly: 0.0000 != ....`

backends/sycl-ref/ceed-sycl-ref-operator.sycl.cpp
```diff
@@ -693,26 +703,30 @@ static inline int CeedOperatorAssembleDiagonalSetup_Sycl(CeedOperator op) {
...
+          eval_mode_out[num_eval_mode_out] = eval_mode;
+          num_eval_mode_out += 1;
           break;
         case CEED_EVAL_GRAD:
-          CeedCallBackend(CeedRealloc(num_eval_mode_in + dim, &eval_mode_in));
-          for (CeedInt d = 0; d < dim; d++) eval_mode_in[num_eval_mode_in + d] = eval_mode;
-          num_eval_mode_in += dim;
+          CeedCallBackend(CeedRealloc(num_eval_mode_out + dim, &eval_mode_out));
+          for (CeedInt d = 0; d < dim; d++) eval_mode_out[num_eval_mode_out + d] = eval_mode;
+          num_eval_mode_out += dim;
```
jeremylt, line 729: I think this should fix it.... shame on me

jrwrigh: That passes!

jeremylt:
Huzzah, at long last. I think this is good to add now, and I can spin up the companion PR to MFEM. Mildly disruptive but not a huge deal as it just leaks if not fixed downstream

jeremylt:
#1673 takes priority. Will rework this branch after that merges

jeremylt:
Rebased, manually tested CUDA, HIP. CI won't run on Noether yet due to the failing fluids tests

jrwrigh:
I'll try and fix those when I get back from lunch and my next meeting (which have now combined into the same event).

jrwrigh:
(the bin files just needed to be updated if you want to do it yourself).

jrwrigh:
https://github.com/CEED/libCEED/pull/1681

jeremylt:
@jrwrigh CI has been updated, so HONEE can be updated now

PR #1648, opened 2024-08-22
Reuse GPU E-vecs

jeremylt:
This PR reuses active input e-vec buffers for output e-vecs where able. These are one of the biggest allocations of memory that the user doesn't see.

~~Adding this for AtPoints is a bit trickier. It is probably best as a follow-up~~

Note: manually tested with Ratel as well

### DIFF
--- a/backends/cuda-ref/ceed-cuda-ref-operator.c
+++ b/backends/cuda-ref/ceed-cuda-ref-operator.c
@@ -310,6 +310,7 @@ static int CeedOperatorSetup_Cuda(CeedOperator op) {
     }
     CeedCallBackend(CeedFree(&rstr_used));
   }
+  impl->has_shared_e_vecs = true;
   CeedCallBackend(CeedOperatorSetSetupDone(op));
   return CEED_ERROR_SUCCESS;
 }
@@ -574,7 +575,7 @@ static int CeedOperatorSetupAtPoints_Cuda(CeedOperator op) {
   impl->num_inputs  = num_input_fields;
   impl->num_outputs = num_output_fields;
 
-  // Set up infield and outfield e_vecs and q_vecs
+  // Set up infield and outfield e-vecs and q-vecs
   // Infields
   CeedCallBackend(CeedOperatorSetupFields_Cuda(qf, op, true, true, impl->skip_rstr_in, NULL, impl->e_vecs, impl->q_vecs_in, 0, num_input_fields,
                                                max_num_points, num_elem));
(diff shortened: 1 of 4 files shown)


jeremylt:
I did not add this change on the CPU side, as a 1 or 8 element block for the Opt backends is pretty negligible memory that doesn't seem worth the extra complexity in the code compared to the memory savings on the GPU side

backends/cuda-ref/ceed-cuda-ref-operator.c
```diff
@@ -310,7 +349,7 @@ static inline int CeedOperatorSetupInputs_Cuda(CeedInt num_input_fields, CeedQFu
         uint64_t state;
 
         CeedCallBackend(CeedVectorGetState(vec, &state));
-        if (state != impl->input_states[i] && !impl->skip_rstr_in[i]) {
+        if ((state != impl->input_states[i] || vec == in_vec) && !impl->skip_rstr_in[i]) {
```
jrwrigh, line 352: I'm not sure I follow this change. Does `impl->input_states` not correctly track the state of active inputs? I guess that makes sense since the user could hand over a brand new `CeedVector` as the active input and the operator has no way of knowing that?

backends/cuda-ref/ceed-cuda-ref-operator.c
```diff
@@ -1557,6 +1635,17 @@ static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Cuda(CeedOperator op, C
   max_num_points = impl->max_num_points;
   for (CeedInt i = 0; i < num_elem; i++) num_points[i] = max_num_points;
 
+  // Create separate output e-vecs
+  if (impl->has_shared_e_vecs) {
```
jrwrigh, line 1639: I don't know the normal call structure for this, but are the shared e vecs re-setup when another OperatorApply is done? I'm thinking in non-linear cases where we continuously build the diagonal throughout the simulation.

backends/cuda-ref/ceed-cuda-ref-operator.c
```diff
@@ -264,14 +264,53 @@ static int CeedOperatorSetup_Cuda(CeedOperator op) {
...
+    CeedInt              num_used  = 0;
+    CeedElemRestriction *rstr_used = NULL;
+
+    for (CeedInt i = 0; i < num_input_fields; i++) {
+      bool                is_used = false;
+      CeedVector          vec_i;
+      CeedElemRestriction rstr_i;
+
+      CeedCallBackend(CeedOperatorFieldGetVector(op_input_fields[i], &vec_i));
+      if (vec_i != CEED_VECTOR_ACTIVE) continue;
```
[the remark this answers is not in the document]
jeremylt: In general I think we save more work by not repeating restrictions from passive inputs that haven't changed that we'd gain by reusing those buffers

backends/cuda-ref/ceed-cuda-ref-operator.c
jeremylt: The input states do keep track of active input states fine. The issue is that if I'm using the same buffer for input and output e-vecs corresponding to the active inputs/outptus, then I always need to re-do the active input restriction as the data is always overwritten. Since we rarely rerun the operator with identical input data, I don't see this costing us much here

backends/cuda-ref/ceed-cuda-ref-operator.c
jeremylt: Nope, for this case once we swap away from shared input/output e-vecs, then that's the state the operator stays in. This only effects AtPoints and hopefully we'll be moving away from actually computing the diagonal for the AtPoints operator itself but instead working off an operator using the background mesh quadrature, so this shouldn't be a big issue in the future

backends/cuda-ref/ceed-cuda-ref-operator.c
jrwrigh: Ok. I would say it might be worth adding a comment to that effect to clarify the purpose of `vec == in_vec` here. Something like
```c
// Always apply restriction to active inputs due to potential reuse of e-vecs
```

PR #1649, opened 2024-08-23
memcheck - invalidate arrays before freeing

jeremylt:
Invalidating array data before  freeing to catch using stale copies of the pointer.


backends/memcheck/ceed-memcheck-vector.c
```diff
@@ -49,15 +49,21 @@ static int CeedVectorSetArray_Memcheck(CeedVector vec, CeedMemType mem_type, Cee
 
   CeedCheck(mem_type == CEED_MEM_HOST, CeedVectorReturnCeed(vec), CEED_ERROR_BACKEND, "Can only set HOST memory for this backend");
 
+  if (impl->array_allocated) {
+    for (CeedSize i = 0; i < length; i++) impl->array_allocated[i] = NAN;
+  }
```
rezgarshakeri, line 54: How setting array to `NAN` helps?

jeremylt:
```C
  Ceed       ceed;
  CeedVector x;
  CeedInt    len = 10;

  CeedInit(argv[1], &ceed);

  CeedVectorCreate(ceed, len, &x);
  CeedVectorSetValue(x, 1.0);

  {
    const CeedScalar *read_array, *misbehavior;

    CeedVectorGetArrayRead(x, CEED_MEM_HOST, &read_array);
    misbehavior = read_array;
    CeedVectorRestoreArrayRead(x, &read_array);

    // I want this to show garbage because we've violated access assumptions
      printf("%f\n", misbehavior[0]);
    }
  }

  CeedVectorDestroy(&x);
  CeedDestroy(&ceed);
  return 0;
}
```

rezgarshakeri (approved):

PR #1652, opened 2024-08-29
dealii - remove duplicate GPU vec allocations

jeremylt:
I think part of the issue was that many copies of the input and output vectors were being created, filling up the device memory.

### DIFF
--- a/examples/deal.II/bps.h
+++ b/examples/deal.II/bps.h
@@ -27,7 +27,8 @@
 #include <deal.II/matrix_free/tools.h>
 
 // libCEED includes
-#include <ceed/ceed.h>
+#include <ceed.h>
+#include <ceed/backend.h>
 
 // QFunction source
 #include "bps-qfunctions.h"
@@ -160,6 +161,8 @@ public:
    */
   ~OperatorCeed()
   {
+    CeedVectorDestroy(&src_ceed);
+    CeedVectorDestroy(&dst_ceed);
     CeedOperatorDestroy(&op_apply);
     CeedDestroy(&ceed);
   }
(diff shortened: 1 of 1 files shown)


examples/deal.II/bps.h
```diff
@@ -322,30 +329,43 @@ class OperatorCeed : public OperatorBase<Number>
...
+        x.set_array(src);
+        y.set_array(dst);
 
         // apply operator
-        CeedOperatorApply(op_apply, src_ceed(), dst_ceed(), CEED_REQUEST_IMMEDIATE);
+        CeedOperatorApply(op_apply, x(), y(), CEED_REQUEST_IMMEDIATE);
+
+        // pull arrays back to deal.II
+        x.sync_to_host();
+        y.sync_to_host();
```
jeremylt, line 343: @peterrum I'm not sure if this is a usual pattern for deal.II, but I think that the current `VectorTypeCeed` aren't being destroyed eagerly enough, resulting in multiple device memory allocations which are filling up our device memory during CI.

Ideally there should be 1 persistent libCEED input vector and 1 persistent libCEED output vector corresponding to the `CeedOperator` so we don't have to repeatedly allocate device memory. That's what we do with our PETSc, MFEM, and Nek5000 codes.

Please let me know if there are ways I can make this more idiomatic for deal.II or C++.

peterrum: @jeremylt Thank you working on this. The implementation looks reasonable and only allocating a single vector makes total sense! My original code only was thinking about CPUs where one could use vector-views.

For copying vectors from/to host/device, we have the following similar functions
- https://github.com/chaos-polymtl/lethe/blob/0ba8ab139de5d031da44aae8d319f94633440cfb/prototypes/kokkos_poisson/kokkos_poisson.cc#L278-L281
- https://github.com/chaos-polymtl/lethe/blob/0ba8ab139de5d031da44aae8d319f94633440cfb/prototypes/kokkos_poisson/kokkos_poisson.cc#L295-L298

Looking back at the code, it would be the best to copy the vector from/to the device only once before starting the CG iterations. This is what is the code snippet above. But to make this working for libceed vectors would be more work (see https://github.com/chaos-polymtl/lethe/blob/0ba8ab139de5d031da44aae8d319f94633440cfb/prototypes/kokkos_poisson/kokkos_poisson.cc#L235: here the vector has an additional template parameter for the memory space). 

One question: how did you notice this issue? Did I miss an issue somewhere? Sorry if I did not reply there :(

peterrum (approved):

examples/deal.II/bps.h
jeremylt: You definitely did not miss an issue. I've been hunting down problems with CI failures that only show on ROCm and aren't consistent with which tests fail. I found a bunch of small memory leaks and the issue started consistently triggering on the deal.II example. Other work decreasing our device memory usage masked the failures again, so I started looking at decreasing total device memory usage in the deal.II example.

That snippet you linked to seems to be a good analog of how we tackle the same issue with PETSc, which is reassuring. I'll tweak the names to reflect the above usage.

PR #1654, opened 2024-09-03
gpu - counting points correctly

jeremylt:
Adding the number of points with padding correctly is important


zatkins-dev (approved):
Counting is indeed quite hard

PR #1657, opened 2024-09-05
minor - move 2 functions to clean up docs

jeremylt:
This should do it - the CI build looks good to me


jrwrigh (approved):
Thanks!

Looks like it was the `@addtogroup` that caused the misplaced function documentation.

PR #1667, opened 2024-09-19
minor - drop unneeded casts

jeremylt:
oops


jrwrigh (approved):

jrwrigh:
For the errors on Noether in the fluids example:
```
[WARNING] yaksa: 1 leaked handle pool objects
```

I've only recently seen those on Aurora. Possibly due to latest PETSc `main`?

jeremylt:
Yeah, looks like PETSc is leaking something somewhere maybe? Its for all PETSc examples

jeremylt:
This isn't expressing in Ratel CI - I wonder if there's something fishy with the int64 builds on Noether?

jeremylt:
Hmm, tried setting the opt flags to O2 since MPI is supposed to clean up everything at higher optimization levels, but it doesn't seem to have helped. This error is, according to my initial research, supposed to show when there is a missing `MPI_Type_Free` https://github.com/pmodels/yaksa/discussions/254

jeremylt:
Confirmed, this only occurs for the int64 build, and not the int32 build

PR #1672, opened 2024-09-23
ceed - add Ceed[Get,Restore]WorkVector

jeremylt:
This will require some more work to get the usage in the GPU backends looking clean

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -138,6 +138,43 @@ int CeedRegisterImpl(const char *prefix, int (*init)(const char *, Ceed), unsign
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Create a work vector space for a `ceed`
+
+  @param[in,out] ceed `Ceed` to create work vector space for
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Developer
+**/
+static int CeedWorkVectorsCreate(Ceed ceed) {
+  CeedCall(CeedCalloc(1, &ceed->work_vectors));
+  return CEED_ERROR_SUCCESS;
+}
+
+/**
+  @brief Destroy a work vector space for a `ceed`
+
+  @param[in,out] ceed `Ceed` to destroy work vector space for
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Developer
+**/
+static int CeedWorkVectorsDestroy(Ceed ceed) {
+  if (!ceed->work_vectors) return CEED_ERROR_SUCCESS;
+  for (CeedSize i = 0; i < ceed->work_vectors->num_vecs; i++) {
+    CeedCheck(!ceed->work_vectors->is_in_use[i], ceed, CEED_ERROR_ACCESS, "Work vector %" CeedSize_FMT " checked out but not returned");
+    ceed->ref_count += 2;  // Note: increase ref_count to prevent Ceed destructor from triggering again
+    CeedCall(CeedVectorDestroy(&ceed->work_vectors->vecs[i]));
+    ceed->ref_count -= 1;  // Note: restore ref_count
+  }
+  CeedCall(CeedFree(&ceed->work_vectors->is_in_use));
+  CeedCall(CeedFree(&ceed->work_vectors->vecs));
+  CeedCall(CeedFree(&ceed->work_vectors));
+  return CEED_ERROR_SUCCESS;
+}
+
 /// @}
 
 /// ----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


interface/ceed.c
```diff
@@ -751,6 +784,66 @@ int CeedReference(Ceed ceed) {
...
+  // Long enough vector was not found
+  if (i == ceed->work_vectors->num_vecs) {
+    if (ceed->work_vectors->num_vecs == 0) {
+      ceed->work_vectors->num_vecs = 1;
+      CeedCall(CeedCalloc(ceed->work_vectors->num_vecs, &ceed->work_vectors->vecs));
+      CeedCall(CeedCalloc(ceed->work_vectors->num_vecs, &ceed->work_vectors->is_in_use));
+    } else {
+      ceed->work_vectors->num_vecs++;
+      CeedCall(CeedRealloc(ceed->work_vectors->num_vecs, &ceed->work_vectors->vecs));
+      CeedCall(CeedRealloc(ceed->work_vectors->num_vecs, &ceed->work_vectors->is_in_use));
```
jedbrown, line 823: Formally this is a quadratic algorithm. The numbers are small so it may be harmless, but growable data structures usually decouple capacity from current size so that, e.g., capacity can be doubled each time reallocation is needed.

interface/ceed.c
```diff
@@ -751,6 +784,66 @@ int CeedReference(Ceed ceed) {
...
+    }
+    CeedCallBackend(CeedVectorCreate(ceed, len, &ceed->work_vectors->vecs[i]));
```
jedbrown, line 825: This is going to increment the reference count on `ceed`, which is a reference-counted cycle and thus a memory leak given the current logic in `CeedDestroy`. There are two ways to avoid this. One is for `CeedRestoreWorkVector` to drop its reference to the `Ceed` (putting the object in an invalid state, but we might accept that while it's essentially just a memory allocation, reinstating the `Ceed` when it's checked out) or to count those cyclic references in the `CeedDestroy` logic.

jeremylt: Ugh, that's right.

jeremylt: ok, should be good now
I'm taking the strategy of leaving un-refcounted vectors in there, bumping the ceed's refcount when checking them out, decreasing the refcount when checking back in

jeremylt:
The actual usage will involve some serious re-organization of the GPU operator code, so I think its probably best to merge this on its own.

interface/ceed.c
```diff
@@ -751,6 +788,81 @@ int CeedReference(Ceed ceed) {
...
+    }
+  }
+  // Long enough vector was not found
+  if (i == ceed->work_vectors->num_vecs) {
+    if (ceed->work_vectors->max_num_vecs == 0) {
+      ceed->work_vectors->max_num_vecs = 1;
+      CeedCall(CeedCalloc(ceed->work_vectors->max_num_vecs, &ceed->work_vectors->vecs));
+      CeedCall(CeedCalloc(ceed->work_vectors->max_num_vecs, &ceed->work_vectors->is_in_use));
+    } else if (ceed->work_vectors->max_num_vecs == i) {
+      ceed->work_vectors->max_num_vecs *= 2;
```
jrwrigh, line 825: `max_num_vecs` makes me think the variable should be static (i.e. the number of work vectors before we start to destroy the last untouched one). I can't think of a better name though.

interface/ceed.c
```diff
@@ -751,6 +788,81 @@ int CeedReference(Ceed ceed) {
...
+      CeedCall(CeedRealloc(ceed->work_vectors->max_num_vecs, &ceed->work_vectors->vecs));
+      CeedCall(CeedRealloc(ceed->work_vectors->max_num_vecs, &ceed->work_vectors->is_in_use));
+    }
+    ceed->work_vectors->num_vecs++;
+    CeedCallBackend(CeedVectorCreate(ceed, len, &ceed->work_vectors->vecs[i]));
+    ceed->ref_count--;  // Note: ref_count manipulation to prevent a ref-loop
+  }
+  // Return pointer to work vector
+  ceed->work_vectors->is_in_use[i] = true;
+  *vec                             = NULL;
```
jrwrigh, line 835: May be misremembering, but don't we normally call `CeedVectorDestroy()` rather than setting to `NULL`?

interface/ceed.c
jrwrigh: Naming aside, having the number of stored vectors always increase seems like it might be a foot-gun at some later point. I feel like we should have a hard limit before it starts to cull the stored vecs.

Granted, I don't see exponential growth of stored vecs being an issue, but just a thought.

jrwrigh (approved):

interface/ceed.c
jeremylt: This is due to a C-ism.

The variable may be uninitialized and contain junk, but if we call RreferenceCopy on junk then we'll segfault. We can either set the pointer to NULL ourselves before ReferenceCopy (current code), or require the user to pass in a valid pointer to a vec they haven't used yet, or we could use Reference with a manual copy.

interface/ceed.c
jeremylt: We can pick a limit, but I guess it feels arbitrary to me since each vector can be arbitrarily large already.

interface/ceed.c
jeremylt: I've thought about of making a helper for the combo of
```C
foo = NULL;
Ceed*ReferenceCopy(bar, &foo);
```
but the main reason I haven't is because I really don't want users generally doing such a thing as that becomes a possible leak

jrwrigh: Ah, yeah, I [ran into that for `ReferenceCopy`](https://gitlab.com/phypid/honee/-/blob/main/src/qdata.c?ref_type=heads#L61-64). It's not a big ask for users to initialize those as `NULL`.

PR #1682, opened 2024-10-04
ci - allow early failure on GitLab CI

jeremylt:
Copying this over from Ratel


jeremylt:
Failed as expected, so we can merge

jrwrigh (approved):

PR #1696, opened 2024-10-16
JiT include update

jeremylt:
Update JiT to `-I` all the JiT directories set on the Ceed.

Note: Most of the changes are repeated ripple effects from the core changes it `ceed-[cuda,hip]-compile.h`.

1) Allow searching of dirs set with CeedAddJitSourceRoot for #include <foo.h>

2) Add CEED_RUNNING_JIT_PASS to guard files that cannot be included for JiT (math.h or std*.h)

3) Prefer ceed/types.h over ceed.h in QF source

4) Redirect ceed.h to ceed/types.h over ceed/ceed.h when CEED_RUNNING_JIT_PASS is set

5) Use include statements instead of loading string buffers for JiT

~~Ok, here's a rough prototype of an idea @jedbrown. This allows `#include <foo.h>` (stripping out `#include <std*.h>`) and would add all JiT dirs as `-I/include/this/dir` flags to the CUDA/HIP JiT compilers.~~

~~Very rough and not rigorously tested but seems to work locally with hacky manual testing.~~


backends/cuda/ceed-cuda-compile.cpp
```diff
@@ -93,6 +93,7 @@ int CeedCompile_Cuda(Ceed ceed, const char *source, CUmodule *module, const Ceed
       + std::to_string(prop.major) + std::to_string(prop.minor);
   opts[1] = arch_arg.c_str();
   opts[2] = "-Dint32_t=int";
+  opts[3] = "-I/home/jeremy/Dev/libCEED/include/ceed/jit-source/"
```
jeremylt, line 96: would loop over all set JiT dirs here to add them all

interface/ceed-jit-tools.c
```diff
@@ -216,6 +216,9 @@ int CeedLoadSourceToInitializedBuffer(Ceed ceed, const char *source_file_path, C
       bool  is_ceed_header    = next_left_chevron && (next_new_line - next_left_chevron > 0) &&
                             (!strncmp(next_left_chevron, "<ceed/jit-source/", 17) || !strncmp(next_left_chevron, "<ceed/types.h>", 14) ||
                              !strncmp(next_left_chevron, "<ceed/ceed-f32.h>", 17) || !strncmp(next_left_chevron, "<ceed/ceed-f64.h>", 17));
+      bool is_std_header =
+          next_left_chevron && (next_new_line - next_left_chevron > 0) &&
+          (!strncmp(next_left_chevron, "<std", 4) || !strncmp(next_left_chevron, "<math.h>", 8) || !strncmp(next_left_chevron, "<ceed.h>", 8));
```
jeremylt, line 221: I'm worried about this list of headers that CPU wants but GPU hates on being brittle

jedbrown (commented):
Probably this means we could delete code related to actually finding those headers and reading them. At a glance, this looks good and presumably reduces the volume of output when using `CEED_DEBUG=1`.

jrwrigh:
> Probably this means we could delete code related to actually finding those headers and reading them.

 I'm guessing the OpenCL online compiler has similar functionality, but we'll need to check that for the SYCL backends.

tests/test-include/fake-sys-include.h
```diff
@@ -0,0 +1,10 @@
+#define FAKE_SYS_SCALE_ONE 1
+
+// Note - files included this way cannot transitively include any files CUDA/ROCm won't compile
+// These are bad
+// #include <ceed.h>
+// #include <math.h>
+// #include <stddef.h>
```
jeremylt, line 7: This is the sharp edge here and why we can't currently totally replace all our JiT infrastructure with this approach (yet?). A simple include of `ceed.h` which is in a lot of QFunction code brings in these lines
```
#include <stdarg.h>
#include <stdbool.h>
#include <stdio.h>
```
and CUDA/ROCm hate those includes.

So currently this PR keeps the chasing of the local includes via "" to build a string buffer but permits system includes via <> to add additional source dirs inside of the user QFunction source.

jeremylt: I think we can merge this as is and make a second attack at this once we're ready to revamp SYCL (I just got my access to the machine but need to sign up and such)

jedbrown: That makes sense. We could move the necessary declarations to `ceedtypes.h` and have qfunction headers only include those, not `ceed.h`.

jeremylt: I think `ceed/types.h` has everything that could be used in a QFunction

jedbrown: Maybe we should explicitly recommend that. No need to force people to change, but making the compilation environment (and context present in an IDE) as close as possible to the JIT environment would be a good thing for low-friction workflow.

jeremylt: We could have an ifdef that swaps `ceed.h` to include `ceed/types.h` instead of `ceed/ceed.h` during JiT?

jeremylt: Note to me:

1) use `code << #include <ceed/jit-source/cuda/cuda-jit.h>;`

2) could maybe `code << #include " << source_path;`

3) then could `-I << source_dir;` for above file?

include/ceed.h
```diff
@@ -1 +1,5 @@
+#ifdef CEED_RUNNING_JIT_PASS
+#include "ceed/types.h"
+#else
 #include "ceed/ceed.h"
+#endif
```
jeremylt, line 5: This should be transparent to the user

tests/test-include/fake-sys-include.h
jeremylt: Ok, 1 was really easy to do and should save us some headache. 2 and 3 can probably wait for a follow-up?

jeremylt:
Need to run full suite tomorrow, but seems fine with Ratel (should be)

include/ceed.h
jedbrown: Yeah, though the IDE (`clangd` or other) will assume you get the whole `ceed/ceed.h` and thus not notify if you use functions that should not be used in qfunctions. So this is fine to preserve compatibility, but we should also move toward only including `ceed/types.h` so that qfunction sources are checked in a similar context to how they will be during JIT (where it's harder to work with error messages).

jeremylt: agreed (latest commit)

jeremylt:
Ratel is happy. Noether looks happy.

I say we merge this and do further improvements as follow-up.

NamjaeChoi:
Hello, I checked out your branch and tested it, but the error persists:
```/data/choin/moose/libCEED/interface/ceed-jit-tools.c:138 in CeedLoadSourceToInitializedBuffer(): Couldn't open source file: /data/choin/projects/moose/framework/build/header_symlinks/Kokkos_Core.hpp```
This file is in another directory and I provided that directory to `CeedAddJitSourceRoot`. `~/header_symlinks` is where the QFunction source is located.

jeremylt:
This remains really difficult for me to fully understand without the actual code generating the issue

jeremylt:
Guessing here, if you are using a local include `#include "Kokkos_Core.hpp"` instead of `#include <Kokkos_Core.hpp>`, then as we talked about in the other thread, that local include is still being interpreted as a relative path from the QFunction source file during code generation.

NamjaeChoi:
Ah so `#include "header.h"` still searches for the local directory only?

jeremylt:
Yes, at this time we're still interpreting `#include "foo.h"` as a local file

NamjaeChoi:
Changing to `#include <header.h>` seems to work. Let me try further. Thank you!

NamjaeChoi:
Do you plan to change the behavior of `#include "header.h"` as well?

jeremylt:
Not in this PR, no.

PR #1699, opened 2024-10-21
Add CeedAddJitDefine

jeremylt:
Now that we switched to heavily using `-Ifoo/bar`, this adds the ability for users to set `-Dfoo=bar`

### DIFF
--- a/interface/ceed.c
+++ b/interface/ceed.c
@@ -1347,7 +1347,7 @@ int CeedAddJitSourceRoot(Ceed ceed, const char *jit_source_root) {
   Ceed ceed_parent;
 
   CeedCall(CeedGetParent(ceed, &ceed_parent));
-  CeedCheck(!ceed_parent->num_jit_source_roots_readers, ceed, CEED_ERROR_ACCESS, "Cannot add JiT source root, read access has been granted");
+  CeedCheck(!ceed_parent->num_jit_source_roots_readers, ceed, CEED_ERROR_ACCESS, "Cannot add JiT source root, read access has not been restored");
 
   CeedInt index       = ceed_parent->num_jit_source_roots;
   size_t  path_length = strlen(jit_source_root);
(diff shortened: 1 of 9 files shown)


interface/ceed.c
```diff
@@ -910,6 +920,43 @@ int CeedRestoreJitSourceRoots(Ceed ceed, const char ***jit_source_roots) {
...
+  @param[out] jit_defines String such as `foo=bar`, used as `-Dfoo=bar` in JiT
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedRestoreJitDefines(Ceed ceed, const char ***jit_defines) {
+  *jit_defines = NULL;
+  return CEED_ERROR_SUCCESS;
+}
```
jrwrigh, line 958: Not sure how I feel about a Restore function that doesn't actually do anything? Is there something we would plan on doing later with this function?

jrwrigh (approved):

interface/ceed.c
jeremylt: My long-term plan is to make any get that isn't a scalar value need a restore so we don't have to remember the magic list of exceptions in our head

jrwrigh: Gotcha. Should we add a lock or something to kinda enforce it? We don't have to remember the exception, but there's no consequence for not remembering to use it at all.

jeremylt: Maybe a reader count? It could be bad to miss an option added last minute 

I should put in a branch to tidy all this up with all our Get*

jrwrigh: Yeah, reader count sounds good.

jrwrigh: > It could be bad to miss an option added last minute

You mean don't allow `CeedAddJitDefine` to be called if the reader count is non-zero?

jeremylt: See c6d6a888b8f8436f07194058d4fda63c5f1fa761

interface/ceed.c
```diff
@@ -1290,17 +1347,52 @@ int CeedAddJitSourceRoot(Ceed ceed, const char *jit_source_root) {
...
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedAddJitDefine(Ceed ceed, const char *jit_define) {
+  Ceed ceed_parent;
+
+  CeedCall(CeedGetParent(ceed, &ceed_parent));
+  CeedCheck(!ceed_parent->num_jit_defines_readers, ceed, CEED_ERROR_ACCESS, "Cannot add JiT define, read access has been granted");
```
jrwrigh, line 1380:
```suggestion
  CeedCheck(!ceed_parent->num_jit_defines_readers, ceed, CEED_ERROR_ACCESS, "Cannot add JiT define, read access has not been restored");
```

interface/ceed.c
jrwrigh: Looks good to me. Small suggestion for the error message.

PR #1702, opened 2024-10-22
Ceed*Get[CeedObject] Needs Destroy

jeremylt:
Requiring Ceed objects returned from `Ceed*Get[CeedObject]` to be destroyed

- `Ceed*GetCeed`
- `CeedGetParent`
- `CeedGetDelegate`
- `CeedGetObjectFallback`
- `CeedGetOperatorFallbackCeed`
- `CeedQFunctionGetContext`
- `CeedOperatorGetQFunction`
- `CeedOperatorGetFallbackParentCeed`

### DIFF
--- a/rust/libceed/src/operator.rs
+++ b/rust/libceed/src/operator.rs
@@ -442,11 +442,7 @@ impl<'a> OperatorCore<'a> {
     // Error handling
     #[doc(hidden)]
     fn check_error(&self, ierr: i32) -> crate::Result<i32> {
-        let mut ptr = std::ptr::null_mut();
-        unsafe {
-            bind_ceed::CeedOperatorGetCeed(self.ptr, &mut ptr);
-        }
-        crate::check_error(ptr, ierr)
+        unsafe { crate::check_error(bind_ceed::CeedOperatorReturnCeed(self.ptr), ierr) }
     }
 
     // Common implementations
--- a/rust/libceed/src/operator.rs
+++ b/rust/libceed/src/operator.rs
@@ -33,22 +33,21 @@ impl<'a> OperatorField<'a> {
     ) -> crate::Result<Self> {
         let vector = {
             let mut vector_ptr = std::ptr::null_mut();
-            let ierr = unsafe { bind_ceed::CeedOperatorFieldGetVector(ptr, &mut vector_ptr) };
-            ceed.check_error(ierr)?;
+            ceed.check_error(unsafe {
+                bind_ceed::CeedOperatorFieldGetVector(ptr, &mut vector_ptr)
+            })?;
             crate::Vector::from_raw(vector_ptr)?
         };
         let elem_restriction = {
             let mut elem_restriction_ptr = std::ptr::null_mut();
-            let ierr = unsafe {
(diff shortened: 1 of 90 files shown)


jedbrown:
Is MFEM going to need updates to track this?

jedbrown (approved):
Do the language bindings (Python, Julia, and Rust) need updates for these?

jeremylt:
Surprisingly, the updates MFEM needs for v0.13 are pretty minimal. I've got a branch open, and I don't think this PR will add any new changes. But Ratel and HONEE will need a couple of changes

rust/libceed/src/vector.rs
```diff
@@ -348,11 +348,7 @@ impl<'a> Vector<'a> {
     // Error handling
     #[doc(hidden)]
     fn check_error(&self, ierr: i32) -> crate::Result<i32> {
-        let mut ptr = std::ptr::null_mut();
-        unsafe {
-            bind_ceed::CeedVectorGetCeed(self.ptr, &mut ptr);
-        }
-        crate::check_error(ptr, ierr)
+        unsafe { crate::check_error(bind_ceed::CeedVectorReturnCeed(self.ptr), ierr) }
     }
```
jedbrown, line 352: As a matter of style/ergonomics, we can make a safe private function `Vector::ceed(&self) -> Ceed`.

We could also make `crate::check_error` accept a closure so we aren't eagerly making real calls just to obtain a `Ceed` we won't use. Then this function would look like `crate::check_error(|| self.ceed(), ierr)`. That seems more in the spirit of standard library interfaces like [`ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else)

The ergonomics above are perhaps kinda moot in that I think the unsafe FFI blocks should use a closure, perhaps similar to RSMPI `with_uninitialized` (which provisions a suitable pointer type for an output argument and does not currently do error checking, but could be extended for that).

jeremylt: Something like 11544396610b36de1cb2f0d18032eefe5c670568?

jedbrown: Yup, exactly. Feels nicer to have less/narrower `unsafe`.

rust/libceed/src/operator.rs
```diff
@@ -2105,7 +2088,7 @@ impl<'a> Operator<'a> {
         let mut ptr_coarse = std::ptr::null_mut();
         let mut ptr_prolong = std::ptr::null_mut();
         let mut ptr_restrict = std::ptr::null_mut();
-        let ierr = unsafe {
+        self.op_core.check_error(unsafe {
```
jedbrown, line 2091: In the future (definitely a different PR), this could be `let (op_coarse, op_prolong, op_restrict) = ceed.op_core.check_error_with_3(|ptr_coarse, ptr_prolong, ptr_restrict| unsafe { ... })?;`.

Note that `Operator::from_raw` should be `unsafe` because it will dereference a raw pointer. Just noticed that while browsing (also not for this PR, and I'm not trying to snipe you into working on it).

PR #1703, opened 2024-10-23
fluids: Update MatCeed to be up-to-date with HONEE and Ratel

jrwrigh:
Update's MatCeed in the fluids example to be up to date with HONEE and Ratel.

Note this requires Noether's PETSc install to be updated to reflect the changes in PETSc MRs !7900 and !7960.


jeremylt:
We still don't have a fix for the CUDA issue with fluids, yes? I can rebuild all the CI PETSc in the morning

jrwrigh:
Yeah, still no fix to the CUDA fluids issue. At least not that I've tested. Should probably make a PETSc issue for that, just so it's a known regression/change of behavior.

KennethEJansen:
While I am a strong advocate of getting this reviewed, it touches a lot of code I have very little experience with so I think this is an @jedbrown or @jeremylt review.
I see mention of a CUDA issue. I can offer up that all the Polaris tests I posted on ALCF-CNDA with a date of 241023 PETSc were with this branch since main would not exit cleanly  there. The results were not run through a formal comparison but convergence and performance appeared to be correct.

jedbrown (approved):

PR #1709, opened 2024-10-27
fix(sycl): Replaces pragma once with include guards

jrwrigh:
OpenCL doesn't really like pragma once evidently. I think we've been 'getting away' with it previously as the JIT processing automatically doesn't allow for nested includes, but the same is not done for <ceed/types.h>?

### DIFF
--- a/include/ceed/types.h
+++ b/include/ceed/types.h
@@ -7,7 +7,8 @@
 
 /// @file
 /// Public header for types and macros used in user QFunction source code
-#pragma once
+#ifndef CEED_QFUNCTION_DEFS_H
+#define CEED_QFUNCTION_DEFS_H
 
 #ifndef CEED_RUNNING_JIT_PASS
 #include <stddef.h>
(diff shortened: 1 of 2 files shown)


jeremylt:
I'd double check the other Sycl JiT files 

Note - I'm out next week and won't be approving anything

jeremylt (approved):

jedbrown:
Rebased after merging #1703

include/ceed/types.h
```diff
@@ -7,7 +7,8 @@
 
 /// @file
 /// Public header for types and macros used in user QFunction source code
-#pragma once
+#ifndef CEED_QFUNCTION_DEFS_H
```
jeremylt, line 10: Small note - I've seen some people recommend both. Not sure I fully understand the argument why though

jrwrigh: Yeah, both is a bit perplexing... Maybe to avoid possible guard name conflicts?

jeremylt: I think for a while `#pragma once` gave faster compile times, but nowadays that's less of a factor

PR #1711, opened 2024-11-15
GPU Shared AtPoints Bases

jeremylt:
This PR adds AtPoints to `/gpu/[cuda,hip]/shared`, which is a blocker for a `/gpu/[cuda,hip]/gen` AtPoints capability.

~~Work in progress. The `t35*` series tests pass, but the `t59*` series don't all pass.~~

~~There seems to be an issue with the 2D/3D transpose interp and grad with multiple elements.~~

Passing for libCEED t* tests. `petsc-bpsswarm` isn't passing yet though. There is also some issue with a few Ratel tests where they get stuck, probably on a `__syncthreads()`. (MMS tests it seems).

### DIFF
--- a/backends/cuda-shared/ceed-cuda-shared-basis.c
+++ b/backends/cuda-shared/ceed-cuda-shared-basis.c
@@ -64,7 +64,7 @@ static int CeedBasisApplyTensorCore_Cuda_shared(CeedBasis basis, bool apply_add,
       if (dim == 1) {
         CeedInt elems_per_block = CeedIntMin(ceed_Cuda->device_prop.maxThreadsDim[2], CeedIntMax(512 / thread_1d,
                                                                                                  1));  // avoid >512 total threads
-        CeedInt grid            = num_elem / elems_per_block + ((num_elem / elems_per_block * elems_per_block < num_elem) ? 1 : 0);
+        CeedInt grid            = num_elem / elems_per_block + (num_elem % elems_per_block > 0);
         CeedInt shared_mem      = elems_per_block * thread_1d * sizeof(CeedScalar);
 
         if (t_mode == CEED_TRANSPOSE) {
@@ -77,7 +77,7 @@ static int CeedBasisApplyTensorCore_Cuda_shared(CeedBasis basis, bool apply_add,
         const CeedInt opt_elems[7] = {0, 32, 8, 6, 4, 2, 8};
         // elems_per_block must be at least 1
         CeedInt elems_per_block = CeedIntMax(thread_1d < 7 ? opt_elems[thread_1d] / num_comp : 1, 1);
-        CeedInt grid            = num_elem / elems_per_block + ((num_elem / elems_per_block * elems_per_block < num_elem) ? 1 : 0);
+        CeedInt grid            = num_elem / elems_per_block + (num_elem % elems_per_block > 0);
         CeedInt shared_mem      = elems_per_block * thread_1d * thread_1d * sizeof(CeedScalar);
 
         if (t_mode == CEED_TRANSPOSE) {
@@ -88,7 +88,7 @@ static int CeedBasisApplyTensorCore_Cuda_shared(CeedBasis basis, bool apply_add,
         }
       } else if (dim == 3) {
         CeedInt elems_per_block = 1;
-        CeedInt grid            = num_elem / elems_per_block + ((num_elem / elems_per_block * elems_per_block < num_elem) ? 1 : 0);
+        CeedInt grid            = num_elem / elems_per_block + (num_elem % elems_per_block > 0);
         CeedInt shared_mem      = elems_per_block * thread_1d * thread_1d * sizeof(CeedScalar);
 
         if (t_mode == CEED_TRANSPOSE) {
@@ -115,7 +115,7 @@ static int CeedBasisApplyTensorCore_Cuda_shared(CeedBasis basis, bool apply_add,
       if (dim == 1) {
         CeedInt elems_per_block = CeedIntMin(ceed_Cuda->device_prop.maxThreadsDim[2], CeedIntMax(512 / thread_1d,
                                                                                                  1));  // avoid >512 total threads
-        CeedInt grid            = num_elem / elems_per_block + ((num_elem / elems_per_block * elems_per_block < num_elem) ? 1 : 0);
+        CeedInt grid            = num_elem / elems_per_block + (num_elem % elems_per_block > 0);
         CeedInt shared_mem      = elems_per_block * thread_1d * sizeof(CeedScalar);
 
         if (t_mode == CEED_TRANSPOSE) {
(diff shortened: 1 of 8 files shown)


jeremylt:
Yay - it works locally but fails on Nother. Those are always fun to debug

jeremylt:
Cuda is more betterer now. Need to check thread block sizes setup for Hip for 3D

jeremylt:
Ok, now every element other than the first in 3D is wrong for HIP. Progress, but super bizarre how it doesn't seem to behave as I expect. Almost the same logic as CUDA so I have to miss some subtle change between HIP shared and CUDA shared as those two have slightly diverged.

jeremylt:
Ok, the kernels work now. We can optimize them in the future

jeremylt:
Local testing for Ratel passes

include/ceed/jit-source/cuda/cuda-shared-basis-tensor-at-points.h
```diff
@@ -0,0 +1,252 @@
...
+        InterpAtPoints1d<BASIS_NUM_COMP, BASIS_NUM_PTS, BASIS_Q_1D>(data, i, r_C, r_X, r_V);
+      } else if (BASIS_DIM == 2) {
+        InterpAtPoints2d<BASIS_NUM_COMP, BASIS_NUM_PTS, BASIS_Q_1D>(data, i, r_C, r_X, r_V);
+      } else if (BASIS_DIM == 3) {
+        InterpAtPoints3d<BASIS_NUM_COMP, BASIS_NUM_PTS, BASIS_Q_1D>(data, i, r_C, r_X, r_V);
+      }
+      for (CeedInt j = 0; j < BASIS_NUM_COMP; j++) {
+        if (i < BASIS_NUM_PTS) d_V[elem * BASIS_NUM_PTS + num_elem * BASIS_NUM_PTS * j + i] = r_V[j];
+      }
+    }
```
jeremylt, line 73: This probably needs to be a template for Gen, might need to tweak the writing out as well

backends/cuda-shared/ceed-cuda-shared-basis.c
```diff
@@ -323,17 +322,76 @@ static int CeedBasisApplyAtPointsCore_Cuda_shared(CeedBasis basis, bool apply_ad
...
+        CeedInt elems_per_block = CeedIntMax(thread_1d < 7 ? opt_elems[thread_1d] / num_comp : 1, 1);
+        CeedInt grid            = num_elem / elems_per_block + ((num_elem / elems_per_block * elems_per_block < num_elem) ? 1 : 0);
+        CeedInt shared_mem      = elems_per_block * thread_1d * thread_1d * sizeof(CeedScalar);
 
-      CeedCallBackend(
-          CeedRunKernel_Cuda(ceed, is_transpose ? data->InterpTransposeAtPoints : data->InterpAtPoints, num_elem, block_size, interp_args));
+        CeedCallBackend(CeedRunKernelDimShared_Cuda(ceed, is_transpose ? data->InterpTransposeAtPoints : data->InterpAtPoints, grid, thread_1d,
+                                                    thread_1d, elems_per_block, shared_mem, interp_args));
+      } else if (dim == 3) {
+        CeedInt elems_per_block = 1;
```
zatkins-dev, line 352: Is there not a case where we would do more than one element per block?

backends/cuda-shared/ceed-cuda-shared-basis.c
```diff
@@ -323,17 +322,76 @@ static int CeedBasisApplyAtPointsCore_Cuda_shared(CeedBasis basis, bool apply_ad
...
+        CeedInt grid            = num_elem / elems_per_block + ((num_elem / elems_per_block * elems_per_block < num_elem) ? 1 : 0);
```
zatkins-dev, line 353:
```suggestion
        CeedInt grid            = num_elem / elems_per_block + ((num_elem % elems_per_block > 0) ? 1 : 0);
```
Is this the same? I think it's a lot more clear than the integer division

zatkins-dev: same for the other places where this condition occurs

backends/cuda-shared/ceed-cuda-shared-basis.c
zatkins-dev: The block sizes seem to be determined differently for HIP, what's the motivation there?

include/ceed/jit-source/cuda/cuda-shared-basis-tensor-at-points-templates.h
```diff
@@ -0,0 +1,568 @@
+// Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and other CEED contributors.
```
zatkins-dev, line 1: Hmm, it feels weird that the only difference between the HIP and CUDA files is the type of the `SharedData_*` parameter. I suppose for clarity (and consistency), it makes sense to fully separate HIP and CUDA backends, but for ease of maintenance, is it possible we could do something like, for example:
```c
template <typename SharedData_t, int NUM_COMP, int NUM_POINTS, int Q_1D>
inline __device__ void InterpAtPoints1d(SharedData_t &data, const CeedInt p, const CeedScalar *__restrict__ r_C, const CeedScalar *r_X,
                                        CeedScalar *__restrict__ r_V) {
```

This isn't necessarily a suggestion for this PR, just a thought about simplifications we could make in the future to reduce duplication

zatkins-dev (approved):
Overall looks good! I have a few clarifying questions, but I think this is good to merge

include/ceed/jit-source/cuda/cuda-shared-basis-tensor-at-points-templates.h
jeremylt: There are some other small differences too. I'm a little hesitant to make more template parameters right now given some of the frustration PETSc has felt doing something similar.

backends/cuda-shared/ceed-cuda-shared-basis.c
jeremylt: @nbeams probably can speak to it better but I think its a little different in order to deliver better performance ?

backends/cuda-shared/ceed-cuda-shared-basis.c
jeremylt: Yeah.... Not sure why that was the original logic

jeremylt:
```suggestion
        CeedInt grid            = num_elem / elems_per_block + (num_elem % elems_per_block > 0);
```
This is probably the way to go?

backends/cuda-shared/ceed-cuda-shared-basis.c
nbeams: I think it probably is tied to https://github.com/CEED/libCEED/commit/9e31c45b1261f08aabe3aa77b1bb7169e37ea122 -- in early days of the HIP backend development, it was recommended to make sure to include launch bounds on kernels, so the AMD compiler could optimize things like register usage. Not sure how much of a difference it would really make with the latest compilers.

PR #1653, opened 2024-08-30
Rust - shrink  prelude

jeremylt:
Ok, dropped a lot of the contents of the prelude, mirroring changes in petsc-rs.

Also made some clippy fixes.

fixes #1599

PR #1712, opened 2024-12-02
Version 1.0

jeremylt:
~~Blocked by #1788~~ 
~~Blocked by #1711~~
~~Blocked by #1653~~
~~Blocked by #1881~~
~~Blocked by #1875~~
Blocked by ~~#1818~~ ~~#1919~~
Blocked by #1813

Draft release notes:

Release of libCEED v1.0 :tada: 

### New features

- Add `CeedOperatorCreateAtPoints` which evaluates the `CeedQFunction` at arbitrary locations in each element, for use in Particle in Cell, Material Point Method, and similar methods.
- Add `CeedElemRestrictionGetLLayout` to provide L-vector layout for strided `CeedElemRestriction` created with `CEED_BACKEND_STRIDES`.
- Add `CeedVectorReturnCeed` and similar when parent `Ceed` context for a libCEED object is only needed once in a calling scope.
- Enable `#pragma once` for all JiT source; remove duplicate includes in JiT source string before compilation.
- Allow user to set additional compiler options for CUDA and HIP JiT.
Specifically, directories set with `CeedAddJitSourceRoot(ceed, "foo/bar")` will be used to set `-Ifoo/bar` and defines set with `CeedAddJitDefine(ceed, "foo=bar")` will be used to set `-Dfoo=bar`.
- Added non-tensor basis support to code generation backends `/gpu/cuda/gen` and `/gpu/hip/gen`.
- Added support to code generation backends `/gpu/cuda/gen` and `/gpu/hip/gen` for operators with both tensor and non-tensor bases.
- Added support for QFunctions written in Rust for CUDA backends.
- Add `CeedGetGitVersion()` to access the Git commit and dirty state of the repository at build time.
- Add `CeedGetBuildConfiguration()` to access compilers, flags, and related information about the build environment.

### Interface changes

- Add `bool` field type for `CeedQFunctionContext` and related interfaces to use `bool` fields.
- `CEED_BASIS_COLLOCATED` removed; users should only use `CEED_BASIS_NONE`.
- Remove unneeded pointer for `CeedElemRestrictionGetELayout`.
- Change QFunction source include file handling in JiT compilers
    - Add `CEED_RUNNING_JIT_PASS` compiler definition for wrapping header files that device JiT compilers cannot read
    - Users should now prefer `#include <ceed/types.h>` rather than `#include <ceed.h>` in QFunction source files
- Require use of `Ceed*Destroy()` on Ceed objects returned from `Ceed*Get*()`.
- Rename `CeedCompositeOperatorCreate()` to `CeedOperatorCreateComposite()` for uniformity.
- Rename `CeedCompositeOperator*()` to `CeedOperatorComposite*()` for uniformity.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,7 +1,7 @@
 cff-version: 1.2.0
 title: "libCEED: Efficient Extensible Discretization"
-version: 0.12.0
-date-released: 2023-10-31
+version: 1.0.0
+date-released: 2025-08-15
 license:  BSD-2-Clause
 message: "Please cite the following works when using this software."
 authors:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ On this page we provide a summary of the main API changes, new features and exam
 
 ### Interface changes
 
+### New features
+
+### Examples
+
+(v1-0)=
+
+## v1.0 (August 15, 2025)
+
+### Interface changes
+
 - Add `bool` field type for `CeedQFunctionContext` and related interfaces to use `bool` fields.
 - `CEED_BASIS_COLLOCATED` removed; users should only use `CEED_BASIS_NONE`.
 - Remove unneeded pointer for `CeedElemRestrictionGetELayout`.
--- a/Doxyfile
+++ b/Doxyfile
@@ -48,7 +48,7 @@ PROJECT_NAME           = libCEED
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = v0.12.0
+PROJECT_NUMBER         = v1.0.0
 
 # Using the PROJECT_BRIEF tag one can provide an optional one line description
 # for a project that appears at the top of each page and should give viewer a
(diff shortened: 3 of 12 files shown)


CITATION.cff
```diff
@@ -1,7 +1,7 @@
 cff-version: 1.2.0
 title: "libCEED: Efficient Extensible Discretization"
-version: 0.12.0
-date-released: 2023-10-31
+version: 1.0.0
+date-released: 2025-08-15
```
jeremylt, line 4: ToDo: Update when we pick a release date

zatkins-dev:
It seems like the blockers are all resolved -- is the only remaining issue updating `libCEED_jll` with Yggdrasil?

jeremylt:
The main thing is that I don't have the permissions to do the Rust and Python language binding packaging, so this is only a PR than can be merged bp Jed when he has some free hours to handle all that

zatkins-dev:
Oh, we should get API tokens for that and use github actions to deploy automatically

jeremylt:
I know parts of Python are automated but I'm unsure if it is all of it

PR #1716, opened 2024-12-09
Gen AtPoints Updates

jeremylt:
2 small changes I discovered while working on AtPoints for *gen

1) Swapped to standard naming of the restriction kernels
2) Dropped looping templates, plan is to handle batches of points in gen like slabs of 3D


backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -215,13 +214,15 @@ static int CeedOperatorBuildKernelRestriction_Cuda_gen(std::ostringstream &code,
...
   CeedSize                  l_size;
+  CeedRestrictionType       rstr_type = CEED_RESTRICTION_STANDARD;
   CeedElemRestriction_Cuda *rstr_data;
   CeedElemRestriction       elem_rstr;
   CeedBasis                 basis;
 
   // Get field data
   CeedCallBackend(CeedOperatorFieldGetElemRestriction(op_field, &elem_rstr));
   if (elem_rstr != CEED_ELEMRESTRICTION_NONE) {
+    CeedCallBackend(CeedElemRestrictionGetType(elem_rstr, &rstr_type));
```
jrwrigh, line 225: Side note: `CeedRestrictionType` isn't documented currently. Honestly not sure what it is or does.

include/ceed/jit-source/cuda/cuda-gen-templates.h
```diff
@@ -144,16 +144,9 @@ inline __device__ void writeDofsStrided2d(SharedData_Cuda &data, const CeedInt e
 //------------------------------------------------------------------------------
 // L-vector -> E-vector, offsets provided
 //------------------------------------------------------------------------------
-// TODO: remove "Dofs" and "Quads" in the following function names?
-//   - readDofsOffset3d -> readOffset3d ?
```
jrwrigh, line 148: Hey, we do eventually come back to TODO comments! haha

jrwrigh (approved):

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
jeremylt: I should go ahead and toss a docstring on it. Short version is that its an enum of the different types of restrictions: standard, strided, oriented, curl oriented, and at points.

PR #1717, opened 2024-12-11
Fix hip shared atpoints

jeremylt:
Drop down to 1 elem per block like CUDA does (not sure why those diverged in the shared backends but this fixes the bug)


backends/hip-shared/ceed-hip-shared-basis.c
```diff
@@ -406,7 +406,7 @@ static int CeedBasisApplyAtPointsCore_Hip_shared(CeedBasis basis, bool apply_add
         CeedCallBackend(CeedRunKernelDimShared_Hip(ceed, is_transpose ? data->InterpTransposeAtPoints : data->InterpAtPoints, grid, thread_1d,
                                                    thread_1d, elems_per_block, shared_mem, interp_args));
       } else if (dim == 3) {
-        const CeedInt elems_per_block = CeedIntMax(block_size / (thread_1d * thread_1d), 1);
+        const CeedInt elems_per_block = 1;
```
jeremylt, line 409: This is the relevant change - the changes in CUDA are just formatting that annoyed me

PR #1718, opened 2025-01-02
Disable ROCm CI

jeremylt:
Temporarily disable ROCm CI due to the card being removed

See #1719 #1920 for future follow-up


jeremylt:
Ugh, we also need to update the Sycl backend for some changes in Sycl

jeremylt:
Disabling SYCL tests for now, as it seems more complex than a simple fix

PR #1721, opened 2025-01-03
Add non-tensor shared

jeremylt:
This is a prereq for the long awaited */gen non-tensor support.


backends/cuda-shared/kernels/cuda-shared-basis-nontensor.cu
```diff
@@ -0,0 +1,25 @@
...
+//
+// SPDX-License-Identifier: BSD-2-Clause
+//
+// This file is part of CEED:  http://github.com/ceed
+
+#include <ceed.h>
+#include <cuda.h>
+
+const int               MAX_SIZE = 52, MAX_DIM = 3;
+__constant__ CeedScalar c_B[MAX_SIZE * MAX_SIZE * MAX_DIM];
```
jeremylt, line 12: This limit means `P * Q * dim` cannot exceed `52 * 52 * 3`... which is rather small, but there's only so much shared memory to go around

jeremylt: We could fallback to ref in this case and warn that */gen is not going to work?

jrwrigh: Does the 52 limit include components? Ie for a quadratic line element, are we limited to `floor(52/3)` components?

jeremylt: excluding components - we don't duplicate rows in the interp and grad matrices for components

jeremylt: That number is device specific (no surprise there). I'm going to make an update to /cuda/shared to swap over to how /hip/shared handles this so we get a error instead of this silent memory access problem

jeremylt: Ok, my most recent approach seems to help a lot

jeremylt:
Investigating error on PETSc BP3:
```
Test: petsc-bps BP3, tet elements
  $ build/petsc-bps -ceed /gpu/cuda/shared -test -problem bp3 -degree 3 -ksp_max_it_clip 50,50 -simplex
ERROR: returncode = 15
Output: 
NO MESSAGE
FAIL: stderr
Output: 
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 8 FPE: Floating Point Exception,probably divide by zero
[0]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[0]PETSC ERROR: or see https://petsc.org/release/faq/#valgrind and https://petsc.org/release/faq/
[0]PETSC ERROR: or try https://docs.nvidia.com/cuda/cuda-memcheck/index.html on NVIDIA CUDA systems to find memory corruption errors
[0]PETSC ERROR: configure using --with-debugging=yes, recompile, link, and run 
[0]PETSC ERROR: to get more information on the crash.
Abort(59) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 59) - process 0
```

jeremylt:
Ok, I need to actually test the HIP code, but we've got everything working on the CUDA side of the house

PR #1724, opened 2025-01-13
Fix hip/shared NonTensor

jeremylt:
Thought I had these. Oops


zatkins-dev (approved):
I can't test at the moment, but I think think looks good to me. If you can run ratel tests with this locally, I would say it's good.

jeremylt:
The only issue should be shared data limits on whatever card is being used

PR #1726, opened 2025-01-14
Fallback if nontensor shared uses too much mem

### LINKED ISSUE #1725: Shared NonTensor Check Shared Mem Size
There has to be a way to query the amount of shared mem available and fall back to ref for NonTensor if the kernel we JiT needs too much memory.

jeremylt:
Fixes #1725 

We check the max required shared memory for a nontensor basis with `/gpu/*/shared` and fallback to `/gpu/*/ref` if the size needed exceeds the max available on the device

### DIFF
--- a/backends/cuda-shared/ceed-cuda-shared-basis.c
+++ b/backends/cuda-shared/ceed-cuda-shared-basis.c
@@ -628,11 +628,21 @@ int CeedBasisCreateH1_Cuda_shared(CeedElemTopology topo, CeedInt dim, CeedInt nu
   CeedBasis_Cuda_shared *data;
 
   CeedCallBackend(CeedBasisGetCeed(basis, &ceed));
-  CeedCallBackend(CeedCalloc(1, &data));
 
-  // Check max sizes
-  CeedCheck(dim <= 3, ceed, CEED_ERROR_BACKEND, "Backend does not implement nontensor bases with dim > 3");
-  CeedCheck(num_nodes * num_qpts * dim < 52 * 52 * 3, ceed, CEED_ERROR_BACKEND, "Backend does not implement nontensor bases with P * Q this large");
+  // Check shared memory size
+  {
+    Ceed_Cuda *cuda_data;
+
+    CeedCallBackend(CeedGetData(ceed, &cuda_data));
+    if (((size_t)num_nodes * (size_t)num_qpts * (size_t)dim + (size_t)CeedIntMax(num_nodes, num_qpts)) * sizeof(CeedScalar) >
+        cuda_data->device_prop.sharedMemPerBlock) {
+      CeedCallBackend(CeedBasisCreateH1Fallback(ceed, topo, dim, num_nodes, num_qpts, interp, grad, q_ref, q_weight, basis));
+      CeedCallBackend(CeedDestroy(&ceed));
+      return CEED_ERROR_SUCCESS;
+    }
+  }
+
+  CeedCallBackend(CeedCalloc(1, &data));
 
   // Copy basis data to GPU
   CeedCallBackend(CeedBasisGetNumQuadratureComponents(basis, CEED_EVAL_INTERP, &q_comp_interp));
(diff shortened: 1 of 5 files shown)


tests/t319-basis.c
```diff
@@ -116,7 +116,7 @@ int main(int argc, char **argv) {
   for (CeedInt dim = 1; dim <= 3; dim++) {
     CeedVector x_corners, x_from, x_to, u_from, u_to, du_to;
     CeedBasis  basis_x, basis_from, basis_to, basis_project;
-    CeedInt    p_from = 3, p_to = 4, q = 4, x_dim = CeedIntPow(2, dim), p_from_dim = CeedIntPow(p_from, dim), p_to_dim = CeedIntPow(p_to, dim);
+    CeedInt    p_from = 4, p_to = 5, q = 6, x_dim = CeedIntPow(2, dim), p_from_dim = CeedIntPow(p_from, dim), p_to_dim = CeedIntPow(p_to, dim);
```
jeremylt, line 119: This pushes the sizes up on my local card high enough to force a swap to ref

backends/cuda-shared/ceed-cuda-shared-basis.c
```diff
@@ -628,11 +628,20 @@ int CeedBasisCreateH1_Cuda_shared(CeedElemTopology topo, CeedInt dim, CeedInt nu
...
-  CeedCheck(dim <= 3, ceed, CEED_ERROR_BACKEND, "Backend does not implement nontensor bases with dim > 3");
-  CeedCheck(num_nodes * num_qpts * dim < 52 * 52 * 3, ceed, CEED_ERROR_BACKEND, "Backend does not implement nontensor bases with P * Q this large");
+  // Check shared memory size
+  {
+    Ceed_Cuda *cuda_data;
+
+    CeedCallBackend(CeedGetData(ceed, &cuda_data));
+    if (((size_t)num_nodes * (size_t)num_qpts * (size_t)dim + (size_t)CeedIntMax(num_nodes, num_qpts)) * sizeof(CeedScalar) >
+        cuda_data->device_prop.sharedMemPerBlock) {
+      CeedCallBackend(CeedBasisCreateH1Fallback(ceed, topo, dim, num_nodes, num_qpts, interp, grad, q_ref, q_weight, basis));
```
jeremylt, line 639: This is the first part of this new logic - before trying to compile a kernel, we check if the shared mem required exceeds the capacity of the device, and if so, we go use a fallback impl (ref)

interface/ceed-basis.c
```diff
@@ -600,6 +600,42 @@ static int CeedBasisApplyAtPoints_Core(CeedBasis basis, bool apply_add, CeedInt
...
+                              const CeedScalar *grad, const CeedScalar *q_ref, const CeedScalar *q_weight, CeedBasis basis) {
+  CeedInt P = num_nodes, Q = num_qpts, dim = 0;
+  Ceed    delegate;
+
+  CeedCall(CeedGetObjectDelegate(ceed, &delegate, "Basis"));
+  CeedCheck(delegate, ceed, CEED_ERROR_UNSUPPORTED, "Backend does not implement BasisCreateH1");
+
+  CeedCall(CeedReferenceCopy(delegate, &(basis)->ceed));
+  CeedCall(CeedBasisGetTopologyDimension(topo, &dim));
+  CeedCall(delegate->BasisCreateH1(topo, dim, P, Q, interp, grad, q_ref, q_weight, basis));
```
jeremylt, line 634: This is the second half here. Once the shared backend has kicked this back, we grab the delegate (ref), swap to that ceed in the basis struct, then do the ref basis setup.

backends/cuda-shared/ceed-cuda-shared-basis.c
zatkins-dev: Seems reasonable to me!

zatkins-dev (commented):
Looks good! Always good to make these little robustness improvements to save users the hassle.

zatkins-dev (approved):

PR #1727, opened 2025-01-14
Fix issue in block sizing for GPU shared basis

zatkins-dev:
This change fixes a crash when applying a `CeedBasis` with eval type `CEED_EVAL_WEIGHT` for order 2 elements.

I'm not sure if the cuda change is also necessary?

### DIFF
--- a/backends/cuda-shared/ceed-cuda-shared-basis.c
+++ b/backends/cuda-shared/ceed-cuda-shared-basis.c
@@ -486,18 +486,21 @@ static int CeedBasisApplyNonTensorCore_Cuda_shared(CeedBasis basis, bool apply_a
       }
     } break;
     case CEED_EVAL_WEIGHT: {
-      CeedInt Q;
+      CeedInt P, Q;
 
       CeedCheck(data->d_q_weight_1d, ceed, CEED_ERROR_BACKEND, "%s not supported; q_weights_1d not set", CeedEvalModes[eval_mode]);
+      CeedCallBackend(CeedBasisGetNumNodes(basis, &P));
       CeedCallBackend(CeedBasisGetNumQuadraturePoints(basis, &Q));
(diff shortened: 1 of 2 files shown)


jeremylt (approved):
Odd that P comes into play here since this doesn't really have anything to do with nodes, but :shrug:

backends/cuda-shared/ceed-cuda-shared-basis.c
```diff
@@ -486,18 +486,22 @@ static int CeedBasisApplyNonTensorCore_Cuda_shared(CeedBasis basis, bool apply_a
...
+      CeedInt thread = CeedIntMax(Q, P);
+
       void *weight_args[] = {(void *)&num_elem, (void *)&data->d_q_weight_1d, &d_v};
 
       {
         // avoid >512 total threads
-        CeedInt elems_per_block = CeedIntMin(ceed_Cuda->device_prop.maxThreadsDim[2], CeedIntMax(512 / Q, 1));
+        CeedInt elems_per_block = CeedIntMin(ceed_Cuda->device_prop.maxThreadsDim[2], CeedIntMax(512 / thread, 1));
         CeedInt grid            = num_elem / elems_per_block + (num_elem % elems_per_block > 0);
+        CeedInt shared_mem      = elems_per_block * thread * sizeof(CeedScalar);
```
jeremylt, line 502: CI doesn't like this variable since its unused

zatkins-dev: Yep, just fixed

jeremylt (approved):

jeremylt:
Figured out why: https://github.com/CEED/libCEED/blob/79881bbeb9d78b353436bc0c40893fd7a3eee562/include/ceed/jit-source/hip/hip-shared-basis-nontensor.h#L186

PR #1730, opened 2025-01-16
Update SetArray to keep old arrays for CEED_COPY_VALUES

### LINKED ISSUE #1728: Need `gen` support for restriction-only identity operators
Trying to apply such an operator (effectively a restricted copy operation) results in the error:
```
libCEED/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp:935 in CeedOperatorBuildKernel_Cuda_gen(): Backend does not implement restriction only identity operators
```

We use these operators for MPM.

Since these are such a specific type of operation, we could probably add support with a specialized kernel.

jeremylt:
Fixes #1728 

Digging in, I think this is the best fix


backends/cuda/ceed-cuda-common.c
```diff
@@ -53,10 +53,11 @@ static inline int CeedSetDeviceGenericArray_Cuda(Ceed ceed, const void *source_a
                                                  void *target_array_owned, void *target_array_borrowed, void *target_array) {
   switch (copy_mode) {
     case CEED_COPY_VALUES:
-      if (!*(void **)target_array_owned) CeedCallCuda(ceed, cudaMalloc(target_array_owned, size_unit * num_values));
-      if (source_array) CeedCallCuda(ceed, cudaMemcpy(*(void **)target_array_owned, source_array, size_unit * num_values, cudaMemcpyDeviceToDevice));
-      *(void **)target_array_borrowed = NULL;
-      *(void **)target_array          = *(void **)target_array_owned;
```
jeremylt, line 59: here is where we were always dumping the old borrowed array

python/tests/test-1-vector.py
```diff
@@ -359,7 +359,7 @@ def test_126(ceed_resource, capsys):
     a = np.arange(10, 10 + n, dtype=ceed.scalar_type())
     x.set_array(a, cmode=libceed.USE_POINTER)
 
-    a2 = np.arange(10, n, dtype=ceed.scalar_type())
+    a2 = np.arange(0, n, dtype=ceed.scalar_type())
```
jeremylt, line 362: That's an embarrassing oops on my part from years ago

PR #1734, opened 2025-01-24
Reduce memory overhead for HIP backends on MI300A GPUs

zatkins-dev:
Prevents double allocations for `CeedVector` when using HIP vector with unified addressing and XNACK.

Also, updates more of the HIP vector operations to use `hipBLAS` functions rather than custom kernels.


backends/hip-ref/ceed-hip-ref-basis.c
```diff
@@ -214,7 +220,8 @@ static int CeedBasisApplyAtPointsCore_Hip(CeedBasis basis, bool apply_add, const
     CeedCallBackend(CeedBasisGetNumNodes(basis, &num_nodes));
     length =
         (CeedSize)num_elem * (CeedSize)num_comp * (t_mode == CEED_TRANSPOSE ? (CeedSize)num_nodes : ((CeedSize)max_num_points * (CeedSize)q_comp));
-    CeedCallHip(ceed, hipMemset(d_v, 0, length * sizeof(CeedScalar)));
+    if (hip_data->has_unified_addressing) memset(d_v, 0, length * sizeof(CeedScalar));
+    else CeedCallHip(ceed, hipMemset(d_v, 0, length * sizeof(CeedScalar)));
```
jeremylt, line 224: Is there a reason we don't use the `CeedVector` interface for this (I can't recall)? I suspect that would likely be a better fix?

zatkins-dev: I kinda assumed that `length` didn't match the vector length. So setting the whole vector to zero would be problematic. I didn't dig into it, though

backends/hip/ceed-hip-common.h
```diff
@@ -72,6 +72,7 @@ typedef struct {
   hipblasHandle_t        hipblas_handle;
   struct hipDeviceProp_t device_prop;
   int                    opt_block_size;
+  int                    has_unified_addressing;
```
jeremylt, line 75: int?

backends/hip/ceed-hip-compile.cpp
```diff
@@ -56,8 +56,8 @@ int CeedCompile_Hip(Ceed ceed, const char *source, hipModule_t *module, const Ce
   // With ROCm 4.5, need to include these definitions specifically for hiprtc (but cannot include the runtime header)
   else {
     code << "#include <stddef.h>\n";
-    code << "#define __forceinline__ inline __attribute__((always_inline))\n";
-    code << "#define HIP_DYNAMIC_SHARED(type, var) extern __shared__ type var[];\n";
+    // code << "#define __forceinline__ inline __attribute__((always_inline))\n";
+    // code << "#define HIP_DYNAMIC_SHARED(type, var) extern __shared__ type var[];\n";
```
jeremylt, line 60: dead code

backends/hip/ceed-hip-common.h
zatkins-dev: That's the type in the `hipDeviceProps_t` struct, I think maybe they just don't like bools

backends/hip/ceed-hip-compile.cpp
zatkins-dev: I actually didn't mean to commit that, although I think that those lines are unnecessary for modern ROCm

backends/hip-ref/ceed-hip-ref-basis.c
jeremylt: I think this might predate my efforts to strictly use the API between objects internally and I just missed it in the transition. I think we're probably safe to use the API here

backends/hip/ceed-hip-common.h
jeremylt: Fair enough, if HIP does it, we can tolerate it

backends/hip-ref/ceed-hip-ref-basis.c
```diff
@@ -35,20 +35,12 @@ static int CeedBasisApplyCore_Hip(CeedBasis basis, bool apply_add, const CeedInt
...
-    CeedInt  num_comp, q_comp, num_nodes, num_qpts;
-    CeedSize length;
-
-    CeedCallBackend(CeedBasisGetNumComponents(basis, &num_comp));
-    CeedCallBackend(CeedBasisGetNumQuadratureComponents(basis, eval_mode, &q_comp));
-    CeedCallBackend(CeedBasisGetNumNodes(basis, &num_nodes));
-    CeedCallBackend(CeedBasisGetNumQuadraturePoints(basis, &num_qpts));
-    length = (CeedSize)num_elem * (CeedSize)num_comp * (t_mode == CEED_TRANSPOSE ? (CeedSize)num_nodes : ((CeedSize)num_qpts * (CeedSize)q_comp));
-    CeedCallHip(ceed, hipMemset(d_v, 0, length * sizeof(CeedScalar)));
+  else {
```
jeremylt, line 38: I lost the comment here, but the style is to use {} for both clauses if one of them needs it
```

File: backends/hip-ref/ceed-hip-ref-basis.c [hunk 5]
```diff
@@ -202,19 +196,10 @@ static int CeedBasisApplyAtPointsCore_Hip(CeedBasis basis, bool apply_add, const
...
-
-    CeedCallBackend(CeedBasisGetNumComponents(basis, &num_comp));
-    CeedCallBackend(CeedBasisGetNumQuadratureComponents(basis, eval_mode, &q_comp));
-    CeedCallBackend(CeedBasisGetNumNodes(basis, &num_nodes));
-    length =
-        (CeedSize)num_elem * (CeedSize)num_comp * (t_mode == CEED_TRANSPOSE ? (CeedSize)num_nodes : ((CeedSize)max_num_points * (CeedSize)q_comp));
-    CeedCallHip(ceed, hipMemset(d_v, 0, length * sizeof(CeedScalar)));
+  else {
+    // Clear v for transpose operation
+    if (is_transpose) CeedCallBackend(CeedVectorSetValue(v, 0.0));
```
jeremylt, line 201: Can you add this fix to CUDA too please?

zatkins-dev: added!

backends/hip/ceed-hip-compile.cpp
```diff
@@ -167,6 +167,9 @@ int CeedGetKernel_Hip(Ceed ceed, hipModule_t module, const char *name, hipFuncti
 // Run HIP kernel
 //------------------------------------------------------------------------------
 int CeedRunKernel_Hip(Ceed ceed, hipFunction_t kernel, const int grid_size, const int block_size, void **args) {
+  Ceed_Hip *data;
+
+  CeedCallBackend(CeedGetData(ceed, &data));
```
jeremylt, line 172: Why this change?

zatkins-dev: oops, that's dead code. removed

backends/hip-ref/ceed-hip-ref-basis.c
```diff
@@ -114,6 +107,7 @@ static int CeedBasisApplyAtPointsCore_Hip(CeedBasis basis, bool apply_add, const
   const CeedScalar *d_x, *d_u;
   CeedScalar       *d_v;
   CeedBasis_Hip    *data;
+  Ceed_Hip         *hip_data;
```
jeremylt, line 110: another stray

backends/hip-ref/ceed-hip-ref-basis.c
```diff
@@ -126,6 +120,7 @@ static int CeedBasisApplyAtPointsCore_Hip(CeedBasis basis, bool apply_add, const
   }
 
   CeedCallBackend(CeedBasisGetCeed(basis, &ceed));
+  CeedCallBackend(CeedGetData(ceed, &hip_data));
```
jeremylt, line 123: and here

backends/hip-ref/ceed-hip-ref-vector.c
```diff
@@ -95,7 +95,21 @@ static inline int CeedVectorSyncD2H_Hip(const CeedVector vec) {
...
+  Ceed_Hip *hip_data;
+
+  CeedCallBackend(CeedGetData(CeedVectorReturnCeed(vec), &hip_data));
+
+  // Unified memory doesn't require syncing, always use the host pointer
+  if (hip_data->has_unified_addressing) {
+    CeedVector_Hip *impl;
+
+    CeedCallBackend(CeedVectorGetData(vec, &impl));
+    CeedCallHip(CeedVectorReturnCeed(vec), hipDeviceSynchronize());
```
zatkins-dev, line 108: Ratel seems to work fine without this line, and is faster

jedbrown: Does `CeedVectorSyncArray` mean that one could immediately start an `MPI_Send`? If the host doesn't know that the previous kernel (writing to the array) has completed, then it would be racy to call `MPI_Send`. (Might be rare to trip, but we don't want that kind of bug.)

If our sends are using a kernel for packing (on the same stream), then the host doesn't need to know when the earlier stuff completes, but we still need to sync after the packing kernel.

zatkins-dev: that's a fair point, I think that we need to be a bit more careful and only sync when the host needs the data. Otherwise this acts as a hard sync with the GPU, which seems to have performance impacts.

jrwrigh:
FYI, generally prefer [rebase to merge for dev branches](https://yarchive.net/comp/linux/git_merges_from_upstream.html). It doesn't matter for squash-merges (the commit history gets nuked anyways), but for normal merges it helps the git history be more regular.

zatkins-dev:
> FYI, generally prefer [rebase to merge for dev branches](https://yarchive.net/comp/linux/git_merges_from_upstream.html). It doesn't matter for squash-merges (the commit history gets nuked anyways), but for normal merges it helps the git history be more regular.

Yeah generally I agree - I need to strip down this branch and rebuild it probably, it's currently a mess due to changes at the AMD workshop.

jeremylt:
I think we want to merge this before the review so we can have libCEED 0.13 and Ratel 0.4 with this? Is the question of the sync call the big blocker right now?

jeremylt:
Restriction offset arrays may also want this

zatkins-dev:
> I think we want to merge this before the review so we can have libCEED 0.13 and Ratel 0.4 with this? Is the question of the sync call the big blocker right now?

I think so? To be honest, I've been more focused on MPM work in Ratel and haven't had much time to work on this. 

If you have more time and desire to extract the changes into a clean branch, I'd be happy to review. Otherwise, I probably won't get to it until late this week at the earliest, more likely next week.

jeremylt:
No rush - it was just a thought that crossed my mind as we look forward to future activities for libCEED

jeremylt:
I'd like to get this in for the release. If you're still tight on time, I can tidy up the branch. The discussion above about syncing seems to be the real sticking point though.

zatkins-dev:
> I'd like to get this in for the release. If you're still tight on time, I can tidy up the branch. The discussion above about syncing seems to be the real sticking point though.

if you have time, that would be great. Ultimately, I think we need to sync when a sync is requested for correctness.

jeremylt:
See https://github.com/CEED/libCEED/pull/1788

jeremylt:
Transferred to #1788

PR #1738, opened 2025-01-31
Guard divide by zero

jeremylt:
How about this as a fix?


jeremylt:
Confirmed this fixed it

zatkins-dev:
Do we need to add any flops if the operator has no elements?

jeremylt:
Good point, fixed

zatkins-dev (approved):
lgtm!

PR #1750, opened 2025-02-12
gpu - prefer cu/hipBlas over handrolls

jeremylt:
Ok, this swaps to faster cu/hipBlas utils over our hand rolled kernels.

The functions called here are in newer versions of cu/hipBlas

- ROCm 6.0 released Dec 2023
- CUDA 12.0 released Dec 2022

I don't think we have users who rely upon older versions of ROCm/CUDA, though we could add some `#if (CUDA_VERSION < ....)` and `#if (HIP_VERSION < ....)` and use the hand rolled versions if we think its important.


jeremylt:
Ok, since Noether uses an older CUDA, I added the logic to preserve the old behavior for old CUDA. We can always take that out when appropriate in the future.

jeremylt:
Noether has an older CUDA, my laptop has a newer CUDA, and Odyssey has a newer ROCM - tests pass on all of them

jedbrown:
Looks good. We can update CUDA on Noether too (Debian testing is currently on cuda-12.2). Should do that in the next few months regardless.

PR #1759, opened 2025-02-19
Drop IBM Test for Now

jeremylt:
https://github.com/uraimo/run-on-arch-action/issues/160

Run on Arch know about the issue. One possible workaround is to use Clang instead of GCC.


jeremylt:
Ok, workaround does not work. changing this PR to delete this for now.

PR #1762, opened 2025-02-26
Mixed Tensor/NonTensor for Gen

jeremylt:
Last step in the *gen refactor. This will allow us to run operators that have a mix of tensor product and non-tensor bases. 2D is easier, 3D will take more thought.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
+++ b/backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
@@ -685,9 +685,8 @@ static int CeedOperatorBuildKernelQFunction_Cuda_gen(std::ostringstream &code, C
         if (is_at_points) {
           // Accumulator for point data
           code << "    CeedScalar r_c" << var_suffix << "[num_comp" << var_suffix << "*" << (max_dim >= 3 ? Q_name : "1") << "];\n";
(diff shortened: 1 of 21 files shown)


jeremylt:
Sample "flat" version

```C++
template <int P_1D, int Q_1D>
inline __device__ void WeightTensor2dFlattened(SharedData_Cuda &data, const CeedScalar *__restrict__ q_weight_1d, CeedScalar *w) {
  const int max = P_1D < Q_1D ? P_1D : Q_1D;

  WeightTensor2d_Core<Q_1D>(data, data.t_id_x % max, data.t_id_x / max, q_weight_1d, w);
}
```

jeremylt:
Ok, I need to check on the memory in the `data.slice` for this approach

Also.... I'm not sure how 3D will be tackled

jeremylt:
Ok, there's going to need to be more disentangling, as the operator I'm trying to target has different dim for different nodal spaces

jeremylt:
Ok, separate dims for each field now, but theres some bug that's giving wrong results

jeremylt:
ugh, T_1D is wrong for this strategy. Pondering

jeremylt:
Getting closer. 2D Tensor + 3D NonTensor is working now. Need 3D Tensor + 3D NonTensor next.

jeremylt:
I'm starting to think this was not worth any of the time spent here - I think the restrictions also need to be hacked around with. I'm going to abandon this effort for now and switch to AtPoints assembly

jeremylt:
Ok, ready for review. Plan is squash + merge

PR #1764, opened 2025-02-28
Add `CeedElemRestrictionGetMinPointsInElement` and `CeedElemRestrictionGetMinMaxPointsInElement`

zatkins-dev:
Adds the ability to query minimum points per element and unifies min/max computation under a single function to reduce repeat work.


jeremylt (approved):

jeremylt:
Good call with the tests, thanks

zatkins-dev:
np, I figured it was good practice

PR #1770, opened 2025-03-05
Add ex03 mass+diff

jeremylt:
This adds a simple usage of multiple eval modes per input/output vector, which allows us to check for excessive restriction calls.

This is what I was thinking for BP1+3 and BP2+4 as well, but it doesn't need to be me who adds it.

### DIFF
--- a/.github/workflows/c-fortan-test-ppc64le.yml
+++ b/.github/workflows/c-fortan-test-ppc64le.yml
@@ -37,4 +37,4 @@ jobs:
           uname -a
           make info
           make -j
-          make prove -j
+          make prove -j search="t5 ex"
--- a/examples/ceed/ex3-volume.h
+++ b/examples/ceed/ex3-volume.h
@@ -14,73 +14,82 @@ struct BuildContext {
 
 /// libCEED Q-function for building quadrature data for a mass + diffusion operator
 CEED_QFUNCTION(build_mass_diff)(void *ctx, const CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out) {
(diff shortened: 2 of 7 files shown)


examples/ceed/ex3-volume.h
```diff
@@ -0,0 +1,151 @@
...
+      break;
+  }
+  return CEED_ERROR_SUCCESS;
+}
+
+/// libCEED Q-function for applying a mass + diffusion operator
+CEED_QFUNCTION(apply_mass_diff)(void *ctx, const CeedInt Q, const CeedScalar *const *in, CeedScalar *const *out) {
+  struct BuildContext *build_data = (struct BuildContext *)ctx;
+  // in[0], out[0] have shape [1,   nc=1, Q]
+  // in[1], out[1] have shape [dim, nc=1, Q]
```
jrwrigh, line 92: `nc` is number of components? Should maybe change this to more human name? Assuming this is used in other examples too.

.github/workflows/c-fortan-test-ppc64le.yml
```diff
@@ -37,4 +37,4 @@ jobs:
           uname -a
           make info
           make -j
-          make prove -j
+          make prove -j search="t5 ex"
```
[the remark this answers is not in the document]
jrwrigh: If it is intentional, should this be a separate PR for clearer "revert" later?

examples/ceed/ex3-volume.h
jeremylt: For some reason, we didn't used the VLA casts here, but I think it makes the code clearer, so I added it

PR #1791, opened 2025-03-21
Include tensor basis header for pure-tensor operators

### LINKED ISSUE #1790: HIP gen tensor-only operators always fall-back
Looks like the wrong bool was used here:

https://github.com/CEED/libCEED/blob/f97c94cc964affa1826363c99dc41a0292e8d752/backends/hip-gen/ceed-hip-gen-operator-build.cpp#L1207-L1209

zatkins-dev:
Fixes #1790


jeremylt (approved):
Can you check the same file for CUDA please 

Thanks for the fast fix!

zatkins-dev:
cuda is good!

PR #1799, opened 2025-04-11
AtPoints assembly for CPU

jeremylt:
WIP here, but far enough along to make a PR

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -1525,6 +1525,212 @@ static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Ref(CeedOperator op, Ce
 }
 
 //------------------------------------------------------------------------------
+// Assemble Operator AtPoints
+//------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


backends/ref/ceed-ref-operator.c
```diff
@@ -1552,6 +1552,248 @@ static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Ref(CeedOperator op, Ce
...
+              // LCOV_EXCL_STOP
+            }
+          }
+          // ---- Copy output
+          {
+            const CeedScalar *output;
+
+            CeedCallBackend(CeedVectorGetArrayRead(impl->e_vecs_out[j], CEED_MEM_HOST, &output));
+            for (CeedInt k = 0; k < elem_size_active; k++) {
+              assembled[offset + e * elem_size_active * elem_size_active + s * elem_size_active + k] += output[k];
```
jeremylt, line 1769: need to address multi-component case

jeremylt: I think I've got it with the update but need to make a test to check

jeremylt: nope, the test was bad, lol

PR #1800, opened 2025-04-18
Fix edge case in offset padding for empty cells

zatkins-dev:
Fixes a weird edge case where if the first cell is empty, then the offset for all points is out of bounds.


jeremylt:
Something is wrong with GitHub's mac image

jeremylt:
Ok, give the pipelines some time, but I'll merge this fix to main in a sec and you can rebase: #1801

Issue #1774, opened 2025-03-12
Adding Python Example for PETSc Bake-Off Problem


tylercollins5737:
Hello libCEED Team,

I noticed an open issue #1763 and I'm also interested in adding to the library of python examples, specifically starting with the first PETSc bake-off problem. It was mentioned that some advanced methods might not be available but could be added. Should I just go for it and report any missing or needed methods?

Appreciate your time and any suggestions!

Cheers,
Tyler

jedbrown:
Yup, go ahead and get started. You'll probably use petsc4py for generating the mesh and handling the solver, with libceed for applying the operator. You may need to look at the petsc4py examples at least as much as libceed examples to see how to implement things like matrix-free operators (which will call `ceed.Operator.apply`).

Let us know if you get stuck. It's usually best to create a draft PR with your work along the way so you have a place to ask questions where we'll see it in context.

tylercollins5737:
> Yup, go ahead and get started. You'll probably use petsc4py for generating the mesh and handling the solver, with libceed for applying the operator. You may need to look at the petsc4py examples at least as much as libceed examples to see how to implement things like matrix-free operators (which will call `ceed.Operator.apply`).
> 
> Let us know if you get stuck. It's usually best to create a draft PR with your work along the way so you have a place to ask questions where we'll see it in context.

Sounds good and thank you for the tips!


> I thought [#1763](https://github.com/CEED/libCEED/issues/1763) was about that particular person offering to make more Python examples? Worth clarifying in that issue

That was a good point, and I got clarification so we aren't overlapping in our efforts.

arrowguy234:
@tylercollins5737 Hii, I am also interested in implementing Bakeoff problem in python , So can you specify which Bakeoff problem you are implementing so that our work doesn't overlap, I am flexible to discuss the example.

PR #1806, opened 2025-04-19
Python PETSc Example BP1

tylercollins5737:
Hello!

This is a Python implementation of bpsraw.c for BP1 and was initially discussed in #1774 .  I'm aiming to recreate the benchmark behavior using libceed and petsc4py. I appreciate any feedback or indications I'm headed in the right direction!

It does run, but this is a work in progress. Remaining tasks:
- [x] fix the matrix operations to work like the MatShell C implementation
- [x] get PETSc Vec scatter working
- [x] QFunction 
- [x] Error function
- [x] add meaningful comments
- [x] improve the printed output metrics 
- [x] create a test

### DIFF
--- a/examples/python/bps_test.py
+++ b/examples/python/bps_test.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017-2026, Lawrence Livermore National Security, LLC and other CEED contributors
+# All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# This file is part of CEED:  http://github.com/ceed
+
+import pytest
+from argparse import Namespace
+
+# The BP examples use PETSc, unlike ex1-ex3, so skip them where it is unavailable
+pytest.importorskip("petsc4py")
+import bpsraw  # noqa: E402
+
+# -------------------------------------------------------------------------------
+
+
+def test_401(ceed_resource):
+    args = Namespace(
+        ceed=ceed_resource,
+        problem='bp1',
+        degree=1,
+        q_extra=None,
+        local=1000,
+        test=True,
+        benchmark=False,
+        write_solution=False,
+        ksp_max_it_clip=[15, 15],
+    )
+    assert bpsraw.example_bps(args) == 0
+
+# -------------------------------------------------------------------------------
+
+
+def test_402(ceed_resource):
+    args = Namespace(
+        ceed=ceed_resource,
+        problem='bp1',
+        degree=2,
+        q_extra=None,
+        local=1000,
+        test=True,
+        benchmark=False,
+        write_solution=False,
+        ksp_max_it_clip=[15, 15],
+    )
+    assert bpsraw.example_bps(args) == 0
+
+# -------------------------------------------------------------------------------
+
+
+def test_403(ceed_resource):
+    args = Namespace(
+        ceed=ceed_resource,
+        problem='bp1',
+        degree=3,
+        q_extra=None,
+        local=1000,
+        test=True,
+        benchmark=False,
+        write_solution=False,
+        ksp_max_it_clip=[15, 15],
+    )
+    assert bpsraw.example_bps(args) == 0
+
+# -------------------------------------------------------------------------------
--- a/examples/python/bps_test.py
+++ b/examples/python/bps_test.py
@@ -16,7 +16,7 @@ import bpsraw  # noqa: E402
 # -------------------------------------------------------------------------------
 
 
-def test_401(ceed_resource):
+def test_bp1_degree_1(ceed_resource):
     args = Namespace(
         ceed=ceed_resource,
         problem='bp1',
@@ -28,12 +28,12 @@ def test_401(ceed_resource):
         write_solution=False,
         ksp_max_it_clip=[15, 15],
     )
-    assert bpsraw.example_bps(args) == 0
+    bpsraw.example_bps(args)
 
 # -------------------------------------------------------------------------------
 
 
-def test_402(ceed_resource):
+def test_bp1_degree_2(ceed_resource):
     args = Namespace(
         ceed=ceed_resource,
         problem='bp1',
@@ -45,12 +45,12 @@ def test_402(ceed_resource):
         write_solution=False,
         ksp_max_it_clip=[15, 15],
     )
-    assert bpsraw.example_bps(args) == 0
+    bpsraw.example_bps(args)
 
 # -------------------------------------------------------------------------------
 
 
-def test_403(ceed_resource):
+def test_bp1_degree_3(ceed_resource):
     args = Namespace(
         ceed=ceed_resource,
         problem='bp1',
@@ -62,6 +62,6 @@ def test_403(ceed_resource):
         write_solution=False,
         ksp_max_it_clip=[15, 15],
     )
-    assert bpsraw.example_bps(args) == 0
+    bpsraw.example_bps(args)
 
 # -------------------------------------------------------------------------------
--- a/examples/python/bpsraw.py
+++ b/examples/python/bpsraw.py
@@ -1,0 +1,659 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017-2026, Lawrence Livermore National Security, LLC and other CEED contributors.
+# All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# This file is part of CEED:  http://github.com/ceed
+#
+# libCEED example solving a CEED benchmark problem (BP) with PETSc
+#
+# This is a Python port of examples/petsc/bpsraw.c and uses the same QFunctions,
+# from examples/petsc/qfunctions/bps.
+#
+# Sample runs:
+#
+#     python bpsraw.py
+#     python bpsraw.py -problem bp1 -degree 3
+#     mpiexec -n 4 python bpsraw.py -problem bp1 -degree 3 -ceed /gpu/cuda
+
+import argparse
+import math
+import os
+import sys
+
+import numpy as np
+import petsc4py
+from mpi4py import MPI
+
+import libceed
+import ex_common as common
+
+# BP types, matching the bp_types[] enum used by bpsraw.c
+BP_TYPES = ("bp1", "bp2", "bp3", "bp4", "bp5", "bp6")
+
+# Per-BP settings, matching the bp_options[] table in bpsraw.c
+BP_OPTIONS = {
+    "bp1": {
+        "num_comp_u": 1,
+        "q_data_size": 1,
+        "q_extra": 1,
+        "source": "bp1.h",
+        "setup_geo": "SetupMassGeo",
+        "setup_rhs": "SetupMassRhs",
+        "apply": "Mass",
+        "error": "Error",
+        "in_mode": libceed.EVAL_INTERP,
+        "out_mode": libceed.EVAL_INTERP,
+        "q_mode": libceed.GAUSS,
+    },
+}
+
+
+def int_pair(value):
+    """Parse a pair of comma separated integers, as PetscOptionsIntArray does"""
+    entries = [int(entry) for entry in value.split(",")]
+    if len(entries) != 2:
+        raise argparse.ArgumentTypeError("expected two comma separated "
+                                         f"integers, got '{value}'")
+    return entries
+
+
+def parse_arguments(argv=None):
+    """Parse the options owned by this example
+
+    Options that are not recognized here are left for PETSc.
+
+    Args:
+        argv: Argument list, defaults to sys.argv[1:]
+
+    Returns:
+        tuple: (parsed arguments, remaining arguments for PETSc)
+    """
+    parser = argparse.ArgumentParser(
+        description="CEED BPs in PETSc", add_help=False)
+    parser.add_argument("-problem", default="bp1", choices=BP_TYPES,
+                        help="CEED benchmark problem to solve")
+    parser.add_argument("-degree", type=int, default=1,
+                        help="Polynomial degree of tensor product basis")
+    parser.add_argument("-q_extra", type=int, default=None,
+                        help="Number of extra quadrature points")
+    parser.add_argument("-ceed", default="/cpu/self",
+                        help="CEED resource specifier")
+    parser.add_argument("-local", type=int, default=1000,
+                        help="Target number of locally owned nodes per process")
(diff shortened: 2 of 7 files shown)


valeriabarra:
@tylercollins5737 , thank you for your work. Are you able to address the reviewer's comments?

examples/python/bpsraw.py
```diff
@@ -0,0 +1,155 @@
...
+        e_vec = ceed.Vector(num_dofs)
+        r_vec = ceed.Vector(num_dofs)
+        e_vec.set_value(0.0)
+        r_vec.set_value(0.0)
+        with e_vec.array_write() as arr:
+            arr[i] = 1.0
+        op.apply(e_vec, r_vec)
+        with r_vec.array_read() as r_arr:
+            matrix.setValues(range(num_dofs), [i], r_arr)
+    matrix.assemble()
```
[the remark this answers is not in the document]
tylercollins5737: Apologies for leaving this a bit stale. Addressing this led to a lot of other changes. I had trouble setting the matrix type to "shell" and applying context, kept getting segmentation faults. So, instead of using the "shell" type, I used the "python" type and set the context through a python class. Pulled that implementation from the petsc4py 2DPoisson example (about half way down) --> [link](https://petsc.org/release/petsc4py/petsc_python_types.html#petsc-python-matrix-type).  I'll tidy up what I have a submit an update soon.

tylercollins5737:
@valeriabarra I was able to get it working using the "python" matrix type, MatAIJ was not the best. I believe the "python" type and MatCreateShell can achieve the same thing.

examples/python/bpsraw.py
```diff
@@ -0,0 +1,423 @@
...
+        self._xglob = np.zeros(ndofs, dtype=float)
+        self.u = ceed.Vector(ndofs)
+        self.v = ceed.Vector(ndofs)
+
+    def mult(self, A: PETSc.Mat, x: PETSc.Vec, y: PETSc.Vec):
+        lo, hi = x.getOwnershipRange()
+        self._xglob.fill(0.0)
+        self._xglob[lo:hi] = x.array_r
+        self.u.set_array(self._xglob,cmode=libceed.USE_POINTER)
+        self.v.set_value(0.0)
```
jeremylt, line 91:
```suggestion
        self.v.set_value(0.0)

```
this shouldn't be needed

examples/python/bpsraw.py
```diff
@@ -0,0 +1,423 @@
...
+        self.op.apply(self.u, self.v)
+        with self.v.array_read() as va:
+            y.setValues(range(lo, hi), va[lo:hi])
+        y.assemble()
```
jeremylt, line 95: Ideally, you'd set v to use y's array before you apply the libCEED operator

examples/python/bpsraw.py
```diff
@@ -0,0 +1,423 @@
...
+        for j in range(p[1]):
+            for k in range(p[2]):
+                if [i, j, k] == list(i_rank):
+                    return start
+                m_nodes = GlobalNodes(p, (i, j, k), degree, mesh_elem)
+                start += m_nodes[0] * m_nodes[1] * m_nodes[2]
+    return -1
+
+def CreateRestriction(ceed, mesh_elem, P, num_comp, ndofs, offsets):
+    return ceed.ElemRestriction(mesh_elem, P, num_comp, 1, ndofs, offsets, cmode=libceed.USE_POINTER)
```
jeremylt, line 40: If this just wraps another function call, I think its clearer to use the function call itself

examples/python/bpsraw.py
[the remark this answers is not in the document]
jeremylt: Overall this is looking like the right approach. The big thing is to try to avoid copying around data during frequent operations, like the matrix application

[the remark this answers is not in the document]
tylercollins5737: Thank you for all the suggestions! Definitely, keeping the data in place would mean less overhead and better efficiency

examples/python/bpsraw.py
tylercollins5737: Totally agree, I have another CreateRestriction function that reflects C but its a WIP

examples/python/bpsraw.py
tylercollins5737: That makes more sense, thank you!

jeremylt:
Hi @tylercollins5737, I'm hoping to have this ready in case next semester's students want to build on this for BP 2-6. This is good work so far and will be very helpful to future users. Is there anything I can do to help here? I'm happy to help out on any pieces that have you stuck so we can get this moving again.

zatkins-dev:
@tylercollins5737 are you planning on coming back to this? I think there's a lot of good work here! If not, it may be time to close this PR.

tylercollins5737:
> @tylercollins5737 are you planning on coming back to this? I think there's a lot of good work here! If not, it may be time to close this PR.

@zatkins-dev  Yes! I should have have a pretty complete update soon, ideally by the end of the weekend.

zatkins-dev:
Awesome! Happy to keep it open then, thanks!

tylercollins5737:
- QFunctions are in the PR now (`qfunctions/bps/bp1.h`, `common.h`, plus `qfunctions.c`), using the `ex3_volume.py` approach you pointed me to. The gallery has `Mass3DBuild`/`MassApply` but nothing like `SetupMassRhs` or `Error`, so those have to be C QFunctions. 
- MatMult no longer copies data. The PETSc `Vec` arrays go to libCEED with `CEED_USE_POINTER`, on the host and on the device.
- Added tests (BP1, degrees 1–3) and matched the CLI to `bpsraw.c` (`-problem bp1`, `-ksp_max_it_clip`, `-write_solution`).
- Output matches `bpsraw.c` on CPU and GPU, across 1, 2 and 4 ranks and degrees 1–4. For example at 1 rank, degree 3: 16 iterations, max error 1.730043e-09, same as the C.

Two notes:
1. I copied `bp1.h`/`common.h` into `examples/python/qfunctions/bps/`, the way ex1–ex3 keep their own copies, so the example is self contained. 
2. `CeedVectorTakeArray` isn't wrapped in the Python bindings, so the borrowed arrays get rebound on the next MatMult instead of detached. Works as written, but not the same as the C.

examples/python/ex_test.py
```diff
@@ -11,6 +11,7 @@
 import ex1_volume
 import ex2_surface
 import ex3_volume
+import bpsraw
```
jeremylt: since this file relies upon PETSc4Py, we'll want to test it separately from ex1-3, which only need libCEED

examples/python/bpsraw.py
```diff
@@ -0,0 +1,659 @@
...
+        self.l_to_g.end(X, self.X_loc, addv=PETSc.InsertMode.INSERT, mode=PETSc.Scatter.Mode.REVERSE)
+
+        # Setup libCEED vectors
+        x_handle = self.set_ceed_array(self.x_ceed, self.X_loc, "r")
+        y_handle = self.set_ceed_array(self.y_ceed, self.Y_loc, "w")
+
+        # Apply libCEED operator
+        self.op_apply.apply(self.x_ceed, self.y_ceed)
+
+        # Restore arrays; the C also detaches them with CeedVectorTakeArray, which the Python bindings do not
```
jeremylt: We really should just add CeedVectorTakeArarry to the Python bindings - @zatkins-dev or I can hit that, though dunno how long it'll take to publish that

examples/python/bpsraw.py
```diff
@@ -0,0 +1,659 @@
...
+petsc4py.init([sys.argv[0]] + _petsc_argv)
+from petsc4py import PETSc  # noqa: E402
+
+# petsc4py exposes the converged reasons as integer attributes, not an enum
+KSP_CONVERGED_REASONS = {value: name for name, value
+                         in vars(PETSc.KSP.ConvergedReason).items()
+                         if isinstance(value, int)}
+
+
+def Split3(size, reverse=False):
```
jeremylt: we should use the same function naming, style throughout. Also a brief docstring for each would be handy (though BPsRaw probably omitted that? If so, then no worries here)

jeremylt:
I have just a handful of comments, but this looks great!

examples/python/ex_test.py
tylercollins5737: Sounds good, I'll create a separate test.py

examples/python/bpsraw.py
tylercollins5737: Ah, yeah I forgot to rename them. I had them styled like C so I could spot differences easier. BPsRaw didn't really have any, but I'll fill them out.

tylercollins5737:
So the BP tests skip when petsc4py isn't installed, and right now that's every CI setup because GitHub Actions has no PETSc, and GitLab CI has PETSc but not petsc4py. So they only really run locally at the moment.

examples/python/bpsraw.py
tylercollins5737: Just let me know about this. I can open a PR for it.

examples/python/bps_test.py
```diff
@@ -0,0 +1,67 @@
...
+        problem='bp1',
+        degree=1,
+        q_extra=None,
+        local=1000,
+        test=True,
+        benchmark=False,
+        write_solution=False,
+        ksp_max_it_clip=[15, 15],
+    )
+    assert bpsraw.example_bps(args) == 0
```
jeremylt: Hmm, we either should use `assert` in the other test cases or not use it here. I *think* but am not sure that not having the `assert` results in this test just returning the error code of `example_bps`

examples/python/bps_test.py
```diff
@@ -0,0 +1,67 @@
...
+from argparse import Namespace
+
+# The BP examples use PETSc, unlike ex1-ex3, so skip them where it is unavailable
+pytest.importorskip("petsc4py")
+import bpsraw  # noqa: E402
+
+# -------------------------------------------------------------------------------
+
+
+def test_401(ceed_resource):
```
jeremylt: I'd pick a clearer set of test names, like `test_bp_1_degree_1` or such? Once this merges, I'll be adding an issue for a student to expand this to BP2, BP3, and BP4, so naming the tests by the BP will help us there

examples/python/bpsraw.py
jeremylt: https://github.com/CEED/libCEED/pull/1995

tylercollins5737: Awesome!

examples/python/bps_test.py
tylercollins5737: Gotcha, I'll just remove them and match the current example tests

examples/python/bps_test.py
tylercollins5737: Sounds good! I think that naming convention makes perfect sense.

tylercollins5737:
1. assert: removed them, now matching ex_test.py.
2. names: renamed to `test_bp1_degree_{1,2,3}` — BP-keyed for the
   BPs expansion.
3. take_array: thanks for #1995! bpsraw.py now detaches with
   `take_array`, matching the C code.

jeremylt:
Ok, we need a cross check from someone who has petsc-py set up locally, but this looks awesome and ready to go as far as I can tell

maybe @zatkins-dev has petsc-py?

zatkins-dev:
I should be able to test locally today!

zatkins-dev:
Looks like the source branch needs to be rebased/merged from main -- it still doesn't have the take_array binding.

zatkins-dev:
> Ok, we need a cross check from someone who has petsc-py set up locally, but this looks awesome and ready to go as far as I can tell

But we're passing once I do that rebase locally, so I think this is good to go.

zatkins-dev:
@jeremylt we should be able to set up a test for this through CI too, I can draft an MR for it.

jeremylt:
Yeah, would be good to add, I just didn't have time for it

tylercollins5737:
> Looks like the source branch needs to be rebased/merged from main -- it still doesn't have the take_array binding.

Yeah my bad, I had just added those lines locally. Rebased onto main!

zatkins-dev:
Looks good to me locally! I can add the test to CI in a follow-up.

zatkins-dev:
Alternatively, @tylercollins5737 can add this line to `.github/workflows/python-test-with-style.yml`

```diff
diff --git a/.github/workflows/python-test-with-style.yml b/.github/workflows/python-test-with-style.yml
index 5cf592c7..38db3e27 100644
--- a/.github/workflows/python-test-with-style.yml
+++ b/.github/workflows/python-test-with-style.yml
@@ -27,6 +27,7 @@ jobs:
       run: |
         pip install -r requirements.txt
         pip install -r requirements-test.txt
+        pip install -r requirements-test-petsc.txt
     - name: Python test
       env:
         CC: ${{ matrix.compiler }}
```

and add a new file `requirements-test-petsc.txt`:
```
mpi4py
petsc
petsc4py
```

tylercollins5737:
> Alternatively, @tylercollins5737 can add this line to `.github/workflows/python-test-with-style.yml`
> 
> ```diff
> diff --git a/.github/workflows/python-test-with-style.yml b/.github/workflows/python-test-with-style.yml
> index 5cf592c7..38db3e27 100644
> --- a/.github/workflows/python-test-with-style.yml
> +++ b/.github/workflows/python-test-with-style.yml
> @@ -27,6 +27,7 @@ jobs:
>        run: |
>          pip install -r requirements.txt
>          pip install -r requirements-test.txt
> +        pip install -r requirements-test-petsc.txt
>      - name: Python test
>        env:
>          CC: ${{ matrix.compiler }}
> ```
> 
> and add a new file `requirements-test-petsc.txt`:
> 
> ```
> mpi4py
> petsc
> petsc4py
> ```

Cool, just pushed those changes.

zatkins-dev:
let me try to diagnose, this is a PETSc issue

zatkins-dev:
alright, you can drop that last commit and then rebase from main and it should work

tylercollins5737:
Dropped the last commit and rebased to main.

zatkins-dev (approved):
Awesome, I think this looks great and we're passing tests. Assuming @jeremylt is on board, I'm good with merging!

jeremylt:
bedtime here, I'll do one last pass in the morning then merge!

tylercollins5737:
Excellent news! Thank you both for the help and letting me work through this. Truly amazing what's been built in libCEED. Cheers!

jeremylt (approved):
ToDo: Create issue so to extend this to BP 2-4

jeremylt:
thanks for all the work here - this is a great add for users to see how to use libCEED in a Python based HPC project

PR #1807, opened 2025-04-21
op - minor at points diagonal improvements

jeremylt:
Trimming out some extra operations


jeremylt:
Tested the changes for GPU against Ratel too, but I think the kernel launch latency is hiding the real speadups (and a gen version of this will be better anyways)

jeremylt:
Ok, need to find a leak

PR #1810, opened 2025-04-23
gen - unload module when destroying op

jeremylt:
Oops.... we never caught this before because we didn't typically have many OP to destroy in the middle of a simulation


zatkins-dev (approved):
Before:
image

After:
image

PR #1812, opened 2025-04-25
Reset work vectors in operator setup to save memory

zatkins-dev:
This should only matter for situations where many operators are created throughout a single execution. Prior to this change, minor increases in max vector length would lead to a growing cache of unused work vectors. 

Also, adds a utility to log current work vector usage when new vectors are allocated when in debug mode.

### DIFF
--- a/backends/cuda-ref/ceed-cuda-ref-operator.c
+++ b/backends/cuda-ref/ceed-cuda-ref-operator.c
@@ -353,8 +353,16 @@ static int CeedOperatorSetup_Cuda(CeedOperator op) {
       CeedCallBackend(CeedElemRestrictionDestroy(&rstr_i));
     }
   }
-
   CeedCallBackend(CeedClearWorkVectors(CeedOperatorReturnCeed(op), impl->max_active_e_vec_len));
+  {
+    // Create two work vectors for diagonal assembly
+    CeedVector temp_1, temp_2;
+
+    CeedCallBackend(CeedGetWorkVector(CeedOperatorReturnCeed(op), impl->max_active_e_vec_len, &temp_1));
+    CeedCallBackend(CeedGetWorkVector(CeedOperatorReturnCeed(op), impl->max_active_e_vec_len, &temp_2));
+    CeedCallBackend(CeedRestoreWorkVector(CeedOperatorReturnCeed(op), &temp_1));
+    CeedCallBackend(CeedRestoreWorkVector(CeedOperatorReturnCeed(op), &temp_2));
+  }
   CeedCallBackend(CeedOperatorSetSetupDone(op));
   CeedCallBackend(CeedQFunctionDestroy(&qf));
   return CEED_ERROR_SUCCESS;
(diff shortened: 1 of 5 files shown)


jeremylt:
Can we add a `t131` testing these new functions? I was thinking we'd add 3 vectors, view the cache, clear so 1-2 are deleted, then view the cache again? This would be a test with target output in `tests/output`

zatkins-dev:
I ended up tweaking the interface to output the memory usage, rather than just print it. That way, the test is checking the actual memory values and not just output

backends/cuda-ref/ceed-cuda-ref-operator.c
```diff
@@ -353,6 +353,8 @@ static int CeedOperatorSetup_Cuda(CeedOperator op) {
       CeedCallBackend(CeedElemRestrictionDestroy(&rstr_i));
     }
   }
+
+  CeedCallBackend(CeedClearWorkVectors(CeedOperatorReturnCeed(op), impl->max_active_e_vec_len));
```
jeremylt, line 357: For this to work as expected, we also want to get and restore at least one vec of that length, I think?

jeremylt: Probably after clearing the cache would be the most memory friendly way to do that

interface/ceed.c
```diff
@@ -817,6 +817,62 @@ int CeedReference(Ceed ceed) {
...
+int CeedClearWorkVectors(Ceed ceed, CeedSize min_len) {
+  if (!ceed->work_vectors) return CEED_ERROR_SUCCESS;
+  for (CeedInt i = 0; i < ceed->work_vectors->num_vecs; i++) {
+    if (ceed->work_vectors->is_in_use[i]) continue;
+    CeedSize vec_len;
+    CeedCall(CeedVectorGetLength(ceed->work_vectors->vecs[i], &vec_len));
+    if (vec_len < min_len) {
+      ceed->ref_count += 2;  // Note: increase ref_count to prevent Ceed destructor from triggering again
+      CeedCall(CeedVectorDestroy(&ceed->work_vectors->vecs[i]));
+      ceed->ref_count -= 1;  // Note: restore ref_count
```
jeremylt, line 863:
```suggestion
      ceed->ref_count += 2;  // Note: increase ref_count to prevent Ceed destructor from triggering
      CeedCall(CeedVectorDestroy(&ceed->work_vectors->vecs[i]));
      ceed->ref_count -= 1;  // Note: restore ref_count
```
We probably don't need this, but its good to keep just in case. Minor wording change since this doesn't trigger via a call to `CeedDestroy()`

backends/cuda-ref/ceed-cuda-ref-operator.c
zatkins-dev: I mean, the next time that we ask for a vector, we'll be asking for one of that size. I don't think it matters too much to get the vector here, that would alloc a vector that would be unused if, e.g., the operator was destroyed without calling `Apply`

jeremylt: Not necessarily. If we use a smaller Op first, and then a bigger Op, then we can have more variables in the cache than we want.

jeremylt: Also, the memory is not allocated in a vector until its explicitly used in some way

zatkins-dev: Ah, I suppose then.

jeremylt: Probably should cache 2 vecs so diagonal assembly doesn't accidentally accrue too many small vectors. If the second is never used, no big deal as its only the CeedVec struct that'd be allocated

jeremylt (approved):
Awesome, thanks for this update. I'll add it to Nother and the Ratel dockerfiles on Monday

PR #1814, opened 2025-04-28
pc - multigrid level setup for AtPoints

jeremylt:
Turning on multigrid level setup. Small change in the interface, so most of the work was setting up the test.

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -885,7 +885,23 @@ static int CeedSingleOperatorMultigridLevel(CeedOperator op_fine, CeedVector p_m
   CeedCheck(!is_composite, ceed, CEED_ERROR_UNSUPPORTED, "Automatic multigrid setup for composite operators not supported");
 
   // Coarse Grid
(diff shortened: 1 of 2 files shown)


interface/ceed-preconditioning.c
```diff
@@ -885,7 +885,23 @@ static int CeedSingleOperatorMultigridLevel(CeedOperator op_fine, CeedVector p_m
...
-  CeedCall(CeedOperatorCreate(ceed, op_fine->qf, op_fine->dqf, op_fine->dqfT, op_coarse));
+  {
+    bool is_at_points;
+
+    CeedCall(CeedOperatorIsAtPoints(op_fine, &is_at_points));
+    if (is_at_points) {
+      CeedVector          point_coords;
+      CeedElemRestriction rstr_points;
+
+      CeedCall(CeedOperatorCreateAtPoints(ceed, op_fine->qf, op_fine->dqf, op_fine->dqfT, op_coarse));
```
jeremylt, line 896: Note: Here only the fine and coarse grid operators are AtPoints. There's no benefit to making the restriction and prolongation AtPoints that I can see since we're simply transferring between FEM grids.

jedbrown: I agree.

PR #1818, opened 2025-05-05
backends - use defines over weak

jeremylt:
@jedbrown this is still WIP but I think this is roughly what we need to fix Mac not respecting weak symbols?

ToDo: Always build `backends/ceed-backends.c`


jeremylt:
Note to me - fix priorities

Makefile
```diff
@@ -662,17 +671,17 @@ endif
 # Building core library components
 # ------------------------------------------------------------
 
-# File names *-weak.c contain weak symbol definitions, which must be listed last
-# when creating shared or static libraries.
-weak_last = $(filter-out %-weak.o,$(1)) $(filter %-weak.o,$(1))
+libceed-backend-list.o = $(libceed-backend-list.c:%.c=$(OBJDIR)/%.o)
+.PHONY: libceed-backend-list
+libceed-backend-list: $(OBJDIR)/backends/ceed-backend-list.o
```
jeremylt, line 676: Something is wrong here. I am trying to force this file to always build.

jeremylt: This file failing to rebuild is why CI is failing

jeremylt: Fixed

backends/ceed-backend-list.c
```diff
@@ -0,0 +1,331 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and other CEED contributors.
```
jeremylt, line 1: Note - part of the thinking here was that this would be one file so the list always matches the current compilation env. I was also thinking that since we add backends so infrequently this would help keep everything in one spot for easy reference.

PR #1819, opened 2025-05-06
Zach/mi300a fixes

zatkins-dev:
Reworks the stream implementation for `/gpu/hip/gen` to avoid creating and destroying streams on every operator apply.

Updates hipblas calls to only sync stream, this matters on MI300A since hipblas seems to use an async stream. Avoids a full device sync.

Also makes working vectors come from the `Vector` object delegate to avoid bad ref behavior.

### DIFF
--- a/backends/hip-gen/ceed-hip-gen-operator.c
+++ b/backends/hip-gen/ceed-hip-gen-operator.c
@@ -22,9 +22,20 @@
 static int CeedOperatorDestroy_Hip_gen(CeedOperator op) {
   Ceed                  ceed;
   CeedOperator_Hip_gen *impl;
+  bool                  is_composite;
 
   CeedCallBackend(CeedOperatorGetCeed(op, &ceed));
   CeedCallBackend(CeedOperatorGetData(op, &impl));
+  CeedCallBackend(CeedOperatorIsComposite(op, &is_composite));
+  if (is_composite) {
+    CeedInt num_suboperators;
+
+    CeedCall(CeedCompositeOperatorGetNumSub(op, &num_suboperators));
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      if (impl->streams[i]) CeedCallHip(ceed, hipStreamDestroy(impl->streams[i]));
+      impl->streams[i] = NULL;
+    }
+  }
   if (impl->module) CeedCallHip(ceed, hipModuleUnload(impl->module));
   if (impl->points.num_per_elem) CeedCallHip(ceed, hipFree((void **)impl->points.num_per_elem));
   CeedCallBackend(CeedFree(&impl));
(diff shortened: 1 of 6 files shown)


backends/hip-gen/ceed-hip-gen-operator.c
```diff
@@ -22,9 +22,20 @@
...
 
   CeedCallBackend(CeedOperatorGetCeed(op, &ceed));
   CeedCallBackend(CeedOperatorGetData(op, &impl));
+  CeedCallBackend(CeedOperatorIsComposite(op, &is_composite));
+  if (is_composite) {
+    CeedInt num_suboperators;
+
+    CeedCall(CeedCompositeOperatorGetNumSub(op, &num_suboperators));
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      if (impl->streams[i]) CeedCallHip(ceed, hipStreamDestroy(impl->streams[i]));
```
jeremylt, line 35: curiosity - is there a noticeable cost to creating and destroying the streams as we were?

zatkins-dev: yeah, it's a heap alloc and destroy. But more importantly, it wasn't running in parallel since the `hipStreamDestroy` call implies a `hipStreamSynchronize`.

interface/ceed.c
```diff
@@ -858,7 +858,7 @@ int CeedClearWorkVectors(Ceed ceed, CeedSize min_len) {
     CeedSize vec_len;
     CeedCall(CeedVectorGetLength(ceed->work_vectors->vecs[i], &vec_len));
     if (vec_len < min_len) {
-      ceed->ref_count += 2;  // Note: increase ref_count to prevent Ceed destructor from triggering
+      ceed->ref_count += 2;  // Note: increase ref_count to prevent Ceed destructor from triggering again
```
jeremylt, line 861:
```suggestion
      ceed->ref_count += 2;  // Note: increase ref_count to prevent Ceed destructor from triggering
```

backends/hip-gen/ceed-hip-gen-operator.c
zatkins-dev: So using them wasn't actually giving us anything over just putting it in the default stream

interface/ceed.c
jeremylt: CeedClearWorkVectors would also need to use the object delegate

zatkins-dev: yep, just added

backends/hip-gen/ceed-hip-gen-operator.c
zatkins-dev: technically, I think it's UB whether it blocks or not, which is even worse I think

interface/ceed.c
```diff
@@ -852,6 +861,15 @@ int CeedGetWorkVectorMemoryUsage(Ceed ceed, CeedScalar *usage_mb) {
   @ref Backend
 **/
 int CeedClearWorkVectors(Ceed ceed, CeedSize min_len) {
+  if (!ceed->VectorCreate) {
+    Ceed delegate;
+
+    CeedCall(CeedGetObjectDelegate(ceed, &delegate, "Vector"));
+    CeedCheck(delegate, ceed, CEED_ERROR_UNSUPPORTED, "Backend does not implement VectorCreate");
```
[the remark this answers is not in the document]
zatkins-dev: I agree, the linter doesn't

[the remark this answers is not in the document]
zatkins-dev: because it thinks that delegate can be `NULL`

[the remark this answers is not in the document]
jeremylt: Bad linter

PR #1826, opened 2025-05-19
rust - update rust examples to match c/python

jeremylt:
Updates the naming convention to match C/Python better and added ex3


jedbrown (approved):
LGTM. There's perhaps excess duplication here. We could consolidate to one crate with multiple binaries (so they can all share the lib part) or split out a utility crate that all the binaries would use.

jeremylt:
Yeah, at some point it would be nice to tidy that up more. I have mixed feelings sometimes about how much code to collapse together as these are sorta intended as tutorial examples?

PR #1831, opened 2025-05-30
cpu - remove vectorize warnings for Clang -O3

jeremylt:
I don't think this has any performance issues, but in this branch I've tried to silence Clang's vectorization warnings at -O3 by only marking the innermost restriction loops with CeedPragmaSIMD. Given the difficulty of actually vectorizing nested loops unless the inner loop range is fixed, I'd guess this has no effect on our actual performance?

### DIFF
--- a/backends/ref/ceed-ref-restriction.c
+++ b/backends/ref/ceed-ref-restriction.c
@@ -17,8 +17,8 @@
 // Core ElemRestriction Apply Code
 //------------------------------------------------------------------------------
 static inline int CeedElemRestrictionApplyStridedNoTranspose_Ref_Core(CeedElemRestriction rstr, const CeedInt num_comp, const CeedInt block_size,
-                                                                      CeedInt start, CeedInt stop, CeedInt num_elem, CeedInt elem_size,
-                                                                      CeedSize v_offset, const CeedScalar *__restrict__ uu,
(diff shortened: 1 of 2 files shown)


backends/ref/ceed-ref-restriction.c
```diff
@@ -28,8 +28,8 @@ static inline int CeedElemRestrictionApplyStridedNoTranspose_Ref_Core(CeedElemRe
     // CPU backend strides are {1, elem_size, elem_size*num_comp}
     // This if branch is left separate to allow better inlining
     for (CeedSize e = start * block_size; e < stop * block_size; e += block_size) {
-      CeedPragmaSIMD for (CeedSize k = 0; k < num_comp; k++) {
-        CeedPragmaSIMD for (CeedSize n = 0; n < elem_size; n++) {
```
jedbrown, line 32: Rationale for reordering these lines should probably be separate from removing the pragmas.

jedbrown (commented):
I don't anticipate any impact on performance from removing the pragma in these positions.

backends/ref/ceed-ref-restriction.c
jeremylt: Ahh, oops, was trying some things and forgot to separate these commits.

My thinking here is that num_comp is one of the things we specialize on, so this pushes all of the loops that can be statically unrolled together towards the inside.

PR #1834, opened 2025-06-12
doc - improve internal API documentation

### LINKED ISSUE #1767: Internal API and Design Documentation
We should expand our documentation of the internal API and design to better support onboarding and future development. This is especially important for our GPU backends, especially the code generation backends `/gpu/cuda/gen` and `/gpu/hip/gen`

jeremylt:
fixes #1767 

Ok, this is an effort to better explain the internal API and overall design. Is this helpful? Did I miss things you'd like to see?

### DIFF
--- a/doc/sphinx/source/index.md
+++ b/doc/sphinx/source/index.md
@@ -8,9 +8,9 @@ intro
 gettingstarted
 libCEEDapi
 examples/index
+api/index
 ffi
 gpu
-api/index
 precision
 libCEEDdev
 Contributing <CONTRIBUTING>


doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
+Then in the creation of the backend specific implementation of the object, typically found in `/backends/[impl]/ceed-[impl]-[object].c`, the developer creates the backend implementation of the specific method and calls {c:func}`CeedSetBackendFunction` to set this implementation of the method for the backend.
+Any supplemental functions intended to be used in the interface or by the backends may be added to the backend API in the header `/include/ceed/backend.h`.
+The basic implementation of these functions should also be placed in `/interface/*.c` files.
 
-In addition to those automatically enforced style rules, libCEED tends to follow the following code style conventions:
+LibCEED generally follows a "CPU first" implementation strategy when adding new functionality to the user facing API.
+If there are no performance specific considerations, it is generally recommended to include a basic CPU default implementation in `/interface/*.c`.
+Any new functions must be well documented and well tested.
```
jrwrigh:
```suggestion
Any new functions must be well documented and tested.
```

I don't think we want to emphasize testing for testings-sake; the tests should be minimal to cover the usage of the function.

This is probably what you intended, but I interpret "well ____" as "more is better". That's probably true for the documentation, but testing should be a bit more precise in it's implementation, imo.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
+Once the user facing API and the default implementation are in place and verified correct via tests, then the developer can focus on hardware specific implementations, such as for CUDA or HIP backends.
```
jrwrigh:
```suggestion
Once the user facing API and the default implementation are in place and verified correct via tests, then the developer can focus on hardware specific implementations (AVX, CUDA, HIP, etc.) if deemed necessary for performance.
```

Just to re-emphasize a previous point that not everything need be put into a computational backend.

Also maybe should point out that a CPU-only implementation is perfectly valid to make a PR on and get into the library; the implementations in other backends aren't necessary to contribute a feature.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
-- Type names: `PascalCase` or language specific style
-- Constant names: `CAPS_SNAKE_CASE` or language specific style
+## Backend Inheritance
 
-Also, documentation files should have one sentence per line to help make git diffs clearer and less disruptive.
+There are three mechanisms by which a Ceed backend can inherit implementation from another Ceed backend.
+These options are set in the backend initialization routine.
 
-## Clang-tidy
+1. Delegation - Developers may use {c:func}`CeedSetDelegate()` to set a backend that will provide the implementation of any unimplemented Ceed objects.
```
jrwrigh: I think these are best stated in the opposite order; what is it, and how to do it. So
```suggestion
1. Delegation - If a backend does not implement some functionality, it may use (delegate) the implementation of that functionality to a  "delegate" backend. Developers may use {c:func}`CeedSetDelegate()` to set the delegate backend.
```

jrwrigh: Note that I changed "unimplemented Ceed objects" to "unimplemented functionality". If I understand correct, there is only ever one definition of a Ceed object, but the functions using that object may change.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
+If there are no performance specific considerations, it is generally recommended to include a basic CPU default implementation in `/interface/*.c`.
+Any new functions must be well documented and well tested.
+Once the user facing API and the default implementation are in place and verified correct via tests, then the developer can focus on hardware specific implementations, such as for CUDA or HIP backends.
 
-- Variable names: `snake_case`
-- Strut members: `snake_case`
-- Function and method names: `PascalCase` or language specific style
-- Type names: `PascalCase` or language specific style
-- Constant names: `CAPS_SNAKE_CASE` or language specific style
+## Backend Inheritance
```
jrwrigh, line 22: I find the wording of this section confusing. I think the explanation I'd prefer is something like.

> A Ceed backend is not required to implement every function for every Ceed object. 
> Instead, it may inherit function implementations from other backends, hence "Backend Inheritance". 
> This inheritance system operates on 3 scopes: per-backend, per-object, and `CeedOperator` methods.
> These are known as delegation, object delegation, and operator fallback:
>
> 1. Delegation - ....
> 2. Object Delegation - ...
> ....

jrwrigh: An example with a diagram would be helpful too. Could probably do the diagram in mermaid as well.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
+The `/cpu/self/opt/*` backends update the {ref}`CeedOperator` to apply the action of the operator in 1 or 8 element batches, depending upon if the blocking strategy is used.
+This reduced the memory required to utilize this backend significantly.
 
-Every symbol that is used in the source file `foo.c` should be defined in `foo.c`, `foo.h`, or in a header file `#include`d in one of these two locations.
-Please check your code by running the tool [`include-what-you-use`](https://include-what-you-use.org/) to see recommendations for changes to your source.
-Most issues reported by `include-what-you-use` should be fixed; however this rule is flexible to account for differences in header file organization in external libraries.
-If you have `include-what-you-use` installed in a sibling directory to libCEED or set the environment variable `IWYU_CC`, then you can use the makefile target `make iwyu`.
+The `/cpu/self/avx/*` and `/cpu/self/xsmm/*` backends delegate to the corresponding `/cpu/self/opt/*` backends.
+These backends update the `CeedTensorContract` objects.
```
jrwrigh:
```suggestion
These backends update the `CeedTensorContract` objects using AVX intrinsics and libXSMM functions, respectively.
```

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
 
-Header files should be listed in alphabetical order, with installed headers preceding local headers and `ceed` headers being listed first.
-The `ceed-f64.h` and `ceed-f32.h` headers should only be included in `ceed.h`.
+The `/cpu/self/memcheck/*` backends are delegate to the `/cpu/self/ref/*` backends.
```
jrwrigh:
```suggestion
The `/cpu/self/memcheck/*` backends delegate to the `/cpu/self/ref/*` backends.
```

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
+These backends replace many of the implementations with methods that include more verification checks and a memory management model that more closely matches the memory management for GPU backends.
```
jrwrigh, line 63: Probably also need to mention that it requires valgrind headers?

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
-For example, the comment
+The `/gpu/*/ref` backends provide basic functionality.
+In these backends, the operator is applied in multiple separate kernel launches, following the libCEED operator decomposition, where first {ref}`CeedElemRestriction` kernels map from the L-vectors to E-vectors, then {ref}`CeedBasis` kernels map from the E-vectors to Q-vectors, then the {ref}`CeedQFunction` kernel provides the action of the user quadrature point function, and the transpose {ref}`CeedBasis` and {ref}`CeedElemRestriction` kernels are applied.
+These kernels apply to all points across all elements in order to maximize the amount of work each kernel launch has.
 
-```c
-// u has shape [dim, num_comp, Q, num_elem]
-```
+The `/gpu/*/shared` backends delegate to the corresponding `/gpu/*/ref` backends.
+These backends use shared memory to reduce the amount of memory required for the {ref}`CeedBasis` kernels.
```
jrwrigh: Maybe a bit more on this (if there is)? Possibly just for my own curiosity. If the `ref` backends do each step sequentially, does this break it up at all?

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
 
-```c
-for (d=0; d<dim; d++)
-  for (c=0; c<num_comp; c++)
-    for (q=0; q<Q; q++)
-      for (e=0; e<num_elem; e++)
-        u[((d*num_comp + c)*Q + q)*num_elem + e] = ...
-```
+The `/gpu/*/magma` backends delegate to the corresponding `/gpu/*/ref` backends.
+These backends provide better performance for {ref}`CeedBasis` kernels but do not have the improvements from the `/gpu/*/gen` backends for {ref}`CeedOperator`.
```
jrwrigh, line 87: If magma only helps with the basis, could/should it delegate to `/gpu/*/shared` instead? If not, why not?

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
 
-This ordering is sometimes referred to as row-major or C-style.
-Note that flattening such as
+The `/*/*/occa` backends are not part of any family and generally are not well maintained.
```
jrwrigh: > generally are not well maintained.

Maybe we could phrase it nicer? IDK, the tone seems like it's implicitly blaming somebody. maybe something like.

> ... and do not recieve much active development or maintenance.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
 
-```c
-// u has shape [dim, num_comp, Q*num_elem]
-```
+## Internal Layouts
 
-and
+Ceed backends are free to use any **E-vector** and **Q-vector** data layout, to include never fully forming these vectors, so long as the backend passes the `t5**` series tests and all examples.
```
jrwrigh:
```suggestion
Ceed backends are free to use any **E-vector** and **Q-vector** data layout (including never fully forming these vectors) so long as the backend passes the `t5**` series tests and all examples.
```

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
+There are several common layouts for **L-vectors**, **E-vectors**, and **Q-vectors**, detailed below:
 
-```c
-// u has shape [dim*num_comp, Q, num_elem]
-```
+- **L-vector** layouts
 
-are purely implicit -- one just indexes the same array using the appropriate convention.
+  - **L-vectors** described by a {ref}`CeedElemRestriction` have a layout described by the `offsets` array and `comp_stride` parameter.
```
jrwrigh:
```suggestion
  - **L-vectors** described by a "standard" {ref}`CeedElemRestriction` have a layout described by the `offsets` array and `comp_stride` parameter.
```

to contrast it with a "strided" restriction

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
+    Data for node `i`, component `j`, element `k` can be found in the **L-vector** at index `i*strides[0] + j*strides[1] + k*strides[2]`.
 
-## `restrict` Semantics
+- **E-vector** layouts
 
-QFunction arguments can be assumed to have `restrict` semantics.
-That is, each input and output array must reside in distinct memory without overlap.
+  - If possible, backends should use {c:func}`CeedElemRestrictionSetELayout()` to use the `t2**` tests.
+    If the backend uses a strided **E-vector** layout, then the data for node `i`, component `j`, element `k` in the **E-vector** is given by `i*layout[0] + j*layout[1] + k*layout[2]`.
+  - Backends may choose to use a non-strided **E-vector** layout; however, the `t2**` tests will not function correctly in this case and the tests will need to be whitelisted for the backend to pass the test suite.
```
jrwrigh: > need to be whitelisted for the backend to pass the test suite.

That should probably be documented somewhere at some point.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
+
+- **Q-vector** layouts
 
-## CeedVector Array Access Semantics
+  - When the size of a {ref}`CeedQFunction` field is greater than `1`, data for quadrature point `i` component `j` can be found in the **Q-vector** at index `i + Q*j`.
```
jrwrigh: Should (re)define what `Q` is in this context.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,109 @@
...
-## `restrict` Semantics
+- **E-vector** layouts
 
-QFunction arguments can be assumed to have `restrict` semantics.
-That is, each input and output array must reside in distinct memory without overlap.
+  - If possible, backends should use {c:func}`CeedElemRestrictionSetELayout()` to use the `t2**` tests.
+    If the backend uses a strided **E-vector** layout, then the data for node `i`, component `j`, element `k` in the **E-vector** is given by `i*layout[0] + j*layout[1] + k*layout[2]`.
+  - Backends may choose to use a non-strided **E-vector** layout; however, the `t2**` tests will not function correctly in this case and the tests will need to be whitelisted for the backend to pass the test suite.
+
+- **Q-vector** layouts
```
jrwrigh, line 109: The documentation for the Q-vector layout and the L-vector layout should also be in "normal" documentation somewhere, since those are directly user facing. I know the L-vector layouts are described in the "normal" and "strided" restriction functions, but I don't think Q-vector is described anywhere.

doc/sphinx/source/libCEEDdev.md
jeremylt: not quite, lemmie push a better version

doc/sphinx/source/libCEEDdev.md
jeremylt:
```suggestion
Once the user facing API and the default implementation are in place and verified correct via tests, then the developer can focus on hardware specific implementations (AVX, CUDA, HIP, etc.) as necessary.
```

doc/sphinx/source/libCEEDdev.md
jeremylt:
```suggestion
  - **L-vectors** described by a standard {ref}`CeedElemRestriction` have a layout described by the `offsets` array and `comp_stride` parameter.
```

doc/sphinx/source/libCEEDdev.md
jeremylt: There isn't much more to say. All this backend does is replace the `CeedBasis` kernels with faster ones

doc/sphinx/source/libCEEDdev.md
jeremylt: It should not - shared replaces only the CeedBasis, and that's also what MAGMA replaces, so it would be an extra layer of indirection that isn't needed

doc/sphinx/source/libCEEDdev.md
jeremylt: Users should not reach into Q vecs or L vecs

doc/sphinx/source/libCEEDdev.md
jeremylt: It's nicer that "unmaintained liability" but I'll soften it a bit :)

doc/sphinx/source/libCEEDdev.md
jeremylt: hopefully the wording is better now?

doc/sphinx/source/libCEEDdev.md
jrwrigh: Understood!

doc/sphinx/source/libCEEDdev.md
jeremylt: Wait, users do touch L-vecs. But the description of L-vec layout is with the `CeedElemRestriction*Create` documentation.

Users *don't* touch E-vecs, but we've still some documentation with `CeedElemRestrictionGetELayout`

We should probably say how to index into the input/output arrays for the user QF but I haven't found a satisfying place for that.

doc/sphinx/source/libCEEDdev.md
jeremylt: This is only dev facing, so I'm not sure where else to document it? Also, edited to remove the phrase "whitelist"

doc/sphinx/source/libCEEDdev.md
jrwrigh: I think the L vec documentation in the `CeedElemRestriction*Create` is probably adequate, but indexing into the Q vecs within a QFunction isn't documented to my knowledge. Or at least I always had to reference an already-implemented QFunction to remember how it was ordered.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,113 @@
...
-For example, the comment
+The `/gpu/*/ref` backends provide basic functionality.
+In these backends, the operator is applied in multiple separate kernel launches, following the libCEED operator decomposition, where first {ref}`CeedElemRestriction` kernels map from the L-vectors to E-vectors, then {ref}`CeedBasis` kernels map from the E-vectors to Q-vectors, then the {ref}`CeedQFunction` kernel provides the action of the user quadrature point function, and the transpose {ref}`CeedBasis` and {ref}`CeedElemRestriction` kernels are applied to go back to the E-vectors and finally the L-vectors.
+These kernels apply to all points across all elements in order to maximize the amount of work each kernel launch has.
 
-```c
-// u has shape [dim, num_comp, Q, num_elem]
-```
+The `/gpu/*/shared` backends delegate to the corresponding `/gpu/*/ref` backends.
+These backends use shared memory to reduce the amount of memory required for the {ref}`CeedBasis` kernels and improve performance.
```
nbeams: Do the shared basis kernels actually reduce memory use compared to ref? I thought the sole purpose of `shared` was improved performance.

doc/sphinx/source/libCEEDdev.md
```diff
@@ -16,26 +16,28 @@ The basic implementation of these functions should also be placed in `/interface
...
+   This delegate {ref}`Ceed` will provide the implementation of any libCeed objects that parent backend does not implement.
+2. Object delegation  - Developers may use {c:func}`CeedSetObjectDelegate` to set a delegate {ref}`Ceed` object for a specific libCEED object.
+   This delegate {ref}`Ceed` will only provide the implementation of that specific libCeed object for the parent backend.
    Object delegation has higher precedence than delegation.
-3. Operator fallback - Developers may use {c:func}`CeedSetOperatorFallbackResource()` to set a {ref}`Ceed` resource that will provide the implementation of unimplemented {ref}`CeedOperator` methods.
-   A fallback {ref}`Ceed` with this resource will only be instantiated if a method is called that is not implemented by the parent {ref}`Ceed`.
-   In order to use the fallback mechanism, the parent {ref}`Ceed` and fallback resource must use compatible **E-vector** and **Q-vector** layouts.
+3. Operator fallback - Developers may use {c:func}`CeedSetOperatorFallbackResource` to set a {ref}`Ceed` resource that will provide the implementation of any unimplemented {ref}`CeedOperator` methods.
```
jrwrigh:
```suggestion
3. Operator fallback - Developers may use {c:func}`CeedSetOperatorFallbackResource` to set a {ref}`Ceed` object that will provide the implementation of any unimplemented {ref}`CeedOperator` methods.
```

Assuming this is correct? All other references are to "`Ceed` objects" rather than a resource.

jrwrigh: Oh, this just specifies the backend that it will create a new `Ceed` object with in case it needs it, while the inheritance method 1 and 2 give it a `Ceed` object explicitly?

jrwrigh: Perhaps there's some information missing about what exactly defines a "backend"? (Or maybe I just need to read it more carefully...)

jeremylt: This one I wanted to say "resource". For the delegates we set a Ceed object because we will need it. For the fallback we set a resource because we may not need it.

See cuda/gen as an example
```C
static int CeedInit_Cuda_gen(const char *resource, Ceed ceed) {
  char      *resource_root;
  const char fallback_resource[] = "/gpu/cuda/ref";
  Ceed       ceed_shared;
  Ceed_Cuda *data;

  CeedCallBackend(CeedGetResourceRoot(ceed, resource, ":", &resource_root));
  CeedCheck(!strcmp(resource_root, "/gpu/cuda") || !strcmp(resource_root, "/gpu/cuda/gen"), ceed, CEED_ERROR_BACKEND,
            "Cuda backend cannot use resource: %s", resource);
  CeedCallBackend(CeedFree(&resource_root));

  CeedCallBackend(CeedCalloc(1, &data));
  CeedCallBackend(CeedSetData(ceed, data));
  CeedCallBackend(CeedInit_Cuda(ceed, resource));
// Setting a Ceed object here to delegate to 
  CeedCallBackend(CeedInit("/gpu/cuda/shared", &ceed_shared));
  CeedCallBackend(CeedSetDelegate(ceed, ceed_shared));
  CeedCallBackend(CeedDestroy(&ceed_shared));
// Setting a Ceed resource in case we need to fallback
  CeedCallBackend(CeedSetOperatorFallbackResource(ceed, fallback_resource));

  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "QFunctionCreate", CeedQFunctionCreate_Cuda_gen));
  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "OperatorCreate", CeedOperatorCreate_Cuda_gen));
  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "CompositeOperatorCreate", CeedOperatorCreate_Cuda_gen));
  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "OperatorCreateAtPoints", CeedOperatorCreate_Cuda_gen));
  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "Destroy", CeedDestroy_Cuda));
  return CEED_ERROR_SUCCESS;
}
```

jrwrigh: Maybe going too hard in the exact/math definitions, but I think it might be useful to define what exactly the difference between a `Ceed` object, a `Ceed` resource, and a backend is.

Then subsequently, how those concepts interact with libCeed objects and `CeedOperator` methods.

Then I think talking about inheritance will make a bit more sense. Or at least it'll be easier to discuss with terminology and the "normal" process defined.

doc/sphinx/source/libCEEDdev.md
jrwrigh: Yep, better. Further followup comments in https://github.com/CEED/libCEED/pull/1834/files/86fd8a671f562887629dc2d04d370f9231c28b46..24516c14203fcd8d9bfdcfd37524a063a2f9a31c#r2143691395

doc/sphinx/source/libCEEDdev.md
jrwrigh: Yeah, it might be nice to have documentation for the test suite (kinda like what [PETSc has](https://petsc.org/release/developers/testing/)). This information would fit in something there.

I think for right now just saying what needs to happen is fine. Developers can ask how to do it (or snoop in the code and see for themselves).

doc/sphinx/source/libCEEDdev.md
jrwrigh: (side note from the code snippet; probably don't need to have ``fallback_resource` as a separate variable and just have it as a raw argument to `CeedSetOperatorFallbackResource`)?

doc/sphinx/source/libCEEDdev.md
```diff
@@ -1,98 +1,113 @@
...
-#include <ceed.h>
-#include <ceed/backend.h>
-#include <stdbool.h>
-#include <string.h>
-#include "ceed-avx.h"
-```
+### GPU Backends
 
-## Shape
+The CUDA, HIP, and SYCL backend families all follow the same basic design.
```
nbeams: I think the SYCL backends are different enough to warrant a note about that somewhere (since this is for developers and not users). For example, right now, I think `sycl-ref` doesn't use runtime compilation for basis kernels, unlike `cuda-ref` and `hip-ref`. Or there are some subtleties to keep in mind for the RTC in SYCL backends, since it uses OpenCL (for now, maybe there will be Intel support for SYCL RTC in the future). I think @kris-rowe had to be careful about making sure things like block dimensions are in the right order (SYCL is "backward" from CUDA, and then we have another layer added with the OpenCL RTC). Maybe he has some thoughts about whether we should add some SYCL-specific tips for developers?

doc/sphinx/source/libCEEDdev.md
jeremylt: I believe so, the working arrays are spread across theads. Its a mix though with the limited size of the shared memory

doc/sphinx/source/libCEEDdev.md
jeremylt: looks like the shapes are described in `CeedQFunctionUser`

doc/sphinx/source/libCEEDdev.md
nbeams: Maybe I don't understand -- by "the amount of memory required," I assumed we meant size of global GPU memory allocated. Does it actually mean the shared memory and/or private variables? ...Does `shared` actually use *less* shared memory than `ref`? This is about to blow my mind if true... :)

jeremylt: Ref does have these big buffers that shared doesn't
```
  __shared__ CeedScalar s_mem[BASIS_Q_1D * BASIS_P_1D + 2 * BASIS_BUF_LEN];
  CeedScalar           *s_interp_1d = s_mem;
  CeedScalar           *s_buffer_1  = s_mem + BASIS_Q_1D * BASIS_P_1D;
  CeedScalar           *s_buffer_2  = s_buffer_1 + BASIS_BUF_LEN;
```
but shared has these
```
  __shared__ CeedScalar s_B[BASIS_P_1D * BASIS_Q_1D];
  data.slice  = slice + data.t_id_z * BASIS_T_1D * (BASIS_DIM > 1 ? BASIS_T_1D : 1);
```

I think that's a net win for shared, but a little bit?

jeremylt: I don't really want to count it all out though, so I'll just delete this bit

doc/sphinx/source/libCEEDdev.md
jeremylt: updated

doc/sphinx/source/libCEEDdev.md
nbeams: But `shared`'s `slice` is dynamic shared memory that's specified in the call to `CeedRunKernelDimShared_Cuda`, and depends on `elems_per_block`, so that gets more complicated to compare...

doc/sphinx/source/libCEEDdev.md
jeremylt: I'm trying to avoid a definitions block because that broke up the flow and hurt clarity, I think?

doc/sphinx/source/libCEEDdev.md
jeremylt: I added some clarification. I'd definitely be for adding any specific notes about special quirks for any backend 'family'

doc/sphinx/source/libCEEDdev.md
```diff
@@ -26,18 +26,18 @@ There are three mechanisms by which a Ceed backend can inherit implementations f
 
 1. Delegation - Developers may use {c:func}`CeedSetDelegate` to set a general delegate {ref}`Ceed` object.
    This delegate {ref}`Ceed` will provide the implementation of any libCeed objects that parent backend does not implement.
+   For example, the `/cpu/self/xsmm/serial` backend implements the `CeedTensorContract` object itself but delegates all other functionality to the `/cpu/self/opt/serial` backend.
+
 2. Object delegation  - Developers may use {c:func}`CeedSetObjectDelegate` to set a delegate {ref}`Ceed` object for a specific libCEED object.
    This delegate {ref}`Ceed` will only provide the implementation of that specific libCeed object for the parent backend.
    Object delegation has higher precedence than delegation.
```
jrwrigh: I like the moving of the delegation example to the delegation bullet. Maybe we could add/move an example for the object delegation as well?

doc/sphinx/source/libCEEDdev.md
jrwrigh: Definitely better! The next logical question that I can't quite figure out though is why have the fallback vs (either) delegation. The choice between full delegation and object delegation seems pretty straight forward (object is just a more precisely scoped version of the full), but I'm not sure why delegation can't handle the problems that fallback aims to solve.

jrwrigh: > I'm trying to avoid a definitions block because that broke up the flow and hurt clarity, I think?

I generally agree with that for user-facing documentation, but for developer documentation, precision == clarity IMO. Honestly, even from a users perspective, I personally like precise definitions. This is software to solve math problems after all.

jrwrigh: You can also interleave definitions with description too. Most of my favorite math textbooks and monographs do that.

doc/sphinx/source/libCEEDdev.md
jeremylt: It turns out that we previously used object delegation but don't currently, so there is no example of it anymore. The option is still there if we need it again

doc/sphinx/source/libCEEDdev.md
jeremylt: The fallback is for *partial* delegation where most but not all of the functionality is already supported by the parent backed

jeremylt: Clarified in latest push

jrwrigh: Ah, ok. That makes sense. Advanced functionality in this case would be something like the AtPoints stuff, where initially it was only implemented for `ref` backends (and thus `gen` would fallback to `ref` for those calls)?

jeremylt: I think its just preconditioning support actually. I could just make a list

jeremylt: I called out a specific example and made it clearer that this is only for preconditioner support

jrwrigh: Cool!

PR #1838, opened 2025-06-17
Clarify __syncthreads usage

jeremylt:
This fixes the intermittent t354 failures in CI

Basically, we really just need `__syncthreads` when we're changing the contents shared memory buffers so I moved them to clarify this (and dropped a couple)


jedbrown (commented):
I know this is tedious to check for correctness. Would it be helpful to document on each function which arguments are shared memory and what the pre/post-conditions are regarding synchronization?

PR #1839, opened 2025-06-17
GPU Gen Full Assembly at Points

zatkins-dev:
Adds full single operator assembly at points for CUDA and HIP gen backends.

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator.c
+++ b/backends/cuda-gen/ceed-cuda-gen-operator.c
@@ -448,6 +448,7 @@ static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Cuda_gen(CeedOperator o
 
     CeedCallBackend(
         CeedTryRunKernelDimShared_Cuda(ceed, data->assemble_diagonal, NULL, grid, block[0], block[1], block[2], shared_mem, &is_run_good, opargs));
+    CeedCallCuda(ceed, cudaDeviceSynchronize());
 
     // Restore input arrays
     for (CeedInt i = 0; i < num_input_fields; i++) {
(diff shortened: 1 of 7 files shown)


backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -448,6 +448,7 @@ static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Cuda_gen(CeedOperator o
 
     CeedCallBackend(
         CeedTryRunKernelDimShared_Cuda(ceed, data->assemble_diagonal, NULL, grid, block[0], block[1], block[2], shared_mem, &is_run_good, opargs));
+    CeedCallCuda(ceed, cudaDeviceSynchronize());
```
jeremylt, line 451: Do you need sync if it is on the NULL steam?

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -497,6 +498,171 @@ static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Cuda_gen(CeedOperator o
...
+        CeedCallBackend(CeedFree(&points_per_elem));
+      }
+    }
+
+    // Get context data
+    CeedCallBackend(CeedQFunctionGetInnerContextData(qf, CEED_MEM_DEVICE, &qf_data->d_c));
+
+    // Assembly array
+    CeedCallBackend(CeedVectorGetArray(assembled, CEED_MEM_DEVICE, &assembled_array));
+    CeedScalar *assembled_array_offset = assembled_array + offset;
```
jeremylt:
```suggestion
    CeedScalar *offset_assembled_array = &assembled_array[offset];
```
minor, maybe this name? Also, a while back we swapped to using &foo[offset] instead of foo + offset for all the GPU backends

interface/ceed-preconditioning.c
```diff
@@ -2465,12 +2465,35 @@ int CeedOperatorLinearAssemble(CeedOperator op, CeedVector values) {
...
     CeedCall(CeedOperatorGetNumElements(op, &num_elem));
     if (num_elem == 0) return CEED_ERROR_SUCCESS;
+    has_linear_assemble_single = op->LinearAssembleSingle != NULL;
+  } else {
+    CeedCall(CeedCompositeOperatorGetNumSub(op, &num_suboperators));
+    CeedCall(CeedCompositeOperatorGetSubList(op, &sub_operators));
+    has_linear_assemble_single = true;
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      has_linear_assemble_single = has_linear_assemble_single && sub_operators[i]->LinearAssembleSingle != NULL;
+    }
```
jeremylt: I think it should be fine for some suboperators to use the interface impl and some to use the backend impl

zatkins-dev: hmm so the idea here was to short-circuit the fallback if all of the suboperators support single operator assembly. If I understand correctly, `CeedOperatorLinearAssemble` should assemble all of the suboperators, so the `else if` `else` options are mutually exclusive. The only way we get to the default interface at the bottom is if some of the suboperators don't have a backend specific `LinearAssembleSingle`, but their fallback operators do. In that case, is it better to use the (possibly more optimized) fallback `LinearAssemble`, or to use the  `LinearAssembleSingle` on each suboperator and never try the fallback `LinearAssemble`?

backends/cuda-gen/ceed-cuda-gen-operator.c
zatkins-dev: I don't think it hurts? We do for HIP on MI300As, so I figured it was good for consistency.

backends/cuda-gen/ceed-cuda-gen-operator.c
zatkins-dev: oh fair, works for me

interface/ceed-preconditioning.c
jeremylt: The priority I used for diagonal was
```C
  if (op->LinearAssembleAddDiagonal) {
    // Backend version
  } else if (is_composite) {
    // Default to summing contributions of suboperators
  } else {
    // Operator fallback
  }
  // Interface version
```
then inside of the composite, each sub uses the same priority order of 1) backend impl 2) fallback impl 3) interface impl

The thinking was to use the backend or fallback impl on as many of the subops as possible

zatkins-dev: oh, I think that was my original intention here, specifically see other thread

interface/ceed-preconditioning.c
```diff
@@ -2465,12 +2465,35 @@ int CeedOperatorLinearAssemble(CeedOperator op, CeedVector values) {
...
+      for (CeedInt k = 0; k < num_suboperators; k++) {
+        CeedCall(CeedSingleOperatorAssemble(sub_operators[k], offset, values));
+        CeedCall(CeedSingleOperatorAssemblyCountEntries(sub_operators[k], &single_entries));
+        offset += single_entries;
+      }
+    } else {
+      CeedCall(CeedSingleOperatorAssemble(op, offset, values));
+    }
+    return CEED_ERROR_SUCCESS;
   } else {
```
zatkins-dev:
```suggestion
  } else if (is_composite && has_linear_assemble_single) {
    // Default to summing contributions of suboperators
    CeedCall(CeedVectorSetValue(values, 0.0));
    if (num_suboperators > 0 && sub_operators[0]) {
      CeedCall(CeedCompositeOperatorGetNumSub(op, &num_suboperators));
      CeedCall(CeedCompositeOperatorGetSubList(op, &sub_operators));
      for (CeedInt k = 0; k < num_suboperators; k++) {
        CeedCall(CeedSingleOperatorAssemble(sub_operators[k], offset, values));
        CeedCall(CeedSingleOperatorAssemblyCountEntries(sub_operators[k], &single_entries));
        offset += single_entries;
      }
    }
    return CEED_ERROR_SUCCESS;
  } else {
```

Would this be better @jeremylt

jeremylt: Do we need to check for `has_linear_assemble_single` or can we let `CeedSingleOperatorAssemble` determine the best way to assemble the single operator?

zatkins-dev: It's possible that none of the suboperators have `LinearAssembleSingle`, but the composite operator does have a fallback with `LinearAssemble`, right?

zatkins-dev: In that case we want to fallback and use the `LinearAssemble` before we try (and fail) to use single operator assemble

zatkins-dev:
I don't understand how I failed codecov before but it's fine now

interface/ceed-preconditioning.c
jeremylt: Currently we only have implementations of `LinearAssembleSingle` and none of `LinearAssemble`.

We really, really want to avoid gen going down to ref, as that blows up our memory usage. We could find ourselves in a situation with the code in this PR where only some of the suboperators of the composite operator support `LinearAssembleSingle` in gen, and I'd like as many of those suboperators to use the native gen assembly as possible instead of falling back to ref preemptively

zatkins-dev: hmm alright, that makes sense

jeremylt: Really, since assembly is much less frequent than operator application, I'm less worried about handling this as one big composite operation as we've got bigger perf targets that should be more impactful

zatkins-dev: Alright, just updated. lmk what you think

jeremylt: Close, I think we need this?
```C
int CeedOperatorLinearAssemble(CeedOperator op, CeedVector values) {
  bool          is_composite;
  CeedInt       num_suboperators, offset = 0;
  CeedSize      single_entries = 0;
  CeedOperator *sub_operators;

  CeedCall(CeedOperatorCheckReady(op));
  CeedCall(CeedOperatorIsComposite(op, &is_composite));

  // Early exit for empty operator
  if (!is_composite) {
    CeedInt num_elem = 0;

    CeedCall(CeedOperatorGetNumElements(op, &num_elem));
    if (num_elem == 0) return CEED_ERROR_SUCCESS;
  }

  if (op->LinearAssemble) {
    // Backend version
    CeedCall(op->LinearAssemble(op, values));
    return CEED_ERROR_SUCCESS;
  } else if (is_composite) {
    // Default to summing contributions of suboperators
    CeedCall(CeedVectorSetValue(values, 0.0));
    CeedCall(CeedCompositeOperatorGetNumSub(op, &num_suboperators));
    for (CeedInt k = 0; k < num_suboperators; k++) {
      CeedCall(CeedSingleOperatorAssemble(sub_operators[k], offset, values));
      CeedCall(CeedSingleOperatorAssemblyCountEntries(sub_operators[k], &single_entries));
      offset += single_entries;
    }
    return CEED_ERROR_SUCCESS;
  } else {
    // Operator fallback
    CeedOperator op_fallback;

    CeedCall(CeedOperatorGetFallback(op, &op_fallback));
    if (op_fallback) {
      CeedCall(CeedOperatorLinearAssemble(op_fallback, values));
      return CEED_ERROR_SUCCESS;
    }
  }

  // Default to interface version if non-composite and no fallback
  CeedCall(CeedVectorSetValue(values, 0.0));
  CeedCall(CeedSingleOperatorAssemble(op, offset, values));
  return CEED_ERROR_SUCCESS;
}
```

zatkins-dev: no that doesn't work, specifically `/gpu/cuda/gen` skips on `t596`. That's because we don't have a working `LinearAssemble` on the fallback, but we passed in a non-composite operator.

zatkins-dev: the path is then:
- no `op->LinearAssemble`
- not composite
- there is an `op_fallback`, so call `CeedOperatorLinearAssemble(op_fallback, values)`
- `op_fallback` throws an error because it doesn't have a valid `LinearAssemble`

zatkins-dev: so we never actually make it to the default at the bottom

jeremylt: Oh, duh
```C
int CeedOperatorLinearAssemble(CeedOperator op, CeedVector values) {
  bool          is_composite;
  CeedInt       num_suboperators, offset = 0;
  CeedSize      single_entries = 0;
  CeedOperator *sub_operators;

  CeedCall(CeedOperatorCheckReady(op));
  CeedCall(CeedOperatorIsComposite(op, &is_composite));

  // Early exit for empty operator
  if (!is_composite) {
    CeedInt num_elem = 0;

    CeedCall(CeedOperatorGetNumElements(op, &num_elem));
    if (num_elem == 0) return CEED_ERROR_SUCCESS;
  }

  if (op->LinearAssemble) {
    // Backend version
    CeedCall(op->LinearAssemble(op, values));
    return CEED_ERROR_SUCCESS;
  } else if (is_composite) {
    // Default to summing contributions of suboperators
    CeedCall(CeedVectorSetValue(values, 0.0));
    CeedCall(CeedCompositeOperatorGetNumSub(op, &num_suboperators));
    for (CeedInt k = 0; k < num_suboperators; k++) {
      CeedCall(CeedSingleOperatorAssemble(sub_operators[k], offset, values));
      CeedCall(CeedSingleOperatorAssemblyCountEntries(sub_operators[k], &single_entries));
      offset += single_entries;
    }
    return CEED_ERROR_SUCCESS;
  } else if (op->LinearAssembleSingle) {
    CeedCall(CeedVectorSetValue(values, 0.0));
    CeedCall(CeedSingleOperatorAssemble(op, offset, values));
    return CEED_ERROR_SUCCESS;
  } else {
    // Operator fallback
    CeedOperator op_fallback;

    CeedCall(CeedOperatorGetFallback(op, &op_fallback));
    if (op_fallback) {
      CeedCall(CeedOperatorLinearAssemble(op_fallback, values));
      return CEED_ERROR_SUCCESS;
    }
  }

  // Default to interface version if non-composite and no fallback
  CeedCall(CeedVectorSetValue(values, 0.0));
  CeedCall(CeedSingleOperatorAssemble(op, offset, values));
  return CEED_ERROR_SUCCESS;
}
```
there we go, this should be better

jeremylt: Basically, at the bottom we still need to try the interface version before quitting

zatkins-dev: yep that works, thanks!

jeremylt:
Oh, CodeCov is having some errors. So ignore that

PR #1840, opened 2025-06-20
feat(sycl): Define CEED_RUNNING_JIT_PASS in JIT

jrwrigh:
Update to include `CEED_RUNNING_JIT_PASS` to the SYCL backend's definitions. This was added in #1696, but not to the SYCL backend.

Confirmed this works on Aurora.

### DIFF
--- a/backends/sycl/ceed-sycl-compile.sycl.cpp
+++ b/backends/sycl/ceed-sycl-compile.sycl.cpp
@@ -61,7 +61,7 @@ static int CeedJitAddDefinitions_Sycl(Ceed ceed, const std::string &kernel_sourc
 // TODO: Add architecture flags, optimization flags
 //------------------------------------------------------------------------------
 static inline int CeedJitGetFlags_Sycl(std::vector<std::string> &flags) {
-  flags = {std::string("-cl-std=CL3.0"), std::string("-Dint32_t=int")};
+  flags = {std::string("-cl-std=CL3.0"), std::string("-Dint32_t=int"), std::string("-DCEED_RUNNING_JIT_PASS=1")};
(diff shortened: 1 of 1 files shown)


jrwrigh:
@kris-rowe @uumesh Any foreseen issues with this?

kris-rowe:
Not that I can think of. Does the kernel generation in the SYCL backend need to be updated to use include flags?

uumesh:
Is this addition just for consistency? I don't think it will have any effect unless the JIT source is also modified.

jrwrigh:
> Is this addition just for consistency? I don't think it will have any effect unless the JIT source is also modified.

For consistency, and also to allow including character arrays in the header files ([see MR here](https://gitlab.com/phypid/honee/-/merge_requests/123)). The latter is more for coding convenience than anything else.

jrwrigh:
> Does the kernel generation in the SYCL backend need to be updated to use include flags?

It'd probably be nice in the future, but it's not blocking anything in HONEE.

uumesh:
That should be fine. I also see that the flag adds a guard to exclude std/math libraries from the QFunctions. I am not sure if that will affect the compilation of OpenCL kernels. For example in the advection functions [here](https://github.com/CEED/libCEED/pull/1696/files#diff-bab1280d31cf5e98c99acf2a7c3e708571414a19ea4c67ccdc48eef6fd2334de)
If you can this test passes, I think the rest should be fine as well.

jrwrigh:
It's been passing tests in HONEE, I'll double check with the libCEED tests real quick though. (HONEE should be a super set of the libCEED fluids tests).

PR #1845, opened 2025-06-26
OP Get Name Helper

jeremylt:
Minor improvement in debugging output


interface/ceed-preconditioning.c
```diff
@@ -1838,14 +1838,16 @@ int CeedOperatorGetFallback(CeedOperator op, CeedOperator *op_fallback) {
...
+      const char *resource, *resource_fallback, *op_name;
 
       CeedCall(CeedGetOperatorFallbackCeed(ceed, &ceed_fallback));
       CeedCall(CeedGetResource(ceed, &resource));
       CeedCall(CeedGetResource(ceed_fallback, &resource_fallback));
+      CeedCall(CeedOperatorGetName(op, &op_name));
 
       CeedDebug256(ceed, CEED_DEBUG_COLOR_SUCCESS, "---------- CeedOperator Fallback ----------\n");
-      CeedDebug(ceed, "Falling back from %s operator at address %p to %s operator at address %p\n", resource, op, resource_fallback, op->op_fallback);
+      CeedDebug(ceed, "CeedOperator %s: falling back from %s operator at address %p to %s operator at address %p\n", op_name, resource, op,
```
jrwrigh: The nomenclature of "<backend_resource> operator" is because it's the operator based on that resource?

Seems a bit odd at first glance, but I think it makes sense in the context of it being in the Fallback functionality.

jrwrigh (approved):

jeremylt:
I'll need to do some more fixes here

interface/ceed-preconditioning.c
jeremylt: Yeah, honestly I didn't think much about the specific nomenclature when I first wrote this years ago since its a debug statement. I clarified it a bit

PR #1850, opened 2025-07-02
Add sychronization calls after hipMemcpyDeviceToDevice

zatkins-dev:
Turns out that `hipMemcpy` with `hipMemcpyDeviceToDevice` is asynchronous on the host. This MR adds synchronization calls to prevent any potential race conditions.

### DIFF
--- a/backends/hip-ref/ceed-hip-ref-qfunctioncontext.c
+++ b/backends/hip-ref/ceed-hip-ref-qfunctioncontext.c
@@ -194,6 +194,7 @@ static int CeedQFunctionContextSetDataDevice_Hip(const CeedQFunctionContext ctx,
       impl->d_data_borrowed = NULL;
       impl->d_data          = impl->d_data_owned;
       CeedCallHip(ceed, hipMemcpy(impl->d_data, data, ctx_size, hipMemcpyDeviceToDevice));
+      CeedCallHip(ceed, hipDeviceSynchronize());
     } break;
     case CEED_OWN_POINTER:
       impl->d_data_owned    = data;
--- a/backends/hip/ceed-hip-common.c
+++ b/backends/hip/ceed-hip-common.c
@@ -68,7 +68,10 @@ static inline int CeedSetDeviceGenericArray_Hip(Ceed ceed, const void *source_ar
           *(void **)target_array = *(void **)target_array_owned;
         }
       }
-      if (source_array) CeedCallHip(ceed, hipMemcpy(*(void **)target_array, source_array, size_unit * num_values, hipMemcpyDeviceToDevice));
+      if (source_array) {
+        CeedCallHip(ceed, hipMemcpy(*(void **)target_array, source_array, size_unit * num_values, hipMemcpyDeviceToDevice));
+        CeedCallHip(ceed, hipDeviceSynchronize());
+      }
       break;
     case CEED_OWN_POINTER:
       CeedCallHip(ceed, hipFree(*(void **)target_array_owned));


jeremylt:
Can you check if CUDA needs the same?

zatkins-dev:
I looked and honestly I have no idea. I can check more next week.

zatkins-dev:
This thread has some insight, the answer is we probably don't need to sync, but it isn't host synchronous. I'm not sure if hip is the same way, i.e. if it's okay that it's not synchronous on host.

https://forums.developer.nvidia.com/t/are-cudamemcpy-and-cudamalloc-blocking-synchronous/308368/2

jeremylt:
We run the suboperators of a composite operator on separate streams.

What would happen if we 
1) do a device to device transfer that doesn't immediately resolve
2) get the address of the (stale) memory, believing its up to date
3) launch an operator kernel with this stale memory

nbeams:
Ah, I see. So in general, a `Ceed` object is associated with one stream only, with everything submitted there, with the suboperators being an exception? But then for each suboperator, everything happens in its own stream.

Looking at the code in https://github.com/CEED/libCEED/commit/6eee1ffc21ba1b6c5586e8a6fa353666fb694027, it looks like we're already syncing each stream before moving on, so I don't see a problem. I assume there's no "crossing the streams" between the suboperators which is why we can put them in separate streams to start? Even if there were dependencies, it'd be better to have a way to handle that other than putting a whole device sync in `CeedSetDeviceGenericArray`, which would be more heavy-handed than we really need.

jeremylt:
We're the ones doing the operating

nbeams:
By "we", do you mean libCEED itself -- but as called by some other code/application (e.g., could be previous calls to libCEED that haven't finished yet)?

Right now, the operator apply with suboperators is the outlier that behaves differently than the other backend functions for CUDA/HIP, so I think ideally, it should be that function's "responsibility" to handle it, rather than forcing a device sync on every devicetodevice copy call (which might mean we end up waiting on another queue that has no impact on us). But to make sure that previous libCEED calls are finished before we split into the other streams for the suboperators, we'd have to add a sync on the NULL stream (which all the other libCEED calls use)...which has the same effect, since it forces any other streams to wait, too. And that would probably be worse in terms of performance since I don't know how often the devicetodevice copy is actually getting called in a typical workflow...

Unless you want to implement `CeedSetStream` for CUDA/HIP, and say that if you want to use composite operators, that requires using libCEED with a non-NULL stream. Then we could wait on that stream at the beginning of `CeedOperatorApplyAddComposite_[Hip/Cuda]_gen`, before splitting into the separate streams for the suboperators. But that's a big change.

jeremylt:
If a user is bypassing our access model, there's nothing we can do about that. It will always be a bug.

> have we added the possibility that even if they're using the NULL stream only, there could be ordering problems when applying a libCEED Operator?

No, we synchronize before returning

nbeams:
> No, we synchronize before returning

I know that, the question was just a rephrasing of my point about things not finishing before the Operator applies (i.e., that's where we introduce the ordering problem). In fact, I take back what I said about the Vector state monitoring, because `CeedVectorRestoreArray` also doesn't wait on any kernels to finish, it's just a host operation, right? So a previous `CeedOperator` could have launched its last/only kernel, writing output into a Vector, and then the composite operator starts with the same Vector as input, but its data isn't finished being updated on the device yet. When everyone uses the same stream, it doesn't matter; if we introduce other streams, it (theoretically) could?

PR #1853, opened 2025-07-10
gen - add support for mixed precision operators

zatkins-dev:
Adds support for mixed precision operators, set at creation time, for the CUDA gen backend. This is made possible by defining a second scalar type `CeedScalarCPU`, which is always double precision, and changing the `CeedScalar` type during the JiT pass if the `CEED_JIT_MIXED_PRECISION` define is set.

The inputs to the device kernels are always `CeedScalarCPU` arrays, to avoid having to muck around with multiple pointers and such in a `CeedVector`. In gen, we only do things to the input and output arrays at the beginning and end of the kernel, so all of the computation will be happening with the single precision `CeedScalar` arrays we copy values into. This approach minimizes the code differences between mixed and full precision runs, essentially just requiring the helper functions to have extra template parameters to ensure the input types are correct.

The support for mixed precision operators is at the backend level, while the actual usage of mixed precision operations is defined per-operator to provide maximal flexibility.

This can be extended to the CUDA ref backend too, though the benefits will likely be more mild.

@jeremylt and @nbeams, does this seem like a reasonable approach?


zatkins-dev:
Should we change the way flops are reported (say, cut them in half) for mixed precision operators?

jrwrigh:
> Should we change the way flops are reported (say, cut them in half) for mixed precision operators?

I don't think so; a FLOP is a FLOP. There can be some other mechanism for adjusting that if the end-user wants to, but by default I don't think the count should be changed.

interface/ceed-operator.c
```diff
@@ -634,6 +634,46 @@ int CeedOperatorIsSetupDone(CeedOperator op, bool *is_setup_done) {
...
+/**
+  @brief Set a `CeedOperator` to use reduced precision for operator application
+
+  @param[in] op `CeedOperator`
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedOperatorSetMixedPrecision(CeedOperator op) {
```
jrwrigh: This should take a `bool` if we want to go with this interface.

jrwrigh: I think a better interface would specify the precision that the operator will use internally via an enum of somesort. This will make it more future proof in case we want to experiment with other precisions. Something like `CEED_PRECISION_SINGLE`, `CEED_PRECISION_DOUBLE`, etc. (if such an enum doesn't already exist).

interface/ceed-basis.c
```diff
@@ -1303,7 +1303,7 @@ int CeedSymmetricSchurDecomposition(Ceed ceed, CeedScalar *mat, CeedScalar *lamb
 
   // Reduce sub and super diagonal
   CeedInt    p = 0, q = 0, itr = 0, max_itr = n * n * n * n;
-  CeedScalar tol = CEED_EPSILON;
+  CeedScalar tol = 10 * CEED_EPSILON;
```
jrwrigh, line 1332: Stray temp fix?

zatkins-dev: no that's a real fix. It isn't able to get to machine precision, now that `CEED_EPSILON` is machine precision

jeremylt: probably worth putting into a separate PR if it applies to main?

interface/ceed-operator.c
zatkins-dev: Yeah I agree. I'll think about the best way to support it; I can't imagine a use case for using a higher precision for the operator than for the entire solve, and it needs to be able to be a compile time constant for the JiT to work

zatkins-dev: Okay I think changing that makes the interface way more clear. Also, I can think of a use case; numerically unstable implementations of QFunctions that people aren't interesting in fixing.

zatkins-dev: Just updated, lmk what you think

interface/ceed-basis.c
zatkins-dev: it technically only applies here, since `CEED_EPSILON` is a bit bigger on main, but fair

interface/ceed-operator.c
```diff
@@ -634,6 +634,49 @@ int CeedOperatorIsSetupDone(CeedOperator op, bool *is_setup_done) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Set a `CeedOperator` to use reduced precision for operator application
```
nbeams: This only mentions using reduced precision in the operator, but the tests show using whichever is the "opposite" of the base precision, no? From the comments, it seems like this is a change that was introduced after the initial addition and the docstring just needs to be updated.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
```diff
@@ -1285,7 +1285,7 @@ extern "C" int CeedOperatorBuildKernel_Cuda_gen(CeedOperator op, bool *is_good_b
   code << tab << "// s_G_[in,out]_i: Gradient matrix, shared memory\n";
   code << tab << "// -----------------------------------------------------------------------------\n";
   code << tab << "extern \"C\" __global__ void " << operator_name
-       << "(CeedInt num_elem, void* ctx, FieldsInt_Cuda indices, Fields_Cuda fields, Fields_Cuda B, Fields_Cuda G, CeedScalar *W, Points_Cuda "
+       << "(CeedInt num_elem, void* ctx, FieldsInt_Cuda indices, Fields_Cuda fields, Fields_Cuda B, Fields_Cuda G, CeedScalarCPU *W, Points_Cuda "
```
nbeams: Personally, I'm not a fan of the `CeedScalarCPU` name, especially since it's the precision that will be used by other GPU things beyond the mixed precision operators, as well, not just on CPU. Is there a reason why we can't keep this as `CeedScalar` (the base precision), and use a different name for the precision that will be changed in the operator? Like `CeedScalarOpCompute`, `CeedScalarMixedCompute`, some other better name someone else comes up with...
Or `CeedScalarCPU` could maybe be `CeedScalarBase`?

zatkins-dev: The issue with that is that user code, e.g. in QFunctions, uses `CeedScalar` internally. By changing what `CeedScalar` means, we can prevent needing all user code to use a different scalar type to support mixed precision operators.

I think `CeedScalarBase` would be reasonable, though.

interface/ceed-operator.c
zatkins-dev: Ah good catch, that docstring is old

tests/t506-operator-mixed.c
```diff
@@ -0,0 +1,176 @@
...
+  CeedQFunctionAddInput(qf_mass, "rho", 1, CEED_EVAL_NONE);
+  CeedQFunctionAddInput(qf_mass, "u", num_comp, CEED_EVAL_INTERP);
+  CeedQFunctionAddOutput(qf_mass, "v", num_comp, CEED_EVAL_INTERP);
+
+  // 'Small' Operators
+  CeedOperatorCreate(ceed, qf_setup, CEED_QFUNCTION_NONE, CEED_QFUNCTION_NONE, &op_setup_small);
+  CeedOperatorSetField(op_setup_small, "weight", CEED_ELEMRESTRICTION_NONE, basis_x_small, CEED_VECTOR_NONE);
+  CeedOperatorSetField(op_setup_small, "x", elem_restriction_x, basis_x_small, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetField(op_setup_small, "rho", elem_restriction_q_data_small, CEED_BASIS_NONE, CEED_VECTOR_ACTIVE);
+  CeedOperatorSetPrecision(op_setup_small, CEED_SCALAR_TYPE == CEED_SCALAR_TYPE == CEED_SCALAR_FP32 ? CEED_SCALAR_FP64
```
jedbrown: ternary `==` isn't a thing in C. Maybe we should have a helper to deal with this logic.

jedbrown: How much of these examples is actually different between the mixed and non-mixed versions? Seems like mostly copy/paste and I don't love that quantity of duplication.

Also, we usually only use the same number for the C and equivalent Fortran test, not for further variability. Could we just make some of the regular ones able to use mixed precision, possibly by running both variants?

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
jedbrown: Yeah, that is the "at rest" or "storage" or "outer" representation (not opposed to "base" either).

What do we anticipate doing about `qdata`? Will it also always be converted for the qfunction or might it be stored in the inner representation? (I can imagine running kernels is f32 while Krylov runs in f64, though it looks like we've been going the other way in examples.)

tests/t506-operator-mixed.c
zatkins-dev: The ternary is a typo.

zatkins-dev: There's essentially no difference. I could also just make it a command line flag for those tests.

backends/cuda-gen/ceed-cuda-gen-operator-build.cpp
zatkins-dev: The qdata is copied into arrays of type `CeedScalar` inside the kernel. So the user will keep using their base precision and it will be internally converted for the QFunction.

zatkins-dev: The examples go both ways - both higher and lower precision for the internal operator. I think that the f64 base, f32 inner is the primary use case though.

jedbrown: We may later want to support qdata at rest in the inner representation (specifically, when the inner representation is smaller). I can imagine that being done by making `CeedVector` handle two pointers (it won't even need axpy and dot product functionality).

jeremylt: You're anticipating a pref benefit from copying single->single instead of double->single? Makes sense, though we'd probably want to be a bit thoughtful about possible cases where the vector is created in double and edited in single?

zatkins-dev: That kind of dual pointer complexity was what I was trying to avoid with this approach. For GPUs, that adds substantially more complexity due to already having 2+ pointers.

jedbrown: The perf benefit would be from having the heaviest data structure in the operator be half the size at rest. I definitely don't want to scope-creep it into this PR, but I think it's something very natural for people to want. The complexity is sort of an outer product with the host/device memory spaces, though we don't need to implement the algebraic operations if we're only using that functionality for qdata.

zatkins-dev:
I think there is still a lot of value here. The biggest sticking point is the at-rest data type, which I agree is pretty important for a lot of memory constrained cases.

I think the most straightforward approach is:
- Add `CeedVector[Get|Restore]Array[|Read|Write]F32`, which give/restore a F32 version of the underlying array. These helpers can then be used to get the low-precision at-rest data for mixed-precision operators. This solves the issue of user data (like passive inputs) being in the heavier format, since we can just move it to the lighter format. This increases the overall memory size, but maximizes flexibility. It's possible that we want to use the same q data for different precision operators -- this ensures that it just works without users having to manage separate vectors.
- Probably also need  `CeedVector[Get|Restore]Array[|Read|Write]F64` for parity.
- Some kind of runtime setting for each vector which decides whether to use float or double for cublas/hipblas operations would also be useful, but isn't completely necessary for this PR.

zatkins-dev:
I do not think that it is worth the complexity to support type generic interfaces of other functions which take raw pointers, even if there could be performance gains there. The concern about float input/output with double in the computation is interesting, though that is potentially a lot of data movement to set up inputs and outputs. I think that could still be supported with the current approach here, just with some extra setter functions.

Overall, I think that the user interface needs to stay in terms of `CeedScalar`, set at compile time, and all mixed precision should be done opaquely. The users should input and output `CeedVector`s which they can request the desired data type from.

PR #1854, opened 2025-07-11
Minor reduction in AtPoints grad FLOPs

jeremylt:
@zatkins-dev 

What do you think about this adjustment to reduce grad FLOPs in an effort to close the difference between grad and interp?


include/ceed/jit-source/cuda/cuda-shared-basis-tensor-at-points-templates.h
```diff
@@ -376,22 +376,20 @@ inline __device__ void GradAtPoints3d(SharedData_Cuda &data, const CeedInt p, co
     CeedScalar buffer[Q_1D];
     CeedScalar chebyshev_x[Q_1D];
 
-    for (CeedInt comp = 0; comp < NUM_COMP; comp++) {
-      // Get z contraction value
-      ChebyshevPolynomialsAtPoint<Q_1D>(r_X[2], chebyshev_x);
-      CeedScalar z = chebyshev_x[k];
+    // Get z contraction values
```
jeremylt, line 379: Lifting this here cuts out `(num_comp - 1)` evals of cheby and dcheby

PR #1856, opened 2025-07-15
ceed - replace usage of `ceed->op_fallback_parent` with `ceed->parent`

zatkins-dev:
The `ceed->op_fallback_parent` member is redundant and creates potential data mismatches between a given `Ceed` context and its children. This removes the need to manually copy over JiT source roots and defines from the parent to fallback `Ceed` and prevents possible footguns when using the `Ceed*Parent*` interfaces.

### DIFF
--- a/include/ceed-impl.h
+++ b/include/ceed-impl.h
@@ -96,7 +96,7 @@ struct Ceed_private {
   Ceed         parent;
   ObjDelegate *obj_delegates;
   int          obj_delegate_count;
-  Ceed         op_fallback_ceed, op_fallback_parent;
+  Ceed         op_fallback_ceed;
   const char  *op_fallback_resource;
   char       **jit_source_roots;
   CeedInt      num_jit_source_roots, max_jit_source_roots, num_jit_source_roots_readers;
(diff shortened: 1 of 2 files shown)


jeremylt (commented):
The parent and fallback parent are different logically though

zatkins-dev:
So perhaps they are logically distinct, but abstractly I think they serve the same role

zatkins-dev:
Hmm that seems reasonable, since then parent would be the same anyway. To me, fallback is essentially a weak delegate. In fact, I feel like we could probably replace most instances of `CeedOperatorGetFallback` with `CeedGetObjectDelegate("Operator", ...)` and have things work fine. I'd have to test it though. 

I understand if that's mixing different logical components too much though.

jeremylt:
I am not sure that the delegate will work. A delegate is designed to fully replace the full object functionally.

jeremylt:
Really, Ratel running correctly with this branch for GPU would verify the changes

zatkins-dev:
The JiT paths no longer need to be copied at all, since, the JiT paths come from the parent. I've been running Ratel with this all day, so I think it's good!

jeremylt:
Cool, my only quibble now is that the developer documentation needs to be updated to reflect this

jeremylt (approved):
Thanks for this update

PR #1861, opened 2025-07-16
op - Add `CeedOperatorApplyAddActive()` for only summing into active outputs

zatkins-dev:
See https://gitlab.com/micromorph/ratel/-/merge_requests/1164 for more discussion.

Often, the passive outputs of a `CeedOperator` should always be overwritten; e.g., auxiliary data from a residual evaluation used for Jacobian evaluation. Without this interface, one must use `CeedOperatorApply()` even if the desired behavior is to add into the active output vector, resulting in needing an extra scratch vector.

### DIFF
--- a/interface/ceed-operator.c
+++ b/interface/ceed-operator.c
@@ -1076,10 +1076,10 @@ int CeedOperatorAtPointsSetPoints(CeedOperator op, CeedElemRestriction rstr_poin
 
 /**
   @brief Get a boolean value indicating if the `CeedOperator` was created with `CeedOperatorCreateAtPoints`
-    
+
   @param[in]  op           `CeedOperator`
   @param[out] is_at_points Variable to store at points status
-  
+
   @return An error code: 0 - success, otherwise - failure
 
   @ref User
@@ -2142,7 +2142,7 @@ int CeedOperatorRestoreContextBooleanRead(CeedOperator op, CeedContextFieldLabel
   This computes the action of the operator on the specified (active) input, yielding its (active) output.
   All inputs and outputs must be specified using @ref CeedOperatorSetField().
 
-  Note: Calling this function asserts that setup is complete and sets the `CeedOperator` as immutable.
+  @note Calling this function asserts that setup is complete and sets the `CeedOperator` as immutable.
 
   @param[in]  op      `CeedOperator` to apply
   @param[in]  in      `CeedVector` containing input state or @ref CEED_VECTOR_NONE if there are no active inputs
@@ -2223,6 +2223,10 @@ int CeedOperatorApply(CeedOperator op, CeedVector in, CeedVector out, CeedReques
   This computes the action of the operator on the specified (active) input, yielding its (active) output.
   All inputs and outputs must be specified using @ref CeedOperatorSetField().
 
+  @note Calling this function asserts that setup is complete and sets the `CeedOperator` as immutable.
+  @warning This function adds into ALL outputs, including passive outputs. To only add into the active output, use `CeedOperatorApplyAddActive()`.
+  @see `CeedOperatorApplyAddActive()`
+
   @param[in]  op      `CeedOperator` to apply
   @param[in]  in      `CeedVector` containing input state or @ref CEED_VECTOR_NONE if there are no active inputs
   @param[out] out     `CeedVector` to sum in result of applying operator (must be distinct from `in`) or @ref CEED_VECTOR_NONE if there are no active outputs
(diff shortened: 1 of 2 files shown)


interface/ceed-operator.c
```diff
@@ -2259,6 +2263,88 @@ int CeedOperatorApplyAdd(CeedOperator op, CeedVector in, CeedVector out, CeedReq
...
+
+        CeedCall(CeedOperatorGetFields(sub_operators[i], NULL, NULL, &num_output_fields, &output_fields));
+        for (CeedInt j = 0; j < num_output_fields; j++) {
+          CeedVector vec;
+
+          CeedCall(CeedOperatorFieldGetVector(output_fields[j], &vec));
+          if (vec != CEED_VECTOR_ACTIVE && vec != CEED_VECTOR_NONE) {
+            CeedCall(CeedVectorSetValue(vec, 0.0));
+            CeedCall(CeedVectorDestroy(&vec));
+          }
```
jeremylt:
```suggestion
          if (vec != CEED_VECTOR_ACTIVE && vec != CEED_VECTOR_NONE) {
            CeedCall(CeedVectorSetValue(vec, 0.0));
          }
          CeedCall(CeedVectorDestroy(&vec));
```

interface/ceed-operator.c
```diff
@@ -2259,6 +2263,88 @@ int CeedOperatorApplyAdd(CeedOperator op, CeedVector in, CeedVector out, CeedReq
...
+        bool       is_active;
+        CeedVector vec;
+
+        CeedCall(CeedOperatorFieldGetVector(output_fields[i], &vec));
+        is_active = vec == CEED_VECTOR_ACTIVE;
+        if (is_active) vec = out;
+        if (!is_active && vec != CEED_VECTOR_NONE) {
+          CeedCall(CeedVectorSetValue(vec, 0.0));
+          CeedCall(CeedVectorDestroy(&vec));
+        }
```
jeremylt:
```suggestion
        if (!is_active && vec != CEED_VECTOR_NONE) {
          CeedCall(CeedVectorSetValue(vec, 0.0));
        }
        CeedCall(CeedVectorDestroy(&vec));
```

interface/ceed-operator.c
zatkins-dev: Why the double destroy?

interface/ceed-operator.c
```diff
@@ -2259,6 +2263,88 @@ int CeedOperatorApplyAdd(CeedOperator op, CeedVector in, CeedVector out, CeedReq
...
+
+      CeedCall(CeedOperatorGetFields(op, NULL, NULL, &num_output_fields, &output_fields));
+      // Zero all output vectors
+      for (CeedInt i = 0; i < num_output_fields; i++) {
+        bool       is_active;
+        CeedVector vec;
+
+        CeedCall(CeedOperatorFieldGetVector(output_fields[i], &vec));
+        is_active = vec == CEED_VECTOR_ACTIVE;
+        if (is_active) vec = out;
```
jeremylt: don't need `is_active`

interface/ceed-operator.c
```diff
@@ -2259,6 +2263,88 @@ int CeedOperatorApplyAdd(CeedOperator op, CeedVector in, CeedVector out, CeedReq
...
+            CeedCall(CeedVectorDestroy(&vec));
+          }
+        }
+      }
+      // ApplyAdd
+      CeedCall(CeedOperatorApplyAdd(op, in, out, request));
+    }
+  } else {
+    // Standard Operator
+    if (op->ApplyAddActive) {
```
jeremylt: We shouldn't need an `ApplyAddActive` method

interface/ceed-operator.c
zatkins-dev: Ah I see, the `CEED_VECTOR_NONE` and `CEED_VECTOR_ACTIVE` are ignored for both `CeedVectorReferenceCopy` and `CeedVectorDestroy`

interface/ceed-operator.c
zatkins-dev: Fair, I just added for consistency. I can remove

interface/ceed-operator.c
jeremylt: I forgot that the browser interface doesn't let me drag lines and copies instead. Should be just moving the line

jeremylt:
We can drop some duplication by:

Apply zeros active vectors and calls ApplyAddActive

ApplyAddActive zeros passive vectors and calls ApplyAdd

PR #1868, opened 2025-07-18
Pull BasisIsCollocated helper to interface level

jeremylt:
This will help add this to gen/shared cleanly

### DIFF
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -783,8 +783,8 @@ int CeedBasisIsTensor(CeedBasis basis, bool *is_tensor) {
 /**
   @brief Determine if given `CeedBasis` has nodes collocated with quadrature points
 
-  @param[in]  basis     `CeedBasis`
-  @param[out] is_tensor Variable to store collocated status
+  @param[in]  basis         `CeedBasis`
+  @param[out] is_collocated Variable to store collocated status
 
   @return An error code: 0 - success, otherwise - failure
 
--- a/interface/ceed-basis.c
+++ b/interface/ceed-basis.c
@@ -781,6 +781,32 @@ int CeedBasisIsTensor(CeedBasis basis, bool *is_tensor) {
 }
 
 /**
+  @brief Determine if given `CeedBasis` has nodes collocated with quadrature points
+
+  @param[in]  basis     `CeedBasis`
+  @param[out] is_tensor Variable to store collocated status
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedBasisIsCollocated(CeedBasis basis, bool *is_collocated) {
+  if (basis->is_tensor_basis && (basis->Q_1d == basis->P_1d)) {
+    *is_collocated = true;
+
+    for (CeedInt i = 0; i < basis->P_1d; i++) {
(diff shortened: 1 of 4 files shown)


interface/ceed-basis.c
```diff
@@ -780,6 +780,32 @@ int CeedBasisIsTensor(CeedBasis basis, bool *is_tensor) {
...
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedBasisIsCollocated(CeedBasis basis, bool *is_collocated) {
+  if (basis->is_tensor_basis && (basis->Q_1d == basis->P_1d)) {
+    *is_collocated = true;
+
+    for (CeedInt i = 0; i < basis->P_1d; i++) {
+      *is_collocated = *is_collocated && (fabs(basis->interp_1d[i + basis->P_1d * i] - 1.0) < 10 * CEED_EPSILON);
```
jrwrigh, line 798: What if the 1D basis and quadrature points aren't assigned? Is `interp_1d` and the like always defined?

interface/ceed-basis.c
```diff
@@ -780,6 +780,32 @@ int CeedBasisIsTensor(CeedBasis basis, bool *is_tensor) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Determine if given `CeedBasis` has nodes collocated with quadrature points
+
+  @param[in]  basis     `CeedBasis`
+  @param[out] is_tensor Variable to store collocated status
```
jrwrigh:
```suggestion
  @param[out] is_collocated Variable to store collocated status
```

jrwrigh (changes requested):

interface/ceed-basis.c
jeremylt: Those are always assigned for a tensor basis. Impossible to make a tensor basis without them

interface/ceed-basis.c
jeremylt: fixed via manual commit and rebase as this forces spacing changes

interface/ceed-basis.c
jrwrigh: Sure. I'm thinking about this for non-tensor basis, which can have collocated a collocated basis-quadrature pair.

jeremylt: currently not something we've implemented so we'd have to revisit when we find a use case

jrwrigh: My point is that a function called `CeedBasisIsCollocated()` should say whether the basis is collocated rather than whether it's collocated and tensor. I know this is a backend function, but it should still be clear what it does.

This would also make it more clear which cases are implemented and which are not in the backend code. Something like

```c
CeedBasisIsTensor(basis, &is_tensor);
CeedBasisIsCollocated(basis, &is_collocated);

if (is_tensor && is_collocated) {
... // optimized code path (e.g. CeedBasisApply == memcpy)
} else {
... // normal code path
}
```

is more clear than if we relied on `CeedBasisIsCollocated()` to be the entire determination of whether to take a more optimized code path.

jeremylt: Sure, but we don't currently have any non-tensor collocated sort of logic, so I don't want to spend too much effort trying to get ahead of use cases that don't currently exist

jrwrigh: Fair. The limitation should be documented then.
> Checking a non-tensor basis is not implemented and will always return false

PR #1869, opened 2025-07-18
ci - update test suite for parity with Ratel

zatkins-dev:
Adds `make test subsearch="..."`, where `subsearch` is interpreted as a regular expression used to filter test cases by name.

Lots of people have asked for this, it wasn't too bad to add in.

Also, adds some useful output (timing, better error formatting) to TAP.

I used that output to shorten some excessively long tests (>30 s).

The parsing for arguments to TESTCASE(...) are far more robust
- quotation marks are now optional and can be escaped as `\"`
- there can be spaces after the commas separating arguments
- parentheses are allowed, so long as they match
- the only disallowed character is `=` when quotation marks aren't used, otherwise it's fine

The arguments `cgns_tol`, `csv_rtol`, and `csv_ztol` are now supported for per-test case tolerance settings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -797,7 +797,7 @@ NPROC_POOL ?= 1
 export NPROC_POOL
 
 run-% : $(OBJDIR)/%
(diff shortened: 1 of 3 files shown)


jeremylt:
Why the drop in codecov? I don't see an obvious reason

zatkins-dev:
I really don't know...the fortran tests are definitely running

PR #1870, opened 2025-07-19
deal.II examples: some clean up

peterrum:


examples/deal.II/bps.h
```diff
@@ -670,15 +656,15 @@ class OperatorCeed : public OperatorBase<Number>
...
     CeedQFunctionSetContext(qf_build, build_ctx);
 
     // 6) put everything together
     CeedOperatorCreate(ceed, qf_build, CEED_QFUNCTION_NONE, CEED_QFUNCTION_NONE, &op_build);
     CeedOperatorSetField(op_build, "geo", geo_restriction, geo_basis, CEED_VECTOR_ACTIVE);
     CeedOperatorSetField(
-      op_build, "weights", CEED_ELEMRESTRICTION_NONE, geo_basis, CEED_VECTOR_NONE);
+      op_build, "metric_data", CEED_ELEMRESTRICTION_NONE, geo_basis, CEED_VECTOR_NONE);
     CeedOperatorSetField(
       op_build, "qdata", q_data_restriction, CEED_BASIS_NONE, CEED_VECTOR_ACTIVE);
```
jrwrigh: I'm not sure if changing the name of the operator field makes sense? The (old) `"weights"` field uses `CEED_EVAL_WEIGHT` means that this is in fact just going to be the quadrature weights. 

If I understand what you're doing, maybe you instead want to change `"qdata"` to be `"metric_data"`?

jeremylt: I think that indeed, our typical  phrase "qdata" is more commonly called "metric data" as it is the information to get from the physical element to the reference element.

Small quibble - we typically use spaces instead of underscores in the string field names, as they are intended to be human readable when we print out the Operator description (so `"metric data"` instead of `"metric_data"` is usual for libCEED)

peterrum: > I'm not sure if changing the name of the operator field makes sense? The (old) "weights" field uses CEED_EVAL_WEIGHT means that this is in fact just going to be the quadrature weights.

True. My fault.

> If I understand what you're doing, maybe you instead want to change "qdata" to be "metric_data"?

Good idea! Done!

jeremylt (approved):
Looks good to me, thanks!

PR #1875, opened 2025-07-24
Operator Name Consistency

jeremylt:
We have `CeedBasisCreateH1` and similar for almost all objects. I'm not sure why we treat `CeedCompositeOperatorCreate` differently?

### DIFF
--- a/include/ceed/ceed.h
+++ b/include/ceed/ceed.h
@@ -409,7 +409,7 @@ CEED_EXTERN int CeedQFunctionContextDestroy(CeedQFunctionContext *ctx);
 
 CEED_EXTERN int CeedOperatorCreate(Ceed ceed, CeedQFunction qf, CeedQFunction dqf, CeedQFunction dqfT, CeedOperator *op);
 CEED_EXTERN int CeedOperatorCreateAtPoints(Ceed ceed, CeedQFunction qf, CeedQFunction dqf, CeedQFunction dqfT, CeedOperator *op);
-CEED_EXTERN int CeedCompositeOperatorCreate(Ceed ceed, CeedOperator *op);
+CEED_EXTERN int CeedOperatorCreateComposite(Ceed ceed, CeedOperator *op);
 CEED_EXTERN int CeedOperatorReferenceCopy(CeedOperator op, CeedOperator *op_copy);
 CEED_EXTERN int CeedOperatorSetField(CeedOperator op, const char *field_name, CeedElemRestriction rstr, CeedBasis basis, CeedVector vec);
 CEED_EXTERN int CeedOperatorGetFields(CeedOperator op, CeedInt *num_input_fields, CeedOperatorField **input_fields, CeedInt *num_output_fields,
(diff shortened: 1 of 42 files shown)


jeremylt:
Note - Julia hating this is expected as it is a beaking change to the API. We should only merge this branch right before release

jeremylt:
Yeah, preserving the stubs to the handful of old names would be easy.

include/ceed/ceed.h
```diff
@@ -473,6 +473,14 @@ CEED_EXTERN int  CeedOperatorApplyAddActive(CeedOperator op, CeedVector in, Ceed
 CEED_EXTERN int  CeedOperatorAssemblyDataStrip(CeedOperator op);
 CEED_EXTERN int  CeedOperatorDestroy(CeedOperator *op);
 
+// Compatibility with previous composite CeedOperator naming
+#define CeedCompositeOperatorCreate(a, b) CeedOperatorCreateComposite(a, b)
+#define CeedCompositeOperatorAddSub(a, b) CeedOperatorCompositeAddSub(a, b)
+#define CeedCompositeOperatorGetNumSub(a, b) CeedOperatorCompositeGetNumSub(a, b)
+#define CeedCompositeOperatorGetSubList(a, b) CeedOperatorCompositeGetSubList(a, b)
+#define CeedCompositeOperatorGetSubByName(a, b) CeedOperatorCompositeGetSubByName(a, b, c)
+#define CeedCompositeOperatorGetMultiplicity(a, b, c, d) CeedOperatorCompositeGetMultiplicity(a, b, c, d)
```
jrwrigh: We could also put deprecation warning attribute on these too? Would require changing to actual function definitions rather than macro functions.

jedbrown: It gives better error messages anyway if these are replaced. They could be `static inline` unless they might need to be real symbols (callable via FFI or without recompiling source).

jeremylt: see the latest commit

jrwrigh: I think the `[[ deprecated() ]]` version of the attribute is [C23 only](https://en.cppreference.com/w/c/language/attributes/deprecated.html). We'll probably need to setup a compiler dependent macro to handle it.

jeremylt: I think its also a compiler extension in some cases

jrwrigh: It is, but the format pre-C23 format was different (e.g. `__attribute__ ((deprecated))` for GCC). PETSc should have a list of them for their deprecation macros.

jeremylt: I mean that I think using the new syntax may be supported by some compilers when compiling for an old standard but not in pedantic mode

jeremylt: of course I forgot that we had a pedantic build in CI

jrwrigh: Ah, gotcha. Do we have a way of verifying when those would be extensions were available in compilers? I'm guessing it would be sometime before C23 was fully available, but not super far past that?

jedbrown: Is there a downside to using the `__attribute__` syntax unless `__STDC_VERSION__ >= 202311L`?

jeremylt: Not that I know? Hopefully this isn't something that we'll need much after the 1.0 release

Issue #1707, opened 2024-10-24
The Python binding rebuilds and reinstalls libceed.so and the headers which is not necessary


yurivict:
As a binding, it should use the pre-installed libceed.so

Could you please consider fixing the Python build such that it would only build parts necessary for the binding and not the whole thing?

All redundant files:
```
%%PYTHON_SITELIBDIR%%/libceed/include/ceed.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/backend.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/ceed-f32.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/ceed-f64.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/ceed.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/cuda.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/fortran.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/hip.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-atomic-add-fallback.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-gen-templates.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-jit.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-ref-basis-nontensor-templates.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-ref-basis-nontensor.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-ref-basis-tensor-at-points.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-ref-basis-tensor.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-ref-operator-assemble-diagonal.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-ref-operator-assemble.h
%%PYTHON_SITELIBDIR%%/libceed/include/ceed/jit-source/cuda/cuda-ref-qfunction.h
[...]

PR #1879, opened 2025-07-30
python - update pyproject.toml and setup.py to use declarative packaging

zatkins-dev:
Move most static data over to `pyproject.toml` and simplify `setup.py`.

Other things:
- We can list multiple authors/maintainers, do we want to?
- Need to address #1707


zatkins-dev:
I think the underlying issues I wanted to solve aren't really possible through pip installations. In order to actually support portable python installation, we should probably instead look at making a `conda` package (ugh).

PR #1873, opened 2025-07-21
Draft: Rust and cuda clang support

SirAlienTheGreat:
Todo before merging:
* update examples to use 2d array notation

PR #1881, opened 2025-07-31
Rust and cuda clang support (#1873)

jeremylt:
This is a staging branch for #1873 so we can run Noether CI and fix any issue there before doing final approval and merging


jeremylt:
Ok, the blocking issue here seems to be around permissions to create files/folders in the runner

PR #1883, opened 2025-08-04
doc - list HONEE in libCEED backends svg

jeremylt:
Swapping so we list both Ratel and HONEE in our docs


doc/img/libCEEDBackends.svg
jedbrown, line 1: Can we include the tex source for this? Also, it'd be a fine time to nudge the left column up for better vertical symmetry.

jeremylt: How about the latest?

jrwrigh: The latest commit added the source, but didn't change the vertical spacing?

jeremylt: oops, how about now?

jrwrigh: Looks like the gap between MFEM and HONEE is larger than the other two gaps?

jrwrigh: Yeah, the difference in the y coordinate location is 1.4 for MFEM and HONEE, but 1.2 for the others.

doc/img/libCEEDBackends.tex
```diff
@@ -0,0 +1,187 @@
...
+\begin{tikzpicture}
+
+\begin{scope}[shift={(0,-0.6)}]
+  \node at (1.0,6.1) {\large Application};
+
+  \draw[
+    top color=ceed@red!10!white,
+    bottom color=ceed@red!40!white,
+    ceed@red!60!black,
+  ] (0.0,-0.8) rectangle ++(1.6,0.8)
```
jrwrigh:
```suggestion
  ] (0.0,-0.6) rectangle ++(1.6,0.8)
```

I think this will do it.

doc/img/libCEEDBackends.tex
```diff
@@ -0,0 +1,187 @@
...
+  node[pos=.5,align=center,color=black] {MFEM};
+  \draw[-stealth, line width=0.5pt] (1.6, -0.8+0.4) -- ++(1.6,2.4-0.6);
+
+  % PETSc
```
jrwrigh: Also need to update the comments to match the block

doc/img/libCEEDBackends.svg
jeremylt: ok, looking better now

jrwrigh: Yep, that all checks out for me!

PR #1886, opened 2025-08-13
ref - Ensure at-points operators skip cells with no points

zatkins-dev:
Previous behavior resulted in garbage values being placed into the output vector.

### DIFF
--- a/backends/ref/ceed-ref-operator.c
+++ b/backends/ref/ceed-ref-operator.c
@@ -1054,6 +1054,7 @@ static int CeedOperatorApplyAddAtPoints_Ref(CeedOperator op, CeedVector in_vec, 
     // Setup points for element
     CeedCallBackend(CeedElemRestrictionApplyAtPointsInElement(rstr_points, e, CEED_NOTRANSPOSE, point_coords, impl->point_coords_elem, request));
     CeedCallBackend(CeedElemRestrictionGetNumPointsInElement(rstr_points, e, &num_points));
+    if (num_points <= 0) continue;
 
     // Input basis apply
     CeedCallBackend(CeedOperatorInputBasisAtPoints_Ref(e, num_points_offset, num_points, qf_input_fields, op_input_fields, num_input_fields, in_vec,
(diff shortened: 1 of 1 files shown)


backends/ref/ceed-ref-operator.c
```diff
@@ -1054,6 +1054,7 @@ static int CeedOperatorApplyAddAtPoints_Ref(CeedOperator op, CeedVector in_vec,
     // Setup points for element
     CeedCallBackend(CeedElemRestrictionApplyAtPointsInElement(rstr_points, e, CEED_NOTRANSPOSE, point_coords, impl->point_coords_elem, request));
     CeedCallBackend(CeedElemRestrictionGetNumPointsInElement(rstr_points, e, &num_points));
+    if (num_points <= 0) continue;
```
jeremylt: Can `num_points < 0` be true? It shouldn't be possible?

I guess if you want the conditional just in case, then I'd say
```suggestion
    if (num_points < 1) continue;
```

zatkins-dev: It definitely shouldn't be possible, but just in case

backends/ref/ceed-ref-operator.c
```diff
@@ -1054,6 +1054,7 @@ static int CeedOperatorApplyAddAtPoints_Ref(CeedOperator op, CeedVector in_vec,
     // Setup points for element
     CeedCallBackend(CeedElemRestrictionApplyAtPointsInElement(rstr_points, e, CEED_NOTRANSPOSE, point_coords, impl->point_coords_elem, request));
     CeedCallBackend(CeedElemRestrictionGetNumPointsInElement(rstr_points, e, &num_points));
+    if (num_points < 1) continue;
```
jeremylt, line 1057: oooh, you'll probably need to do cleanup here and destroy `rstr_points` and `point_coords`

jeremylt: wait, no, I'm wrong

zatkins-dev: I don't think so? That happens outside the loop

PR #1889, opened 2025-08-19
assembly - add `CeedOperatorLinearAssembleGetNumEntries`

zatkins-dev:
Adds an external interface for getting the number of entries in the COO data for an operator.

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -957,6 +957,42 @@ static int CeedOperatorAssemblyCountEntriesSingle(CeedOperator op, CeedSize *num
 }
 
 /**
+  @brief Count number of entries for assembled `CeedOperator`
+
+  @param[in]  op          `CeedOperator` to assemble
+  @param[out] num_entries Number of entries in assembled representation
+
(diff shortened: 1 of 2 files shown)


interface/ceed-preconditioning.c
```diff
@@ -956,6 +956,40 @@ static int CeedOperatorAssemblyCountEntriesSingle(CeedOperator op, CeedSize *num
...
+**/
+int CeedOperatorLinearAssembleGetNumEntries(CeedOperator op, CeedSize *num_entries) {
+  bool          is_composite;
+  CeedInt       num_suboperators;
+  CeedOperator *sub_operators;
+
+  CeedCall(CeedOperatorCheckReady(op));
+  CeedCall(CeedOperatorIsComposite(op, &is_composite));
+
+  *num_entries = 0;
```
jrwrigh: _Technically_ only the composite branch needs to be zeroed. This is being really pedantic, but it might help the readability an bit to have it just before the loop over suboperators.

zatkins-dev: fair point, updated!

PR #1890, opened 2025-08-30
at-points - ensure reference point coordinates are always up to date

zatkins-dev:
Minor change to make the behavior of the `cuda/ref` and `hip/ref` operators the same as gen and `/cpu/self/ref` by always applying the reference coordinate element restriction during calls to `OperatorApply`. This allows for in-place updating of reference coordinates without having to change any interfaces.

### DIFF
--- a/backends/cuda-ref/ceed-cuda-ref-operator.c
+++ b/backends/cuda-ref/ceed-cuda-ref-operator.c
@@ -848,12 +848,12 @@ static int CeedOperatorApplyAddAtPoints_Cuda(CeedOperator op, CeedVector in_vec,
   CeedCallBackend(CeedGetWorkVector(ceed, impl->max_active_e_vec_len, &active_e_vec));
 
   // Get point coordinates
-  if (!impl->point_coords_elem) {
+  {
     CeedVector          point_coords = NULL;
     CeedElemRestriction rstr_points  = NULL;
 
     CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, &point_coords));
-    CeedCallBackend(CeedElemRestrictionCreateVector(rstr_points, NULL, &impl->point_coords_elem));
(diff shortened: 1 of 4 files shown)


backends/cuda-ref/ceed-cuda-ref-operator.c
```diff
@@ -848,12 +848,12 @@ static int CeedOperatorApplyAddAtPoints_Cuda(CeedOperator op, CeedVector in_vec,
...
+  {
     CeedVector          point_coords = NULL;
     CeedElemRestriction rstr_points  = NULL;
 
     CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, &point_coords));
-    CeedCallBackend(CeedElemRestrictionCreateVector(rstr_points, NULL, &impl->point_coords_elem));
+    if (!impl->point_coords_elem) CeedCallBackend(CeedElemRestrictionCreateVector(rstr_points, NULL, &impl->point_coords_elem));
     CeedCallBackend(CeedElemRestrictionApply(rstr_points, CEED_NOTRANSPOSE, point_coords, impl->point_coords_elem, request));
     CeedCallBackend(CeedVectorDestroy(&point_coords));
     CeedCallBackend(CeedElemRestrictionDestroy(&rstr_points));
```
jeremylt:
```suggestion
    if (!impl->point_coords_elem) CeedCallBackend(CeedElemRestrictionCreateVector(rstr_points, NULL, &impl->point_coords_elem));
    {
      uint64_t state;

      CeedCallBackend(CeedVectorGetState(point_coords, &state));
      if (impl->points_state != state) {
        CeedCallBackend(CeedElemRestrictionApply(rstr_points, CEED_NOTRANSPOSE, point_coords, impl->point_coords_elem, request));
      }
    }
    CeedCallBackend(CeedVectorDestroy(&point_coords));
    CeedCallBackend(CeedElemRestrictionDestroy(&rstr_points));
```
This will require an extra field in `CeedOperator_Cuda`, but this will prevent extra calls.

zatkins-dev:
HIP will still need your patch

PR #1904, opened 2025-11-07
view - add Ceed*[Get/Set]ViewTabs for all objects

### LINKED ISSUE #1902: Add `Ceed*SetNumViewTabs` for all libCEED objects
Add functions similar to `CeedOperatorSetNumViewTabs`, but for other objects.

jeremylt:
Closes #1902 

This makes the output inline with other logging for Ratel


jeremylt:
`TypeError: pathlib.Path | None is not callable` is at the heart of the issue with MacOS

PR #1908, opened 2025-11-20
minor - move Ceed*GetNumViewTabs to public header

zatkins-dev:


jeremylt (commented):
Minor nit - can you move the corresponding function definitions too? I try to keep the same function ordering in the .h and .c files for quick finding

jeremylt (approved):

Issue #1905, opened 2025-11-17
Question on padding for qfunctions with blocked backends


hughcars:
Hello libceed developers, 

Thank you for the library, it's very impressive and extremely enabling for [Palace](https://github.com/awslabs/palace), we're using it for almost all of our linear algebra at this stage. Given that, we've been investigating some data structure optimizations and have some questions about how the padded backends work:

1) Where is the padding for a blocked backend defined, and how does the "padded data" get computed from the valid data, whilst not being accumulated/used for calculations in the output vector? 
2) How is this padding affected by the choice of stride pattern a user provides? Are users allowed to specify strides that aren't `CEED_STRIDES_BACKEND` in `CeedElemRestrictionCreateStrided` and work with blocked backends?
3) Is changing the data storage pattern for a passive vector from all of i-th component then all of (i+1)-th etc. to all components for quad point j, then all for quad j+1, a bad idea for a non-cpu backend?

Long version: 
We have QFunctions that use a passive vector where have stored the geometric factor data, computed in advance, along with two additional pieces of scalar data. We do this as there are many kernels that reuse the geometric factor data and have traded storage for time. This data is stored as a QVector, given there's `nelem * nquad_per_elem * n_component (=11)` pieces of data. An issue we've noticed through some cpu profiling however is that for the `CEED_STRIDES_BACKEND` the 11 components of the data are being stored with a stride of `Q` in the data vector, which for a large enough `Q` (88 for `/cpu/self/ref/blocked` for instance), is causing a cache miss for each value of the data drawn (this was all found using vtune, so we're moderately convinced it's a real effect). Namely all the 0-th component terms come first, then all the 1-th component, then all the 2-th etc, meaning we get ~10 L3 cache misses per quadrature point (resulting in a lot of DRAM calls).

Given this, we wanted to try changing the data storage to be coalesced by components, I'm not sure of the terminology for this within libCEED, but basically have all 11 components of data that are needed adjacent in the data vector, rather than all of the first components, then all of the second components etc. (thinking about the data vector as a matrix `Q x 11`, this amounts to row major vs column major storage). Thus a load for the first data component would almost certainly pull all the data for the given quadrature point into the cache in one go, and we'd no longer need a DRAM access (which is what has been observed with vtune).

Changing the `strides` from `CEED_STRIDES_BACKEND` to 
```
  CeedInt strides[3] = {
    geom_data_size,            // Stride between nodes
    1,                         // Stride between components
[...]

PR #1910, opened 2025-12-12
Use compiler definition for Intel oneAPI

Sbozzolo:
`__INTEL_COMPILER` no longer exists in LLVM-based Intel compilers. The new macro is `__INTEL_LLVM_COMPILER`.

https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-2/use-predefined-macros-to-specify-intel-compilers.html


laylagi:
Thanks @Sbozzolo for the catch. This fixes the warnings I reported [here](https://github.com/CEED/libCEED/issues/1905#issuecomment-3573747159).

jeremylt (approved):
Thanks for this fix!

PR #1913, opened 2026-01-07
Experimenting with Weak Workaround

jeremylt:


backends/ceed-backend-weak.c
```diff
@@ -37,8 +37,8 @@ static int CeedRegister_Weak(const char *name, int num_prefixes, ...) {
 }
 // LCOV_EXCL_STOP
 
-#define CEED_BACKEND(name, num_prefixes, ...)       \
-  CEED_INTERN int name(void) __attribute__((weak)); \
-  int             name(void) { return CeedRegister_Weak(__func__, num_prefixes, __VA_ARGS__); }
+#define CEED_BACKEND(name, num_prefixes, ...)                 \
+  CEED_INTERN int           name(void) __attribute__((weak)); \
+  int __attribute__((weak)) name(void) { return CeedRegister_Weak(#name, num_prefixes, __VA_ARGS__); }
```
jeremylt, line 42: Listing both the definition and declaration as week is supposed to help. It does not.

It is also supposed to help to make both the weak and strong symbol extern (I'd rather not) but that did not help either.

jeremylt: Also, trying libtool failed as well

... MacOS seems to really not want you to make static libraries

PR #1917, opened 2026-01-14
Add core CeedObject to handle common functionality

jeremylt:
@zatkins-dev this should simplify some of the logging logic, as we won't have to have different view logic for each Ceed* 

We can now:
```C
CeedObjectSetNumViewTabs((CeedObj)foo, 42);
CeedObjectView((CeedObj)foo, stdout);
CeedObjectDestroy((CeedObject *)&foo);
```

~~I'll be doing other stuff for the next few days, so feel free to tinker with this if you like. No worries if not - I'll finish it next week sometime~~

### DIFF
--- a/interface/ceed-object.c
+++ b/interface/ceed-object.c
@@ -19,7 +19,11 @@
 /// @{
 
 /**
-  @brief Create a `CeedObject`
+  @brief Create a `CeedObject`.
+
+  Note: This interface takes a `CeedObject` and not a pointer to a `CeedObject` like other `Ceed*Create` interfaces.
+          This `CeedObject` will have already been allocated a the first part of the `Ceed*` struct.
+          This function is only intended to be called inside of `Ceed*Create` functions.
 
   @param[in]  ceed             `Ceed` object to reference
   @param[in]  view_function    `Ceed*` function for viewing the `obj`


zatkins-dev:
> Thoughts - Do we also want CeedObjectDestroy ? (Easy to add) And do we want the view tabs to be added to the CeedObject? I'm thinking yes for both

I would say yes for both

interface/ceed-object.c
```diff
@@ -0,0 +1,132 @@
...
+
+  @ref Backend
+**/
+int CeedObjectCreate(Ceed ceed, int (*view_function)(CeedObject, FILE *), CeedObject obj) {
+  obj->ceed = NULL;
+  if (ceed) CeedCall(CeedReferenceCopy(ceed, &obj->ceed));
+  obj->ViewFunction = view_function;
+  obj->ref_count    = 1;
+  return CEED_ERROR_SUCCESS;
+}
```
zatkins-dev: Note: I understand why, but this is technically a different interface (taking a `CeedObject` instead of a `CeedObject *`) than other `*Create` methods in libCEED. Do we want to add a note that this function that `obj` is the address of a `CeedObject_private` (more accurate than the current wording)?

jrwrigh:
I think another addition here might be to add a `name` field to `CeedObject` though I think that can be in a follow-on PR.

interface/ceed-object.c
jeremylt: I'm hesitant to mention the private struct as that's something the users should not know about

PR #1919, opened 2026-01-22
Fix Weak Symbols on MacOS

jeremylt:
Fixes static library builds on MacOS.

The trick is to beak the weak definitions into separate files. Once the weak definition object file is load once, that file beats object files loaded later, so we need to split based upon the units that are activated/deactivated separately.

### DIFF
--- a/backends/ceed-backend-list-avx.h
+++ b/backends/ceed-backend-list-avx.h
@@ -1,0 +1,13 @@
+// Copyright (c) 2017-2026, Lawrence Livermore National Security, LLC and other CEED contributors.
+// All Rights Reserved. See the top-level LICENSE and NOTICE files for details.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+//
+// This file is part of CEED:  http://github.com/ceed
+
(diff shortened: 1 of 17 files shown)


backends/ceed-backend-weak.c
```diff
@@ -37,8 +37,7 @@ static int CeedRegister_Weak(const char *name, int num_prefixes, ...) {
 }
 // LCOV_EXCL_STOP
 
-#define CEED_BACKEND(name, num_prefixes, ...)       \
-  CEED_INTERN int name(void) __attribute__((weak)); \
-  int             name(void) { return CeedRegister_Weak(__func__, num_prefixes, __VA_ARGS__); }
+#define CEED_BACKEND(name, num_prefixes, ...) \
+  int __attribute__((weak)) name(void) { return CeedRegister_Weak(__func__, num_prefixes, __VA_ARGS__); }
```
jeremylt: it really feels like this weak attribute is supposed to go on the definition and not the declaration anyways, even if we can't get this working on MacOS

PR #1921, opened 2026-01-29
ex - add fortran ex1

jeremylt:
Adding a Fortran ex1, hopefully this makes it easier for future students to add ex2 and ex3


jeremylt:
unfortunately, coverage for fortran files will always be wrong, as fortran compilers strip all comments out by default (so strips `! LCOV_EXCL_START`) and `-cpp -C` does not in fact enable the C preprocessor and force the compiler to retain C style comments, as the gfortran documentation says it would

PR #1924, opened 2026-02-04
Fix gcc `-march=native` for arm mac

hughcars:
When building with spack on ARM (Apple Silicon), GCC's `-march=native` fails:

```
  cc1: error: unknown value 'apple-m1' for '-march'
  cc1: note: valid arguments are: armv8-a armv8.1-a ... native
  cc1: note: did you mean '-mcpu=apple-m1'?
```

This occurs because spack's compiler wrapper causes `-march=native` to resolve to CPU names like `apple-m1`, but GCC's `-march` doesn't accept CPU names on ARM—only `-mcpu` does. Fix by detecting architecture via `uname -m` and using `-mcpu=native` on non-x86 platforms.

Reproducer with spack on macOS ARM:

```bash
mkdir test && cd test
echo 'spack:
  specs:
  - libceed@develop %gcc' > spack.yaml
spack env activate .
spack install
```

Also changes the macos CI to use gcc/clang from homebrew, as `gcc` `clang` are aliases for apple-clang.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,8 +112,12 @@ CC_VENDOR := $(subst (GCC),gcc,$(subst icc_orig,icc,$(CC_VENDOR)))
 CC_VENDOR := $(if $(filter cc,$(CC_VENDOR)),gcc,$(CC_VENDOR))
 FC_VENDOR := $(if $(FC),$(firstword $(filter GNU ifort ifx XL,$(shell $(FC) --version 2>&1 || $(FC) -qversion))))
 
+# Host architecture for setting appropriate flags
+UNAME_M := $(shell uname -m)
+
 # Default extra flags by vendor
-MARCHFLAG.gcc           := -march=native
+# GCC: use -march=native only on x86 (where -mcpu doesn't exist); use -mcpu=native elsewhere
+MARCHFLAG.gcc           := $(if $(filter x86_64 i%86,$(UNAME_M)),-march=native,-mcpu=native)
 MARCHFLAG.clang         := $(MARCHFLAG.gcc)
 MARCHFLAG.icc           :=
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -112,9 +112,13 @@ CC_VENDOR := $(subst (GCC),gcc,$(subst icc_orig,icc,$(CC_VENDOR)))
...
 
+# Host architecture for setting appropriate flags
+UNAME_M := $(shell uname -m)
+
 # Default extra flags by vendor
-MARCHFLAG.gcc           := -march=native
-MARCHFLAG.clang         := $(MARCHFLAG.gcc)
+# GCC: use -march=native only on x86 (where -mcpu doesn't exist); use -mcpu=native elsewhere
+MARCHFLAG.gcc           := $(if $(filter x86_64 i%86,$(UNAME_M)),-march=native,-mcpu=native)
+MARCHFLAG.clang         := -march=native
```
jedbrown:
```suggestion
MARCHFLAG.clang         := $(MARCHFLAG.gcc)
```
I believe clang works the same way as gcc in this regard so I think this line should be reverted.
```console
$ clang -E --target=aarch64 -march=apple-m1 -x c /dev/null
clang: error: unsupported argument 'apple-m1' to option '-march='
```

In contrast, this is successful:
```console
$ clang -E --target=aarch64 -mcpu=apple-m1 -x c /dev/null
```

hughcars: I'd wanted to try and figure out the details here before engaging either of you (the PR description isn't even right any more), but thank you for taking an early look. I think you're right, and had undone it, the `UNAME_M` approach here is I think better than the `-c` vs `-E` I originally tried (but broke the powerpc CI).

jeremylt (approved):

PR #1927, opened 2026-02-06
Scalar Multigrid Multiplicity

### LINKED ISSUE #1307: Scalar Multiplicity Vector
Yeah, it is. But the important part is that is results in 1 scalar per node rather than 1 scalar per component per node (which is what `CeedElemRestrictionGetMultiplicity` gives out). So if there's an easy *Edit: and robust* way of getting out the components that correspond to a single component, then we could just do that and invert that vector.

_Originally posted by @jrwrigh in https://github.com/CEED/libCEED/pull/1223#discussion_r1304839240_
            

Local node multiplicity is the same across all components - we could save memory movement by letting a ElemRestriction provide a scalar multiplicity vector.

jeremylt:
Right now we're dragging around a `num_comp` duplication of the multiplicity vector for multigrid. This uses a scalar multiplicity instead. This should save a fair amount of memory, but the PR needs to be cleaned up to do all of this on the target memory space

We save `(p-1)^3 * (num_comp - 1)` values per element (roughly) in exchange for adding `3 p^2 - 3 p + 1` values, which is generally a good trade. I've included a check to only use this new approach when it is favorable in terms of storage needed. (most relevant for gen backends, as CPU will keep both the lvec and evec unless its a strided restriction with backend strides)

Fixes #1307

### DIFF
--- a/interface/ceed-preconditioning.c
+++ b/interface/ceed-preconditioning.c
@@ -1013,7 +1013,7 @@ static int CeedOperatorMultigridLevelCreateSingle_Core(CeedOperator op_fine, Cee
                                                        CeedOperator *op_prolong, CeedOperator *op_restrict) {
   bool                is_composite;
   Ceed                ceed;
-  CeedInt             num_comp, num_input_fields, num_output_fields;
+  CeedInt             dim              = 0, num_comp, num_input_fields, num_output_fields;
   CeedVector          mult_vec         = NULL;
   CeedElemRestriction rstr_p_mult_fine = NULL, rstr_fine = NULL;
   CeedOperatorField  *input_fields, *output_fields;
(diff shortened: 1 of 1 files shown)


gallery/scale/ceed-scale-scalar.c
```diff
@@ -0,0 +1,29 @@
...
+#include <ceed/backend.h>
+#include <ceed/jit-source/gallery/ceed-scale-scalar.h>
+#include <string.h>
+
+/**
+  @brief  Set fields for vector scaling `CeedQFunction` that scales inputs
+**/
+static int CeedQFunctionInit_ScaleScalar(Ceed ceed, const char *requested, CeedQFunction qf) {
+  // Check QFunction name
+  const char *name = "Scalar Scale";
```
jeremylt: Not my favorite name

interface/ceed-preconditioning.c
```diff
@@ -1097,21 +1097,48 @@ static int CeedOperatorMultigridLevelCreateSingle_Core(CeedOperator op_fine, Cee
...
+      CeedCall(CeedElemRestrictionGetNumElements(rstr_p_mult_fine, &num_elem));
+      CeedCall(CeedElemRestrictionGetElementSize(rstr_p_mult_fine, &elem_size));
+      CeedCall(CeedVectorCreate(ceed, num_elem * elem_size, &mult_vec));
+      CeedCall(CeedVectorReciprocal(mult_e_vec));
+      CeedCall(CeedElemRestrictionGetELayout(rstr_p_mult_fine, layout));
+      CeedCall(CeedVectorGetArrayWrite(mult_vec, CEED_MEM_HOST, &mult_array));
+      CeedCall(CeedVectorGetArrayRead(mult_e_vec, CEED_MEM_HOST, &mult_e_array));
+      for (CeedInt i = 0; i < num_elem; i++) {
+        for (CeedInt j = 0; j < elem_size; j++) mult_array[j + i * elem_size] = mult_e_array[j * layout[0] + i * layout[2]];
+      }
```
jeremylt: This is the part that needs to be on device

PR #1930, opened 2026-02-18
Add CeedOperatorCompositeSetSequential and fix edge cases in CUDA gen at points assembly

zatkins-dev:
These are changes needed to support functionality in Ratel (https://gitlab.com/micromorph/ratel/-/merge_requests/1213).

First, I need to be able to turn off the parallel sub-operator functionality. 

The particular use case is that I need to project a quantity from different mesh elements to the same set of points. The points here are "shared" between the two sides - they have the same quadrature weights and areas, etc. I then need to use both of those at-points vectors as inputs to the "real" QFunctions which use each of the interpolated values to compute the physics at quadrature points. These "real" QFunctions then interpolate back to each of the two sets of mesh elements. That's a use-case that's fundamentally incompatible with the way that our at points operators work, which group points by which element they are inside of. We strictly need 2 "setup" operators (one for each set of mesh elements) that just interp to points, then 2 "output" operators (again, one for each set of mesh elements) to compute the physics and interp-transpose back to the mesh nodes. There is never a situation where I would only run the "setup" or "output" -- they are explicitly coupled. But, the math is perfectly well-defined if that sequence is allowed. 

Second, the linear assemble diagonal (and maybe also the normal assembly) were trying to write to passive outputs (particularly with CeedEvalNone), which is incorrect. This fixes that by adding an additional flag to the QFunction builder function.

Specifically, previously the point loop would have something like this:
```c
        // EvalMode: none
        const CeedInt comp_stride_out_1 = 1;
        WritePoint<num_comp_out_1, comp_stride_out_1, max_num_points>(data, elem, i, points.num_per_elem[elem], indices.outputs[1], r_s_out_1, d_out_1);

```
Here, `d_out_1` is a passive (CeedEvalNone`) output. We should instead just not include inactive outputs in the QFunction build routine.

### DIFF
--- a/tests/t599-operator.c
+++ b/tests/t599-operator.c
@@ -1,0 +1,146 @@
+/// @file
+/// Test creation, action, and destruction for mass matrix operator at points using sequential composite operator
+/// \test Test creation, action, and destruction for mass matrix operator at points using sequential composite operator
+#include "t599-operator.h"
+
+#include <ceed.h>
+#include <math.h>
+#include <stdio.h>
+
+int main(int argc, char **argv) {
+  Ceed    ceed;
+  CeedInt num_elem_1d = 3, num_elem = num_elem_1d * num_elem_1d, dim = 2, p = 3, q = 5;
+  CeedInt num_nodes = (num_elem_1d * (p - 1) + 1) * (num_elem_1d * (p - 1) + 1), num_points_per_elem = 4, num_points = num_elem * num_points_per_elem;
+  CeedVector          x_points, u, v, u_points;
+  CeedElemRestriction elem_restriction_x_points, elem_restriction_u_points, elem_restriction_u;
+  CeedBasis           basis_u;
+  CeedQFunction       qf_to_points, qf_from_points;
+  CeedOperator        op_to_points, op_from_points, op_mass;
+  bool                is_at_points, is_sequential;
+
+  CeedInit(argv[1], &ceed);
+
+  CeedVectorCreate(ceed, dim * num_points, &x_points);
+  {
+    CeedScalar x_array[dim * num_points];
+
+    for (CeedInt e = 0; e < num_elem; e++) {
+      for (CeedInt d = 0; d < dim; d++) {
+        x_array[num_points_per_elem * (e * dim + d) + 0] = 0.25;
+        x_array[num_points_per_elem * (e * dim + d) + 1] = d == 0 ? -0.25 : 0.25;
+        x_array[num_points_per_elem * (e * dim + d) + 2] = d == 0 ? 0.25 : -0.25;
+        x_array[num_points_per_elem * (e * dim + d) + 3] = 0.25;
+      }
+    }
+    CeedVectorSetArray(x_points, CEED_MEM_HOST, CEED_COPY_VALUES, x_array);
+  }
+  {
+    CeedInt ind_x[num_elem + 1 + num_points];
+
+    for (CeedInt i = 0; i <= num_elem; i++) ind_x[i] = num_elem + 1 + i * num_points_per_elem;
+    for (CeedInt i = 0; i < num_points; i++) ind_x[num_elem + 1 + i] = i;
+    CeedElemRestrictionCreateAtPoints(ceed, num_elem, num_points, dim, num_points * dim, CEED_MEM_HOST, CEED_COPY_VALUES, ind_x,
+                                      &elem_restriction_x_points);
+    CeedElemRestrictionCreateAtPoints(ceed, num_elem, num_points, 1, num_points, CEED_MEM_HOST, CEED_COPY_VALUES, ind_x, &elem_restriction_u_points);
+    CeedElemRestrictionCreateVector(elem_restriction_u_points, &u_points, NULL);
+    CeedVectorSetValue(u_points, 0);
+  }
+
+  {
+    CeedInt ind_u[num_elem * p * p];
+
+    for (CeedInt e = 0; e < num_elem; e++) {
+      CeedInt elem_xy[2] = {1, 1}, n_d[2] = {0, 0};
+
+      for (CeedInt d = 0; d < dim; d++) n_d[d] = num_elem_1d * (p - 1) + 1;
+      {
+        CeedInt r_e = e;
+
+        for (CeedInt d = 0; d < dim; d++) {
+          elem_xy[d] = r_e % num_elem_1d;
+          r_e /= num_elem_1d;
+        }
+      }
+      CeedInt num_nodes_in_elem = p * p, *elem_nodes = ind_u + e * num_nodes_in_elem;
+
+      for (CeedInt n = 0; n < num_nodes_in_elem; n++) {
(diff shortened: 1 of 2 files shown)


backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -304,18 +304,36 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
   CeedCallBackend(CeedOperatorGetCeed(op, &ceed));
   CeedCall(CeedOperatorCompositeGetNumSub(op, &num_suboperators));
   CeedCall(CeedOperatorCompositeGetSubList(op, &sub_operators));
+  CeedCall(CeedOperatorCompositeIsSequential(op, &is_sequential));
   if (input_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArrayRead(input_vec, CEED_MEM_DEVICE, &input_arr));
   if (output_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArray(output_vec, CEED_MEM_DEVICE, &output_arr));
-  for (CeedInt i = 0; i < num_suboperators; i++) {
-    CeedInt num_elem = 0;
+  if (is_sequential) {
```
jeremylt: Is this maybe clearer to put the if inside of the loop?

interface/ceed-operator.c
```diff
@@ -1390,6 +1390,48 @@ int CeedOperatorCompositeGetSubByName(CeedOperator op, const char *op_name, Ceed
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Set whether the sub-operators of the composite `CeedOperator` must be run sequentially.
+
+  Note: This value currently only affects the GPU `/gpu/cuda/gen` and `/gpu/hip/gen` backends
+
+  @param[in] op            Composite `CeedOperator`
+  @param[in] is_sequential Flag value to set, if true, forces the composite `CeedOperator` to execute sequentially
```
jeremylt:
```suggestion
  @param[in] is_sequential Flag value to set, if `true`, forces the composite `CeedOperator` to execute sequentially
```

backends/cuda-gen/ceed-cuda-gen-operator.c
zatkins-dev: I avoided that because the sequential needs to declare the stream outside of the loop

jeremylt: sure, so it would be multiple if statements. The amount of repeated code is just a bit uncomfortable

zatkins-dev:
> Codecov is correct - this should be tested, if we can cook a representative reason why we would want this

Yeah, I didn't add tests yet. I think that splitting the input and output parts of the operator like described above (but with only one mesh element set) would be a fine test.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -304,18 +304,36 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
...
+      CeedCall(CeedOperatorGetNumElements(sub_operators[i], &num_elem));
+      if (num_elem > 0) {
+        cudaStream_t stream = NULL;
+
+        CeedCallCuda(ceed, cudaStreamCreate(&stream));
+        CeedCallBackend(CeedOperatorApplyAddCore_Cuda_gen(sub_operators[i], stream, input_arr, output_arr, &is_run_good[i], request));
+        CeedCallCuda(ceed, cudaStreamDestroy(stream));
+      }
     }
   }
```
zatkins-dev:
```suggestion
  if (is_sequential) CeedCallCuda(ceed, cudaStreamCreate(&stream));
  for (CeedInt i = 0; i < num_suboperators; i++) {
    CeedInt num_elem = 0;

    CeedCall(CeedOperatorGetNumElements(sub_operators[i], &num_elem));
    if (num_elem > 0) {
      if (!is_sequential) CeedCallCuda(ceed, cudaStreamCreate(&stream));
      CeedCallBackend(CeedOperatorApplyAddCore_Cuda_gen(sub_operators[i], stream, input_arr, output_arr, &is_run_good[i], request));
      if (!is_sequential) CeedCallCuda(ceed, cudaStreamDestroy(stream));
    }
  }
  if (is_sequential) CeedCallCuda(ceed, cudaStreamDestroy(stream));
```

@jeremylt is this what you were thinking?

zatkins-dev: it's definitely shorter, though I don't think the control flow is more clear

jeremylt: Yeah. I would hope that the conditional evaluation is cheap compared to the kernel execution time so we win on overall reduced repetition and (relative) simplicity at a quick glance

zatkins-dev: I mean I don't think there's any perf difference, the compiler is going to figure out what we're doing

jeremylt: At the end of the day - I'm not too fussed either way so I'm happy to defer to your preference.

Don't forget the same changes in the HIP backend

zatkins-dev: hip ends up looking a bit different FYI

interface/ceed-operator.c
```diff
@@ -1390,6 +1390,48 @@ int CeedOperatorCompositeGetSubByName(CeedOperator op, const char *op_name, Ceed
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Set whether the sub-operators of the composite `CeedOperator` must be run sequentially.
+
+  Note: This value currently only affects the GPU `/gpu/cuda/gen` and `/gpu/hip/gen` backends
```
jeremylt:
```suggestion
  Note: This value currently only affects the GPU `/gpu/cuda/gen` and `/gpu/hip/gen` backends.
```

interface/ceed-operator.c
```diff
@@ -1390,6 +1390,48 @@ int CeedOperatorCompositeGetSubByName(CeedOperator op, const char *op_name, Ceed
...
+  CeedCall(CeedOperatorIsComposite(op, &is_composite));
+  CeedCheck(is_composite, CeedOperatorReturnCeed(op), CEED_ERROR_MINOR, "Only defined for a composite operator");
+  op->is_sequential = is_sequential;
+  return CEED_ERROR_SUCCESS;
+}
+
+/**
+  @brief Get whether the sub-operators of the composite `CeedOperator` must be run sequentially.
+
+  Note: This value currently only affects the GPU `/gpu/cuda/gen` and `/gpu/hip/gen` backends
```
jeremylt:
```suggestion
  Note: This value currently only affects the GPU `/gpu/cuda/gen` and `/gpu/hip/gen` backends.
```

jeremylt (approved):

PR #1935, opened 2026-03-06
interface: fix uninitialized use in CeedOperatorMultigridLevelCreateSingle_Core

zatkins-dev:
@jeremylt this seems to have slipped through when you made this change. Do we not run with `-Werror` in CI?


interface/ceed-preconditioning.c
jeremylt: Really, it should be for `p_mult_full`, so we need to raise this line to where you made your changes

jeremylt (approved):

PR #1942, opened 2026-03-17
Add chipStar (SPIR-V) support for HIP backends

pvelesko:
Purpose:

Add support for [chipStar](https://github.com/CHIP-SPV/chipStar), enabling libCEED HIP backends (`/gpu/hip/ref`, `/gpu/hip/shared`, `/gpu/hip/gen`) to run on any OpenCL or Level Zero GPU (Intel, etc.) via SPIR-V.

LLM/GenAI Disclosure: Claude Code Opus 4.6 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,8 @@ endif
 ifeq (,$(filter-out undefined default,$(origin ARFLAGS)))
   ARFLAGS = $(if $(DARWIN),cr,crD)
 endif
+# chipStar sets HIP_DIR instead of ROCM_DIR
+ROCM_DIR ?= ${HIP_DIR}
 NVCC ?= $(CUDA_DIR)/bin/nvcc
 NVCC_CXX ?= $(CXX)
 HIPCC ?= $(ROCM_DIR)/bin/hipcc
@@ -88,6 +90,15 @@ ifneq ($(EMSCRIPTEN),)
   EM_LDFLAGS = -s TOTAL_MEMORY=256MB
 endif
 
+# Detect chipStar (SPIR-V HIP) vs AMD ROCm
+HIP_LIB_NAME = amdhip64
+ifneq ($(ROCM_DIR),)
+  ifneq ($(wildcard $(ROCM_DIR)/bin/hipconfig),)
+    ifneq (,$(findstring __HIP_PLATFORM_SPIRV__,$(shell $(ROCM_DIR)/bin/hipconfig 2>/dev/null)))
+      HIP_LIB_NAME = CHIP
+    endif
+  endif
+endif
 # ASAN must be left empty if you don't want to use it
 ASAN ?=
 
@@ -555,16 +566,23 @@ ifneq ($(CUDA_LIB_DIR),)
 endif
 
 # HIP Backends
-HIP_LIB_DIR := $(wildcard $(foreach d,lib lib64,$(ROCM_DIR)/$d/libamdhip64.${SO_EXT}))
+HIP_LIB_DIR := $(wildcard $(foreach d,lib lib64,$(ROCM_DIR)/$d/lib${HIP_LIB_NAME}.${SO_EXT}))
 HIP_LIB_DIR := $(patsubst %/,%,$(dir $(firstword $(HIP_LIB_DIR))))
 HIP_BACKENDS = /gpu/hip/ref /gpu/hip/shared /gpu/hip/gen
 ifneq ($(HIP_LIB_DIR),)
-  HIPCONFIG_CPPFLAGS := $(subst =,,$(shell $(ROCM_DIR)/bin/hipconfig -C))
-  $(hip-all.c:%.c=$(OBJDIR)/%.o) $(hip-all.c:%=%.tidy): CPPFLAGS += $(HIPCONFIG_CPPFLAGS)
-  ifneq ($(CXX), $(HIPCC))
-    $(hip-all.cpp:%.cpp=$(OBJDIR)/%.o) $(hip-all.cpp:%=%.tidy): CPPFLAGS += $(HIPCONFIG_CPPFLAGS)
+  ifeq ($(HIP_LIB_NAME),CHIP)
+    # chipStar hipconfig -C emits clang-only flags; keep only -D/-I/-include for gcc
+    HIPCONFIG_CPPFLAGS := $(shell $(ROCM_DIR)/bin/hipconfig -C)
+    HIPCONFIG_CPPFLAGS_C := $(filter-out --offload% -nohipwrapperinc --hip-path% --target%,$(HIPCONFIG_CPPFLAGS)) -I$(ROCM_DIR)/include
+  else
+    HIPCONFIG_CPPFLAGS := $(subst =,,$(shell $(ROCM_DIR)/bin/hipconfig -C))
+    HIPCONFIG_CPPFLAGS_C := $(HIPCONFIG_CPPFLAGS)
   endif
-  PKG_LIBS += -L$(abspath $(HIP_LIB_DIR)) -lamdhip64 -lhipblas
+  $(hip-all.c:%.c=$(OBJDIR)/%.o) $(hip-all.c:%=%.tidy): CPPFLAGS += $(HIPCONFIG_CPPFLAGS_C)
+  ifneq ($(CXX), $(HIPCC))
+    $(hip-all.cpp:%.cpp=$(OBJDIR)/%.o) $(hip-all.cpp:%=%.tidy): CPPFLAGS += $(HIPCONFIG_CPPFLAGS_C)
+  endif
+  PKG_LIBS += -L$(abspath $(HIP_LIB_DIR)) -l${HIP_LIB_NAME} -lhipblas
   LIBCEED_CONTAINS_CXX = 1
   libceed.c     += $(hip-all.c)
   libceed.cpp   += $(hip-all.cpp)
--- a/backends/hip-gen/ceed-hip-gen-operator-build.cpp
+++ b/backends/hip-gen/ceed-hip-gen-operator-build.cpp
@@ -12,6 +12,7 @@
 #include <ceed/gen-tools.h>
 #include <ceed/jit-tools.h>
 
+#include <cstring>
 #include <iostream>
 #include <sstream>
 #include <string>
--- a/README.md
+++ b/README.md
@@ -208,6 +208,10 @@ The `/gpu/hip/*` backends provide GPU performance strictly using HIP.
 They are based on the `/gpu/cuda/*` backends.
 ROCm version 4.2 or newer is required.
 
+The `/gpu/hip/*` backends can also run on non-AMD GPUs (e.g., Intel) via [chipStar](https://github.com/CHIP-SPV/chipStar), which implements HIP on top of SPIR-V through Level Zero or OpenCL.
+To build against chipStar, set `HIP_DIR` to the chipStar install prefix (in place of `ROCM_DIR`); libCEED's Makefile detects chipStar by inspecting `hipconfig` and automatically enables the required code paths.
+At runtime, chipStar's own environment variables (e.g., `CHIP_BE=level0` or `CHIP_BE=opencl`, `CHIP_DEVICE_TYPE`, `CHIP_PLATFORM`) select the backend and device — see the chipStar documentation for details.
+
 The `/gpu/sycl/*` backends provide GPU performance strictly using SYCL.
 They are based on the `/gpu/cuda/*` and `/gpu/hip/*` backends.
 


pvelesko:
@jeremylt When will the CI run for this PR?

jeremylt:
We have to manually trigger any jobs from forks, so just let us know when you want to run CI. I just approved CI running for the current push.

include/ceed/jit-source/hip/hip-ref-basis-nontensor.h
```diff
@@ -20,37 +20,65 @@
 //------------------------------------------------------------------------------
 extern "C" __global__ void Interp(const CeedInt num_elem, const CeedScalar *__restrict__ d_B, const CeedScalar *__restrict__ d_U,
                                   CeedScalar *__restrict__ d_V) {
+#ifdef __HIP_PLATFORM_SPIRV__
+  CeedInt elem = blockIdx.x * blockDim.z + threadIdx.z;
```
jeremylt: Style - variable declaration blocks should be followed by a blank line

Style - multi line if statements without { } are not permitted 

The contributing guide outlines style rules like that which are harder for our linter to catch

include/ceed/jit-source/hip/hip-ref-basis-nontensor.h
```diff
@@ -20,37 +20,65 @@
...
+  if (elem < num_elem)
+    Contract<BASIS_NUM_COMP, BASIS_Q_COMP_INTERP, BASIS_P, BASIS_Q>(elem, BASIS_P, BASIS_Q, BASIS_P * num_elem, BASIS_Q * num_elem,
+                                                                    BASIS_NUM_COMP * BASIS_Q * num_elem, d_B, d_U, d_V);
+#else
   for (CeedInt elem = blockIdx.x * blockDim.z + threadIdx.z; elem < num_elem; elem += gridDim.x * blockDim.z) {
     Contract<BASIS_NUM_COMP, BASIS_Q_COMP_INTERP, BASIS_P, BASIS_Q>(elem, BASIS_P, BASIS_Q, BASIS_P * num_elem, BASIS_Q * num_elem,
```
jeremylt, line 24: Is this line the same as above? We try to reduce duplication but it seems like these preprocessor guards are giving a lot of duplication

include/ceed/jit-source/hip/hip-ref-basis-nontensor.h
```diff
@@ -60,7 +88,13 @@ extern "C" __global__ void Weight(const CeedInt num_elem, const CeedScalar *__re
   const CeedInt t_id = threadIdx.x;
   // TODO load q_weight in shared memory if blockDim.z > 1?
 
+#ifdef __HIP_PLATFORM_SPIRV__
+  CeedInt elem = blockIdx.x * blockDim.z + threadIdx.z;
+  if (elem < num_elem)
```
jeremylt: These style changes will need to be checked all over

Makefile
```diff
@@ -88,6 +91,17 @@ ifneq ($(EMSCRIPTEN),)
...
 
+HIP_CONFIG_RES := $(shell $(ROCM_DIR)/bin/hipconfig)
+ifneq (,$(findstring __HIP_PLATFORM_SPIRV__,$(HIP_CONFIG_RES)))
+  HIP_LIB_NAME = CHIP
+else ifneq (,$(findstring __HIP_PLATFORM_HCC__,$(HIP_CONFIG_RES)))
+  HIP_LIB_NAME = amdhip64
+else ifneq (,$(findstring __HIP_PLATFORM_AMD__,$(HIP_CONFIG_RES)))
+  HIP_LIB_NAME = amdhip64
+else 
+  $(error "HIP platform not supported")
```
jeremylt: We must support platforms that have no HIP components installed  (by not attempting to build with HIP)

jeremylt (changes requested):

pvelesko:
@jeremylt I've addressed the review feedback in the latest push:

1. **Makefile** ([db0d71ce → eb092232](https://github.com/CEED/libCEED/pull/1942/files)): `hipconfig` invocation is now guarded by `ROCM_DIR` non-empty AND `hipconfig` binary present, so non-HIP machines build cleanly. Default `HIP_LIB_NAME=amdhip64`.

2. **Style fixes (blank lines, brace-around-if)** + **duplication reduction**: introduced [`include/ceed/jit-source/hip/hip-element-loop.h`](https://github.com/CEED/libCEED/pull/1942/files) with three macros:
   - `CEED_HIP_ELEM_LOOP_BEGIN(num_elem)` / `_END` — for kernels with no shared-memory ops
   - `CEED_HIP_ELEM_LOOP_BEGIN_SHARED(num_elem)` / `_END_SHARED` — for kernels that need to reach `__syncthreads()` unconditionally
   - `CEED_HIP_IF_ELEM_VALID` — guards individual loads/stores; expands to `if (elem < num_elem)` on chipStar and to nothing on AMD/NVIDIA (where the for-loop already enforces validity)

   Both `hip-ref-basis-nontensor.h` and `hip-shared-basis-tensor.h` now use the macros — no more duplicated `#ifdef __HIP_PLATFORM_SPIRV__` blocks. The code-generator (`ceed-hip-gen-operator-build.cpp`) emits the macro names too, so the JIT'd source benefits from the same single source of truth.

   Net result: `hip-shared-basis-tensor.h` shrank from ~760 to ~600 lines (and the diff vs upstream is much smaller).

3. **Verified**: clean build with chipStar, `t320-basis` and `ex1-volume` (5M DOFs, 3D) pass on all 3 HIP backends (ref/shared/gen). Both AMD and SPIRV macro paths preprocess and compile cleanly.

Could you re-trigger CI when convenient? Thanks!

pvelesko:
Follow-up: replaced the stderr line filter with `CHIP_LOGLEVEL=crit` set in the test process environment ([2b8414e8](https://github.com/CEED/libCEED/pull/1942/commits)). Cleaner — chipStar suppresses the noise at source instead of filtering after the fact. Could you re-trigger CI?

jrwrigh:
@jeremylt We disabled the HIP backend checks when we removed the AMD card from Noether, right? Did we ever try and get it back going with the NVIDIA card? I'm afraid the CI may not be actually testing the code here (assuming I understand how chip star interacts with the backends).

jeremylt:
> @jeremylt We disabled the HIP backend checks when we removed the AMD card from Noether, right? Did we ever try and get it back going with the NVIDIA card? I'm afraid the CI may not be actually testing the code here (assuming I understand how chip star interacts with the backends).

Correct - CI is only testing that the build system is correct and that the HIP backend compiles correctly.

tests/junit_common.py
```diff
@@ -763,6 +763,9 @@ def init_process():
     global my_env
     my_env = os.environ.copy()
     my_env['CEED_ERROR_HANDLER'] = 'exit'
+    # Suppress chipStar runtime informational/warning lines on stderr so they
+    # do not get classified as test failures by JUnit's stderr check.
+    my_env.setdefault('CHIP_LOGLEVEL', 'crit')
```
jeremylt: @zatkins-dev do we want to stick the ability to add extra env settings in somewhere so this change can move out of the common junit file?

jeremylt:
Overall this is looking pretty great. I need to do manual testing of the changes since ROCm testing isn't part of CI (as discussed above). Can you post the output of `make prove -j` for the full test suite so we have documented in the PR the test suite passing on a machine using chipStar?

tests/junit_common.py
zatkins-dev: This could probably be added as a per test case feature? Is there not a way to suppress this when calling the chipStar library? It would make more sense to suppress these from inside libCEED unless the use expressly opts in to seeing this debug output.

tests/junit_common.py
```diff
@@ -244,7 +244,7 @@ def has_cgnsdiff() -> bool:
                           stdout=subprocess.PIPE,
                           stderr=subprocess.PIPE,
                           env=my_env)
-    return 'not found' not in proc.stderr.decode('utf-8')
+    return 'not found' not in proc.stderr.decode('utf-8', errors='replace')
```
zatkins-dev: Is this still necessary? I would prefer to avoid changes in this file if possible

tests/junit_common.py
```diff
@@ -498,7 +498,7 @@ def diff_cgns(test_cgns: Path, true_cgns: Path, cgns_tol: float) -> str:
                           stderr=subprocess.PIPE,
                           env=my_env)
 
-    return proc.stderr.decode('utf-8') + proc.stdout.decode('utf-8')
+    return proc.stderr.decode('utf-8', errors='replace') + proc.stdout.decode('utf-8', errors='replace')
```
zatkins-dev: Same here.

tests/junit_common.py
```diff
@@ -634,8 +634,8 @@ def run_test(index: int, test: str, spec: TestSpec, backend: str,
                              classname=source_path.parent,
                              elapsed_sec=time.time() - start,
                              timestamp=time.strftime('%Y-%m-%d %H:%M:%S %Z', time.localtime(start)),
-                             stdout=proc.stdout.decode('utf-8'),
-                             stderr=proc.stderr.decode('utf-8'),
+                             stdout=proc.stdout.decode('utf-8', errors='replace'),
+                             stderr=proc.stderr.decode('utf-8', errors='replace'),
```
zatkins-dev: And same here.

include/ceed/jit-source/hip/hip-element-loop.h
```diff
@@ -0,0 +1,53 @@
...
+
+#define CEED_HIP_ELEM_LOOP_BEGIN(num_elem) \
+  for (CeedInt elem = blockIdx.x * blockDim.z + threadIdx.z; elem < num_elem; elem += gridDim.x * blockDim.z) {
+#define CEED_HIP_ELEM_LOOP_END }
+
+#define CEED_HIP_ELEM_LOOP_BEGIN_SHARED(num_elem) \
+  for (CeedInt elem = blockIdx.x * blockDim.z + threadIdx.z; elem < num_elem; elem += gridDim.x * blockDim.z) {
+#define CEED_HIP_ELEM_LOOP_END_SHARED }
+
+#define CEED_HIP_IF_ELEM_VALID
```
zatkins-dev: This entire file seems unnecessary and has the effect of obfuscating the changes to the HIP backend code. If the functionality is this different (i.e., changing from having loops to not having loops), there should be a separate backend to support those changes.

Also, I do not understand why these aren't needed in the other JIT files with element loops (e.g. `include/ceed/jit-source/hip/hip-ref-basis-tensor.h`).

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -12,6 +12,7 @@
 #include <ceed/gen-tools.h>
 #include <ceed/jit-tools.h>
 
+#include <cstring>
```
zatkins-dev, line 15: remove unnecessary include 
```suggestion
```

zatkins-dev (changes requested):
I think that hijacking the HIP backend for this support is interesting, but creates more brittle and confusing code. I think that this ought to be its own backend, perhaps utilizing the same inner kernels, to clarify the differences with the existing HIP backend

include/ceed/jit-source/hip/hip-element-loop.h
zatkins-dev: The more I think about it, how is this doing the same thing when it doesn't perform the strided loop? Depending on the element type and order, we often need to do more than one element per GPU thread. 

This PR doesn't change anything about the dispatch, so I'm unsure how it can change the kernel code like this and keep the same results.

jrwrigh:
> I think that hijacking the HIP backend for this support is interesting, but creates more brittle and confusing code. I think that this ought to be its own backend, perhaps utilizing the same inner kernels, to clarify the differences with the existing HIP backend

I obviously don't work on the backends that much, so take my opinion with a grain of salt. But the point of chipStar is to just operate straight on HIP code and the changes made, while a bit confusing, are pretty minimal. Perhaps maybe there's some cleaner way to implement the changes.

zatkins-dev:
> the changes made, while a bit confusing, are pretty minima

I think my main issue is why is this necessary and in what situations. I don't understand it, which means I can't maintain it.

jeremylt:
The GPU backends are in a tricky spot with maintainability today, let alone 2 months from now, so I think its a good concern to voice. I agree that the why about not looping over elements is not clear to me either.

pvelesko:
@zatkins-dev I wouldn't call this hijacking since chipStar is a full HIP implementation+runtime. 

The only real change here is the hoisting of `syncthreads` call outside of the inner loop. SPIR-V semantics dictate that all threads must reach `syncthreads` and this case we have cases where some threads are not participating in the calculation thus never reach `syncthreads` and thus hang.

These changes are benign but I can run some performance benchmarks on MI50 before and after this refactor and report performance.

Having two HIP backends doesn't make sense to me even in the case where there is a performance impact on the AMD side. 

If I observe a performance degradation on the AMD side, we can discuss how to best set this up.

jeremylt:
If the goal is to prevent some threads from not hitting sync points, I would tackle this differently. I have done that before with other kernels. Let me take a look at the code again.

zatkins-dev:
Hijacking may have been a strong word, I meant no offense.

I am not concerned about performance, since the macro restores the existing code. I am concerned about how this can be doing the same computation when it removes a loop over elements. I am also confused as to why only the two files that you changed need these changes. As one of the people who will have to maintain this code in the future, I need to know when I should or should not be using these annotations, and I currently have no mental model for that.

include/ceed/jit-source/hip/hip-ref-basis-nontensor.h
```diff
@@ -20,37 +21,37 @@
...
 extern "C" __global__ void Interp(const CeedInt num_elem, const CeedScalar *__restrict__ d_B, const CeedScalar *__restrict__ d_U,
                                   CeedScalar *__restrict__ d_V) {
-  for (CeedInt elem = blockIdx.x * blockDim.z + threadIdx.z; elem < num_elem; elem += gridDim.x * blockDim.z) {
-    Contract<BASIS_NUM_COMP, BASIS_Q_COMP_INTERP, BASIS_P, BASIS_Q>(elem, BASIS_P, BASIS_Q, BASIS_P * num_elem, BASIS_Q * num_elem,
-                                                                    BASIS_NUM_COMP * BASIS_Q * num_elem, d_B, d_U, d_V);
-  }
+  CEED_HIP_ELEM_LOOP_BEGIN(num_elem)
+  Contract<BASIS_NUM_COMP, BASIS_Q_COMP_INTERP, BASIS_P, BASIS_Q>(elem, BASIS_P, BASIS_Q, BASIS_P * num_elem, BASIS_Q * num_elem,
+                                                                  BASIS_NUM_COMP * BASIS_Q * num_elem, d_B, d_U, d_V);
+  CEED_HIP_ELEM_LOOP_END
```
zatkins-dev: Additional question: this doesn't even call syncthreads (nor does anything else in this file) -- why does this need different behavior?

jeremylt:
I would use the approach in https://github.com/CEED/libCEED/pull/1950 instead of the macro magic.

I think it should do what you're trying to do, but clearer so it can be maintained.

I do think that for a big change like this, it is better to first create an issue so design decisions can be discussed first.

pvelesko:
After discussion with @jeremylt, I've pulled all kernel changes out of this PR. @jeremylt's PR #1950 handles the syncthreads fix with a cleaner loop-padding approach that works across all platforms without preprocessor guards. This PR now focuses only on the build-system and test-infrastructure changes needed for chipStar support.

### What's in this PR now

| Commit | Description |
|--------|-------------|
| `eb092232` | Makefile: chipStar HIP platform detection and dynamic lib name |
| `5b80b42c` | Makefile: use SYCLCXX as linker for libceed.so when SYCL is enabled |
| `f10a5a85` | Makefile: filter chipStar clang-only flags from HIPCONFIG_CPPFLAGS for gcc |
| `ad0af6e3` | hip-gen: add explicit `#include <cstring>` for chipStar compatibility |
| `8d62fa67` | tests/junit: add `--env KEY=VAL` flag for subprocess environment |

### Answers to inline review comments

**@zatkins-dev on `hip-ref-basis-nontensor.h` line 27** ("why does this file need different behavior?"):
That change has been reverted to upstream. It was unnecessarily included in the earlier version of the PR.

**@zatkins-dev on `hip-element-loop.h` line 51** (correctness concerns):
`hip-element-loop.h` has been deleted — the entire macro-based approach is gone. The kernel fix is now entirely in PR #1950.

**@zatkins-dev on `ceed-hip-gen-operator-build.cpp` line 15** (`#include <cstring>`):
The include stays. chipStar HIP headers do not transitively include `<cstring>`, which causes build failures for the four pre-existing `strcmp()` calls in this file (lines 1191, 1215, 2164, 2188 — `resource_root` comparisons). AMD ROCm headers include it transitively, so this is a no-op on non-chipStar builds.

### Note on PR #1950

I verified the loop-padding approach empirically on chipStar and found a formula bug: `elem_loop_bound = num_elem * ceil(num_elem/stride)` deadlocks when `stride > num_elem`. I posted the fix with a reproducer on PR #1950.

### chipStar test invocation (for reference)

```bash
make junit BACKENDS='/gpu/hip/shared /gpu/hip/gen /gpu/hip/ref' JUNIT_ARGS='--env CHIP_LOGLEVEL=crit'
```

### Test results

With the loop-padding approach (corrected formula equivalent to what I posted on #1950), the following pass on chipStar (Intel Arc A770, Level Zero):

- `t310-basis`, `t314-basis`, `t316-basis` on `/gpu/hip/shared` — all pass
- Basis tests previously deadlocked with the wrong formula (`hipErrorOutOfMemory` is chipStar's symptom for a workgroup barrier deadlock)

Full `make prove` results will be posted after #1950 merges and the kernel fix is in main.

backends/hip-gen/ceed-hip-gen-operator-build.cpp
pvelesko: it's needed for the four pre-existing `strcmp()` calls in this file (resource_root comparisons). chipStar HIP headers don't transitively include `<cstring>`; AMD ROCm headers do. The explicit include is a no-op on non-chipStar builds.

jeremylt:
It looks like you heavily use Claude or something similar in your workflow - don't forget to update the LLM usage the disclosure so we know how you do so.

Note - you are replying to human generated review comments with machine generated messages. I think it's worth keeping in mind that is a potential frustration point

tests/junit.py
```diff
@@ -2,6 +2,26 @@
 from junit_common import *
 
 
+def parse_env_assignment(arg: str) -> Tuple[str, str]:
```
jeremylt: @zatkins-dev is this how you want this to work and should this be in the core? JUNIT has become mainly yours, so I don't want to make changes without your OK.

Makefile
```diff
@@ -812,6 +844,8 @@ realsearch = $(search:%=%%)
 matched = $(foreach pattern,$(realsearch),$(filter $(OBJDIR)/$(pattern),$(tests) $(allexamples)))
 subsearch ?= .*
 JUNIT_BATCH ?= ''
+# Extra arguments forwarded to tests/junit.py (e.g. --env CHIP_LOGLEVEL=crit)
+JUNIT_ARGS ?=
```
jeremylt: Ideally, we should automatically suppress the Chipstar warnings when we build with Chipstar so the test suite passes - so JUNIT_ARGS should be set appropriately by default if Chipstar is enabled

Makefile
```diff
@@ -88,6 +91,23 @@ ifneq ($(EMSCRIPTEN),)
...
 
+# Detect HIP platform (chipStar/SPIR-V vs AMD ROCm) only when ROCM_DIR is set
+# and hipconfig is actually present, so machines without HIP can build other backends.
+HIP_LIB_NAME = amdhip64
+ifneq ($(ROCM_DIR),)
+  ifneq ($(wildcard $(ROCM_DIR)/bin/hipconfig),)
+    HIP_CONFIG_RES := $(shell $(ROCM_DIR)/bin/hipconfig 2>/dev/null)
+    ifneq (,$(findstring __HIP_PLATFORM_SPIRV__,$(HIP_CONFIG_RES)))
+      HIP_LIB_NAME = CHIP
+    else ifneq (,$(findstring __HIP_PLATFORM_HCC__,$(HIP_CONFIG_RES)))
```
jeremylt:
```suggestion
      HIP_LIB_NAME = CHIP
      JUNIT_ARGS  += --env CHIP_LOGLEVEL=crit
    else ifneq (,$(findstring __HIP_PLATFORM_HCC__,$(HIP_CONFIG_RES)))
```
I think this will work?

tests/junit.py
zatkins-dev: I think this is mostly reasonable -- it's a fine feature to support in the libCEED junit suite if we need it. 

I do think that if our tests are triggering chipstar warnings, then either
1. The warnings are indicating an actual issue, in which case we should fix whatever is causing the warning
2. The warning is not a real issue, in which case chipStar should not be emitting it by default.

It is pretty normal to expect libraries to be quiet/have a way to programmatically suppress output rather than only relying on environment variables, as environment variables are not portable and are subject to annoying requirements as in this case.

Just to be clear, does running this not work? We use environment variables like PETSC_OPTIONS in this way for the test suite, so it would be odd if that didn't work for chipstar.
```sh
export CHIP_LOGLEVEL=crit
make prove -j
```
If so, we should just do that instead if/when we add it to CI, as that would be easier.

I also think that we could (and perhaps should) set the environment within libCEED using the C standard library so that users are not forced to also set this environment variable to suppress output in their test suites. 


TLDR, we should just set the environment before calling make junit, and if that doesn't work can chipStar please add a way to set this log level via a library function or something and/or can we fix whatever is causing the warnings?

pvelesko: The reason why chipStar is throwing warnings is because you're passing JIT compile options that we don't support and thus we throw warning that JIT options are ignored.

Setting the env var before `make prove` works.

jeremylt: Oh, which options are causing the trouble? I'd rather we just not pass those options when we're using Chipstar

pvelesko:
```
  CHIP warning: hiprtc: ignored option: '-default-device'
  CHIP warning: hiprtc: ignored option: '--gpu-architecture=unavailable'
  CHIP warning: hiprtc: ignored option: '-munsafe-fp-atomics'
```

jeremylt: Ok, https://github.com/CEED/libCEED/pull/1954

This way we stop the issue at its root instead of hiding it

Makefile
jeremylt: With my PR #1954, we should be able to drop the changes to `junit.py` and the relevant changes in the Makefile as well

pvelesko:
@jeremylt rebased

jeremylt:
Thanks for the work on this, it looks just about ready @pvelesko 

Can you post the output from `make prove -j` here showing that the HIP backends pass with these changes? I like to include that in PRs when the change is not covered by CI (we don't currently have CI for our HIP backends working).

---

@jrwrigh you may want to look at this. Unclear how much the perf difference is chipStar vs SYCL and how much is my perf improvements to the HIP backend family. (but this could provide motivation to find someone with the skills to rework the SYCL backend now that I've overhauled CUDA/HIP backends). I wonder how this looks for HONEE's performance on Intel hardware.

jrwrigh:
I don't have access to Aurora at the moment. Or at least I'm 90% sure I don't have access to any computer time on there. I can try and double check that tomorrow or Monday.

pvelesko:
I tested on my local system and on Aurora. All tests pass except for one (unrelated to this PR, working on it). On Aurora haven't been able to run SYCL tests at all:

```
    terminate called after throwing an instance of 'sycl::_V1::exception'
      what():  No kernel named _ZTSZZN4sycl3_V16detail16NDRangeReductionILNS1_9reduction8strategyE1EE3runINS1_9auto_nameELi1E...CeedVectorNorm_Sycl... was found

    Exit code 134 (SIGABRT from std::terminate()).
```



```
  HIP (chipStar) Performance — ex3-volume, 50× CeedOperatorApply:

  ┌─────────────────┬──────────┬────────┬──────────────────────┐
  │     Backend     │ 500K DOF │ 4M DOF │ 4M DOF (DOF·apply/s) │
  ├─────────────────┼──────────┼────────┼──────────────────────┤
  │ /gpu/hip/ref    │ 486 ms   │ 848 ms │ 235M                 │
  ├─────────────────┼──────────┼────────┼──────────────────────┤
  │ /gpu/hip/shared │ 467 ms   │ 768 ms │ 260M                 │
  ├─────────────────┼──────────┼────────┼──────────────────────┤
  │ /gpu/hip/gen    │ 451 ms   │ 653 ms │ 306M                 │
  └─────────────────┴──────────┴────────┴──────────────────────┘
```

jeremylt:
Well, like I said above, will merge this once you post the output of `make prove`.

pvelesko:
[aurora-make-prove-hip.txt](https://github.com/user-attachments/files/26868398/aurora-make-prove-hip.txt)

pvelesko:
Aurora Benchmarking Results:

```
(200 runs × 50M nodes):

  ┌───────────────────┬────────────┬─────────────┬────────────┐
  │      Backend      │ ex1-volume │ ex2-surface │ ex3-volume │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ hip/gen level0    │ 4.6s       │ 7.6s        │ 9.1s       │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ hip/gen opencl    │ 4.7s       │ 7.6s        │ 9.2s       │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ sycl/gen          │ 5.8s       │ 8.6s        │ 9.7s       │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ hip/shared level0 │ 5.6s       │ 11.4s       │ 12.2s      │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ hip/shared opencl │ 5.7s       │ 11.5s       │ 12.2s      │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ sycl/shared       │ 6.9s       │ 10.6s       │ 12.4s      │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ sycl/ref          │ 7.0s       │ 12.0s       │ 14.3s      │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ hip/ref level0    │ 7.3s       │ 14.4s       │ 16.7s      │
  ├───────────────────┼────────────┼─────────────┼────────────┤
  │ hip/ref opencl    │ 7.6s       │ 14.8s       │ 17.3s      │
  └───────────────────┴────────────┴─────────────┴────────────┘
```

pvelesko:
@jeremylt make prove results attached btw

jeremylt:
It just occurred to me that we need to tell the users how to turn on this feature.

Can you add 2-3 sentences explaining what chipStar is and what environment variables the user needs to set here:
https://github.com/CEED/libCEED/blob/5a7ad6a3b630f6659fc947b7e79a4f82a605ff49/README.md?plain=1#L209

That's the last thing I think we need to merge!

pvelesko:
@jeremylt readme updated

jeremylt:
And libCEED now officially supports chipStar :tada:

PR #1950, opened 2026-04-09
gpu - pad out elem loop for shared/gen

jeremylt:
Purpose:

Ensure all threads hit all `syncthreads()` for #1942

Closes: #N/A

LLM/GenAI Disclosure:

None

### DIFF
--- a/backends/hip-gen/ceed-hip-gen-operator-build.cpp
+++ b/backends/hip-gen/ceed-hip-gen-operator-build.cpp
@@ -500,13 +500,17 @@ static int CeedOperatorBuildKernelRestriction_Hip_gen(std::ostringstream &code, 
         CeedInt comp_stride;
 
         CeedCallBackend(CeedElemRestrictionGetLVectorSize(elem_rstr, &l_size));
+        code << tab << "if (e < num_elem) {\n";
+        tab.push();
         code << tab << "const CeedInt l_size" << var_suffix << " = " << l_size << ";\n";
         CeedCallBackend(CeedElemRestrictionGetCompStride(elem_rstr, &comp_stride));
-        code << tab << "const CeedInt comp_stride" << var_suffix << " = " << comp_stride << ";\n";
+        code << tab << "const CeedInt comp_stride" << var_suffix << " = " << comp_stride << ";\n\n";
         data->indices.outputs[i] = (CeedInt *)rstr_data->d_offsets;
         code << tab << "WriteLVecStandard" << (is_all_tensor ? max_dim : 1) << "d<num_comp" << var_suffix << ", comp_stride" << var_suffix << ", "
              << P_name << ">(data, l_size" << var_suffix << ", elem, indices.outputs[" << i << "], r_e" << var_suffix << ", d" << var_suffix
              << ");\n";
+        tab.pop();
+        code << tab << "}\n";
         break;
       }
       case CEED_RESTRICTION_STRIDED: {
@@ -520,11 +524,15 @@ static int CeedOperatorBuildKernelRestriction_Hip_gen(std::ostringstream &code, 
         if (!has_backend_strides) {
           CeedCallBackend(CeedElemRestrictionGetStrides(elem_rstr, strides));
         }
+        code << tab << "if (e < num_elem) <\n";
+        tab.push();
         code << tab << "const CeedInt strides" << var_suffix << "_0 = " << strides[0] << ", strides" << var_suffix << "_1 = " << strides[1]
-             << ", strides" << var_suffix << "_2 = " << strides[2] << ";\n";
+             << ", strides" << var_suffix << "_2 = " << strides[2] << ";\n\n";
         code << tab << "WriteLVecStrided" << (is_all_tensor ? max_dim : 1) << "d<num_comp" << var_suffix << ", " << P_name << ", strides"
              << var_suffix << "_0, strides" << var_suffix << "_1, strides" << var_suffix << "_2>(data, elem, r_e" << var_suffix << ", d" << var_suffix
              << ");\n";
+        tab.pop();
+        code << tab << "}\n";
         break;
       }
       case CEED_RESTRICTION_POINTS:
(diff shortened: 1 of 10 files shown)


nbeams:
I don't know about shared, but for `hip-gen`, I looked into this awhile back when we first started testing chipStar (so long ago it still had a different name...). I investigated some different kernel options; IIRC none of them were identical to what you've done here, but I think one was very similar (adding checks around the element restriction to avoid trying to read/write out of memory and removing the current loop bounds). When running the `hip-gen` backend on AMD hardware/with `hiprtc`, it caused some pretty large performance losses for the Poisson operator, especially as I increased basis function order. The compiler output showed increased register usage which led to a drop in occupancy. I have some old notes I could dig up from a past quarterly report with the actual numbers.

Anyway, I would recommend some updated performance testing before merging this for all backends. In case it's still an issue, would there be a way to know that the kernel will be built with chipStar and only add the element check in that case?

jeremylt:
These changes should not have any effect on register pressure I don't think? Here I am keeping the same strategy we currently have but making sure every thread is working during the last block of elements by padding with valid dummy data

nbeams:
Oh, I guess I didn't look closely enough at the code here. I also tried a version that had any "leftover" threads doing a dummy read/write, though I think they were all reading from the same (valid) element rather than padded data (which could definitely affect things). Anyway, it also had performance drops over what we currently had in `hip-gen`.

Just a warning since I didn't expect the perf drops I saw before I did the tests, either. It's not exactly the same code and of course `hiprtc` has changed since then, so no idea if it will be a problem, but I'd still recommend checking just to be sure before merging.

jeremylt:
For sure. If we see a performance difference, then I think the way to go for ChipStar would be to make chipstar backends `/gpu/hip/chipstar/shared` and `/gpu/hip/chipstar/gen` that delegate back to the current shared/gen code and that code would check the resource string for the root `/gpu/hip/chipstar` to determine if it needs to do the padding elements

pvelesko:
The `elem_loop_bound` formula has a bug when `stride > num_elem` (last block has more threads than remaining elements):

**Example** (`t314-basis` on `/gpu/hip/shared`, p=8 q=10):
```
num_elem = 63, blockDim.z = 64, gridDim.x = 1
stride = 64
elem_loop_bound = 63 * ceil(63/64) = 63 * 1 = 63
Thread 63: e=63, 63 < 63 → false → skips loop → misses __syncthreads() inside ContractX1d → deadlock
```

**Fix** — multiply by `stride`, not `num_elem`:
```cpp
const CeedInt stride          = gridDim.x * blockDim.z;
const CeedInt elem_loop_bound = stride * ((num_elem + stride - 1) / stride);
```

With the wrong formula, `t314-basis` and `t316-basis` on `/gpu/hip/shared` deadlock and return `hipErrorOutOfMemory` on chipStar (that error code is how chipStar signals a workgroup barrier deadlock). With the corrected formula they pass.

There is also a typo in `backends/hip-gen/ceed-hip-gen-operator-build.cpp` in the `CEED_RESTRICTION_STRIDED` case — `<\n` should be `{\n`:
```cpp
code << "if (e < num_elem) <\n"   // typo: < should be {
```

jeremylt:
That's not quite the correct fix - it's logically inconsistent with what the word stride means in the codebase. But now that I see where the issue is I can create in the a fix. Thanks

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -520,11 +524,15 @@ static int CeedOperatorBuildKernelRestriction_Hip_gen(std::ostringstream &code,
         if (!has_backend_strides) {
           CeedCallBackend(CeedElemRestrictionGetStrides(elem_rstr, strides));
         }
+        code << tab << "if (e < num_elem) <\n";
```
zatkins-dev:
```suggestion
        code << tab << "if (e < num_elem) {\n";
```

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -1495,8 +1507,15 @@ extern "C" int CeedOperatorBuildKernel_Hip_gen(CeedOperator op, bool *is_good_bu
   // Loop over all elements
   code << "\n" << tab << "// Element loop\n";
   code << tab << "__syncthreads();\n";
-  code << tab << "for (CeedInt elem = blockIdx.x*blockDim.z + threadIdx.z; elem < num_elem; elem += gridDim.x*blockDim.z) {\n";
+  code << tab << "#if CEED_HIP_USE_CHIPSTAR\n";
+  code << tab << "// Pad out elements so all threads hit syncthreads()\n";
+  code << tab << "const CeedInt elem_loop_bound = (gridDim.x*blockDim.z) * ceil(1.0*num_elem/(gridDim.x*blockDim.z));\n\n";
+  code << tab << "#else\n";
+  code << tab << "const CeedInt elem_loop_bound = num_elem;\n\n";
+  code << tab << "#endif\n";
```
jeremylt, line 1515: @zatkins-dev does this make it clearer now what's happening?

zatkins-dev: Yeah, I think so!

jeremylt:
@pvelesko can you confirm these changes do what you need? If not then I can merge and there's just a couple of small tweaks I'd like to request for your branch

include/ceed/jit-source/hip/hip-shared-basis-nontensor.h
```diff
@@ -35,10 +35,21 @@ extern "C" __launch_bounds__(BASIS_INTERP_BLOCK_SIZE) __global__
   __syncthreads();
 
   // Apply basis element by element
-  for (CeedInt elem = blockIdx.x * blockDim.z + threadIdx.z; elem < num_elem; elem += gridDim.x * blockDim.z) {
+#if CEED_HIP_USE_CHIPSTAR
```
pvelesko, line 38: Why not `__HIP_PLATFORM_SPIRV__` ?  HIP defines these for all platforms `__HIP_PLATFORM_AMD__` and `__HIP_PLATFORM_NVIDIA__`

pvelesko:
> @pvelesko can you confirm these changes do what you need? If not then I can merge and there's just a couple of small tweaks I'd like to request for your branch

Yes all tests are passing after rebasing my PR on top of this one.

include/ceed/jit-source/hip/hip-shared-basis-nontensor.h
jeremylt: Better clarity of intent. This part of the codebase is a maintenance vulnerability because fewer people are familiar with it.

But, I think there's a better place for me to put the setting of that macro value

Issue #1830, opened 2025-05-30
CeedVectorFilter


jeremylt:
@zatkins-dev would it be adequate to your purposes to add this in libCEED? (Though you could access it via PETSc also)

https://petsc.org/release/manualpages/Vec/VecFilter/

jedbrown:
I feel it. FWIW, it was once `Chop` and changed to `Filter` in https://gitlab.com/petsc/petsc/-/merge_requests/6699.

A closely related name is the ["shrink" operation](https://en.wikipedia.org/wiki/Bregman_method#Linearized_Bregman).

jeremylt:
I think its "filter" like "high pass filter" instead of filter as seen in functional programming?

jeremylt:
That feels reasonable - I don't like deviating from the names PETSc uses if we can avoid it, even if I hate the name PETSc gave a function

wostrie2:
I would like to implement this. I am new to libCEED

wostrie2:
Ok, thank you for giving me a place to start and pointing out different things that will need to be addressed. Yes, it does feel like a lot all at once. I will start looking at things and hit you up with questions.

jeremylt:
I would recommend this order, personally

1) Interface layer changes (with the default version of the logic)

2) Documentation (goes above the function in `ceed-vector.c`, looks like this:)
```
/**
  @brief Filter a `CeedVector` by a threshold value.
         All entries of `x` with an absolute value less than or equal to `threshold` are set to `0.0`.

  @param[in,out] x         `CeedVector` for filtering
  @param[in]     threshold Threshold value

  @return An error code: 0 - success, otherwise - failure

  @ref User
**/
```

3) Test

4) CUDA/HIP

---

Yeah, I would fork and clone the repo. You can run a specific example with the command `make prove PROVE_OPTS=-v search=t100` (`t1**` tests are vector tests)

I can help you address any issues you run into with CUDA/HIP

wostrie2:
So, I was just checking some basic syntax with google AI, and it recommended implementing the logic as follows:

CeedPragmaSIMD
  for (CeedSize i = 0; i < length; i++) {
    if (x_array[i] <= threshold && x_array[i] >= -threshold) {
      x_array[i] = 0.0;
    }
  }

where the macro acts as a hint to the compiler to vectorize the loop, and the change from fabs() to direct comparisons is for type safety.

@jeremylt  is this a good use of this macro?

jeremylt:
[...]

PR #1951, opened 2026-04-15
Adding filter function

wostrie2:
Purpose:

Add CeedVectorFilter function to filter or clip a CeedVector using a threshold value. Vector entries with absolute value less than `threshold` are clipped to `0.0`.

This may be considered a WIP pull request.

Default implementation of the function and a test, t129-vector.c, are implemented. 

GPU backends have NOT been implemented yet.

Addresses Issue #1830 

LLM/GenAI Disclosure: Google AI mode was used to check syntax for the function, and its suggestion to use the `CeedPragmaSIMD` macro was adopted. Google AI mode was used to write piece of the test for this function, `t129-vector.c`


tests/t129-vector.c
```diff
@@ -0,0 +1,81 @@
...
+      printf("  Result: FAIL\n");
+      test_errors++;
+    }
+    CeedVectorDestroy(&x);
+  }
+
+  // Cleanup and Exit
+  CeedDestroy(&ceed);
+  return test_errors;
+}
```
jeremylt:
```suggestion
}

```

jeremylt:
This looks pretty good. The test is a rather different style than what we usually use, so I'll have some comments for you there soon. Let me know if you have questions as you keep working!

interface/ceed-vector.c
```diff
@@ -756,6 +756,44 @@ int CeedVectorScale(CeedVector x, CeedScalar alpha) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Filters or clips a `CeedVector` using a threshold value. 
+  All entries in `x` with an absolute value less than or equal to `threshold` are set to `0.0`.
+
+  @param[in,out] x     `CeedVector` to filter
+  @param[in]     threshold clipping threshold or tolerance
```
jeremylt:
```suggestion
  @param[in,out] x         `CeedVector` to filter
  @param[in]     threshold clipping threshold or tolerance
```
minor style fix

interface/ceed-vector.c
```diff
@@ -756,6 +756,44 @@ int CeedVectorScale(CeedVector x, CeedScalar alpha) {
...
+  CeedCall(CeedVectorHasValidArray(x, &has_valid_array));
+  CeedCheck(has_valid_array, CeedVectorReturnCeed(x), CEED_ERROR_BACKEND,
+            "CeedVector has no valid data to scale, must set data with CeedVectorSetValue or CeedVectorSetArray");
+
+  // Return early for empty vector
+  CeedCall(CeedVectorGetLength(x, &length));
+  if (length == 0) return CEED_ERROR_SUCCESS;
+
+  // Backend implementation
+  if (x->Filter) return x->Filter(x, threshold);
```
jeremylt:
```suggestion
  if (x->Filter) {
    CeedCall(x->Filter(x, threshold));
    return CEED_ERROR_SUCCESS;
  }
```
Really, for clearer stack traces, we should do this. I need to fix in a few places (like CeedVectorScale)

jeremylt (commented):
The bulk of this outside of `t129-vector.c` looks really good. Lemmie cook up a separate comment about what exactly I would do for the test though

jeremylt:
Ok, here is what I would do for the tests.

You should think of the tests as an extension of the documentation. To that end, they should all have a similar look/feel so that it is easier for people to review the tests and see how specific functions work.

So, I would start with an *existing* test instead of making a brand new one.

Let's start with `t123-vector.c`, the `CeedVectorScale()` test.

---

```C
/// @file
/// Test scaling a vector
/// \test Test scaling of a vector

//TESTARGS(name="length 10") {ceed_resource} 10
//TESTARGS(name="length 0") {ceed_resource} 0
#include <ceed.h>
#include <math.h>
#include <stdio.h>
#include <stdlib.h>

  Ceed       ceed;
  CeedVector x;
  CeedInt    len = 10;

  CeedInit(argv[1], &ceed);
  len = argc > 2 ? atoi(argv[2]) : len;

  CeedVectorCreate(ceed, len, &x);
  {
    CeedScalar array[len];

    for (CeedInt i = 0; i < len; i++) array[i] = 10 + i;
    CeedVectorSetArray(x, CEED_MEM_HOST, CEED_COPY_VALUES, array);
  }
  {
    // Sync memtype to device for GPU backends
    CeedMemType type = CEED_MEM_HOST;
    CeedGetPreferredMemType(ceed, &type);
    CeedVectorSyncArray(x, type);
  }
  CeedVectorScale(x, -0.5);

  {
    const CeedScalar *read_array;

    CeedVectorGetArrayRead(x, CEED_MEM_HOST, &read_array);
      if (fabs(read_array[i] + (10.0 + i) / 2) > 1e-14) {
        // LCOV_EXCL_START
        printf("Error in alpha x at index %" CeedInt_FMT ", computed: %f actual: %f\n", i, read_array[i], -(10.0 + i) / 2);
        // LCOV_EXCL_STOP
      }
    }
    CeedVectorRestoreArrayRead(x, &read_array);
  }

  CeedVectorDestroy(&x);
  CeedDestroy(&ceed);
  return 0;
}
```

---

I like to consider the edge cases here. Two come to mind immediately
* positive vs negative values
* less than vs equal to the threshold

The test file `t123-vector.c` suggests a third edge case
* length 0 vectors

So that should guide our modification of this code to make a test for `CeedVectorFilter`.

If we change
```C
    for (CeedInt i = 0; i < len; i++) array[i] = 10 + i;
 ```
into a different pattern, then we can build array contents that hit our edge case concerns.
Perhaps something like `2 + i * pow(-1, i)`?

I would then replace the `CeedVectorScale` call with a `CeedVectorFilter` call (to do this, you probably want to create a `tolerance` variable?`

I would then replace the logic following `CeedVectorGetArrayRead` to ensure that the contents of `read_array` are the same as the pattern you picked above unless the value should be filtered out.

It may be worth doing this twice, the second time with a bigger value of `threshold`. That way you could check with a threshold value that is the same as one of the array entries and a threshold value that is in between two array entry values.

Does that all make sense?

**Comment by wostrie2** [comment]
Thanks, Jeremey! I think that's all making sense. Ah, it makes sense how you are explaining the tests in general. I should have asked about the test for the Scale function.

Are you thinking of just making one vector and changing the threshold for the different test cases?

**Comment by jeremylt** [comment]
Yeah, that is how I would do it. I don't see a reason why the vector itself would make a big difference, unless its of a special length (length 0 specifically).

CI has given the following style fix:
```diff
index 4abc4d6..d829c7a 100644
--- a/interface/ceed-vector.c
+++ b/interface/ceed-vector.c
@@ -789,8 +789,7 @@ int CeedVectorFilter(CeedVector x, CeedScalar threshold) {
   // Default implementation
   CeedCall(CeedVectorGetArray(x, CEED_MEM_HOST, &x_array));
   assert(x_array);
-  CeedPragmaSIMD
-  for (CeedSize i = 0; i < length; i++) {
+  CeedPragmaSIMD for (CeedSize i = 0; i < length; i++) {
     if (fabs(x_array[i]) <= threshold) x_array[i] = 0.0;
   }
   CeedCall(CeedVectorRestoreArray(x, &x_array));
```

wostrie2:
Cool, I made the style fix suggested by CI.

For the test, should I let it run the case of the zero-length vector the same way that t123 does?

Oh, never mind, I see that you mentioned this above: "The test file t123-vector.c suggests a third edge case

length 0 vectors"

hmmm, google AI is recommending wrapping the call to create the vector with a check that len > 0, but I think t123 does NOT do this?

```
CeedVectorCreate(ceed, len, &x);
  if (len > 0) {
    CeedScalar array[len];
    for (CeedInt i = 0; i < len; i++) array[i] = (1.0 + i) * pow(-1, i);
    CeedVectorSetArray(x, CEED_MEM_HOST, CEED_COPY_VALUES, array);
  }
```

What do you think?

jeremylt:
What will the for loop do if len is 0?

I don't think Google AI is making a sensible suggestion here.

Like with any code or code output, I would consider the Google AI output to be wrong until you check it to see if it makes sense and does anything useful.

wostrie2:
The loop should be skipped if len is 0, so things should function as intended. And this pattern seems fine in t123, which is why I didn't want to deviate from it without asking about it.

Google AI was pointing out that a VLA of length 0 is undefined behavior according to the C standard. Which seems to be true from a quick check on stack exchange, but I am very inexperienced in C, so I didn't know if this mattered in this case.

Yes, that is a sound assumption for any AI output.

The new test runs fine, and I think the test cases are good. I didn't break anything else - all other tests are passing.

But `make test` and `make prove` are ignoring my test. It built and it runs fine manually. Do you have any suggestions?

jeremylt:
That's a good point about a length 0 VLA technically being undefined behavior. Fortunately, we are neither reading nor writing anything, because all of our logic depends upon the length, so it's a benign theoretical issue for us.



Can you say help about what happens when you run `make prove -j search=t PROVE_OPTS=-v`? Maybe posting the output that looks like this new test isn't running?

wostrie2:
TLDR; It looks fine now :) Successful output from make prove at the bottom of this comment. Thank you!



Yep, sorry for not explaining that. When I run `make test` or `make prove` I see the t128 and t130 results and nothing for t129.

I deleted the build folder, and make test builds t129 but doesn't run it.
```
    LINK.c build/t127-vector
      LINK.c build/t128-vector
      LINK.c build/t129-vector
      LINK.c build/t130-vector
      LINK.c build/t131-vector
```

Ok, it looks good now!

I ran `make prove -j search=t PROVE_OPTS=-v`:

```
t128-vector ............. 
TAP version 13
1..1
# Subtest: t128-vector
    1..4
    ok 1 - t128-vector, /cpu/self/ref/serial (0.0058553218841552734 s)
    ok 2 - t128-vector, /cpu/self/ref/blocked (0.00545191764831543 s)
    ok 3 - t128-vector, /cpu/self/opt/serial (0.0036628246307373047 s)
    ok 4 - t128-vector, /cpu/self/opt/blocked (0.004467964172363281 s)
ok 1 - t128-vector
ok
t129-vector ............. 
TAP version 13
1..5
# Subtest: filter none
    1..4
    ok 1 - filter none, /cpu/self/ref/serial (0.005805015563964844 s)
    ok 2 - filter none, /cpu/self/ref/blocked (0.0048220157623291016 s)
    ok 3 - filter none, /cpu/self/opt/serial (0.004112958908081055 s)
    ok 4 - filter none, /cpu/self/opt/blocked (0.004045009613037109 s)
ok 1 - filter none
# Subtest: filter all
    1..4
    ok 1 - filter all, /cpu/self/ref/serial (0.0037598609924316406 s)
    ok 2 - filter all, /cpu/self/ref/blocked (0.005825042724609375 s)
    ok 3 - filter all, /cpu/self/opt/serial (0.004752159118652344 s)
    ok 4 - filter all, /cpu/self/opt/blocked (0.004585981369018555 s)
ok 2 - filter all
# Subtest: between values
    1..4
    ok 1 - between values, /cpu/self/ref/serial (0.003821849822998047 s)
    ok 2 - between values, /cpu/self/ref/blocked (0.0047149658203125 s)
    ok 3 - between values, /cpu/self/opt/serial (0.005877017974853516 s)
    ok 4 - between values, /cpu/self/opt/blocked (0.003863096237182617 s)
ok 3 - between values
# Subtest: equal to value
    1..4
    ok 1 - equal to value, /cpu/self/ref/serial (0.004484891891479492 s)
    ok 2 - equal to value, /cpu/self/ref/blocked (0.0059239864349365234 s)
    ok 3 - equal to value, /cpu/self/opt/serial (0.004312276840209961 s)
    ok 4 - equal to value, /cpu/self/opt/blocked (0.00548100471496582 s)
ok 4 - equal to value
# Subtest: empty vector
    1..4
    ok 1 - empty vector, /cpu/self/ref/serial (0.005505084991455078 s)
    ok 2 - empty vector, /cpu/self/ref/blocked (0.004116058349609375 s)
    ok 3 - empty vector, /cpu/self/opt/serial (0.0053708553314208984 s)
    ok 4 - empty vector, /cpu/self/opt/blocked (0.0046498775482177734 s)
ok 5 - empty vector
ok
t130-vector ............. 
```

jeremylt:
Awesome. You can use `search=t129` to run the new test specifically

wostrie2:
ok I rebased and pushed changes.

t129 is now very much along the lines of what you recommended.

There are changes in backend files in my PR now. I assume that these are related changes? I confess to not recognizing anything related there, but perhaps this is not unexpected since I'm not familiar with CUDA or HIP.

jeremylt:
Those changes should not be there. It looks like you somehow did something strange with the rebase and changes to main are showing up in your branch now

tests/t129-vector.c
```diff
@@ -0,0 +1,60 @@
...
+  if (argc > 3) tolerance = (CeedScalar)atof(argv[3]);
+
+  CeedVectorCreate(ceed, len, &x);
+  {
+    CeedScalar array[len];
+
+    for (CeedInt i = 0; i < len; i++) array[i] = (1.0 + i)* pow(-1, i);
+    CeedVectorSetArray(x, CEED_MEM_HOST, CEED_COPY_VALUES, array);
+  }
+  {
```
jeremylt, line 32: This block is not needed

tests/t129-vector.c
```diff
@@ -0,0 +1,60 @@
+/// @file
+/// Test filtering a vector using a threshold or tolerance
+/// \test Test filtering a vector using a threshold or tolerance
+
+//TESTARGS(name="filter none") {ceed_resource} 10 0.0
```
jeremylt, line 5: I would only specialize on length 0 or not here. I would use different fixed tolerances inside of the test file

wostrie2:
> Those changes should not be there. It looks like you somehow did something strange with the rebase and changes to main are showing up in your branch now

ok, I'm going to reset my branch to before the rebase, fetch to make sure I have the latest official main, and rebase on top of that.

wostrie2:
Ok, I reset my branch, rebased on top of the latest main, and pushed. It looks like only the relevant files are in this PR now.

tests/t129-vector.c
wostrie2: Should I delete this block? I just kept it from t123.

Will that functionality be handled by the GPU implementations of the function? Or is it handled by the function already? Apologies, my understanding of the structure is a bit hazy.

jeremylt: Yes, you should delete it. We aren't testing that function in this file

jeremylt: Actually, I think you should stick this between the first and second calls to your new filter function. That way we test the GPU backends for both host and device side

tests/t129-vector.c
wostrie2: ok, shoot, I thought that putting the tolerances in TESTARGS would be a good way to cover the test cases without adding repetitive code to the test. Also, it seemed like a nice way to easily label the test cases.

If I use fixed tolerances in the test, it seems like the verification block should become a function, and then I will be repeatedly changing the tolerance, calling filter, and verifying.

Am I thinking about this clearly? Did you have something else in mind?

jeremylt: I like your function idea.

Normally, yes we would want to avoid a lot of copy and paste. But specifically here in this folder we tend not to use the test args, so this is about consistency with the other tests in the folder.

wostrie2: Got it. Should I still use TESTARGS for length not-zero and length zero? Or not use them at all?

jeremylt: Yea, I would still use them for length 0 or not

wostrie2:
I refactored the test to move the test cases out of TESTARGS and into the test, and created a couple of helper functions: one to populate the vector with values and reset it between filter calls, and one to verify a filtered vector.

The block for syncing for GPU backends has been moved to be between the first and second filter calls.

The test builds and runs locally.

tests/t129-vector.c
```diff
@@ -0,0 +1,89 @@
...
+// Test builds the vector [1.0, -2.0, 3.0, -4.0, 5.0, -6.0, 7.0, -8.0, 9.0, -10.0]
+// and filters it using different tolerances
+static int InitVector(CeedVector x, CeedInt len) {
+  if (len <= 0) return 0; // Nothing to set for an empty vector
+
+  CeedScalar array[len];
+  for (CeedInt i = 0; i < len; i++) array[i] = (1.0 + i) * pow(-1, i);
+  CeedVectorSetArray(x, CEED_MEM_HOST, CEED_COPY_VALUES, array);
+  return 0;
+}
```
jeremylt, line 20:
```suggestion
#include <stdlib.h>

// Test builds the vector [1.0, -2.0, 3.0, -4.0, 5.0, -6.0, 7.0, -8.0, 9.0, -10.0]
// and filters it using different tolerances
static int InitVector(CeedVector x, CeedInt len) {
  if (len <= 0) return 0; // Nothing to set for an empty vector

  CeedScalar array[len];
  
  for (CeedInt i = 0; i < len; i++) array[i] = (1.0 + i) * pow(-1, i);
  CeedVectorSetArray(x, CEED_MEM_HOST, CEED_COPY_VALUES, array);
  return 0;
}
```
minor

tests/t129-vector.c
```diff
@@ -0,0 +1,89 @@
...
+        // LCOV_EXCL_START
+        printf("Error in filtered vector at index %" CeedInt_FMT ", computed: %f actual: %f\n", 
+               i, read_array[i], expected_value);
+        // LCOV_EXCL_STOP
+        VF_error_count++;
+      }
+    }
+    CeedVectorRestoreArrayRead(x, &read_array);
+    return VF_error_count;
+  }
```
jeremylt:
```suggestion
static int VerifyFilter(CeedVector x, CeedInt len, CeedScalar tolerance) {
    const CeedScalar *read_array;

    CeedVectorGetArrayRead(x, CEED_MEM_HOST, &read_array);
    for (CeedInt i = 0; i < len; i++) {
      CeedScalar initial_value = (1.0 + i) * pow(-1, i);
      CeedScalar expected_value = (fabs(initial_value) <= tolerance) ? 0.0 : initial_value;

      if (fabs(read_array[i] - expected_value) > 1e-14) {
        // LCOV_EXCL_START
        printf("Error in filtered vector at index %" CeedInt_FMT ", computed: %f actual: %f\n", 
               i, read_array[i], expected_value);
        // LCOV_EXCL_STOP
      }
    }
    CeedVectorRestoreArrayRead(x, &read_array);
    return 0;
  }
```
We don't typically return non-zero error codes  from these tests unless there is an error *inside* of libCEED

tests/t129-vector.c
```diff
@@ -0,0 +1,89 @@
...
+
+int main(int argc, char **argv) {
+  Ceed       ceed;
+  CeedVector x;
+  CeedInt    len = 10;
+  CeedScalar tolerance = 1e-10;
+  CeedInt    error_count = 0;
```
jeremylt: Similarly, I'd drop the `error_count` here too

jeremylt (commented):
Looks pretty good - just a couple of small changes

tests/t129-vector.c
wostrie2: Ok interesting, so tests always show as passing, and you just count on error reporting inside the test to show if there's an issue?

I feel like maybe I'm not clear about what you mean by "an error inside of libCEED". Do you mean an error in something that's part of libCEED's core functionality as opposed to an error in the new filter function? Sorry, just trying to make sure I understand the intention here.

jeremylt: Not quite

libCEED itself will raise a non-zero error code if there is a problem inside of the library (so anything from the files in `interface/` or `backend/` or similar). That error code will halt any test. The tests themselves don't return anything other than 0. The test harness captures anything written to stdout or stderr to flag errors

We could change the tests to return an error code if there is an issue, but that would involve reworking a lot of the tests, so it would be out of scope for this PR.

wostrie2: Ok, got it. Thanks for explaining! I will change things accordingly and comment when I push.

wostrie2:
I have deleted the error counting, made the VerifyFilter function and test return 0, and pushed the updated test.

jeremylt:
This is looking pretty good. Did you want to take a try at adding this to the CUDA ref backend?

wostrie2:
> This is looking pretty good. Did you want to take a try at adding this to the CUDA ref backend?

Yes! I do want to. I'm super interested, but I'm not sure how my timing is going to work. I'm a student with too many things due speeding toward the end of the term (like 3 weeks). I want to manage that without adding something overwhelming to my to-do list. Maybe it's not a big deal, since it's such a simple function? (he asked foolishly, after rewriting the test for the fourth time lol) 

Can you maybe give me an idea what's involved and point me at an example to look at? Is CeedVectorScale still a good case to look at on the CUDA side of things?

jeremylt:
Its overall not too bad - you've done the trickiest work already imho

The CUDA backends all use the `CeedVector` functionality found in [`/backends/cuda-ref/ceed-cuda-ref-vector.c`](https://github.com/CEED/libCEED/blob/main/backends/cuda-ref/ceed-cuda-ref-vector.c), so you only need to make changes in one place.

You'll want to put this new functionality after `CeedVectorScale_Cuda`, but that's not the best example to look at. I'd look at `CeedVectorReciprocal_Cuda` because its simpler.

------

Big picture - GPU vectors hold two copies of the data, one on the host (CPU) and one on the device (GPU).

1) You'll need a outer function that determines if the device or host memory needs to be updated. https://github.com/CEED/libCEED/blob/b0010ad8933c18b4a9c0085d246b295207ee26ce/backends/cuda-ref/ceed-cuda-ref-vector.c#L658
(ooh, I made a typo in my comment there and should fix that - [I say typo, but really it was a sloppy copy-paste])

2) You'll then make a CPU version of the function: https://github.com/CEED/libCEED/blob/b0010ad8933c18b4a9c0085d246b295207ee26ce/backends/cuda-ref/ceed-cuda-ref-vector.c#L643

3) Then you write the GPU version:
  a) In the CPU code you put the function definition: https://github.com/CEED/libCEED/blob/b0010ad8933c18b4a9c0085d246b295207ee26ce/backends/cuda-ref/ceed-cuda-ref-vector.c#L653
  b) And the GPU code itself has two parts (This is in the file `/backends/cuda-ref/kernels/cuda-ref-vector.cu`):
    1) The code that calls the GPU kernel: https://github.com/CEED/libCEED/blob/b0010ad8933c18b4a9c0085d246b295207ee26ce/backends/cuda-ref/kernels/cuda-ref-vector.cu#L95
    2) And the GPU kernel itself: https://github.com/CEED/libCEED/blob/b0010ad8933c18b4a9c0085d246b295207ee26ce/backends/cuda-ref/kernels/cuda-ref-vector.cu#L84
  For the GPU code, it may be more useful to also look at the `CeedDeviceScale_Cuda` and `scaleValueK`

4) Finally, you need to 'register' the backend function so the code in `/interface/vector.c` knows to use the backend implementation: https://github.com/CEED/libCEED/blob/b0010ad8933c18b4a9c0085d246b295207ee26ce/backends/cuda-ref/ceed-cuda-ref-vector.c#L870

------

Ok, hopefully this makes it look manageable?

wostrie2:
Ok, thank you! :) 

That's a lot of awesome guidance. Thanks for your support all through this.

I'd like to do it, but I'm concerned about the timing. Is it going to put anyone out if it takes me 3 weeks to start working on this part of things?

Apologies for the delay. The end of my school term is out of control.

tests/t129-vector.c
```diff
@@ -0,0 +1,86 @@
...
+
+  {
+    // Sync memtype to device for GPU backends
+    CeedMemType type = CEED_MEM_HOST;
+    CeedGetPreferredMemType(ceed, &type);
+    CeedVectorSyncArray(x, type);
+  }
+
+  // Test Case 2 - tolerance equal to a vector value
+  InitVector(x, len);
```
jeremylt, line 64:
```suggestion

  // Test Case 2 - tolerance equal to a vector value
  InitVector(x, len);
  {
    // Sync memtype to device for GPU backends
    CeedMemType type = CEED_MEM_HOST;
    CeedGetPreferredMemType(ceed, &type);
    CeedVectorSyncArray(x, type);
  }
```
minor - this moves the memory to the GPU before filtering to ensure that GPU backends will execute the filtering logic on the GPU if they can (`InitVector` makes the memory on the host (CPU) the most up to date, so that's where the computation would take place without this change)

jeremylt:
We're in no rush - its definitely a busy time of the semester.

Depending upon your preferences, we could do one of two things

1) Wait, you add the GPU changes when you get a chance, and we merge this branch then

2) We split this work into two - we merge the interface changes and the test now and you make a second PR to cover the GPU implementation

Either works fine for me

jeremylt:
Style fix requested by CI:
```diff
index c8fe932..d08a613 100644
--- a/tests/t129-vector.c
+++ b/tests/t129-vector.c
@@ -11,7 +11,7 @@
 // Test builds the vector [1.0, -2.0, 3.0, -4.0, 5.0, -6.0, 7.0, -8.0, 9.0, -10.0]
 // and filters it using different tolerances
 static int InitVector(CeedVector x, CeedInt len) {
-  if (len <= 0) return 0; // Nothing to set for an empty vector
+  if (len <= 0) return 0;  // Nothing to set for an empty vector
 
   CeedScalar array[len];
   for (CeedInt i = 0; i < len; i++) array[i] = (1.0 + i) * pow(-1, i);
@@ -20,29 +20,28 @@ static int InitVector(CeedVector x, CeedInt len) {
 }
 
 static int VerifyFilter(CeedVector x, CeedInt len, CeedScalar tolerance) {
-    const CeedScalar *read_array;
+  const CeedScalar *read_array;
 
-    CeedVectorGetArrayRead(x, CEED_MEM_HOST, &read_array);
-    for (CeedInt i = 0; i < len; i++) {
-      CeedScalar initial_value = (1.0 + i) * pow(-1, i);
-      CeedScalar expected_value = (fabs(initial_value) <= tolerance) ? 0.0 : initial_value;
+  CeedVectorGetArrayRead(x, CEED_MEM_HOST, &read_array);
+  for (CeedInt i = 0; i < len; i++) {
+    CeedScalar initial_value  = (1.0 + i) * pow(-1, i);
+    CeedScalar expected_value = (fabs(initial_value) <= tolerance) ? 0.0 : initial_value;
 
-      if (fabs(read_array[i] - expected_value) > 1e-14) {
-        // LCOV_EXCL_START
-        printf("Error in filtered vector at index %" CeedInt_FMT ", computed: %f actual: %f\n", 
-               i, read_array[i], expected_value);
-        // LCOV_EXCL_STOP
-      }
+    if (fabs(read_array[i] - expected_value) > 1e-14) {
+      // LCOV_EXCL_START
+      printf("Error in filtered vector at index %" CeedInt_FMT ", computed: %f actual: %f\n", i, read_array[i], expected_value);
+      // LCOV_EXCL_STOP
     }
-    CeedVectorRestoreArrayRead(x, &read_array);
-    return 0;
   }
+  CeedVectorRestoreArrayRead(x, &read_array);
+  return 0;
+}
 
 int main(int argc, char **argv) {
   Ceed       ceed;
   CeedVector x;
-  CeedInt    len = 10;
-  CeedScalar tolerance = 1e-10; 
+  CeedInt    len       = 10;
+  CeedScalar tolerance = 1e-10;
 
   CeedInit(argv[1], &ceed);
   len = argc > 2 ? atoi(argv[2]) : len;
```

wostrie2:
Ok I've edited to take care of the style fixes requested by CI. Built and ran it, and pushed changes.

jeremylt:
A couple more style fixes:
```diff
+++ b/tests/t129-vector.c
@@ -24,7 +24,7 @@ static int VerifyFilter(CeedVector x, CeedInt len, CeedScalar tolerance) {
 
   CeedVectorGetArrayRead(x, CEED_MEM_HOST, &read_array);
   for (CeedInt i = 0; i < len; i++) {
-    CeedScalar initial_value = (1.0 + i) * pow(-1, i);
+    CeedScalar initial_value  = (1.0 + i) * pow(-1, i);
     CeedScalar expected_value = (fabs(initial_value) <= tolerance) ? 0.0 : initial_value;
 
     if (fabs(read_array[i] - expected_value) > 1e-14) {
@@ -32,7 +32,7 @@ static int VerifyFilter(CeedVector x, CeedInt len, CeedScalar tolerance) {
       printf("Error in filtered vector at index %" CeedInt_FMT ", computed: %f actual: %f\n", i, read_array[i], expected_value);
       // LCOV_EXCL_STOP
     }
-    }
+  }
   CeedVectorRestoreArrayRead(x, &read_array);
   return 0;
 }
@@ -41,7 +41,7 @@ int main(int argc, char **argv) {
   Ceed       ceed;
   CeedVector x;
   CeedInt    len       = 10;
-  CeedScalar tolerance = 1e-10; 
+  CeedScalar tolerance = 1e-10;
 
   CeedInit(argv[1], &ceed);
   len = argc > 2 ? atoi(argv[2]) : len;
```

jeremylt:
@valeriabarra this PR is in pretty good shape. We've just got a style fix and a requirement I forgot to tell @wostrie2 about.

@wostrie2, whenever we make a change to the interface, we add a brief description to the CHANGELOG (https://github.com/CEED/libCEED/blob/db9619436dd97d50def0cffa9b0d091bdfd79936/CHANGELOG.md) so users know its there. Something like "Add `CeedVectorFilter` to zero values in a `CeedVector` below a specified threshold value" under the heading **New features** at the top for current changes to main (or however you'd phrase it)

@wostrie2 wants to take a crack at adding CUDA/HIP backend implementations as well. That's certainly helpful and I plan on supporting them as needed through this, but this PR is already a value add with the interface level changes and default implementation. Thank you for the hard work here!

valeriabarra:
Amazing! Yes, I also believe the other implementations could follow in separate PRs, but it is up to the author and maintainers to decide on that. I gave @wostrie2 access to some hardware to be able to test the CUDA implementation. I think he might still need/want to also test on AMD GPUs (which I could not offer at the moment).

jeremylt:
Fantastic! I'm happy to do the GPU bit in this PR or a followup as @wostrie2 prefers. 

We can help with testing for AMD hardware. Once the CUDA impl works, its mostly a copy-paste from CUDA to HIP, so I'm not seeing that as a big vulnerability to this development effort.

wostrie2:
> Fantastic! I'm happy to do the GPU bit in this PR or a followup as @wostrie2 prefers.
> 
> We can help with testing for AMD hardware. Once the CUDA impl works, its mostly a copy-paste from CUDA to HIP, so I'm not seeing that as a big vulnerability to this development effort.

I'd like to do the GPU part in a followup PR, as long as this PR will remain accessible - you've already added good notes for how to start on the GPU work.

I'll take care of the changelog and the style fix, then start on the GPU tasks.

wostrie2:
Ok, I added a note to the CHANGELOG and made the style changes.

jeremylt:
Yup, you will be able to see this PR under the closed PRs. And you can @ me if you have more questions.

Thank you for your work here; it's a nice new feature. 🚀 

I'll merge once CI finishes

jeremylt (approved):

wostrie2:
> Yup, you will be able to see this PR under the closed PRs. And you can @ me if you have more questions.
> 
> Thank you for your work here; it's a nice new feature. 🚀
> 
> I'll merge once CI finishes

Thanks, @jeremylt! Happy to be able to work on this. Thank you for making it very approachable :)

PR #1955, opened 2026-04-19
SYCL: link libceed.so with SYCLCXX; fix JiT 1-byte heap overflow

pvelesko:
## Purpose

Two fixes for SYCL / JiT on Intel oneAPI (e.g. Aurora): link `libceed.so` with the SYCL-aware toolchain when `CC=gcc`, and fix a one-byte heap overflow in `CeedLoadSourceToInitializedBuffer` when JiT source has `#include` at the start of a file (or back-to-back includes), which showed up as heap corruption / `realloc(): invalid pointer` during SYCL JiT.

## 1. `make`: use `SYCLCXX` as linker for `libceed.so` when SYCL is enabled

When `CC=gcc`, `SYCL_FLAG` is empty so the `libceed.so` link step used `g++` without `-fsycl`. The linker does not merge SYCL device sections from `.sycl.cpp` objects, so the shared library only had host stubs and the SYCL backend failed at runtime with errors like **No kernel named …** (e.g. `CeedVectorNorm_Sycl`).

**Change:** If `SYCL_LIB_DIR` is set, link `libceed.so` with `$(SYCLCXX)` (e.g. `icpx`) and add `-fsycl` / `-fno-sycl-id-queries-fit-in-int` to `CEED_LDFLAGS`. **PKG_LIBS:** use `$(filter -fsycl …,$(SYCLFLAGS))` so `ceed.pc` carries `-fsycl` even when `CC` is not Intel.

## 2. `jit-tools`: fix 1-byte heap overflow in `CeedLoadSourceToInitializedBuffer`

On the `#pragma` and `#include` paths the code did `CeedRealloc(current_size + copy_size + 1, …)` but the next line is `memcpy(&(*buffer)[current_size], "\n", 2)`, which writes **two** bytes (the string literal includes the trailing `\0`).

When **`copy_size == 0`** (e.g. the first line of the file is `#include <…>`, or the next preprocessor line is another `#include` so `file_offset` already points at `#`), that allocation is **one byte short** → **heap buffer overflow** (one byte past the end of the `realloc` region).

**Fix:** Allocate `current_size + copy_size + 2` and place the final `'\0'` at `[current_size + copy_size + 1]`, matching the same pattern already used for the “rest of file” tail copy.

### AddressSanitizer repro (pre-fix)

With the buggy `+1` realloc, loading a minimal JiT file whose **first line** is an include, e.g.:

```c
#include <ceed/types.h>
#include <ceed/ceed-f32.h>
```

and calling `CeedLoadSourceToBuffer` under ASAN reports, for example:

```
ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 2 at ... thread T0
    #0 memcpy
    #1 CeedLoadSourceToInitializedBuffer .../ceed-jit-tools.c:212
    #2 CeedLoadSourceAndInitializeBuffer
    #3 CeedLoadSourceToBuffer
...
N-byte region [...] allocated by ... CeedReallocArray ... ceed.c:326
```

The fixed code completes successfully for the same input.

## Commits

- `make: use SYCLCXX as linker for libceed.so when SYCL is enabled`
- `jit-tools: fix 1-byte heap overflow in CeedLoadSourceToInitializedBuffer`

## LLM / GenAI disclosure

Pull request description drafted with assistance from an AI coding tool (Cursor). Commits and code changes Claude Opus 4.7

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -579,7 +579,7 @@ ifneq ($(SYCL_DIR),)
   SYCL_LIB_DIR := $(patsubst %/,%,$(dir $(firstword $(SYCL_LIB_DIR))))
 endif
 ifneq ($(SYCL_LIB_DIR),)
-  PKG_LIBS += $(SYCL_FLAG) -lze_loader
+  PKG_LIBS += $(filter -fsycl -fno-sycl-id-queries-fit-in-int,$(SYCLFLAGS)) -lze_loader
   LIBCEED_CONTAINS_CXX = 1
   libceed.sycl  += $(sycl-core.cpp) $(sycl-ref.cpp) $(sycl-shared.cpp) $(sycl-gen.cpp)
   BACKENDS_MAKE += $(SYCL_BACKENDS)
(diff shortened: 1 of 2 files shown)


pvelesko:
[aurora-make-prove-sycl.txt](https://github.com/user-attachments/files/26868800/aurora-make-prove-sycl.txt)

`make prove` run results from Aurora after these fixes

jeremylt (approved):

jeremylt (approved):

PR #1956, opened 2026-04-21
fix(examples): Update examples for latest PETSc changes

jrwrigh:
Purpose:

Update the examples to be compatible with the latest PETSc versions.

Note this will require updating the PETSc build on Noether.

LLM/GenAI Disclosure:

None


examples/solids/elasticity.c
```diff
@@ -507,7 +507,7 @@ int main(int argc, char **argv) {
         // ---------- Smoother KSP
         PetscCall(PCMGGetSmoother(pc, level, &ksp_smoother));
         PetscCall(KSPSetDM(ksp_smoother, level_dms[level]));
-        PetscCall(KSPSetDMActive(ksp_smoother, PETSC_FALSE));
+        PetscCall(KSPSetDMActive(ksp_smoother, KSP_DMACTIVE_ALL, PETSC_FALSE));
```
jrwrigh, line 510: @jeremylt @zatkins-dev I noticed that this change was not present on Ratel. From the [PETSc MR](https://gitlab.com/petsc/petsc/-/merge_requests/8990/), I believe this is the correct change, though let me know if not.

jeremylt: We made this change in Ratel and we did it the same  way

jeremylt (approved):
Working on the CI build update currently, will reply when ready

examples/solids/elasticity.c
jrwrigh: Whoops, looks like I hadn't pulled Ratel locally since early Feburary. That'd explain it.

jeremylt:
We also need to swap the test cases over to using `ts_time_step` instead of `ts_dt`

jeremylt (approved):

jeremylt:
Huh, new issue: https://gitlab.com/libceed/libCEED/-/jobs/14025365570

jrwrigh:
Looks like they're all CUDA backends. Did you revert the CUDA change in PETSc that we've been having to recycle?

```
commit 8176a7140471166274ad21eefd68452483cbaa22 (HEAD -> main)
Author: Jeremy <jeremy@jeremylt.org>
Date:   Mon Jun 16 17:14:11 2025 -0600

    Revert "CUVEC: fix copy from device to default to copy to device"

    This reverts commit 7ada2a5134eeb4113b6693316dd2660902f08a96.
```

We really should figure out what's going on there...

jeremylt:
ah crap - I ran the rebuild script without reverting that one commit

PR #1963, opened 2026-05-04
at-points: change at-points operators to use Chebyshev polynomials of the first kind (normalized)

zatkins-dev:
Purpose:

We're currently using Chebyshev polynomials of the second kind, which don't have their range bounded to the same [-1,1] as their domain. I'm really not sure why we would be using the second kind, since they're numerically less stable and have a more complicated derivative. 

This doesn't cause any issues with tests here or in Ratel.

LLM/GenAI Disclosure:

None

### DIFF
--- a/include/ceed/jit-source/cuda/cuda-ref-basis-tensor-at-points.h
+++ b/include/ceed/jit-source/cuda/cuda-ref-basis-tensor-at-points.h
@@ -15,21 +15,22 @@
 template <int Q_1D>
 inline __device__ void ChebyshevPolynomialsAtPoint(const CeedScalar x, CeedScalar *chebyshev_x) {
   chebyshev_x[0] = 1.0;
-  chebyshev_x[1] = 2 * x;
+  chebyshev_x[1] = x;
   for (CeedInt i = 2; i < Q_1D; i++) chebyshev_x[i] = 2 * x * chebyshev_x[i - 1] - chebyshev_x[i - 2];
 }
 
 template <int Q_1D>
 inline __device__ void ChebyshevDerivativeAtPoint(const CeedScalar x, CeedScalar *chebyshev_dx) {
-  CeedScalar chebyshev_x[3];
+  CeedScalar chebyshev_x[2];
 
-  chebyshev_x[1]  = 1.0;
-  chebyshev_x[2]  = 2 * x;
+  chebyshev_x[0]  = 1.0;
+  chebyshev_x[1]  = 2 * x;
   chebyshev_dx[0] = 0.0;
-  chebyshev_dx[1] = 2.0;
+  chebyshev_dx[1] = 1.0;
   for (CeedInt i = 2; i < Q_1D; i++) {
-    chebyshev_x[(i + 1) % 3] = 2 * x * chebyshev_x[(i + 0) % 3] - chebyshev_x[(i + 2) % 3];
-    chebyshev_dx[i]          = 2 * x * chebyshev_dx[i - 1] + 2 * chebyshev_x[(i + 0) % 3] - chebyshev_dx[i - 2];
+    // dT_i/dx = i * dU_{i-1}/dx
+    chebyshev_dx[i]    = i * chebyshev_x[(i + 1) % 2];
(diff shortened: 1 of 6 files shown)


zatkins-dev:
we could possibly squeeze an extra register out of the cuda and hip impls, since they use 3 doubles for the derivative calculation right now and we should only need two with this change.

interface/ceed-basis.c
```diff
@@ -47,7 +47,7 @@ const CeedBasis CEED_BASIS_NONE = &ceed_basis_none;
 **/
 static int CeedChebyshevPolynomialsAtPoint(CeedScalar x, CeedInt n, CeedScalar *chebyshev_x) {
   chebyshev_x[0] = 1.0;
-  chebyshev_x[1] = 2 * x;
+  chebyshev_x[1] = x;
```
jeremylt, line 52: Can you toss references to Chebyshev of the first kind for the future grad student trying to grok what we're doing here? (Also for the derivative function)

include/ceed/jit-source/cuda/cuda-ref-basis-tensor-at-points.h
```diff
@@ -15,21 +15,22 @@
...
 }
 
 template <int Q_1D>
 inline __device__ void ChebyshevDerivativeAtPoint(const CeedScalar x, CeedScalar *chebyshev_dx) {
-  CeedScalar chebyshev_x[3];
+  CeedScalar chebyshev_x[2];
 
-  chebyshev_x[1]  = 1.0;
-  chebyshev_x[2]  = 2 * x;
+  chebyshev_x[0]  = 1.0;
```
jeremylt, line 26: specifically we should note here that the second kind is used to make the derivatives of the first kind (probably why I stuck with the second kind throughout? I don't recall)

jeremylt: (I know you have it there mathematically, but I'm trying to picture the theoretical new grad student)

interface/ceed-basis.c
zatkins-dev: Yeah for sure, though tbh the options are wikipedia or a book (or both)

interface/ceed-basis.c
```diff
@@ -47,7 +47,7 @@ const CeedBasis CEED_BASIS_NONE = &ceed_basis_none;
 **/
 static int CeedChebyshevPolynomialsAtPoint(CeedScalar x, CeedInt n, CeedScalar *chebyshev_x) {
   chebyshev_x[0] = 1.0;
```
zatkins-dev, line 51:
```suggestion
  // Chebyshev polynomials of the first kind, 3 term recurrence: T_0(x) = 1, T_1(x) = x, T_n = 2 x T_{n-1} - T_{n-2}
  // For more info, see e.g. https://en.wikipedia.org/wiki/Chebyshev_polynomials#Recurrence_definition
  chebyshev_x[0] = 1.0;
```

interface/ceed-basis.c
```diff
@@ -69,12 +69,13 @@ static int CeedChebyshevDerivativeAtPoint(CeedScalar x, CeedInt n, CeedScalar *c
   chebyshev_x[1]  = 1.0;
   chebyshev_x[2]  = 2 * x;
```
zatkins-dev, line 74:
```suggestion
  // Note, these are Chebyshev polynomials of the 2nd kind, used for derivatives
  // See e.g. https://en.wikipedia.org/wiki/Chebyshev_polynomials#Differentiation_and_integration for the recurrence
  chebyshev_x[1]  = 1.0;
  chebyshev_x[2]  = 2 * x;
```

interface/ceed-basis.c
jeremylt: Yea, Wikipedia is probably good for a quick reference to get people oriented, and I like the refs being here in this file. That makes the most sense I think

jeremylt (approved):

jeremylt:
~~Checked CI locally - happy to merge if you want me to~~

~~Ope, I have a CUDA build issue.~~

Nevermind, that's the CUDA/GCC-16 issue, all is good

zatkins-dev:
I think I'm good with merging!

PR #1969, opened 2026-05-21
HIP: Add support for HIP 7+

zatkins-dev:
Purpose:

Update libCEED to support ROCm 7+

Specific changes:
- We need to link the `-lhiprtc` library explicitly now
- We can no longer include `<stddef.h>, which led to
  - Change `NULL` to `nullptr` in JiT sources
  - Pull `int32_t` and `int64_t` from the `__hip_internal` namespace
- Due to the above, I noticed an untested edge case for assembly that could cause seg faults

Closes: NA

LLM/GenAI Disclosure:

I used Gemini to generate the cut command in the makefile, pretty embarrassing in hindsight.

### DIFF
--- a/backends/hip-gen/ceed-hip-gen-operator-build.cpp
+++ b/backends/hip-gen/ceed-hip-gen-operator-build.cpp
@@ -302,7 +302,7 @@ static int CeedOperatorBuildKernelFieldData_Hip_gen(std::ostringstream &code, Ce
 
         CeedCallBackend(CeedBasisIsCollocated(basis, &is_collocated));
         if ((is_active && skip_active_load) || (is_collocated && !is_at_points)) {
-          code << tab << "CeedScalar *s_B" << var_suffix << " = NULL;\n";
+          code << tab << "CeedScalar *s_B" << var_suffix << " = nullptr;\n";
         } else {
           code << tab << "__shared__ CeedScalar s_B" << var_suffix << "[" << P_name << "*" << Q_name << "];\n";
           code << tab << "LoadMatrix<" << P_name << ", " << Q_name << ">(data, B." << option_name << "[" << i << "], s_B" << var_suffix << ");\n";
@@ -341,7 +341,7 @@ static int CeedOperatorBuildKernelFieldData_Hip_gen(std::ostringstream &code, Ce
 
           CeedCallBackend(CeedBasisIsCollocated(basis, &is_collocated));
           if ((is_active && skip_active_load) || (is_collocated && !is_at_points)) {
-            code << tab << "CeedScalar *s_B" << var_suffix << " = NULL;\n";
+            code << tab << "CeedScalar *s_B" << var_suffix << " = nullptr;\n";
           } else {
             code << tab << "__shared__ CeedScalar s_B" << var_suffix << "[" << P_name << "*" << Q_name << "];\n";
             code << tab << "LoadMatrix<" << P_name << ", " << Q_name << ">(data, B." << option_name << "[" << i << "], s_B" << var_suffix << ");\n";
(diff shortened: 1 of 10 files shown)


interface/ceed-preconditioning.c
```diff
@@ -712,18 +712,19 @@ static int CeedOperatorAssembleSymbolicSingle(CeedOperator op, CeedSize offset,
   Note: If the value of `assembled` or `rstr` passed to this function are non-`NULL` , then it is assumed that they hold valid pointers.
         These objects will be destroyed if `*assembled` or `*rstr` is the only reference to the object.
 
-  @param[in]  op         `CeedOperator` to assemble `CeedQFunction`
-  @param[in]  use_parent Boolean flag to check for fallback parent implementation
-  @param[out] assembled  `CeedVector` to store assembled `CeedQFunction` at quadrature points
-  @param[out] rstr       `CeedElemRestriction` for `CeedVector` containing assembled `CeedQFunction`
-  @param[in]  request    Address of @ref CeedRequest for non-blocking completion, else @ref CEED_REQUEST_IMMEDIATE
+  @param[in]  op            `CeedOperator` to assemble `CeedQFunction`
```
jeremylt, line 715: note to us - we should see if we can lint for the correct alignment here automatically

backends/hip-gen/ceed-hip-gen-operator-build.cpp
```diff
@@ -302,7 +302,7 @@ static int CeedOperatorBuildKernelFieldData_Hip_gen(std::ostringstream &code, Ce
 
         CeedCallBackend(CeedBasisIsCollocated(basis, &is_collocated));
         if ((is_active && skip_active_load) || (is_collocated && !is_at_points)) {
-          code << tab << "CeedScalar *s_B" << var_suffix << " = NULL;\n";
+          code << tab << "CeedScalar *s_B" << var_suffix << " = nullptr;\n";
```
jeremylt, line 305: \<snark> Thanks C++ for *two* sorts of `NULL` \</snark>

zatkins-dev: yep, so fun

jeremylt: lol, I put in pesudo HTML tags of `<snark>` and `</snark>` and GitHub ate them

backends/hip/ceed-hip-compile.cpp
```diff
@@ -120,12 +120,16 @@ static int CeedCompileCore_Hip(Ceed ceed, const char *source, const char *name,
     CeedCallHip(ceed, hipRuntimeGetVersion(&runtime_version));
     if (runtime_version < 40400000) {
       code << "#include <hip/hip_runtime.h>\n\n";
-    }
-    // With ROCm 4.5, need to include these definitions specifically for hiprtc (but cannot include the runtime header)
-    else {
+    } else if (runtime_version < 70000000) {
+      // With ROCm 4.5, need to include these definitions specifically for hiprtc (but cannot include the runtime header)
```
jeremylt: Should clarify the version range for those not used to their system of numbering
```suggestion
      // With ROCm 4.5 and onwards, need to include these definitions specifically for hiprtc (but cannot include the runtime header)
```

backends/hip/ceed-hip-compile.cpp
```diff
@@ -120,12 +120,16 @@ static int CeedCompileCore_Hip(Ceed ceed, const char *source, const char *name,
...
       code << "#include <stddef.h>\n";
       code << "#define __forceinline__ inline __attribute__((always_inline))\n";
       code << "#define HIP_DYNAMIC_SHARED(type, var) extern __shared__ type var[];\n\n";
+    } else {
+      code << "using __hip_internal::int32_t;\n";
```
jeremylt, line 130: For my curiosity, why swapping stddef for these two 'using' statements?

zatkins-dev: stddef is no longer found, it just causes a compilation error

jeremylt: that's an odd choice from the ROCm team

backends/hip/ceed-hip-compile.cpp
jeremylt: really, this commenting is confused a bit by the repetition of the two defines.... maybe else clause with a sub-if delineating stddef vs the two using statements would be clearer to the reader?

backends/hip/ceed-hip-compile.cpp
zatkins-dev: Everything about ROCm 7 has been super odd to me, including not packaging support for cards that their libraries still claim to support (though given the libraries take like 30 GB now, I can see the reasoning)

jeremylt: I'm sure this will be made better by the tendency of LLM generated code to bloat the total project size

backends/hip/ceed-hip-compile.cpp
jeremylt: Unresolving with the above comment

zatkins-dev: how about this?

jeremylt (approved):

PR #1970, opened 2026-05-22
hip(compile) - add definition of NULL for user codes

zatkins-dev:
Purpose:

Minor fix -- user codes should be able to use `NULL`.

Closes: NA

LLM/GenAI Disclosure:

None


jeremylt (approved):

jeremylt:
uggggghhhhhh




thank you

zatkins-dev:
I noticed the issue when I got Ratel building with ROCm 7 (woo!)

PR #1973, opened 2026-05-26
format - fix regressions due to clang-format-22, enforce braces around if-else

zatkins-dev:
Purpose:
This MR adds enforcement of the if-else block formatting requirements (single line if statement with no braces is okay, if-else must always have braces).

Right now we have if-else blocks with no braces, which is something we intentionally tried to rid of. This is a tricky fix, as it is hard to both allow single line if statements while disallowing single line if-else blocks AND to require braces around the if and else blocks in the latter case. The combination that works is adding `AllowShortIfStatementsOnASingleLine: WithoutElse` to the `.clang-format` config and using `clang-tidy --fix` with the `readability-braces-around-statements` check and the option `readability-braces-around-statements.ShortStatementLines` set to `1`. This combo allows single line if statements with no braces, but automatically adds braces for any if-else blocks.

Closes: NA

LLM/GenAI Disclosure:

None


zatkins-dev:
@jeremylt this one was a lot weirder to get right, and we're not going to be able to tidy the JiT source. I updated all existing format violations, future cases will have to be manually checked.

jeremylt:
What's the issue with tidying the JiT souce? (so we know for the future)

zatkins-dev:
> What's the issue with tidying the JiT souce? (so we know for the future)

clang-tidy has to be able to compile the file in order to parse it, and due to that being device code (and needing various define statements) that isn't possible for the JiT source.

jeremylt:
Perfect - can you note in the dev notes which files must be manually checked and why? Gotta leave those breadcrumbs for future us

jeremylt (approved):
This all looks good to me. If CI is happy, I'm happy

zatkins-dev:
I think it's just codecov, which I'm not surprised by

PR #1968, opened 2026-05-20
Magma use ceed includes

jeremylt:
Purpose:

This PR updates the Magma backend(s) to use includes to compile the basis kernels instead of reading the contents of these files to a string buffer.

Closes: #N/A

LLM/GenAI Disclosure:

None

By submitting this PR, the author certifies to its contents as described by the [Developer's Certificate of Origin](https://developercertificate.org/).
Please follow the [Contributing Guidelines](https://github.com/CEED/libCEED/blob/main/CONTRIBUTING.md) for all PRs.

PR #1974, opened 2026-05-27
Magma include: use using dim instead of num_comp

Sbozzolo:
Found in
https://github.com/awslabs/palace/actions/runs/26517671781/job/78111329701?pr=738

PR #1968 changes the `snprintf` to use `num_comp` instead of `dim` as a suffix, so that our 1D fields try to load the 3D headers


jeremylt:
Ack, thank you for this too.

@zatkins-dev to do for me is to make Magma-CUDA in CI, and to document the CI update stuff for you

jeremylt (approved):

zatkins-dev:
Ah, yes I noticed this yesterday but forgot to make an MR, thanks!

jeremylt:
Update - added Magma back to our CI, so this should be caught in our dev workflow now

PR #1980, opened 2026-06-05
Fix strcmp namespace in HIP

NamjaeChoi:
`strcmp` requires `std` namespace when used from `cstdlib` header. It fails to compile with the latest hipcc.


jeremylt:
which version of ROCm, for posterity?

jeremylt (approved):

NamjaeChoi:
We are currently testing 7.2.4.

jeremylt:
I think @zatkins-dev 7.2 working well but 7.3 currently is buggy with the same code?

NamjaeChoi:
Additional information that may be helpful for clarity is that we are building libCEED through PETSc.

zatkins-dev:
> I think @zatkins-dev 7.2 working well but 7.3 currently is buggy with the same code?

At least on GFX906, 7.2+ cause a ton of test failures. 7.1.x is the last known good version.

zatkins-dev:
This fix is totally reasonable though, we should fix namespacing issues

NamjaeChoi:
Actually the code compiles as is. Sorry for a confusion, even though the change isn't harmful. I realized that the libCEED version PETSc is pointing is missing `#include <cstring>`.

jeremylt:
Ah the correct fix is probably to update the libCEED hash PETSc is using

PR #1981, opened 2026-06-07
fluids - change HIP backends to use VECHIP and MATAIJHIPSPARSE by default

zatkins-dev:
Purpose:

We used Kokkos in the past, but PETSc now has good HIP support.

Closes: NA

LLM/GenAI Disclosure:

None


jeremylt (commented):
Do Solids and the PETSc examples need this too?

zatkins-dev:
> Do Solids and the PETSc examples need this too?

Nope, I think both of those were fixed back when we were in Oregon.

jeremylt (approved):

PR #1982, opened 2026-06-08
gpu(hip) - fix compilation error in gen at-points diagonal assembly

zatkins-dev:
Purpose:

Fixes a minor bug in HIP gen backends which prevented compilation.
We need some way to catch these kind of bugs.

Closes: NA

LLM/GenAI Disclosure:

None


jeremylt (commented):
are both e and elem supposed to be defined at this point?

zatkins-dev:
> are both e and elem supposed to be defined at this point?

Not sure. The loop is over elem.

jeremylt:
hmm, this should be compared to the operator application kernel to make sure we are not breaking HipStar support

zatkins-dev:
Okay @jeremylt, I think this is the right fix for ChipStar support

jeremylt (approved):

PR #1987, opened 2026-07-08
cuda-gen: Add CUDA Graph capture and replay for composite operators

Nafees01:
## Summary

Adds CUDA Graph capture and replay to the 'cuda-gen' backend for composite operators. On the second apply, the kernel sequence is captured into a cudaGraph; all subsequent applies replay the instantiated graph. QFunction contexts are re-synced to device before each replay so time- and load-dependent parameters stay correct.

Sub-operators that cannot be captured, for example, contact operators that allocate memory (cudaMalloc) during apply, which CUDA doesn't allow inside a capture. When this happens, the capture is properly closed to leave the stream in a clean state, any leftover CUDA error is cleared, and that operator falls back to running normally on every call.

Validated against both /gpu/cuda/ref and cuda-gen without graphs across static, quasistatic, contact, multi-material, and dynamic elasticity examples. Strain energy and displacements match to ~1e-13 in every case, with zero graph fallbacks on capturable operators.

Performance-wise, graphs are within ±1% of cuda-gen without graphs across all sizes, so no regression, but no real speedup either. The reason is that cuda-gen already fuses everything into a few large kernels, so there's very little launch overhead left for graphs to remove. 

Validated correctness against /gpu/cuda/ref and cuda-gen (CEED_DISABLE_GRAPH=1) on the following Ratel examples:
- ex01-static-elasticity-linear-mms
- ex02-quasistatic-elasticity-linear-platen (contact)
- ex02-quasistatic-elasticity-neo-hookean-current-sinker (multi-material)
- ex03-dynamic-elasticity-mooney-rivlin-current
Strain energy matches to ~1e-13 in all cases. Zero graph fallbacks on capturable operators.


zatkins-dev:
> Sub-operators that cannot be captured, for example, contact operators that allocate memory (cudaMalloc) during apply

Are there specific operators you were unable to use in Ratel? None of the QFunctions there should alloc during operator application.

zatkins-dev:
There's some style errors here, be sure to `make format` each commit

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    disable_graph_checked = true;
+  }
+  if (disable_graph) {
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      CeedCallBackend(CeedOperatorApplyAdd(sub_operators[i], input_vec, output_vec, request));
     }
+    return CEED_ERROR_SUCCESS;
   }
-  if (is_sequential) CeedCallCuda(ceed, cudaStreamDestroy(stream));
-  if (input_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorRestoreArrayRead(input_vec, &input_arr));
```
zatkins-dev, line 330: This removed the `is_sequential` checks, which are there to ensure that all suboperators run in a single stream.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    return CEED_ERROR_SUCCESS;
+  }
 
-    CeedCall(CeedOperatorGetNumElements(sub_operators[i], &num_elem));
-    if (num_elem > 0) {
-      if (!is_sequential) CeedCallCuda(ceed, cudaStreamCreate(&stream));
-      CeedCallBackend(CeedOperatorApplyAddCore_Cuda_gen(sub_operators[i], stream, input_arr, output_arr, &is_run_good[i], request));
-      if (!is_sequential) CeedCallCuda(ceed, cudaStreamDestroy(stream));
+  // CEED_DISABLE_GRAPH=1: run suboperators directly, no capture/replay (for benchmarking).
+  static bool disable_graph = false;
```
zatkins-dev: This should be stored on a per-operator (or at least per-`Ceed`) basis, not as static variables. It's okay to use the environment variable to set a default value, but there should be a setter (e.g. `CeedSetUseCudaGraph(Ceed ceed, bool use_graph)`)

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
-  if (input_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorRestoreArrayRead(input_vec, &input_arr));
-  if (output_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorRestoreArray(output_vec, &output_arr));
-  CeedCallCuda(ceed, cudaDeviceSynchronize());
 
-  // Fallback on unsuccessful run
-  for (CeedInt i = 0; i < num_suboperators; i++) {
-    if (!is_run_good[i]) {
-      CeedOperator op_fallback;
+  // No real I/O buffers to capture; just run directly.
+  if (input_vec == CEED_VECTOR_NONE || output_vec == CEED_VECTOR_NONE) {
```
zatkins-dev: Should this be?
```suggestion
  if (input_vec == CEED_VECTOR_NONE && output_vec == CEED_VECTOR_NONE) {
```
It's possible to have a passive input and active output

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+  static bool disable_graph = false;
+  static bool disable_graph_checked = false;
+  if (!disable_graph_checked) {
+    char *env_val = getenv("CEED_DISABLE_GRAPH");
+    disable_graph = (env_val != NULL && strcmp(env_val, "1") == 0);
+    disable_graph_checked = true;
+  }
+  if (disable_graph) {
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      CeedCallBackend(CeedOperatorApplyAdd(sub_operators[i], input_vec, output_vec, request));
```
zatkins-dev: This call pattern prevents each suboperator from running on separate streams. You won't see a perf difference until you have large, multioperator applications, but the perf difference can be significant.

zatkins-dev: The disabled graph case should restore the previous implementation, with multi-stream operators, sequential checks, etc

zatkins-dev: Given you do this a bunch of places, I'd recommend a helper function which looks like:
```c
static int CeedOperatorApplyAddComposite_NoGraph_Cuda_gen(CeedOperator op, CeedVector input_vec, CeedVector output_vec, CeedRequest *request) {
  bool              is_run_good[CEED_COMPOSITE_MAX] = {false}, is_sequential;
  CeedInt           num_suboperators;
  const CeedScalar *input_arr  = NULL;
  CeedScalar       *output_arr = NULL;
  Ceed              ceed;
  CeedOperator     *sub_operators;
  cudaStream_t      stream = NULL;

  CeedCall(CeedOperatorCompositeGetNumSub(op, &num_suboperators));
  CeedCall(CeedOperatorCompositeGetSubList(op, &sub_operators));
  CeedCall(CeedOperatorCompositeIsSequential(op, &is_sequential));
  if (input_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArrayRead(input_vec, CEED_MEM_DEVICE, &input_arr));
  if (output_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArray(output_vec, CEED_MEM_DEVICE, &output_arr));
  if (is_sequential) CeedCallCuda(ceed, cudaStreamCreate(&stream));
  for (CeedInt i = 0; i < num_suboperators; i++) {
    CeedInt num_elem = 0;

    CeedCall(CeedOperatorGetNumElements(sub_operators[i], &num_elem));
    if (num_elem > 0) {
      if (!is_sequential) CeedCallCuda(ceed, cudaStreamCreate(&stream));
      CeedCallBackend(CeedOperatorApplyAddCore_Cuda_gen(sub_operators[i], stream, input_arr, output_arr, &is_run_good[i], request));
      if (!is_sequential) CeedCallCuda(ceed, cudaStreamDestroy(stream));
    }
  }
  if (is_sequential) CeedCallCuda(ceed, cudaStreamDestroy(stream));
  if (input_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorRestoreArrayRead(input_vec, &input_arr));
  if (output_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorRestoreArray(output_vec, &output_arr));
  CeedCallCuda(ceed, cudaDeviceSynchronize());

  // Fallback on unsuccessful run
  for (CeedInt i = 0; i < num_suboperators; i++) {
    if (!is_run_good[i]) {
      CeedOperator op_fallback;

      CeedDebug(ceed, "\nFalling back to /gpu/cuda/ref CeedOperator for ApplyAdd\n");
      CeedCallBackend(CeedOperatorGetFallback(sub_operators[i], &op_fallback));
      CeedCallBackend(CeedOperatorApplyAdd(op_fallback, input_vec, output_vec, request));
    }
  }
  CeedCallBackend(CeedDestroy(&ceed));
  return CEED_ERROR_SUCCESS;
}
```

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
-  CeedCallCuda(ceed, cudaDeviceSynchronize());
 
-  // Fallback on unsuccessful run
-  for (CeedInt i = 0; i < num_suboperators; i++) {
-    if (!is_run_good[i]) {
-      CeedOperator op_fallback;
+  // No real I/O buffers to capture; just run directly.
+  if (input_vec == CEED_VECTOR_NONE || output_vec == CEED_VECTOR_NONE) {
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      CeedCallBackend(CeedOperatorApplyAdd(sub_operators[i], input_vec, output_vec, request));
```
zatkins-dev: Need to use the same helper here

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    }
+    return CEED_ERROR_SUCCESS;
+  }
+
+
+  // Phase 1: first call runs directly so lazy allocations happen before capture.
+  if (!impl->warmup_done) {
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      CeedCallBackend(CeedOperatorApplyAdd(sub_operators[i], input_vec, output_vec, request));
```
zatkins-dev: Should use `CeedOperatorApplyAddCore_Cuda_gen` directly, we intentionally don't go through the interface here

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    if (err != cudaSuccess) {
+      capture_ok = false;
+    }
+
+    if (capture_ok) {
+      // Check errors by hand (not CeedCallBackend) so we always reach the
+      // cudaStreamEndCapture below. Some sub-operators (e.g. contact, which calls
+      // cudaMalloc) invalidate the capture, and we still need to close it out.
+      for (CeedInt i = 0; i < num_suboperators; i++) {
+        if (CeedOperatorApplyAdd(sub_operators[i], input_vec, output_vec, CEED_REQUEST_IMMEDIATE)) {
```
zatkins-dev: `CeedOperatorApplyAddCore_Cuda_gen` has a `bool *is_run_good` output parameter for exactly this purpose.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    impl->graph_launches = 0;
+
+    if (capture_ok) printf("[CUDA Graph] Graph created for operator '%s'\n", op_name ? op_name : "unnamed");
+    else printf("[CUDA Graph] Capture disabled for operator '%s'; falling back to direct apply\n", op_name ? op_name : "unnamed");
+
+    // Capture doesn't run the kernels, so apply directly to get this call's output.
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      CeedCallBackend(CeedOperatorApplyAdd(sub_operators[i], input_vec, output_vec, request));
+    }
+    return CEED_ERROR_SUCCESS;
```
zatkins-dev:
```suggestion
```

Can we just play out the graph (i.e. don't return, exit if statement, continue) instead of applying the kernels for a second time?

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -885,6 +1070,16 @@ int CeedOperatorCreate_Cuda_gen(CeedOperator op) {
...
   CeedCallBackend(CeedOperatorSetData(op, impl));
+  
+  impl->graph_created            = false;
+  impl->warmup_done              = false;
+  impl->graph                    = NULL;
+  impl->graph_instance           = NULL;
+  impl->graph_launches           = 0;
+  impl->fallbacks                = 0;
+  impl->captured_input_ptr       = NULL;
+
```
zatkins-dev:
```suggestion
```

All of these assignments are implicitly done via `CeedCalloc(1, &impl);`

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    }
+
+    cudaError_t err = cudaGraphLaunch(impl->graph_instance, stream);
+    if (err != cudaSuccess) {
+      printf("CUDA Graph launch failed: %s - using fallback\n", cudaGetErrorString(err));
+      CeedOperator op_fallback;
+      CeedCallBackend(CeedOperatorGetFallback(op, &op_fallback));
       CeedCallBackend(CeedOperatorApplyAdd(op_fallback, input_vec, output_vec, request));
+      return CEED_ERROR_SUCCESS;
     }
```
zatkins-dev: We shouldn't fall back here -- we should call the helper I mentioned above and mark the graph as bad. The fallback should be a last case scenario.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    cudaGraphDestroy(impl->graph);
+    impl->graph = NULL;
+  }
+  impl->graph_created      = false;
+  impl->captured_input_ptr = NULL;
+  impl->fallbacks++;
+
+  CeedOperator op_fallback;
+  CeedCallBackend(CeedOperatorGetFallback(op, &op_fallback));
+  CeedCallBackend(CeedOperatorApplyAdd(op_fallback, input_vec, output_vec, request));
```
zatkins-dev: Same here -- use the helper, not fall back.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    for (CeedInt i = 0; i < num_suboperators; i++) {
+      CeedCallBackend(CeedOperatorApplyAdd(sub_operators[i], input_vec, output_vec, request));
+    }
+    return CEED_ERROR_SUCCESS;
+  }
+
+  // Phase 3: replay.
+  {
+    // graph_instance == NULL means capture failed; direct-apply every call.
+    if (!impl->graph_instance) {
```
zatkins-dev: Use helper here

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    // If the input buffer moved since capture, recapture.
+    {
+      const CeedScalar *in_ptr;
+      bool              ptr_ok;
+
+      CeedCallBackend(CeedVectorGetArrayRead(input_vec, CEED_MEM_DEVICE, &in_ptr));
+      ptr_ok = (in_ptr == impl->captured_input_ptr);
+      CeedCallBackend(CeedVectorRestoreArrayRead(input_vec, &in_ptr));
+      if (!ptr_ok) goto use_fallback;
+    }
```
zatkins-dev: This check seems like it should happen before the graph is built, that way you can rebuild the graph if needed instead of waiting until the next apply call

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    CeedQFunction qf  = NULL;
+    void         *d_c = NULL;
+
+    CeedCallBackend(CeedOperatorGetQFunction(sub_operators[i], &qf));
+    CeedCallBackend(CeedQFunctionGetInnerContextData(qf, CEED_MEM_DEVICE, &d_c));
+    CeedCallBackend(CeedQFunctionRestoreInnerContextData(qf, &d_c));
+    CeedCallBackend(CeedQFunctionDestroy(&qf));
+  }
+  return CEED_ERROR_SUCCESS;
+}
```
zatkins-dev: This same issue is going to happen for any passive inputs which have been updated on the CPU -- there should be a more robust way to handle this

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,51 +318,216 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+    char *env_val = getenv("CEED_FORCE_BASELINE");
+    force_baseline = (env_val != NULL && strcmp(env_val, "1") == 0);
+    force_baseline_checked = true;
+  }
+  if (force_baseline) {
+    CeedOperator op_fallback;
+    CeedCallBackend(CeedOperatorGetFallback(op, &op_fallback));
+    CeedCallBackend(CeedOperatorApplyAdd(op_fallback, input_vec, output_vec, request));
+    return CEED_ERROR_SUCCESS;
+  }
```
zatkins-dev: This isn't the right way to benchmark against `ref` -- use the `/gpu/cuda/ref` backend directly.
```suggestion
```

zatkins-dev (changes requested):
I'm sure you're still working on this, but here's an initial review to correct some high-level design issues and removed functionality that I think should be fixed.

zatkins-dev:
I would be interested to see if this speeds up e.g. /gpu/cuda/ref, when recording the graph over the entire `CeedOperatorApplyCore_Cuda_Ref` function

jeremylt:
Note - merging main into your branch instead of rebasing means I'd recommend we squash-merge this PR

Nafees01:
> > Sub-operators that cannot be captured, for example, contact operators that allocate memory (cudaMalloc) during apply
> 
> Are there specific operators you were unable to use in Ratel? None of the QFunctions there should alloc during operator application.

The only example where the contact (Nitsche) sub-operator couldn't be captured was the quasistatic contact case (ex02-quasistatic-elasticity-linear-platen). Everything else I tried static, dynamic, multi-material, neo-Hookean, Mooney-Rivlin captured and replayed fine.     

And you're right that the QFunction doesn't allocate; the cudaMalloc comes from libCEED's own device-side handling of the contact data, not the QFunction itself. It runs either way correctly; the contact operators just fall back to a normal apply rather than being replayed.

zatkins-dev:
> > > Sub-operators that cannot be captured, for example, contact operators that allocate memory (cudaMalloc) during apply
> 
> > 
> 
> > Are there specific operators you were unable to use in Ratel? None of the QFunctions there should alloc during operator application.
> 
> 
> 
> The only example where the contact (Nitsche) sub-operator couldn't be captured was the quasistatic contact case (ex02-quasistatic-elasticity-linear-platen). Everything else I tried static, dynamic, multi-material, neo-Hookean, Mooney-Rivlin captured and replayed fine.     
> 
> 
> 
> And you're right that the QFunction doesn't allocate; the cudaMalloc comes from libCEED's own device-side handling of the contact data, not the QFunction itself. It runs either way correctly; the contact operators just fall back to a normal apply rather than being replayed.
> 
> 

This is a pretty big red flag to me. The Nitsche contact operators don't actually do anything out of the ordinary on the libCEED side. Them failing to work means that something about this setup (very possibly something with passive inputs and outputs) is violating the libCEED interface assumptions.

Nafees01:
> > > > Sub-operators that cannot be captured, for example, contact operators that allocate memory (cudaMalloc) during apply
> > 
> > 
> > > 
> > 
> > 
> > > Are there specific operators you were unable to use in Ratel? None of the QFunctions there should alloc during operator application.
> > 
> > 
> > The only example where the contact (Nitsche) sub-operator couldn't be captured was the quasistatic contact case (ex02-quasistatic-elasticity-linear-platen). Everything else I tried static, dynamic, multi-material, neo-Hookean, Mooney-Rivlin captured and replayed fine.
> > And you're right that the QFunction doesn't allocate; the cudaMalloc comes from libCEED's own device-side handling of the contact data, not the QFunction itself. It runs either way correctly; the contact operators just fall back to a normal apply rather than being replayed.
> 
> This is a pretty big red flag to me. The Nitsche contact operators don't actually do anything out of the ordinary on the libCEED side. Them failing to work means that something about this setup (very possibly something with passive inputs and outputs) is violating the libCEED interface assumptions.

Thanks for flagging this. On re-checking linear-platen with CEED_DEBUG=1, the contact operators appear to fail earlier than graph capture: cuda-gen hits an NVRTC compile error (GradTransposeTensor3dFlattened defined twice in cuda-shared-basis-tensor-flattened-templates.h) and then falls back to /gpu/cuda/ref. So this looks like a pre-existing cuda-gen JIT issue triggered by the contact operator's basis configuration, rather than something specific to Nitsche or CUDA graphs.

zatkins-dev:
> On re-checking linear-platen with CEED_DEBUG=1, the contact operators appear to fail earlier than graph capture: cuda-gen hits an NVRTC compile error (GradTransposeTensor3dFlattened defined twice in cuda-shared-basis-tensor-flattened-templates.h) and then falls back to /gpu/cuda/ref

Sweet, good catch. I'll fix that separately.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Good catch, restored. The is_sequential handling now lives in CeedOperatorApplyAddComposite_NoGraph_Cuda_gen, which is the shared non-graph path for composite operators.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Addressed. Removed the static disable_graph variable. Graph enable/disable is now stored per-operator as impl->use_graph in CeedOperator_Cuda_gen.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Yes, good catch; updated to &&. We only skip graph capture when both input_vec and output_vec are CEED_VECTOR_NONE.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: > The disabled graph case should restore the previous implementation, with multi-stream operators, sequential checks, etc

Restored. The graph-disabled path no longer loops over suboperators with the public CeedOperatorApplyAdd interface. It now goes through CeedOperatorApplyAddComposite_NoGraph_Cuda_gen, which brings back the original multi-stream behavior: is_sequential composites run on one shared stream, non-sequential composites get per-suboperator streams for overlap, and we call CeedOperatorApplyAddCore_Cuda_gen directly with per-suboperator /ref fallback only as a last resort. All non-graph routes (disable graph, passive-only composite, warmup, capture/launch failure) use this helper.

Nafees01: > Given you do this a bunch of places, I'd recommend a helper function which looks like:
> 
> ```c
> static int CeedOperatorApplyAddComposite_NoGraph_Cuda_gen(CeedOperator op, CeedVector input_vec, CeedVector output_vec, CeedRequest *request) {
>   bool              is_run_good[CEED_COMPOSITE_MAX] = {false}, is_sequential;
>   CeedInt           num_suboperators;
>   const CeedScalar *input_arr  = NULL;
>   CeedScalar       *output_arr = NULL;
>   Ceed              ceed;
>   CeedOperator     *sub_operators;
>   cudaStream_t      stream = NULL;
> 
>   CeedCallBackend(CeedOperatorGetCeed(op, &ceed));
>   CeedCall(CeedOperatorCompositeGetNumSub(op, &num_suboperators));
>   CeedCall(CeedOperatorCompositeGetSubList(op, &sub_operators));
>   CeedCall(CeedOperatorCompositeIsSequential(op, &is_sequential));
>   if (input_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArrayRead(input_vec, CEED_MEM_DEVICE, &input_arr));
>   if (output_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArray(output_vec, CEED_MEM_DEVICE, &output_arr));
>   if (is_sequential) CeedCallCuda(ceed, cudaStreamCreate(&stream));
>   for (CeedInt i = 0; i < num_suboperators; i++) {
>     CeedInt num_elem = 0;
> 
>     CeedCall(CeedOperatorGetNumElements(sub_operators[i], &num_elem));
>     if (num_elem > 0) {
>       if (!is_sequential) CeedCallCuda(ceed, cudaStreamCreate(&stream));
>       CeedCallBackend(CeedOperatorApplyAddCore_Cuda_gen(sub_operators[i], stream, input_arr, output_arr, &is_run_good[i], request));
>       if (!is_sequential) CeedCallCuda(ceed, cudaStreamDestroy(stream));
>     }
>   }
>   if (is_sequential) CeedCallCuda(ceed, cudaStreamDestroy(stream));
>   if (input_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorRestoreArrayRead(input_vec, &input_arr));
>   if (output_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorRestoreArray(output_vec, &output_arr));
>   CeedCallCuda(ceed, cudaDeviceSynchronize());
> 
>   // Fallback on unsuccessful run
>   for (CeedInt i = 0; i < num_suboperators; i++) {
>     if (!is_run_good[i]) {
>       CeedOperator op_fallback;
> 
>       CeedDebug(ceed, "\nFalling back to /gpu/cuda/ref CeedOperator for ApplyAdd\n");
>       CeedCallBackend(CeedOperatorGetFallback(sub_operators[i], &op_fallback));
>       CeedCallBackend(CeedOperatorApplyAdd(op_fallback, input_vec, output_vec, request));
>     }
>   }
>   CeedCallBackend(CeedDestroy(&ceed));
>   return CEED_ERROR_SUCCESS;
> }
> ```

Done; added CeedOperatorApplyAddComposite_NoGraph_Cuda_gen as suggested. It centralizes the non-graph composite path (is_sequential stream handling, core apply, sync, per-suboperator ref fallback). All routes that shouldn't use a graph go through it: graph disabled, passive-only composite, warmup, and capture/launch failure fallbacks.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Updated. This path now calls CeedOperatorApplyAddComposite_NoGraph_Cuda_gen instead of looping over suboperators with the public ApplyAdd interface. Same helper is used for the other non-graph routes.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Updated. Warmup now goes through CeedOperatorApplyAddComposite_NoGraph_Cuda_gen, which calls CeedOperatorApplyAddCore_Cuda_gen directly with is_run_good tracking the same as the original composite apply path. The capture loop also uses Core directly rather than the public interface.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Updated. The capture loop now calls CeedOperatorApplyAddCore_Cuda_gen directly with an is_run_good flag instead of going through the public ApplyAdd interface.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Done. After a successful capture, we no longer re-apply the suboperators; we set graph_created and fall through to cudaGraphLaunch, so the first post-capture call is served by replaying the graph rather than running the kernels a second time.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Removed the redundant zero-inits. CeedCalloc already zeroes the struct, so only use_graph is set explicitly (from CEED_DISABLE_GRAPH). Also dropped the graph_launches and fallbacks counters from the struct.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Removed whole-operator ref fallback there too; uses the NoGraph helper.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fixed. Routed through the NoGraph helper instead of whole-operator fallback.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fixed. Capture failure and other non-graph paths now use the NoGraph helper instead of the old Phase 3 inline direct-apply.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Done. Moved the input pointer check before graph build. If the device address changed, we set need_build and recapture immediately in the same apply instead of falling back and waiting until the next call.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Good point; extended pre-replay sync beyond QFunction contexts. CeedCompositeRefreshForReplay_Cuda_gen now touches each sub-operator's passive input vectors (and at-points coordinates) with GetArrayRead(CEED_MEM_DEVICE) / Restore, along with the existing context refresh.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Done. Removed CEED_FORCE_BASELINE. Ref benchmarks use -ceed /gpu/cuda/ref directly.

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -330,7 +398,7 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
   if (output_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorRestoreArray(output_vec, &output_arr));
   CeedCallCuda(ceed, cudaDeviceSynchronize());
 
-  // Fallback on unsuccessful run
+  // Fall back to /gpu/cuda/ref for any sub-operator that couldn't run here.
```
jeremylt:
```suggestion
  // Fall back to /gpu/cuda/ref for any sub-operator that couldn't run here
```

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -31,7 +33,19 @@ static int CeedOperatorDestroy_Cuda_gen(CeedOperator op) {
...
-  if (impl->points.num_per_elem) CeedCallCuda(ceed, cudaFree((void **)impl->points.num_per_elem));
+  if (impl->points.num_per_elem) CeedCallCuda(ceed, cudaFree((void *)impl->points.num_per_elem));
+
+  if (impl->graph_instance) {
+    cudaGraphExecDestroy(impl->graph_instance);
+    impl->graph_instance = NULL;
+  }
+  if (impl->graph) {
+    cudaGraphDestroy(impl->graph);
+    impl->graph = NULL;
```
jeremylt, line 44: these calls probably should be wrapped in `CeedCallCuda`, right?

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -344,6 +412,124 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
...
+    if (capture_ok && (err != cudaSuccess || !graph)) capture_ok = false;
+    if (capture_ok) {
+      impl->graph = graph;
+      if (cudaGraphInstantiate(&impl->graph_instance, impl->graph, 0) != cudaSuccess) {
+        CeedCallCuda(ceed, cudaGraphDestroy(impl->graph));
+        impl->graph = NULL;
+        capture_ok  = false;
+      }
+    } else if (graph) {
+      cudaGraphDestroy(graph);
```
jeremylt: same here

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -344,6 +412,124 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
...
+        bool is_run_good = true;
+
+        if (CeedOperatorApplyAddCore_Cuda_gen(sub_operators[i], capture_stream, input_arr, output_arr, &is_run_good, request) || !is_run_good) {
+          capture_ok = false;
+          break;
+        }
+      }
+    }
+
+    err = cudaStreamEndCapture(capture_stream, &graph);
```
jeremylt, line 482:
```suggestion
    err = cudaStreamBeginCapture(capture_stream, cudaStreamCaptureModeThreadLocal);
    if (err != cudaSuccess) capture_ok = false;
    if (capture_ok) {
      // Still call EndCapture if capture is invalidated mid-way.
      for (CeedInt i = 0; i < num_suboperators; i++) {
        bool is_run_good = true;

        if (CeedOperatorApplyAddCore_Cuda_gen(sub_operators[i], capture_stream, input_arr, output_arr, &is_run_good, request) || !is_run_good) {
          capture_ok = false;
          break;
        }
      }
    }
    err = cudaStreamEndCapture(capture_stream, &graph);

```

backends/cuda-ref/ceed-cuda-ref-qfunctioncontext.c
```diff
@@ -36,7 +36,16 @@ static inline int CeedQFunctionContextSyncH2D_Cuda(const CeedQFunctionContext ct
...
-  CeedCallCuda(ceed, cudaMemcpy(impl->d_data, impl->h_data, ctx_size, cudaMemcpyHostToDevice));
+
+  // Use async memcpy during CUDA Graph capture for compatibility
+  enum cudaStreamCaptureStatus capture_status;
+  cudaStreamIsCapturing(cudaStreamPerThread, &capture_status);
+  if (capture_status != cudaStreamCaptureStatusNone) {
+    CeedCallCuda(ceed, cudaMemcpyAsync(impl->d_data, impl->h_data, ctx_size, cudaMemcpyHostToDevice, cudaStreamPerThread));
+  } else {
+    CeedCallCuda(ceed, cudaMemcpy(impl->d_data, impl->h_data, ctx_size, cudaMemcpyHostToDevice));
+  }
```
jeremylt, line 48:
```suggestion
  // Use async memcpy during CUDA Graph capture for compatibility
  enum cudaStreamCaptureStatus capture_status;

  cudaStreamIsCapturing(cudaStreamPerThread, &capture_status);
  if (capture_status != cudaStreamCaptureStatusNone) {
    CeedCallCuda(ceed, cudaMemcpyAsync(impl->d_data, impl->h_data, ctx_size, cudaMemcpyHostToDevice, cudaStreamPerThread));
  } else {
    CeedCallCuda(ceed, cudaMemcpy(impl->d_data, impl->h_data, ctx_size, cudaMemcpyHostToDevice));
  }
```

backends/cuda-ref/ceed-cuda-ref-vector.c
```diff
@@ -326,7 +326,17 @@ static int CeedVectorSetValue_Cuda(CeedVector vec, CeedScalar val) {
   }
   if (impl->d_array) {
     if (val == 0) {
-      CeedCallCuda(CeedVectorReturnCeed(vec), cudaMemset(impl->d_array, 0, length * sizeof(CeedScalar)));
+      // Check if we're in CUDA Graph capture mode
+      enum cudaStreamCaptureStatus capture_status;
+      cudaStreamIsCapturing(cudaStreamPerThread, &capture_status);
+
+      if (capture_status != cudaStreamCaptureStatusNone) {
```
jeremylt:
```suggestion
      enum cudaStreamCaptureStatus capture_status;

      cudaStreamIsCapturing(cudaStreamPerThread, &capture_status);
      if (capture_status != cudaStreamCaptureStatusNone) {
```

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -885,6 +1071,13 @@ int CeedOperatorCreate_Cuda_gen(CeedOperator op) {
   CeedCallBackend(CeedOperatorGetCeed(op, &ceed));
   CeedCallBackend(CeedCalloc(1, &impl));
   CeedCallBackend(CeedOperatorSetData(op, impl));
+
+  {
+    const char *env_val = getenv("CEED_DISABLE_GRAPH");
+
+    impl->use_graph = !(env_val && !strcmp(env_val, "1"));
+  }
```
jeremylt: Generally not great if the only way to control something is via an env variable. I'd also add a way to set this per operator from user facing code in `ceed-operator.c` with `CeedOperatorSetCudaEnableCudaGraph(op, enable_graph)` or similar.

Q: Is this graph also valid to use with ref composite operators, or no?

Either way, I'd call `CeedOperatorSetCudaEnableCudaGraph` here in this block instead of directly modifying `impl->use_graph` here, and I'd also call it when capture fails.

Also, maybe it's a me thing, but I'd go with `CEED_ENABLE_CUDA_GRAPH 0` and `CEED_ENABLE_CUDA_GRAPH 1` and assume `CEED_ENABLE_CUDA_GRAPH 1` is the default?

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fixed.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Good catch; wrapped both destroys in `CeedCallCuda` in CeedOperatorDestroy

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fixed, wrapped in `CeedCallCuda` here as well.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Applied

backends/cuda-ref/ceed-cuda-ref-qfunctioncontext.c
Nafees01: Applied

backends/cuda-ref/ceed-cuda-ref-vector.c
Nafees01: Applied

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: **Programmatic control**: Added `CeedOperatorSetEnableCudaGraph(op, enable_graph)` in `include/ceed/cuda.h / interface/ceed-cuda.c` (same pattern as `CeedQFunctionSetCUDAUserFunction`). The cuda-gen backend registers a `SetEnableCudaGraph` hook on operator creation.

**Env var**: Switched from `CEED_DISABLE_GRAPH` to `CEED_ENABLE_CUDA_GRAPH`. Default is on (unset or any value other than `0/false`). `CEED_ENABLE_CUDA_GRAPH=0` disables graphs.

**Create / failure paths**: Operator create reads the env var and calls `CeedOperatorSetEnableCudaGraph` instead of setting `impl->use_graph` directly. Capture and launch failure also call `CeedOperatorSetEnableCudaGraph(op, false)` rather than touching the impl field.

**Ref composite operators**: Not in this PR. Graph capture/replay is only implemented for `/gpu/cuda/gen` composite ApplyAddComposite. `/gpu/cuda/ref` still uses the standard multi-kernel apply path with no graph support; adding graphs there would be a separate follow-up

**Verification**: On noether, `multi-density-force` with 50 TS steps (`-ts_time_step 0.02`): gen+graph 4.26 s, gen with `CEED_ENABLE_CUDA_GRAPH`=0 6.02 s, ref 6.15 s. Strain energy and displacement match across all three. CUDA graphs are **~29%** faster than no-graph gen (1.41×) and **~31%** faster than ref (1.44×).

jeremylt: Woof, that's a lengthy comment that looks LLM generated.  If it was, don't forget to disclose LLM usage!

I say woof because.... that's a lot to parse all at once, which makes it less good at communication. If you are and want to use LLMs to write comments, I would definitely try to make a human summary that is easier to understand to go with it. If its human written - same thing really, I'd make a small summary to preceed such a lengthy reply

interface/ceed-cuda.c
```diff
@@ -29,3 +29,22 @@ int CeedQFunctionSetCUDAUserFunction(CeedQFunction qf, CUfunction f) {
   }
   return CEED_ERROR_SUCCESS;
 }
+
+/**
+  @brief Enable or disable CUDA Graph capture/replay for a `CeedOperator`
```
jeremylt: We should probably note that this function (and the one above) are silently successful if the backend doesn't support this feature

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fair point, that was too much to read at once. I did try to keep it concise, but I kept adding summary bits and asked the LLM to clean up the phrasing afterward. From next time, I will use a short summary so that it's easy to understand. My bad on the length. 

Here is the short summary: added a real setter `CeedOperatorSetEnableCudaGraph` instead of hardcoding the env var check, so users/backends can turn graphs on/off directly. Renamed the env var to `CEED_ENABLE_CUDA_GRAPH` (on by default, 0 turns it off). Capture/launch failures now go through that same setter. Ref backend still isn't touched; that's a separate follow-up. Tested on noether with 50 steps of multi-density-force, graph is ~30% faster than both no-graph and ref, and results match across all three.

interface/ceed-cuda.c
Nafees01: Makes sense. Planning to add something like: if the backend does not support `CUfunction` pointers for QFunctions, the call succeeds without effect, and if the backend does not support CUDA Graphs for operators, the call succeeds without effect. Does that sound good?

backends/cuda-gen/ceed-cuda-gen-operator.c
jeremylt: sounds good, and much easier for me to understand than what the LLM did with your words ❤️

interface/ceed-cuda.c
jeremylt: Yea, that works for me. Maybe also something like 'When this function has no effect, it is logged in debugging output."

Nafees01: Done!

interface/ceed-cuda.c
```diff
@@ -14,6 +14,9 @@
 /**
   @brief Set CUDA function pointer to evaluate action at quadrature points
 
+  If the backend does not support `CUfunction` pointers for QFunctions, the call succeeds without effect.
+  When unsupported, a message is emitted via `CeedDebug`.
```
jeremylt:
```suggestion
  @brief Set CUDA function pointer to evaluate action at quadrature points.

  If the backend does not support `CUfunction` pointers for QFunctions, then the call succeeds without effect.
  When unsupported, a message is emitted via `CeedDebug`.
```

minor

interface/ceed-cuda.c
```diff
@@ -29,3 +32,25 @@ int CeedQFunctionSetCUDAUserFunction(CeedQFunction qf, CUfunction f) {
   }
   return CEED_ERROR_SUCCESS;
 }
+
+/**
+  @brief Enable or disable CUDA Graph capture/replay for a `CeedOperator`
+
+  If the backend does not support CUDA Graphs for operators, the call succeeds without effect.
+  When unsupported, a message is emitted via `CeedDebug`.
```
jeremylt:
```suggestion
  @brief Enable or disable CUDA Graph capture/replay for a `CeedOperator`.

  If the backend does not support CUDA Graphs for operators, then the call succeeds without effect.
  When unsupported, a message is emitted via `CeedDebug`.
```

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -344,6 +413,124 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
...
 
+static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector input_vec, CeedVector output_vec, CeedRequest *request) {
+  Ceed                   ceed;
+  CeedOperator_Cuda_gen *impl;
+  CeedOperator          *sub_operators;
+  CeedInt                num_suboperators;
+
+  ceed = CeedOperatorReturnCeed(op);
+  CeedCall(CeedOperatorCompositeGetNumSub(op, &num_suboperators));
+  CeedCall(CeedOperatorCompositeGetSubList(op, &sub_operators));
```
zatkins-dev: All CeedCall in this file should be CeedCallBackend

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -344,6 +413,124 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
...
+
+  bool need_build = !impl->graph_created;
+
+  if (!need_build && input_vec != CEED_VECTOR_NONE) {
+    const CeedScalar *in_ptr;
+
+    CeedCallBackend(CeedVectorGetArrayRead(input_vec, CEED_MEM_DEVICE, &in_ptr));
+    need_build = in_ptr != impl->captured_input_ptr;
+    CeedCallBackend(CeedVectorRestoreArrayRead(input_vec, &in_ptr));
+  }
```
zatkins-dev, line 443: Do we also need to capture and check for the output vector pointer?

interface/ceed-cuda.c
Nafees01: Done.

interface/ceed-cuda.c
Nafees01: Done

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fixed.

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Good point! The graph captures the output device pointer too, not just input. I added the same recapture check for output. Ratel reused the same I/O buffers in our runs, so we didn’t notice it before, but this makes the general case safer.

backends/cuda-ref/ceed-cuda-ref-vector.c
```diff
@@ -326,7 +326,17 @@ static int CeedVectorSetValue_Cuda(CeedVector vec, CeedScalar val) {
...
   if (impl->d_array) {
     if (val == 0) {
-      CeedCallCuda(CeedVectorReturnCeed(vec), cudaMemset(impl->d_array, 0, length * sizeof(CeedScalar)));
+      // Check if we're in CUDA Graph capture mode
+      enum cudaStreamCaptureStatus capture_status;
+
+      cudaStreamIsCapturing(cudaStreamPerThread, &capture_status);
+      if (capture_status != cudaStreamCaptureStatusNone) {
+        // During capture, use async memset with cudaStreamPerThread
+        CeedCallCuda(CeedVectorReturnCeed(vec), cudaMemsetAsync(impl->d_array, 0, length * sizeof(CeedScalar), cudaStreamPerThread));
```
zatkins-dev: Note to me and @jeremylt -- we should probably make these async/non-blocking during normal execution; it shouldn't affect determinism and will likely improve perf. If blocking is needed, that's likely a bug anyway.

zatkins-dev (commented):
I think this looks pretty good. I'll hold off for @jeremylt on final approval, but this is a really cool feature add! Thanks @Nafees01!

backends/cuda-ref/ceed-cuda-ref-vector.c
jeremylt: I think honestly than would be cleaner and clearer

Nafees01:
> I think this looks pretty good. I'll hold off for @jeremylt on final approval, but this is a really cool feature add! Thanks @Nafees01!

Thanks @zatkins-dev and thanks @jeremylt for all the detailed feedback - really helped get this in good shape. Appreciate you both taking the time!

backends/cuda-ref/ceed-cuda-ref-vector.c
jeremylt: yea, the more I think on it, the better off I think we are just defaulting to async and skipping this special case logic

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -284,7 +300,10 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
   // Try to run kernel
   if (input_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArrayRead(input_vec, CEED_MEM_DEVICE, &input_arr));
   if (output_vec != CEED_VECTOR_NONE) CeedCallBackend(CeedVectorGetArray(output_vec, CEED_MEM_DEVICE, &output_arr));
-  CeedCallBackend(CeedOperatorApplyAddCore_Cuda_gen(op, NULL, input_arr, output_arr, &is_run_good, request));
+  enum cudaStreamCaptureStatus capture_status;
+  cudaStreamIsCapturing(cudaStreamPerThread, &capture_status);
```
jeremylt: call should be wrapped in error catching macro

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -284,7 +300,10 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
+  CUstream stream_to_use = (capture_status != cudaStreamCaptureStatusNone) ? cudaStreamPerThread : NULL;
```
jeremylt:
```suggestion
  CUstream stream_to_use = (capture_status != cudaStreamCaptureStatusNone) ? cudaStreamPerThread : NULL;

```
Always blank line after variable declaration

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -299,7 +318,58 @@ static int CeedOperatorApplyAdd_Cuda_gen(CeedOperator op, CeedVector input_vec,
...
-static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector input_vec, CeedVector output_vec, CeedRequest *request) {
+// Push passive inputs and QFunction context to device before graph replay.
+static int CeedCompositeRefreshForReplay_Cuda_gen(CeedOperator *sub_operators, CeedInt num_suboperators) {
+  for (CeedInt i = 0; i < num_suboperators; i++) {
+    bool                is_at_points;
+    CeedInt             num_input_fields, num_output_fields;
+    CeedOperatorField  *op_input_fields, *op_output_fields;
+    CeedQFunction       qf  = NULL;
+    CeedQFunctionField *qf_input_fields;
+    void               *d_c = NULL;
```
jeremylt: variable name should use words

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -344,6 +414,134 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
   return CEED_ERROR_SUCCESS;
 }
 
+static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector input_vec, CeedVector output_vec, CeedRequest *request) {
+  Ceed                   ceed;
+  CeedOperator_Cuda_gen *impl;
+  CeedOperator          *sub_operators;
+  CeedInt                num_suboperators;
+
+  ceed = CeedOperatorReturnCeed(op);
```
jeremylt: This function should not be used to get a persistent ceed
```suggestion
  CeedCallBackend(CeedOperatorGetCeed(op, &ceed));
```

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -344,6 +414,134 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
...
+    if (impl->graph) CeedCallCuda(ceed, cudaGraphDestroy(impl->graph));
+    impl->graph              = NULL;
+    impl->graph_instance     = NULL;
+    impl->graph_created       = false;
+    impl->captured_input_ptr  = NULL;
+    impl->captured_output_ptr = NULL;
+    CeedCallBackend(CeedOperatorSetEnableCudaGraph(op, false));
+    return CeedOperatorApplyAddComposite_NoGraph_Cuda_gen(op, input_vec, output_vec, request);
+  }
+  return CEED_ERROR_SUCCESS;
```
jeremylt, line 534: And destroy that Ceed down here

jeremylt: really, before each of the return statements

jeremylt: You could, alternatively, use `CeedOperatorReturnCeed()` exactly in the sites where you use the `ceed`. Just don't use `CeedOperatorReturnCeed()` to get around having to destroy a persistent ceed variable

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -885,24 +1094,36 @@ int CeedOperatorCreate_Cuda_gen(CeedOperator op) {
   CeedCallBackend(CeedOperatorGetCeed(op, &ceed));
   CeedCallBackend(CeedCalloc(1, &impl));
   CeedCallBackend(CeedOperatorSetData(op, impl));
-  CeedCall(CeedOperatorIsComposite(op, &is_composite));
+
+  CeedCallBackend(CeedOperatorIsComposite(op, &is_composite));
```
jeremylt:
```suggestion
  CeedCallBackend(CeedOperatorIsComposite(op, &is_composite));
```

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -885,24 +1094,36 @@ int CeedOperatorCreate_Cuda_gen(CeedOperator op) {
...
     CeedCallBackend(CeedSetBackendFunction(ceed, "Operator", op, "ApplyAdd", CeedOperatorApplyAdd_Cuda_gen));
   }
-  CeedCall(CeedOperatorIsAtPoints(op, &is_at_points));
+  CeedCallBackend(CeedOperatorIsAtPoints(op, &is_at_points));
   if (is_at_points) {
     CeedCallBackend(CeedSetBackendFunction(ceed, "Operator", op, "LinearAssembleAddDiagonal",
                                            CeedOperatorLinearAssembleAddDiagonalAtPoints_Cuda_gen));
     CeedCallBackend(CeedSetBackendFunction(ceed, "Operator", op, "LinearAssembleSingle", CeedOperatorAssembleSingleAtPoints_Cuda_gen));
   }
+
```
jeremylt: ok, either add blank lines before those if/else blocks or don't please - the mix is confusing

backends/cuda-gen/ceed-cuda-gen.h
```diff
@@ -25,6 +26,14 @@ typedef struct {
   Fields_Cuda    G;
   CeedScalar    *W;
   Points_Cuda    points;
+
+  bool              use_graph;
```
jeremylt, line 31:
```suggestion
  // Graph capture data
  bool              use_graph;
```
Either a comment here or no blank line

backends/cuda-ref/ceed-cuda-ref-vector.c
```diff
@@ -326,7 +326,17 @@ static int CeedVectorSetValue_Cuda(CeedVector vec, CeedScalar val) {
...
+      } else {
+        // Normal execution, use blocking memset
+        CeedCallCuda(CeedVectorReturnCeed(vec), cudaMemset(impl->d_array, 0, length * sizeof(CeedScalar)));
+      }
```
jeremylt:
```suggestion
      CeedCallCuda(CeedVectorReturnCeed(vec), cudaMemsetAsync(impl->d_array, 0, length * sizeof(CeedScalar), cudaStreamPerThread));
```

backends/cuda-ref/ceed-cuda-ref-vector.c
Nafees01: removed the `capture-status` check and always use `cudaMemsetAsync`

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: wrapped `cudaStreamIsCapturing` in `CeedCallCuda`

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fixed

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fixed

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Fixed

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Dropped the persistent `ceed` and using `CeedOperatorReturnCeed()`

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: Done

backends/cuda-gen/ceed-cuda-gen-operator.c
Nafees01: I agree with you. Sorry for the confusion, and that’s fair enough. I think I didn’t pay much attention to blank lines while coding. I’ll definitely keep that in mind in future. Thanks again for pointing it out.

backends/cuda-gen/ceed-cuda-gen.h
Nafees01: Fixed and noted

backends/cuda-ref/ceed-cuda-ref-vector.c
Nafees01: Removed

backends/cuda-gen/ceed-cuda-gen-operator.c
jeremylt: Looks good. Generally we want to prefer `CeedOperatorGetCeed` unless the ceed is only used in one or two places or the function has multiple exit points.

backends/cuda-gen/ceed-cuda-gen-operator.c
jeremylt: It can be a nice way to logically group chunks with some visual space, but its gotta be consistent. I look out for it because it can make a difference in how easy it is for someone to start to understand the codebase while reading

jeremylt:
CI wants some style fixes and the release notes to be updated

jeremylt:
There's been some CI changes, so this probably needs to be rebased before it can be run on Noether. Also, only branches on the main repo can run on Noether so we probably want to merge this into a local branch so Noether CI can run.

Nafees01:
> CI wants some style fixes and the release notes to be updated

Updated CHANGELOG and fixed the style issues.

Nafees01:
> There's been some CI changes, so this probably needs to be rebased before it can be run on Noether. Also, only branches on the main repo can run on Noether so we probably want to merge this into a local branch so Noether CI can run.

Rebase had some conflicts, so I merged the latest upstream/main into cuda-graph-dev instead. The branch is now up to date with main.

jeremylt:
This has been a very old branch, but I think learning how to handle those conflicts and get a clean branch history is a skill to develop eventually.

We can't just put all of these commits on main, so we need to squash+merge. And we still need Noether CI to run, so we'll squash+merge to a staging branch first

jeremylt (approved):

jeremylt:
More style fixes are needed

Nafees01:
> More style fixes are needed

Ran clang-format-22. Style CI should be clean now.

Nafees01:
> This has been a very old branch, but I think learning how to handle those conflicts and get a clean branch history is a skill to develop eventually.
> 
> We can't just put all of these commits on main, so we need to squash+merge. And we still need Noether CI to run, so we'll squash+merge to a staging branch first

Sounds good, and thanks for the advice. I'll keep pushing fixes here. Thanks for switching the base to test/cuda-graph so Noether can run.

jeremylt:
Testing on Noether over in https://github.com/CEED/libCEED/pull/2006

PR #1988, opened 2026-07-11
Adding CUDA GPU support for Filter function

wostrie2:
Purpose: adding CUDA GPU functionality for the Filter function, which filters or clips a vector to a threshold value by setting vector components to 0.0 if their absolute value is less than `threshold`.

Closes: N/A

LLM/GenAI Disclosure: Google AI spotted a missing parenthesis and saved me a compile error.

Helped by: Jeremy L. Thompson is a human who offered a lot of kind and helpful support while I was working on this. 

### DIFF
--- a/backends/cuda-ref/ceed-cuda-ref-vector.c
+++ b/backends/cuda-ref/ceed-cuda-ref-vector.c
@@ -718,6 +718,36 @@ static int CeedVectorScale_Cuda(CeedVector x, CeedScalar alpha) {
 }
 
 //------------------------------------------------------------------------------
+// Filter or clip a vector using a threshold value on the host
+//------------------------------------------------------------------------------
+static int CeedHostFilter_Cuda(CeedScalar *x_array, CeedScalar threshold, CeedSize length) {
+  CeedPragmaSIMD for (CeedSize i = 0; i < length; i++) {
+    if (fabs(x_array[i]) <= threshold) x_array[i] = 0.0;
+  }
+  return CEED_ERROR_SUCCESS;
+}
+
+//------------------------------------------------------------------------------
+// Filter or clip a vector using a threshold value on device (impl in .cu file)
+//------------------------------------------------------------------------------
+int CeedDeviceFilter_Cuda(CeedScalar *x_array, CeedScalar threshold, CeedSize length);
+
+//------------------------------------------------------------------------------
+// Filter or clip a vector using a threshold value
+//------------------------------------------------------------------------------
+static int CeedVectorFilter_Cuda(CeedVector vec, CeedScalar threshold) {
+  CeedSize         length;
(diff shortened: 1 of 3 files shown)


jrwrigh (approved):
This looks fine to me, but I'll wait for someone who's more familiar with the GPU backends to double check before merging.

Not sure what's going on with Noether CI though.

jeremylt:
Noether won't run for forks.

I can verify the tests myself tomorrow

jeremylt:
Ok, we need to fix t129, because your code is all correct but the test is not executing on the GPU.

So, here's what has to happen.

I'd insert this code block into your `InitVec
```
  {
    // Sync memtype to device for GPU backends
    CeedMemType type = CEED_MEM_HOST;

    CeedGetPreferredMemType(CeedVectorReturnCeed(x), &type);
    CeedVectorSyncArray(x, type);
  }
```

That will pull the memory up to the device, which then will trigger the code you added during the test.

While you're in there, can you update a small style fix? A block (or line) of variable declarations must always be followed by an empty line, per our style guide, so `CeedScalar array[len]` on line 16 needs a blank line after it.

(We may need to add this fix to other vector tests, need to check)

Lemmie know if you have questions about this explanation!

wostrie2:
> Ok, we need to fix t129, because your code is all correct but the test is not executing on the GPU.
> 
> So, here's what has to happen.
> 
> I'd insert this code block into your `InitVec
> 
> ```
>   {
>     // Sync memtype to device for GPU backends
>     CeedMemType type = CEED_MEM_HOST;
> 
>     CeedGetPreferredMemType(CeedVectorReturnCeed(x), &type);
>     CeedVectorSyncArray(x, type);
>   }
> ```
> 
> That will pull the memory up to the device, which then will trigger the code you added during the test.
> 
> While you're in there, can you update a small style fix? A block (or line) of variable declarations must always be followed by an empty line, per our style guide, so `CeedScalar array[len]` on line 16 needs a blank line after it.
> 
> (We may need to add this fix to other vector tests, need to check)
> 
> Lemmie know if you have questions about this explanation!

Ok, I added that block at the end of the InitVector function, and I added a blank line after the variable declaration on line 16.

Should I delete the equivalent block for memory syncing that appears between test cases 1 and 2?

jeremylt:
Yeah, go ahead and remove the one outside of the init function since it isn't helping us anymore

wostrie2:
ok, just pushed those changes to t129-vector.c

jeremylt (approved):
Awesome, this is working for me locally and running on the GPU! Nice work

wostrie2:
> Awesome, this is working for me locally and running on the GPU! Nice work

Thanks for all your support with this!!

jeremylt:
Sure thing; I'm happy to help

PR #1990, opened 2026-07-15
cuda: Remove compile-time dependence on maximum number of points per elements

zatkins-dev:
Purpose:

Reduce the required recompilation of restriction, basis, and operator kernels by changing maximum number of points per element from a compile-time to run-time parameter

Closes: #

LLM/GenAI Disclosure:

None

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator.c
+++ b/backends/cuda-gen/ceed-cuda-gen-operator.c
@@ -174,17 +174,20 @@ static int CeedOperatorApplyAddCore_Cuda_gen(CeedOperator op, CUstream stream, c
   CeedCallBackend(CeedOperatorIsAtPoints(op, &is_at_points));
   if (is_at_points) {
     // Coords
-    CeedVector vec;
+    CeedVector          vec;
+    CeedElemRestriction rstr_points = NULL;
 
-    CeedCallBackend(CeedOperatorAtPointsGetPoints(op, NULL, &vec));
+    CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, &vec));
     CeedCallBackend(CeedVectorGetArrayRead(vec, CEED_MEM_DEVICE, &data->points.coords));
+    CeedCallBackend(CeedElemRestrictionGetMaxPointsInElement(rstr_points, &data->points.max_num_points));
+    CeedCallBackend(CeedElemRestrictionGetCompStride(rstr_points, &data->points.coords_comp_stride));
     CeedCallBackend(CeedVectorDestroy(&vec));
+    CeedCallBackend(CeedElemRestrictionDestroy(&rstr_points));
 
     // Points per elem
     if (num_elem != data->points.num_elem) {
-      CeedInt            *points_per_elem;
-      const CeedInt       num_bytes   = num_elem * sizeof(CeedInt);
-      CeedElemRestriction rstr_points = NULL;
+      CeedInt      *points_per_elem;
+      const CeedInt num_bytes = num_elem * sizeof(CeedInt);
 
       data->points.num_elem = num_elem;
       CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, NULL));
@@ -608,11 +611,15 @@ static int CeedOperatorLinearAssembleAddDiagonalAtPoints_Cuda_gen(CeedOperator o
 
     // Point coordinates
     {
-      CeedVector vec;
+      CeedVector          vec;
+      CeedElemRestriction rstr_points = NULL;
 
-      CeedCallBackend(CeedOperatorAtPointsGetPoints(op, NULL, &vec));
+      CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, &vec));
       CeedCallBackend(CeedVectorGetArrayRead(vec, CEED_MEM_DEVICE, &data->points.coords));
+      CeedCallBackend(CeedElemRestrictionGetMaxPointsInElement(rstr_points, &data->points.max_num_points));
+      CeedCallBackend(CeedElemRestrictionGetCompStride(rstr_points, &data->points.coords_comp_stride));
       CeedCallBackend(CeedVectorDestroy(&vec));
+      CeedCallBackend(CeedElemRestrictionDestroy(&rstr_points));
 
       // Points per elem
       if (num_elem != data->points.num_elem) {
@@ -774,11 +781,15 @@ static int CeedOperatorAssembleSingleAtPoints_Cuda_gen(CeedOperator op, CeedInt 
 
     // Point coordinates
     {
-      CeedVector vec;
+      CeedVector          vec;
+      CeedElemRestriction rstr_points = NULL;
 
-      CeedCallBackend(CeedOperatorAtPointsGetPoints(op, NULL, &vec));
+      CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, &vec));
       CeedCallBackend(CeedVectorGetArrayRead(vec, CEED_MEM_DEVICE, &data->points.coords));
+      CeedCallBackend(CeedElemRestrictionGetMaxPointsInElement(rstr_points, &data->points.max_num_points));
+      CeedCallBackend(CeedElemRestrictionGetCompStride(rstr_points, &data->points.coords_comp_stride));
       CeedCallBackend(CeedVectorDestroy(&vec));
+      CeedCallBackend(CeedElemRestrictionDestroy(&rstr_points));
 
       // Points per elem
       if (num_elem != data->points.num_elem) {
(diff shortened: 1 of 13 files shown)


jeremylt:
I'll check more tomorrow or Friday, but at a blush it seems reasonable. HIP coming too?

And did you check in Ratel when we know the max number of points changed?

Other thought - we should profile as I thought the data access optimization of knowing the indexing stride was noticeable, but I can recall why I thought that.

zatkins-dev:
I can profile and add HIP next week -- moving this weekend and wanted to get this proof of concept up for review.

zatkins-dev:
> And did you check in Ratel when we know the max number of points changed?

Technically we trash the operators in Ratel whenever things change -- it will be difficult to fully test until we have the operator rebuild interface.

zatkins-dev:
> we should profile as I thought the data access optimization of knowing the indexing stride was noticeable

I would expect that the real impact is that the indexing stride is uniform across the thread block. I also think the loop unrolling for batches of points (which is definitely broken by this) would be a bigger impact. We can probably recover perf via manual unrolling or by using a fixed batch size instead.

backends/cuda-ref/ceed-cuda-ref-basis.c
```diff
@@ -126,6 +127,7 @@ static int CeedBasisApplyAtPointsCore_Cuda(CeedBasis basis, bool apply_add, cons
 
   // Check padded to uniform number of points per elem
   for (CeedInt i = 1; i < num_elem; i++) max_num_points = CeedIntMax(max_num_points, num_points[i]);
+  data->num_points = max_num_points;
```
jeremylt, line 130: I'd probably call this `data->max_num_points`, dunno why I named it the way I did before

backends/cuda-ref/ceed-cuda-ref-basis.c
```diff
@@ -212,15 +208,15 @@ static int CeedBasisApplyAtPointsCore_Cuda(CeedBasis basis, bool apply_add, cons
   // Basis action
   switch (eval_mode) {
     case CEED_EVAL_INTERP: {
-      void         *interp_args[] = {(void *)&num_elem, &data->d_chebyshev_interp_1d, &data->d_points_per_elem, &d_x, &d_u, &d_v};
-      const CeedInt block_size    = CeedIntMin(CeedIntPow(Q_1d, dim), max_block_size);
+      void *interp_args[] = {(void *)&num_elem, (void *)&data->num_points, &data->d_chebyshev_interp_1d, &data->d_points_per_elem, &d_x, &d_u, &d_v};
```
jeremylt, line 211: we should see if we have (or should establish) a consistent pattern for the inputs to these kernels

backends/cuda-ref/ceed-cuda-ref-restriction.c
```diff
@@ -155,7 +153,15 @@ static inline int CeedElemRestrictionApply_Cuda_Core(CeedElemRestriction rstr, C
 
         CeedCallBackend(CeedRunKernel_Cuda(ceed, impl->ApplyNoTranspose, grid, block_size, args));
       } break;
-      case CEED_RESTRICTION_POINTS:
+      case CEED_RESTRICTION_POINTS: {
+        CeedInt max_num_points;
+
+        CeedCallBackend(CeedElemRestrictionGetMaxPointsInElement(rstr, &max_num_points));
```
jeremylt, line 159: this should never change for the *restriction* so we can save the effort of recomputing in on every application here

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -174,17 +174,20 @@ static int CeedOperatorApplyAddCore_Cuda_gen(CeedOperator op, CUstream stream, c
...
   if (is_at_points) {
     // Coords
-    CeedVector vec;
+    CeedVector          vec;
+    CeedElemRestriction rstr_points = NULL;
 
-    CeedCallBackend(CeedOperatorAtPointsGetPoints(op, NULL, &vec));
+    CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, &vec));
     CeedCallBackend(CeedVectorGetArrayRead(vec, CEED_MEM_DEVICE, &data->points.coords));
+    CeedCallBackend(CeedElemRestrictionGetMaxPointsInElement(rstr_points, &data->points.max_num_points));
```
jeremylt, line 182: this should only change if the restriction changes, see the other comment about avoiding re-computing

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -174,17 +174,20 @@ static int CeedOperatorApplyAddCore_Cuda_gen(CeedOperator op, CUstream stream, c
...
+    CeedCallBackend(CeedElemRestrictionGetCompStride(rstr_points, &data->points.coords_comp_stride));
```
jeremylt, line 183: ... should this be allowed to change?

backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -774,11 +781,15 @@ static int CeedOperatorAssembleSingleAtPoints_Cuda_gen(CeedOperator op, CeedInt
...
     // Point coordinates
     {
-      CeedVector vec;
+      CeedVector          vec;
+      CeedElemRestriction rstr_points = NULL;
 
-      CeedCallBackend(CeedOperatorAtPointsGetPoints(op, NULL, &vec));
+      CeedCallBackend(CeedOperatorAtPointsGetPoints(op, &rstr_points, &vec));
       CeedCallBackend(CeedVectorGetArrayRead(vec, CEED_MEM_DEVICE, &data->points.coords));
+      CeedCallBackend(CeedElemRestrictionGetMaxPointsInElement(rstr_points, &data->points.max_num_points));
```
jeremylt, line 789: same here - should only re-compute if the rstr changed (really, should be something we can query if we create an interface to update the points rstr, which we should only allow if the number of elements does not change)

backends/cuda-ref/ceed-cuda-ref-basis.c
```diff
@@ -105,7 +105,7 @@ static int CeedBasisApplyAdd_Cuda(CeedBasis basis, const CeedInt num_elem, CeedT
 static int CeedBasisApplyAtPointsCore_Cuda(CeedBasis basis, bool apply_add, const CeedInt num_elem, const CeedInt *num_points,
                                            CeedTransposeMode t_mode, CeedEvalMode eval_mode, CeedVector x_ref, CeedVector u, CeedVector v) {
   Ceed              ceed;
-  CeedInt           Q_1d, dim, max_num_points = num_points[0];
+  CeedInt           P_1d, Q_1d, dim, max_num_points = num_points[0];
```
jeremylt, line 108: why did this need to get pulled up?

jeremylt:
Yeah, overall looks pretty good; I just had a few questions. But we definitely shouldn't merge anything without understanding the performance implications for sure

backends/cuda-gen/ceed-cuda-gen-operator.c
zatkins-dev: I'm relatively sure that is usually also max num points, see the shared basis functions.

jeremylt: hmm, if that's the case this a separate duplicate value shouldn't exist

backends/cuda-gen/ceed-cuda-gen-operator.c
zatkins-dev: does that function not cache the result? if not, we should cache and invalidate when updating the points

backends/cuda-gen/ceed-cuda-gen-operator.c
zatkins-dev: yeah, i think that it's because of the different striding between CPU and GPU, but we know what the striding will be so we can reduce the complexity.

backends/cuda-gen/ceed-cuda-gen-operator.c
jeremylt: The function could cache yea. But the restriction should never update it because you would need a new restriction

include/ceed/jit-source/cuda/cuda-gen-templates.h
```diff
@@ -24,27 +24,29 @@ inline __device__ void LoadMatrix(SharedData_Cuda &data, const CeedScalar *__res
...
 //------------------------------------------------------------------------------
-template <int NUM_COMP, int COMP_STRIDE, int NUM_PTS>
-inline __device__ void ReadPoint(SharedData_Cuda &data, const CeedInt elem, const CeedInt p, const CeedInt points_in_elem,
-                                 const CeedInt *__restrict__ indices, const CeedScalar *__restrict__ d_u, CeedScalar *r_u) {
-  const CeedInt ind = indices[p + elem * NUM_PTS];
+template <int NUM_COMP>
+inline __device__ void ReadPoint(SharedData_Cuda &data, const CeedInt elem, const CeedInt p, const CeedInt max_num_points,
+                                 const CeedInt points_in_elem, const CeedInt     comp_stride, const CeedInt *__restrict__ indices,
+                                 const CeedScalar *__restrict__ d_u, CeedScalar *r_u) {
+  const CeedInt ind = indices[p + elem * max_num_points];
```
jeremylt, line 31: this is I think where the perf impact might be most notable - I think knowing the indexing pattern at compile time helps here? Unsure

PR #1992, opened 2026-07-17
cuda - fix grad3d flat templates for gen

### LINKED ISSUE #1991: NVRTC Compile error in GradTransposeTensor3dFlattened
> > > > Sub-operators that cannot be captured, for example, contact operators that allocate memory (cudaMalloc) during apply
> > 
> > 
> > > 
> > 
> > 
> > > Are there specific operators you were unable to use in Ratel? None of the QFunctions there should alloc during operator application.
> > 
> > 
> > The only example where the contact (Nitsche) sub-operator couldn't be captured was the quasistatic contact case (ex02-quasistatic-elasticity-linear-platen). Everything else I tried static, dynamic, multi-material, neo-Hookean, Mooney-Rivlin captured and replayed fine.
> > And you're right that the QFunction doesn't allocate; the cudaMalloc comes from libCEED's own device-side handling of the contact data, not the QFunction itself. It runs either way correctly; the contact operators just fall back to a normal apply rather than being replayed.
> 
> This is a pretty big red flag to me. The Nitsche contact operators don't actually do anything out of the ordinary on the libCEED side. Them failing to work means that something about this setup (very possibly something with passive inputs and outputs) is violating the libCEED interface assumptions.

Thanks for flagging this. On re-checking linear-platen with CEED_DEBUG=1, the contact operators appear to fail earlier than graph capture: cuda-gen hits an NVRTC compile error (GradTransposeTensor3dFlattened defined twice in cuda-shared-basis-tensor-flattened-templates.h) and then falls back to /gpu/cuda/ref. So this looks like a pre-existing cuda-gen JIT issue triggered by the contact operator's basis configuration, rather than something specific to Nitsche or CUDA graphs.

_Originally posted by @Nafees01 in https://github.com/CEED/libCEED/pull/1987#issuecomment-4996208345_

jeremylt:
Purpose:

Fix my oops. That was totally copypasta

Closes: #1991 

LLM/GenAI Disclosure:

None, pure human stupidity here

### DIFF
--- a/.github/workflows/release-notes.yml
+++ b/.github/workflows/release-notes.yml
@@ -20,7 +20,8 @@ jobs:
     - name: Check release notes
       run: |
         git fetch origin main
-        if git diff origin/main --exit-code include/ceed/*; then
+        if git diff origin/main --exit-code include/*.h && \
+           git diff origin/main --exit-code include/ceed/*.h; then
           echo "No public interface changes detected"
         elif git diff origin/main --exit-code CHANGELOG.md; then
           echo "API changes detected, but release notes not updated"
--- a/include/ceed/jit-source/cuda/cuda-shared-basis-tensor-flattened-templates.h
+++ b/include/ceed/jit-source/cuda/cuda-shared-basis-tensor-flattened-templates.h
@@ -570,15 +570,15 @@ inline __device__ void GradTransposeTensor3dFlattened(SharedData_Cuda &data, Cee
 
   if (Q_1D != T_1D) QUnpack3d<NUM_COMP * 3, Q_1D, T_1D>(data, t_id_x, t_id_y, t_id_z, r_U);
   for (CeedInt comp = 0; comp < NUM_COMP; comp++) {
-    ContractTransposeZ3dFlattened<NUM_COMP, t_id_x, t_id_y, t_id_z, P_1D, Q_1D, T_1D>(data, &r_U[comp + 0 * NUM_COMP], c_B, r_t1);
-    ContractTransposeY3dFlattened<NUM_COMP, t_id_x, t_id_y, t_id_z, P_1D, Q_1D, T_1D>(data, r_t1, c_B, r_t2);
-    ContractTransposeX3dFlattened<NUM_COMP, t_id_x, t_id_y, t_id_z, P_1D, Q_1D, T_1D>(data, r_t2, c_G, &r_V[comp]);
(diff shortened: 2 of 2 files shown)


.github/workflows/release-notes.yml
```diff
@@ -20,7 +20,8 @@ jobs:
     - name: Check release notes
       run: |
         git fetch origin main
-        if git diff origin/main --exit-code include/ceed/*; then
+        if git diff origin/main --exit-code include/*.h && \
+           git diff origin/main --exit-code include/ceed/*.h; then
```
jeremylt, line 24: ooops, we really only wanted this to check if the header files changed, not the JiT files

zatkins-dev:
Does HIP also need this fix?

jeremylt:
oddly no, I did that one correctly

zatkins-dev:
huh, fair enough. Did you make sure it isn't still falling back? or should I check

jeremylt:
I didn't double check

zatkins-dev:
Seems to compile okay now. I'm now getting:
```
CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES: max_threads_per_block 256 on block size (27,1,18), shared_size 8352, num_regs 250
```

jeremylt:
Yea, I'm not surprised. Flattening should only work for really low order

zatkins-dev:
Sounds good. Also, is there a reason we fall back straight to ref? In this case, fallback to shared would be faster

jeremylt:
Honestly, probably an oversight

jeremylt:
Oh, you know what I bet it is? Actually, should be using shared bases, just ref operator evaluation

zatkins-dev:
Isn't that what the shared operator application does anyhow?

jeremylt:
Yea. There is no 'shared operator' so I directly fell back to the compatible backend that implements the target functionality so it was clear (to me) where that operator application/assembly logic actually came from

zatkins-dev:
Gotcha. We should switch the fallback to shared then?

jeremylt:
I personally wouldn't because it is the *operator* implementation that is falling back and ref is what provides that fallback operator implementation, but I don't feel strongly

zatkins-dev:
Hmm, will it still use the shared bases with the ref operator fallback?

zatkins-dev:
yes it will

zatkins-dev (approved):
Okay, I think this is good! I'm going to have a follow up fixing block sizes for the non tensor gen, we can make it work for more cases

PR #2002, opened 2026-08-07
feat(hip): add vector filter backend

tomatotomata:
## Summary

- add the `CeedVectorFilter` host and HIP-device implementations to `/gpu/hip/ref`
- register the backend operation so the existing public API uses the HIP path
- keep the threshold semantics aligned with the existing CUDA reference backend

This follows the remaining HIP work identified in #1830 and ports the already-merged CUDA implementation from #1988. The existing `tests/t129-vector.c` covers the public operation for vector sizes including zero and threshold boundary cases.

## Validation

- `git diff --check` passed
- local HIP compilation was not available because this Windows checkout has neither `hipcc` nor GNU `make`; the HIP source will need the repository CI environment for compile and test validation

## LLM/GenAI disclosure

I used Codex to help inspect the existing CUDA implementation and draft the HIP adaptation. I reviewed the resulting diff and am responsible for the submitted code.

### DIFF
--- a/backends/hip-ref/ceed-hip-ref-vector.c
+++ b/backends/hip-ref/ceed-hip-ref-vector.c
@@ -828,6 +828,36 @@ static int CeedVectorScale_Hip(CeedVector x, CeedScalar alpha) {
 }
 
 //------------------------------------------------------------------------------
+// Filter or clip a vector using a threshold value on the host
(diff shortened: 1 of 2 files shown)


jeremylt:
The associated issue is intended for students to work on - not mass automated submissions

PR #2008, opened 2026-08-19
ctx - expand CeedContextFieldType to support additional field types

### LINKED ISSUE #2004: Support additional `CeedContextFieldType` values
## Summary

Add new enum values to `CeedContextFieldType` for single-precision floating point, 64-bit integer, byte, and `CeedInt` and `CeedScalar` types.

## Possible Implementation

- [ ] Expand the `CeedContextFieldType` interface to include:
```c
typedef enum {
  /// Byte value, C type of char
  CEED_CONTEXT_FIELD_BYTE = 1,
  /// CeedScalar value
  CEED_CONTEXT_FIELD_SCALAR = 2,
  /// Single precision value
  CEED_CONTEXT_FIELD_FLOAT = 3,
  /// Double precision value
  CEED_CONTEXT_FIELD_DOUBLE = 4,
  /// CeedInt value
  CEED_CONTEXT_FIELD_INT = 5,
  /// 32 bit integer value
  CEED_CONTEXT_FIELD_INT32 = 6,
  /// 64 bit integer value
  CEED_CONTEXT_FIELD_INT64 = 7,
  /// Boolean value
  CEED_CONTEXT_FIELD_BOOL = 8,
} CeedContextFieldType;
```

- [ ] Add strings/update indices in `CeedContextFieldTypes`.

- [ ] Add associated public registration functions:
```c
int CeedQFunctionContextRegisterByte(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values, const char *field_description);
int CeedQFunctionContextRegisterScalar(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values, const char *field_description);
int CeedQFunctionContextRegisterFloat(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values, const char *field_description);
int CeedQFunctionContextRegisterInt(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values, const char *field_description);
int CeedQFunctionContextRegisterInt64(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values, const char *field_description);
```

- [ ] Add the associated public `Get[TYPE]Read`/`Restore[TYPE]Read`/`Set[TYPE]` functions (byte used as an example):
```c
int CeedOperatorGetContextByteRead(CeedOperator op, CeedContextFieldLabel field_label, size_t *num_values, const char **values);
int CeedOperatorRestoreContextByteRead(CeedOperator op, CeedContextFieldLabel field_label, size_t *num_values, const char **values);
int CeedOperatorSetContextByte(CeedOperator op, CeedContextFieldLabel field_label, char *values);
int CeedQFunctionContextGetByteRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, const char **values);
int CeedQFunctionContextRestoreByteRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const char **values);
int CeedQFunctionContextSetByte(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, char *values);
```

- [ ] Update `CeedQFunctionContextRegisterGeneric`, `CeedQFunctionContextGetGenericRead`.

- [ ] Regenerate Julia bindings.

- [ ] Add new values to `tests/t407-qfunction`.

ksd-2005:
Purpose:

Adds seven new `CeedContextFieldType` values - `BYTE`, `INT8`, `INT`, `INT64`, `SIZE`, `SCALAR`,
and `FLOAT` - alongside the existing `DOUBLE`, `INT32`, and `BOOL`, with matching
`CeedQFunctionContextRegisterCeed*` / `Set*` / `Get*Read` / `Restore*Read` functions on both
`CeedQFunctionContext` and `CeedOperator`.

As we agreed in the issue discussion, this renumbers the existing `CeedContextFieldType` enum
(an intentional ABI break, approved ahead of the 1.0 release) and gives every accessor function
a `Ceed`-prefixed name for the libCEED-typedef'd types, as per @jeremylt 's and @zatkins-dev 's suggestion.

`CEED_CONTEXT_FIELD_SCALAR`, `FLOAT`, and `DOUBLE` are kept as distinct field types.

One thing I want to mention is: 
The discussion we had earlier didn't say an exact enum ordering for the final 10-type set - the only ordering anyone said was the case order of
the `switch` @jeremylt  posted (BOOL, BYTE, INT8, INT, INT32, INT64, SIZE, SCALAR, FLOAT, DOUBLE).
I've used 
``` cpp
typedef enum {
  /// Byte value, C type of char
  CEED_CONTEXT_FIELD_BYTE = 1,
  /// CeedScalar value
  CEED_CONTEXT_FIELD_SCALAR = 2,
  /// Single precision value
  CEED_CONTEXT_FIELD_FLOAT = 3,
  /// Double precision value
  CEED_CONTEXT_FIELD_DOUBLE = 4,
  ///8 bit integer value
  CEED_CONTEXT_FIELD_INT8 = 5,
  /// CeedInt value
  CEED_CONTEXT_FIELD_INT = 6,
  /// 32 bit integer value
  CEED_CONTEXT_FIELD_INT32 = 7,
  /// 64 bit integer value
  CEED_CONTEXT_FIELD_INT64 = 8,
  /// CeedSize value
  CEED_CONTEXT_FIELD_SIZE = 9,
  /// Boolean value
  CEED_CONTEXT_FIELD_BOOL = 10,
} CeedContextFieldType;
```

I am happy to change the ordering if anything else was intended.

This PR is currently a draft. The below tasks remain, before it's ready for full review:

- [ ] Extend `tests/t407-qfunction.c` to cover all ten field types (currently covers three)
- [ ] Confirm the enum ordering above
- [ ] `make prove-all` clean run including `examples/fluids`

These will be completed by tomorrow afternoon

Closes: #2004 

LLM/GenAI Disclosure:

I used Claude to help me understand the existing `CeedQFunctionContext`
 in `interface/ceed-qfunctioncontext.c` and `interface/ceed-operator.c` before
writing this change, and to review my diff against the design decisions from the issue thread -
it pointed out a few bugs (a `CeedSize`/`size_t` signature mismatch that failed to compile, a
stale `CeedContextFieldTypes[]` string table) which **I then fixed myself.** 

**ALL CODE SOURCE CHANGES IN THIS PR ARE WRITTEN BY ME.**

### DIFF
--- a/include/ceed/backend.h
+++ b/include/ceed/backend.h
@@ -409,18 +409,46 @@ CEED_EXTERN int  CeedQFunctionContextGetGenericRead(CeedQFunctionContext ctx, Ce
                                                     size_t *num_values, void *value);
 CEED_EXTERN int  CeedQFunctionContextRestoreGenericRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, CeedContextFieldType field_type,
                                                         void *value);
-CEED_EXTERN int  CeedQFunctionContextSetDouble(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, double *values);
-CEED_EXTERN int  CeedQFunctionContextGetDoubleRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values,
-                                                   const double **values);
-CEED_EXTERN int  CeedQFunctionContextRestoreDoubleRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const double **values);
-CEED_EXTERN int  CeedQFunctionContextSetInt32(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, int32_t *values);
-CEED_EXTERN int  CeedQFunctionContextGetInt32Read(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values,
-                                                  const int32_t **values);
-CEED_EXTERN int  CeedQFunctionContextRestoreInt32Read(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const int32_t **values);
-CEED_EXTERN int  CeedQFunctionContextSetBoolean(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, bool *values);
-CEED_EXTERN int  CeedQFunctionContextGetBooleanRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values,
-                                                    const bool **values);
-CEED_EXTERN int  CeedQFunctionContextRestoreBooleanRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const bool **values);
+CEED_EXTERN int CeedQFunctionContextSetCeedByte(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, char *values);  
+CEED_EXTERN int  CeedQFunctionContextGetCeedByteRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values,
+                                                       const char **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedByteRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const char **values); 
+CEED_EXTERN int CeedQFunctionContextSetCeedScalar(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, CeedScalar *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedScalarRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, 
+                                                      const CeedScalar **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedScalarRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const CeedScalar **values);
+CEED_EXTERN int CeedQFunctionContextSetCeedFloat(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, float *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedFloatRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, 
+                                                      const float **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedFloatRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const float **values);
+CEED_EXTERN int  CeedQFunctionContextSetCeedDouble(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, double *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedDoubleRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values,
+                                                        const double **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedDoubleRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const double **values);
+CEED_EXTERN int  CeedQFunctionContextSetCeedInt8(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, CeedInt8 *values);    
+CEED_EXTERN int  CeedQFunctionContextGetCeedInt8Read(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values,
+                                                      const CeedInt8 **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedInt8Read(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const CeedInt8 **values);
+CEED_EXTERN int  CeedQFunctionContextSetCeedInt(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, CeedInt *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedIntRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, 
+                                                      const CeedInt **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedIntRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const CeedInt **values);
+CEED_EXTERN int  CeedQFunctionContextSetCeedInt32(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, int32_t *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedInt32Read(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, 
+                                                      const int32_t **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedInt32Read(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const int32_t **values);
+CEED_EXTERN int  CeedQFunctionContextSetCeedInt64(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, int64_t *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedInt64Read(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, 
+                                                      const int64_t **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedInt64Read(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const int64_t **values);
+CEED_EXTERN int  CeedQFunctionContextSetCeedSize(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, CeedSize *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedSizeRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, 
+                                                      const CeedSize **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedSizeRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const CeedSize **values);
+CEED_EXTERN int  CeedQFunctionContextSetCeedBoolean(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, bool *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedBooleanRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values,
+                                                         const bool **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedBooleanRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const bool **values);
 CEED_EXTERN int  CeedQFunctionContextGetDataDestroy(CeedQFunctionContext ctx, CeedMemType *f_mem_type, CeedQFunctionContextDataDestroyUser *f);
 CEED_EXTERN int  CeedQFunctionContextReference(CeedQFunctionContext ctx);
 
--- a/include/ceed/ceed.h
+++ b/include/ceed/ceed.h
@@ -417,11 +417,25 @@ CEED_EXTERN int CeedQFunctionContextGetData(CeedQFunctionContext ctx, CeedMemTyp
 CEED_EXTERN int CeedQFunctionContextGetDataRead(CeedQFunctionContext ctx, CeedMemType mem_type, void *data);
 CEED_EXTERN int CeedQFunctionContextRestoreData(CeedQFunctionContext ctx, void *data);
 CEED_EXTERN int CeedQFunctionContextRestoreDataRead(CeedQFunctionContext ctx, void *data);
-CEED_EXTERN int CeedQFunctionContextRegisterDouble(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values,
+CEED_EXTERN int CeedQFunctionContextRegisterCeedByte(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values,
                                                    const char *field_description);
-CEED_EXTERN int CeedQFunctionContextRegisterInt32(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values,
+CEED_EXTERN int CeedQFunctionContextRegisterCeedScalar(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values,
+                                                   const char *field_description);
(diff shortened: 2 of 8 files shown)


ksd-2005:
Please also let me know, if I have to change any coding style( I followed the exact way in which other functions are written in the codebase, but i want to know if I can change anything else), or can improve by making the code concise anywhere, I would like to learn on ways to improve the code.

jeremylt:
Thanks for the PR. It's evening for me here in the EU so I'll look in the morning and let you know!

zatkins-dev:
A note on the ordering: I think I would prefer the order for the switch, it's a bit more intuitive to be. I also think that the enum needs a zero value -- we often use `CeedCalloc` to get memory and that automatically zeros the allocated memory. Enum values should be valid in that case.

zatkins-dev:
Also, thank you for the very clear explanation of AI use!

include/ceed/backend.h
```diff
@@ -409,18 +409,46 @@ CEED_EXTERN int  CeedQFunctionContextGetGenericRead(CeedQFunctionContext ctx, Ce
...
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedByteRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const char **values); 
+CEED_EXTERN int CeedQFunctionContextSetCeedScalar(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, CeedScalar *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedScalarRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, 
+                                                      const CeedScalar **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedScalarRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const CeedScalar **values);
+CEED_EXTERN int CeedQFunctionContextSetCeedFloat(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, float *values);
+CEED_EXTERN int  CeedQFunctionContextGetCeedFloatRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, size_t *num_values, 
+                                                      const float **values);
+CEED_EXTERN int  CeedQFunctionContextRestoreCeedFloatRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const float **values);
```
zatkins-dev, line 423: Just to note, only the `Ceed` types should have the `Ceed` prefix on the type name for these functions.

For example, `CeedScalar` should be as you have it, but `float` should be
```suggestion
CEED_EXTERN int  CeedQFunctionContextRestoreFloatRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const float **values);
```

zatkins-dev: Thats going to be a pretty big diff, so I'll let you handle it locally instead of me suggesting all of them.

zatkins-dev (commented):
This is a great start, a couple high-level things on a first look.

zatkins-dev:
I'm not sure that the Julia changes need to all be here -- some of that should probably be a separate PR.

include/ceed/backend.h
jeremylt: Yea, to clarify, I meant *only* the four types with `Ceed` in their name should have that added to the signature, so `CeedInt`, `CeedScalar`, `CeedSize`, and `CeedInt8`

include/ceed/ceed.h
```diff
@@ -417,11 +417,25 @@ CEED_EXTERN int CeedQFunctionContextGetData(CeedQFunctionContext ctx, CeedMemTyp
 CEED_EXTERN int CeedQFunctionContextGetDataRead(CeedQFunctionContext ctx, CeedMemType mem_type, void *data);
 CEED_EXTERN int CeedQFunctionContextRestoreData(CeedQFunctionContext ctx, void *data);
 CEED_EXTERN int CeedQFunctionContextRestoreDataRead(CeedQFunctionContext ctx, void *data);
-CEED_EXTERN int CeedQFunctionContextRegisterDouble(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values,
+CEED_EXTERN int CeedQFunctionContextRegisterCeedByte(CeedQFunctionContext ctx, const char *field_name, size_t field_offset, size_t num_values,
```
jeremylt, line 420: Note, these (and the other fn definitions and declarations) should use the ordering of the enum provided

include/ceed/types.h
```diff
@@ -274,12 +274,26 @@ typedef enum {
 /// Denotes type of data stored in a CeedQFunctionContext field
 /// @ingroup CeedQFunction
 typedef enum {
+  /// Byte value, C type of char
+  CEED_CONTEXT_FIELD_BYTE = 1,
```
jeremylt, line 278: As Zach said, I'd use the order I provided, starting at 0

julia/LibCEED.jl/src/generated/libceed_bindings.jl
```diff
@@ -1239,6 +1723,14 @@ function CeedQFunctionSetData(qf, data)
     ccall((:CeedQFunctionSetData, libceed), Cint, (CeedQFunction, Ptr{Cvoid}), qf, data)
 end
 
+function CeedQFunctionIsImmutable(qf, is_immutable)
```
jeremylt, line 1726: there's a bunch of other stuff added here - I'd only add the Julia changes that correspond to C API changes in this PR

julia/LibCEED.jl/src/generated/libceed_bindings.jl
```diff
@@ -1459,6 +2079,10 @@ function CeedHouseholderApplyQ(mat_A, mat_Q, tau, t_mode, m, n, k, row, col)
     ccall((:CeedHouseholderApplyQ, libceed), Cint, (Ptr{CeedScalar}, Ptr{CeedScalar}, Ptr{CeedScalar}, CeedTransposeMode, CeedInt, CeedInt, CeedInt, CeedInt, CeedInt), mat_A, mat_Q, tau, t_mode, m, n, k, row, col)
 end
 
+function CeedMatrixPseudoinverse(ceed, mat, m, n, mat_pinv)
```
jeremylt, line 2082: This definitely should not be exposed here

julia/LibCEED.jl/src/generated/libceed_bindings.jl
```diff
@@ -1443,6 +2055,14 @@ function CeedOperatorGetFallbackParentCeed(op, parent)
     ccall((:CeedOperatorGetFallbackParentCeed, libceed), Cint, (CeedOperator, Ptr{Ceed}), op, parent)
 end
 
+function CeedOperatorLinearAssembleQFunctionBuildOrUpdateFallback(op, build_objects, assembled, rstr, request)
+    ccall((:CeedOperatorLinearAssembleQFunctionBuildOrUpdateFallback, libceed), Cint, (CeedOperator, Bool, Ptr{CeedVector}, Ptr{CeedElemRestriction}, Ptr{CeedRequest}), op, build_objects, assembled, rstr, request)
+end
+
+function CeedOperatorAssembleSingle(op, offset, values)
```
jeremylt, line 2062: These also don't belong

julia/LibCEED.jl/src/generated/libceed_bindings.jl
```diff
@@ -1407,6 +2011,14 @@ function CeedOperatorGetNumArgs(op, num_args)
     ccall((:CeedOperatorGetNumArgs, libceed), Cint, (CeedOperator, Ptr{CeedInt}), op, num_args)
 end
 
+function CeedOperatorHasTensorBases(op, has_tensor_bases)
+    ccall((:CeedOperatorHasTensorBases, libceed), Cint, (CeedOperator, Ptr{Bool}), op, has_tensor_bases)
+end
+
+function CeedOperatorIsImmutable(op, is_immutable)
+    ccall((:CeedOperatorIsImmutable, libceed), Cint, (CeedOperator, Ptr{Bool}), op, is_immutable)
```
jeremylt, line 2019: And these are internal utilities that don't belong... really just only add the fn's from this PR and no others to be safe

interface/ceed-qfunctioncontext.c
```diff
@@ -387,6 +429,165 @@ int CeedQFunctionContextRestoreGenericRead(CeedQFunctionContext ctx, CeedContext
...
+  @param[in]  ctx         `CeedQFunctionContext`
+  @param[in]  field_label Label for field to restore
+  @param[out] values      Pointer to context values
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedQFunctionContextRestoreCeedByteRead(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, const char **values) {
+  CeedCheck(field_label, CeedQFunctionContextReturnCeed(ctx), CEED_ERROR_UNSUPPORTED, "Invalid field label");
```
jeremylt, line 482: These changes emphasize that probably this check should move from being duplicated in all of these functions to being in `CeedQFunctionContextRestoreGenericRead`

interface/ceed-qfunctioncontext.c
```diff
@@ -387,6 +429,165 @@ int CeedQFunctionContextRestoreGenericRead(CeedQFunctionContext ctx, CeedContext
...
+  @param[in,out] ctx         `CeedQFunctionContext`
+  @param[in]     field_label Label for field to set
+  @param[in]     values      Values to set
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedQFunctionContextSetCeedByte(CeedQFunctionContext ctx, CeedContextFieldLabel field_label, char * values) {
+  CeedCheck(field_label, CeedQFunctionContextReturnCeed(ctx), CEED_ERROR_UNSUPPORTED, "Invalid field label");
```
jeremylt, line 447: same comment here, all these replicated checks probably should go to `CeedQFunctionContextSetGeneric`

jeremylt:
@ksd-2005 how's this coming? Need anything clarified or any other sort of help?

jeremylt:
Just checking in @ksd-2005. We can help with the tidy up if you're busy with the semester

ksd-2005:
Hi @jeremylt @zatkins-dev , sorry for going quiet for so long. I had a run of exams and assignments over the past few weeks and couldn't get back to this until now. Thanks for your patience, and for the offer to help tidy things up if I got stuck.

I've pushed updates addressing the review:

- **Enum order:** now follows the switch order, starting at `CEED_CONTEXT_FIELD_BOOL = 0`, so zeroed memory from `CeedCalloc` is a valid type. This meant `CeedOperatorGetContextFieldLabel` could no longer use `type == 0` to mean "not set yet" when building a composite label (a BOOL field on the first sub-operator would skip the type check), so it now uses `field_found` for that instead.
- **Naming:** the `Ceed` prefix is only on `CeedInt`, `CeedInt8`, `CeedSize`, and `CeedScalar`, the plain C types use `Byte`, `Float`, `Double`, `Int32`, `Int64`, and `Boolean`.
- **Ordering:** declarations and definitions follow the enum order in `ceed.h`, `backend.h`, and both `.c` files.
- **Duplicated checks:** the `field_label` NULL check now lives once in `CeedQFunctionContextSetGeneric`/`GetGenericRead`/`RestoreGenericRead` instead of in every typed wrapper, matching how the `CeedOperator` side already does it.
- **Julia:** I dropped the full regeneration and the Clang.jl compat bump. The bindings file now only adds the enum values and the context-field functions from this PR. I also included the Boolean functions, since they were never added to the Julia bindings when `bool` support was added, I can drop them if you'd prefer.

**Possible bug I found, want to confirm before fixing:** in `CeedOperatorGetContextFieldLabel` (`interface/ceed-operator.c`), both composite-label error branches free the label and then read from it to build the error message:

```c
CeedCall(CeedFree(&new_field_label));
return CeedError(CeedOperatorReturnCeed(op), CEED_ERROR_INCOMPATIBLE, "Incompatible field types on sub-operator contexts. %s != %s",
                 CeedContextFieldTypes[new_field_label->type], CeedContextFieldTypes[new_field_label_i->type]);
```

Since `CeedFree` sets the pointer to `NULL`, `new_field_label->type` (and `new_field_label->num_values` in the second branch) dereferences `NULL`, so hitting a mismatch would crash `CEED_ERROR_INCOMPATIBLE`. `new_field_label->sub_labels` is also never freed on these paths. The branches are wrapped in `LCOV_EXCL`, so I don't think any test is testing them. My proposed fix is to save the value before freeing, and free `sub_labels` too:

```c
if (field_found && new_field_label->type != new_field_label_i->type) {
  CeedContextFieldType type = new_field_label->type;

  CeedCall(CeedFree(&new_field_label->sub_labels));
  CeedCall(CeedFree(&new_field_label));
  return CeedError(CeedOperatorReturnCeed(op), CEED_ERROR_INCOMPATIBLE, "Incompatible field types on sub-operator contexts. %s != %s",
                   CeedContextFieldTypes[type], CeedContextFieldTypes[new_field_label_i->type]);
}
```

(and the same pattern with `size_t num_values = new_field_label->num_values;` for the second branch). The composite-operator test I'm planning (BOOL field on the first sub-operator, DOUBLE with the same name on the second, expecting `CEED_ERROR_INCOMPATIBLE`) would hit this path, so it would need the fix to pass. Does this look right and also would you like it in this PR or a separate one?

Still to do before this is complete:

- [ ] Extend `t407-qfunction` to cover all ten field types
- [ ] Add a composite-operator case in `t525-operator` for mismatched field types with BOOL on the first sub-operator (depends on the fix above)
- [ ] Clean `make prove-all`

I'm planning to have the tests in by tomorrow. Let me know if anything above looks off.

jeremylt:
Good catch, that looks like the right fix. A separate PR makes sense since it should be a single commit fix we can merge that fast.

Thanks for the hard work so far!

zatkins-dev:
Overall looking good, thanks!

> My proposed fix is to save the value before freeing, and free `sub_labels` too:

This is actually a case of returning an error, which we're trying to eliminate altogether. 

I think a more appropriate fix would be to use `CeedCheck`:

```c
CeedCheck(!field_found || new_field_label->type == new_field_label_i->type, CeedOperatorReturnCeed(op), CEED_ERROR_INCOMPATIBLE, 
  "Incompatible field types on sub-operator contexts. %s != %s",
  CeedContextFieldTypes[type], CeedContextFieldTypes[new_field_label_i->type]);
```

jeremylt:
See, I like returning an error code personally for 'minor' errors like this because it means the repl session (Python and Julia) stays valid

zatkins-dev:
> See, I like returning an error code personally for 'minor' errors like this because it means the repl session (Python and Julia) stays valid

Hmm fair point; I feel like we could also enable exception-based error handling for REPL use instead of the abort handler. That would in theory let us hard-fail on some errors while soft-failing on others. Wayyyy outside scope for this though, I'm fine with the originally proposed change if you are.

jeremylt:
Yeah, that's been on the backburner forever since we don't have an active Python community, but the student projects may change that.

Issue #2009, opened 2026-08-19
CPU: even-odd (symmetric) factorization for the AVX basis-apply — benchmark + RFC


mohitt31:
Following a suggestion from @jedbrown to open an issue describing thoughts and plans — here's the RFC for adding an even-odd (symmetric) factorization path to the AVX tensor contraction kernel.

## Proposal

Add an even-odd factorization code path to `CeedTensorContractApply_Avx` in `backends/avx/ceed-avx-tensor.c`. For bases whose 1D interpolation/gradient matrix has the point symmetry of GLL/GL nodes (`B[i,j] = ±B[n-1-i, n-1-j]`), this precomputes symmetric and antisymmetric halves and uses them to halve the FMA count in the inner contraction loop.

## Evidence

I've benchmarked this in [mf-kernels](https://github.com/mohitt31/mf-kernels), a standalone set of five 1D tensor-contraction kernel variants (naive, pitfall, explicit AVX2, register-blocked AVX2, and even-odd AVX2). The isolated effect of even-odd vs register-blocked AVX2 on the bare 1D contraction:

| p | blocked (GFLOP/s) | even-odd (GFLOP/s) | speedup |
|---|---:|---:|---:|
| 5 | 20.82 | 21.73 | 1.04× |
| 7 | 21.58 | 25.92 | 1.20× |
| 9 | 21.09 | 28.36 | 1.34× |

(AMD EPYC Zen 3, single core, AVX2+FMA, gcc 13.3, `-O3 -march=native`. GFLOP/s uses the standard algorithm's FLOP count so the algorithmic saving shows as a wall-clock gain, not a smaller denominator.)

The gain grows with order because the ratio of halved FMAs to fixed overhead improves. At p=9, even-odd reaches ~28 GFLOP/s on the bare contraction.

The same repo also has a full 3D sum-factorized operator benchmark (BP1 mass + BP3 Poisson, p=1..8, structured hex mesh, ~0.5M DOFs) against MFEM native partial assembly and libCEED's `/cpu/self/avx/blocked` and `/cpu/self/xsmm/blocked` backends. This establishes the contraction kernels in a realistic operator context and confirms correctness (worst-case max-abs difference vs MFEM's own output: 3.8e-15 across all variants and orders).

## Where this does NOT help — honest caveats

[...]

PR #2010, opened 2026-08-21
Even-odd decomposition for tensor contractions

mohitt31:
## Summary

Exploit centro-symmetry of GLL-to-Gauss interpolation and gradient matrices to halve FMA count in 1D tensor contractions for the `/cpu/self/avx` backend.

- Split symmetric (interp) / antisymmetric (grad) basis matrices into half-size even/odd components
- Fold input, two half-contractions via existing blocked/remainder/single kernels, unfold output
- Lazy cache (8 entries) keyed by `(t_ptr, t_mode, B, J)` with content validation — no setup-time cost, correctly handles assembly paths that reuse pointers with different data
- Activates only when `min(B,J) >= 4` and matrix passes symmetry check; standard path otherwise
- Handles odd B/J, TRANSPOSE mode, both centro-symmetric and centro-antisymmetric cases

RFC: #2009

## Benchmark results

Native x86_64 (Google Cloud Shell, `gcc -O3 -march=native`), 3D tensor contractions, q=p+1, 10k iterations.

**ncomp=1 speedup** (opt/blocked baseline / avx/blocked even-odd):

| p | interp | grad | interpT | gradT |
|---|--------|------|---------|-------|
| 5 | 1.01x | 1.12x | 1.00x | 1.06x |
| 6 | 1.15x | 1.15x | 1.09x | 1.11x |
| 7 | 1.75x | 1.84x | 1.25x | 1.58x |
| 8 | 1.49x | 1.58x | 1.88x | 1.76x |
| 9 | 1.64x | 1.75x | 1.77x | 1.61x |
| 10 | 1.78x | 1.60x | 1.68x | 1.77x |
| 11 | 1.84x | 1.92x | 1.83x | 1.84x |
| 12 | 1.91x | 1.89x | 1.88x | 1.95x |

**ncomp=3 speedup**:

| p | interp | grad | interpT | gradT |
|---|--------|------|---------|-------|
| 5 | 1.08x | 1.14x | 1.10x | 1.12x |
| 6 | 1.23x | 1.21x | 1.16x | 1.17x |
| 7 | 1.74x | 1.90x | 1.28x | 1.60x |
| 8 | 1.52x | 1.61x | 1.85x | 1.64x |
| 9 | 1.56x | 1.70x | 1.54x | 1.62x |
| 11 | 1.81x | 1.88x | 1.67x | 1.87x |
| 12 | 1.83x | 1.84x | 1.85x | 1.77x |

Crossover at p~5-6, reaching ~1.9x at p=12. Below p=4 the even-odd path is gated off (min dim threshold).

## Test plan

- All existing basis and operator tests pass with zero regressions against both `/cpu/self/avx/blocked` and `/cpu/self/avx/serial`
- Tested interp and grad in forward and transpose modes across p=2..12, dim=1..3, ncomp=1 and ncomp=3
- Verified correctness for symmetric (interp_1d) and antisymmetric (grad_1d) matrices
- Edge cases: odd p (middle column fold), odd q (middle row unfold), C=1 serial path
- Operator assembly (t566) verified — cache validates matrix contents on every hit

## LLM usage disclosure

Used Claude (Anthropic) for implementation assistance: drafting the even-odd fold/unfold logic, cache data structures, symmetry detection, and debugging the middle-row unfolding for antisymmetric matrices with odd J. Also used for diagnosing the t566 assembly regression (stale cache when pointer reused with different contents).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Specifically, directories set with `CeedAddJitSourceRoot(ceed, "foo/bar")` will 
 - Add `CeedGetBuildConfiguration()` to access compilers, flags, and related information about the build environment.
 - Add support for full `CeedOperator` assembly for operators with multiple active fields with different bases for CPU backends and `/gpu/cuda/ref` and `/gpu/hip/gen` backends.
 - Add `CeedVectorFilter` to zero out components of a `CeedVector` that have absolute value below a specified threshold value.
+- Add `CeedOperatorSetEnableCudaGraph` for CUDA Graph capture/replay on `/gpu/cuda/gen` composite operators. Enabled by default; use `CEED_ENABLE_CUDA_GRAPH=0` to turn off.
 
 ### Examples
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ Specifically, directories set with `CeedAddJitSourceRoot(ceed, "foo/bar")` will 
 - Add support for full `CeedOperator` assembly for operators with multiple active fields with different bases for CPU backends and `/gpu/cuda/ref` and `/gpu/hip/gen` backends.
 - Add `CeedVectorFilter` to zero out components of a `CeedVector` that have absolute value below a specified threshold value.
 - Add `CeedOperatorSetEnableCudaGraph` for CUDA Graph capture/replay on `/gpu/cuda/gen` composite operators. Enabled by default; use `CEED_ENABLE_CUDA_GRAPH=0` to turn off.
+- Add even-odd (centro-symmetry) decomposition for tensor contractions, which halves the work when the 1D basis matrix is centro-symmetric.
+`CeedBasisGetEvenOddDecompositionInterp1D` and `CeedBasisGetEvenOddDecompositionGrad1D` report the symmetry of a basis and give the cached half-matrices, and `CeedTensorContractApplyEvenOdd` applies a contraction with them, so every CPU backend benefits rather than only `/cpu/self/avx/*`.
 
 ### Examples
 
--- a/backends/avx/ceed-avx-tensor.c
+++ b/backends/avx/ceed-avx-tensor.c
@@ -258,20 +258,19 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
 }
 
 //------------------------------------------------------------------------------
-// Even-Odd: Standard Dispatch (reused for half-contractions)
+// Tensor Contract - Common Sizes
 //------------------------------------------------------------------------------
-static inline int CeedTensorContract_Avx_StandardDispatch(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J,
-                                                          const CeedScalar *restrict t, CeedTransposeMode t_mode, const CeedInt add,
-                                                          const CeedScalar *restrict u, CeedScalar *restrict v) {
-  const CeedInt blk_size = 8;
-
-  if (C == 1) {
-    CeedTensorContract_Avx_Single(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8);
-  } else {
-    if (C >= blk_size) CeedTensorContract_Avx_Blocked(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8);
-    if (C % blk_size) CeedTensorContract_Avx_Remainder(contract, A, B, C, J, t, t_mode, add, u, v, 8, 8);
-  }
-  return CEED_ERROR_SUCCESS;
+static int CeedTensorContract_Avx_Blocked_4_8(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
+                                              CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
+  return CeedTensorContract_Avx_Blocked(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8);
+}
+static int CeedTensorContract_Avx_Remainder_8_8(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
+                                                CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
+  return CeedTensorContract_Avx_Remainder(contract, A, B, C, J, t, t_mode, add, u, v, 8, 8);
+}
+static int CeedTensorContract_Avx_Single_4_8(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
+                                             CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
+  return CeedTensorContract_Avx_Single(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8);
 }
 
 //------------------------------------------------------------------------------
@@ -368,7 +367,7 @@ static int CeedTensorContract_Avx_EvenOddPopulate(CeedTensorContract_Avx *data, 
   e->J         = J;
   e->B_half    = (B + 1) / 2;
   e->J_half    = (J + 1) / 2;
-  e->validated = false;
+  e->is_validated = false;
   e->t_even    = NULL;
   e->t_odd     = NULL;
   e->symmetry  = CeedTensorContract_Avx_DetectSymmetry(B, J, t, t_mode);
@@ -390,8 +389,8 @@ static int CeedTensorContract_Avx_EvenOddPopulate(CeedTensorContract_Avx *data, 
 // On the first hit for an entry, verify the pointer's contents haven't changed. Stable pointers (basis interp_1d/grad_1d) pass and are trusted for
 // all future calls. Scratch buffers (e.g. BTD_mat in operator assembly) that change contents are invalidated and skipped permanently.
 static int CeedTensorContract_Avx_EvenOddValidate(CeedTensorContract_Avx_EvenOddEntry *entry, const CeedScalar *t) {
-  if (entry->validated) return CEED_ERROR_SUCCESS;
-  entry->validated = true;
+  if (entry->is_validated) return CEED_ERROR_SUCCESS;
+  entry->is_validated = true;
 
   if (memcmp(entry->t_copy, t, (size_t)entry->B * entry->J * sizeof(CeedScalar))) {
     CeedCallBackend(CeedFree(&entry->t_even));
@@ -440,10 +439,23 @@ static int CeedTensorContract_Avx_EvenOdd(CeedTensorContract contract, CeedInt A
   }
 
   // Half-contractions: w_even = t_even * u_even, w_odd = t_odd * u_odd
-  memset(w_even, 0, A * J_half * C * sizeof(CeedScalar));
-  memset(w_odd, 0, A * J_half * C * sizeof(CeedScalar));
-  CeedTensorContract_Avx_StandardDispatch(contract, A, B_half, C, J_half, entry->t_even, CEED_NOTRANSPOSE, true, u_even, w_even);
-  CeedTensorContract_Avx_StandardDispatch(contract, A, B_half, C, J_half, entry->t_odd, CEED_NOTRANSPOSE, true, u_odd, w_odd);
+  memset(w_even, 0, sizeof(w_even));
+  memset(w_odd, 0, sizeof(w_odd));
+  if (C == 1) {
+    CeedTensorContract_Avx_Single_4_8(contract, A, B_half, C, J_half, entry->t_even, CEED_NOTRANSPOSE, true, u_even, w_even);
+    CeedTensorContract_Avx_Single_4_8(contract, A, B_half, C, J_half, entry->t_odd, CEED_NOTRANSPOSE, true, u_odd, w_odd);
+  } else {
+    const CeedInt blk_size = 8;
+
+    if (C >= blk_size) {
+      CeedTensorContract_Avx_Blocked_4_8(contract, A, B_half, C, J_half, entry->t_even, CEED_NOTRANSPOSE, true, u_even, w_even);
+      CeedTensorContract_Avx_Blocked_4_8(contract, A, B_half, C, J_half, entry->t_odd, CEED_NOTRANSPOSE, true, u_odd, w_odd);
+    }
+    if (C % blk_size) {
+      CeedTensorContract_Avx_Remainder_8_8(contract, A, B_half, C, J_half, entry->t_even, CEED_NOTRANSPOSE, true, u_even, w_even);
+      CeedTensorContract_Avx_Remainder_8_8(contract, A, B_half, C, J_half, entry->t_odd, CEED_NOTRANSPOSE, true, u_odd, w_odd);
+    }
+  }
 
   // Unfold output: recombine even/odd results into lower and upper halves
   for (CeedInt a = 0; a < A; a++) {
@@ -499,7 +511,14 @@ static int CeedTensorContractApply_Avx(CeedTensorContract contract, CeedInt A, C
   }
 
   // Standard path
-  CeedTensorContract_Avx_StandardDispatch(contract, A, B, C, J, t, t_mode, true, u, v);
+  if (C == 1) {
+    CeedTensorContract_Avx_Single_4_8(contract, A, B, C, J, t, t_mode, true, u, v);
+  } else {
+    const CeedInt blk_size = 8;
+
+    if (C >= blk_size) CeedTensorContract_Avx_Blocked_4_8(contract, A, B, C, J, t, t_mode, true, u, v);
+    if (C % blk_size) CeedTensorContract_Avx_Remainder_8_8(contract, A, B, C, J, t, t_mode, true, u, v);
+  }
   return CEED_ERROR_SUCCESS;
 }
 
--- a/backends/avx/ceed-avx-tensor.c
+++ b/backends/avx/ceed-avx-tensor.c
@@ -258,15 +258,18 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
 //------------------------------------------------------------------------------
 static int CeedTensorContract_Avx_Blocked_4_8(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
                                               CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
-  return CeedTensorContract_Avx_Blocked(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8);
+  CeedCallBackend(CeedTensorContract_Avx_Blocked(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8));
+  return CEED_ERROR_SUCCESS;
 }
 static int CeedTensorContract_Avx_Remainder_8_8(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
                                                 CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
-  return CeedTensorContract_Avx_Remainder(contract, A, B, C, J, t, t_mode, add, u, v, 8, 8);
+  CeedCallBackend(CeedTensorContract_Avx_Remainder(contract, A, B, C, J, t, t_mode, add, u, v, 8, 8));
+  return CEED_ERROR_SUCCESS;
 }
 static int CeedTensorContract_Avx_Single_4_8(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
                                              CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
-  return CeedTensorContract_Avx_Single(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8);
+  CeedCallBackend(CeedTensorContract_Avx_Single(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8));
+  return CEED_ERROR_SUCCESS;
 }
 
 //------------------------------------------------------------------------------
--- a/backends/avx/ceed-avx-tensor.c
+++ b/backends/avx/ceed-avx-tensor.c
@@ -343,55 +343,61 @@ static int CeedTensorContract_Avx_ComputeHalfMatrices(CeedInt B, CeedInt J, cons
 }
 
 //------------------------------------------------------------------------------
-// Even-Odd: Cache Lookup and Populate
+// Even-Odd: Lookup and Populate
 //------------------------------------------------------------------------------
-static CeedTensorContract_Avx_CacheEntry *CeedTensorContract_Avx_CacheLookup(CeedTensorContract_Avx *data, const CeedScalar *t,
-                                                                             CeedTransposeMode t_mode, CeedInt B, CeedInt J) {
-  for (CeedInt i = 0; i < data->n_cached; i++) {
-    if (data->cache[i].t_ptr == t && data->cache[i].t_mode == t_mode && data->cache[i].B == B && data->cache[i].J == J) return &data->cache[i];
+static CeedTensorContract_Avx_EvenOddEntry *CeedTensorContract_Avx_EvenOddLookup(CeedTensorContract_Avx *data, const CeedScalar *t,
+                                                                                  CeedTransposeMode t_mode, CeedInt B, CeedInt J) {
+  for (CeedInt i = 0; i < data->n_entries; i++) {
+    if (data->entries[i].t_ptr == t && data->entries[i].t_mode == t_mode && data->entries[i].B == B && data->entries[i].J == J) {
+      return &data->entries[i];
+    }
   }
   return NULL;
 }
 
-// The t matrix contents may change between applies at the same pointer (e.g. operator assembly rewrites a scratch buffer in place), so the entry
-// stores a copy of t and must be refreshed whenever the contents no longer match.
-static int CeedTensorContract_Avx_CacheEntryRefresh(CeedTensorContract_Avx_CacheEntry *e, const CeedScalar *t) {
-  memcpy(e->t_copy, t, (size_t)e->B * e->J * sizeof(CeedScalar));
-  e->symmetry = CeedTensorContract_Avx_DetectSymmetry(e->B, e->J, t, e->t_mode);
+static int CeedTensorContract_Avx_EvenOddPopulate(CeedTensorContract_Avx *data, const CeedScalar *t, CeedTransposeMode t_mode, CeedInt B, CeedInt J,
+                                                   CeedTensorContract_Avx_EvenOddEntry **entry) {
+  *entry = NULL;
+  if (data->n_entries >= CEED_AVX_EVEN_ODD_MAX_ENTRIES) return CEED_ERROR_SUCCESS;
+
+  CeedTensorContract_Avx_EvenOddEntry *e = &data->entries[data->n_entries];
+
+  e->t_ptr     = t;
+  e->t_mode    = t_mode;
+  e->B         = B;
+  e->J         = J;
+  e->B_half    = (B + 1) / 2;
+  e->J_half    = (J + 1) / 2;
+  e->validated = false;
+  e->t_even    = NULL;
+  e->t_odd     = NULL;
+  e->symmetry  = CeedTensorContract_Avx_DetectSymmetry(B, J, t, t_mode);
+
+  CeedCallBackend(CeedMalloc(B * J, &e->t_copy));
+  memcpy(e->t_copy, t, (size_t)B * J * sizeof(CeedScalar));
 
   if (e->symmetry != 0) {
-    if (!e->t_even) {
-      CeedCallBackend(CeedMalloc(e->J_half * e->B_half, &e->t_even));
-      CeedCallBackend(CeedMalloc(e->J_half * e->B_half, &e->t_odd));
-    }
-    CeedTensorContract_Avx_ComputeHalfMatrices(e->B, e->J, t, e->t_mode, e->B_half, e->J_half, e->t_even, e->t_odd);
-  } else {
-    CeedCallBackend(CeedFree(&e->t_even));
-    CeedCallBackend(CeedFree(&e->t_odd));
+    CeedCallBackend(CeedMalloc(e->J_half * e->B_half, &e->t_even));
+    CeedCallBackend(CeedMalloc(e->J_half * e->B_half, &e->t_odd));
+    CeedTensorContract_Avx_ComputeHalfMatrices(B, J, t, t_mode, e->B_half, e->J_half, e->t_even, e->t_odd);
   }
+
+  data->n_entries++;
+  *entry = e;
   return CEED_ERROR_SUCCESS;
 }
 
-static int CeedTensorContract_Avx_CachePopulate(CeedTensorContract_Avx *data, const CeedScalar *t, CeedTransposeMode t_mode, CeedInt B, CeedInt J,
-                                                CeedTensorContract_Avx_CacheEntry **entry) {
-  *entry = NULL;
-  if (data->n_cached >= CEED_AVX_EVEN_ODD_CACHE_MAX) return CEED_ERROR_SUCCESS;
+// On the first hit for an entry, verify the pointer's contents haven't changed. Stable pointers (basis interp_1d/grad_1d) pass and are trusted for
+// all future calls. Scratch buffers (e.g. BTD_mat in operator assembly) that change contents are invalidated and skipped permanently.
+static int CeedTensorContract_Avx_EvenOddValidate(CeedTensorContract_Avx_EvenOddEntry *entry, const CeedScalar *t) {
+  if (entry->validated) return CEED_ERROR_SUCCESS;
+  entry->validated = true;
 
-  CeedTensorContract_Avx_CacheEntry *e = &data->cache[data->n_cached];
-
-  e->t_ptr  = t;
-  e->t_mode = t_mode;
-  e->B      = B;
-  e->J      = J;
-  e->B_half = (B + 1) / 2;
-  e->J_half = (J + 1) / 2;
-  e->t_even = NULL;
-  e->t_odd  = NULL;
-  CeedCallBackend(CeedMalloc(B * J, &e->t_copy));
-  CeedCallBackend(CeedTensorContract_Avx_CacheEntryRefresh(e, t));
-
-  data->n_cached++;
-  *entry = e;
+  if (memcmp(entry->t_copy, t, (size_t)entry->B * entry->J * sizeof(CeedScalar))) {
+    CeedCallBackend(CeedFree(&entry->t_even));
+    CeedCallBackend(CeedFree(&entry->t_odd));
+    entry->symmetry = 0;
+  }
   return CEED_ERROR_SUCCESS;
 }
 
@@ -399,7 +405,7 @@ static int CeedTensorContract_Avx_CachePopulate(CeedTensorContract_Avx *data, co
 // Even-Odd: Apply
 //------------------------------------------------------------------------------
 static int CeedTensorContract_Avx_EvenOdd(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J,
-                                          const CeedTensorContract_Avx_CacheEntry *entry, const CeedScalar *restrict u, CeedScalar *restrict v) {
+                                          const CeedTensorContract_Avx_EvenOddEntry *entry, const CeedScalar *restrict u, CeedScalar *restrict v) {
   const CeedInt B_half = entry->B_half;
   const CeedInt J_half = entry->J_half;
   const int     sym    = entry->symmetry;
@@ -410,59 +416,59 @@ static int CeedTensorContract_Avx_EvenOdd(CeedTensorContract contract, CeedInt A
   memset(u_even, 0, sizeof(u_even));
   memset(u_odd, 0, sizeof(u_odd));
 
-  // Fold input along b
+  // Fold input: combine lower and upper halves into even/odd components
   for (CeedInt a = 0; a < A; a++) {
     for (CeedInt b = 0; b < B / 2; b++) {
-      const CeedInt lo = (a * B + b) * C;
-      const CeedInt hi = (a * B + (B - 1 - b)) * C;
-      const CeedInt fo = (a * B_half + b) * C;
+      const CeedInt idx_lower  = (a * B + b) * C;
+      const CeedInt idx_upper  = (a * B + (B - 1 - b)) * C;
+      const CeedInt idx_folded = (a * B_half + b) * C;
 
       for (CeedInt c = 0; c < C; c++) {
-        u_even[fo + c] = u[lo + c] + u[hi + c];
-        u_odd[fo + c]  = u[lo + c] - u[hi + c];
+        u_even[idx_folded + c] = u[idx_lower + c] + u[idx_upper + c];
+        u_odd[idx_folded + c]  = u[idx_lower + c] - u[idx_upper + c];
       }
     }
     if (B % 2) {
-      const CeedInt mid_in  = (a * B + B / 2) * C;
-      const CeedInt mid_out = (a * B_half + B / 2) * C;
+      const CeedInt idx_mid_in  = (a * B + B / 2) * C;
+      const CeedInt idx_mid_out = (a * B_half + B / 2) * C;
 
       for (CeedInt c = 0; c < C; c++) {
-        u_even[mid_out + c] = u[mid_in + c];
-        u_odd[mid_out + c]  = (CeedScalar)0.0;
+        u_even[idx_mid_out + c] = u[idx_mid_in + c];
+        u_odd[idx_mid_out + c]  = (CeedScalar)0.0;
       }
     }
   }
 
-  // Half-contractions: w_e = t_e * u_e, w_o = t_o * u_o
+  // Half-contractions: w_even = t_even * u_even, w_odd = t_odd * u_odd
   memset(w_even, 0, A * J_half * C * sizeof(CeedScalar));
   memset(w_odd, 0, A * J_half * C * sizeof(CeedScalar));
   CeedTensorContract_Avx_StandardDispatch(contract, A, B_half, C, J_half, entry->t_even, CEED_NOTRANSPOSE, true, u_even, w_even);
   CeedTensorContract_Avx_StandardDispatch(contract, A, B_half, C, J_half, entry->t_odd, CEED_NOTRANSPOSE, true, u_odd, w_odd);
 
-  // Unfold output along j
+  // Unfold output: recombine even/odd results into lower and upper halves
   for (CeedInt a = 0; a < A; a++) {
     for (CeedInt j = 0; j < J / 2; j++) {
-      const CeedInt eo   = (a * J_half + j) * C;
-      const CeedInt v_lo = (a * J + j) * C;
-      const CeedInt v_hi = (a * J + (J - 1 - j)) * C;
+      const CeedInt idx_half      = (a * J_half + j) * C;
+      const CeedInt idx_out_lower = (a * J + j) * C;
+      const CeedInt idx_out_upper = (a * J + (J - 1 - j)) * C;
 
       if (sym == 1) {
         for (CeedInt c = 0; c < C; c++) {
-          v[v_lo + c] += w_even[eo + c] + w_odd[eo + c];
-          v[v_hi + c] += w_even[eo + c] - w_odd[eo + c];
+          v[idx_out_lower + c] += w_even[idx_half + c] + w_odd[idx_half + c];
+          v[idx_out_upper + c] += w_even[idx_half + c] - w_odd[idx_half + c];
         }
       } else {
         for (CeedInt c = 0; c < C; c++) {
-          v[v_lo + c] += w_even[eo + c] + w_odd[eo + c];
-          v[v_hi + c] += -w_even[eo + c] + w_odd[eo + c];
+          v[idx_out_lower + c] += w_even[idx_half + c] + w_odd[idx_half + c];
+          v[idx_out_upper + c] += -w_even[idx_half + c] + w_odd[idx_half + c];
         }
       }
     }
     if (J % 2) {
-      const CeedInt eo    = (a * J_half + J / 2) * C;
-      const CeedInt v_mid = (a * J + J / 2) * C;
+      const CeedInt idx_half    = (a * J_half + J / 2) * C;
+      const CeedInt idx_out_mid = (a * J + J / 2) * C;
 
-      for (CeedInt c = 0; c < C; c++) v[v_mid + c] += w_even[eo + c] + w_odd[eo + c];
+      for (CeedInt c = 0; c < C; c++) v[idx_out_mid + c] += w_even[idx_half + c] + w_odd[idx_half + c];
     }
   }
   return CEED_ERROR_SUCCESS;
@@ -482,12 +488,12 @@ static int CeedTensorContractApply_Avx(CeedTensorContract contract, CeedInt A, C
 
   CeedCallBackend(CeedTensorContractGetData(contract, &data));
   if (data && B >= CEED_AVX_EVEN_ODD_MIN_DIM && J >= CEED_AVX_EVEN_ODD_MIN_DIM) {
-    CeedTensorContract_Avx_CacheEntry *entry = CeedTensorContract_Avx_CacheLookup(data, t, t_mode, B, J);
+    CeedTensorContract_Avx_EvenOddEntry *entry = CeedTensorContract_Avx_EvenOddLookup(data, t, t_mode, B, J);
 
     if (!entry) {
-      CeedCallBackend(CeedTensorContract_Avx_CachePopulate(data, t, t_mode, B, J, &entry));
-    } else if (memcmp(entry->t_copy, t, (size_t)B * J * sizeof(CeedScalar))) {
-      CeedCallBackend(CeedTensorContract_Avx_CacheEntryRefresh(entry, t));
+      CeedCallBackend(CeedTensorContract_Avx_EvenOddPopulate(data, t, t_mode, B, J, &entry));
+    } else {
+      CeedCallBackend(CeedTensorContract_Avx_EvenOddValidate(entry, t));
     }
     if (entry && entry->t_even) return CeedTensorContract_Avx_EvenOdd(contract, A, B, C, J, entry, u, v);
   }
@@ -505,10 +511,10 @@ static int CeedTensorContractDestroy_Avx(CeedTensorContract contract) {
 
   CeedCallBackend(CeedTensorContractGetData(contract, &data));
   if (data) {
-    for (CeedInt i = 0; i < data->n_cached; i++) {
-      CeedCallBackend(CeedFree(&data->cache[i].t_copy));
-      CeedCallBackend(CeedFree(&data->cache[i].t_even));
-      CeedCallBackend(CeedFree(&data->cache[i].t_odd));
+    for (CeedInt i = 0; i < data->n_entries; i++) {
+      CeedCallBackend(CeedFree(&data->entries[i].t_copy));
+      CeedCallBackend(CeedFree(&data->entries[i].t_even));
+      CeedCallBackend(CeedFree(&data->entries[i].t_odd));
     }
     CeedCallBackend(CeedFree(&data));
   }
--- a/backends/avx/ceed-avx-tensor.c
+++ b/backends/avx/ceed-avx-tensor.c
@@ -353,6 +353,25 @@ static CeedTensorContract_Avx_CacheEntry *CeedTensorContract_Avx_CacheLookup(Cee
   return NULL;
 }
 
+// The t matrix contents may change between applies at the same pointer (e.g. operator assembly rewrites a scratch buffer in place), so the entry
+// stores a copy of t and must be refreshed whenever the contents no longer match.
+static int CeedTensorContract_Avx_CacheEntryRefresh(CeedTensorContract_Avx_CacheEntry *e, const CeedScalar *t) {
+  memcpy(e->t_copy, t, (size_t)e->B * e->J * sizeof(CeedScalar));
+  e->symmetry = CeedTensorContract_Avx_DetectSymmetry(e->B, e->J, t, e->t_mode);
+
+  if (e->symmetry != 0) {
+    if (!e->t_even) {
+      CeedCallBackend(CeedMalloc(e->J_half * e->B_half, &e->t_even));
+      CeedCallBackend(CeedMalloc(e->J_half * e->B_half, &e->t_odd));
+    }
+    CeedTensorContract_Avx_ComputeHalfMatrices(e->B, e->J, t, e->t_mode, e->B_half, e->J_half, e->t_even, e->t_odd);
+  } else {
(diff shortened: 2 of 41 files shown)


backends/avx/ceed-avx-tensor.c
```diff
@@ -254,40 +258,259 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
...
+static int CeedTensorContract_Avx_DetectSymmetry(CeedInt B, CeedInt J, const CeedScalar *t, CeedTransposeMode t_mode) {
+  CeedInt t_stride_0 = B, t_stride_1 = 1;
+
+  if (t_mode == CEED_TRANSPOSE) {
+    t_stride_0 = 1;
+    t_stride_1 = J;
+  }
+
+  const CeedScalar tol = 1e-14;
+  int              sym = 0;
```
jeremylt, line 288: First thing - variable names should be words

backends/avx/ceed-avx-tensor.c
```diff
@@ -254,40 +258,259 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
...
+//------------------------------------------------------------------------------
+// Even-Odd: Cache Lookup and Populate
 //------------------------------------------------------------------------------
-static int CeedTensorContract_Avx_Blocked_4_8(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
-                                              CeedTransposeMode t_mode, const CeedInt add, const CeedScalar *restrict u, CeedScalar *restrict v) {
-  return CeedTensorContract_Avx_Blocked(contract, A, B, C, J, t, t_mode, add, u, v, 4, 8);
+static CeedTensorContract_Avx_CacheEntry *CeedTensorContract_Avx_CacheLookup(CeedTensorContract_Avx *data, const CeedScalar *t,
+                                                                             CeedTransposeMode t_mode, CeedInt B, CeedInt J) {
+  for (CeedInt i = 0; i < data->n_cached; i++) {
+    if (data->cache[i].t_ptr == t && data->cache[i].t_mode == t_mode && data->cache[i].B == B && data->cache[i].J == J) return &data->cache[i];
```
jeremylt: What is this?

backends/avx/ceed-avx-tensor.c
```diff
@@ -254,40 +258,259 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
...
+  }
+  return NULL;
+}
+
+// The t matrix contents may change between applies at the same pointer (e.g. operator assembly rewrites a scratch buffer in place), so the entry
```
jeremylt: Rewriting a scratch buffer should have no effect on application of a basis. What is the actual problem that happened so we can directly fix it?

backends/avx/ceed-avx-tensor.c
```diff
@@ -254,40 +258,259 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
...
+  CeedScalar u_even[A * B_half * C], u_odd[A * B_half * C];
+  CeedScalar w_even[A * J_half * C], w_odd[A * J_half * C];
+
+  memset(u_even, 0, sizeof(u_even));
+  memset(u_odd, 0, sizeof(u_odd));
+
+  // Fold input along b
+  for (CeedInt a = 0; a < A; a++) {
+    for (CeedInt b = 0; b < B / 2; b++) {
+      const CeedInt lo = (a * B + b) * C;
```
jeremylt: real words here especially please

backends/avx/ceed-avx.h
```diff
@@ -8,5 +8,24 @@
...
+
+typedef struct {
+  const CeedScalar *t_ptr;
+  CeedTransposeMode t_mode;
+  CeedInt           B, J, B_half, J_half;
+  int               symmetry;
+  CeedScalar       *t_copy;
+  CeedScalar       *t_even;
+  CeedScalar       *t_odd;
+} CeedTensorContract_Avx_CacheEntry;
```
jeremylt: Is there a more descriptive name than 'cache'?

jeremylt (changes requested):
First comments added.

The performance comparison seems to compare opt/blocked to avx/blocked? You should compare avx/blocked on main to avx/blocked on this branch. Comparing to the opt in this branch checks how much handwritten AVX instructions help against whatever the compiler emits itself. 

You should compare avx/serial on main and this branch.

let's please talk human to human for discussion in this review

backends/avx/ceed-avx-tensor.c
mohitt31: Fixed, renamed variables throughout to full words.

backends/avx/ceed-avx-tensor.c
mohitt31: It's the even-odd entry lookup, matches a (t_ptr, t_mode, B, J) tuple against stored entries. Renamed to EvenOddLookup and added a section header.

backends/avx/ceed-avx-tensor.c
mohitt31: Right, basis matrices (interp_1d, grad_1d) have stable contents so pointer-keyed lookup is fine for those.

The issue was the assembly path in ceed-preconditioning.c (line 1006). It calls CeedTensorContractApply with a scratch BTD_mat buffer that gets rewritten each (comp_in, comp_out) iteration. In t566 (p=3, q=3, dim=2, num_comp=2), the first 9x9 BTD_mat happened to be centro-symmetric, so the entry stored half-matrices for it. Next iteration same pointer, different contents, stale hit.

Replaced the per-call memcmp with one-time validation: on the first hit, verify contents still match. Stable pointers pass once and are trusted after that (zero overhead). Scratch buffers that change contents get permanently invalidated.

backends/avx/ceed-avx-tensor.c
mohitt31: Done. idx_lower, idx_upper, idx_folded in fold. idx_half, idx_out_lower, idx_out_upper, idx_out_mid in unfold.

backends/avx/ceed-avx.h
mohitt31: Renamed to EvenOddEntry with n_entries and entries[] in the container struct. Functions are EvenOddLookup, EvenOddPopulate, EvenOddValidate. Macro is CEED_AVX_EVEN_ODD_MAX_ENTRIES.

backends/avx/ceed-avx-tensor.c
jeremylt: was this comment written by a human as I requested?

@zatkins-dev design question for you here - something about this coupling of the TensorContraction back up to its Basis so explicitly doesn't sit right but I don't see any obvious alternative

backends/avx/ceed-avx.h
```diff
@@ -8,5 +8,25 @@
...
+
+#define CEED_AVX_EVEN_ODD_MAX_ENTRIES 8
+#define CEED_AVX_EVEN_ODD_MIN_DIM 4
+
+typedef struct {
+  const CeedScalar *t_ptr;
+  CeedTransposeMode t_mode;
+  CeedInt           B, J, B_half, J_half;
+  int               symmetry;
+  bool              validated;
```
jeremylt: this would be `is_validated` per our convention

backends/avx/ceed-avx-tensor.c
```diff
@@ -254,40 +258,265 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
...
+      } else if (sym == 1) {
+        if (diff_sym / scale >= tol) return 0;
+      } else {
+        if (diff_ant / scale >= tol) return 0;
+      }
+    }
+  }
+  return sym;
 }
-static int CeedTensorContract_Avx_Single_4_8(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t,
```
jeremylt, line 267: This function was to help the compiler recognize a fixed size in would see frequently. Is there a particular reason it was removed in this work?

backends/avx/ceed-avx-tensor.c
```diff
@@ -254,40 +258,265 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
...
+static int CeedTensorContract_Avx_EvenOdd(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J,
+                                          const CeedTensorContract_Avx_EvenOddEntry *entry, const CeedScalar *restrict u, CeedScalar *restrict v) {
+  const CeedInt B_half = entry->B_half;
+  const CeedInt J_half = entry->J_half;
+  const int     sym    = entry->symmetry;
+
+  CeedScalar u_even[A * B_half * C], u_odd[A * B_half * C];
+  CeedScalar w_even[A * J_half * C], w_odd[A * J_half * C];
+
+  memset(u_even, 0, sizeof(u_even));
```
jeremylt, line 415: why not just declare these to be arrays of 0s instead of using raw memset?

backends/avx/ceed-avx-tensor.c
```diff
@@ -254,40 +258,265 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
...
-
   if (!add) {
     for (CeedInt q = 0; q < A * J * C; q++) v[q] = (CeedScalar)0.0;
   }
 
-  if (C == 1) {
-    // Serial C=1 Case
-    CeedTensorContract_Avx_Single_4_8(contract, A, B, C, J, t, t_mode, true, u, v);
-  } else {
-    // Blocks of 8 columns
```
jeremylt, line 287: this structure that the compiler can leverage still exists... I don't see why removing this would be helpful

jeremylt (changes requested):
thanks, a few more comments

backends/avx/ceed-avx-tensor.c
mohitt31: First off, honestly, I used Claude to help write my last reply. That was my mistake, and I will just use my own words from now on.

About the t566 test failure (where results were exactly 1/3 of what they should be): the issue was in ceed-preconditioning.c around line 1006.

The BTD_mat buffer is reused in a loop. My cache was only checking the memory pointer, so it accidentally kept using old cached data even when the actual data inside the buffer changed.

I fixed this by replacing the per-call data check with a one-time validation on the first cache hit.

If the data matches on that first try, it is marked as trusted (zero overhead). If it fails, it completely skips the cache. This keeps static matrices (like interp_1d) fast, while safely ignoring scratch buffers (like BTD_mat).

backends/avx/ceed-avx-tensor.c
```diff
@@ -254,40 +258,265 @@ static inline int CeedTensorContract_Avx_Single(CeedTensorContract contract, Cee
...
+      return &data->entries[i];
+    }
+  }
+  return NULL;
+}
+
+static int CeedTensorContract_Avx_EvenOddPopulate(CeedTensorContract_Avx *data, const CeedScalar *t, CeedTransposeMode t_mode, CeedInt B, CeedInt J,
+                                                   CeedTensorContract_Avx_EvenOddEntry **entry) {
+  *entry = NULL;
+  if (data->n_entries >= CEED_AVX_EVEN_ODD_MAX_ENTRIES) return CEED_ERROR_SUCCESS;
```
zatkins-dev, line 360: If we're already allocating here and cache entries are pretty light, we may as well let this cache grow arbitrarily large. We should just realloc it if we need to expand path the default size.

backends/avx/ceed-avx-tensor.c
zatkins-dev: I'm inclined to say that it's alright, since the backend controls both. I agree that the structure is a bit odd. I need to look closer at the current relationship between the basis and tensor contraction. Is is one to one? Or one contraction per Ceed context?

I'll look in detail tomorrow.

jeremylt: yea, that makes sense now.... but I feel like there has to be a simpler solution. 

I think at its core the complication is that we use a contraction for Basis applications and for general purpose work. Doing aBasis application alone doesn't need this safeguard.

This dual usage is mixing poorly with the TensorContraction holding Basis data in this change, I think. But I don't see an obvious design fix.

Maybe the detection of the symmetry and related work on the Basis matrices should be in the Basis object. Could have the Basis query the TensorContraction if it supports this split and if so the Basis stores the required info and calls TensorContractionApplyEvenOdd? That keeps the data in the object that actually owns it and makes this logic in the TensorContraction simpler.

backends/avx/ceed-avx.h
mohitt31: Fixed! Renamed it to is_validated.

backends/avx/ceed-avx-tensor.c
mohitt31: My bad on removing these, I've put them back. I originally thought the compiler would just figure out the constants through inlining, but I was wrong. The compiler actually needs to see those exact sizes right at the call site to fully unroll and vectorize the loops. The wrappers lock in those numbers (like 4,8 and 8,8) so the compiler can do its job.

backends/avx/ceed-avx-tensor.c
mohitt31: I really wanted to use = {0}, but since A, B_half, and C are runtime values, these are Variable-Length Arrays (VLAs). C doesn't let us initialize VLAs that way - GCC just throws a "variable-sized object may not be initialized" error. So I kept the memset, but I changed it to use sizeof(array) instead of doing the math manually. Much cleaner now!

backends/avx/ceed-avx-tensor.c
mohitt31: Put this back for the exact same reason as the wrapper functions. StandardDispatch was passing the sizes as variables, which basically hid the constants from the compiler. Now it calls the wrappers directly.

backends/avx/ceed-avx-tensor.c
mohitt31: Good catch, that makes sense. I'll change this to a dynamically growing array. I'll start with a small default size (like 4) and just double it with CeedRealloc whenever it fills up. I'll include this in my next push.

backends/avx/ceed-avx-tensor.c
mohitt31: Just to add a bit of context here: right now, it's exactly one tensor contraction per basis. The cache lives inside the contraction's backend data, so it naturally stays scoped to that specific basis. But I'm totally open to restructuring if you see a cleaner way to handle it!

backends/avx/ceed-avx-tensor.c
jeremylt: ugh, that's right

How big can these get? Calloc/Free may be a better fit. I've been trying to keep VLAs to small allocations and indexing casts only

backends/avx/ceed-avx-tensor.c
mohitt31: That's actually a really clean way to separate things. If the Basis handles the symmetry check and holds the half-matrices, then the TensorContraction won't need any of this caching or validation logic at all. It would just take the pre-computed even and odd matrices directly.

I can definitely prototype this! The Basis would check the symmetry right when it's created, store the t_even and t_odd matrices, and then just call a new TensorContractionApplyEvenOdd function. Would you like me to try this approach in the next revision?

backends/avx/ceed-avx-tensor.c
mohitt31: Really good point. If we are dealing with a 3D problem where C gets large, these VLAs could take up way too much space on the stack. I'll swap them out and use CeedCalloc and CeedFree for the main working arrays instead. I'll only keep VLAs for really small, fixed-size stuff where we know it's safe.

jeremylt:
Some of these comments still feel LLM generated. Can you please not use any LLMs to generate any replies to my human generated questions and comments? You are permitted to use LLMs in your personal development process as long as you disclose all of the usage, but in my personal review process I need to talk to the human who is responsible for the changes to make sure we're making the right choices for the codebase

backends/avx/ceed-avx-tensor.c
zatkins-dev: These (and many other) variables are still not full words.

backends/avx/ceed-avx-tensor.c
zatkins-dev: Yeah, I think that we only want to do this extra work for stable inputs, i.e. basis applications. Related question: is there a reason why we only want to do this with AVX? None of the added code uses AVX intrinsics, would it also have a benefit for /cpu/self/opt?

backends/avx/ceed-avx-tensor.c
```diff
@@ -269,34 +273,284 @@ static int CeedTensorContract_Avx_Single_4_8(CeedTensorContract contract, CeedIn
...
+    }
+    if (entry && entry->t_even) return CeedTensorContract_Avx_EvenOdd(contract, A, B, C, J, entry, u, v);
+  }
+
+  // Standard path
   if (C == 1) {
-    // Serial C=1 Case
     CeedTensorContract_Avx_Single_4_8(contract, A, B, C, J, t, t_mode, true, u, v);
   } else {
-    // Blocks of 8 columns
```
zatkins-dev, line 287: Comments should not be removed

backends/avx/ceed-avx-tensor.c
```diff
@@ -269,34 +273,284 @@ static int CeedTensorContract_Avx_Single_4_8(CeedTensorContract contract, CeedIn
...
+    const CeedInt blk_size = 8;
+
     if (C >= blk_size) CeedTensorContract_Avx_Blocked_4_8(contract, A, B, C, J, t, t_mode, true, u, v);
-    // Remainder of columns
```
zatkins-dev, line 289: Comments should not be removed

backends/avx/ceed-avx-tensor.c
```diff
@@ -269,34 +273,284 @@ static int CeedTensorContract_Avx_Single_4_8(CeedTensorContract contract, CeedIn
...
+//------------------------------------------------------------------------------
+static int CeedTensorContract_Avx_DetectSymmetry(CeedInt B, CeedInt J, const CeedScalar *t, CeedTransposeMode t_mode) {
+  CeedInt t_stride_0 = B, t_stride_1 = 1;
+
+  if (t_mode == CEED_TRANSPOSE) {
+    t_stride_0 = 1;
+    t_stride_1 = J;
+  }
+
+  const CeedScalar tol = 1e-14;
```
zatkins-dev, line 287: Why this tolerance? Generally, tolerances should be well justified and/or based on `CEED_EPSILON`

backends/avx/ceed-avx-tensor.c
zatkins-dev: `CeedGetWorkVector` is a good option for memory that would likely be reused in subsequent calls, as it does not require allocating every application. It could also go in the EvenOdd data structure

backends/avx/ceed-avx-tensor.c
jeremylt: The more I think on it, the more I think the TensorContraction object should be the one that supports (or doesn't) `TensorContractionApplyEvenOdd` and the basis checks for this support and holds the needed decompositions of the basis matrices if needed. Honestly, the default even/odd logic could even be at the interface level. Then we keep the changes to the single CPU basis application implementation and the even/odd logic at the interface level and every single CPU backend gets the changes all at once.

backends/avx/ceed-avx-tensor.c
```diff
@@ -269,34 +273,284 @@ static int CeedTensorContract_Avx_Single_4_8(CeedTensorContract contract, CeedIn
...
+  for (CeedInt j = 0; j < J_half; j++) {
+    for (CeedInt b = 0; b < B / 2; b++) {
+      CeedScalar val         = t[j * t_stride_0 + b * t_stride_1];
+      CeedScalar mirror      = t[j * t_stride_0 + (B - 1 - b) * t_stride_1];
+      t_even[j * B_half + b] = (val + mirror) * (CeedScalar)0.5;
+      t_odd[j * B_half + b]  = (val - mirror) * (CeedScalar)0.5;
+    }
+    if (B % 2) {
+      t_even[j * B_half + B / 2] = t[j * t_stride_0 + (B / 2) * t_stride_1];
+      t_odd[j * B_half + B / 2]  = (CeedScalar)0.0;
```
zatkins-dev, line 338: The ordering of the `t_even` and `t_odd` matrices need not depend on the transpose mode, right? The `t` matrix is symmetric or anti-symmetric regardless of whether we are looking at the base matrix or its transpose. As such, the split can be performed on just the base matrix. Then, the `EvenOddApply` function should index into the `t_even` or `t_odd` matrix such that it extracts the proper entries for the particular transpose mode. This would save one setup call for each matrix that is used in both `CEED_TRANSPOSE` and `CEED_NOTRANSPOSE` mode.

Am I missing something?

backends/avx/ceed-avx-tensor.c
zatkins-dev: Just to give a roadmap for what that would look like IMO (@jeremylt please let me know if this looks sensible to you):

- Add an enum `CeedSymmetryType`:
```c
typedef enum {
  /// Unknown/uninitialized symmetry
  CEED_SYMMETRY_UNKNOWN = 0,
  /// No symmetry
  CEED_SYMMETRY_NONE = 1,
  /// Symmetric over both indices
  CEED_SYMMETRY_SYMMETRIC = 2,
  /// Anti-Symmetric over both indices
  CEED_SYMMETRY_ANTISYMMETRIC = 3,
}
```
- Add `CeedBasisDetectSymmetry()`, `CeedBasisComputeHalfMatrices()`, and `CeedBasisGetEvenOddDecomposition[Interp|Grad|CollocatedGrad]1D()` to `interface/ceed-basis.c` (and `include/ceed/backend.c`)
```c
CeedInt CeedBasisDetectSymmetry(CeedBasis basis, const CeedInt M, const CeedInt N, const CeedScalar *matrix, const CeedSymmetryType* symmetry_type);
CeedInt CeedBasisComputeHalfMatrices(CeedBasis basis, const CeedInt M, const CeedInt N, const CeedScalar *matrix, const CeedSymmetryType symmetry_type, CeedScalar *matrix_even, CeedScalar *matrix_odd);
// Use the above two functions within CeedBasisGetEvenOddDecomposition[Interp|Grad|CollocatedGrad]1D()
// Sample implementation
CeedInt CeedBasisGetEvenOddDecompositionInterp1D(CeedBasis basis, CeedSymmetryType *symmetry_type, const CeedScalar **interp_1d_even, const CeedScalar **interp_1d_odd) {
  if (basis->interp_1d_symmetry_type == CEED_SYMMETRY_UNKNOWN) {
    CeedScalar *interp_1d;
    CeedInt P_1d, Q_1d, P_half, Q_half;
    
    CeedCall(CeedBasisGetNumNodes1D(basis, &P_1d));
    CeedCall(CeedBasisGetNumQuadraturePoints1D(basis, &Q_1d));
    P_half = (P_1d + 1) / 2;
    Q_half = (Q_1d + 1) / 2;
    CeedCall(CeedBasisGetInterp1D(basis, &interp_1d));
    
    // Note: have to add interp_1d_symmetry_type to the CeedBasis_private struct
    CeedCall(CeedBasisDetectSymmetry(basis, P_1d, Q_1d, interp_1d, &basis->interp_1d_symmetry_type));
    if (basis->interp_1d_symmetry_type != CEED_SYMMETRY_NONE) {
      // Note: have to add these to the CeedBasis_private struct
      CeedCallBackend(CeedCalloc(P_half * Q_half, &basis->interp_1d_even)); 
      CeedCallBackend(CeedCalloc(P_half * Q_half, &basis->interp_1d_odd));
      CeedCall(CeedBasisComputeHalfMatrices(basis, P_1d, Q_1d, basis->interp_1d, basis->interp_1d_symmetry_type, basis->interp_1d_even, basis->interp_1d_odd));
    }
  }
  if (symmetry_type) *symmetry_type = basis->interp_1d_symmetry_type;
  if (interp_1d_even) *interp_1d_even = basis->interp_1d_even;
  if (interp_1d_odd) *interp_1d_odd = basis->interp_1d_odd;
  return CEED_ERROR_SUCCESS;
}
CeedInt CeedBasisGetEvenOddDecompositionGrad1D(CeedBasis basis, CeedSymmetryType *symmetry_type, const CeedScalar **grad_1d_even, const CeedScalar **grad_1d_odd);
CeedInt CeedBasisGetEvenOddDecompositionCollocatedGrad1D(CeedBasis basis, CeedSymmetryType *symmetry_type, const CeedScalar **collo_grad_1d_even, const CeedScalar **collo_grad_1d_odd);
```

- Add a public interface to `interface/ceed-tensor.c` for applying Even-Odd decompositions, the implementation may look something like this:
```c
// Scratch must be preallocated of size 2 * (A * (B + 1) / 2 * C) + 2 * (A * (J + 1) / 2 * C)
int CeedTensorContractApplyEvenOdd(CeedTensorContract contract, CeedInt A, CeedInt B, CeedInt C, CeedInt J, const CeedScalar *restrict t_even,
                                   const CeedScalar *restrict t_odd, const CeedSymmetryType symmetry_type, CeedTransposeMode t_mode, const bool add,
                                   CeedScalar *restrict scratch, const CeedScalar *restrict u, CeedScalar *restrict v) {
  const CeedInt B_half = (B + 1) / 2;
  const CeedInt J_half = (J + 1) / 2;

  CeedScalar *u_even = scratch, *u_odd = &u_even[A * B_half * C];
  CeedScalar *w_even = &u_even[A * B_half * C], *w_odd = &w_even[A * J_half * C];

  // Fold input: combine lower and upper halves into even/odd components
  for (CeedInt a = 0; a < A; a++) {
    for (CeedInt b = 0; b < B / 2; b++) {
      const CeedInt idx_lower  = (a * B + b) * C;
      const CeedInt idx_upper  = (a * B + (B - 1 - b)) * C;
      const CeedInt idx_folded = (a * B_half + b) * C;

      for (CeedInt c = 0; c < C; c++) {
        u_even[idx_folded + c] = u[idx_lower + c] + u[idx_upper + c];
        u_odd[idx_folded + c]  = u[idx_lower + c] - u[idx_upper + c];
      }
    }
    if (B % 2) {
      const CeedInt idx_mid_in  = (a * B + B / 2) * C;
      const CeedInt idx_mid_out = (a * B_half + B / 2) * C;

      for (CeedInt c = 0; c < C; c++) {
        u_even[idx_mid_out + c] = u[idx_mid_in + c];
        u_odd[idx_mid_out + c]  = (CeedScalar)0.0;
      }
    }
  }
  
  // Half-contractions: w_even = t_even * u_even, w_odd = t_odd * u_odd
  // Add == false so that the w arrays are cleared before writing
  CeedCall(CeedTensorContractApply(contract, A, B_half, C, J_half, t_even, t_mode, false, u_even, w_even));
  CeedCall(CeedTensorContractApply(contract, A, B_half, C, J_half, t_odd, t_mode, false, u_odd, w_odd));
  
 if (!add) {
    for (CeedInt q = 0; q < A * J * C; q++) v[q] = (CeedScalar)0.0;
  }

  // Unfold output: recombine even/odd results into lower and upper halves
  for (CeedInt a = 0; a < A; a++) {
    for (CeedInt j = 0; j < J / 2; j++) {
      const CeedInt idx_half      = (a * J_half + j) * C;
      const CeedInt idx_out_lower = (a * J + j) * C;
      const CeedInt idx_out_upper = (a * J + (J - 1 - j)) * C;

      if (sym == CEED_SYMMETRY_SYMMETRIC) {
        for (CeedInt c = 0; c < C; c++) {
          v[idx_out_lower + c] += w_even[idx_half + c] + w_odd[idx_half + c];
          v[idx_out_upper + c] += w_even[idx_half + c] - w_odd[idx_half + c];
        }
      } else if (sym == CEED_SYMMETRY_ANTISYMMETRIC) {
        for (CeedInt c = 0; c < C; c++) {
          v[idx_out_lower + c] += w_even[idx_half + c] + w_odd[idx_half + c];
          v[idx_out_upper + c] += -w_even[idx_half + c] + w_odd[idx_half + c];
        }
      }
    }
    if (J % 2) {
      const CeedInt idx_half    = (a * J_half + J / 2) * C;
      const CeedInt idx_out_mid = (a * J + J / 2) * C;

      for (CeedInt c = 0; c < C; c++) v[idx_out_mid + c] += w_even[idx_half + c] + w_odd[idx_half + c];
    }
  }
  return CEED_ERROR_SUCCESS;
}
```

- Update `backends/ref/ceed-basis-ref.c` to call the appropriate `CeedBasisGetEvenOddDecomposition*` function to get the split before calling `CeedTensorContractApplyEvenOdd`.

zatkins-dev: See my other comment here https://github.com/CEED/libCEED/pull/2010#discussion_r3854842838 for discussion on how we should use the transpose mode.

jeremylt: does the ApplyEvenOdd even need to be backend specific?

zatkins-dev: good point -- it definitely doesn't. We should be able to just call `TensorContractApply` on the half matrices. Let me update my prior comment.

zatkins-dev: updated. I also baked in a preallocated scratch array -- I think the basis is better equipped to prealloc and cache a sufficiently large array than the tensor contraction is.

jeremylt: probably will be worth specializing on C being a multiple of 8 in that code at the interface level similar to how opt and avx do

zatkins-dev: Are you talking about the loops for building u_even/u_odd and for recombining to v?

jeremylt: yeah, if I'm recalling correctly, the compiler can emit vectorized instructions for those if it knows enough

zatkins-dev:
@mohitt31 Do you plan to continue with this PR? If not, I will go ahead and implement it with the changes requested and credit you in the commits.

mohitt31:
Hey guys, sorry for the delay. I had my mid-sem exams. I'm still working on this.
I read all the comments and Zach's plan. Moving the even-odd logic to the interface is a good idea. I will push a new update in a few days. I will do exactly what you guys suggested:

1.Add CeedSymmetryType enum
2.Do detection and half-matrix storage in the Basis
3.Put ApplyEvenOdd at the interface level
4.Change the avx backend back to normal
5.I will also change the tolerance to CEED_EPSILON and fix the other small things.

And Jeremy , I got your point about LLM replies. Sorry for that. I'l write everything myself from now on.

zatkins-dev:
No worries, thanks for the update!

backends/ref/ceed-ref-basis.c
```diff
@@ -58,6 +58,13 @@ static int CeedBasisApplyCore_Ref(CeedBasis basis, bool apply_add, CeedInt num_e
 
     CeedCallBackend(CeedBasisGetNumNodes1D(basis, &P_1d));
     CeedCallBackend(CeedBasisGetNumQuadraturePoints1D(basis, &Q_1d));
+
+    CeedSymmetryType   interp_symmetry, grad_symmetry;
+    const CeedScalar  *interp_1d_even = NULL, *interp_1d_odd = NULL;
+    const CeedScalar  *grad_1d_even = NULL, *grad_1d_odd = NULL;
+
+    CeedCallBackend(CeedBasisGetEvenOddDecompositionInterp1D(basis, &interp_symmetry, &interp_1d_even, &interp_1d_odd));
+    CeedCallBackend(CeedBasisGetEvenOddDecompositionGrad1D(basis, &grad_symmetry, &grad_1d_even, &grad_1d_odd));
```
jeremylt: This should be called in `CeedBasisCreateTensorH1` so this decomposition is only computed once.

We'll need to modify the `CeedBasis_Ref` object to
```C
typedef struct {
  bool             is_collocated;
  CeedScalar      *collo_grad_1d;
  CeedSymmetryType interp_symmetry;
  CeedScalar      *interp_1d_even, *interp_1d_odd;
  CeedSymmetryType grad_symmetry;
  CeedScalar      *grad_1d_even, *grad_1d_odd;
  CeedSymmetryType collo_grad_symmetry;
  CeedScalar      *collo_grad_1d_even, *collo_grad_1d_odd;
} CeedBasis_Ref;
```

And then `CeedbasisDestoryTensor_Ref` would need the update
```C
static int CeedBasisDestroyTensor_Ref(CeedBasis basis) {
  CeedBasis_Ref *impl;

  CeedCallBackend(CeedBasisGetData(basis, &impl));
  CeedCallBackend(CeedFree(&impl->collo_grad_1d));
  CeedCallBackend(CeedFree(&impl->interp_1d_even));
  CeedCallBackend(CeedFree(&impl->interp_1d_odd));
  CeedCallBackend(CeedFree(&impl->grad_1d_even));
  CeedCallBackend(CeedFree(&impl->grad_1d_odd));
  CeedCallBackend(CeedFree(&impl->collo_grad_1d_even));
  CeedCallBackend(CeedFree(&impl->collo_grad_1d_odd));
  CeedCallBackend(CeedFree(&impl));
  return CEED_ERROR_SUCCESS;
}
```

interface/ceed-basis.c
```diff
@@ -770,6 +770,175 @@ int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *collo_grad_1d) {
...
+  Checks whether matrix[row][col] = +/- matrix[num_rows-1-row][num_cols-1-col] using a relative tolerance based on `CEED_EPSILON`.
+
+  @param[in]  num_rows      Number of rows
+  @param[in]  num_cols      Number of columns
+  @param[in]  matrix        Row-major matrix data
+  @param[out] symmetry_type Detected symmetry type
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
```
jeremylt, line 785: really, `@ref Developer`

interface/ceed-basis.c
```diff
@@ -770,6 +770,175 @@ int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *collo_grad_1d) {
...
+  @param[in]  num_cols      Number of columns in original matrix
+  @param[in]  matrix        Row-major original matrix
+  @param[in]  num_rows_half (num_rows + 1) / 2
+  @param[in]  num_cols_half (num_cols + 1) / 2
+  @param[out] even          Even half-matrix, shape [num_rows_half, num_cols_half]
+  @param[out] odd           Odd half-matrix, shape [num_rows_half, num_cols_half]
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
```
jeremylt, line 844: Same here - this isn't ever used in the backends

backends/ref/ceed-ref-basis.c
jeremylt: ohhh, I see what you did

You stuck this on the `basis` object. That can work, but then we'd still want the `collo_grad_1d_even` and `collo_grad_1d_odd` here in the backend code

include/ceed/types.h
```diff
@@ -225,6 +225,19 @@ typedef enum {
...
+typedef enum {
+  /// Symmetry has not been checked yet
+  CEED_SYMMETRY_UNKNOWN = 0,
+  /// No centro-symmetry detected
+  CEED_SYMMETRY_NONE = 1,
+  /// Centro-symmetric: t[j][b] = t[J-1-j][B-1-b]
+  CEED_SYMMETRY_SYMMETRIC = 2,
+  /// Centro-antisymmetric: t[j][b] = -t[J-1-j][B-1-b]
+  CEED_SYMMETRY_ANTISYMMETRIC = 3,
+} CeedSymmetryType;
```
jeremylt, line 239: We'll want to add a `CeedSymmetryTypes` array for pretty-printing

jeremylt:
Note - CI requests style fixes. Also, this change needs to be summarized in CHANGELOG.md

zatkins-dev:
Should we add a way to disable this functionality? I feel like `CeedBasisUseEvenOdd_Ref` should take the `CeedBasis` as an argument and check if it has been disabled. We should probably have a function interface (say, `CeedBasisSetUseEvenOdd`) and an environment variable (maybe `CEED_BASIS_USE_EVEN_ODD`) so that it can be adjusted on a per-basis or global basis.

That would make benchmarking considerably easier, which we should also do prior to merging.

jeremylt:
> Should we add a way to disable this functionality? I feel like `CeedBasisUseEvenOdd_Ref` should take the `CeedBasis` as an argument and check if it has been disabled. We should probably have a function interface (say, `CeedBasisSetUseEvenOdd`) and an environment variable (maybe `CEED_BASIS_USE_EVEN_ODD`) so that it can be adjusted on a per-basis or global basis.

I think this is a good idea. See this commit for something similar: https://github.com/CEED/libCEED/pull/1987/changes/c7c5b1ac1bdf6521d0b2322cd5d5e48e3eda838d

momoquq:
Thanks — adding some native test results for `7b18a0a7ea91`.

I tested this revision on a Kunpeng AArch64 server with a 512-bit SVE vector length: C-only, FP64, GCC 12.3.1, with key flags `-O3 -march=armv8-a+sve -ffp-contract=fast -fopenmp-simd`.

`make prove search=t3` passed for all 44 selected t3xx test binaries on each of these resources — 176 test/backend combinations, with no skipped cases:

- `/cpu/self/sve/serial`
- `/cpu/self/sve/blocked`
- `/cpu/self/opt/serial`
- `/cpu/self/opt/blocked`

In addition to these upstream tests, I ran local correctness checks on the same node for each resource: 120 TensorContract cases against a scalar reference and four BasisApply cases (1D interp/grad, forward/transpose), with a normalized-error tolerance of 1e-12. Both the direct TensorContract checks and the exercised BasisApply paths reached the expected SVE or opt contraction implementations.

This covers the selected FP64 correctness tests. I haven’t run the full suite, FP32, or BP/performance tests yet; I’ll follow up with BP results.

I’ve also been working on SVE kernels, but other commitments have slowed my progress. This PR is further along than my implementation. Thanks for the great work! I’d like to build on it and help with follow-up optimizations.

By the way, I have access to SME-capable hardware and would like to contribute an SME backend. For an initial implementation, would I reusing the existing Basis implementations and specializing `CeedTensorContract`, as in this PR?

jeremylt:
Hey, @mohitt31 how's it going? One of us can tag in to handle the small number of clean-up items if you're busy (understandably) with the semester

mohitt31:
Here is summary of the things covered in this PR:

Cleanups & Core changes:

* Did some doc cleanup. Changed `@ref Backend` to `@ref Developer` for `CeedBasisDetectSymmetry` and `CeedBasisComputeHalfMatrices` (the two existing static helpers).
* Added `CeedSymmetryTypes` array.
* Added `CeedBasisSetUseEvenOdd` and `CEED_BASIS_USE_EVEN_ODD` env var so we can toggle it.
* Added `CeedComputeEvenOddDecomposition` in `backend.h`. Extracted this from both basis getters so the ref backend can decompose its `collo_grad` matrix without duplicating code.
* Implemented `collo_grad_1d_even` and `collo_grad_1d_odd` inside `CeedBasis_Ref`.

Why default is off:

* I kept the even-odd feature default off. The fact is even-odd changes the summation order, which causes a ~1 ulp shift.
* Because of this shift, `CompositeOperator::apply_add` doctest was failing when even-odd is on. Its basis is p=3, q=4 so the `collo_grad` matrix is 4x4, which crosses the threshold. The existing tolerance in that test is `10 * EPSILON` (for context, similar tests in that same file use `200 * EPSILON`). Since I didn't want to touch the Rust doctest tolerance, keeping default off solves this.
* By keeping default off, all existing tests run exactly the same code path and remain bit-identical.

Note on helper signature:

* For the helper signature, Zach suggested to take `CeedBasis`. But I used `CeedBasis_Ref *impl` instead. The reason is `impl` is already fetched before and the helper is called inside the `for d` loop, so it didn't make sense to fetch it again.

Testing:

* Added a new test `t366-basis.c` which checks two cases: 1) a matrix that is only centro-symmetric in the top-left quadrant, and 2) antisymmetric matrix in transpose mode. Both of these fails without the fixes.
* C suite is passing 266/266 on macOS + x86_64 (AVX included). Checked both default and forced-on modes.
* Checked even-odd on vs off: all 324 configs agree under `1e-12` relative tolerance (not bit-identical because of the ulp shift).
* Rust doctests passing 111/0.

Benchmarks:

* Because of the new toggle, A/B testing is possible now without rebuilding. Benchmark numbers are still pending, I will post them here later
*

backends/ref/ceed-ref-basis.c
```diff
@@ -305,7 +384,17 @@ int CeedBasisCreateTensorH1_Ref(CeedInt dim, CeedInt P_1d, CeedInt Q_1d, const C
 
   CeedCallBackend(CeedSetBackendFunction(ceed, "Basis", basis, "Apply", CeedBasisApply_Ref));
   CeedCallBackend(CeedSetBackendFunction(ceed, "Basis", basis, "ApplyAdd", CeedBasisApplyAdd_Ref));
+  CeedCallBackend(CeedSetBackendFunction(ceed, "Basis", basis, "SetUseEvenOdd", CeedBasisSetUseEvenOdd_Ref));
   CeedCallBackend(CeedSetBackendFunction(ceed, "Basis", basis, "Destroy", CeedBasisDestroyTensor_Ref));
+
+  {
+    const char *env_val      = getenv("CEED_BASIS_USE_EVEN_ODD");
```
jeremylt, line 392: @zatkins-dev I'm fine with this being here for now, but really we should clean-up/standardize this. As I see it, I think the interface layer is supposed to communicate with the user and the backend is supposed to implement to spec. To that end, this function and the similar CUDA graph one should be moved up to the interface level. Might make a good student issue.

(put this comment in the wrong place)

backends/ref/ceed-ref-basis.c
```diff
@@ -296,6 +372,9 @@ int CeedBasisCreateTensorH1_Ref(CeedInt dim, CeedInt P_1d, CeedInt Q_1d, const C
   if (Q_1d >= P_1d && !impl->is_collocated) {
     CeedCallBackend(CeedMalloc(Q_1d * Q_1d, &impl->collo_grad_1d));
     CeedCallBackend(CeedBasisGetCollocatedGrad(basis, impl->collo_grad_1d));
+    // The collocated gradient is backend data, so its decomposition is computed and cached here rather than on the CeedBasis
```
jeremylt:
```suggestion
```
I'm fine leaving this comment out now that I understand

backends/ref/ceed-ref-basis.c
zatkins-dev: I agree -- I think that we should have a centralized place for runtime environment options so that they are clear and consistent.

interface/ceed-basis.c
```diff
@@ -770,6 +770,195 @@ int CeedBasisGetCollocatedGrad(CeedBasis basis, CeedScalar *collo_grad_1d) {
...
+  @param[in]  num_rows      Number of rows
+  @param[in]  num_cols      Number of columns
+  @param[in]  matrix        Row-major matrix data
+  @param[out] symmetry_type Detected symmetry type
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Developer
+**/
+static int CeedBasisDetectSymmetry(CeedInt num_rows, CeedInt num_cols, const CeedScalar *matrix, CeedSymmetryType *symmetry_type) {
```
jeremylt: Minor, I'd move these two right after the static inline def of `CeedBasisCreateProjectionMatrices` since that's the section of this file where these sorts of functions go

backends/ref/ceed-ref-basis.c
```diff
@@ -58,6 +69,13 @@ static int CeedBasisApplyCore_Ref(CeedBasis basis, bool apply_add, CeedInt num_e
 
     CeedCallBackend(CeedBasisGetNumNodes1D(basis, &P_1d));
     CeedCallBackend(CeedBasisGetNumQuadraturePoints1D(basis, &Q_1d));
+
+    CeedSymmetryType  interp_symmetry, grad_symmetry;
+    const CeedScalar *interp_1d_even = NULL, *interp_1d_odd = NULL;
+    const CeedScalar *grad_1d_even = NULL, *grad_1d_odd = NULL;
+
+    CeedCallBackend(CeedBasisGetEvenOddDecompositionInterp1D(basis, &interp_symmetry, &interp_1d_even, &interp_1d_odd));
+    CeedCallBackend(CeedBasisGetEvenOddDecompositionGrad1D(basis, &grad_symmetry, &grad_1d_even, &grad_1d_odd));
```
zatkins-dev: I don't think that we want to compute the decomposition matrices for grad unless we are using the grad eval mode. Some bases will never need to use these.

jeremylt: Yeah, I'd push these down to where they are used

backends/ref/ceed-ref-basis.c
```diff
@@ -105,13 +128,26 @@ static int CeedBasisApplyCore_Ref(CeedBasis basis, bool apply_add, CeedInt num_e
...
+            const CeedSymmetryType t_symmetry     = is_notranspose ? interp_symmetry : impl->collo_grad_symmetry;
+            const CeedScalar      *t_even         = is_notranspose ? interp_1d_even : impl->collo_grad_1d_even;
+            const CeedScalar      *t_odd          = is_notranspose ? interp_1d_odd : impl->collo_grad_1d_odd;
+
+            for (CeedInt d = 0; d < dim; d++) {
+              const CeedInt     is_add = (t_mode == CEED_TRANSPOSE) && (d > 0);
+              const CeedScalar *in     = is_notranspose ? (d == 0 ? u : tmp[d % 2]) : &u[d * num_qpts * num_comp * num_elem];
+              CeedScalar       *out    = is_notranspose ? (d == dim - 1 ? interp : tmp[(d + 1) % 2]) : interp;
+
+              if (CeedBasisUseEvenOdd_Ref(impl, t_symmetry, P, Q)) {
```
zatkins-dev, line 145: We could definitely compute this bool a single time for each basis mat, but I think this is probably fine for an internal helper.

jeremylt:
I'm actually fine adjusting the Rust test tols and turning this on by default if there is an order at which this change wins for all backends.

backends/ref/ceed-ref-basis.c
jeremylt: Honestly, for cpu/gen I think I would like a bool precomputed.

backends/ref/ceed-ref-basis.c
zatkins-dev: I'm inclined to say that it's too small/requires too much guidance for how we would want it done for a student contribution, though.

backends/ref/ceed-ref-basis.c
jeremylt: (not urgent, can be follow-up)

backends/ref/ceed-ref-basis.c
jeremylt: maybe if bundled with the other changes, but yea, maybe not the best candidate

backends/ref/ceed-ref-basis.c
```diff
@@ -296,6 +372,9 @@ int CeedBasisCreateTensorH1_Ref(CeedInt dim, CeedInt P_1d, CeedInt Q_1d, const C
   if (Q_1d >= P_1d && !impl->is_collocated) {
     CeedCallBackend(CeedMalloc(Q_1d * Q_1d, &impl->collo_grad_1d));
     CeedCallBackend(CeedBasisGetCollocatedGrad(basis, impl->collo_grad_1d));
+    // The collocated gradient is backend data, so its decomposition is computed and cached here rather than on the CeedBasis
+    CeedCallBackend(CeedComputeEvenOddDecomposition(Q_1d, Q_1d, impl->collo_grad_1d, &impl->collo_grad_symmetry, &impl->collo_grad_1d_even,
+                                                    &impl->collo_grad_1d_odd));
```
zatkins-dev, line 378: I'm hesitant to support precomputing this for all bases; the collocated gradient should probably just be promoted to an interface/`CeedBasisGetCollocatedGrad` should cache its output. I can do that real quick.

mohitt31:
Thanks for the feedback! I've pushed the requested updates:

Changes in this commit:

* Restricted the gradient decomposition strictly to `grad` evaluation mode (verified that `grad_1d_even` correctly remains `NULL` on an `interp`-only basis).
* Removed the outdated comment.
* Relocated both static helpers to immediately follow `CeedBasisCreateProjectionMatrices`.

Notes on remaining items:

* `collo_grad` caching: The even-odd decomposition for `collo_grad` is implemented in this PR. However, per Zach's note, promoting `CeedBasisGetCollocatedGrad` to the interface level to cache its output is left for him to handle separately.
* `precomputed` boolean: Deferred to a follow-up PR, as Jeremy noted it's not urgent.
* Environment variable interface: Retained in its current location for now, as Jeremy confirmed this is fine for this iteration.

Testing:

* C Suite: 266/266 passed on macOS and x86_64 (AVX included). Checked under both `default` and `forced-on` configurations.
* Rust Suite: 111/0 passed.

jeremylt:
There's merge conflicts and changes to the test environment since this branch started. Can you rebase on current main please?

CHANGELOG.md
```diff
@@ -42,6 +42,10 @@ Specifically, directories set with `CeedAddJitSourceRoot(ceed, "foo/bar")` will
 - Add string names of `ElemRestrictionType`s for human-readable output.
 - Add vector-length-agnostic Arm SVE serial and blocked CPU backends.
 - Add `/cpu/self/gen/serial` and `/cpu/self/gen/blocked` CPU just-in-time (JiT) compiled backends, inspired by the CUDA and ROCm JiT complied backends. JiT compiler and optimizer options can be set at compile or runtime with the environment variables `CEED_CPU_JIT_CXX` and `CEED_CPU_JIT_OPT`, respectively.
+- Add even-odd (centro-symmetry) decomposition for tensor contractions, which halves the arithmetic when the 1D basis matrix is centro-symmetric.
```
jeremylt, line 45: This release note is very long. It should be boiled down to two sentences (human written please) and summarize any important user facing options ta enable/disable this.

backends/ref/ceed-ref-basis.c
```diff
@@ -9,10 +9,21 @@
...
 #include "ceed-ref.h"
 
+//------------------------------------------------------------------------------
+// Even-odd decomposition is only worth using on a centro-symmetric matrix that is
+// large enough for the halved contraction to outweigh the fold and unfold, and
+// only when it has not been turned off for this basis.
+//------------------------------------------------------------------------------
+static inline bool CeedBasisUseEvenOdd_Ref(const CeedBasis_Ref *impl, CeedSymmetryType symmetry_type, CeedInt B, CeedInt J) {
+  return impl->use_even_odd && (symmetry_type == CEED_SYMMETRY_SYMMETRIC || symmetry_type == CEED_SYMMETRY_ANTISYMMETRIC) &&
+         B >= CEED_EVEN_ODD_MIN_DIM && J >= CEED_EVEN_ODD_MIN_DIM;
```
jeremylt: can we compute the total flops in each case by hand to determine what the best size is and cross check with performance testing?

backends/ref/ceed-ref-basis.c
```diff
@@ -9,10 +9,21 @@
...
 
 #include "ceed-ref.h"
 
+//------------------------------------------------------------------------------
+// Even-odd decomposition is only worth using on a centro-symmetric matrix that is
+// large enough for the halved contraction to outweigh the fold and unfold, and
+// only when it has not been turned off for this basis.
+//------------------------------------------------------------------------------
+static inline bool CeedBasisUseEvenOdd_Ref(const CeedBasis_Ref *impl, CeedSymmetryType symmetry_type, CeedInt B, CeedInt J) {
+  return impl->use_even_odd && (symmetry_type == CEED_SYMMETRY_SYMMETRIC || symmetry_type == CEED_SYMMETRY_ANTISYMMETRIC) &&
```
jeremylt: really, if the user demands that even_odd is used, then we should use it (and not if they demand not) otherwise we use the default dim/order thresholds

jeremylt:
As a reminder by the way, the discussion in this PR should be human-to-human and not written with an LLM

backends/ref/ceed-ref.h
```diff
@@ -31,9 +31,20 @@ typedef struct {
                CeedRequest *);
 } CeedElemRestriction_Ref;
 
+/// Whether the user has asked for the even-odd decomposition on this basis.
+/// Unset leaves the choice to the size threshold; the other two override it.
+typedef enum {
+  CEED_BASIS_EVEN_ODD_UNSET = 0,
```
jeremylt: This is overkill.

The logic only needs to be

if (bool from user is set) return bool from user
else return result of default check

@zatkins-dev does your helper have a flag to check if the user set a value? would be useful

backends/ref/ceed-ref-basis.c
```diff
@@ -9,10 +9,24 @@
...
+#include <stdlib.h>
 #include <string.h>
 
 #include "ceed-ref.h"
 
+//------------------------------------------------------------------------------
+// The decomposition needs a centro-symmetric matrix, so that is checked first.
+// Past that, an explicit request from the user decides, and only when there is
+// none does the size threshold choose: below it the fold and unfold cost more
+// than the halved contraction saves.
```
jeremylt:
```suggestion
// Check user setting or default to order rule of thumb
```
way too verbose, check the style of other comments in the codebase

interface/ceed-basis.c
```diff
@@ -2517,6 +2706,33 @@ int CeedBasisGetCurl(CeedBasis basis, const CeedScalar **curl) {
   return CEED_ERROR_SUCCESS;
 }
 
+/**
+  @brief Enable or disable the even-odd decomposition for a `CeedBasis`
```
jeremylt, line 2749: way too long of a description; should be a single human readable sentence

tests/t366-basis.c
```diff
@@ -0,0 +1,145 @@
...
+
+int main(int argc, char **argv) {
+  Ceed             ceed;
+  const CeedInt    num_comp = 2, num_elem = 3;
+  const CeedScalar tol = CEED_SCALAR_TYPE == CEED_SCALAR_FP32 ? 1.e-4 : 1.e-11;
+
+  CeedInit(argv[1], &ceed);
+
+  // Orders straddle CEED_EVEN_ODD_MIN_DIM so both the plain and the even-odd contraction are covered.
+  // The high orders are kept to one and two dimensions to hold the test's run time down.
```
jeremylt: Is this comment needed or is it unchecked LLM over-commenting? Really the whole file need small, focused comments instead of long ones

backends/ref/ceed-ref.h
zatkins-dev: the way the env processing works is you can set the bool to the default and it will only be changed if the user sets it. I think an explicit getter for if the user set it would be good though

jeremylt: yea, we need basis default calculations if the user doesn't opt in/out so I think that a helper to determine if the user set the flag is needed for that

jeremylt: yea, we need basis default calculations if the user doesn't opt in/out so I think that a helper to determine if the user set the flag is needed for that

jeremylt:
This is a cool feature that could have big performance benefits. However, given the current level of human-to-human discussion in this PR, I'm not certain that libCEED's expectations are being met in terms of code ownership and assuming full responsibility for the PR content. Especially due to the level of impact this change could have, I think author responsibility is critical in this PR.

I totally understand @mohitt31 if it is the case that you don't feel like you have the time, interest, or experience to conduct human-to-human technical discussions and assume full responsibility for all lines of code in this PR. I do not want to assert a value judgement in any way here.

However, both in terms of the scope of impact and adherence to the contribution guidelines I feel like this is an essential requirement.

If for whatever reason this requirement cannot be met, I would suggest that we squash+merge this contribution into a staging branch to preserve attribution. Then a libCEED core member could assume responsibility and perform a line by line review followed by a commit to adapt the contribution to better align the code with libCEED standards wrt norms, clarity, maintainability, extensibility, etc. This member would assume responsibility for managing downstream consequences of that PR. Then an additional core libCEED member would perform a review upon this branch prior to merging.

This feels very nonstandard to me, but I really feel like in terms of both libCEED contributing guidelines and scope of impact we need a clear responsible party for this change.

Thoughts? Suggestions? I don't want to overstep here but I really think that we need a developer who fully assumes responsibility for this change and it's consequences and I worry that this isn't happening in this PR right now (for whatever reasons, that I, again, do not want to assign a value judgement to).

mohitt31:
ok iam ok with staging branch idea of this
yes i agree ai is being used a lot in this pr, sorry for that
thanks for your time and effort

PR #2016, opened 2026-08-27
HIP Vec Filter

### LINKED ISSUE #1830: CeedVectorFilter

jeremylt:
Purpose:

Really, this felt to simple to be a useful student issue anymore

Closes: #1830

LLM/GenAI Disclosure:

none

### DIFF
--- a/backends/hip-ref/ceed-hip-ref-vector.c
+++ b/backends/hip-ref/ceed-hip-ref-vector.c
@@ -281,7 +281,7 @@ static int CeedVectorSetArray_Hip(const CeedVector vec, const CeedMemType mem_ty
 // Copy host array to value strided
 //------------------------------------------------------------------------------
 static int CeedHostCopyStrided_Hip(CeedScalar *h_array, CeedSize start, CeedSize stop, CeedSize step, CeedScalar *h_copy_array) {
-  for (CeedSize i = start; i < stop; i += step) h_copy_array[i] = h_array[i];
+  CeedPragmaSIMD for (CeedSize i = start; i < stop; i += step) h_copy_array[i] = h_array[i];
   return CEED_ERROR_SUCCESS;
 }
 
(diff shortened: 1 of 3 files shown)


backends/hip-ref/ceed-hip-ref-vector.c
```diff
@@ -827,11 +827,41 @@ static int CeedVectorScale_Hip(CeedVector x, CeedScalar alpha) {
   return CEED_ERROR_SUCCESS;
 }
 
+//------------------------------------------------------------------------------
+// Filter or clip a vector using a threshold value on the host
+//------------------------------------------------------------------------------
+static int CeedHostFilter_Hip(CeedScalar *x_array, CeedScalar threshold, CeedSize length) {
+  CeedPragmaSIMD for (CeedSize i = 0; i < length; i++) {
```
zatkins-dev, line 834: Worth noting: the `CeedPragmaSIMD` isn't doing a lot here above `-O1`, still doesn't hurt though (https://godbolt.org/z/3Mfc1qfce)

backends/hip-ref/ceed-hip-ref-vector.c
```diff
@@ -281,7 +281,7 @@ static int CeedVectorSetArray_Hip(const CeedVector vec, const CeedMemType mem_ty
 // Copy host array to value strided
 //------------------------------------------------------------------------------
 static int CeedHostCopyStrided_Hip(CeedScalar *h_array, CeedSize start, CeedSize stop, CeedSize step, CeedScalar *h_copy_array) {
```
zatkins-dev, line 283: Are these host functions actually tested? I think the way we set up the tests, they may not be. We should probably double-check.

zatkins-dev (approved):
Looks good and passes cleanly locally, thanks!

backends/hip-ref/ceed-hip-ref-vector.c
jeremylt: yea, figured it was more clarity of intent than anything else

backends/hip-ref/ceed-hip-ref-vector.c
jeremylt: That would probably be a good student issue - modify the tests suite to ensure all these Host functions are tested

PR #2006, opened 2026-08-18
cuda-gen: Add CUDA Graph capture and replay for composite operators

jeremylt:
Squash of #1987 merged into staging branch for Noether testing. See original PR for discussion - this is only a testing branch

PR #2019, opened 2026-09-01
Fix CUDA Graph implementation

zatkins-dev:
Purpose:

#2006 broke Ratel tests due to (1) something being wrong with the async memory handling, and (2) the graph implementation allowing code to progress from an unrecoverable state (specifically, any errors in `CeedOperatorApplyAddCore_Cuda_gen` were ignored). 

This MR fixes these by:
1. Reverting the async memory operations. We need to do a thorough think about how best to allow for CUDA streams, as synchronization is very tricky. I have a very rough idea of how to approach it (that was the direction I went first), but actually making sure it's 100% correct seems impossible. C is not a language built for asynchronicity.
2. Reordering and adding back some error checking to the CUDA Graph implementation. The refresh of operator data/input arrays now occurs before the graph is built. These cannot get invalidated by building the graph, and this removes the need to have async memory operations as they happen pre-capture. As mentioned above, we also need to fail loudly on any error in `CeedOperatorApplyAddCore_Cuda_gen`. This is mainly because, depending on where the error occurred, the input and output vectors may have been checked out for reading/writing without being restored. This is what was ultimately breaking many Ratel tests after removing the async functions; the graph would fail to build due to `[CUDA_ERROR_STREAM_CAPTURE_IMPLICIT](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ggc6c391505e117393cc2558fff6bfc2e9960a55453736ec87ca941f9bc2d80abe)`, which led to crashing at cleanup time due to the vectors having outstanding readers.

Closes: Bug in main

LLM/GenAI Disclosure:

None

### DIFF
--- a/backends/cuda-gen/ceed-cuda-gen-operator.c
+++ b/backends/cuda-gen/ceed-cuda-gen-operator.c
@@ -424,7 +424,8 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
   CeedCallBackend(CeedOperatorCompositeGetSubList(op, &sub_operators));
   CeedCallBackend(CeedOperatorGetData(op, &impl));
   if (!impl->use_graph || (input_vec == CEED_VECTOR_NONE && output_vec == CEED_VECTOR_NONE)) {
-    return CeedOperatorApplyAddComposite_NoGraph_Cuda_gen(op, input_vec, output_vec, request);
+    CeedCallBackend(CeedOperatorApplyAddComposite_NoGraph_Cuda_gen(op, input_vec, output_vec, request));
+    return CEED_ERROR_SUCCESS;
   }
   if (!impl->warmup_done) {
     CeedCallBackend(CeedOperatorApplyAddComposite_NoGraph_Cuda_gen(op, input_vec, output_vec, request));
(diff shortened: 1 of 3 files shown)


backends/cuda-gen/ceed-cuda-gen-operator.c
```diff
@@ -424,7 +424,8 @@ static int CeedOperatorApplyAddComposite_Cuda_gen(CeedOperator op, CeedVector in
   CeedCallBackend(CeedOperatorCompositeGetSubList(op, &sub_operators));
   CeedCallBackend(CeedOperatorGetData(op, &impl));
   if (!impl->use_graph || (input_vec == CEED_VECTOR_NONE && output_vec == CEED_VECTOR_NONE)) {
-    return CeedOperatorApplyAddComposite_NoGraph_Cuda_gen(op, input_vec, output_vec, request);
+    CeedCallBackend(CeedOperatorApplyAddComposite_NoGraph_Cuda_gen(op, input_vec, output_vec, request));
+    return CEED_ERROR_SUCCESS;
```
jeremylt, line 428: we should do a pass to remove all of that sort of returns so the error tracing is clearer

jeremylt (approved):

backends/cuda-gen/ceed-cuda-gen-operator.c
zatkins-dev: true, it's a tempting but bad way to propagate errors

zatkins-dev:
you may have to merge, I don't think I can override the failed checks

PR #2021, opened 2026-09-10
Add Arm SVE CPU backends

### LINKED ISSUE #1984: SVE Backend
@MoMo569377793 we were working on an SVE backend but it appears you got much further. Would you be interested in working with us to upstream that effort in your fork to libCEED's main branch?

hughcars:
## Summary

Add Arm SVE resources for serial and blocked CPU execution:

```text
/cpu/self/sve/serial
/cpu/self/sve/blocked
```

The resources delegate all Basis and Operator behavior to `/cpu/self/opt/*` and override only `CeedTensorContract` with a vector-length-agnostic SVE implementation.

Closes #1984.

## What changed

- One FP32/FP64 SVE tensor contraction in the shape of the AVX backend: a per-slice kernel with four-, two-, and predicated one-vector column blocks and 4/2/1-row tiles, accumulators retained across the reduction dimension, and a `C == 1` kernel vectorized over the output rows. The runtime vector length is handled without fixed-width assumptions.
- Tiles advance a fixed set of row pointers rather than indexing from loop counters, so GCC 14 and Clang 18 no longer strength-reduce each address into its own spilled induction variable; the same source now performs within a few percent across the three compilers (figure below).
- SVE support is detected with a compile probe using the active compiler target and flags, with weak unavailable-backend registration otherwise.
- CI runs the full existing suite for every enabled backend on ARM natively and under QEMU at 128- and 2048-bit SVE (GCC/Clang, FP32/FP64); no SVE-specific tests are added.
- The IBM Power job now cross-compiles and emulates only the test binaries under `qemu-ppc64le`, running the full suite in ~1 min instead of a subset in ~7.
- `t360-basis` used an FP64 tolerance that failed on FP32 builds regardless of backend; it now uses `CeedScalar` and scales with `CEED_EPSILON`.

The implementation uses the SVE instruction subset. It does not require SVE2 and does not add SME support. There is no runtime ISA dispatch; a library built with SVE enabled must run on an SVE-capable target.

## Scope

No Basis, Operator, or restriction overrides, no shape dispatch tables, no fixed vector-length policies, no workload-specific dispatch. The generic even-odd work in #2010 is complementary and can use this backend through the existing `CeedTensorContractApply` interface.

## Performance

All numbers at **`-O3 -mcpu=native -ffp-contract=fast`**.

CEED BP3, `benchmarks/benchmark.sh -r petsc-bpsraw.sh -b bp3 -n 16 -p 16`, Graviton4 (Neoverse V2, 128-bit SVE, c8g.16xlarge), GCC 13, libCEED ae7e0bc30. Peak MDoF/s per degree, median of 3 sweeps; parentheses are the ratio to `/cpu/self/xsmm/blocked` (LIBXSMM 7c0898c).

| p | opt/blocked | sve/blocked | xsmm/blocked |
|---|---|---|---|
| 1 | 37.6 (0.88) | 47.1 (1.11) | 42.5 |
| 2 | 122.0 (0.75) | 173.1 (1.06) | 163.2 |
| 3 | 181.1 (0.66) | 272.0 (1.00) | 273.3 |
| 4 | 215.1 (0.62) | 337.4 (0.97) | 349.3 |
| 5 | 232.1 (0.58) | 378.5 (0.94) | 403.1 |
| 6 | 227.1 (0.56) | 384.8 (0.96) | 402.8 |
| 7 | 234.1 (0.52) | 426.5 (0.95) | 447.2 |
| 8 | 228.1 (0.51) | 423.4 (0.95) | 444.5 |

Compiler sensitivity of the SVE kernel, same benchmark on c8g.4xlarge, previous head b97b129 vs ae7e0bc30, GCC 13 / GCC 14 / Clang 18. Relative to the GCC 13 build, GCC 14 moves from 0.91–0.95 to 0.99–1.01 and Clang 18 from 0.94–0.98 to 0.97–0.99.

On Graviton3 (Neoverse V1, 256-bit SVE, c7g.16xlarge) the same binary recipe is at or above `xsmm/blocked` at every degree (1.00–1.18); on Graviton5 (Neoverse V3, 128-bit, c9g.16xlarge) the profile matches Graviton4. See the [issue comment](https://github.com/CEED/libCEED/pull/2021#issuecomment-5706112256) for the three-generation plot.

## Update — ae7e0bc30

Kernel restructured for compiler-robust codegen (fixed row pointers, AVX-style slice kernel, `C == 1` kernel). GCC 14 and Clang 18 builds are now within 3% of GCC 13 (were 10%); whole-operator set 0.81× the previous head. BP3 across Graviton3/4/5 (c7g/c8g/c9g 16xlarge, 16 ranks, GCC 13 `-O3 -mcpu=native`): at or above libxsmm at every degree on the 256-bit Graviton3; ahead at p ≤ 2 and 4–6% behind above that on the 128-bit Graviton4/5.

bp3_graviton_generations_ae7e0bc30

## Validation

Full existing suite on GCC 13, GCC 14, and Clang 18, FP64 and FP32, native on Graviton3 (256-bit), Graviton4 and Graviton5 (128-bit), and under QEMU at 128- and 2048-bit vector lengths; all enabled CPU backends. Results are bitwise identical to `/cpu/self/ref`. Whole-operator loops on 49 H1/H(curl)/H(div) workloads (GCC 13 and Clang 18, -O1/-O3, both precisions) run in 0.81× the time of the previous head in aggregate with no case more than 3% slower. Clang 22 formatting.

## LLM/GenAI disclosure

OpenAI GPT-family and Anthropic Claude coding agents contributed substantially to this PR: implementation and tuning of the SVE contraction, the CI changes, benchmarking and analysis, and drafting this description. Kernel tiling was selected by automated benchmark campaigns gated on correctness and performance. Independent read-only reviews were done with GPT-6, Claude, and GPT-5.6 models.


jeremylt:
I'm a bit worried about seeing that the SVE tensor contraction logic is leaking into the SVE Basis? It would be really good if we could use the base CPU ref Basis implementation and only have special case logic in the TensorContract object.

hughcars:
That makes sense. I removed the SVE Basis override in `b97b1292b`. The shared ref/opt Basis still reaches `CeedTensorContractApply_Sve` through `CeedTensorContractStridedApply`; this only removes the SVE-specific `q_comp * Q` flattening policy.

The TensorContract-only version passed the full suite, GCC/Clang, FP32/FP64, native Graviton4, QEMU from 128–2048 bits, and focused H1, H(div), and H(curl) tests. Correctness was unchanged.

| Metric | Previous | Shared Basis | Change |
|---|---:|---:|---:|
| Primary SVE / best | 1.136 | 1.165 | +2.5% |
| H1 / H(curl) | 1.162 / 1.111 | 1.186 / 1.145 | +2.0% / +3.1% |
| Held-out | 1.079 | 1.111 | +3.0% |
| Speedup over opt | 5.89× | 5.82× | −1.2% |
| SVE source lines | 624 | 464 | −25.6% |

The largest single regression was 12.8% in one FP32 non-tensor tet case, but the aggregate cost is about 2–3%. That seems worth the smaller backend and the cleaner boundary.

I also tested moving the flattening into `CeedTensorContractStridedApply` on a separate branch from `main`. The result was backend-dependent rather than a general win. Values below are geometric-mean runtime changes; negative is faster. The 16-case suite was used during SVE autoresearch, while the 22 held-out cases were excluded from tuning.

| Backend | 16-case suite (`/blocked`) | 22 held-out cases (`/blocked`) | 8 serial cases (`/serial`) |
|---|---:|---:|---:|
| ref | +5.0% | +5.6% | neutral |
| opt | −0.4% | neutral | +3.6% |
| XSMM | −1.7% | −0.9% | +3.5% |

So I do not think the unconditional generic version is worth proposing. It would need a backend-owned cost decision or another TensorContract hook. I could take a look at that as a separate follow-up if you'd be open to it.

tests/t366-tensor.c
```diff
@@ -0,0 +1,182 @@
...
+static void GuardedArrayDestroy(GuardedArray *array) {
+#if defined(__linux__)
+  if (array->mapping && array->mapping != MAP_FAILED) munmap(array->mapping, array->mapping_size);
+#else
+  free(array->data);
+#endif
+  array->data = NULL;
+}
+
+static CeedInt GetVectorLength(void) {
```
zatkins-dev:
```suggestion
static CeedInt GetVectorLength() {
```

Style: no explicit `void` argument for functions with no args. Without further context, I'm inclined to say this should be a preprocessor macro instead of a function.

tests/t366-tensor.c
```diff
@@ -0,0 +1,182 @@
...
+
+typedef struct {
+  CeedScalar *data;
+#if defined(__linux__)
+  void  *mapping;
+  size_t mapping_size;
+#endif
+} GuardedArray;
+
+static int GuardedArrayCreate(CeedSize length, GuardedArray *array) {
```
zatkins-dev: Definitely need docstrings and some comments for these helpers. I *think* this is allocating page-aligned memory with some flags set, but I don't know enough to follow the details.

.github/workflows/c-fortran-test-arm64.yml
```diff
@@ -26,3 +26,55 @@ jobs:
...
+          bits=$(qemu-aarch64 -cpu "$cpu" build/sve-vector-length)
+          echo "vq=$vq bits=$bits"
+          test "$bits" -eq "$((128 * vq))"
+          for resource in /cpu/self/sve/serial /cpu/self/sve/blocked; do
+            echo "vq=$vq resource=$resource test=t366"
+            qemu-aarch64 -cpu "$cpu" build/t366-tensor "$resource"
+            echo "vq=$vq resource=$resource test=t367"
+            qemu-aarch64 -cpu "$cpu" build/t367-basis "$resource"
+          done
+        done
```
zatkins-dev: What's this test doing exactly? We have the ability to filter tests based e.g. on the CEED backend, so we possibly don't need this

tests/t367-basis.c
```diff
@@ -0,0 +1,167 @@
...
+  CeedScalar                *div          = calloc((CeedSize)test.num_qpts * num_nodes, sizeof(*div));
+  CeedScalar                *q_ref        = calloc((CeedSize)dim * test.num_qpts, sizeof(*q_ref));
+  CeedScalar                *q_weight     = calloc(test.num_qpts, sizeof(*q_weight));
+  CeedScalar                *input_array  = malloc(input_size * sizeof(*input_array));
+  CeedScalar                *output_array = malloc(output_size * sizeof(*output_array));
+  CeedScalar                *expected     = malloc(output_size * sizeof(*expected));
+  const volatile CeedScalar *interp_ref = interp, *input_ref = input_array;
+  int                        failed = !interp || !div || !q_ref || !q_weight || !input_array || !output_array || !expected;
+
+  if (failed) goto cleanup_arrays;
```
zatkins-dev: style: avoid `goto` for error handling, even in tests

tests/t366-tensor.c
```diff
@@ -0,0 +1,182 @@
+/// @file
+/// Test tensor contraction semantics and vector-length tails
+/// \test Test tensor contraction semantics and vector-length tails
+//TESTARGS(only="cpu") {ceed_resource}
+#if defined(__linux__)
```
jeremylt, line 5: These two test files confuse me. Can you explain why the existing tests are insufficient and these two tests need to be executed by all backends?

.github/workflows/c-fortran-test-arm64.yml
```diff
@@ -26,3 +26,55 @@ jobs:
...
+        os: [ubuntu-24.04-arm]
+        compiler: [gcc, clang]
+        precision: [fp64, fp32]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - name: Environment setup
+      uses: actions/checkout@v5
+    - name: Install QEMU
```
jeremylt, line 42: I was originally thinking of testing SVE via emulation of supporting hardware, like the IBM Power job. If this runs faster, then we should switch to this for both SVE and IBM Power CI.

.github/workflows/c-fortran-test-arm64.yml
```diff
@@ -26,3 +26,55 @@ jobs:
...
+    - name: Configure precision
+      if: matrix.precision == 'fp32'
+      run: sed -i 's/#include "ceed-f64.h"/#include "ceed-f32.h"/' include/ceed/types.h
+    - name: Build SVE tests
+      env:
+        CC: ${{ matrix.compiler }}
+        FC: ""
+        SVE_OPT: -O2 -g -ffp-contract=fast -march=armv8-a+sve
+      run: |
+        make -j4 OPT="$SVE_OPT" build/t366-tensor build/t367-basis
```
jeremylt, line 53: all backends should be run against the full test suite in CI, not a subset of the tests

tests/t366-tensor.c
hughcars: Agreed, redundant. Removed in fdbd83d; the full suite under QEMU at VL 128 and 2048 covers it.

.github/workflows/c-fortran-test-arm64.yml
hughcars: Done in 9d6789b: Power now uses the same mechanism as ARM — native cross-compile, only the test binaries under `qemu-ppc64le`. Full suite passes; the job went from ~7 min to ~1 min.

.github/workflows/c-fortran-test-arm64.yml
hughcars: Fixed in fdbd83d: full `make prove`, all CPU backends, GCC/Clang, FP64/FP32, native + QEMU. Fortran skipped on FP32 only (`real*8`), as on x86.

tests/t360-basis.c
```diff
@@ -36,11 +36,13 @@ int main(int argc, char **argv) {
     // Check area
     {
       const CeedScalar *v_array;
-      CeedScalar        area = 0.0;
+      const double      expected  = 2.0 * CeedIntPow(2, dim);
+      const double      tolerance = fmax(5E-6, 100. * CEED_EPSILON * expected);
+      double            area      = 0.0;
```
jeremylt: Should these be doubles or `CeedScalar`?

backends/sve/ceed-sve-serial.c
```diff
@@ -0,0 +1,39 @@
...
+  CeedCallBackend(CeedDestroy(&ceed_ref));
+
+  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "TensorContractCreate", CeedTensorContractCreate_Sve));
+  return CEED_ERROR_SUCCESS;
+}
+
+//------------------------------------------------------------------------------
+// Backend Register
+//------------------------------------------------------------------------------
+CEED_INTERN int CeedRegister_Sve_Serial(void) { return CeedRegister("/cpu/self/sve/serial", CeedInit_Sve, 35); }
```
jeremylt:
```suggestion
CEED_INTERN int CeedRegister_Sve_Serial(void) { return CeedRegister("/cpu/self/sve/serial", CeedInit_Sve, 36); }
```
They should never collide, but we don't duplicate priority numbers

backends/sve/ceed-sve-blocked.c
```diff
@@ -0,0 +1,39 @@
...
+  CeedCallBackend(CeedDestroy(&ceed_ref));
+
+  CeedCallBackend(CeedSetBackendFunction(ceed, "Ceed", ceed, "TensorContractCreate", CeedTensorContractCreate_Sve));
+  return CEED_ERROR_SUCCESS;
+}
+
+//------------------------------------------------------------------------------
+// Backend Register
+//------------------------------------------------------------------------------
+CEED_INTERN int CeedRegister_Sve_Blocked(void) { return CeedRegister("/cpu/self/sve/blocked", CeedInit_Sve, 30); }
```
jeremylt:
```suggestion
CEED_INTERN int CeedRegister_Sve_Blocked(void) { return CeedRegister("/cpu/self/sve/blocked", CeedInit_Sve, 31); }
```

tests/junit_common.py
```diff
@@ -607,6 +608,11 @@ def run_test(index: int, test: str, spec: TestSpec, backend: str,
     for i, arg in enumerate(run_args):
         if '{ceed_resource}' in arg:
             run_args[i] = arg.replace('{ceed_resource}', backend.replace('/', '-'))
+    # Prefix the test executable, not mpiexec, and preserve arguments for shell execution.
+    runner = shlex.join(shlex.split(os.environ.get('CEED_TEST_RUNNER', '')))
```
jeremylt: @zatkins-dev for review of the changes here since you know this file best

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ These functions will be removed when the SYCL backends are updated to reflect th
 
 ### New features
 
+- Add vector-length-agnostic Arm SVE serial and blocked CPU backends.
```
jeremylt: New entries go at the bottom of this list, so it is in chronological order instead of reverse chronological

jeremylt (changes requested):
Looks good, just a couple of minor things from me

jeremylt:
> So I do not think the unconditional generic version is worth proposing. It would need a backend-owned cost decision or another TensorContract hook. I could take a look at that as a separate follow-up if you'd be open to it.

Yeah, I'd be curious to see that, though the PR that adds an even-odd decomposition of the basis application may change that investigation, so you may want to wait on #2010

momoquq:
Thanks — adding some native test results for `7b18a0a7ea91`.

I tested this revision on a Kunpeng AArch64 server with a 512-bit SVE vector length: C-only, FP64, GCC 12.3.1, with key flags `-O3 -march=armv8-a+sve -ffp-contract=fast -fopenmp-simd`.

`make prove search=t3` passed for all 44 selected t3xx test binaries on each of these resources — 176 test/backend combinations, with no skipped cases:

- `/cpu/self/sve/serial`
- `/cpu/self/sve/blocked`
- `/cpu/self/opt/serial`
- `/cpu/self/opt/blocked`

In addition to these upstream tests, I ran local correctness checks on the same node for each resource: 120 TensorContract cases against a scalar reference and four BasisApply cases (1D interp/grad, forward/transpose), with a normalized-error tolerance of 1e-12. Both the direct TensorContract checks and the exercised BasisApply paths reached the expected SVE or opt contraction implementations.

This covers the selected FP64 correctness tests. I haven’t run the full suite, FP32, or BP/performance tests yet; I’ll follow up with BP results.

I’ve also been working on SVE kernels, but other commitments have slowed my progress. This PR is further along than my implementation. Thanks for the great work! I’d like to build on it and help with follow-up optimizations.

By the way, I have access to SME-capable hardware and would like to contribute an SME backend. For an initial implementation, would I reusing the existing Basis implementations and specializing `CeedTensorContract`, as in this PR?

jeremylt:
Thanks @momoquq 

Yeah, I prefer to keep the TensorContraction related logic in those objects whenever possible to keep the new backends capable of meeting the full range CeedBasis behavior.

We can definitely modify the TensorContraction interface as needed for performance

hughcars:
Thank you very much for the feedback! Happy to get to contribute to the project, I'll reply in line to the comments. Since I marked this ready a colleague pointed out to me that gcc14 apparently has better support for this instructions so I decided to try even more experiments, and unfortuantely found that what I have right now is actually a bit too gcc13 tuned, and that using gcc14 actually resulted in worse performance. I am working on a fix, and hopefully can test on some more compilers. I suspect a more robust solution is possible, and that it was just over fitting. I don't want the performance to be so compiler specific.

tests/t360-basis.c
hughcars: Good catch, should be CeedScalar, fixed in 034274f.

backends/sve/ceed-sve-serial.c
hughcars: Done in 034274f.

backends/sve/ceed-sve-blocked.c
hughcars: Done in 034274f.

CHANGELOG.md
hughcars: Ah, moved in 034274f.

hughcars:
Ok, pushed the fix in ae7e0bc. Turned out the old kernel was computing every address from the loop counters, and gcc14 and clang were both strength reducing each of those into its own induction variable and then spilling. I restructured it to walk a fixed set of row pointers, closer to how the avx backend does it, and added a proper C=1 kernel while I was in there.

On Graviton4 at -O3 this is +2% on gcc13, +8% on gcc14, +4% on clang18 for hex BP3, so all three compilers are now within ~3% of each other rather than 10%. On my wider set of operator workloads it's about 20% faster overall and ~25% on serial layouts. Still bitwise identical to ref.

Since I had the boxes up I also ran BP3 on Graviton3/4/5 (c7g/c8g/c9g 16xlarge, 16 ranks, gcc13 -O3), plot below. Interesting that on the 256-bit Graviton3 the sve backend is at or above libxsmm at every degree, while on the 128-bit parts it's ahead at p≤2 and 4-6% behind above that. One gotcha for anyone trying this on Graviton5: clang-18 doesn't know Neoverse-V3 yet, so `-mcpu=native` falls back to generic without SVE and the backend gets disabled; use `-mcpu=neoverse-v2`.

CEED BP3 peak throughput across Graviton generations

I also tried a few of the obvious next steps — pair loads with lane-indexed FMLA to halve the t broadcasts, taller 8-row tiles in the narrow column blocks, and a 3-row tile for the odd remainders — but they added a lot of complexity without buying a consistent win across shapes, compilers and vector lengths (a few percent on hex at p≥3, a few percent lost on the small p2 tet cases and at -O1), so I opted to keep this simpler version.

@momoquq thanks for running it on SVE-512, that's the first native result at a vector length other than 128, really useful.

tests/junit_common.py
zatkins-dev: It's honestly a pretty elegant solution, but I don't love that it would be introducing a libCEED specific option into the `junit_common.py` file. I think I would prefer an interface added to the `SuiteSpec` class. I think it would be reasonable to change `SuiteSpec.get_run_path()` to return either a `Path` or a list of `str`, then add this functionality into `CeedTestSuite.get_run_path()`. Thoughts?

jeremylt: Seems like a good change to me. @hughcars would you like to tackle this or have @zatkins-dev do the change since they're the owner on this file? Seems like this is otherwise in really good shape? We can always do more follow-up tuning in future PRs too.

hughcars: Happy to, done in 3d8948c: `get_run_path` can return either a `Path` or a list of args, and `CeedSuiteSpec.get_run_path` handles `CEED_TEST_RUNNER`, so nothing libCEED specific is left in junit_common.py. Thanks @zatkins-dev, that's a much cleaner seam.

jeremylt: Looks good. I'll wait for @zatkins-dev to reply, but is there anything else open as a ToDo for this PR?

hughcars: Nothing else on my side, modulo review feedback.

zatkins-dev (approved):
Looks good to me!

tests/junit_common.py
zatkins-dev: looks good to me, thanks!

jeremylt:
Thanks for this work @hughcars ! This has been an open ToDo for way too long and I'm excited to finally better support these chips.

PR #2023, opened 2026-09-10
make - minor fixes

jeremylt:
Purpose:

- I don't think 'tidy-fix' needs to be forced to be serial

- used call quiet for Clang Tidy

- Added `make lint` for format+tidy since those are both in CI.

Closes: #N/A

LLM/GenAI Disclosure:

None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1045,33 +1045,35 @@ format-c  :
 	@$(MAKE) -j1 --no-print-directory --no-keep-going tidy-fix
 
 format-py :
-	$(AUTOPEP8) $(AUTOPEP8_OPTS) $(format.py)
+	$(call quiet,AUTOPEP8) $(AUTOPEP8_OPTS) $(format.py)
 
-format-ot:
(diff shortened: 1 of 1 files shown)


zatkins-dev:
I think it does need to be serial *if* there are fixes applied. Otherwise, multiple fixes can be applied simultaneously, leading to write conflicts.

zatkins-dev:
You can test that by intentionally introducing a tidy error into a file included by more than one file

jeremylt:
But the same make target should not be duplicated?

jeremylt:
Locally, introducing a style error into `ceed.h` and running `make format -j` worked fine

jeremylt:
Clang agrees its a theoretical issue, so I'll drop that and continue to suffer the slowness :sob:

zatkins-dev:
> But the same make target should not be duplicated?

The issue is that the checks (and fixes) are applied recursively to included files. Very annoying in this case, since the desired fix doesn't actually need that kind of semantic context.

PR #2024, opened 2026-09-11
CPU JiT Backend

### LINKED ISSUE #1239: Clang JIT CPU Backend
Clang 16 now supports JIT. An interesting small project could be to create a `/cpu/self/clang-jit` backend that provides JITed tensor contraction kernels. If we see performance that is in the neighborhood of AVX or libXSMM, this could be a way to ship a faster CPU backend with fewer dependencies.

See Serac for reference:
https://github.com/LLNL/serac/blob/prototype/adjoints_with_internal_variables/tests/jit/basic_jit.cpp
https://github.com/LLNL/serac/blob/prototype/adjoints_with_internal_variables/include/JIT.hpp

(This repo comes from a member of Jamie's Smith team)

jeremylt:
Purpose:

This PR is for a CPU JiT backend. At the moment it doesn't run, but the bones and basic vision should (I hope) be clear. I am wondering if we can make some performance gains by compiling to the specific basis and restriction parameters.

In principle, once this works, we can drop in XSMM support easily.

Closes: #1239

LLM/GenAI Disclosure:

Describe any LLM and GenAI usage here.

None


jeremylt:
Sample generated code:
```C++
#define BLOCK_SIZE 8

#include <ceed/jit-source/cpu-gen/cpu-jit.h>

#include <ceed/jit-source/cpu-gen/cpu-gen-utils.h>

// Ceed object templates
#include <ceed/jit-source/cpu-gen/cpu-gen-restriction-templates.h>
#include <ceed/jit-source/cpu-gen/cpu-gen-basis-templates.h>

// User QFunction source
#include "/home/jeremy/Dev/libCEED/tests/t500-operator.h"

static int CeedCpuGenOperator_setup(void *ctx, const FieldData_Cpu_Gen *inputs, FieldData_Cpu_Gen *outputs) {
  // Operator constants
  const CeedInt Q = 8;
  const CeedInt Q_1d = 8;
  const CeedInt num_elem = 15;
  const CeedInt num_blocks = (num_elem / BLOCK_SIZE) + !!(num_elem % BLOCK_SIZE);

  // Field constants
  // -- Input Fields
  // ---- Input Field 0: weight
  const CeedInt num_q_comp_in_0 = 1;
  const CeedInt dim_in_0 = 1;
  const CeedInt P_1d_in_0 = 2;
  // ---- Input Field 1: dx
  const CeedInt elem_size_in_1 = 2;
  const CeedInt num_comp_in_1 = 1;
  const CeedInt num_q_comp_in_1 = 1;
  const CeedInt dim_in_1 = 1;
  const CeedInt P_1d_in_1 = 2;
  
  // -- Output Fields
  // ---- Output Field 0: rho
  const CeedInt elem_size_out_0 = 8;
  const CeedInt num_comp_out_0 = 1;
  const CeedInt num_q_comp_out_0 = num_comp_out_0;
  const CeedInt dim_out_0 = 1;
  const CeedInt P_out_0 = elem_size_out_0;

  // Scratch restriction buffer space
  const CeedInt max_e_vec_buffer_size = 2;
  CeedScalar e_vec_scratch[max_e_vec_buffer_size * BLOCK_SIZE];

  // Loop over blocks
  for (CeedInt block = 0; block < num_blocks; block++) {
    // -- Input ElemRestrictions and Bases
    // ---- Input Field 0: weight
    // ------ Restriction Type: none
    // ------ Evaluation Mode: quadrature weights
    CeedScalar q_vec_in_0[Q * BLOCK_SIZE];
    CeedCall(CeedBasis_Apply_Weight_Tensor_1D<BLOCK_SIZE, Q_1d>(inputs[0].weights, q_vec_in_0));
    // ---- Input Field 1: dx
    // ------ Restriction Type: offset
    CeedScalar *e_vec_in_1 = e_vec_scratch;
    {
      const CeedInt comp_stride = 1;
      
      CeedCall(CeedElemRestriction_Apply_NoTranspose_Offset<BLOCK_SIZE, num_comp_in_1, elem_size_in_1, num_elem, comp_stride>(block, inputs[1].offsets, inputs[1].array, e_vec_in_1));
    }
    // ------ Evaluation Mode: gradient
    CeedScalar q_vec_in_1[num_q_comp_in_1 * Q * BLOCK_SIZE];
    CeedCall(CeedBasis_Apply_NoTranspose_Grad_Tensor_1D<BLOCK_SIZE, num_comp_in_1, P_1d_in_1, Q_1d>(inputs[1].interp, inputs[1].grad, e_vec_in_1, q_vec_in_1));
    
    // -- QFunction
    // ---- QFunction inputs
    const CeedScalar* q_vecs_in[2] = {
      q_vec_in_0,
      q_vec_in_1,
    };

    // ---- QFunction outputs
    CeedScalar q_vec_out_0[num_q_comp_out_0 * Q * BLOCK_SIZE];
    CeedScalar* q_vecs_out[1] = {
      q_vec_out_0,
    };

    // ---- Call User QFunction
    CeedCall(setup(ctx, Q * BLOCK_SIZE, q_vecs_in, q_vecs_out));

    // -- Output Bases and ElemRestrictions
    // ---- Output Field 0: rho
    // ------ Evaluation Mode: none
    CeedScalar *e_vec_out_0 = q_vec_out_0;
    // ------ Restriction Type: strided
    {
      const CeedInt strides_0 = 1, strides_1 = 8, strides_2 = 8;
      
      CeedCall(CeedElemRestriction_ApplyAdd_Transpose_Strided<BLOCK_SIZE, num_comp_out_0, elem_size_out_0, num_elem, strides_0, strides_1, strides_2>(block, e_vec_out_0, outputs[0].array));
    }
  }
  return CEED_ERROR_SUCCESS;
}
```

jeremylt:
@zatkins-dev there's bugs somewhere in my templated code, but this compiles and runs now, so I think the idea is workable

zatkins-dev:
i'll take a look, i like templates

jeremylt:
As with [cuda,hip]/gen, `CEED_DEBUG=1` turn on printing out the generated code, but it also gets dumped into the folder `temp` (same folder for CUDA + Rust)

zatkins-dev:
> ```c++
>   const CeedInt Q = 8;
>   const CeedInt Q_1d = 8;
>   const CeedInt num_elem = 15;
>   const CeedInt num_blocks = (num_elem / BLOCK_SIZE) + !!(num_elem % BLOCK_SIZE);
> 
>   // Field constants
>   // -- Input Fields
>   // ---- Input Field 0: weight
>   const CeedInt num_q_comp_in_0 = 1;
>   const CeedInt dim_in_0 = 1;
>   const CeedInt P_1d_in_0 = 2;
>   // ---- Input Field 1: dx
>   const CeedInt elem_size_in_1 = 2;
>   const CeedInt num_comp_in_1 = 1;
>   const CeedInt num_q_comp_in_1 = 1;
>   const CeedInt dim_in_1 = 1;
>   const CeedInt P_1d_in_1 = 2;
>   
>   // -- Output Fields
>   // ---- Output Field 0: rho
>   const CeedInt elem_size_out_0 = 8;
>   const CeedInt num_comp_out_0 = 1;
>   const CeedInt num_q_comp_out_0 = num_comp_out_0;
>   const CeedInt dim_out_0 = 1;
>   const CeedInt P_out_0 = elem_size_out_0;
> ```

If we're using C++ to compile this, these should really be `constexpr` to give the compiler extra hints

jeremylt:
Feel free to push direct to the branch - my puppy needs exercise *now*, according to him :)

jeremylt:
just occurred to me - changing the JiT compile flags to -O0 -g would probably help see what the error is

zatkins-dev:
I found it

zatkins-dev:
rather, I found the cause of the seg fault -- there's still at least something wrong

jeremylt:
Awesome. I figure there are multiple bugs - I got excited and started coding way too fast.

zatkins-dev:
aha! it was the tensor contractions

jeremylt:
Ok, t500 rens now, so it compiles and executes the operator function without issues. But there's a memory leak somewhere, and t501 fails, so there's a correctness issue. And AtPoints still needs to be piped in

jeremylt:
Awesome, hopefully our fixes complement each other

it should be relatively easy to setup the even-odd stuff and to replace this cantraction with XSMM, I think

zatkins-dev:
Hmmm this is working locally...I wonder what the issue is

jeremylt:
If dlopen is too fiddly, we may want to use LLVM instead

zatkins-dev:
We're going to need to be careful -- having multiple functions with the same name seems to result in unpredictable behavior (see t506)

backends/cpu-gen/ceed-cpu-compile.cpp
```diff
@@ -0,0 +1,275 @@
...
+static inline int CeedJitGetOpts_Cpu(Ceed ceed, const char ***opts, int *num_opts) {
+  int opts_count = 2;
+
+  // Standard options
+  CeedCallBackend(CeedCalloc(opts_count, opts));
+  // TODO: Revert
+  // CeedCallBackend(CeedStringAllocCopy("-march=native", (char **)&(*opts)[0]));
+  // CeedCallBackend(CeedStringAllocCopy("-O3", (char **)&(*opts)[1]));
+  CeedCallBackend(CeedStringAllocCopy("-g", (char **)&(*opts)[0]));
+  CeedCallBackend(CeedStringAllocCopy("-O0", (char **)&(*opts)[1]));
```
jrwrigh, line 66: I'm guessing y'all have probably already thought about this, but having the compiler flags be editable/overwritable would be nice. There are times when I've noticed that `-02` is actually faster than `-O3` in some cases (most of which I never fully understand).

zatkins-dev: Definitely agree

zatkins-dev:
I think tensor is now working right. Non-tensor is still giving bad results, and at-points is still unimplemented

zatkins-dev:
It's so close, I think the collocated grad has some bug still (and fortran is being weird). But basically all tests pass!

backends/cpu-gen/ceed-cpu-gen-operator-build.cpp
```diff
@@ -0,0 +1,957 @@
...
+
+  // Compute key
+  {
+    std::size_t hash = std::hash<std::string>{}(code.str());
+
+    // Wrapper function with hash
+    code << tab << "extern \"C\" int CeedOperator_" << hash << "(void *ctx, const InputFieldData_Cpu_Gen *inputs, OutputFieldData_Cpu_Gen *outputs) {\n";
+    tab.push();
+    code << tab << "CeedCall(" << operator_name << "(ctx, inputs, outputs));\n";
+    code << tab << "return CEED_ERROR_SUCCESS;\n";
```
zatkins-dev: I think that if we want to do this, the function that is being wrapped needs to be static inline and not extern. Otherwise we can run into symbol issues again

zatkins-dev:
Also, I think that the CeedCall in the tensor contractions is going to interfere with some vectorization/optimization potential, since those are suddenly potential exit points. Would be worth investigating.

backends/cpu-gen/ceed-cpu-gen-operator-build.cpp
jeremylt: good catch

zatkins-dev: we may also want to add the equivalent of CEED_INTERN to them to prevent visibility further

backends/cpu-gen/ceed-cpu-compile.cpp
jeremylt: really, we should probably store the OPT string from the CPU compilation and reuse it

jeremylt:
ToDo:
- [x] Pipe in AtPoints (shouldn't be too bad)
- [x] Rust (and other reasons) probably want us to guard if we want to 'turn on' CXX (by default no unless GPU, XSMM, or CPU/GEN requested?)
- [x] ICX might not have a CXX compiler installed?
- [ ] How to handle external libraries called by user QFunctions (Enzyme!)
- [ ] ...

jeremylt:
Noether's CPU jobs are happy :tada:

jeremylt:
We should note in the documentation or somewhere that this is in large part based upon what I learned maintaining and adding features to @YohannDudouit 's GPU gen backends

jedbrown:
Nick work, team. At this point, linking libclang versus calling out to ordinary compilers would be a simple change (no need to address in this PR). I think what you have is a bit more portable because lots of systems don't have libclang. It'll be interesting to compare perf.

zatkins-dev:
@jeremylt bad news -- the JiT backend was actually falling back due to the incorrect function having the `static inline` annotation. We're still failing a few tests after fixing that

zatkins-dev:
We're officially down to just fortran errors!

zatkins-dev:
I've got a very rough at-points impl set up. The idea to pad the blocked e vector, like with the GPU gen, but then to only apply the QFunction at the real points. This makes the array sizes and vectorization happy while avoiding too much extra work.

zatkins-dev:
Currently I've only got the 1D interp/grad and 2D interp set up, but I think the general approach for the contractions is making sense at this point. It also makes very clear what the extra cost is, IMO.

backends/cpu-gen/ceed-cpu-gen-operator-build.cpp
```diff
@@ -0,0 +1,1056 @@
...
+      } break;
+      case CEED_RESTRICTION_STRIDED: {
+        CeedInt strides[3];
+
+        CeedCallBackend(CeedElemRestrictionGetStrides(elem_rstr, strides));
+        CeedCallBackend(CeedElemRestrictionCreateBlockedStrided(ceed, num_elem, elem_size, block_size, num_comp, l_size, strides, &block_rstr));
+      } break;
+      // LCOV_EXCL_START
+      case CEED_RESTRICTION_POINTS: {
+        CeedCallBackend(CeedElemRestrictionReferenceCopy(elem_rstr, &block_rstr));
```
jeremylt, line 174: We don't have a blocked way to apply AtPoints, so this should be an error, not quietly playing along with a case we can't support (and that the if clause forbid)

include/ceed/jit-source/cpu-gen/cpu-gen-basis-templates.h
```diff
@@ -0,0 +1,441 @@
...
+    for (CeedInt j = 0; j < Q_1D; j++) {
+      for (CeedInt b = 0; b < BLOCK_SIZE; b++) v[(i * Q_1D + j) * BLOCK_SIZE + b] = weights_1d[j];
+    }
+  }
+  for (CeedInt i = 0; i < Q_1D; i++) {
+    for (CeedInt j = 0; j < Q_1D; j++) {
+      for (CeedInt k = 0; k < Q_1D; k++) {
+        const CeedScalar w = weights_1d[j] * v[((i * Q_1D + j) * Q_1D + k) * BLOCK_SIZE];
+
+        for (CeedInt b = 0; b < BLOCK_SIZE; b++) v[((i * Q_1D + j) * Q_1D + k) * BLOCK_SIZE + b] = w;
```
jeremylt: why copy it 8 times when its an intermediate calculation?

include/ceed/jit-source/cpu-gen/cpu-gen-basis-templates.h
```diff
@@ -0,0 +1,441 @@
...
+    }
+  }
+  return CEED_ERROR_SUCCESS;
+}
+
+template <CeedInt BLOCK_SIZE, CeedInt Q_1D>
+static inline int CeedBasis_Apply_Weight_Tensor_3D(const CeedScalar *weights_1d, CeedScalar *v) {
+  for (CeedInt i = 0; i < Q_1D * Q_1D; i++) {
+    for (CeedInt j = 0; j < Q_1D; j++) {
+      for (CeedInt b = 0; b < BLOCK_SIZE; b++) v[(i * Q_1D + j) * BLOCK_SIZE + b] = weights_1d[j];
```
jeremylt: same, why replicate this 8x?

interface/ceed-basis.c
```diff
@@ -1143,6 +1143,55 @@ int CeedBasisSetTensorContract(CeedBasis basis, CeedTensorContract contract) {
...
+
+  @param[in,out] basis           `CeedBasis`
+  @param[out]    basis_chebyshev Pointer to store `CeedBasis`, or `NULL` if not needed
+  @param[out]    vec_chebyshev   Pointer to store `CeedVector`, or `NULL` if not needed
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
+**/
+int CeedBasisGetChebyshevData(CeedBasis basis, CeedBasis *basis_chebyshev, CeedVector *vec_chebyshev) {
```
jeremylt, line 1157: Why not just get the array we want instead of making another basis?

zatkins-dev: We can do that too, the Basis object just already has the chebyshev basis object. This is more because I didn't want to rewrite the same logic in a fourth (?) place.

include/ceed/jit-source/cpu-gen/cpu-gen-basis-templates.h
zatkins-dev: To be honest, the existing implementation was wrong, so I basically manually unrolled the ref implementation. Feel free to optimize it down, but just make sure the examples still pass (since they're the main places 3d is tested).

jeremylt: we should fix this in ref too, lol

backends/cpu-gen/ceed-cpu-gen-operator-build.cpp
zatkins-dev: I think I have a blocked strategy working right now, it's just that the blocking only really helps the basis operations (which may or may not matter, but it would be good to see).

interface/ceed-basis.c
jeremylt: ah, right. We should pull this into a separate PR

zatkins-dev: agreed, it's currently duplicated in interface, cuda gen and hip gen at least

zatkins-dev:
tbh I have no idea how the fortran support works, if you have any idea why it's not working I'd appreciate. I assume there's some kind of ABI incompatibility?

jeremylt:
I thought Fortran was working?

zatkins-dev:
> I thought Fortran was working?

No, the whole backend was silently failing due to the wrong function being marked extern (there's some comments about it from earlier today). There's a couple fortran tests failing.

jeremylt:
Looks like most of the Fortran failures are a compilation issue (on Noether). Uncertain about 531 as that has no QFunction dota which really is the only sharp edge with Fortran

jeremylt:
its all the user QFunction that's failing for ex*

zatkins-dev:
hmm, at least locally, it was giving wrong answers without any other errors.

backends/cpu-gen/ceed-cpu-gen-operator-build.cpp
jeremylt: Currently `is_at_points` blocks the `block_size` from being anything other than 1

jeremylt:
Ugh, I switched away from using the inner context, but I shouldn't have . I knew better! See here: https://github.com/CEED/libCEED/blob/3381e6d3ca5750f9dad1bcf0a0bec6afee7db4aa/interface/ceed-fortran.c#L690
This fixes Fortran

jeremylt:
The AtPoints gradient is getting twisted up, which is causing a bunch of issues. Instead of at a point for 591 getting 0.5, 0.0, 0.0, 0.5 we get instead 0.5, 0.5, 0.0, 0.0

jeremylt:
The Fortran t531 issue is really strange and I honestly don't know what exactly is wrong there

zatkins-dev:
I got at-points operators working! They're also _much_ faster than the existing CPU implementation, to the point that I think we should rewrite the interface implementation to use the approach in this MR of precomputing chebyshev evaluations at the reference points. For example, running locally on a Zen 3 processor with OPT flags:

```
O3 -g -march=native -fPIC -fno-omit-frame-pointer -ffp-contract=fast
```

<details>
<summary>/cpu/self/avx/serial</summary>

```console
$ build/petsc-bpsswarm -problem bp4 -dm_plex_dim 3 -dm_plex_box_faces 40,40,40 -dm_plex_simplex 0 -swarm uniform -points_per_cell 64 -degree 3 -q_extra 0 -benchmark -ceed /cpu/self/avx/serial

-- CEED Benchmark Problem 4 -- libCEED + PETSc --
  MPI:
    Hostname                                : zatkins-dev
    Total ranks                             : 1
    Ranks per compute node                  : 1
  PETSc:
    PETSc Vec Type                          : seq
  libCEED:
    libCEED Backend                         : /cpu/self/avx/serial
    libCEED Backend MemType                 : host
  Mesh:
    Solution Order (P)                      : 4
    Quadrature  Order (Q)                   : 4
    Additional quadrature points (q_extra)  : 0
    Global nodes                            : 1685159
    Local Elements                          : 64000
    Owned nodes                             : 1685159
    DoF per node                            : 3
  Swarm:
    Global points                           : 4096000
    Local points                            : 4096000
    Avg points per cell                     : 64
    Point distribution                      : uniform
  KSP:
    KSP Type                                : cg
    KSP Convergence                         : DIVERGED_ITS
    Total KSP Iterations                    : 5
    Final rnorm                             : 1.435122e+00
  Performance:
    L2 Error                                : 6.132548e-03
    CG Solve Time                           : 22.196 (22.196) sec
    DoFs/Sec in CG                            : 1.13883 (1.13883) million
```

</details>

<details>
<summary>/cpu/self/gen/serial</summary>

```console
$ build/petsc-bpsswarm -problem bp4 -dm_plex_dim 3 -dm_plex_box_faces 40,40,40 -dm_plex_simplex 0 -swarm uniform -points_per_cell 64 -degree 3 -q_extra 0 -benchmark -ceed /cpu/self/gen/serial

-- CEED Benchmark Problem 4 -- libCEED + PETSc --
  MPI:
    Hostname                                : zatkins-dev
    Total ranks                             : 1
    Ranks per compute node                  : 1
  PETSc:
    PETSc Vec Type                          : seq
  libCEED:
    libCEED Backend                         : /cpu/self/gen/serial
    libCEED Backend MemType                 : host
  Mesh:
    Solution Order (P)                      : 4
    Quadrature  Order (Q)                   : 4
    Additional quadrature points (q_extra)  : 0
    Global nodes                            : 1685159
    Local Elements                          : 64000
    Owned nodes                             : 1685159
    DoF per node                            : 3
  Swarm:
    Global points                           : 4096000
    Local points                            : 4096000
    Avg points per cell                     : 64
    Point distribution                      : uniform
  KSP:
    KSP Type                                : cg
    KSP Convergence                         : DIVERGED_ITS
    Total KSP Iterations                    : 5
    Final rnorm                             : 1.435122e+00
  Performance:
    L2 Error                                : 6.132548e-03
    CG Solve Time                           : 8.29756 (8.29756) sec
    DoFs/Sec in CG                            : 3.04636 (3.04636) million
```

</details>

zatkins-dev:
The difference is much large for higher degree -- I suspect this is asymptotically better.

<details>
<summary>/cpu/self/avx/serial</summary>

```console
$ build/petsc-bpsswarm -problem bp4 -dm_plex_dim 3 -dm_plex_box_faces 40,40,40 -dm_plex_simplex 0 -swarm uniform -points_per_cell 64 -degree 4 -q_extra 0 -benchmark -ceed /cpu/self/avx/serial

-- CEED Benchmark Problem 4 -- libCEED + PETSc --
  MPI:
    Hostname                                : zatkins-dev
    Total ranks                             : 1
    Ranks per compute node                  : 1
  PETSc:
    PETSc Vec Type                          : seq
  libCEED:
    libCEED Backend                         : /cpu/self/avx/serial
    libCEED Backend MemType                 : host
  Mesh:
    Solution Order (P)                      : 5
    Quadrature  Order (Q)                   : 5
    Additional quadrature points (q_extra)  : 0
    Global nodes                            : 4019679
    Local Elements                          : 64000
    Owned nodes                             : 4019679
    DoF per node                            : 3
  Swarm:
    Global points                           : 4096000
    Local points                            : 4096000
    Avg points per cell                     : 64
    Point distribution                      : uniform
  KSP:
    KSP Type                                : cg
    KSP Convergence                         : DIVERGED_ITS
    Total KSP Iterations                    : 5
    Final rnorm                             : 5.465017e+00
  Performance:
    L2 Error                                : 3.354294e-01
    CG Solve Time                           : 41.2598 (41.2598) sec
    DoFs/Sec in CG                            : 1.46136 (1.46136) million
```

</details>

<details>
<summary>/cpu/self/gen/serial</summary>

```console
$ build/petsc-bpsswarm -problem bp4 -dm_plex_dim 3 -dm_plex_box_faces 40,40,40 -dm_plex_simplex 0 -swarm uniform -points_per_cell 64 -degree 4 -q_extra 0 -benchmark -ceed /cpu/self/gen/serial

-- CEED Benchmark Problem 4 -- libCEED + PETSc --
  MPI:
    Hostname                                : zatkins-dev
    Total ranks                             : 1
    Ranks per compute node                  : 1
  PETSc:
    PETSc Vec Type                          : seq
  libCEED:
    libCEED Backend                         : /cpu/self/gen/serial
    libCEED Backend MemType                 : host
  Mesh:
    Solution Order (P)                      : 5
    Quadrature  Order (Q)                   : 5
    Additional quadrature points (q_extra)  : 0
    Global nodes                            : 4019679
    Local Elements                          : 64000
    Owned nodes                             : 4019679
    DoF per node                            : 3
  Swarm:
    Global points                           : 4096000
    Local points                            : 4096000
    Avg points per cell                     : 64
    Point distribution                      : uniform
  KSP:
    KSP Type                                : cg
    KSP Convergence                         : DIVERGED_ITS
    Total KSP Iterations                    : 5
    Final rnorm                             : 5.465017e+00
  Performance:
    L2 Error                                : 3.354294e-01
    CG Solve Time                           : 11.7102 (11.7102) sec
    DoFs/Sec in CG                            : 5.14896 (5.14896) million
```

</details>

jeremylt:
nice! So fixing the CXX CI issues and debugging t531-f are the outstanding issues

backends/cpu-gen/ceed-cpu-compile.cpp
```diff
@@ -219,7 +219,7 @@ static inline int CeedCompileCore_Cpu(Ceed ceed, const char *source, const char
     }
 
     // Compile wrapper kernel
-    std::string command = std::string(cxx) + " -shared -fPIC -rdynamic";
+    std::string command = std::string(cxx) + " -shared -fPIC -rdynamic -fsplit-stack";
```
jeremylt: need to cross check these for full support with all compilers (GCC, Clang, Intel)

jeremylt: yeah, looks like ARM didn't like '-fsplit-stack`

jeremylt:
Sanity check: generated code is nearly identical
```shell
$ diff temp/function_1630473341_CeedOperator_601880389243708707.cpp temp/function_1639939767_CeedOperator_10122776004029177837.cpp 
33c33
<   // ---- Input Field 1: qdata
---
>   // ---- Input Field 1: q data
66c66
<     // ---- Input Field 1: qdata
---
>     // ---- Input Field 1: q data
108c108
< extern "C" int CeedOperator_601880389243708707(void *ctx, const PointsData_Cpu_Gen *points, const InputFieldData_Cpu_Gen *inputs, OutputFieldData_Cpu_Gen *outputs) {
---
> extern "C" int CeedOperator_10122776004029177837(void *ctx, const PointsData_Cpu_Gen *points, const InputFieldData_Cpu_Gen *inputs, OutputFieldData_Cpu_Gen *outputs) {
```

backends/cpu-gen/ceed-cpu-compile.cpp
zatkins-dev: ugh that makes sense, they probably have a different flag for it.

jeremylt:
This t531-f bug is wild. The output from the QFs is identical to other (passing) cases. The output basis action and restriction are the same. The linearized output is correct from the new backends..... but the output from the diff operator is wrong. Subtle memory corruption maybe?

zatkins-dev:
Are we wanting to do assembly in this PR? Or push that to a follow-up?

jeremylt:
Follow-up I think. Also XSMM jit backend and using the even/odd stuff

zatkins-dev:
I figured out the t531 fortran bug! Nothing so complicated, just a simple indexing error in the fortran header. Since most backends only ever use the C or Fortran QFunction source, we never had a way to notice the difference.

jeremylt:
Good catch, that's not where I was looking at all

jeremylt:
ToDo:
- [x] Fix ICC issue
- [x] Fix macos Rust issue (no CXX by default?)
- [x] Move Basis function to separate PR (and add to release notes)
- [x] Add this new backend to release notes
- [ ] Create issue with planned future improvements

backends/cpu-gen/ceed-cpu-gen-operator-build.cpp
jeremylt: Do we want to take a crack at the blocked version here or leave it 1 elem at a time for this initial (big) push?

zatkins-dev: let me try increasing the block size and seeing how bad it breaks

jeremylt:
cool, big todo is an issue with future hopes and plans (xsmm, blocked at points, even/odd, assembly, world peace)

jeremylt:
Squash+merge because... uhhh... I did not make a clean history

backends/cpu-gen/ceed-cpu-gen-operator-build.cpp
zatkins-dev: yeah no those tensor contractions are going to be annoying bc there isn't really a correct place to put the block size. Serial is already super fast, let's just stick with that for now.

zatkins-dev:
I added a follow-up thread here: #2032

There are sub-issues for each of the major tasks, as I like the dopamine of closing issues. Feel free to edit/add anything to them as you see necessary

backends/cpu-gen/ceed-cpu-compile.cpp
```diff
@@ -0,0 +1,301 @@
...
+        // Only one process needs to succeed
+        mkdir("temp", 0777);
+        chmod("temp", 0777);
+      }
+    }
+
+    // Write code to temp file
+    {
+      std::string filename = filename_base + ".cpp";
+      FILE       *file     = fopen(filename.c_str(), "w");
```
jeremylt, line 174: ToDo: Should check this cache and just use the file if it exists already (would required dropping the build_id from the filename though). Follow-up issue

jeremylt:
Something is funky when I run `make junit -j` for Ratel. I get a lot of `/bin/sh: 1: build/ex02-quasistatic: Text file busy`

jeremylt:
Ratel tests pass:
```shell
Running unit tests
- Testing with libCEED backends: /cpu/self
- Testing on 1 processes
prove -j 64 --exec 'python3 tests/junit.py' t000-init t001-view t002-view t003-ts-monitor t004-ts-checkpoint t005-config t010-eigensolver t050-mpm t100-static-elasticity t101-static-elasticity t102-static-elasticity t103-static-elasticity t110-static-elasticity t111-static-elasticity t120-static-elasticity t121-static-elasticity t122-static-elasticity t123-static-elasticity t130-static-elasticity-volume t203-quasistatic-elasticity-ts-monitor t211-quasistatic-elasticity t221-quasistatic-elasticity t222-quasistatic-elasticity t300-polygon-clipping t301-triangulate ex01-static ex02-quasistatic ex03-dynamic :: --petsc-arch arch-serial-cpu --ceed-backends /cpu/self    --with-uhyper 1  --mode tap --nproc 1 --pool-size 1 --search '.*'
t010-eigensolver ........................ ok                            
t005-config ............................. ok                            
t000-init ............................... ok                            
t301-triangulate ........................ ok                            
t300-polygon-clipping ................... ok                            
t001-view ............................... ok                            
t003-ts-monitor ......................... ok                            
t101-static-elasticity .................. ok                            
t002-view ............................... ok                            
t123-static-elasticity .................. ok                            
t122-static-elasticity .................. ok                            
t100-static-elasticity .................. ok                            
t111-static-elasticity .................. ok                            
t103-static-elasticity .................. ok                            
t130-static-elasticity-volume ........... ok                            
t211-quasistatic-elasticity ............. ok                            
t050-mpm ................................ ok                            
t221-quasistatic-elasticity ............. ok                            
t120-static-elasticity .................. ok                            
t121-static-elasticity .................. ok                            
t110-static-elasticity .................. ok                            
t102-static-elasticity .................. ok                            
t222-quasistatic-elasticity ............. ok                            
t203-quasistatic-elasticity-ts-monitor .. ok                            
ex03-dynamic ............................ ok                            
ex01-static ............................. ok                            
t004-ts-checkpoint ...................... ok                            
ex02-quasistatic ........................ ok       
All tests successful.
Files=28, Tests=314, 928 wallclock secs ( 0.15 usr  0.04 sys + 1896.45 cusr 132.28 csys = 2028.92 CPU)
Result: PASS
```

though, I think we'll continue to want to use other backends for CI (like we do with CUDA) so we don't pay for compile time.

We could flip the priority orders around to make this new backend something you have to request. Thoughts?

backends/cpu-gen/ceed-cpu-compile.cpp
zatkins-dev: it's not that simple unfortunately -- what if the user qfunction changed?

jeremylt: Yeah, we'll have to think on it because there's some sharp corners. But this has been a desire for a while on the GPU side.

jeremylt:
@zatkins-dev if you're happy with the SVE PR, then I'd like to merge that first, rebase this one, then update the required tests for merging, and then merge this one.

I'm a little worried about the possibility of race conditions in the temp files so I'm going to drop this to a lower priority number so its opt-in.

jeremylt:
t594 has a transient failure now, so that PR must have introduced or exposed a bug

PR #2027, opened 2026-09-15
Add specialization for `num_comp = 2` in  CeedElemRestrictionApply

Sbozzolo:
`CeedElemRestrictionCreate_Ref` missed specialization for the case with `num_comp = 2`. I found that adding this has no to marginally positive performance implication depending on the compiler and microarchitecture.

In the best case I've seen (on Graviton3 with GCC 15), this lead to a 15 % improvement in microbenchmarks. On Apple Silicon and GCC, this led to at most 3 % improvement. No improvement with GCC 11 or with Apple Clang (but also no regression).

Found in https://github.com/awslabs/palace/pull/932#discussion_r4009619217 (by an LLM, I assume).


jeremylt:
Yea, I was leaving it out since we don't have anyone that I am aware of using it in production, but no harm adding it.

Aside, I think it might be interesting to see if #2024 has a performance impact for y'all

jeremylt (approved):

PR #2030, opened 2026-09-17
CeedBasisGetChebyshevData

jeremylt:
Purpose:

Cherry-pick out CeedBasisGetChebyshevData and use it to reduce repeated code.

Closes: #N/A

LLM/GenAI Disclosure:

None


zatkins-dev:
There's one more: around `ceed-basis.c:516`

interface/ceed-basis.c
```diff
@@ -503,29 +503,7 @@ static int CeedBasisApplyAtPoints_Core(CeedBasis basis, bool apply_add, CeedInt
...
-    CeedCall(CeedBasisCreateTensorH1(ceed, dim, num_comp, P_1d, Q_1d, chebyshev_interp_1d, chebyshev_grad_1d, q_ref_1d, chebyshev_q_weight_1d,
-                                     &basis->basis_chebyshev));
-
-    // Cleanup
-    CeedCall(CeedFree(&chebyshev_interp_1d));
-    CeedCall(CeedFree(&chebyshev_grad_1d));
-    CeedCall(CeedFree(&chebyshev_q_weight_1d));
-    CeedCall(CeedDestroy(&ceed));
-  }
+  if (!basis->basis_chebyshev) CeedCall(CeedBasisGetChebyshevData(basis, &basis->basis_chebyshev, &basis->vec_chebyshev));
```
zatkins-dev: This isn't right -- the `basis->basis_chebyshev` and `basis->vec_chebyshev` are now considered implementation details and shouldn't be accessed directly. Instead, this should be a local basis and vec which are destroyed at the end.

jeremylt: fair - I was distracted trying to figure out why CUDA doesn't like this

backends/cuda-ref/ceed-cuda-ref-basis.c
```diff
@@ -167,15 +167,16 @@ static int CeedBasisApplyAtPointsCore_Cuda(CeedBasis basis, bool apply_add, cons
...
     // -- Create interp matrix to Chebyshev coefficients
     if (!data->d_chebyshev_interp_1d) {
-      CeedSize    interp_bytes;
-      CeedScalar *chebyshev_interp_1d;
+      CeedSize          interp_bytes;
+      const CeedScalar *chebyshev_interp_1d;
+      CeedBasis         basis_chebyshev;
 
+      CeedCallBackend(CeedBasisGetChebyshevData(basis, &basis_chebyshev, NULL));
+      CeedCallBackend(CeedBasisGetInterp1D(basis, &chebyshev_interp_1d));
```
zatkins-dev:
```suggestion
      CeedCallBackend(CeedBasisGetInterp1D(basis_chebyshev, &chebyshev_interp_1d));
```

zatkins-dev: looks like all of them are using the wrong basis

jeremylt: yea, just caught that. What I get for tinkering at 10:30

zatkins-dev (approved):
assuming tests pass I think this is good

PR #2031, opened 2026-09-17
xsmm - use pkgconfig instead of incorrectly assuming required libs

zatkins-dev:
Purpose:

XSMM doesn't always need `-lblas`, we should use their pkgconfig file to determine dependencies.

Closes: #ISSUE_NUMBER

LLM/GenAI Disclosure:

None


Makefile
```diff
@@ -1045,36 +1054,34 @@ format-c  :
 	@$(MAKE) -j1 --no-print-directory --no-keep-going tidy-fix
 
 format-py :
-	$(call quiet,AUTOPEP8) $(AUTOPEP8_OPTS) $(format.py)
+	$(AUTOPEP8) $(AUTOPEP8_OPTS) $(format.py)
```
jeremylt: is there a need to drop these `call quiet`s? I tend to use them and then use V=1 when something goes wrong

zatkins-dev: I think something got messed up, I didn't mean to change these

Makefile
```diff
@@ -262,7 +262,7 @@ CEED_LIBS = -lceed
 libceeds = $(libceed)
 BACKENDS_BUILTIN := /cpu/self/ref/serial /cpu/self/ref/blocked /cpu/self/opt/serial /cpu/self/opt/blocked
 BACKENDS_MAKE := $(BACKENDS_BUILTIN)
-
+pkgconf   = $(shell pkg-config $1 | $(SED) -e 's/^"//g' -e 's/"$$//g')
```
jeremylt, line 265: Is there commonly multiple possibilities here? For the PETSc examples the makefiles assume the presence of pkgconf

zatkins-dev: This is what we have in Ratel; I'm genuinely not sure what it's doing.

zatkins-dev: ah, it strips away leading or trailing double quotes from the pkg-config output

jeremylt: ah. Noether doesn't seem to be going for it

zatkins-dev: actually it was totally unrelated; my IDE decided to use tabs since the file mixes tabs and spaces. Make hates that and got very confused.

jeremylt: oh, fair. Yay for whitespace carrying meaning it shouldn't

zatkins-dev: yeah it's my least favorite thing about Make

zatkins-dev: huh no it was bc the XSMM pkgconfig file uses a colon in the lib string, which was interfering with sed substitution for creating the libCEED pkgconfig file. Now fixed.

jeremylt (approved):
looks right now - ping me in the (my) morning if there was another snag

jeremylt:
@zatkins-dev looks like this triggered a spack bug

zatkins-dev:
> @zatkins-dev looks like this triggered a spack bug

shoot, is the issue that we need to add pkgconfig as a dependency?

jeremylt:
yea - https://github.com/spack/spack-packages/pull/6527

PR #2041, opened 2026-09-18
cpu(gen) - fix for at-points 

zatkins-dev:
Purpose:

- initialize output q vectors with zeros when at-points, since some entries will not be written to if there are different numbers of points per cell
- fix bug in grad notranspose 1d where the wrong matrix was used in a tensor contraction

LLM/GenAI Disclosure:

None


zatkins-dev:
Just a note: we should probably have some tests which sanity check outputs against the `/cpu/self/ref` backend. This went unnoticed because we're effectively testing the implementation against itself.

jeremylt (approved):

jeremylt:
> Just a note: we should probably have some tests which sanity check outputs against the `/cpu/self/ref` backend. This went unnoticed because we're effectively testing the implementation against itself.

Yea, I've thought about that. Also, I think a transient failure has always been a real bug, so we should treat them as such in the future.

jeremylt (approved):

PR #2043, opened 2026-09-19
fix(install): add missing  `ceed/macros.h` to install target

### LINKED ISSUE #2042: install: missing `ceed/macros.h` in install target causes `ceed/backend.h` to fail
Hi!

When trying to compile anything that includes `<ceed/backend.h>` I get:

```
/usr/include/ceed/backend.h:13:10: fatal error: ceed/macros.h: No such file or directory
   13 | #include <ceed/macros.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated
```

This is because the installation target in the Makefile doesn't include `macros.h`:

https://github.com/CEED/libCEED/blob/c7a728a0a0b818b2f093e0cde71e90924f56bd01/Makefile#L995-L1007

## Re-producer:

1. Add the source file `repro.cpp`:

```cpp
#include <ceed.h>
#include <ceed/backend.h>

int main() {
  Ceed ceed;
  CeedCall(CeedInit("/cpu/self", &ceed));
  CeedCall(CeedDestroy(&ceed));
  return 0;
}
```

2. Compile with: `g++ -c $(pkg-config --cflags) repro.cpp`
3. See whether it completes or not.

## System info

OS: ArchLinux x86_64
libCEED version: `v0.12.0-1475-gc7a728a0a`

NOTE: The repro code is from [Palace](https://github.com/awslabs/palace/blob/ae20d251e996e63a1f4d2c8172e311ca7be4f1c5/palace/cmake/PkgConfigHelpers.cmake#L43-L52), which is where I triggered this issue originally.

--- 

Let me know if you need anything else from me,

Thanks!

tskaar:
Purpose:

`ceed/backend.h` depends on `ceed/macros.h`, include `macros.h` to installation target to be able to use `ceed/backend.h`

Closes: #2042 

LLM/GenAI Disclosure:

LLMs were used to track down the original issue, however nothing in the PR itself is LLM/GenAI based.


jeremylt (approved):
Thanks for the fix of my oops!

PR #2045, opened 2026-09-21
basis - add collocated gradient to CeedBasis and cache its computation

zatkins-dev:
Purpose:

Additional `CeedBasis` interface cleanup and unified caching of objects used by multiple backends. Specifically:

- Rename `CeedBasisGetCollocatedGrad` to `CeedBasisGetCollocatedGrad1D` and change signature to take `const CeedScalar **` like other getters; the output array is now set to the internally cached collocated gradient matrix.
- Add `CeedBasisHasCollocatedGrad` to check whether collocated gradient is suppported for a `CeedBasis`
- Update backends which either (a) computed the collo grad, or (b) precomputed and stored the collo grad, to use the caching `CeedBasisGetCollocatedGrad1D` function

Note: Currently failing pedantic check due to empty struct, though it will no longer be empty after #2010. Not sure the best way to deal with that.

Describe the purpose of the PR here.

LLM/GenAI Disclosure:

None


include/ceed/ceed.h
```diff
@@ -353,6 +353,7 @@ CEED_EXTERN int  CeedBasisGetInterp(CeedBasis basis, const CeedScalar **interp);
 CEED_EXTERN int  CeedBasisGetInterp1D(CeedBasis basis, const CeedScalar **interp_1d);
 CEED_EXTERN int  CeedBasisGetGrad(CeedBasis basis, const CeedScalar **grad);
 CEED_EXTERN int  CeedBasisGetGrad1D(CeedBasis basis, const CeedScalar **grad_1d);
+CEED_EXTERN int  CeedBasisGetCollocatedGrad1D(CeedBasis basis, const CeedScalar **collocated_grad_1d);
```
jeremylt: Do users need this?

interface/ceed-basis.c
```diff
@@ -2487,6 +2485,49 @@ int CeedBasisGetGrad1D(CeedBasis basis, const CeedScalar **grad_1d) {
...
 
+/**
+  @brief Return collocated gradient matrix
+
+  @param[in]  basis              `CeedBasis`
+  @param[out] collocated_grad_1d Row-major (`Q_1d * Q_1d`) matrix expressing derivatives of basis functions at quadrature points
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref Backend
```
jeremylt, line 2496: If its backend then should be in the backend header

interface/ceed-fortran.c
```diff
@@ -565,6 +560,14 @@ CEED_EXTERN void fCeedBasisGetGrad1D(int *basis, CeedScalar *grad_1d, int64_t *o
   *offset = grad_1d_c - grad_1d;
 }
 
+#define fCeedBasisGetCollocatedGrad1D FORTRAN_NAME(ceedbasisgetcollocatedgrad1d, CEEDBASISGETCOLLOCATEDGRAD1D)
+CEED_EXTERN void fCeedBasisGetCollocatedGrad1D(int *basis, CeedScalar *collocated_grad_1d, int64_t *offset, int *err) {
+  const CeedScalar *colloc_grad_1d_c;
+
+  CeedCallFortran(CeedBasisGetCollocatedGrad1D(CeedBasis_dict[*basis], &colloc_grad_1d_c));
```
jeremylt, line 567: Honestly, some of these backend functions don't really need to be in the Fortran interface

jeremylt (commented):
Is there a way to mark that struct as nolint?

interface/ceed-basis.c
```diff
@@ -2487,6 +2485,49 @@ int CeedBasisGetGrad1D(CeedBasis basis, const CeedScalar **grad_1d) {
...
+**/
+int CeedBasisGetCollocatedGrad1D(CeedBasis basis, const CeedScalar **collocated_grad_1d) {
+  bool has_collo_grad;
+
+  CeedCall(CeedBasisHasCollocatedGrad(basis, &has_collo_grad));
+  CeedCheck(has_collo_grad, CeedBasisReturnCeed(basis), CEED_ERROR_MINOR, "CeedBasis does not support collocated gradient application");
```
jeremylt, line 2502: Should we just return NULL if this operation isn't supported? It would simplify the backend calling convention.

zatkins-dev: there's a few places where we have an extra condition (usually dim == 3) on using the collocated basis, so the HasCollocatedGrad function is needed. We error if the 1D functions are called on non tensor bases, so I feel like there's precedent for this approach

interface/ceed-fortran.c
zatkins-dev: it's used directly in a test, that's why it's there

include/ceed/ceed.h
zatkins-dev: i'd be fine moving to backend

interface/ceed-fortran.c
jeremylt: Yea, I don't think we need that test. For a while I was mirroring neraly everything through the language interfaces but we should do a scrub to ditch stuff that isn't intended to be for users

interface/ceed-basis.c
jeremylt: makes sense, and sounds more consistent with existing approach

interface/ceed-fortran.c
zatkins-dev: do we want to do that cleanup in a follow-up? I feel like there's quite a few things to change there

jeremylt: Yea, that's what I was thinking. Probably loop through each language wrapper and create a list of stuff that isn't intended to be user facing, compile those lists in an issue, then strip them out?

jeremylt (approved):

interface/ceed-fortran.c
zatkins-dev: See #2046, we can discuss more on this there.

PR #2047, opened 2026-09-21
interface - Refactor environment variables to a unified interface

### LINKED ISSUE #2040: Env Based Flags Shoud Be Interface Level
We currently have backends reading env level flags in some cases and the interface layer in others. It is the job of the interface layer to communicate with the users, not the backends.

zatkins-dev:
This is my rough take at an environment variable refactor. The main goal is to make a single source of truth for all environment variables with consistent interfaces for overriding values programmaticly.

- refactor to make environment variables owned by Ceed context
- automatically generate public getters and setters for each environment variable via preprocessor
- two types are supported: flag (bool) and string
- allows for default values and aliases (e.g. `CEED_DEBUG` or `DEBUG` or `DBG`)
- all environment variables live in a single file `include/ceed/ceed-env-list.h`
- getters and setters are generated using `include/ceed/ceed-env.h` and `interface/ceed-env.c`
- stored values are automatically placed into `Ceed_private` struct

I'm generally hesitant to use macros, but I think this is the easiest way to make adding environmental args in the future clear and consistent.

Purpose:

Describe the purpose of the PR here.

Closes: #2040 

LLM/GenAI Disclosure:

None


zatkins-dev:
Okay, finally passing tests. @jeremylt - i mostly want to know if this approach seems reasonable/any major changes you want to see

jeremylt (commented):
I like the general approach. Q - should the function to get the debug flag be the exception to the pattern and be a static inline function in a header since its shows up all over the place?

jeremylt:
Thinking, it could be that debug is handled as here but the debug printing bypasses this to access the field on the ceed struct directly since it will always be there. Its just all over the codebase so ideally we help the compiler make that call as fast as possible

zatkins-dev:
That's a fair point. I do like the ability to set debug mode in code though, but the perf is a valid reason to not allow. We can still use the env getter, just define a custom getter (like the error handler)

jeremylt:
yea, mirror the interface so it looks same to the backend/library code but custom guts to its not a performance bomb

backends/cpu-gen/ceed-cpu-compile.cpp
```diff
@@ -29,7 +29,6 @@
 
 #define CEED_QUOTE(name) #name
 #define CEED_STRINGIFY(macro) CEED_QUOTE(macro)
-const char *CeedJitOpt = CEED_STRINGIFY(CEED_CPU_JIT_OPT);
 const char *CeedJitCxx = CEED_STRINGIFY(CEED_CPU_JIT_CXX);
```
jeremylt: does this part need to go too?

zatkins-dev: No, we use the baked-in one as a fallback if the user-provided one fails

jeremylt: maybe a different name then to suggest the different value?

zatkins-dev:
> yea, mirror the interface so it looks same to the backend/library code but custom guts to its not a performance bomb

I'm also torn on where the "true" values should live. Like, are these application-wide settings, in which case the values should be mirrored across parent and delegate Ceed objects? Or should it be only propagated to children? I feel like only applying to the specific Ceed context isn't a good idea, at least the delegates should be set.

jeremylt:
could have the delegates recursively go up to the parent, except for debug which is copied on delegates?

interface/ceed.c
```diff
@@ -219,27 +219,61 @@ static int CeedDestroy_Object(CeedObject *ceed) {
...
 // LCOV_EXCL_START
-bool CeedDebugFlagEnv(void) { return getenv("CEED_DEBUG") || getenv("DEBUG") || getenv("DBG"); }
+/**
+  @brief Set the debug state for the `Ceed` context
+
+  @note Custom setter to enable compiler inlining of the getter.
+        Rather than setting only on the root `Ceed`, this setter propogates to all children of the root `Ceed`
+
+  @param[in]   ceed      `Ceed` context
+  @param[out]  is_debug  Pointer to store flag value
```
jeremylt: bad copy paste :)

jeremylt (commented):
minor, looks pretty close

zatkins-dev:
> I like the general approach. Q - should the function to get the debug flag be the exception to the pattern and be a static inline function in a header since its shows up all over the place?

Okay so diving into this a bit, as it is more complicated than I originally thought. Right now, `CeedDebugFlag()` is not inline (still may be inlined via link-time optimization). The main reason is that the inline function definition would need to be able to access the guts of the `Ceed` context, which isn't possible from the `ceed/backend.h` header. We could define it in the `ceed-impl.h` header, but then we're forced to include that in all of the backend code (which we're intentionally not doing).

My solution to this is:
1. Define `CeedDebugFlag()` as a macro in `ceed-impl.h` to manually inline. Since the guts of the `Ceed` context are visible for any file which has the macro defined, this will act as a drop-in replacement without issue. 
```c
#define CeedDebugFlag(ceed) (ceed->ceed_env_EnableDebug)
```
2. Keep the existing implementation of `CeedDebugFlag()` declared in `ceed/backend.h` and defined in `interface/ceed.c`. This will be used anywhere that doesn't include `ceed-impl.h` and will be available for external linkage via including the `ceed/backend.h` header. This has identical perf characteristics to the implementation on main. Note that we have to `#undef` the macro, since otherwise it will replace the function name. Redefining it after only affects the rest of `ceed.c`.
```c
#undef CeedDebugFlag
bool CeedDebugFlag(Ceed ceed) { return ceed->ceed_env_EnableDebug; }
#define CeedDebugFlag(ceed) (ceed->ceed_env_EnableDebug)
```

tl; dr: The implementation of `CeedDebugFlag()` in this PR should be at least as fast as the existing implementation, with minor added complexity in order to inline wherever possible.

jeremylt:
yea, that makes sense and I should have remembered that

.github/workflows/julia-test-with-style.yml
```diff
@@ -36,6 +36,11 @@ jobs:
...
           echo "libceed_path = \"$LIBCEED_LIB\"" >> test/Project.toml
-          [[ "$GITHUB_REF" =~ ^refs/(heads/release|tags/).* ]] || julia --project -e 'import Pkg; Pkg.test("LibCEED"; coverage=true, test_args=["--run-dev-tests"])'
-          git checkout test/Project.toml && julia --project -e 'import Pkg; Pkg.test("LibCEED")'
+          # Test using the built libraries
+          julia --project -e 'import Pkg; Pkg.test("LibCEED"; coverage=true, test_args=["--run-dev-tests"])'
+          git checkout test/Project.toml
+          if [[ "$GITHUB_REF" =~ ^refs/(heads/release|tags/).* ]]; then
+            # Test using built-in libCEED libraries, only do this for release
+            julia --project -e 'import Pkg; Pkg.test("LibCEED")'
+          fi
```
zatkins-dev, line 45: Possibly a controversial change, but why were we testing the Julia interface against the prebuilt binaries? I'm shocked that this didn't cause issues at any other point. IMO, we should only test the interface against the corresponding binary version.

zatkins-dev:
@jeremylt I think I'm happy with the current state. Let me know if there's anything else you think we should do with this.

.github/workflows/julia-test-with-style.yml
jeremylt: yea, that's actually something we want and is on purpose - its why there are two sets of tests for Julia

@pazer has more details but this lets us know how the current interface lines up with the pre built binaries (another reason to release really)

interface/ceed.c
```diff
@@ -219,27 +219,61 @@ static int CeedDestroy_Object(CeedObject *ceed) {
...
+        Rather than setting only on the root `Ceed`, this setter propogates to all children of the root `Ceed`
+
+  @param[in,out] ceed     `Ceed` context
+  @param[in]     is_debug New flag value to set
+
+  @return An error code: 0 - success, otherwise - failure
+
+  @ref User
+**/
+int CeedSetEnableDebug(Ceed ceed, bool is_debug) {
```
jeremylt, line 266: probably need to call this when creating a fallback Ceed?

julia/LibCEED.jl/test/runtests.jl
```diff
@@ -44,10 +44,14 @@ else
...
             @test get_preferred_memtype(c) == MEM_HOST
             @test_throws LibCEED.CeedError create_interior_qfunction(c, "")
-            @test showstr(c) == """
-                Ceed
-                  Ceed Resource: $res
-                  Preferred MemType: host"""
+            @test startswith(
+                showstr(c),
+                """
+Ceed
```
jeremylt, line 50: see, this should move to devtests to pass

.gitlab-ci.yml
```diff
@@ -205,6 +206,8 @@ noether-rust-qfunctions:
     #    Note: PETSC_DIR is set by default in GitLab runner env, unsetting to isolate core tests
     - export PETSC_DIR= PETSC_ARCH=
     - make -k -j$((NPROC_GPU / NPROC_POOL)) JUNIT_BATCH="rust-qfunction" junit search=rustqfunction
+    - export CEED_CUDA_CLANG_CXX=
```
jeremylt, line 209: maybe a note that this checks the default handling?

.github/workflows/julia-test-with-style.yml
zatkins-dev: alright -- i can rewrite the test such that it passes for both

jeremylt (commented):
minor

backends/cpu-gen/ceed-cpu-compile.cpp
zatkins-dev: changed to `CeedJitOptDefault`

interface/ceed.c
zatkins-dev: good call

.gitlab-ci.yml
zatkins-dev: yeah, also this wasn't doing exactly what I wanted. should be better now

julia/LibCEED.jl/test/runtests.jl
zatkins-dev: eh, probably. Outside of scope here though.

jeremylt: No, I mean that's how you get this to pass without breaking anything. When you change the Julia interface between releases, the affected tests moves from this file to the devtests file. That's what the split is for

zatkins-dev:
Okay, I've got the requested feature (get whether an environment variable was set) working. The number of loose values in the `Ceed_private` struct was getting a little silly, so I grouped the state for each variable into its own struct.

interface/ceed.c
zatkins-dev: I'm actually not sure that it's needed with our current conventions of setting fallbacks/delegates in the Init functions for the backend, as the user cannot have called `CeedSetEnableDebug` on the new `Ceed` before its delegate/fallbacks are created. Still, good to be safe.

julia/LibCEED.jl/test/runtests.jl
zatkins-dev: ohhh I see

zatkins-dev: Still -- I don't actually want to check the full string since it depends on the current environment. I think the current change makes sense. I can move to dev tests still if you want.

interface/ceed.c
jeremylt: ahh, I think I used to create these lazily but switched because of exactly this sort of thing

julia/LibCEED.jl/test/runtests.jl
jeremylt: fair - I mostly wanted to make sure that you knew what was going on here since its a sharp corner I've cut myself on

zatkins-dev: I appreciate, I had never dealt with them

include/ceed-impl.h
```diff
@@ -98,6 +98,21 @@ typedef struct CeedObject_private {
   CeedInt num_view_tabs;
 } CeedObject_private;
 
+typedef struct {
+  bool value;
+  bool set_by_user;
+  bool set_in_environment;
+  bool checked_environment;
```
jeremylt: minor - should it be is_? I think we usually start with is/has/other small verb whose class name I'm forgetting?

whatever is fine because I'm talking average gravel diameter for path leading to bike shed here

jeremylt (approved):
thanks for this nice cleanup

include/ceed-impl.h
zatkins-dev: haha perhaps, it just felt really long and like the meaning was pretty obvious.

zatkins-dev: fixed now!

backends/cpu-gen/ceed-cpu-compile.cpp
```diff
@@ -186,7 +190,8 @@ static inline int CeedCompileCore_Cpu(Ceed ceed, const char *source, const char
 
     // First check for user JiT compiler
     if (!cxx) {
-      const char *user_cxx = getenv("CEED_CPU_JIT_CXX");
+      const char *user_cxx;
+      CeedCall(CeedGetCpuJitCxx(ceed, &user_cxx));
```
jeremylt, line 194:
```suggestion
      const char *user_cxx;

      CeedCall(CeedGetCpuJitCxx(ceed, &user_cxx));
```
minor!

backends/cpu-gen/ceed-cpu-compile.cpp
jeremylt, line 198:
```suggestion
      bool is_valid = false;

```

backends/cpu-gen/ceed-cpu-compile.cpp
```diff
@@ -210,18 +215,18 @@ static inline int CeedCompileCore_Cpu(Ceed ceed, const char *source, const char
     if (!cxx) {
       bool is_valid = false;
 
-      CeedDebug(ceed, "Default JiT compiler: %s\n", CeedJitCxx);
+      CeedDebug(ceed, "Default JiT compiler: %s\n", CeedJitCxxDefault);
```
jeremylt, line 219:
```suggestion
      CeedDebug(ceed, "Default JiT compiler: %s\n", CeedJitCxxDefault);
```

backends/cuda/ceed-cuda-compile.cpp
```diff
@@ -190,9 +190,12 @@ static int CeedCompileCore_Cuda(Ceed ceed, const char *source, const char *name,
   CeedDebug256(ceed, CEED_DEBUG_COLOR_SUCCESS, "---------- END OF JIT SOURCE ----------\n");
 
   if (!using_clang) {
+    bool debug;
+
+    CeedCallBackend(CeedGetEnableDebug(ceed, &debug));
     CeedCallNvrtc(ceed, nvrtcCreateProgram(&prog, code.str().c_str(), NULL, 0, NULL, NULL));
 
-    if (CeedDebugFlag(ceed)) {
+    if (debug) {
```
jeremylt, line 198:
```suggestion
CeedCallNvrtc(ceed, nvrtcCreateProgram(&prog, code.str().c_str(), NULL, 0, NULL, NULL));

    if (CeedDebugFlag(ceed)) {
```
this looks like what you switched back to elsewhere?

zatkins-dev: good call, i missed that one

backends/cpu-gen/ceed-cpu-compile.cpp
zatkins-dev: I'm really struggling to see the difference

jeremylt: ugh, github ate the change - I was deleting the unneeded blank line, I think?

jeremylt (approved):

jeremylt:
Should we add a CI job to test the Spack install?

zatkins-dev:
> Should we add a CI job to test the Spack install?

Yes, tho wrong PR I think

jeremylt:
oh, absolutely, I just thought of it here because include files are a common failure point for our install process

